PR #16, opened 2018-04-04
fetching a query string returns maybe(string)

CrowdHailer:
I may have misunderstood what's going on here but I think this is an improvement


lpil:
Ah, thank you!

PR #19, opened 2018-06-29
Brace based syntax

lpil:


rizo:
Now gleam will become mainstream.

lpil:
I've learnt from Reason!

Issue #36, opened 2018-10-12
Replace => with the more Erlang-aligned ->


hengestone:
Just a reminder issue to discuss using `=>` for functions vs `->`, which is more Erlang-aligned.

hengestone:
Thanks for the merge - you beat me to the punch changing the example files on Saturday :)

lpil:
I got a little excited and did it myself :P

PR #37, opened 2018-10-13
Naive search and replace of => to -> change

hengestone:
Attempt at implementing #36 
EUnit tests pass FWIW!


lpil:
Great, thank you for this.

Could you also update the syntax in the .gleam files? A search and replace should work fine as the => token has been removed entirely

PR #50, opened 2018-11-07
rebar3 plugin

lpil:


tsloughter:
Note that there is a new and far better interface for dealing with code paths and compiler in 3.7.0. We kept the old in there with a deprecated so this will still work, but you'll want to move to the new stuff. Sadly haven't documented it yet... But if you look at the new compile modules, like `rebar_compile_mib`, they are pretty small. And look at the use of the new module `rebar_paths`.

lpil:
Thanks for the tip @tsloughter. Is there a particular example I should be looking at?

lpil:
@tsloughter Could we get the callbacks in the behaviour documented/commented? It's unclear to me what the arguments are returns values are, and the examples don't make use of all of them in an obvious way.

tsloughter:
Yea. I'll get around to that soon, at least next week.

lpil:
Thank you. Has rebar 3.7.0 been released? I've run `rebar3 local upgrade` and I'm still on 3.6.2

tsloughter:
Na, it is still 3.7.0-rc2. Hopefully soon.

lpil:
Thanks again, I'll use master

PR #85, opened 2019-03-02
Fix spelling

jwinnie:
Principal (first or highest in rank, importance, value, etc.; chief; foremost) vs. Principle (a fundamental, primary, or general law or truth from which others are derived)

_Side note:_
I'm excited about Gleam! I appreciate the power and abstraction of Haskell, but it has too high of a learning curve and its ML-style syntax is alien to all non-ML programmers. I like Rust's syntax, but the borrow checker & manual memory management annoy me very much. I like Elixir's powerful Actor model, the cool features of the Erlang VM (e.g. hot swapping) and its unique combination of functional & imperative programming, but it is weakly typed, which is a major weakness in a modern programming language.

I think Gleam might be that "just-right" language: statically-typed with a powerful type system, functional but flexible enough, a friendly and easy-to-learn syntax, and concurrent to the core!

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ pub fn has_even_leaf(tree) {
 }
 ```
 
-## Principals
+## Principles
 
 ### Be safe
 


lpil:
Thank you! I am rather rubbish at spelling.

Also thank you for your enthusiasm, it sounds like you also want the language that I want. Hopefully we'll get what we want :)

OvermindDL1:
For note, Rust doesn't use manual memory management, at no point do you free anything manually, that is why the borrow checker exists, not just to manage memory resources but to manage *any* resource (unlike garbage collectors, which are memory only, and quite inefficient at that).

lpil:
I like to think it as static garbage collection :)

OvermindDL1:
> I like to think it as static garbage collection :)

Lol, that's a very good way of thinking about it, the compiler knows about the lifespans statically at compiletime so it can control the resource management perfectly.  ^.^

jwinnie:
I'm no computer scientist, just a Python programmer :smile:

What I mean is that I love Rust, but, in my opinion, it is too low level for most tasks. To make a working program, I have to worry about unwrapping ten million Results and Options, fighting the lifetime/borrow checker, etc., etc. I want my programming language to handle memory and errors automatically without me having to think about it, so I can think about the problems that actually matter for my application.

OvermindDL1:
>  fighting the lifetime/borrow checker,

To be honest though, if fighting the borrow checker then the code was broken to begin with, it is a "Good Thing" it is catching those faults.  ^.^

As for the many nested unwrap's, that sounds like an exceptionally poor API design, what library was that from?!  o.O

PR #137, opened 2019-04-05
Implement str:parse_int and str:parse_float

MainShayne233:
Issue: https://github.com/lpil/gleam/issues/116

I have no idea if this approach is desired, but thought I'd take a stab at it using pre-existing code as a rough guideline. Is this along the right track, or are we going for something different? (maybe pure-gleam implementation)?

Also, have no idea how to verify that it works :p. Was wondering if you could point me in the right direction there.

If green-lit, I'll go ahead w/ `str:parse_float` in the desired method.

### DIFF
--- a/gleam_stdlib/src/str.gleam
+++ b/gleam_stdlib/src/str.gleam
@@ -121,6 +121,34 @@ test parse_int {
   |> expect:equal(_, Error(ParseError))
 }
 
+pub external fn parse_float(String) -> Result(Float, ParseError) = "gleam__stdlib" "parse_float";
+
+test parse_float {
+  "1.23"
+  |> parse_float
+  |> expect:equal(_, Ok(1.23))
+
+  "5.0"
+  |> parse_float
+  |> expect:equal(_, Ok(5.0))
+
+  "0.123456789"
+  |> parse_float
+  |> expect:equal(_, Ok(0.123456789))
+
+  ""
+  |> parse_float
+  |> expect:equal(_, Error(ParseError))
+
+  "what"
+  |> parse_float
+  |> expect:equal(_, Error(ParseError))
+
+  "1"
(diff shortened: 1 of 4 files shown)


lpil:
I think wrapping the Erlang function here is the way to go, at least for now.

> have no idea how to verify that it works :p. Was wondering if you could point me in the right direction there.

Good point, this should be documented. Have you run the tests? To do so run `rebar3 eunit` in the `gleam_stdlib` directory.

gleam_stdlib/src/str.gleam
```diff
@@ -90,6 +93,22 @@ test from_int {
...
+  |> parse_int
+  |> expect:equal(_, Ok(123))
+
+  "-123"
+  |> parse_int
+  |> expect:equal(_, Ok(-123))
+
+  "0123"
+  |> parse_int
+  |> expect:equal(_, Ok(123))
```
lpil: Can we have some failing cases as well please?

MainShayne233: Absolutely!

MainShayne233:
@lpil Awesome! Thanks for the info! I'll expand the tests, write the float implementation, and document how to run tests/work on the project where it seems useful :)

MainShayne233:
@lpil Got tests passing for current changes! :)

Since I didn't feel at liberty to make certain decisions, like what the return type should be, if an integer string should be able to be parsed as a float, etc, I just decided to roll w/ the erlang functions' behaviors and to just have the very simple `Result(Type, ParseError)` return type.

lpil:
This looks great!

Would you like to rebase this into 2 commits (one for each function)? If not I'll squash and merge this in :)

lpil:
I've rebased these commits into master. Thank you! <3

PR #149, opened 2019-04-18
stdlib updates

rawburt:
Add `list:index_map`

https://github.com/lpil/gleam/issues/143

Add `str:concat`

### DIFF
--- a/gleam_stdlib/src/list.gleam
+++ b/gleam_stdlib/src/list.gleam
@@ -54,15 +54,19 @@ pub fn filter(list, fun) {
   do_filter(list, fun, [])
 }
 
-fn do_map(list, fun, acc) {
+fn do_map_with(list, f, acc, map_f) {
   case list {
   | [] -> reverse(acc)
-  | [x | xs] -> do_map(xs, fun, [fun(x) | acc])
+  | [x | xs] -> do_map_with(xs, f, [map_f(f, acc, x) | acc], map_f)
   }
 }
 
-pub fn map(list, fun) {
-  do_map(list, fun, [])
+pub fn map(list, f) {
+  do_map_with(list, f, [], fn(fun, _, x) { fun(x) })
+}
+
+pub fn index_map(list, f) {
+  do_map_with(list, f, [], fn(fun, a, x) { fun(length(a), x) })
 }
 
 fn do_traverse(list, fun, acc) {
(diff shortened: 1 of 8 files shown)


gleam_stdlib/test/list_test.gleam
```diff
@@ -248,3 +250,14 @@ pub fn sort_test() {
   list:sort([{1,2}, {4,5}, {3,2}])
   |> expect:equal(_, [{1,2}, {3,2}, {4,5}])
 }
+
+pub fn index_map_test() {
+  list:index_map([3,4,5], fn(i, x) { {i, x} })
+  |> expect:equal(_, [{0,3},{1,4},{2,5}])
+
+  let f = fn(i, x) {
+    iodata:new(x) |> iodata:append(_, str:from_int(i)) |> iodata:to_string(_)
```
rawburt: Is there a better way to do this? I'm about to do something like `str:join` but maybe I'm missing something obvious about the Gleam/Erlang world

gleam_stdlib/src/list.gleam
```diff
@@ -54,15 +54,19 @@ pub fn filter(list, fun) {
...
 }
 
-pub fn map(list, fun) {
-  do_map(list, fun, [])
+pub fn map(list, f) {
+  do_map_with(list, f, [], fn(fun, _, x) { fun(x) })
+}
+
+pub fn index_map(list, f) {
+  do_map_with(list, f, [], fn(fun, a, x) { fun(length(a), x) })
```
lpil: The Gleam and Erlang compilers are not clever enough to optimise the call to `length` so the accumulator will be traversed on each iteration making this `O(n log n)` I'm afraid. We could get it down to `O(n)` using an implementation something like this:

```rust
fn do_index_map(list, fun, index, acc) {
  case list {
  | [] -> reverse(acc)
  | [x | xs] -> do_index_map(xs, fun, index + 1, [fun(index, x) | acc])
  }
}

pub fn index_map(list, fun) {
  do_index_map(list, fun, 0, [])
}
```

gleam_stdlib/src/str.gleam
```diff
@@ -34,6 +34,10 @@ pub fn replace(string, pattern, with) {
   |> iodata:to_string
 }
 
+pub fn concat(s1, s2) {
+  iodata:new(s1) |> iodata:append(_, s2) |> iodata:to_string(_)
```
lpil: Can we call this `append`? To match `list:append` and `iodata:append`.

lpil:
Mind updating the CHANGELOG as well? :)

PR #168, opened 2019-05-10
list:strict_zip fix test

bsnyder788:


bsnyder788:
@lpil, that test was actually not calling the strict method

PR #174, opened 2019-05-11
list:repeat

bsnyder788:


gleam_stdlib/src/list.gleam
```diff
@@ -263,3 +263,12 @@ pub fn range(start, stop) {
   | order:Lt -> [start | range(start + 1, stop)]
   }
 }
+
+external fn duplicate(t, a) -> List(a) = "lists" "duplicate"
+
+pub fn repeat(a, times) {
+  case times <= 0 {
+    | True -> []
+    | False -> duplicate(times, a)
```
lpil: Can we implement this in Gleam rather than calling out to Erlang? :)

bsnyder788: sure! wasn't sure what the policy was on using Erlang for some of these

gleam_stdlib/test/list_test.gleam
```diff
@@ -298,3 +298,17 @@ pub fn range_test() {
...
+  |> expect:equal(_, [])
+
+  list:repeat(1, 0)
+  |> expect:equal(_, [])
+
+  list:repeat(2, 3)
+  |> expect:equal(_, [2, 2, 2])
+
+  list:repeat("x", 5)
+  |> expect:equal(_, ["x","x","x","x","x"])
```
lpil: Could we have spaces after the `,`s please? Thank you

bsnyder788: fixed!

gleam_stdlib/src/list.gleam
bsnyder788: reimplemented in Gleam :)

lpil:
I generally think we should implement things in Gleam unless in Erlang it is implemented using a BIF or syntax that Gleam does not support. :)

PR #179, opened 2019-05-14
map_dict:fold

bsnyder788:


gleam_stdlib/src/map_dict.gleam
```diff
@@ -86,3 +86,11 @@ pub fn update(dict, key, f) {
   | Error(_) -> put(dict, key, f(Error(NotFound)))
   }
 }
+
+pub fn fold(dict, acc, f) {
+  let kvs = to_list(dict)
```
bsnyder788: I can use another function, i.e. `do_fold(list, acc, f)` to prevent doing this list conversion and map deletion in the  case each time it recurses. Is pattern matching on maps a planned future feature @lpil ?

lpil: It's not a planned feature right now. We could wrap Erlang's maps:fold? Or we could convert to list and use `list:fold` ? The former would probably be faster but I've not tested this in any way.

bsnyder788: I pushed a 2nd commit with the optimization I referenced for the Gleam version. We could definitely wrap the Erlang `map:fold`. I don't think we can do anything with `list:fold` though since the arity is different since we have key, value, and acc in the params for the folding function for maps.

lpil:
Thanks

PR #187, opened 2019-06-29
Release automation

lpil:
Closes https://github.com/lpil/gleam/issues/90

### DIFF
--- a/bin/build-linux-release.sh
+++ b/bin/build-linux-release.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -eu
+
+GLEAM_ROOT=$(pwd)
+TAG=$(git tag --points-at HEAD)
+DOCKER_TAG=lpil/gleam:${TAG:1}
+ARCHIVE=gleam-$TAG-linux-amd64.tar.gz
+CONTAINER_NAME=gleam-linux-builder
+
+cd gleam
+docker build . -t $DOCKER_TAG
+docker push $DOCKER_TAG
+
+TMP=$(mktemp -d)
+cd $TMP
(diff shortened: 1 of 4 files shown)


NobbZ:
Will you also provide prebuilts for MUSL, or only the dockerfile?

lpil:
The prebuild Linux binary in the release page is statically linked against MUSL so should work on any AMD64 Linux distribution :)

NobbZ:
Will you in future always provide the precompiled binary with every tag?

I could probably cut the build times for the AUR package drastically if you guarantee availability of the binary.

lpil:
Yes, from here on out I intend to. In future I'll probably have a build server run these scripts when a new tag is pushed

NobbZ:
Move the project to gitlab and I could help you set it up ;)

But I'm sure it will be possible using circle-ci or travis-ci as well.

lpil:
The limiting factor is mostly macOS and the lack of a build server for it. When Gleam gets bigger I may contact some CI providers such circle to see if they will offer up some free macOS build time.

NobbZ:
Mac OS isn't that important ;)

But yes, I fully understand the considerations… And I assume, the same is true for windows?

lpil:
Lots of people use it, it's plenty important :)

I'd like to provide these for Windows as well but I don't have a Windows computer or any Windows experience, so it's even more difficult.

NobbZ:
I've one available at office… And at least for CI to run tests and to compilation to download the artifact after the fact should be doable. I know there are CI services available that offer CI time on windows for free.

lpil:
Good point! I'm sure we can find what we need somewhere. I'll open an issue to track this.

PR #191, opened 2019-07-01
Adds upgrade target to makefile

AdamBrodzinski:
When trying out Gleam I ran into several issues trying to upgrade to a stable commit because it was just using master which happened to break something and my rebar_gleam didn't update itself.

I made this makefile target for my own use to help out with this upgrade process and I wanted to show it to you in case you thought it's something that would be useful for an interim solution.

Also i'm not sure if it would be overkill to make a `fetch-version` target and a `upgrade-rebar-gleam` target and then calling `upgrade: fetch-version install upgrade-rebar` so that you don't have to call `make install`.

Please feel free to suggest edits or close the PR all together as I realize it may not be a fit for general use.

Thanks and keep up the good work!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,14 @@ build: book gleam gleam_stdlib/gen ## Build all targets
 install: ## Build the Gleam compiler and place it on PATH
 	cd gleam && cargo install --path . --force
 
+.PHONY: upgrade
+upgrade: ## Fetch by version tag using arg version=v0.x.x and install
+	git fetch &&\
+	git fetch --tags &&\
+	git checkout v$(version) &&\
+	make install &&\
+	rebar3 as global plugins upgrade rebar_gleam
+
 .PHONY: help
 help:
 	@cat $(MAKEFILE_LIST) | grep -E '^[a-zA-Z_-]+:.*?## .*$$' | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'


lpil:
Hey! Thanks for this!

While the new asdf plugin (https://github.com/vic/asdf-gleam) makes this less useful but perhaps this could be beneficial.

I'm not sure about the rebar3 update- there's nothing linking the rebar3 plugin version to the Gleam version so unless you're building the latest tag you'll get the wrong version. I think we'd need to remove this. Would it still be useful without the handling of the rebar3 plugin?

AdamBrodzinski:
Oh wow I didn't realize there was an asdf plugin yet! I was thinking the makefile target would be a crutch until one existed.

Great point on the version of the rebar_gleam version. Since an asdf plugin exists I think it's best to close this PR. At best I think a `upgrade_latest_rebar_gleam` target makes sense but is not worth it.

IMHO it feels a bit cumbersome to use rebar (and update the plugin) to generate a Gleam project. Would you consider accepting a PR that has a file generator in the Gleam (Rust) CLI? I'd be happy to add something like `gleam new my_amazing_application ` that spits out the templates in a new folder. I've made several of these before from scratch (but not in Rust yet).

lpil:
You make a good point, currently the rebar3 plugin serves no purpose other than to scaffold the project- it would be easier for users if they didn't have to also manage this plugin.

Would you mind opening an issue to track this feature?

For templating I am a fan of the askama library.
https://github.com/djc/askama

AdamBrodzinski:
Askama looks great! 👍  I'll take a stab at this next weekend.
Closing this PR in favor of #192 .

lpil:
Great! Rust can be quite fiddly at first so feel free to open a work in progress PR if you get checker fighting the borrow checker and want a stone pair of eyes

PR #200, opened 2019-08-01
Anonymous structs

lpil:


OvermindDL1:
These don't seem to be structs, they seem to be Tuples?  Structs/records are generally named tuples, tuples are unnamed (positional) structs/records.

lpil:
There will also be named structs to go with these :)

OvermindDL1:
In that case what about combining both tuples and records into a single 'struct' syntax, like how Rust does?  Tuple structs and named structs (distinguished by using either `()` or `{}`)?

lpil:
In Rust the syntax is

```rust
(1, 2)                           // tuple
MyStruct { first: 1, second: 2 } // struct
```

In Gleam the syntax may be 

```rust
struct(1, 2)                  // anonymous struct
MyStruct(first: 1, second: 2) // named struct
```

I can't use the `(1, 2)` syntax for function calls as it's ambiguous in Gleam because we use `(1, 2)` for function calls and don't have semicolons for delimiting statements.

OvermindDL1:
> In Rust the syntax is

Huh, for basic tuples sure, but I'm talking about tuple structs, in Rust:
```
// Named struct definition:
struct Blah {
  a: i32,
  f: f32,
}
// Creation:
let b = Blah { a: 42, f: 6.28 }
// Access:
b.a
b.f

// Tuple struct definition:
struct Blah(i32, f32)
// Creation
let b = Blah(42, 6.28)
// Access:
b.0
b.1
```

lpil:
Ah right! I've not decided whether Gleam will have tuple structs or not. More thought needs to be done :)

PR #205, opened 2019-08-13
Named structs

lpil:


gleam/src/typ.rs
```diff
@@ -557,6 +557,26 @@ pub enum ValueConstructorVariant {
...
+            ValueConstructorVariant::ModuleFn { .. } => ModuleValueConstructor::Fn,
+            ValueConstructorVariant::NamedStruct { .. } => unimplemented!(),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum ModuleValueConstructor {
+    Enum,
+    Fn,
```
lpil: We need a test for modules that export structs to show that they are rendered correctly when used in other modules (i.e. not as an enum or fn)

PR #207, opened 2019-08-14
Make >, >=, <, <= operate only on Ints

### LINKED ISSUE #171: Make >, >=, <, <= operate only on Ints
Order operators current behaviour can be confusing as the automatic ordering may not align with user expectations:

```rust
enum Order =
  | Gt
  | Eq
  | Lt
```
```rust
Gt > Lt  // => False
```

Restrict them to Ints only, and introduce `>.` etc that operate on Floats.

dirk:
Restrict the comparison operators to only operate on integers to prevent confusion and/or unexpected results when values of different types are compared.

Fixes #171

### DIFF
--- a/gleam/src/typ.rs
+++ b/gleam/src/typ.rs
@@ -958,42 +958,38 @@ impl<'a> Env<'a> {
             },
         );
 
-        let a = env.new_generic_var();
         env.insert_variable(
             ">".to_string(),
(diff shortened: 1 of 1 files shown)


lpil:
If you'd like to try another issue there is for Floats https://github.com/lpil/gleam/issues/208

PR #214, opened 2019-08-15
Add bin-ops and float:compare for comparing Floats

### LINKED ISSUE #208: Create comparison operators for Floats
After #171 the comparison operators only operate on Ints. A new set of operators for Floats would be convenient.

`>.` `>=.` `<.` `<=.`

### LINKED ISSUE #209: float:compare function
To compliment `int:compare`.

Requires #208 to be implemented first.

dirk:
- Adds the `>.` `>=.` `<.` `<=.` binary operators for comparing Floats (fixes #208).
- Adds the `float:compare` function to the stdlib (fixes #209).


gleam/src/erl.rs
```diff
@@ -1674,6 +1674,18 @@ x() ->
     'try':'and'().
 "#,
         },
+        // Translation of Float-specific BinOp into variable-type Erlang term comparison.
+        Case {
+            src: r#"fn x() { 1. <. 2.3 }"#,
+            erl: r#"-module().
+-compile(no_auto_import).
+
+-export([]).
```
lpil: I think this will fail now because of the change to exports, could you rebase on master? Thank you

gleam_stdlib/test/gleam/float_test.gleam
```diff
@@ -37,6 +38,26 @@ pub fn to_string_test() {
   |> expect:equal(_, "-8.1")
 }
 
+pub fn compare_test() {
+  float:compare(0., 0.)
```
lpil: Heh, I totally forgot this syntax is valid.

dirk:
@lpil: Fixed up the integration test to comply with #215. 🙃

PR #215, opened 2019-08-15
Don't emit export directive if there's nothing to export

dirk:
Not sure if this is worth the added complexity, but it does result in more concise generated code.


lpil:
It seems I misconfigured CI to not build for PRs, it should work for future ones.

PR #238, opened 2019-09-16
Disallow positional arguments after labelled arguments

lpil:


OvermindDL1:
Why disallow?  OCaml allows them to be mixed in a definition and thus mixed in the actual place.  However, OCaml only allows labelled arguments to be mixed together that are not separated by positional, once a positional is specified then labelled that are defined 'before' it cannot be further added, though ones immediately after it are now allowed where they were not before.  It allows for a very fluid API.

lpil:
I want to make the API strict to start, we can relax it later if that becomes useful :)

PR #262, opened 2019-09-30
Implement string:compare

MainShayne233:
An attempt at implementing: https://github.com/lpil/gleam/issues/216; basically just leans on erlang's rules for string comparison.

I wasn't sure if there was a way to cleanly return something from `erlang` that'd resolve as an enum-variant in `gleam`, and also couldn't find a way to refer to atoms, so I went w/ integers that map to the `order` enum variants.


gleam_stdlib/src/gleam_stdlib.erl
```diff
@@ -90,3 +90,13 @@ parse_float(String) ->
...
   end.
+
+compare_strings(Lhs, Rhs) ->
+  if
+    Lhs == Rhs ->
+      0;
+    Lhs < Rhs ->
+      -1;
+    true ->
+      1
```
lpil: Looks good!

We could return `lt` `eq` and `gt` rather than ints here to avoid the extra conversion on the Gleam side.

MainShayne233:
@lpil done!

PR #268, opened 2019-10-02
add min() for Int and Float modules

bsnyder788:


gleam_stdlib/src/gleam/int.gleam
```diff
@@ -16,3 +16,10 @@ pub fn compare(a, b) {
     }
   }
 }
+
+pub fn min(a, b) {
+  case a < b {
+  | True -> a
+  | False -> b
+  }
+}
```
lpil: Could we have a newline at the end of this file please?

gleam_stdlib/test/gleam/float_test.gleam
```diff
@@ -137,3 +137,26 @@ pub fn truncate_test() {
   |> float.truncate
   |> expect.equal(_, -7)
 }
+
+pub fn min_test() {
+    float.min(0., 0.)
```
lpil: Can you indent with 2 spaces here and in the int test file please? Thank you

gleam_stdlib/src/gleam/int.gleam
bsnyder788: done

gleam_stdlib/test/gleam/float_test.gleam
bsnyder788: done

PR #270, opened 2019-10-04
Add `brew install gleam` note

chenrui333:
Relates to https://github.com/Homebrew/homebrew-core/pull/44466

The formula has been introduced into homebrew-core, so it would be nice to add instructions for using it. :)


chenrui333:
Also keeping it up to date in this PR, https://github.com/Homebrew/homebrew-core/pull/44914. 😄

PR #273, opened 2019-10-09
[WIP] Update 'stdlib' in document

djKooks:
Hello,
I'm planning to update document for add on standard library guide.

I've update very few module, so please check if format or contents are OK.
If it is, I'll keep working on other libraries.

Thanks.

### DIFF
--- a/book/src/tour/string.md
+++ b/book/src/tour/string.md
@@ -18,3 +18,22 @@ Special characters such as `"` need to be escaped with a `\` character.
 ```rust,noplaypen
 "Here is a double quote -> \" <-"
 ```
+
+## Standard library
+
+To use standard library for 'string', add:
+```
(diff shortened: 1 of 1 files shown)


lpil:
Hi, thanks for the contribution!

I think you're very right in that we need to document the standard library, however I want to keep the tour very simple right now. I don't want to introduce concepts like `import` earlier in the tour.

If you are interested in writing Rust then perhaps it would be good to take a look at https://github.com/lpil/gleam/issues/153 as this will help with learning the standard library.

djKooks:
@lpil understood~ it seems better way to go on like you mentioned. I'll close this one.

lpil:
Adding comments to the Gleam code in `gleam_stdlib` is a great idea!

Issue #232, opened 2019-09-08
Switch CI to GitHub actions


lpil:
GitHub actions offers Windows, OSX, and Linux build machines with less faff than CircleCI and a better integrated experience, so let's try switching.

NobbZ:
GitHub Actions currently do not support caching. So especially if you spend a lot of time on rebuilding things that you could cache on Circle/Travis easily, GHA will cause a lot of slowlyness.

Due to constantly recompiling dependencies, we always had build times of a cold cache, which on travis only happened for builds on master, and there we did this on purpose.

We've decided to stick to travis for that project for now, but we will keep GHA in our mind and watch it closely.

lpil:
This is a concern of mine but after seeing a few projects switch from Circle to GHA I think it may still be worth trying.

If the performance is too poor we can remain on Circle, but given we have very few PRs being opened here speed is less important.

NobbZ:
Yeah, not having caching is fine for some, but we faced build times of 30+ minutes while we had 3 to 5 when the cache was hot… So this was a major issue.

lpil:
Caching is expected to come in November it seems https://stackoverflow.com/questions/55110729/how-do-i-cache-steps-in-github-actions/57958803#57958803

EskiMag:
This is something I would like to do when the GitHub Actions are not beta anymore or if the feature will be available sooner.

lpil:
I'm happy to start work without the caching for now :)

PR #281, opened 2019-10-21
Create simple GitHub Actions workflow.

EskiMag:
This Pull Request is related to issue #232.

I created a simple GitHub Actions workflow which replicates current Circle CI build process. It doesn't use cache, since that is not available in GitHub Actions yet.

What should be the output of this workflow? Is the successfull run of all the commands good enough enough, or should this workflow produce some artifacts or a GitHub Release?

GitHub Actions supports MacOS runners too, so it is possible to even do a release after a new tag is created in the repository along with the linux and macos binaries. I assume you're doing this manually now. Am I right?

### DIFF
--- a/.github/workflows/build_compiler.yaml
+++ b/.github/workflows/build_compiler.yaml
@@ -1,0 +1,14 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build_compiler:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v1
+      - run: cd gleam && cargo test
+      - run: cd gleam && cargo build --release
+      - run: gleam/target/release/gleam --version
+      - run: gleam/target/release/gleam build gleam_stdlib


lpil:
Hi! Thanks for this!

Yes, you're right about the manual release. I would like to make the release process automated later.

How does this know to use an image that has cargo installed?

lpil:
Do you know if there is a way to test this on a branch? I've not managed to figure it out yet!

EskiMag:
Here's a list of what's included in the GitHub virtual environment: https://help.github.com/en/github/automating-your-workflow-with-github-actions/software-in-virtual-environments-for-github-actions#ubuntu-1804-lts

I specified `ubuntu-latest` which at this moment refers to `ubuntu-18.04`. More info is here: https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idruns-on

Currently in `ubuntu-latest`, there's:
* rustup (1.19.0)
* rust (1.37.0)
* cargo (1.37.0)
* rustfmt (1.3.0-stable)
* clippy (0.0.212)
* rustdoc (1.37.0)

Now I realize that maybe using a containerized job would be a better idea. That way we can specify  exactly what will be used. More info here: https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#jobsjob_idcontainer
If there's no container specified, the workflow will run on a host directly - this is the currently proposed way, which will be changed I will change in a moment.

I am going to update the wofklow to use the same container as is used by Circle CI at this moment.

I don't know how to test this on a branch yet. I tried to find this information, but with no luck.

lpil:
> I don't know how to test this on a branch yet. I tried to find this information, but with no luck.

I'll try merging this and see how that goes :)

EskiMag:
OK, it went well. https://github.com/lpil/gleam/commit/c1fd94f6dc3b765bc02f34e8c58762ec63a706ee/checks?check_suite_id=275620440

EskiMag:
If you have some proposals on how to improve the whole CI, let me know somehow. :-)

lpil:
Awesome! I'll make some extra tickets in the issue tracker now :)

Issue #287, opened 2019-10-22
Run rustfmt on CI


lpil:
Note these don't need to be ran for every OS, just on Linux will be fine.

Rustfmt should fail if any files are not correctly formatted.

lpil:
@EskiMag I think you need to comment here for me to assign you

EskiMag:
Please, assign this issue to me. 🙏 😁

EskiMag:
OK, so I wanted to work on this, but I am stuck. I am not familiar with Rust, I wrote maybe 10 lines in it few years ago. I did a clean install via the recommended method, then tried to run `cargo fmt` or `rustfmt` inside the `gleam` directory of this repo, but it's stuck on this error:
```
error: command failed: 'rustfmt'
error: caused by: Permission denied (os error 13)
```
I checked Stackoverflow, but no solution is working for me. I am using macOS Catalina and `rustup --version` is returning `rustup 1.20.2 (13979c968 2019-10-16)`. Any tips, what can be wrong?

OvermindDL1:
What version of rust, not rustup?  Run `rustup update` to make sure it is up to date and post the output here so we can see what components it installs as well?

EskiMag:
I deleted both `~/.cargo` and `~/.rustup`, installed again and it's working. Anyway, thanks @OvermindDL1 for your help. There were probably some old resuides from the past causing the trouble.
Now when I run `cargo fmt`, I can the `src/main.rs` is being formatted.
Is there some switch for the `cargo fmt` command to just check if everything is properly formatted? Or should I combine the `cargo fmt` with `git status` to decide if the check should fail or succeed?

OvermindDL1:
Different versions can format slightly differently.  Generally you just have the CI locked to 'latest' and format everything and commit if any changes.

If you really want to 'just check' (which will fail if someone formatted with a different version at times) then I think it's just `cargo fmt -- --check` if I recall correctly.

PR #295, opened 2019-10-24
Add rustfmt into CI workflow

EskiMag:
This Pull Request solves #287

@lpil is this enough or do you have something else in mind?


lpil:
Hmm... It's not running the build on this PR. Could you try adding `pull_request` to the `on` array

Issue #310, opened 2019-10-28
GitHub actions does not run on gleam-lang GitHub org


lpil:
I have contacted support about this.

lpil:
Seems to be running now!

PR #312, opened 2019-10-30
GitHub Actions macOS workflow for release created

### LINKED ISSUE #290: Compile OSX binary on GitHub actions when a new release is created
https://help.github.com/en/github/automating-your-workflow-with-github-actions/events-that-trigger-workflows#release-event-release

EskiMag:
Closes #290 

This workflow runs only when:
- new release is created directly as public
- draft release is published

It would be nice if this workflow runs also after a new release is drafted, so @lpil can download the built release and perhaps test it manualy before publishing.
Unfortunately, GitHub's documentation for the **release** event is probably outdated. I already created a topic about this on community forum [here](https://github.community/t5/GitHub-Actions/Incorrect-documentation-for-release-event/m-p/36361#M2552)

I also changed the `on` rules for `build_compiler.yaml` workflow. Before the change it was running for every push and that included even tags. Now it runs only for new commits. This should help #310 although it doesn't solve it completely.

### DIFF
--- a/.github/workflows/build_compiler.yaml
+++ b/.github/workflows/build_compiler.yaml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - '*'
+  pull_request:
 
 jobs:
   rustfmt:
--- a/.github/workflows/create_release.yaml
+++ b/.github/workflows/create_release.yaml
@@ -1,0 +1,29 @@
+name: CI
+
+on:
+  release:
+    types: [created, published]
+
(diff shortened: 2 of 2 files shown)


EskiMag:
You can see the run here:
https://github.com/EskiMag/gleam/commit/006e419024f3951efee47ea4e2eece470ebb4464/checks?check_suite_id=288231271

And an example release here:
https://github.com/EskiMag/gleam/releases/tag/0.5.1

lpil:
Great! Thank you! I've learnt a lot from this :)

PR #313, opened 2019-10-31
Setup Linux binary release workflow

### LINKED ISSUE #291: Compile Docker image and Linux binary on GitHub actions when a new vX.X.X tag is created

EskiMag:
Closes #291 

Example release:
https://github.com/EskiMag/gleam/releases/tag/0.5.3

Example run:
https://github.com/EskiMag/gleam/runs/282990410

For some reason, I had to re-run the workflow few times. It was returning false negatives on the `upload-release-asset` step even when the asset was successfully uploaded. I guess it's just a _beta syndrome_.

### DIFF
--- a/.github/workflows/create_release.yaml
+++ b/.github/workflows/create_release.yaml
@@ -27,3 +27,32 @@ jobs:
           asset_path: ${{ steps.create_archive.outputs.ASSET }}
           asset_name: ${{ steps.create_archive.outputs.ASSET }}
           asset_content_type: application/gzip
+
+  build_release_linux:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


lpil:
This builds a dynamically linked binary which will require additional libs to be installed on the OS in order to run. Could you build a statically linked one? My current approach is to build within a docker container and then extract the binary. https://github.com/gleam-lang/gleam/blob/master/bin/build-linux-release.sh

EskiMag:
Oh, OK. Now I understand the whole process in that script. 👍

lpil:
I should have made that clearer, thanks :)

PR #318, opened 2019-11-06
Cache Rust dependnecies in GitHub Actions

### LINKED ISSUE #288: Cache Rust dependencies build step on GitHub actions
This is currently blocked by GitHub having not yet released their caching feature.

https://stackoverflow.com/questions/55110729/how-do-i-cache-steps-in-github-actions/57958803#57958803

EskiMag:
Closes #288 

OK, GitHub increased the per-action limit to 400MB ([here](https://github.com/actions/cache/issues/6#issuecomment-550051416)), so our actions are now running with cache.

* [before](https://github.com/EskiMag/gleam/commit/c627de22fc102b4b0fff875c5a9d96deece066af/checks?check_suite_id=297886003) 9m 57s
* [after](https://github.com/EskiMag/gleam/commit/6088bd98ed1ce0143527872a186be5a747c58346/checks?check_suite_id=297898728) 2m 24s

Quite an improvement.

Just one thing I noticed is the _"Cache hit occurred on the primary key Linux-rust-target, not saving cache."_ notice [here](https://github.com/EskiMag/gleam/runs/290480288) in the _Post Cache Rust dependencies_ step.
I am not sure what that really means, but I think we probably need to use the `restore-keys` option as shown [here](https://github.com/actions/cache/blob/master/README.md).
But I don't know what file is used in Rust to something similar as `mix.lock` in Elixir for example.

### DIFF
--- a/.github/workflows/build_compiler.yaml
+++ b/.github/workflows/build_compiler.yaml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Cache Rust dependencies
-        uses: actions/cache@preview
+        uses: actions/cache@v1.0.1
         with:
           path: target
           key: ${{ runner.os }}-rust-target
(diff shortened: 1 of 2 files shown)


lpil:
Excellent!

> But I don't know what file is used in Rust to something similar as mix.lock in Elixir for example.

There is a Cargo.lock file that can be used in the same way :)

EskiMag:
OK, done. I think we can merge.
[example run](https://github.com/EskiMag/gleam/commit/68c3ac495396857b89df60385c261282e7f6c24b/checks?check_suite_id=298317357)

Issue #303, opened 2019-10-27
Alias unqualified imports


lpil:
```javascript
import gleam/list.{map as list_map}
```

zjhmale:
working on this

lpil:
Thank you

thehabbos007:
Hey!
I started looking at this feature today and I've suceeded in adding the syntax to the grammar and the AST. However, I am unsure how to correctly do something like module alias. 

I thought maybe adding a variable to the environment might be a solution to alias the imports? I'm not quite sure the implications of doing that though :)

lpil:
Great! Thanks for your help.

I was thinking something like this: 

Add a field to the UnqualifiedImport AST node for the alias name. This is sometimes there and sometimes not so I'd make the field `Option<String>`. https://github.com/gleam-lang/gleam/blob/87aa490e6813bbe208472e1a2bf60ed8e9a39ddf/src/ast.rs#L157-L160

Then in the typer when we are processing each UnqualifiedImport insert the variable or type using the alias field if it is present, or the regular name if it is not.
https://github.com/gleam-lang/gleam/blob/87aa490e6813bbe208472e1a2bf60ed8e9a39ddf/src/typ.rs#L2008-L2016

A test like this one for an aliased value and another for an aliased type would be good https://github.com/gleam-lang/gleam/blob/87aa490e6813bbe208472e1a2bf60ed8e9a39ddf/src/project.rs#L817-L855

lpil:
If you get stuck anywhere or no longer have time to continue please open a WIP PR and either we'll help you continue or I'll pick up this work from you, whichever you prefer :)

thehabbos007:
Yes, I had tried to do the same as above in the typer, but the compiled erlang files use the alias instead of the relevant function. I assume I somehow have to map the alias to the real function name, right? Or am i missing something? :)

lpil:
It's possible there's more to do after those steps. The compiler is too large to fit in my head these days :)

Please open a PR and I'll take a look

lpil:
Waiting on #357

PR #361, opened 2019-12-23
Alias unqualified

thehabbos007:
WIP PR for #303 to investigate further complications

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -2002,23 +2002,28 @@ This should not be possible. Please report this crash",
                 };
 
                 // Insert unqualified imports into scope
-                for UnqualifiedImport { name, meta } in &unqualified {
+                for UnqualifiedImport { name, meta, as_name } in &unqualified {
                     let mut imported = false;
 
+                    let unqualified_alias = match &as_name {
+                        None => name.clone(),
+                        Some(alias) => alias.clone(),
+                    };
+
                     if let Some(value) = module_info.value_constructors.get(name) {
-                        env.insert_variable(name.clone(), value.variant.clone(), value.typ.clone());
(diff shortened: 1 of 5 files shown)


src/typ.rs
```diff
@@ -2002,23 +2002,37 @@ This should not be possible. Please report this crash",
...
+                    name,
+                    meta,
+                    as_name,
+                } in &unqualified
+                {
                     let mut imported = false;
 
+                    let unqualified_alias = match &as_name {
+                        None => name.clone(),
+                        Some(alias) => alias.clone(),
```
lpil: Does this compile if we remove the `.clone()` from these two lines? It would be good to only clone the string once (when we insert it) rather than twice.

thehabbos007: Yes it does seem to compile. I've changed it!

lpil:
This looks great so far!

https://github.com/gleam-lang/gleam/blob/a90b31a5adcabf5ad937128ed49dde5328c9b367/src/typ.rs#L736-L759

It looks like I didn't make the variables store enough information to know what the correct value should be. I am currently reworking this data structure in #357 so I can add that information there.

Seeing as it's not possible for you to continue with this until I have finished my work how would you feel about commenting out those tests, adding a `// TODO: ` comment next to them and then merging this in?

thehabbos007:
That seems like a great idea. Thank you for the quick responses by the way, it's great!
Do i just keep the new syntax in the PR?

lpil:
> Do i just keep the new syntax in the PR?

Yes please, keep everything but comment out the tests.

Once I have fixed the other problem I think this will work so there will be nothing more for you to do :)

PR #372, opened 2020-01-08
Update build_compiler.yaml

imbsky:
Fetch performance has been improved since checkout action v2.0.0. Also, the cache limit has been bumped to 2GB since cache action v 1.1.0.


imbsky:
Sure! I'll open it later!

lpil:
Thank you

PR #381, opened 2020-01-17
Enclosing (top level) types shown in error messages

### LINKED ISSUE #378: Type errors are unclear when the top level type is correct but type parameters are not
```
error: Type mismatch

   ┌── /private/var/folders/dc/0tgdh7d11x95m6xx1vgvxgrh0000gq/T/tmp.mflSkcdp/tetromino/src/tetromino.gleam:2:10 ───
   │
 2 │   [1] == [1.0]
   │          ^^^^^
   │


Expected type:

    Int

Found type:

    Float
```

This error message should make it clear that it is `List(Int)` and `List(Float)`

## Option 1

```
Expected type:

    tuple(Int, Int, Float)

Found type:

    tuple(Int, Float, Float)
```

## Option 2

```
Expected type:

    Int

Found type:

    Float

In type:

    tuple(_, Float, _)
```

## Option 3

Something else?

thehabbos007:
Closes #378 
This is going with option 1, but option 2 is possible too!
Changelog might have to be changed to look like the previous error message lines :)

It's such a pleasure to work with this code base, even as rust beginner!

```
error: Type mismatch

   --- /hello_world/src/other.gleam:4:12 ---
   |
 4 |   [1.0] == [1]
   |            ^^^
   |

Expected type:

    List(Float)

Found type:

    List(Int)
```

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -2271,10 +2271,10 @@ enum UnifyError {
     RecursiveType,
 }
 
-fn unifyEnclosedType(e1: &Type, e2: &Type, result: Result<(), UnifyError>) -> Result<(), UnifyError> {
+fn unify_enclosed_type(e1: &Type, e2: &Type, result: Result<(), UnifyError>) -> Result<(), UnifyError> {
     // If types cannot unify, show the type error with the enclosing types, e1 and e2.
     match result {
-        Err(e) =>  Err(UnifyError::CouldNotUnify {
+        Err(_) =>  Err(UnifyError::CouldNotUnify {
             expected: (*e1).clone(),
             given: (*e2).clone(),
         }),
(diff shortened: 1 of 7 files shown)


src/typ.rs
```diff
@@ -2271,6 +2271,21 @@ enum UnifyError {
...
 }
 
+fn unify_enclosed_type(
+    e1: &Type,
+    e2: &Type,
+    result: Result<(), UnifyError>,
+) -> Result<(), UnifyError> {
+    // If types cannot unify, show the type error with the enclosing types, e1 and e2.
+    match result {
+        Err(_) => Err(UnifyError::CouldNotUnify {
```
lpil: Could we match specifically on `Err(UnifyError::CouldNotUnify { .. })`? To be sure it's the right error we're overriding

lpil:
I've rebased these changes onto master, thank you! :)

Issue #368, opened 2019-12-29
Old modules remain in `gen/` directory after recompilation


rjdellecese:
When I create a module, compile my Gleam code, and then rename or delete that module, its generated Erlang code sticks around in the `gen/` directory, even after recompiling again. I would have assumed that upon recompilation, the `gen/` directory would reflect only the Gleam code in my project at that time.

This very rarely has caused any issues, but is a somewhat opaque problem to debug when it does!

lpil:
We're don't have any way of knowing whether modules have been removed or not so we cannot safely clean the gen directory. Should the rebar plugin delete the directory? Or perhaps we could keep track of what was previously generated.

rjdellecese:
Although a bit ham-fisted, is there any reason deleting the directory before recompiling wouldn't work fine, for now?

The other solution (of keeping track of what was previously generated) would probably involve generating a manifest, right? Would that help at all with the future of https://github.com/gleam-lang/gleam/issues/157? My reference in asking is the way the [the Elixir compiler](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/compilers/elixir.ex) utilizes a manifest file.

lpil:
I think the manifest is the way to go, and it will be required in that ticket.

I guess for now we could delete the directory, I can't think of any reason why it would cause problems :)

rjdellecese:
If you haven't already starting working on this I'd love to take a crack at it @lpil (also assuming that wouldn't mean holding up a release!).

NobbZ:
If we delete the `gen` folders content, then every module will be recompiled, this is probably true.

Though if we keep it, will still everything be recompiled or only those modules that need to due to changes of their sources?

If the latter I opt against unconditionally deleting the folder, but instead use heuristics to check if files need to be removed or only remove on `--force` flag.

lpil:
We don't have incremental compilation yet, all modules are always recompiled. More fine grained compilation will come as part of #157

lpil:
We always recompile all Gleam modules and overwrite the Erlang modules, so this shouldn't change the current behaviour of rebar3. Incremental compilation is tracked here #157

PR #383, opened 2020-01-21
Delete gen directory before compilation

rjdellecese:
Closes https://github.com/gleam-lang/gleam/issues/368


src/main.rs
```diff
@@ -119,6 +119,18 @@ fn command_build(root: String) -> Result<(), Error> {
 
     let compiled = crate::project::compile(srcs)?;
 
+    // Delete the gen directory before generating the newly compiled files
+    let gen_dir = root_path.parent().unwrap().join("gen");
```
lpil: Can we `.map_err(|e| Error::FileIO {` for the `parent` call rather than unwrapping it? Thank you.

rjdellecese:
All set now, I think!

lpil:
Just waiting for tests

PR #389, opened 2020-02-08
feat(type system): implement type alias statement

### LINKED ISSUE #71: Type aliases
```rust
type IntMap = Map(Int, Int)
```

zjhmale:
right now, we can use `type IntMap = Map(Int, Int)` semantic in gleam :)

fix #71

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -25,10 +25,21 @@ pub Module: UntypedModule = {
 
 Statement: UntypedStatement = {
     StatementFn => <>,
+    StatementTypeAlias => <>,
     StatementCustomType => <>,
     StatementExternalFn => <>,
     StatementExternalType => <>,
     StatementImport => <>,
+}
+
+StatementTypeAlias : UntypedStatement = {
+    <s:@L> <p:"pub"?> "type" <ta:TypeName> "=" <tr:Type> <e:@L> => Statement::TypeAlias {
+        meta: meta(s, e),
+        public: p.is_some(),
+        alias: ta.0,
+        args: ta.1,
+        resolved_type: tr,
+    }
 }
 
 StatementCustomType: UntypedStatement = {
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -276,6 +276,15 @@ pub struct Module {
     pub name: Vec<String>,
     pub types: HashMap<String, TypeConstructor>,
     pub values: HashMap<String, ValueConstructor>,
+    pub type_aliases: HashMap<String, TypeAliasConstructor>,
+}
+
+impl Module {
+    pub fn resolve_module_type_alias(&self, name: &str) -> Option<&Type> {
+        self.type_aliases
+            .get(name)
+            .map(|TypeAliasConstructor { ref typ, .. }| typ)
+    }
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -298,6 +307,9 @@ pub struct Env<'a> {
 
     // Values defined in the current module
     module_values: HashMap<String, ValueConstructor>,
+
+    // type alias defined in the current module (or the prelude)
+    module_type_aliases: HashMap<String, TypeAliasConstructor>,
 }
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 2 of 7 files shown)


zjhmale:
test cases are still under the way 😰

src/grammar.lalrpop
```diff
@@ -25,12 +25,22 @@ pub Module: UntypedModule = {
...
     StatementCustomType => <>,
     StatementExternalFn => <>,
     StatementExternalType => <>,
     StatementImport => <>,
 }
 
+StatementTypeAlias : UntypedStatement = {
+    <s:@L> <p:"pub"?> "type" <ta:TypeName> "=" <tr:Type> <e:@L> => Statement::TypeAlias {
+        meta: meta(s, e),
+        public: p.is_some(),
```
lpil: This is cool :)

src/typ.rs
```diff
@@ -777,6 +797,49 @@ impl<'a> Env<'a> {
...
+
+    /// Map a type alias in the current scope.
+    ///
+    pub fn insert_type_alias(&mut self, name: String, info: TypeAliasConstructor) {
+        self.module_type_aliases.insert(name, info);
+    }
+
+    /// Lookup a type alias in the current scope.
+    ///
+    pub fn get_real_type(
```
lpil: I find the term `real_type` a little unclear. Could we call this `resolve_type_alias`? Same for other places where `real_type` is used.

src/typ.rs
```diff
@@ -865,27 +928,31 @@ impl<'a> Env<'a> {
...
-                    .collect::<Result<Vec<_>, _>>()?;
-                let info = self
-                    .get_type_constructor(module, name)
-                    .map_err(|e| convert_get_type_constructor_error(e, &meta))?;
-                if args.len() != info.arity {
-                    return Err(Error::IncorrectTypeArity {
-                        meta: meta.clone(),
+                if let Ok(typ) = self.get_real_type(module, name) {
+                    return Ok(typ.to_owned());
+                } else {
```
lpil: As we're returning in the `if branch we can remove the `else` branch and have the code as it was before.

src/typ.rs
```diff
@@ -1341,6 +1425,33 @@ pub fn infer_module(
                 })
             }
 
+            Statement::TypeAlias {
+                meta,
+                public,
+                alias,
+                real_type,
+            } => {
+                let mut type_vars = hashmap![];
```
lpil: I think we should check that the type being aliased exists before we register it. i.e. this should be an error:

```rust
type MyAlias = TypeThatDoesNotExist
```

src/typ.rs
zjhmale: sure

src/typ.rs
zjhmale: maybe changing `get_real_type` to `resolve_type_alias` and `real_type` to `resolved_type` ?

src/typ.rs
zjhmale: this is already handled, I will only `insert_type_alias` if and only if the resolved type is legal, otherwise it will be an unknown type error

image

zjhmale:
- [ ] parametric polymorphism .e.g. `type Option(a) = Result(a, Nil)`.
- [ ] more test cases, both success and error ones.

src/typ.rs
lpil: Ah yes! I forgot that function does that. Thank you :)

src/typ.rs
lpil: sounds good

src/typ.rs
```diff
@@ -1341,6 +1529,43 @@ pub fn infer_module(
...
+                args,
+                resolved_type,
+            } => {
+                let mut type_vars = hashmap![];
+                let resolved_type_var_names = resolved_type.get_type_vars();
+
+                match env.type_from_ast(&resolved_type, &mut type_vars, NewTypeAction::MakeGeneric)
+                {
+                    Ok(typ) => {
+                        let generics = typ.bind_generic_args(&resolved_type_var_names);
```
lpil: I think this approach could be made simpler by re-using existing functionality in the typer and by storing the type AST in the alias rather than the resolved type.

## When a type alias is defined:

1. Resolve the alias once to ensure the type is valid, then discard the resulting type
2. Store the alias name, the parameter names, and the AST of the aliased type

## When a type alias being resolved

1. Iterate across the alias args and insert the given parameter into `type_vars` for that name to seed the scope with those params.
2. Call `env.type_from_ast`, specifying `NewTypeAction::Disallow` to ensure there are no extra type variables that were not already declared.
3. Return the newly constructed type as the resolved type

lpil:
Thank you for this!

I am picking up this work and continuing in #398

PR #401, opened 2020-03-01
Docs POC

gamebox:
This is EXTREMELY rough.  Mostly due to limitations of lalrpop, and having to manually extract doc comments.  But it maintains existing invariants, and it works.  The output is based on Hexdocs for Elixir.

TODO
---
* Clean up bad Rust code.  I don't _really_ know the language, and am probably not being idiomatic nor efficient.
* ~~Move doc writing to a DocWriter that can collect all the modules and generate navigation for a sidebar.~~
* ~~Actually create a visual style.~~
* ~~Parse out doc metadata.~~
* [OUT OF SCOPE]~~Parse out doc tests / examples~~
* ~~Hide behind a flag.~~


src/ast.rs
```diff
@@ -125,6 +127,7 @@ pub enum Statement<ValueConstructor, ModuleValueConstructor, PatternConstructor,
         args: Vec<String>,
         resolved_type: TypeAst,
         public: bool,
+        comments: Vec<String>,
```
gamebox: Remove

src/ast.rs
```diff
@@ -117,6 +118,7 @@ pub enum Statement<ValueConstructor, ModuleValueConstructor, PatternConstructor,
         body: Expr<ValueConstructor, ModuleValueConstructor, PatternConstructor, Type>,
         public: bool,
         return_annotation: Option<TypeAst>,
+        comments: Vec<String>,
```
gamebox: Remove

src/ast.rs
```diff
@@ -133,6 +136,7 @@ pub enum Statement<ValueConstructor, ModuleValueConstructor, PatternConstructor,
         args: Vec<String>,
         public: bool,
         constructors: Vec<RecordConstructor>,
+        comments: Vec<String>,
```
gamebox: Remove

src/ast.rs
```diff
@@ -143,13 +147,15 @@ pub enum Statement<ValueConstructor, ModuleValueConstructor, PatternConstructor,
         retrn: TypeAst,
         module: String,
         fun: String,
+        comments: Vec<String>,
```
gamebox: Remove

src/ast.rs
```diff
@@ -143,13 +147,15 @@ pub enum Statement<ValueConstructor, ModuleValueConstructor, PatternConstructor,
...
     },
 
     ExternalType {
         meta: Meta,
         public: bool,
         name: String,
         args: Vec<String>,
+        comments: Vec<String>,
```
gamebox: Remove

src/doc.rs
```diff
@@ -0,0 +1,354 @@
+use super::ast::{Arg, ArgNames, Statement, TypeAst, TypedModule};
```
gamebox: This could be a directory

src/main.rs
```diff
@@ -124,15 +126,8 @@ fn command_build(root: String) -> Result<(), Error> {
...
-    let gen_dir = root_path
-        .parent()
-        .ok_or_else(|| Error::FileIO {
-            action: error::FileIOAction::FindParent,
-            kind: error::FileKind::Directory,
-            path: root_path.clone(),
-            err: None,
-        })?
-        .join("gen");
+    let gen_dir = root_path.join("gen");
```
gamebox: This was actually a bug I found.  It was always a noop because it was looking outside the project dir

src/doc.rs
lpil: What do you mean?

src/main.rs
lpil: Nice catch!

src/doc.rs
```diff
@@ -0,0 +1,354 @@
...
+        let new_line = RawComment {
+          line_no: line_no,
+          content,
+        };
+        if (line_no - 1) == *l {
+          self.current_block.push(new_line);
+        } else {
+          let copy = std::mem::replace(&mut self.current_block, vec![]);
+          self.blocks.push(copy);
+          self.current_block = vec![];
```
lpil: I don't think we need to do this as we've set `self.current_block` to a new vector with `std::mem::replace(&mut self.current_block, vec![]);`

src/doc.rs
```diff
@@ -0,0 +1,354 @@
...
+        cross_references: None,
+        edit_url: format!("https://www.example.com/{}", module.name_string()),
+      },
+      docs,
+    }
+  }
+}
+
+fn type_to_string(typ: &TypeAst) -> String {
+  "".to_string()
```
lpil: We can use the `typ::pretty` module for this :)

src/doc.rs
gamebox: Sorry, this file is large, and could be split into smaller files

src/doc.rs
gamebox: Oh, nice catch.  That was from previous rounds of my bout with the compiler.

src/doc.rs
gamebox: Awesome.  I'm glad I don't have to reinvent the wheel.

src/doc.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+      }
+      None => {
+        self.current_block.push(RawComment { line_no, content });
+      }
+    }
+    self.num_comments = self.num_comments + 1;
+  }
+
+  pub fn gen_doc_chunk(self: &Self, module: &TypedModule) -> EEP48DocChunk {
```
lpil: I don't think this is the right way to approach this. If we operate on the raw AST we only have basic information about function names etc as we have not done any of the static analysis yet (such as type checking).

We could instead build the documentation from the output of the `project::compile` function which returns `Result<Vec<Compiled>, Error>`. The `Compiled` struct has lots of useful information.

gamebox: I thought that the output of infer_module had the type information?

src/doc.rs
```diff
@@ -0,0 +1,354 @@
...
+struct RawComment {
+  pub line_no: usize,
+  pub content: String,
+}
+
+#[derive(Debug, PartialEq)]
+pub struct DocBlockManager {
+  pub(self) current_block: Vec<RawComment>,
+  pub(self) blocks: Vec<Vec<RawComment>>,
+  num_comments: usize,
```
lpil: Is this used for anything? I couldn't spot it

gamebox: Look in parser.rs, it's what the comment lines get added to.

src/doc.rs
```diff
@@ -0,0 +1,354 @@
...
+impl DocBlockManager {
+  pub fn new() -> Self {
+    DocBlockManager {
+      current_block: vec![],
+      blocks: vec![],
+      num_comments: 0,
+    }
+  }
+
+  pub fn block_count(self: &Self) -> usize {
```
lpil: Not sure this is used either

src/doc.rs
lpil: Yeah memory management is real fiddly in Rust!

src/doc.rs
```diff
@@ -0,0 +1,354 @@
+use super::ast::{Arg, ArgNames, Statement, TypeAst, TypedModule};
+use itertools::Itertools;
+use std::collections::HashMap;
+#[derive(Debug, PartialEq)]
+struct RawComment {
+  pub line_no: usize,
```
lpil: We annotate the AST using the byte index rather than a line number so it might be easier to merge the comments in to the AST if store the byte index here rather than the line numbers

src/doc.rs
lpil: No need to be sorry, this is great :)

src/doc.rs
lpil: Oh I see! We're iterating over the typed module. My bad.

I think it'd be better to use the `Compiled` struct too as it is all the static information collected into a more compact data structure

src/doc.rs
lpil: I mean specifically the num_comments property

gamebox: Ah, it's used in the comment_count method that's module protected.  It's meant primarily for testing purposes.

src/doc.rs
gamebox: Yeah, this needs to be updated, it is the byte index* now.

 * At least the char index, which will be correct as long as the file uses the ASCII subset of utf-8.

lpil: We support unicode so we'll need to check the byte size of the chars as we iterate over them

src/doc.rs
gamebox: I didn't really find that `Compiled` had that much more useful data in it, and this let me generate the docs chunks as we are going without changing the `project::compile` function's signature too much.  This prevents a separate pass from having to happen.

lpil:
Hi @gamebox, this PR seems to be covering multiple features at once. Would you be able to split them into separate PRs so that we can give each a proper review? Thank you

gamebox:
Of course, I'll put the pretty Printer first

gamebox:
Looks like I need to run rustfmt

src/doc.rs
gamebox: Looks like the offset from the iterator is byte index

lpil:
Tests are passing! Awesome :)

Is this ready for a review?

gamebox:
Yes it is!

src/doc/writer.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    fn write_index(self: &Self, module: &Module) -> Result<(), Error> {
+        let rendered = self
+            .registry
+            .render("index", module)
+            .map_err(|_| Error::FileIO {
+                path: self.doc_dir.join("index.html"),
+                action: FileIOAction::Create,
+                kind: FileKind::File,
+                err: Some("Could not parse README as valid markdown".to_string()),
```
lpil: This err should be the argument that this `map_err` closure takes.

src/doc/writer.rs
```diff
@@ -0,0 +1,259 @@
...
+        }
+        self.modules.push(self.chunk_to_module(chunk));
+    }
+
+    fn write_index(self: &Self, module: &Module) -> Result<(), Error> {
+        let rendered = self
+            .registry
+            .render("index", module)
+            .map_err(|_| Error::FileIO {
+                path: self.doc_dir.join("index.html"),
```
lpil: This should be where the error originated, so `REAME.md`

src/doc/writer.rs
```diff
@@ -0,0 +1,259 @@
...
+                action: FileIOAction::Create,
```
lpil: It's a parse error, so `Parse`.

Though looking at the code it could be several other errors here. We can make this more precise in a later PR

src/doc/writer.rs
```diff
@@ -0,0 +1,259 @@
...
+        self.write_to_path(rendered.as_bytes(), self.doc_dir.join("index.html"))
+    }
+
+    pub fn write(self: &Self) -> Result<(), Error> {
+        let readme =
+            read_to_string(self.package_root.join("README.md")).map_err(|_| Error::FileIO {
+                path: self.package_root.join("README.md"),
+                action: FileIOAction::Read,
+                kind: FileKind::File,
+                err: Some("Could not read README.md for doc index".to_string()),
```
lpil: This err should be the argument that this `map_err` closure takes.

src/doc/writer.rs
```diff
@@ -0,0 +1,259 @@
...
+        let filename = format!("{}.html", module.module_name);
+        let doc_dir_path = self.doc_dir.join(&filename);
+        let doc_text = self
+            .registry
+            .render("module", &module)
+            .map_err(|_e| Error::FileIO {
+                action: FileIOAction::Parse,
+                kind: FileKind::File,
+                path: doc_dir_path.clone(),
+                err: None,
```
lpil: This err should be the argument that this `map_err` closure takes.

src/main.rs
```diff
@@ -27,6 +29,13 @@ extern crate lalrpop_util;
 #[macro_use]
 extern crate lazy_static;
 
+extern crate serde_json;
```
lpil: We're using Rust 2018 so no need for these `extern crate` declarations anymore.

src/main.rs
```diff
@@ -148,8 +151,33 @@ fn command_build(root: String) -> Result<(), Error> {
...
+                err: Some(e.to_string()),
+            })?;
+        }
+    }
+
+    let mut doc_writer = doc::writer::DocWriter::new(
+        project_config.name.clone(),
+        project_config
+            .version
+            .unwrap_or("0.0.1".to_string())
```
lpil: We default to `1.0.0` in `new.rs`

src/main.rs
```diff
@@ -180,8 +208,13 @@ fn command_build(root: String) -> Result<(), Error> {
             })?;
         }
     }
-
     println!("Done!");
+
+    if write_docs {
+        doc_writer.write()?;
```
lpil: Probably should come before `println!("Done!");`

src/pretty.rs
```diff
@@ -6,7 +6,10 @@
 
 use im::vector::Vector;
 
-pub trait Documentable {
+pub trait Documentable
+where
+    Self: Clone,
```
lpil: This contraint isn't used, let's remove it

1q
```diff
@@ -0,0 +1,5 @@
+# Please enter a commit message to explain why this merge is necessary,
```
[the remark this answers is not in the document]
gamebox: Nope :-)

src/doc/doc.rs
```diff
@@ -0,0 +1,84 @@
...
+#[derive(Debug, PartialEq, Eq, Copy, Clone)]
+pub enum DocType {
+    Fn,
+    TypeAlias,
+    CustomType,
+    ExternalFn,
+    ExternalType,
+}
+
+impl std::hash::Hash for DocType {
```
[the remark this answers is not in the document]
gamebox: See `src/doc/writer.rs:165`.  It's used to allow this type to be used as the key of a HashMap.

src/doc/doc.rs
```diff
@@ -0,0 +1,84 @@
...
+#[derive(Debug, PartialEq)]
+pub struct ErlAnno {
+    pub line: usize,
+    pub column: usize,
+    pub file: String,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum CrossReference {
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
gamebox: This is not used yet, but will be when we parse links in documentation to point to other documents in the module.  I guess we could practice YAGNI here, but I'd like to keep it for EEP48 spec completeness.

src/doc/doc.rs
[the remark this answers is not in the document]
lpil: Neat!

lpil:
There's a few references to `meta`, which has been renamed. Could you bring this up to date with master please? 🙏

gamebox:
There are?  It passes all the tests... I'll take a look

gamebox:
Sorry but the only references I see to `meta` are to `DocMeta`, which is EEP48 term

lpil:
Oh right! that's my misunderstanding :)

PR #406, opened 2020-03-04
Anonymous function return type annotations

### LINKED ISSUE #404: Annotating return types on anonymous functions
I just installed Gleam v0.7.0 and have been playing with it this afternoon - it's really fun, thanks so much for making it! As a big Elm fan, I've been wanting a similarly friendly statically typed language on the Beam for ages.

I have an anonymous function where I've annotated the type of the argument as follows:

```gleam
pub fn example(x: Int) -> Int {
  let anon = fn(y: Int) {
    1 + y
  } 
  anon(x)
}
```

However, if I also try to annotate the return type:

```gleam
pub fn example(x: Int) -> Int {
  let anon = fn(y: Int) -> Int {
    1 + y
  } 
  anon(x)
}
```

I get:

```gleam
error: Syntax error

let anon = fn(y: Int) -> Int {
                      ^^ Unexpected token

Expected one of "{"
```

I'm not sure if this is expected behaviour? It seems unintuitive to be able to annotate the arguments of an anonymous function but not the return type.

thehabbos007:
Closes #404 

Seems like a great feature to have!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Anonymous functions can now have return type annotations.
+
 ## v0.7.1 - 2020-03-03
 
 - Projects generated with `gleam new` use `stdlib` version 0.7.0.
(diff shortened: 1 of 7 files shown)


thehabbos007:
Oh, I forgot to commit that file, will push it later :)

Issue #425, opened 2020-03-23
Support larger Int literals


lpil:
Currently we are limited to i64, which can overflow easily.

We could store it as a string as we don't actually use the Int value in the compilation process.

theAnnie:
@lpil can I take a look at this?

lpil:
Yes please! :) Let me know if you need any pointers at all

theAnnie:
> Yes please! :) Let me know if you need any pointers at all

@lpil So I've created PR, although I need to change all the Ints in test like 1 -> "1" and it does not seems to be right... Should not I parse it somewhere from String to Int back again?

lpil:
Great job! Yes the tests will need to be updated too. I may be able to help with some Vim regex if that would be useful.

`"1"` is an str in Rust, it'll need to be `"1".to_string()`.

I don't think we'll need to convert them to ints anywhere.

theAnnie:
@lpil ok then ;) I'll ping u when it will be done

lpil:
thank you!

lpil:
Thank you @theAnnie !

PR #437, opened 2020-03-28
Support larger Int literals

theAnnie:
Addresses #425

### DIFF
--- a/src/format/tests.rs
+++ b/src/format/tests.rs
@@ -349,8 +349,8 @@ pub external type Four
     assert_format!(
         r#"fn main() {
   fn() {
-    1
-    2
+    "1"
+    "2"
   }
 }
 "#
@@ -379,14 +379,14 @@ pub external type Four
 
     assert_format!(
         r#"fn main() {
-  run(1)
+  run("1")
 }
 "#
     );
 
     assert_format!(
         r#"fn main() {
-  run(with: 1)
+  run(with: "1")
 }
 "#
     );
@@ -394,8 +394,8 @@ pub external type Four
     assert_format!(
         r#"fn main() {
   run(
-    with: 1,
-    loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong: 1,
+    with: "1",
+    loooooooooooooooooooooooooooooooooooooooooooooooooooooooooooong: "1",
   )
 }
 "#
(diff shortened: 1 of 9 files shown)


src/typ/tests.rs
```diff
@@ -44,35 +44,35 @@ fn field_map_reorder_test() {
             CallArg {
                 location: Default::default(),
                 label: None,
-                value: int(1),
+                value: int("1".to_string()),
```
lpil: We could make this `int` function take `&str` instead of `String` and then call `.to_string()` in the body of the function so that we don't need to repeate `to_string()` where it is used

lpil:
We need up update grammar.lalrpop too! :)

theAnnie:
> We need up update `grammar.lalrpop` too! :)

Done

src/format/tests.rs
```diff
@@ -349,8 +349,8 @@ pub external type Four
     assert_format!(
         r#"fn main() {
   fn() {
-    1
-    2
+    "1"
+    "2"
```
lpil: Looks like we're now formatting Ints as strings too. If we remove the `.surround("\"", "\"")` from `src/format.rs` for `UntypedExpr::Int` and `Pattern::Int` then it should fix it I think :)

src/typ/tests.rs
theAnnie: I'm not sure if I'm getting it right but
`fn infer_int(value: &str, location: SrcSpan) -> Result<TypedExpr, Error> {
    Ok(TypedExpr::Int {
        location,
        value: value.to_string(),
        typ: int(),
    })
}`
does not do the thing (typ.rs)

src/grammar.lalrpop
```diff
@@ -480,7 +480,7 @@ IntLiteral: i64 = {
 Int: UntypedExpr = {
     <s:@L> <value:IntLiteral> <e:@L> => UntypedExpr::Int {
         location: location(s, e),
-        value,
+        value: value.to_string(),
```
lpil: Rather than calling `.to_string()` here I think it would be better to change `IntLiteral` to return a `String`, not an `Int`, that way we're not converting `String -> Int -> String`.

src/typ/tests.rs
lpil: Here is the function: https://github.com/gleam-lang/gleam/blob/3f3000756e99a218d651edbf3e25126b66c2edf8/src/typ/tests.rs#L5-L8

I've written it in the closure syntax for some reason. Can just change `value,` to `value: value.to_string()` I think

theAnnie: Oh, I see. Done

src/format/tests.rs
theAnnie: oh, sorry formatter was stronger then me then me this time ;) fixed

src/grammar.lalrpop
theAnnie: Yep, that's better. Fixed

theAnnie:
> Thank you!

You're welcome! I just pushed small changes to grammar.lalrpop, please do not miss them

Issue #438, opened 2020-03-28
Store Floats as Strings in the AST


lpil:
We don't need to know their value so there is no reason to parse the source here.

Related to #425

theAnnie:
Can you assign me to this one, since I've made something very similar?

lpil:
For sure!

PR #440, opened 2020-03-28
Store Floats as Strings

theAnnie:
Addresses #438

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -313,7 +313,7 @@ impl Documentable for &UntypedPattern {
         match self {
             Pattern::Int { value, .. } => value.clone().to_doc(),
 
-            Pattern::Float { value, .. } => value.to_doc(),
+            Pattern::Float { value, .. } => value.clone().to_doc(),
 
             Pattern::String { value, .. } => value.clone().to_doc().surround("\"", "\""),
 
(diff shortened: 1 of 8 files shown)


theAnnie:
This will fail so far - we are losing rest of the  numer after `.`,  searching for the issue

src/format.rs
```diff
@@ -313,7 +313,7 @@ impl Documentable for &UntypedPattern {
         match self {
             Pattern::Int { value, .. } => value.clone().to_doc(),
 
-            Pattern::Float { value, .. } => value.to_doc(),
+            Pattern::Float { value, .. } => value.clone().to_doc(),
```
lpil: Gleam allows floats to omit the numbers after the `.`. In the Erlang we want the number to always have a number after the `.` so here we can check to see if the `value` string ends with `.` and add a 0 to the end if it does: `format!("{}0", value).to_doc()`

lpil:
Looking good! I've left a comment on what I think the issue is

src/format.rs
```diff
@@ -313,7 +313,13 @@ impl Documentable for &UntypedPattern {
         match self {
             Pattern::Int { value, .. } => value.clone().to_doc(),
 
-            Pattern::Float { value, .. } => value.to_doc(),
+            Pattern::Float { value, .. } => {  
+                if value.ends_with(".") {
+                    format!("{}0", value).to_doc()
+                } else {
+                    value.clone().to_doc()
+                }
```
theAnnie: I'm not sure if thats a good approach. Also for some reason `value.ends_with(".")` always returns false

lpil: Sorry, I have directed you to the wrong file. We want the Erlang printer, not the Gleam formatter.

theAnnie: np, fixed 😸

theAnnie:
> Thank you!

You're welcome 😄

PR #445, opened 2020-04-01
New lexer

lpil:


lpil:
Fun but not super useful atm :)

PR #453, opened 2020-04-11
Format files

### LINKED ISSUE #451: Formatting of Gleam files
https://github.com/gleam-lang/gleam/blob/dc3d6499c4d8503171188940175d1f2a64b7be3f/src/format/command.rs#L14

gamebox:
Closes #451


src/main.rs
```diff
@@ -82,7 +82,10 @@ enum Command {
 
     #[structopt(name = "format", about = "Format source code")]
     Format {
-        #[structopt(help = "files to format", required_unless = "stdin")]
+        #[structopt(help = "location of the project root", default_value = ".")]
```
lpil, line 85: Rather than having the root as an option could we copy the API of Rust format, Elm format, and prettier and use the paths as relative to the current directory if they are not absolute?

I think the file IO functions in Rust may behave like this is we give them the unaltered relative path, though I have not tested

gamebox: Done.  I had other issues that lead me to believe I needed this.

src/format/command.rs
```diff
@@ -10,8 +12,44 @@ pub fn run(stdin: bool, check: bool, files: Vec<String>) -> Result<(), Error> {
...
+            .read(true)
+            .open(path.clone())
+            .expect(&format!("Could not open file: {}", file_path));
+
+        let mut src = String::new();
+        file.read_to_string(&mut src)
+            .expect(&format!("Could not read file: {}", file_path));
+
+        let formatted = crate::format::pretty(src.as_ref()).map_err(|error| Error::Parse {
+            path: PathBuf::from("<standard input>"),
```
lpil, line 35: The error message here says the code is from stanard in, but it's from a file. Let's use the path of the file here to make the error message accurate.

src/format/command.rs
```diff
@@ -10,8 +12,44 @@ pub fn run(stdin: bool, check: bool, files: Vec<String>) -> Result<(), Error> {
...
+            kind: FileKind::File,
+            path: PathBuf::new(),
+            err: Some(format!("Invalid path: {}\n{}", file_path, e.to_string())),
+        })?;
+
+        let mut file = OpenOptions::new()
+            .write(true)
+            .read(true)
+            .open(path.clone())
+            .expect(&format!("Could not open file: {}", file_path));
```
lpil, line 28: It would be good to `Error` for these error cases rather than crashing the program so the user has a better idea how to solve the problem when this happens.

src/format/command.rs
```diff
@@ -10,8 +12,44 @@ pub fn run(stdin: bool, check: bool, files: Vec<String>) -> Result<(), Error> {
...
 
-pub fn format_files(_files: Vec<String>, _check: bool) -> Result<(), Error> {
-    todo!()
+pub fn format_files(files: Vec<String>, _check: bool) -> Result<(), Error> {
+    for file_path in files {
+        let path = PathBuf::from_str(&file_path).map_err(|e| Error::FileIO {
+            action: FileIOAction::Open,
+            kind: FileKind::File,
+            path: PathBuf::new(),
+            err: Some(format!("Invalid path: {}\n{}", file_path, e.to_string())),
```
lpil, line 21: Can we use the error `e` here rather than using our own error? The printed output says this this field is from the underlying library and I don't want to lose any information here.

src/format/command.rs
gamebox: Done.  The problem with copy-pasta

src/format/command.rs
gamebox: I _think_ I addressed this.

lpil:
Just waiting for tests to pass :)

Issue #455, opened 2020-04-14
Number comparisons in case clause guards


lpil:
Currently we can check for equality in case clause guards

```rust
let y = z
case x {
  1 if x == y -> True
  _ -> False
}
```

It would be useful to also be able to test for ordering

```rust
let y = z
case x {
  1 if x > y -> True
  _ -> False
}
```

Add support for Int ordering opereators, starting with `>` in the first PR.

- [x] GtInt `>`
- [x] GtEqInt `>=`
- [x] LtInt `<`
- [x] LtEqInt `<=`
- [x] GtFloat `>.`
- [x] GtEqFloat `>=.`
- [x] LtFloat `<.`
- [x] LtEqFloat `<=.`

## Implementation hints

See this PR for an example: https://github.com/gleam-lang/gleam/pull/458

### New AST 

The data structure that represents clause guards are defined in the compiler here: https://github.com/gleam-lang/gleam/blob/d35c51843362cf7bd4e173bc9a3afa86a0e28123/src/ast.rs#L292

We would want to add an `GtInt` variant, similar to the `Equals` one.

Basic guard methods would need to be updated for this new variant here: https://github.com/gleam-lang/gleam/blob/d35c51843362cf7bd4e173bc9a3afa86a0e28123/src/ast.rs#L292

### Erlang code generation

Erlang code generation would need to be updated to support the new variant here: https://github.com/gleam-lang/gleam/blob/d35c51843362cf7bd4e173bc9a3afa86a0e28123/src/erl.rs#L487

### Type checking

The type checker would need to be updated to support the new variant here: https://github.com/gleam-lang/gleam/blob/d35c51843362cf7bd4e173bc9a3afa86a0e28123/src/typ.rs#L2328

The new type checker code would be very similar to the code for `ClauseGuard::And` _except_ it would also unify `left` and `right` with `int()` to ensure that Int's are being compared by the `>` operator.

### Parsing

The lalrpop parser will need to be updated to parse the new syntax here: https://github.com/gleam-lang/gleam/blob/d35c51843362cf7bd4e173bc9a3afa86a0e28123/src/grammar.lalrpop#L284-L310

The `ClauseGuard4` statement will need to be renamed to `ClauseGuard5`, and a new `ClauseGuard4` will need to be created by copying `ClauseGuard3`, incrementing the numbers in it, and adding the new operator.

### Testing

[...]

PR #458, opened 2020-04-16
GreaterThan guard clause for Integers

chouzar:
This pull request is meant to solve Issue #455 meant to add greater than guard clauses for integer types:

```gleam
case 1, 0 {
  x, y if x > y -> True
  _, _ -> False
}
```

Adding the new operator was a very mechanical process :) thanks a lot for the thorough explanation.

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -304,6 +304,13 @@ pub enum ClauseGuard<Type> {
         right: Box<Self>,
     },
 
+    GreaterThan {
+        location: SrcSpan,
+        typ: Type,
+        left: Box<Self>,
+        right: Box<Self>
+    },
+
     Or {
         location: SrcSpan,
         typ: Type,
@@ -333,6 +340,7 @@ impl<A> ClauseGuard<A> {
             ClauseGuard::Var { location, .. } => location,
             ClauseGuard::Equals { location, .. } => location,
             ClauseGuard::NotEquals { location, .. } => location,
+            ClauseGuard::GreaterThan { location, ..} => location
         }
     }
 }
@@ -345,6 +353,7 @@ impl TypedClauseGuard {
             ClauseGuard::Var { typ, .. } => typ.clone(),
             ClauseGuard::Equals { typ, .. } => typ.clone(),
             ClauseGuard::NotEquals { typ, .. } => typ.clone(),
+            ClauseGuard::GreaterThan { typ, .. } => typ.clone(),
         }
     }
 }
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -304,7 +304,7 @@ pub enum ClauseGuard<Type> {
         right: Box<Self>,
     },
 
-    GreaterThan {
+    GtInt {
         location: SrcSpan,
         typ: Type,
         left: Box<Self>,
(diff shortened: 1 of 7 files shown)


src/grammar.lalrpop
```diff
@@ -300,6 +300,17 @@ ClauseGuard3: UntypedClauseGuard = {
...
+        location: location(s, e),
+        typ: (),
+        left: Box::new(left),
+        right: Box::new(right),
+    },
+
+    ClauseGuard5 => <>,
+}
+
+ClauseGuard5: UntypedClauseGuard = {
```
chouzar, line 313: I'm really clueless about this file but followed the numbering instructions and it worked just fine :) 

Why are names numbered?

src/typ.rs
```diff
@@ -2396,6 +2396,26 @@ fn infer_clause_guard(
...
+            left,
+            right,
+            ..
+        } => {
+            let left = infer_clause_guard(*left, level, env)?;
+            unify(int(), left.typ(), env)
+                .map_err(|e| convert_unify_error(e, left.location()))?;
+            let right = infer_clause_guard(*right, level, env)?;
+            unify(int(), right.typ(), env)
+                .map_err(|e| convert_unify_error(e, right.location()))?;
```
chouzar, line 2411: Ok so here we are just checking the integer types. But I'm guessing rust is able to use the same comparison operator `>` `<` for many different types 🤔 

If that's the case would it be possible to make a broader check? Kind of like using a `Rust::Numeric` type or the like and checking for "numeric" + same type.

src/grammar.lalrpop
lpil: It's to do with operator precedence. The parser tries one set of operators, then another, then another, etc. :)

src/typ.rs
lpil: Gleam doesn't have any features like operator overloading or type classes so the `>` operator only works on Ints. For floats we use `>.`

We might add something here later, but this is what we have for now :)

src/ast.rs
```diff
@@ -304,6 +304,13 @@ pub enum ClauseGuard<Type> {
         right: Box<Self>,
     },
 
+    GreaterThan {
```
lpil, line 307: Can you name this `GtInt`, to match the `GtInt` node elsewhere in the AST?

src/typ.rs
chouzar: Right, I forget that we're just using Rust to parse valid Gleam.

src/ast.rs
chouzar: Modified symbol!

chouzar:
Should I update the branch from here?
image

lpil:
I prefer not to use that button as it adds a merge commit, something I don't want on master. I'm happy to rebase this in as is :)

Thank you again for you work! This is great

lpil:
Oh! I forgot! Would you mind making another PR to update the CHANGELOG.md file please?

chouzar:
> Oh! I forgot! Would you mind making another PR to update the CHANGELOG.md file please?

Sure! Changelog notes are added once a PR is approved?

And for the rest of operators `>=`, `<`, `<=`, etc., one PR per operator is fine?

lpil:
> Sure! Changelog notes are added once a PR is approved?

I normally try to get them in first but I forgot this time.

> And for the rest of operators >=, <, <=, etc., one PR per operator is fine? Or could they go bundled?

Either or. Might be nice to let someone else do the Float ones if there's interest but I don't really mind :)

chouzar:
Ok! grouping by type sounds sensible :)

PR #459, opened 2020-04-16
Representing strings as utf8 binaries in erlang and tests

### LINKED ISSUE #457: Handling Unicode strings correctly
Currently it is impossible to input some Unicode characters.

In Erlang writing binaries directly can cause their bytes to get truncated, meaning:
`<<256>> == <<0>>` and `<<"🌵">> == <<"5">>`

Erlang has a postfix encoding you can supply to handle this so: 
`<<"🌵"/utf8> /= <<"5">>`

and perhaps more importantly:
```
unicode:characters_to_binary(string:to_graphemes("🤷‍♂️")) == <<"🤷‍♂️"/utf8>>.
```

perhaps this line:
https://github.com/gleam-lang/gleam/blob/99d033aa15622c53bb846e02e2407b8a44b8d69a/src/erl.rs#L227

should be `value.to_doc().surround("<<\"", "\"/utf8>>")` ? 

Is there a better way to handle unicode strings?

here's a little test you can add to check if things are working:
```
pub fn unicode_test() {
    "🌵"
    |> should.not_equal(_, "5")
}
```

itsgreggreg:
This addresses the issue in #457 where high number Unicode strings can't be literally input correctly.

This chooses to represent them as utf8 strings in erlang by compiling the gleam `"🌵"` into the erlang `<<"🌵"/utf8>>`.

The current behavior of compiling to `<<"🌵">>` will actually overflow to `<<"5">>` making it impossible to correctly represent "🌵" in gleam.

With this we should be able to correctly literally input arbitrary unicode strings.

This also adds a `core_language` test suite written in gleam that you can run by:
  - `cd test/core_language`
  - `rebar3 eunit`

fixes #457 

Sorry if this is too much! Lemme know what works for y'all!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Strings are now encoded as utf8 binaries in erlang.
 - HTML documentation can now be generated from Gleam code by running `gleam
   build --doc`.
 - Gleam code can be formatted the `gleam format` command.
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -224,7 +224,7 @@ fn atom(value: String) -> Document {
 }
 
 fn string(value: &str) -> Document {
-    value.to_doc().surround("<<\"", "\">>")
+    value.to_doc().surround("<<\"", "\"/utf8>>")
 }
 
 fn tuple(elems: impl Iterator<Item = Document>) -> Document {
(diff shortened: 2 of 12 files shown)


lpil:
Would you mind also updating the changelog file?

itsgreggreg:
Happy to help! Changelog updated.

lpil:
Thank you!!

PR #462, opened 2020-04-18
Don't treat source chars as single byte

itsgreggreg:
`str.chars().enumerate()` ( https://github.com/gleam-lang/gleam/blob/master/src/parser.rs#L55 ) will enumerate over the count of chars, not their byte position. This means that if the count is ever used as a byte index in the original source (https://github.com/gleam-lang/gleam/blob/master/src/parser.rs#L125 ) not only could the position be incorrect, but it might index directly into a multibyte char causing Rust to panic. I have written a test in `test/core_language/test/unicode_test.gleam` that demonstrates this on an unchanged compiler. The panic output looks like:
```
thread 'main' panicked at 'byte index 2410 is not a char boundary; it is inside 'ð§' (bytes 2408..2412) of 

...source file

src/libcore/str/mod.rs:2154:5
```

The three options here are:
 1. Only support ASCII in source
 1. use `char_indices()` which gives you the char and its byte offset
 1. use something like https://crates.io/crates/unicode-segmentation and treat source as a stream of grapheme clusters

To which my thoughts are:
  1. Not a good idea as then literal strings would also be limited to ascii
  1. Probably a fine solution if syntax is always going to be ascii
  1. Perhaps the most robust, possibly slower, would allow for non ascii syntax

I ran into this panic while adding string functions to the standard library after the recent support of erlang `<<""/utf8>>` binaries.

Happy to remove / change the test or add more info if needed!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -387,6 +387,7 @@ dependencies = [
  "strum_macros 0.17.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "termcolor 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "toml 0.5.5 (registry+https://github.com/rust-lang/crates.io-index)",
+ "unicode-segmentation 1.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "walkdir 2.2.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,6 +39,8 @@ strum_macros = "0.17"
 askama = "0.8"
 # Markdown in docs
 pulldown-cmark = "0.7.0"
+# Support unicode in gleam source
+unicode-segmentation = "1.6.0"
 
 [build-dependencies]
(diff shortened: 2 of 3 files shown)


lpil:
Good find! Thank you!

I actually started implementing the third option but stopped as I was working on too many things at once. I would happily accept that contribution if you would be interested in implementing it.

test/core_language/test/unicode_test.gleam
```diff
@@ -1,5 +1,16 @@
 import should
 
+pub fn a_bunch_of_unicode_test() {
+  let unicode = "👨‍👩‍👧‍👧"
```
lpil, line 4: Seeing as this test doesn't require any code to be run we could test this in the normal compiler tests.

Perhaps `parser::strip_extra_test` as that tests the code that is crashing here.

gamebox:
Yeah I remember us discussing this when I implemented the doc comments parsing.  It would be super helpful change @itsgreggreg

itsgreggreg:
Cool dudes, updated this PR to add treating Gleam source as a stream of Unicode grapheme clusters.

PR #463, opened 2020-04-18
Add >. guard clause for floats

tomwhatmore:
This pull request adds the >. float guard mentioned in #455, and closely follows the GtInt version added in #458. Assuming this first one looks good I can add the other 3 float guards to this same PR.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -525,7 +525,7 @@ impl Documentable for &UntypedClauseGuard {
             }
 
             ClauseGuard::GtFloat { left, right, .. } => {
-                left.as_ref().to_doc().append(" >. ").append(right.as_ref())
+                left.as_ref().to_doc().append(" > ").append(right.as_ref())
             }
 
             ClauseGuard::Var { name, .. } => name.to_string().to_doc(),
--- a/src/format.rs
+++ b/src/format.rs
@@ -524,6 +524,10 @@ impl Documentable for &UntypedClauseGuard {
                 left.as_ref().to_doc().append(" > ").append(right.as_ref())
             }
 
+            ClauseGuard::GtFloat { left, right, .. } => {
+                left.as_ref().to_doc().append(" >. ").append(right.as_ref())
+            }
+
             ClauseGuard::Var { name, .. } => name.to_string().to_doc(),
         }
     }
--- a/src/format.rs
+++ b/src/format.rs
@@ -525,19 +525,19 @@ impl Documentable for &UntypedClauseGuard {
             }
 
             ClauseGuard::GtFloat { left, right, .. } => {
-                left.as_ref().to_doc().append(" > ").append(right.as_ref())
+                left.as_ref().to_doc().append(" >. ").append(right.as_ref())
             }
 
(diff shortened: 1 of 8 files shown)


src/format.rs
```diff
@@ -524,6 +524,10 @@ impl Documentable for &UntypedClauseGuard {
                 left.as_ref().to_doc().append(" > ").append(right.as_ref())
             }
 
+            ClauseGuard::GtFloat { left, right, .. } => {
+                left.as_ref().to_doc().append(" >. ").append(right.as_ref())
```
lpil, line 528: In Erlang the operator is `>`, so we'll want to render that rather than `>.` here.

lpil:
Hi @tomwhatmore ! Thank you for your contribution.

This looks great! I've left one small comment above and then we can merge this in.

Would you mind also updating the CHANGELOG.md file? Something like "The `>.` operator is now supported in Guard clauses".

src/format.rs
tomwhatmore: Well spotted, thank you. Have fixed this below.

tomwhatmore:
Hi @lpil, thanks for the quick response. Have fixed the issue in your comment. I'll add the other 3 float guards and update the change log.

lpil:
Great! Thank you

tomwhatmore:
Have added the other 3 and the change log update. Hopefully I got the Erlang operators right this time!

lpil:
Apologies, I've misled you in my comment above. I have you the right instructions but pointed you to the wrong file. We want the Erlang operators in the erl.rs module and the Gleam ones in the format.rs module.

Sorry about that!

src/grammar.lalrpop
```diff
@@ -311,6 +311,50 @@ ClauseGuard4: UntypedClauseGuard = {
 }
 
 ClauseGuard5: UntypedClauseGuard = {
+    <s:@L> <left:ClauseGuard5> ">." <right:ClauseGuard6> <e:@L> => ClauseGuard::GtFloat {
```
lpil, line 314: To match how the parsing of regular expressions works for these operators we can put all of these new operators in the `ClauseGuard4` block, we don't need to create a new block for each one.


```rust
ClauseGuard4: UntypedClauseGuard = {
    <s:@L> <left:ClauseGuard4> ">" <right:ClauseGuard5> <e:@L> => ClauseGuard::GtInt {
        location: location(s, e),
        typ: (),
        left: Box::new(left),
        right: Box::new(right),
    },

    <s:@L> <left:ClauseGuard4> ">." <right:ClauseGuard5> <e:@L> => ClauseGuard::GtFloat {
        location: location(s, e),
        typ: (),
        left: Box::new(left),
        right: Box::new(right),
    },

  // And so on...

    ClauseGuard5 => <>,
}
```

tomwhatmore:
Great! Thanks a lot for all the assistance.
I'll see if there are any other issues that I might be able to contribute to with minimal hand-holding.

lpil:
Always been free to ping me if you need anything!

Issue #460, opened 2020-04-16
Run test/core_language tests on CI


lpil:
```
cd test/core_language
rebar3 eunit
```

Will need to be a new job in the `build_compiler.yaml` GitHub actions workflow that uses https://github.com/gleam-lang/setup-erlang with the compiler build from the Linux workflow step.

DiodonHystrix:
☝️ working on it

lpil:
Thank you @DiodonHystrix

PR #469, opened 2020-04-19
Run tests for core_language in CI

DiodonHystrix:
Implements #460

As specified in the issue I added a separate job for that. Unfortunately, it comes with a drawback. We have to create an artifact of compiled binary to use it. The drawback is that there is currently no official action to remove the artifacts (they expired after 90days).

I can think about two possible solutions:
1. Run tests in the same job with built compiler.
2. Add one more job to clean artifacts...

Or maybe there is something I'm missing, feel free to point it out.

### DIFF
--- a/.github/workflows/build_compiler.yaml
+++ b/.github/workflows/build_compiler.yaml
@@ -30,6 +30,10 @@ jobs:
       - run: cargo test
       - run: cargo build --release
       - run: target/release/gleam --version
+      - uses: actions/upload-artifact@v1
+        with:
+          name: gleam
+          path: target/release/gleam
 
   build_compiler_macos:
     needs: rustfmt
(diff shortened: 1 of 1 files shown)


test/core_language/rebar.config
```diff
@@ -5,6 +5,10 @@
     {test, [
         {pre_hooks, [{compile, "cargo run -- build ."}]},
         {src_dirs, ["src", "test", "gen/src", "gen/test"]}
+    ]},
+    {gh_actions, [
```
lpil, line 9: What's this new profile for?

lpil:
That's for this. What's the problem with leaving the compiled binary for 90 days?

test/core_language/rebar.config
DiodonHystrix: It's for running tests with prebuilt binary `run: rebar3 as gh_actions eunit` couldn't figure out how to achieve that with `cargo run -- build .` which was building a new one instead.

DiodonHystrix:
I saw some information about an upper limit for a number or a storage size of the artifects. I can't find any information in the official docs, though. So it may not be an issue for public repos after all.

test/core_language/rebar.config
lpil: Oh I see! Thanks!

lpil: On CI seems to be running cargo run still, can you see why that is?

DiodonHystrix:
I had opened up the PR too fast, but I finally got it working with just one file change.

Issue #470, opened 2020-04-19
Refactor ClauseGuard to only have a type field for Var


lpil:
https://github.com/gleam-lang/gleam/blob/113c89fc15753d5f7f4efbec3f65af253a64a07e/src/ast.rs#L280-L287

Currently we store a Type value on all nodes, but the value contained only varies on the Var node. Remove this field from all other nodes and update the typ method to return a constant value rather than the contained field.

tomwhatmore:
I can do this unless you'd rather people only did one `good first issue`?

lpil:
Go for it Tom!

CrowdHailer:
Is there anything remaining on this issue?

lpil:
Well spotted, thanks Peter!

PR #471, opened 2020-04-20
Remove superfluous typ fields from ClauseGuards

tomwhatmore:
Addresses #470

- `ClauseGuard::Var` still returns the stored `Type`
- `ClauseGuard::Int` always returns `int()`
- All the others always return `bool()`

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -281,84 +281,72 @@ pub type TypedClauseGuard = ClauseGuard<Arc<typ::Type>>;
 pub enum ClauseGuard<Type> {
     Equals {
         location: SrcSpan,
-        typ: Type,
         left: Box<Self>,
         right: Box<Self>,
     },
 
     NotEquals {
         location: SrcSpan,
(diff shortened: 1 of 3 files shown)


src/ast.rs
```diff
@@ -4,7 +4,9 @@ mod untyped;
 pub use self::typed::TypedExpr;
 pub use self::untyped::UntypedExpr;
 
-use crate::typ::{self, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor};
+use crate::typ::{
+    self, bool, int, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
```
lpil, line 8: Small style nitpick: Could we not import `bool` and `int` and instead in the function call `typ::int()` and `typ::bool()`? It makes it a little clearer where the function is defined (as the name is pretty generic).

tomwhatmore:
Thanks for the feedback, I have updated it.

PR #472, opened 2020-04-20
Run tests for core_language in CI

DiodonHystrix:


DiodonHystrix:
mb was supposed to test it against my own fork 😅

Issue #475, opened 2020-04-21
Change list prepend syntax to use .. spread operator


lpil:
Current syntax:

```rust
[1, 2 | rest]
```

Desired syntax:

```rust
[1, 2, ..rest]
```

This change is motivated by the desire to have a uniform syntax with #474 

## Requirements

This should be supported in both expressions and patterns.

The formatter should render using the new syntax.

The parser should continue to accept the old syntax so that users can upgrade their code using the formatter.

tomwhatmore:
I can do this one if you want?

lpil:
Yes please

tomwhatmore:
Would be nice if the compiler told you the `|` syntax was deprecated, it doesn't currently emit any warnings right?

lpil:
We don't currently have any way of emitting warnings so do not include that in the first PR. Please open an issue for warnings :)

PR #480, opened 2020-04-22
Change list prepend syntax to use .. spread operator

tomwhatmore:
Addresses #475.
The old syntax is still accepted by the parser, but the formatter renders the new syntax.


tomwhatmore:
Change log updated

lpil:
Thanks!

PR #484, opened 2020-04-22
Register imports before registering types

### LINKED ISSUE #478: Type aliasing fails when aliasing an imported type
Given this module
```rust
// one.gleam
pub type One {
  One
}
```
These modules should compile
```rust
// two.gleam
import one.{One}
type Two = One
```
```rust
// three.gleam
import one
type Three = one.One
```

We should add test cases for these here -> https://github.com/gleam-lang/gleam/blob/3530ecc8133c2ea8bd8410828829e13a9b4060e2/src/project/tests.rs

I believe this may be because we process type aliases prior to imports. We should change the typer to process imports first.

gamebox:
Closes #478


lpil:
Looks nice and clean! Looks like the formatter needs to be run but I think we can merge then.

gamebox:
I goofed here

gamebox:
Fixing

lpil:
Oh, could you add a quick test case in here -> https://github.com/gleam-lang/gleam/blob/3530ecc8133c2ea8bd8410828829e13a9b4060e2/src/project/tests.rs

Thanks!

gamebox:
Formatted and with tests

PR #489, opened 2020-04-23
Continue build when rust fmt fails

### LINKED ISSUE #485: Don't require rustfmt job to pass on CI to run tests
https://github.com/gleam-lang/gleam/blob/master/.github/workflows/build_compiler.yaml

If the formatting is off we should still run the tests.

gamebox:
Closes #485


lpil:
Great, thank you. Could we also remove the code that makes the other jobs depend upon rustfmt? I think we can run them all at the same time too for a speed boost

gamebox:
There you go!

Issue #490, opened 2020-04-24
gleam build outputs colour codes for non-ttys


michaeljones:
I'm not sure if I'm using this terms correctly but if I run `gleam build .` from within neovim as a `:make` command then the output is filled with unhandled colour codes.

From my limited understanding it is possible to check if the output is going to a tty or not and if it isn't we might consider not outputting the colour codes. I think it might be configurable in `codespan_reporting::term::Config` though I'm not sure how easy it is make work.

Is that something that you consider gleam doing?

michaeljones:
I believe the check can be done with this crate: https://docs.rs/atty/0.2.11/atty/ which is already in gleam's cargo.lock file.

lpil:
Sounds like a good addition!

In your case specifically are you running make from the Neovim terminal? If so colour codes ought to be supported. Can you share the output that is printed? Thank you

michaeljones:
Thanks for the quick response. I was experimenting with trying to add a `compiler` setting for `gleam`. I'm not massively familiar with it but the rust.vim set up allows you to set `:compiler cargo` and then run `:make check` and it'll do the check and allow you to jump between errors in vim. I suspect it is all meant to be superseded by the language server stuff these days but I'm quite fond of the flow. 

I've added a file into `gleam.vim` called `compiler/gleam.vim` with the following contents:
```
if exists('current_compiler')
    finish
endif

runtime compiler/rustc.vim

let current_compiler = "gleam"

if exists(':CompilerSet') != 2
    command -nargs=* CompilerSet setlocal <args>
endif

CompilerSet makeprg=gleam\ $*
```
And then in my personal configuration a file called `ftplugin/gleam.vim` with:
```
" Gleam Specific
" --------------

" Only do this when not done yet for this buffer
if exists("b:did_ftplugin")
  finish
endif

compiler gleam

map <leader>m :make build .<CR>
```
And when I run `<leader>m` I see:
```
:!gleam build . 2>&1| tee /tmp/nvimRVwmtV/1                                                                                                                                               
                                                                                                                                                                                          
^[[0m^[[1m^[[38;5;9merror^[[0m^[[1m: Syntax error^[[0m                                                                                                                                    
                                                                                                                                                                                          
   ^[[0m^[[34m┌^[[0m^[[0m^[[34m──^[[0m /home/michael/root/projects/gleam-phoenix-mix/src/page_controller.gleam:7:27 ^[[0m^[[34m───^[[0m                                                   
[...]

PR #492, opened 2020-04-24
Check for tty for color codes in errors

michaeljones:
This is to address color codes being written out when there is no tty (#490). It isn't complete but I wanted to check that you're happy with the approach before trying to adjust other errors.

### Commit message

We use the "atty" crate which was already a transitive dependency to check whether standard-error is a tty before deciding whether or not to use colors for errors.

This attempts to follow the essence of the logic used in rustc at the moment:

  https://github.com/rust-lang/rust/blob/7f3df5772439eee1c512ed2eb540beef1124d236/src/librustc_errors/emitter.rs#L533-L544

Running `gleam build . | cat` still inserts color codes which is the same for `cargo build` but running `gleam build . 2>&1 | cat` does not as we're redirecting standard-error too. This also means that it does not insert color codes when run as part of a vim `:make` command which was the initial motivation for this change.

This doesn't cover all the error print outs but that can be done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
   but will continue to work for now. The formatter will output the new syntax.
 - Add new assert syntx for binding variables `assert Ok(x) = result`. In the future
   this will allow you to use a pattern that does not match all values.
+- Color codes are now only emitted in error output for interactive terminal sessions.
 
 ## v0.7.1 - 2020-03-03
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,6 +367,7 @@ name = "gleam"
 version = "0.8.0-dev"
 dependencies = [
  "askama 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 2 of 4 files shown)


lpil:
Would you mind doing the same for this one too? https://github.com/gleam-lang/gleam/blob/88322206418d76f6088ec4ba7d2e072301da9289/src/error.rs#L10

michaeljones:
Done. I'm pretty new to Rust so, whilst I realise this change is tiny, I'd be happy if there is any feedback or adjustments or anything. I'd be happy to learn.

lpil:
Would you mind also updating the CHANGELOG file please?

michaeljones:
Done. Thanks for the prompting. Again, happy if you'd like to reword. Not sure how technical to go and the more technical the language the less confident I am about it in the particular case :) Terminals & tty is a weird world to me despite using them for years :)

lpil:
I think you've made it much clearer than I could have! Thanks once again 💜

lpil:
There was a conflict so I manually rebased these commits onto master to save you having to resolve them. Cheers!

Issue #491, opened 2020-04-24
Raise error if the same variable is used multiple times in a pattern.


CrowdHailer:
Best explained with an example

```rust
case tuple(3, 5) {
  tuple(x, x) -> 
  tuple(x, y) ->
}
```

Currently this will match on the first branch, the value will be 5. This is because Gleam allows variable rebinding.

In both Elixir and Erlang the first branch will only match if the values are equal. As gleam doesn't support this is should be an error to use the same variable twice in one match

lpil:
I agree! Thanks Peter

PR #494, opened 2020-04-25
Raise error if the same variable is used multiple times in a pattern

tomwhatmore:
Addresses #491


tomwhatmore:
Updated the changelog and resolved the conflicts

Issue #474, opened 2020-04-21
Shorthand for discarding remaining fields in a record pattern


lpil:
```rust
type Triple {
  Triple(a: Int, b: Int, c: Int)
}
```
```rust
let Triple(a: the_a, ..) = triple
```

lpil:
@tomwhatmore Did you say you were doing this?

tomwhatmore:
I did. Made some progress last night, hoping to get it finished this evening.

lpil:
Great, thanks! No rush, I just wanted to make sure no one else picked this up.

PR #495, opened 2020-04-25
Support .. operator when binding to record fields

tomwhatmore:
Addresses #474

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -665,6 +665,7 @@ impl<'a> Formatter<'a> {
                 name,
                 args,
                 module: None,
+                with_spread: false,
                 ..
             } => name
                 .to_string()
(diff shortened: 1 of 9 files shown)


src/format.rs
```diff
@@ -867,6 +894,25 @@ where
...
+    if let None = args.peek() {
+        return "()".to_doc();
+    }
+
+    break_("(", "(")
+        .append(concat(args.intersperse(delim(","))))
+        .append(break_(",", ", "))
+        .append("..")
+        .nest(INDENT)
+        .append(break_("", ""))
```
lpil, line 911: Can we have a trailing semicolon?

```rust
break_(",", "")
```

tomwhatmore: Do you mean comma?

lpil: Sorry yes, a comma :)

tomwhatmore: Done!

lpil:
I've rebased these commits onto master as I want to avoid merge commits. Thanks!

Issue #501, opened 2020-04-25
Allow discard to be omitted in list prefix


lpil:
```rust
[x, .._]
```
Becomes
```rust
[x, ..]
```

The formatter should use this syntax

tomwhatmore:
I'm on it

PR #504, opened 2020-04-26
Allow discard to be omitted in list tail

tomwhatmore:
Addresses #501

### DIFF
--- a/src/format/tests.rs
+++ b/src/format/tests.rs
@@ -2289,4 +2289,42 @@ fn module_rewrites_test() {
 }
 ",
     );
+
+    // Formatter removes discard list prefix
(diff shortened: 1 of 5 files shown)


src/format/tests.rs
```diff
@@ -2247,3 +2247,84 @@ fn main() {
...
     );
 }
+
+#[test]
+fn module_rewrites_test() {
+    macro_rules! assert_format_rewrite {
+        ($src:expr, $output:expr  $(,)?) => {
+            assert_eq!(pretty($src).unwrap(), $output);
+        };
+    }
```
lpil, line 2257: Nice!

PR #511, opened 2020-04-27
Always show types for function arguments in generated docs

tomwhatmore:


src/format.rs
```diff
@@ -644,11 +645,22 @@ impl<'a> Formatter<'a> {
...
             .append(typ::pretty::Printer::new().to_doc(return_type.as_ref()))
     }
 
+    // Like fn_args but will always print the types, even if they were implicit in the original source
+    pub fn docs_fn_args(&mut self, args: &[TypedArg]) -> Document {
+        let mut printer = Printer::new();
+        wrap_args(args.iter().map(|arg| {
+            arg.names
+                .to_doc()
+                .append(": ".to_doc().append(printer.to_doc(&arg.typ)))
```
lpil, line 659: Oh sweet, I forgot the type is already there too so we don't even need to zip it.

PR #512, opened 2020-04-27
Custom Type Constructors in Docs

### LINKED ISSUE #483: Include custom type constructors in generated HTML docs
https://github.com/gleam-lang/gleam/blob/7d2f1ae988e2c05608cfc94d7fd3a38909b505a9/templates/documentation_module.html#L41-L77

We'll need to add constructor information to this struct. https://github.com/gleam-lang/gleam/blob/7d2f1ae988e2c05608cfc94d7fd3a38909b505a9/src/doc.rs#L206-L211

We do not need to implement links to types at this point.

gamebox:
Closes #483

### DIFF
--- a/src/doc.rs
+++ b/src/doc.rs
@@ -53,7 +53,6 @@ pub fn generate_html(
     // Generate module documentation pages
     for module in modules {
         let name = module.name.join("/");
-        println!("{}", name);
         let template = ModuleTemplate {
             unnest: module.name.iter().map(|_| "..").intersperse("/").collect(),
             links,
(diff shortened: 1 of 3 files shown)


src/doc.rs
```diff
@@ -71,6 +72,7 @@ pub fn generate_html(
             types: {
                 let mut t: Vec<_> = module.ast.statements.iter().flat_map(type_).collect();
                 t.sort();
+                println!("{:#?}", t);
```
lpil, line 75: Debug print left in here

src/doc.rs
```diff
@@ -53,6 +53,7 @@ pub fn generate_html(
     // Generate module documentation pages
     for module in modules {
         let name = module.name.join("/");
+        println!("{}", name);
```
lpil, line 56: And here

Issue #107, opened 2019-03-17
Display module qualifiers when type errors would be otherwise unclear


lpil:
`Map` and `Map` print the same, thought they may be `app.Map` and `lib.Map`. This can result in confusing errors.

It would be nice if we show as few qualifiers as possible for conciseness. Some design work is to be done here.

zjhmale:
I don't think we can have two `Map`s in a single module, this is forbidden in other languages IIRC 🤔

lpil:
Here's an example

```rust
// one.gleam
pub type Thing {
  Thing
}
```

```rust
// two.gleam

import two

pub type Thing {
  Thing
}

pub fn main() {
  Map == one.Map // Type error
}
```

It should be made clear that the two different types named Map are not from the same module.

zjhmale:
Gocha, let me make it clear, pls don't blame if I got wrong 😅  

```rust
// one.gleam
pub type Thing {
  Thing
}
```

```rust
// two.gleam

import one

pub type Thing {
  Thing
};

pub fn main() {
  Thing == one.Thing // Type error
}
```

at present, gleam will display

```
error: Type mismatch

    ┌── /Users/jihuizheng/Documents/gleam-ttt/ttt/src/ttt.gleam:29:12 ───
    │
 29 │   Thing == one.Thing // Type error
    │            ^^^^^^^^^
    │

Expected type:

    Thing

Found type:

    Thing
```

and the ideal output will be

```
error: Type mismatch

    ┌── /Users/jihuizheng/Documents/gleam-ttt/ttt/src/ttt.gleam:29:12 ───
    │
 29 │   Thing == one.Thing // Type error
    │            ^^^^^^^^^
    │

Expected type:

    two.Thing

Found type:

    one.Thing
```

correct?

lpil:
Yes, something like that.

My original idea was to show a qualifier when the type originates from another module but I'm now more sure that is the right approach as it would introduce a lot of noise. It would be good to only show the qualifier if it is the part of the type that is the cause of the error.

Kn99HN:
> To do this we would want to iterate over the document before printing and record which names are used multiple times.

I want to tackle this issue. But I'm unsure about the approach suggested here. I have an idea of where the error is happening. 

Are you suggesting to iterate over the documents before type checker runs? Would you mind elaborating? Thank you!

lpil:
Hello! Thank you for helping

[...]

PR #513, opened 2020-04-27
Display module qualifiers when type errors would be otherwise unclear

tomwhatmore:
Addresses #107 
The `Error` itself is unchanged, it's just the pretty output that has been updated, so I'm not sure how to provide test cases for the new behavior. If you have any suggestions on that front let me know and I can add them to this PR.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,4 +1,5 @@
 use crate::typ::pretty::Printer;
+use crate::typ::Type;
 use itertools::Itertools;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


lpil:
I'm not sure this is enough as the problem may not be at the top level.

For example, we could have an error from unifying `List(one.Element)` and `List(two.Element)`

tomwhatmore:
Yeah fair enough, this isn't going to cut it. I'll close this and maybe we can come back to it later

PR #514, opened 2020-04-27
gleam format --check

### LINKED ISSUE #509: gleam format --check
Check if files are formatted correctly but do not write them.

gamebox:
Closes #509


gamebox:
This is the current behavior.  It never writes a file, only printing an error for every file that needs formatting.

PR #521, opened 2020-04-30
Return more clear error when expecting functions

### LINKED ISSUE #510: Compiler errors not clear for type mismatches when expecting functions
Here is an example.

```rust
fn bar() -> Int {
    5
}

fn run(foo: fn() -> String) {
    foo()
}

fn demo() {
    run(bar)
}
```

Error message 

```
Expected type:

    String

Found type:

    Int
```

I would expect it to say Expected Type `fn () -> String` or have a message about expected return type.

gamebox:
Closes #510

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,13 @@
 ## Unreleased
 
 - Newly generated projects use stdlib v0.8.0.
+- Will give a clearer error when a function given as an argument to another function
+  doesn't match the type of the parameter.
 
 ## v0.8.0-rc1 - 2020-04-28
 
 - Strings are now encoded as utf8 binaries in the generated Erlang.
-- HTML documentation can now be generated from Gleam code by running `gleam
-  build --doc`.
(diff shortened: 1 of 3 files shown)


gamebox:
Example error, using above example(this is in the test)

```
error: Type mismatch

    ┌── :10:18 ───
    │
 10 │              run(bar)
    │                  ^^^
    │

Expected type:

    fn() -> String

Found type:

    fn() -> Int
```

lpil:
Fantastic! Could we also have a test for when the arguments are different? For example `fn(Int) -> Int != fn(String) -> Int`

Thank you

gamebox:
Example output for that case:

```
error: Type mismatch

    ┌── :10:18 ───
    │
 10 │              run(bar)
    │                  ^^^
    │

Expected type:

    fn(String) -> Int

Found type:

    fn(Int) -> Int
```

gamebox:
Sure!

lpil:
I've rebased these commits onto master, thank you!

Issue #479, opened 2020-04-22
Support compiler warnings


tomwhatmore:
Would be useful for the compiler to be able to omit warnings. Specifically, deprecation notices would be nice when there are syntax changes such as in #475.

I have no idea how much work this would be as I haven't even looked into it yet, but getting this issue up as suggested by @lpil in case anyone has thoughts on the matter.

lpil:
My first thought is this:

We have an enum to represent all possible warnings

```rust
pub enum Warning {
  DeprecatedListPrependSyntax {
    path: PathBuf,
    location: SrcSpan,
  }
}
```

And then the `infer_module` function can return a vector of these warnings. The compile then prints them after compilation, before printing any errors.

tomwhatmore:
That sounds like a good approach to me. Not sure how complicated the pretty formatting is but presumably it could mostly reuse the current error printing logic but with some slightly different colours?

lpil:
Yes that sounds good. We could do the warning printing as a second step later if that's easier, as to not get in the way of the warning collection.

Issue #157, opened 2019-04-20
Incremental compilation


lpil:
May be good to investigate Salsa for this.

gamebox:
Can you assign this to me @lpil just so others know it's active?

gamebox:
Just to clarify should I drop development on this?

gamebox:
My plan was to work on this mostly while I am on paternity leave at the end of June - july

lpil:
I'm not sure how we should approach this feature yet so it would be good to try and do this iteratively if possible, it would make it easier to review and to make design decisions. Happy for you to take the lead though!

Also congratulations! How exciting!

gamebox:
Thanks!  Sorry this was meant for the LSP ticket 🤦‍♂️

gamebox:
This feature is almost working.  Might be working tonight.  Will take me a few days to get the PR together.

gamebox:
I have a working prototype working.  Am testing current compiler invariants against the behavior I see in the new straight-line compiler powered by Salsa.

I want to structure the PRs and the release process as such:

1. PR for the Salsa system itself, completely isolated(minus a couple extra derives and functions needed)
2. PR for adding a flag to compile with the Salsa system(or env var? maybe SALSA=1)
3. PR for adding a flag to incrementally compile(maybe only with SALSA enabled?)
4. PR for running compiler tests with SALSA enabled
5. When all looks good, PR for dropping SourceTree and removing SALSA env var.

Let me know how this looks to you @lpil

lpil:
Sounds good, though I think we may be able to skip the env var step- I don't think we need both systems at once for any reason.

gamebox:
I would usually blue/green a feature like this, but not doing it cuts it down quite a bit

lpil:
Yeah it's a great technique, but I don't think we have enough Gleam code in the world for it to be effective just yet, so if we can cut down the amount of work to do that would be my preference.

lpil:
Gunna close this for more precise issues later

PR #524, opened 2020-04-30
Compiler can now emit warnings

tomwhatmore:
Addresses #479

### DIFF
--- a/src/project.rs
+++ b/src/project.rs
@@ -95,15 +95,15 @@ pub fn analysed(inputs: Vec<Input>) -> Result<Vec<Analysed>, Error> {
 
         println!("Compiling {}", name_string);
 
-        let mut warnings = Vec::new();
         let (result, mut module_warnings) = crate::typ::infer_module(module, &modules_type_infos);
-        module_warnings.drain(..).for_each(|warning| {
-            warnings.push(Warning::Type {
+        let warnings = module_warnings
+            .drain(..)
+            .map(|warning| Warning::Type {
                 path: path.clone(),
                 src: src.clone(),
                 warning,
             })
-        });
+            .collect();
 
         let ast = result.map_err(|error| Error::Type {
             path: path.clone(),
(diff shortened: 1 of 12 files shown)


lpil:
This looks good! Thanks!

I want the warnings to be emitted whether the module succeeds compilation or not, and I don't think we should cache the warnings in #157 , so I don't think we should be storing them on the module.

For this PR we could have the module typing function return `(Result<a, b>, Vec<Warning>)`, though later I think we could either create some kind of telemetry object and pass it in, or we can have it as a property on Env (which I would like to restructure to be an exposed object called Typer, similar to the Formatter and Printer)

tomwhatmore:
Took the warnings out of the modules and now returning them seperately from `infer_module()`

lpil:
Build is failing FYI

tomwhatmore:
Whoops! Not sure what happened there, I'll fix it when I've made some
dinner.

tomwhatmore:
Apparently I deleted the warnings field from the generated `grammar.rs` file instead of `grammar.lalrpop`. Blunder.

lpil:
A classic move! I have to work out how to get my editor from automatically opening that file when it has an error in it.

src/project.rs
```diff
@@ -92,8 +95,21 @@ pub fn analysed(inputs: Vec<Input>) -> Result<Vec<Analysed>, Error> {
 
         println!("Compiling {}", name_string);
 
-        let ast = crate::typ::infer_module(module, &modules_type_infos)
-            .map_err(|error| Error::Type { path, src, error })?;
+        let mut warnings = Vec::new();
+        let (result, mut module_warnings) = crate::typ::infer_module(module, &modules_type_infos);
+        module_warnings.drain(..).for_each(|warning| {
```
lpil, line 100: Could this be `module_warnings.map` ? Would be slightly more concise and the vector wouldn't be grown as new warnings are added.

tomwhatmore: Yeah, probably. `.map` was being a pain when it was part of the module struct but might be ok now. Will give it a go.

lpil: Thanks

tomwhatmore: Yeah that works fine, I was just being dumb earlier. Updated.

lpil: I'm sure you wern't :)

Issue #533, opened 2020-05-01
Warn when a variable is never used


lpil:

michallepicki:
@itsgreggreg @lpil Did https://github.com/gleam-lang/gleam/pull/961 resolve this?

lpil:
Thank you

Issue #532, opened 2020-05-01
Warn when a private function is never referenced


lpil:

michallepicki:
@itsgreggreg @lpil Similarily here, did https://github.com/gleam-lang/gleam/pull/961 resolve this?

lpil:
Thank you

PR #539, opened 2020-05-02
Warn when a private function is never called

tomwhatmore:


src/typ.rs
```diff
@@ -1626,11 +1630,24 @@ pub fn infer_module(
     // Remove private and imported types and values to create the public interface
     env.module_types
         .retain(|_, info| info.public && &info.module == module_name);
-    env.module_values.retain(|_, info| info.public);
+
+    // Move the hashmap of module values out of the env struct, then iterate over it.
+    // Create warnings for uncalled private functions
+    // Finally, keep only the public values.
+    let mut module_values = std::mem::replace(&mut env.module_values, HashMap::new());
```
lpil, line 1637: Couldn't we keep this hashmap on env as before?

src/typ/tests.rs
```diff
@@ -1773,6 +1773,7 @@ fn infer_module_warning_test() {
             ast.name = vec!["my_module".to_string()];
             let (_, warnings) = infer_module(ast, &HashMap::new());
 
+            assert_eq!(true, warnings.len() > 0);
```
lpil, line 1776: Can write `assert!(warning.len() > 0)`, I think it might have a slightly clearer error message.

lpil:
I'm thinking about how we expand this to include local variables and I'm wondering if it would be easier to track the names that have not been referenced rather than the ones that have.

```rust
fn one() { 1 }

fn two() {
  let one = 1
  one + one
}
```

In this code we have referenced "one", but the function "one" was never called. We would need to have a new set at each new scope in order to handle variable shadowing.

Having said that, I think both these approaches will fail for recursively referenced types and functions (we don't have this for functions yet).

```rust
fn a() { b() }
fn b() { a() }

type X {
  X(Y)
}
type Y {
  Y1
  Y2(X)
}
```

These are all referenced by another private type or value, yet they are not used as we would want to warn for them.

This is suddenly looking a lot like memory GC to me. Perhaps we want to create a call-graph for all named values, iterate over the graph starting at the root, and then warn for nodes we don't see during the traversal.

src/typ/tests.rs
tomwhatmore: Yeah that would be loads better, will do.

src/typ.rs
tomwhatmore: We probably can but this seemed like an easier way to make the borrow checker happy. Trying to add warnings to the env inside the retain was giving me the old multiple mutable references. I'm sure I can wrestle it into submission if you'd rather keep it on the env.

tomwhatmore:
I was thinking of using this approach for private functions and the 3 never referenced ones. Unused variables seems like it would be a bit more involved but I hadn't really got to thinking it through properly. Do you think there would be value in adding in naive versions of these warnings that use the approach in this PR (or similar), or would you rather wait and go for a full call-graph approach?

lpil:
I think I would rather wait until we have a solution that works for both this feature and the future requirements, given they are known.

tomwhatmore:
Ok, I'll close this then. Would it be worth collating all those never called / never referenced warning issues into a new call-graph one?

QuinnWilton:
The above comment (https://github.com/gleam-lang/gleam/pull/539#issuecomment-623075436) is good context to keep in mind if anyone takes a stab at implementing #532 #533 #534 #535 or #536

I'm just commenting to link those issues to this PR.

lpil:
I'm probably being annoying here, but I'm wondering if this is worth reviving in order to get _some_ of this functionality, even if it's not the perfect solution.

PR #540, opened 2020-05-02
Log warnings on implicitly discarded results

### LINKED ISSUE #531: Warn when a result is discarded in a Seq
Warn:

```rust
function_that_returns_a_result() // This result value is being implicitly discarded
something_else()
```

Ok:
```rust
let _ = function_that_returns_a_result() // This result value is being excplicitly discarded
something_else()
```

QuinnWilton:
Please feel free to heavily critique any of this code if it isn't idiomatic Rust, I don't know what I'm doing :)

Resolves #531

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -2110,7 +2110,7 @@ fn infer_tuple_index(
 ) -> Result<TypedExpr, Error> {
     let tuple = infer(tuple, level, env)?;
 
-    match &*tuple.typ() {
+    match tuple.typ().as_ref() {
         Type::Tuple { elems } => {
             let typ = elems
                 .get(index as usize)
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -38,6 +38,13 @@ pub enum Type {
 }
 
 impl Type {
+    pub fn is_result(&self) -> bool {
+        match self {
+            Type::App { name, module, .. } if "Result" == name && module.is_empty() => true,
+            _ => false,
+        }
+    }
+
     pub fn is_unbound(&self) -> bool {
         match self {
             Type::Var { typ } => typ.borrow().is_unbound(),
@@ -1000,6 +1007,8 @@ pub enum Warning {
     DeprecatedListPrependSyntax { location: SrcSpan },
 
     Todo { location: SrcSpan },
+
+    ImplicitlyDiscardedResult { location: SrcSpan },
 }
 
 #[derive(Debug, PartialEq)]
@@ -1960,6 +1969,17 @@ fn infer_seq(
 ) -> Result<TypedExpr, Error> {
     let first = infer(first, level, env)?;
     let then = infer(then, level, env)?;
+
+    match first.typ().as_ref() {
+        typ if typ.is_result() => {
+            env.warnings.push(Warning::ImplicitlyDiscardedResult {
+                location: first.location().clone(),
+            });
+        }
+
+        _ => {}
+    }
+
     Ok(TypedExpr::Seq {
         typ: then.typ(),
         first: Box::new(first),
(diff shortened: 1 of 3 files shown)


src/typ.rs
```diff
@@ -38,6 +38,13 @@ pub enum Type {
 }
 
 impl Type {
+    pub fn is_result(&self) -> bool {
+        match self {
+            Type::App { name, .. } if &*name == "Result" => true,
```
QuinnWilton, line 43: Is this the most idiomatic way of checking `String` equality? I was getting really mixed answers from my Googling

lpil, line 43: A style thing: Can we swap `&*name` for `name.as_str()`? I find it a little easier to read.

Also can we check that the module `is_empty`? In case someone implemented a type called Result in some other module.

src/typ.rs
```diff
@@ -1960,6 +1969,17 @@ fn infer_seq(
 ) -> Result<TypedExpr, Error> {
     let first = infer(first, level, env)?;
     let then = infer(then, level, env)?;
+
+    match &*first.typ() {
```
lpil, line 1973: Same style thing again, can we swap `&*` for `as_ref()`? Thank you

QuinnWilton: Do you want me to change this line too (in another commit, of course): https://github.com/gleam-lang/gleam/blob/master/src/typ.rs#L2093

src/warning.rs
```diff
@@ -50,6 +50,17 @@ impl Warning {
...
+                ImplicitlyDiscardedResult { location } => {
+                    let diagnostic = Diagnostic {
+                        title: "Implicitly discarded Result".to_string(),
+                        label: "".to_string(),
+                        file: path.to_str().unwrap().to_string(),
+                        src: src.to_string(),
+                        location: location.clone(),
+                    };
+                    write(buffer, diagnostic, Severity::Warning);
+                    writeln!(buffer, "This code implicitly ignores a Result returned from an expression. Instead, bind the Result to _ if you intend to discard it.").unwrap();
```
lpil, line 62: One of my goals with Gleam is to try and make strong static typing more accessible, so I'd like to try and make the error and warning messaging as newcomer friendly as possible.

How about something more like this:

### Unused result value

The Result value returned by this code is not being used, so any error is being silently ignored.
Check for an error with a case statement, or assign it to the variable `_` if you are sure the error does not matter.

What do you think?

src/typ.rs
lpil: Yes please!

QuinnWilton: It actually looks like the `AsRef` trait isn't implemented on `TypedExpr`. Do you want me to implement it and then use `as_ref()`?

src/warning.rs
QuinnWilton: I like that! I'm a big fan of Elm-style error messages, and think making static typing more accessible is a noble goal. I was hoping you'd have feedback on the copy :)

src/typ.rs
lpil: The left is a String and the right is a &str.

`&*name` is a converstion from String to &str, so it's then &str == &str

I think you might not need to do this if you swap the left and right hand side? I can't remember how the Eq and PartialEq traits work...

I think the string types are badly named in Rust. It would be clearer if `str` was called `string` and `String` was called `StringBuf` (StringBuffer).

QuinnWilton: Thanks for the explanation! Swapping the left and right side works too it seems

QuinnWilton:
I've addressed all of your comments except the `&*` one. I'll keep poking around, but I need to do some more reading to understand why `AsRef` isn't implemented for the type in question / what I need to do to address that.

src/typ.rs
lpil: Sorry I've been unclear. It's implemented on `Arc` (a fancy pointer type) which `typ()` returns so `first.typ().as_ref()`

QuinnWilton: Ah, that makes more sense! Thanks, I pushed up the fix.

PR #548, opened 2020-05-03
Implement --check flag for gleam format --stdin

QuinnWilton:
The first commit also addresses some permission errors I was getting while trying to format files on Mac OS.

Here's what using the command looks like: https://gist.github.com/QuinnWilton/5fcb8b5a1ce8829dd851427f6cd150ff

I tried to keep my changes in `format::command` to a minimum, but if you'd like me to do some refactoring to DRY up some of the shared logic between the stdin and non-stdin case, let me know and I'm happy to :)

I also didn't see existing tests for the `format::command` anywhere, so I wasn't sure where to start with testing this. If you have suggestions I'm happy to give that a try.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -1061,7 +1061,7 @@ but it cannot be found.",
             Error::Format { problem_files } => {
                 let mut files: Vec<_> = problem_files
                     .iter()
-                    .flat_map(|formatted| formatted.path.to_str())
+                    .flat_map(|formatted| formatted.source.to_str())
                     .map(|p| format!("  - {}", p))
                     .collect();
                 files.sort();
--- a/src/format/command.rs
+++ b/src/format/command.rs
@@ -6,21 +6,28 @@ use std::str::FromStr;
 
 #[derive(Debug, PartialEq)]
 pub struct Formatted {
-    pub path: PathBuf,
+    pub source: PathBuf,
+    pub destination: Destination,
     pub input: String,
     pub output: String,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum Destination {
+    Stdout,
+    File { path: PathBuf },
+}
+
 pub fn run(stdin: bool, check: bool, files: Vec<String>) -> Result<(), Error> {
-    if stdin {
-        format_stdin(check)
+    let formatted = if stdin {
+        vec![read_and_format_stdin()?]
     } else {
-        let formatted = read_and_format_paths(files)?;
-        if check {
-            check_formatting(formatted)
-        } else {
-            write_formatted(formatted)
-        }
+        read_and_format_paths(files)?
+    };
+    if check {
+        check_formatting(formatted)
+    } else {
+        write_formatted(formatted)
     }
 }
 
@@ -39,21 +46,34 @@ fn check_formatting(formatted_files: Vec<Formatted>) -> Result<(), Error> {
 
 fn write_formatted(formatted_files: Vec<Formatted>) -> Result<(), Error> {
     for formatted in formatted_files {
-        let path = formatted.path;
-        let mut f = File::create(&path).map_err(|e| Error::FileIO {
-            action: FileIOAction::Create,
-            kind: FileKind::File,
-            path: path.clone(),
-            err: Some(e.to_string()),
-        })?;
+        match formatted {
+            Formatted {
+                destination: Destination::Stdout,
+                output,
+                ..
+            } => {
+                print!("{}", output);
+            }
+            Formatted {
+                destination: Destination::File { path },
+                output,
+                ..
+            } => {
+                let mut f = File::create(&path).map_err(|e| Error::FileIO {
+                    action: FileIOAction::Create,
+                    kind: FileKind::File,
+                    path: path.clone(),
+                    err: Some(e.to_string()),
+                })?;
 
-        f.write_all(formatted.output.as_bytes())
-            .map_err(|e| Error::FileIO {
-                action: FileIOAction::WriteTo,
-                kind: FileKind::File,
-                path: path.clone(),
-                err: Some(e.to_string()),
-            })?;
+                f.write_all(output.as_bytes()).map_err(|e| Error::FileIO {
+                    action: FileIOAction::WriteTo,
+                    kind: FileKind::File,
+                    path: path.clone(),
+                    err: Some(e.to_string()),
+                })?;
+            }
+        }
     }
 
     Ok(())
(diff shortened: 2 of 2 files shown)


src/error.rs
```diff
@@ -1073,6 +1083,16 @@ but it cannot be found.",
                     label,
                 };
 
+                write_project(buffer, diagnostic);
+            }
+            Error::Format {
+                input: FormatInput::STDIN,
+            } => {
+                let diagnostic = ProjectErrorDiagnostic {
+                    title: "STDIN has not been formatted".to_string(),
```
lpil, line 1092: I had a look around and it seems the canonical name is stdin (lowercase), at least according to GNU, Microsoft and the C++ reference.

Could we make this and the constructor lowercase? Thanks

src/format/command.rs
```diff
@@ -114,10 +118,20 @@ pub fn format_stdin(_check: bool) -> Result<(), Error> {
...
+            Err(Error::Format {
+                input: FormatInput::STDIN,
+            })
+        } else {
+            Ok(())
+        }
+    } else {
+        print!("{}", formatted);
+        Ok(())
+    }
```
lpil, line 136: Can we move this `if` to the `run` function, to match the logic for formatting files?

src/format/command.rs
```diff
@@ -33,14 +33,18 @@ fn check_formatting(formatted_files: Vec<Formatted>) -> Result<(), Error> {
...
+            },
+        })
     }
 }
 
 fn write_formatted(formatted_files: Vec<Formatted>) -> Result<(), Error> {
     for formatted in formatted_files {
         let path = formatted.path;
-        let mut f = File::open(&path).map_err(|e| Error::FileIO {
+        let mut f = File::create(&path).map_err(|e| Error::FileIO {
```
lpil, line 47: What does this change do?

src/error.rs
```diff
@@ -1073,6 +1083,16 @@ but it cannot be found.",
...
+                    label: "".to_string(),
```
bausano, line 1093: Alternatively `String::new()`

src/format/command.rs
```diff
@@ -33,14 +33,18 @@ fn check_formatting(formatted_files: Vec<Formatted>) -> Result<(), Error> {
     if problem_files.is_empty() {
         Ok(())
     } else {
-        Err(Error::Format { problem_files })
+        Err(Error::Format {
+            input: FormatInput::Files {
```
bausano, line 37: Can be shortened to `input: FormatInput::Files { problem_files }`.

src/format/command.rs
```diff
@@ -103,7 +107,7 @@ fn format_file(path: PathBuf) -> Result<Formatted, Error> {
     })
 }
 
-pub fn format_stdin(_check: bool) -> Result<(), Error> {
+pub fn format_stdin(check: bool) -> Result<(), Error> {
```
bausano, line 110: What does the `check` argument refer to? Is there a more descriptive name?

lpil: It refers to the `--check` flag. I've requests that this argument be removed from this function and moved to the higher level.

src/format/command.rs
QuinnWilton: This opens the file with write permissions. Using `File::open` I was getting permission errors when it tried to write the newly formatted file, since `File::open` only requests read permissions.

lpil: Thank you

QuinnWilton:
I pushed up a new commit that addresses your comments. It's a bit of a bigger refactor, since I shifted the complexity of handling the stdin / file case out of the error handling layer, and into `format::command`, by introducing `source` and `destination` fields onto the `Formatted` struct.

This feels like a more natural place for this information to live, but I've left the previous version intact if you'd prefer I revert :)

I'll squash the last 3 commits together if everything looks good -- no sense having my rough work in the tree.

src/format/command.rs
```diff
@@ -6,21 +6,29 @@ use std::str::FromStr;
...
 pub struct Formatted {
-    pub path: PathBuf,
+    pub source: PathBuf,
+    pub destination: FormatDestination,
     pub input: String,
     pub output: String,
 }
 
+#[derive(Debug, PartialEq)]
+pub enum FormatDestination {
```
QuinnWilton, line 16: Which is a more idiomatic name in Rust: `FormatDestination` or just `Destination`?

src/format/command.rs
```diff
@@ -97,13 +118,14 @@ fn format_file(path: PathBuf) -> Result<Formatted, Error> {
...
-        path,
+        source: path.clone(),
+        destination: FormatDestination::File { path: path },
         input: src,
         output: formatted,
     })
 }
 
-pub fn format_stdin(_check: bool) -> Result<(), Error> {
+pub fn read_and_format_stdin() -> Result<Vec<Formatted>, Error> {
```
QuinnWilton, line 128: Perhaps it'd be better to just return `Result<Formatted, Error>` and then wrap it in `vec!` at the callsite?

src/format/command.rs
lpil: No idea :) Do what you prefer!

src/format/command.rs
lpil: I think that's a good idea :)

src/format/command.rs
```diff
@@ -6,21 +6,29 @@ use std::str::FromStr;
...
 }
 
+#[derive(Debug, PartialEq)]
+pub enum FormatDestination {
+    Stdout,
+    File { path: PathBuf },
+}
+
 pub fn run(stdin: bool, check: bool, files: Vec<String>) -> Result<(), Error> {
+    let formatted;
```
lpil, line 22: Can we assign directly to the variable please? I much prefer the style of using expressions.

```rust
let formatted = if stdin { // ....
```

src/format/command.rs
```diff
@@ -6,21 +6,29 @@ use std::str::FromStr;
...
-            check_formatting(formatted)
-        } else {
-            write_formatted(formatted)
-        }
+        formatted = read_and_format_paths(files)?;
+    }
+    if check {
+        check_formatting(formatted)
+    } else {
+        write_formatted(formatted)
```
lpil, line 31: This is very nice! Thanks!

src/format/command.rs
QuinnWilton: I didn't realize that was an option! I much prefer that too

lpil: Honestly I didn't know you could not initialise a variable. I guess Rust can keep track of whether variables are initialised in order to make it safe. I still don't like it 😅

QuinnWilton: Yeah, it bothered me too. I'm not used to C-style languages having nice flow-control constructs, so I just assumed it was the best I could do in Rust :P This language is growing on me a little bit.

QuinnWilton:
Pushed up a squashed diff with the comments addressed!

src/format/command.rs
lpil: Rust is Haskell in C's clothing :)

PR #549, opened 2020-05-05
Imported constructor bugs

tomwhatmore:
This is 3 bug fixes and a refactor in one PR, but they are split up into individual commits so it's possible to do them seperately if preferable. The changes are:

1. The arity of ValueConstructorVariant::Record was being set wrong in `infer_module()`, which meant that on importing one into another module it had the arity of its type instead, casuing incorrect erlang to be printed. Fixed by using the correct arity when creating them.

2. Importing a constructor with an unqualified alias ie `import one.{A as b}` would print the incorrect name when generating the erlang if the constructor had an arity > 0. Fixed by making the name printing code the same as for the arity == 0 case.

3. The arity field of ValueConstructorVariant::Record was only being used in one place in the erland printer, and could be calculated there instead. Refactored it out.

4. ModuleSelect expressions always printed just the atom for a constructor, even if it had args. Fix: Now uses the same logic as when printing a var.

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -277,7 +277,6 @@ pub enum ValueConstructorVariant {
     Record {
         name: String,
         field_map: Option<FieldMap>,
-        arity: usize,
     },
 }
 
@@ -372,7 +371,6 @@ impl<'a, 'b> Env<'a, 'b> {
             ValueConstructorVariant::Record {
                 name: "True".to_string(),
                 field_map: None,
-                arity: 0,
             },
             bool(),
         );
@@ -381,7 +379,6 @@ impl<'a, 'b> Env<'a, 'b> {
             ValueConstructorVariant::Record {
                 name: "False".to_string(),
                 field_map: None,
-                arity: 0,
             },
             bool(),
         );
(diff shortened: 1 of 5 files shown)


src/typ.rs
```diff
@@ -277,16 +277,16 @@ pub enum ValueConstructorVariant {
...
 
 impl ValueConstructorVariant {
     fn to_module_value_constructor(&self) -> ModuleValueConstructor {
         match self {
-            ValueConstructorVariant::Record { name, .. } => {
-                ModuleValueConstructor::Record { name: name.clone() }
-            }
+            ValueConstructorVariant::Record { name, field_map } => ModuleValueConstructor::Record {
+                name: name.clone(),
+                arity: field_map.clone().map_or(0, |fm| fm.arity),
```
lpil, line 288: Could we map without cloning? No need to clone the entire field map if we only the arity from it.

tomwhatmore: Excellent suggestion

tomwhatmore: Done

lpil:
Oh, and would you mind updating the changelog with a note about the bugs squished? Thanks

tomwhatmore:
Can do. What sort of level of detail do you want?

lpil:
Doesn't need too much detail, an example of a previous entry is "Fixed a bug where variable names would be incorrectly generated in some situations when variable names are reused during and after a case expression."

tomwhatmore:
Changelog updated

PR #559, opened 2020-05-09
Surround binary operator sub expressions with parentheses

### LINKED ISSUE #448: Erlang printer does not surround binary operator sub expressions with parens where required

QuinnWilton:
Resolves #448 

I'm unconditionally adding parentheses around the sub expressions here, because it seemed like the simplest way of fixing the wrong behaviour. If you'd prefer being smarter about it I can look into doing that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Unreleased
 
 - Numbers are now permitted in module names.
+- Emitted Erlang code correctly adds parentheses around binary subexpressions
+  to preserve precedence.
 
 ## v0.8.0 - 2020-05-07
 
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -279,7 +279,6 @@ fn seq(first: &TypedExpr, then: &TypedExpr, env: &mut Env) -> Document {
         .append(expr(then, env))
 }
 
-// TODO: Surround left or right in parens if required
 fn bin_op(name: &BinOp, left: &TypedExpr, right: &TypedExpr, env: &mut Env) -> Document {
     let op = match name {
         BinOp::And => "andalso",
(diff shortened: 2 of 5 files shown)


QuinnWilton:
Whoops, this obviously doesn't work because it introduces the opposite problem. I'll properly fix the bug tomorrow + add more comprehensive tests.

QuinnWilton:
Yeah, the `.gitignore` file doesn't ignore `.beam` files that show up in the test directory. I noticed that previous commits had been including the `.beam` files when they changed, so I did so here too.

Do you want me to add a new ignore rule and remove those files?

lpil:
Yes please, they should not be included.

QuinnWilton:
I pushed up the .gitignore changes!

I've also spent all morning trying to break the precedence handling, and I'm having trouble finding any failing test cases. I thought that this approach wouldn't work for tightly binding operators, like `*`, but it seems to as far as I can tell. Is there any precedence climbing or anything going on elsewhere in the system, like during the initial parsing?

I'm hesitant to consider this pull request done, because I don't fully understand why it's working, but if you're able to offer any insight into what might be going on, I'd either be more confident in the changes, or able to put together a better fix.

That being said, I did modify it to only wrap subexpressions in parentheses if they are also binary operators. This cleans the emitted Erlang code up substantially:

```erlang
precedence_test() ->
    should:equal(1 + (2 * 3), 7),
    should:equal((3 * 1) + 2, 5),
    should:equal((1 + 2) * 3, 9),
    should:equal(1 + (2 * 3), 7),
    should:equal(3 * (1 + 2), 9),
    should:equal((1 + (2 * 3)) + 4, 11),
    should:equal((2 * 3) + (4 * 5), 26),
    should:equal((2 * (3 + 1)) div 2, 4),
    should:equal((5 + ((3 div 3) * 2)) - (6 * 4), -17).
```

lpil:
Looks great! Could we have some tests in the `erl::tests` module that show how the Erlang should render with parens where required? It'll be a little friendlier for future development than checking against the core_language test project.

QuinnWilton:
That's a great idea! I'll add that in a few hours

QuinnWilton:
I added those tests to the first commit. Thanks for the suggestion! I was hoping there was a better place for those tests than just in the core_language directory.

QuinnWilton:
Updated it! I'll make sure to do that on future changes.

QuinnWilton:
Alright, should be good now!

Issue #465, opened 2020-04-18
Support literal values in guards


lpil:
```rust
case x {
  _ if x == 1 -> "here we are using an Int literal"
}
```

- [x] Int
- [x] Float
- [x] Tuple
- [ ] List
- [x] Records
- [x] String

If you pick up this ticket please no more than one implementation per PR please so that it is easier to review. Thanks!

kyle-sammons:
Working on this!

lpil:
Thank you @kyle-sammons ! Let me know if you would like any pointer on implementation

gamebox:
Just a suggestion but what about.

```
case x {
  1 -> "this is an int literal pattern"
  _ -> ""
}
```

lpil:
That would be a better way to express that, it's more an example to explain the syntax.

A more realistic example might be something like this:

```rust
case x {
  Ok(i) if i > 3 -> True
  _ -> False
}
```

gamebox:
Could you add some detail about what this looks like for compound types?

lpil:
Does this help?

```rust
case x {
  Ok(i) if i == [] || i == [1] -> True
  _ -> False
}
```

lpil:
Hi @kyle-sammons are you still working on this? Thanks

QuinnWilton:
I'm interested in taking a stab at the tuple implementation if no one is currently looking at that :)

kyle-sammons:
@lpil I am, I am. Hopefully finishing up the tuple implementation today. @QuinnWilton you can totally take either the List or Record one if you want to help out!

QuinnWilton:
Thanks @kyle-sammons! I'll likely start work on one of those tasks in the next few days if you haven't gotten to them by then :)

Do you by any chance have your WIP branch pushed up anywhere? I imagine the implementation for tuples / lists / records will be fairly similar, so it'd probably be worth me trying to match your approach as much as possible. No pressure if it isn't ready for public viewing yet though :D

QuinnWilton:
Just an update -- I have not started work on the list implementation. I've been spending my time thinking about another issue (#34), so list and record support are both unclaimed at the moment.

tomwhatmore:
No-one's working on those last two right? I'll do lists now and unless there's any other takers I'm happy to do records too.

kyle-sammons:
@tomwhatmore Hey-o, I've actually been working on the list implementation. If you want to knock out the record one, though, go for it! :)

[...]

PR #563, opened 2020-05-10
Added support for tuple values in guards

kyle-sammons:
This pull request is meant to complete Issue #465, adding support for tuple literals within guards. As requested each type is going to be it's own separate merge request.

```
let x = tuple(1, 2, 3)
case x {
  (1, 2, 3) -> "here we are using a tuple literal"
}
```


src/erl/tests.rs
```diff
@@ -1817,6 +1817,64 @@ main() ->
     assert_erl!(
         r#"
 pub fn main() {
+  let x = tuple(1, 2, 3)
+  case x {
+    tuple(1, 2, 3) -> 1
```
lpil, line 1822: These tests are using a tuple pattern rather than a tuple in the guard. Could we update these to something like `y if y == tuple(1, 2, 3)` please? Thanks

src/format.rs
```diff
@@ -1022,6 +1022,10 @@ impl Documentable for &UntypedClauseGuard {
 
             ClauseGuard::Float { value, .. } => value.to_string().to_doc(),
 
+            ClauseGuard::Tuple { elems, .. } => "tuple"
```
lpil, line 1025: Could we have a test for formatting of Gleam code that uses tuples in guards please?

lpil:
Looks great! I've left a couple notes above about tests. 

In addition to these could we have tests for the sad path in the typer? For example, Test that these return an appropriate errors:

```rust
case 1 {
  y if y == tuple()
}
```


```rust
case tuple(1, 2) {
  y if y == tuple(1, [])
}
```

src/erl/tests.rs
kyle-sammons: Ah, of course, my bad. Fixed!

src/format.rs
kyle-sammons: Fixed!

kyle-sammons:
@lpil I've added the tests as requested. This suggested test, however, ended in a syntax error:
```
case tuple(1, 2) {
  y if y == tuple(1, [])
}
```

I'm not sure if that's correct or not. Thought I'd double check with you just to make sure it isn't a regression.

lpil:
Oops, my bad. I'm half asleep and am being sloppy 😴 

It should be this:

```
case tuple(1, 2) {
  y if y == tuple(1, []) -> 1
}
```

I forgot the `-> 1` bit

kyle-sammons:
Haha, no worries. That still gives me a syntax error, though. I believe that's because the grammar is defined based on `ClauseGuard`s, and there is no list implementation for `ClauseGuard` yet. Section of the code in question:
```
    <s:@L> "tuple" "(" <elems:Comma<ClauseGuard>> ")" <e:@L> => ClauseGuard::Tuple {
        location: location(s, e),
        typ: (),
        elems,
    },
```

lpil:
Oh yes! Some other value that isn't an Int will work in place of that list

kyle-sammons:
Excellent! In that case, I've already pushed up a test that tests against a float and confirms that it fails to unify the types

lpil:
I've rebased these changes onto master, thank you! <3

Issue #566, opened 2020-05-12
Type annotations on let bindings


lpil:
```rust
let x: String = ""
```

QuinnWilton:
I'm interested in working on this.

To clarify the correct behaviour, should it:
1) Add an optional annotation to untyped let expressions
2) During type inferencing, check whether the inferred type matches the annotation
3) Fail with an error if they do not match

lpil:
Yes, that's right!

PR #571, opened 2020-05-15
Support type annotations on `let` and `assert`

QuinnWilton:
This resolves https://github.com/gleam-lang/gleam/issues/566

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -2213,14 +2213,16 @@ fn infer_let(
 ) -> Result<TypedExpr, Error> {
     let value = infer(value, level + 1, env)?;
     let value_typ = generalise(value.typ(), level + 1);
-    let pattern = PatternTyper::new(env, level).unify(pattern, value_typ)?;
+    let pattern = PatternTyper::new(env, level).unify(pattern, value_typ.clone())?;
     let then = infer(then, level, env)?;
     let typ = then.typ();
 
     // Check that any type annotation is accurate.
     if let Some(ann) = annotation {
-        let ann_typ = env.type_from_ast(ann, &mut hashmap![], NewTypeAction::MakeGeneric)?;
-        unify(ann_typ, value.typ(), env).map_err(|e| convert_unify_error(e, value.location()))?;
+        let ann_typ = env
+            .type_from_ast(ann, &mut hashmap![], NewTypeAction::MakeGeneric)
+            .map(|t| instantiate(t, level, &mut hashmap![], env))?;
+        unify(ann_typ, value_typ, env).map_err(|e| convert_unify_error(e, value.location()))?;
     }
 
     Ok(TypedExpr::Let {
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -1753,8 +1753,18 @@ pub fn infer(expr: UntypedExpr, level: usize, env: &mut Env) -> Result<TypedExpr
             value,
             then,
             assert,
+            annotation,
             ..
-        } => infer_let(pattern, *value, *then, assert, level, location, env),
+        } => infer_let(
+            pattern,
+            *value,
+            *then,
+            assert,
+            &annotation,
+            level,
+            location,
+            env,
+        ),
 
         UntypedExpr::Case {
             location,
@@ -2196,6 +2206,7 @@ fn infer_let(
     value: UntypedExpr,
     then: UntypedExpr,
     assert: bool,
+    annotation: &Option<TypeAst>,
     level: usize,
     location: SrcSpan,
     env: &mut Env,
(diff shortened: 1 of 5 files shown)


src/typ.rs
```diff
@@ -2205,6 +2216,13 @@ fn infer_let(
     let pattern = PatternTyper::new(env, level).unify(pattern, value_typ)?;
     let then = infer(then, level, env)?;
     let typ = then.typ();
+
+    // Check that any type annotation is accurate.
+    if let Some(ann) = annotation {
+        let ann_typ = env.type_from_ast(ann, &mut hashmap![], NewTypeAction::MakeGeneric)?;
```
QuinnWilton, line 2222: Is it okay for me to pass in `hashmap![]` as the vector of type variables? I assumed so, based on my understanding of what this function is doing, but I'm a little fuzzy on some of the details

src/typ.rs
```diff
@@ -2205,6 +2216,13 @@ fn infer_let(
...
+        unify(ann_typ, value.typ(), env).map_err(|e| convert_unify_error(e, value.location()))?;
```
QuinnWilton, line 2223: Similarly, is it correct for me to use `value.typ()` directly, without generalising it first?

QuinnWilton:
> This looks awesome Quinn! Not going to really review, but just wanted to say thanks!

Thanks! I accidentally broke the tests before pushing, so I'm fixing those up now :)

src/typ/tests.rs
```diff
@@ -226,6 +226,7 @@ fn infer_test() {
         "let tuple(tuple(_, _) as x, _) = tuple(tuple(0, 1.0), []) x",
         "tuple(Int, Float)"
     );
+    assert_infer!("let x: String = \"\" x", "String");
```
QuinnWilton, line 229: I'd like to add a test for this snippet, but I'm not exactly sure what the correct behaviour is:

```
let x: tuple(x, x) = tuple(5, 5) x
```

Do you have ideas on whether it should unify or not?

src/typ.rs
```diff
@@ -2205,6 +2216,15 @@ fn infer_let(
     let pattern = PatternTyper::new(env, level).unify(pattern, value_typ)?;
     let then = infer(then, level, env)?;
     let typ = then.typ();
+
+    // Check that any type annotation is accurate.
+    if let Some(ann) = annotation {
+        let ann_typ = env
+            .type_from_ast(ann, &mut hashmap![], NewTypeAction::MakeGeneric)
```
QuinnWilton, line 2223: Is it okay for me to pass in hashmap![] as the vector of type variables? I assume so, based on my understanding of what this function is doing, but I'm a little fuzzy on some of the details

lpil: Oops I accidentally resolved the conversation and I don't know how to undo it on mobile.

It's a hashmap rather than a vector, but you are using it correctly, yes. It's how we keep track of what type variables have what names if we are building multiple types at once (i.e. function arguments)

src/typ/tests.rs
```diff
@@ -226,6 +226,18 @@ fn infer_test() {
...
+    assert_infer!("let x: tuple(x, x) = tuple(5, 5) x", "tuple(Int, Int)",);
+    assert_infer!("let x: tuple(x, y) = tuple(5, 5.0) x", "tuple(Int, Float)",);
+    assert_infer!("let [1, 2, ..x]: List(Int) = [1,2,3] x", "List(Int)",);
+    assert_infer!(
+        "let tuple(5, [..x]): tuple(x, List(x)) = tuple(5, [1,2,3]) x",
+        "List(Int)",
+    );
+    assert_infer!(
+        "let tuple(5.0, [..x]): tuple(y, List(x)) = tuple(5.0, [1,2,3]) x",
```
lpil, line 238: Huh, what Erlang does [..x] generate? I'm not sure that should be valid.

lpil:
Could you rebase on master to bring this up to date? Thanks

src/typ.rs
QuinnWilton: Ah yeah, I meant hashmap. I was mostly wondering whether you could think of any cases where I would need to collect type variables when handling let annotations.

It sounds like the answer is no though, thanks!

src/typ/tests.rs
QuinnWilton: Good question, I'll check once I'm on my computer in an hour or so.

QuinnWilton: It looks like it assigns `X` to the entire list.

This:

```
let [..x] = [1,2,3]
```

Generates:

```
X = [1, 2, 3]
```

It's a little bit weird, but I suppose it kind of makes sense, if `..x` means "the rest of the list that `..x` is contained within"?

QuinnWilton:
Rebased!

src/typ/tests.rs
lpil: I think maybe we should remove this syntax

lpil:
Thank you!!

PR #572, opened 2020-05-16
Add the ability to define additional pages to add to the output of `gleam docs`.

### LINKED ISSUE #569: Ability to add extra pages to the gleam docs output.
It's a good Default to add the Readme as a page, however I would like to be able to specify other pages of markdown that should be included.

Not sure where this declaration of markdown pages should go. Perhaps Gleam.toml?

borski:
This resolves #569

### DIFF
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -4,7 +4,7 @@ use crate::{
     ast::{Statement, TypedStatement},
     error::{Error, GleamExpect},
     format, pretty,
-    project::{self, Analysed, ModuleOrigin, OutputFile, ProjectConfig},
+    project::{self, Analysed, DocsPage, ModuleOrigin, OutputFile, ProjectConfig},
 };
 use askama::Template;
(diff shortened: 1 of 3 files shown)


src/docs.rs
```diff
@@ -19,27 +19,40 @@ pub fn build_project(
     // Read and type check project
     let (config, analysed) = project::read_and_analyse(&project_root)?;
 
-    // Get README content
-    let readme =
-        std::fs::read_to_string(project_root.as_ref().join("README.md")).unwrap_or_default();
+    // Initialize pages with the README
+    let mut pages = vec![DocsPage {
+        title: "README".to_string(),
+        path: "index.html".to_string(),
```
lpil, line 25: Can we make this empty and then have the `index.html` be appended for all pages when rendered?

lpil:
I've rebased these changes onto master (to fix a conflict).

Thank you so much for your help!

Issue #562, opened 2020-05-10
Inline calls to external functions


lpil:
If I define this function:

```rust
pub external fn internal(a) -> a = "some_external" "function"
```

When I use it in my code

```rust
internal("hello")
```

It should generate code calling the internal function directly:

```erl
some_external:function(<<"hello">>).
```

Currently it generates a call to `internal(<<"hello">>).`

tomwhatmore:
I can do this one

lpil:
Thank you!

PR #573, opened 2020-05-16
Inline calls to external functions

tomwhatmore:
Addresses #562


src/erl.rs
```diff
@@ -891,19 +872,6 @@ fn incrementing_args_list(arity: usize) -> String {
...
-    let chars: String = incrementing_args_list(arity);
-
-    atom(name.to_string())
-        .append(format!("({}) ->", chars))
-        .append(line())
-        .append(atom(module.to_string()))
-        .append(":")
-        .append(atom(fun.to_string()))
-        .append(format!("({}).", chars))
-        .nest(INDENT)
```
lpil, line 904: I would like to continue to generate the function so that it could be used from Erlang, Elixir, etc.

src/erl/tests.rs
```diff
@@ -1206,13 +1091,10 @@ x() ->
...
 
--export(['receive'/0, 'catch'/1]).
-
-'receive'() ->
-    'try':'and'().
+-export(['catch'/1]).
 
 'catch'(X) ->
-    'receive'().
+    try:'and'().
```
lpil, line 1097: Is this valid Erlang? `try` is used for try/catch

src/typ.rs
```diff
@@ -1428,9 +1428,9 @@ pub fn infer_module(
                         typ: typ.clone(),
                         origin: location.clone(),
                         variant: ValueConstructorVariant::ModuleFn {
-                            name: name.clone(),
+                            name: fun.clone(),
                             field_map: field_map.clone(),
-                            module: module_name.clone(),
+                            module: vec![module.clone()],
```
lpil, line 1433: Wow that was a lot simpler than I expected!

tomwhatmore:
Sorry, I didn't even realise those core_langauage tests were there! I've just been running cargo test. I'll see why they are failing.

src/erl.rs
tomwhatmore: Ah ok, misunderstood you there. I'll bring this back.

src/typ.rs
tomwhatmore: Haha yeah, me too. I actually implemented it in a much more complicated way first and then went "wait, what if I just..."

src/erl/tests.rs
tomwhatmore: I would assume not. It looks like none of the erlang printed in the `call` function escapes the module names at the moment. Should they all be wrapped in `atom()`?

tomwhatmore:
Had to add `pub` to `external fn to_graphemes` in `unicode_test.gleam` to get the core_language tests to pass as the generated function was unused otherwise.

src/erl/tests.rs
lpil: I'll make an issue

lpil:
Good find with the private function there. I think we want to allow private external functions without warning, so I think we should render the function in the generated Erlang if it is public, but not otherwise.

tomwhatmore:
Yeah that makes sense, will do.

PR #586, opened 2020-05-20
Fix code generation when calling functions returned through field access

### LINKED ISSUE #580: Invalid Erlang generated when calling function returned by accessor
```rust
some_record.field()
```

This generates this Erlang

```erlang
erlang:element(2, SomeRecord)()
```

Which is invalid, it need to be this:

```erlang
(erlang:element(2, SomeRecord))()
```

We will also have the same problem for accessing tuples.

QuinnWilton:
Resolves #580

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@
   from outside their own modules.
 - Additional (arbitrary) markdown documentation pages can now be added and
   built with `docs build`.
+- Fix code generation when calling functions returned through either record
+  or tuple access
 
 ## v0.8.1 - 2020-05-19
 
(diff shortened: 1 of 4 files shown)


lpil:
Would you mind updating the changelog? Thanks

lpil:
oops, merge conflict. Sorry, could you fix that?

QuinnWilton:
Yup, on it :)

QuinnWilton:
Fixed! One day I'll remember to update the changelog on my own. Do you think it's worth adding a PR template that includes that on a checklist?

lpil:
> Do you think it's worth adding a PR template that includes that on a checklist?

If it's possible to add them for only PRs then that sounds good

Issue #579, opened 2020-05-17
Preserve type variable names within functions


lpil:
Currently when type variables are used in the annotation for a function and in annotations within the function they will be seen as different type variables, even if they have the same name.

To implement this we will need to persist the hashmap of type var names to type var types in each function. To do this we may want to refactor the typ module to have a Typer struct holding the state with each function being converted into a method on the typer. This would match how the Formatter works.

This code should type check:

```rust
pub type Box(a) {
  Box(value: a)
};

pub fn go(box1: Box(a)) {
  fn(box2: Box(a)) { box1.value == box2.value }
}
```

lpil:
@QuinnWilton Did you want this ticket? Drop a comment and I'll assign you

QuinnWilton:
Yeah, sure! It'll be the largest work I've done on Gleam, so it might take me a bit longer + I'll likely have more questions as I start on it :)

QuinnWilton:
I've been working on this a little bit, and I'm increasingly of the opinion that we probably want two new "Typer" structs, like you mentioned in my draft PR: `ModuleTyper`, and `FnTyper`, and that `Env` should be renamed to `ModuleTyper`. 

`ModuleTyper` should be responsible for inferring module-level statements, and should instantiate a `FnTyper` for each `Statement::Fn` it performs type inference on. That struct can look approximately like this:

```rust
struct FnTyper<'a, 'b, 'c> {
    typer: &'a mut ModuleTyper<'b, 'c>,
    level: usize,
    type_variables: im::HashMap<usize, Arc<Type>>,
}
```

Should `local_values` and `annotated_generic_type` also be moved onto that struct, from `ModuleTyper`?

I think that `infer` should be moved into `FnTyper`, along with all of its helpers for inferring expressions. Whenever `infer_fn` runs, it can create a new `FnTyper` for handling that function. Doing this means that we don't need to cleanup state after that function is inferred, and may allow us to remove this sort of existing cleanup: https://github.com/gleam-lang/gleam/blob/master/src/typ.rs#L3288

I then just need to modify function and let annotations to add their type variables to the state, and modify `instantiate` to use that state, and I think things will work as expected.

Does that align with what you were thinking needs to happen?

lpil:
That sounds like a good improvement, especially the removal of the cleanup code. Thanks for this thorough write up!

QuinnWilton:
[...]

PR #587, opened 2020-05-20
Define crate::typ::Typer and move inference methods into it

QuinnWilton:
This is preliminary work for allowing the typer to track it's own
state, so that it's able to keep track of type variables, unused
types, etc, as described in #579 

This commit has no functional changes, and simply defines the
Typer struct, and moves the type inference functions into it
as methods.

I wanted to make sure I understood what you were suggesting
in #579 before getting too far into the refactor. Is this
approximately what you were hoping for?

I am interested in working on the larger issue, too, however if
you have some time at some point, I wouldn't mind chatting
with you to make sure I understand what your ideal solution
looks like.

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -795,7 +795,7 @@ impl<'a, 'b> Typer<'a, 'b> {
 
             UntypedExpr::Int {
                 location, value, ..
-            } => infer_int(value, location),
+            } => self.infer_int(value, location),
 
             UntypedExpr::Seq { first, then, .. } => infer_seq(*first, *then, level, self),
 
@@ -1033,6 +1033,14 @@ impl<'a, 'b> Typer<'a, 'b> {
             location,
             value,
             typ: string(),
+        })
+    }
+
+    fn infer_int(&mut self, value: String, location: SrcSpan) -> Result<TypedExpr, Error> {
+        Ok(TypedExpr::Int {
+            location,
+            value,
+            typ: int(),
         })
     }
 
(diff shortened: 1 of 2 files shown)


lpil:
I was thinking renaming Env -> Typer and then having all the functions that take it as a last argument to instead be a method on it.

Here's the first step of a similar refactoring which happened in the formatter module when it started to need additional state: https://github.com/gleam-lang/gleam/commit/d35c51843362cf7bd4e173bc9a3afa86a0e28123#diff-0bb16ac13b75e9b568fa4aff61b0e71dR23-R30

Now all functions that require some kind of stateful environment are methods on the Formatter struct https://github.com/gleam-lang/gleam/blob/master/src/format.rs#L33

One question I had in my head was this: Is there a need for Env -> ModuleTyper and there be a FnTyper which holds module function state such as the hashmap of type var names and the level?

QuinnWilton:
> I was thinking renaming Env -> Typer and then having all the functions that take it as a last argument to instead be a method on it.

Ah, that makes sense! I was wondering what the relationship was between `Env` and `Typer`. Yeah, I can do that.

> One question I had in my head was this: Is there a need for Env -> ModuleTyper and there be a FnTyper which holds module function state such as the hashmap of type var names and the level?

I'm not sure yet. If it's possible, it'd be nice to be able to just use another `Env` struct for sub-environments, possibly with a constructor that clones the current environment. That's also something that would be useful for exhaustiveness checking.

I'll need to spend more time in the code to understand how everything ties together though. I'll also likely have a ton of questions for you :)

PR #592, opened 2020-05-21
Add support for custom type record literals in guards

tomwhatmore:
Addresses #465 

This seems to work and passes the tests, but the typing of the `ClauseGuard::Constructor` is a massive hack.

All the inference functions that seem relevant expect `UntypedExpr` rather than `UntypedClauseGuard` so I can't call them directly. I ended up using `infer_value_constructor()` to make construct a `TypedExpr::Var` and then using that with code copied from `do_infer_call_with_known_fun()` but with various bits swapped out to use `ClauseGuard`.

The formatter and printer also end up being very similar to existing logic but with `ClauseGuard` instead of `Expr`.

Suggestions welcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Add lookup for Gleam source code in Mix's `deps` directory.
 - Newly generated Gleam projects use the GitHub action
   `gleam-lang/setup-erlang` v1.1.0.
+- Added support for custom type record literals in guards.
 
 ## v0.8.1 - 2020-05-19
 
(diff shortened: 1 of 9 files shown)


lpil:
It is a bit irksome to duplicate this logic, but I don't have a better alternative. It's relatively easy to understand and it's not the largest chunk of code so I would be happy with this approach if you are.

tomwhatmore:
Ok, fair enough. I'll add a few comments explaining the reasoning behind the duplication and possibility for future improvement and then submit this PR properly.

tomwhatmore:
Rebased onto master

PR #597, opened 2020-05-23
Updated compiler deps to latest

oneness:
@lpil As you have mentioned in the issue thread, I updated all except for `codespan`.  All tests are passing.  BTW, `codespan` did have breaking changes and I assume you know what it is. 

I created a bash script to automate updating the deps thinking we all need this ongoing basis. It is just first cut and have some rough edges. Please have a look and let me know. 

Great work, keep it up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Add lookup for Gleam source code in Mix's `deps` directory.
 - Newly generated Gleam projects use the GitHub action
   `gleam-lang/setup-erlang` v1.1.0.
+- Added support for custom type record literals in guards.
 
 ## v0.8.1 - 2020-05-19
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -23,7 +23,7 @@ dependencies = [
 
 [[package]]
 name = "arrayref"
-version = "0.3.5"
+version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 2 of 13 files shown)


lpil:
Thanks for the help here.

The changes to the Cargo.toml file has resulted in the versions being much more restrictive, meaning that now we would need to edit this file to upgrade the deps. I don't think the Cargo.toml file needs to be edited for this ticket to be done. More information on the cargo deps format can be found here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

oneness:
> Thanks for the help here.
> 
> The changes to the Cargo.toml file has resulted in the versions being much more restrictive, meaning that now we would need to edit this file to upgrade the deps. I think we might not need to any this file much, but if you would like to bump the versions there I would like to keep the same restrictions as was previously present, i.e. `package = "0.6"`
> 
> More information on the cargo deps format can be found here: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

Thanks for the context. I updated the script to just put major.minor as you stated. You can just runt `./bin/update-dependencies.sh` from now on to keep compiler deps up to date.

lpil:
Thanks you but we won't be using the shell script, we will continue to use cargo's build in upgrading features.

oneness:
> Thanks you but we won't be using the shell script, we will continue to use cargo's build in upgrading features.

Sounds good. Please feel free to merge or close this PR.

lpil:
There seems to be some extra commits in this branch now. Would you mind removing them and the Cargo.toml restrictions changes so we can merge this? Thanks

oneness:
> There seems to be some extra commits in this branch now. Would you mind removing them and the Cargo.toml restrictions changes so we can merge this? Thanks

Sure. Let me do a clean fork of current master and do a clean PR.

PR #607, opened 2020-05-26
Preserve type variables across scopes

QuinnWilton:
Resolves https://github.com/gleam-lang/gleam/issues/579

There's a lot of miscellaneous refactoring going on here to make the final change easier. I'm not completely happy with the end result yet, but it does work, so I wanted to PR what I have so that we can start testing 0.9 sooner than later.

In particular, I wish I had a better method for opening / closing new scopes. I don't like that the programmer needs to remember to call `end_scope` in order to keep `warnings` and `uids` in sync. Tests should fail immediately if they do forget, though, so it should be ok for now. For the future, I think if `new_scope` took a function as an argument it'd be able to automatically close the scope after calling it. That was a larger refactoring than I wanted to address now, however.

### DIFF
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -325,6 +325,9 @@ pub struct Typer<'a, 'b> {
     // Values defined in the current function (or the prelude)
     local_values: im::HashMap<String, ValueConstructor>,
 
+    // Type variables in the current scope taken from annotations
+    annotated_type_vars: im::HashMap<String, (usize, Arc<Type>)>,
+
     // Types defined in the current module (or the prelude)
     module_types: HashMap<String, TypeConstructor>,
 
@@ -371,6 +374,7 @@ impl<'a, 'b> Typer<'a, 'b> {
             accessors: HashMap::new(),
             warnings: Vec::new(),
             local_values: hashmap![],
+            annotated_type_vars: hashmap![],
             importable_modules,
             current_module,
         };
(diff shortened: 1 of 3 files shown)


src/typ.rs
```diff
@@ -2863,6 +2853,7 @@ pub fn infer_module(
 
                 // Infer the type
                 let mut fn_typer = typer.new_scope();
+                fn_typer.type_vars.clear();
```
QuinnWilton, line 2856: I needed to clear the `type_vars` here to get tests passing, but I'm not sure entirely sure why yet. I think type variables from custom type constructors may be leaking into the top level scope somewhere, but I haven't managed to confirm the issue + find a less hacky fix.

I'm going to keep looking, but I don't think this introduces any new bugs, because the original code used an empty `type_vars` hashmap here too. My main concern is whether there's a larger systemic problem for all module-level statements, but I haven't constructed any failing testcases yet.

src/typ.rs
```diff
@@ -314,16 +314,20 @@ pub enum PatternConstructor {
...
+    level: usize,
     annotated_generic_types: im::HashSet<usize>,
     importable_modules: &'a HashMap<String, Module>,
     imported_modules: HashMap<String, Module>,
 
     // Values defined in the current function (or the prelude)
     local_values: im::HashMap<String, ValueConstructor>,
 
+    // Type variables in the current scope
+    type_vars: im::HashMap<String, (usize, Arc<Type>)>,
```
lpil, line 329: There will be other type vars in the scope that are not in this hashmap- this is just the type vars from annotations. Perhaps `annotation_type_vars` would be a better name?

src/typ.rs
lpil: Makes sense, thank you. This seems like the same behaviour to me.

tomwhatmore:
Nice one Quinn, this looks great!

I have a fairly easy suggestion for the opening / closing new scopes issue you mentioned, I'll make a PR.

QuinnWilton:
I'd love to see that! I spent a while trying different things, but my Rust-fu was lacking.

tomwhatmore:
You probably didn't have enough angle brackets, that's usually the problem.

PR #608, opened 2020-05-26
Make typing scopes an actual rust scope

tomwhatmore:
Addresses the scoping issue @QuinnWilton brought up as part of #607.

It's a little verbose as you have to explicitly pass values back out of the scope and in cases where  there's more than one you have to construct a tuple to do so, but it does make it impossible to forget to close.


tomwhatmore:
That's a good idea, should definitely be doable. I'll take a look tomorrow.

PR #617, opened 2020-05-29
Compile Erlang to beam

lpil:


test/build_with_gleam/gleam.toml
```diff
@@ -0,0 +1,2 @@
+name = "bob" # The builder
```
QuinnWilton, line 1: 👷

unsek:
Does this mean that Gleam can be compiled directly to EVM machine code rather than stopping at Erlang?

lpil:
Gleam will continue to compile to Erlang (as Elixir does), but we will no longer depend upon rebar3 to compile the Erlang code.

src/build.rs
```diff
@@ -0,0 +1,91 @@
...
+    file::write_outputs(compiled_erlang.as_slice())?;
+
+    // Compile Erlang source into VM bytecode
+    compile_erlang_to_beam(&root, &packages)?;
+
+    Ok(())
+}
+
+#[derive(Debug)]
+pub struct Package {
```
tsloughter, line 56: I'd name this `Application`, if I'm right in how you are using it. Package will be confusing because while a hex package will correlate to a single Application you can have Applications not from Packages.

I started with:

```rust
pub struct Application {
    pub name: String,
    pub vsn: String,
    pub description: String,
    pub id: String,
    pub included_applications: Vec<String>,
    pub applications: Vec<String>,
    pub start_mod: (String, Vec<String>), // is actually `mod` in the .app file
    pub start_phases: Vec<String>,
    pub registered: Vec<String>,
    pub max_t: i64, // max time an app can run for in millseconds, using 0 for infinity
    pub env: Vec<String>,

    // for now this is to be filled in by rebar3
    pub modules: Vec<String>,

}
```

And then was going to write it out with.

```
    let app_name = config.name.to_string();
    let _terms: [(String, String); 0] = [];

    let f = File::create(format!("{}.app.src", app_name)).expect("Unable to open file");
    let mut w = BufWriter::new(f);

    let app_spec = BertTerm::List(vec!());
    let app_resource = BertTerm::Tuple(vec!(BertTerm::Atom("application".to_string()),
                                            BertTerm::Atom(app_name),
                                            app_spec));


    let _ = app_resource.write_as_erlang(&mut w, 2, 5);
    let _ = w.write(b".\n");
```

lpil: The name application is one of the most confusing things in Erlang so I'd like to avoid it in the context of Gleam.

As for the fields, I'm not sure yet what is needed, this is a rough first pass. Thanks for sharing that snippet, it looks very informative

tsloughter: I highly recommend not fighting OTP naming. The naming is unfortunate but is only made worse if there are competing names. It isn't only in the beginning that users will have to work with Erlang directly, even at a point like Elixir's the users are faced with Erlang documentation and dependencies.

If you only meant in the context of the gleam internals like this code I think its still less confusion to not try to map another name to the same concepts.

lpil: It has been a challenge with everyone I've taught Elixir and Erlang to, and apart from you every experienced BEAM user I've discussed this with has agreed that different language would be beneficial. Nothing is set in stone yet but so far the evidence supports not using the overloaded application name.

tsloughter: It would be beneficial if it were able to replace the existing terms. But when those users are still going to have to learn what an OTP Application is it will only make it more confusing that they have to now also translate other terms to the OTP terms.

I don't know that Elixir embraced it early enough, that is definitely the case with OTP Releases. I think there will be much less trouble down the line if from the beginning you spell out to users what Applications, Releases, gen_servers, etc are and not have them first encounter them when they ultimately have to dig into Erlang documentation or OTP logs that will still be mentioning applications.

ferd: Echoing Tristan's warning. We gave the same warning to Elixir developers half a decade ago, only to see most of the decisions that aimed to work _around_ releases have to be undone or shimmied in weird ways to work with how the VM ultimately runs things and is deployed.

If you do plan to change the terminology, you should at least have an early plan and vision on how your terms are going to be bridged with the ERTS concepts because you'll have a much easier time shipping the VM with the OTP libs and apps as a more or less static bundle that boots and shuts down the proper way, than you'd have reinventing the loader, config handling, bundling to tarballs, vm arguments, and portability of the runtime system. 

You'll have to eventually bridge the gap, and if you pick incompatible ways to do it, it won't be easy. It will also either impose limitation on the composability of your stack (elixir + gleam code for example) possibly being hard to make work together, or will still require people to learn both terms at once. If your objective is more independent, then at least that latter part can be ignored for a while.

lpil: There's no plans to change anything in Erlang except for the name "application" which confuses newcomers. Releases will be supported in as easy as possible fashion as soon as possible. :)

tsloughter: Releases tell ERTS how to boot applications, so my warning here is that once a user starts digging into releases they'll encounter "application" and possibly be even more confused than if they were confronted with applications from the start.

PR #620, opened 2020-06-01
Binary syntax

tomwhatmore:
Not sure this is completely finished but it's close enough for some feedback to be useful.

The TL;DR is: very close to Elixir syntax, with slightly more annotations required to keep the type checking happy. When used in a pattern the default type for each segment is integer, in construction the default type is binary. All other types require explicit annotations.

### Compile time errors

1. Mismatched values and types (including default types) in binary segments: `<<2.0>>`, `<<"test">>`, `<<1::binary>>`
2. Bound variables will be unified and produce an error if they don't match `{ <<a::binary>> if a > 1 -> 1 }`
3. Duplicate type options `<<1::integer-binary>>`
4. Invalid type options `<<1::typo>>`
5. Providing units for types that do not accept them <<"asd"::utf32-unit(2)>>

Anything else useful we could check for?


src/binary.rs
```diff
@@ -0,0 +1,198 @@
...
+    {
+        let empty_bts = BinaryTypeSpecifier {
+            typ: None,
+            signedness: None,
+            endianness: None,
+            unit: None,
+            size: None,
+        };
+
+        let parsed_bts = options.iter().try_fold(empty_bts, |mut bts, option| {
```
lpil, line 27: Does this mean that the options can be given in any order?

src/format.rs
```diff
@@ -640,6 +651,71 @@ impl<'a> Formatter<'a> {
         }
     }
 
+    fn segment(&mut self, segment: &UntypedExprBinSegment) -> Document {
+        match segment {
+            UntypedExprBinSegment { value, options, .. } if *options == [] => self.expr(&value),
+
+            UntypedExprBinSegment { value, options, .. } => {
+                // TODO: Is there a better option than these self.clone() calls?
+                let options = options.iter().map(|o| self.clone().segment_option(o));
```
lpil, line 660: Why is it that the formatter is cloned here and on line 662?

src/format.rs
```diff
@@ -974,6 +1061,77 @@ impl<'a> Formatter<'a> {
...
+    fn pattern_segment(&mut self, segment: &UntypedPatternBinSegment) -> Document {
+        match segment {
+            UntypedPatternBinSegment { value, options, .. } if *options == [] => {
+                self.pattern(&value)
+            }
+
+            UntypedPatternBinSegment { value, options, .. } => {
+                // TODO: Is there a better option than these self.clone() calls?
+                let options = options
+                    .iter()
```
lpil, line 1075: Same as above, why is the formatter cloned here?

src/format/tests.rs
```diff
@@ -2571,6 +2606,8 @@ pub fn two() {
 }
 ",
     );
+
+    // Bitstrings
```
lpil, line 2610: Leftover comment ? :)

src/grammar.lalrpop
```diff
@@ -579,6 +583,76 @@ Todo: UntypedExpr = {
     }
 }
 
+Bitstring: UntypedExpr = {
```
lpil, line 586: Is there a difference between a bitstring and a binary? Would it be better to always call it a binary?

src/typ.rs
```diff
@@ -3901,3 +4228,39 @@ pub fn tuple(elems: Vec<Arc<Type>>) -> Arc<Type> {
...
+pub fn binary() -> Arc<Type> {
+    Arc::new(Type::App {
+        args: vec![],
+        public: true,
+        name: "Binary".to_string(),
+        module: vec![],
+    })
+}
+
+pub fn bitstring() -> Arc<Type> {
```
lpil, line 4241: What are these other types for? Would they not all be `Binary`?

src/typ.rs
```diff
@@ -1281,6 +1286,111 @@ impl<'a, 'b> Typer<'a, 'b> {
...
+        location: SrcSpan,
+    ) -> Result<TypedExprBinSegment, Error> {
+        let value = self.infer(value)?;
+
+        let options = options
+            .into_iter()
+            .map(|option| self.infer_segment_option(option))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let type_specifier = BinaryTypeSpecifier::new(options.clone(), true)
```
lpil, line 1319: Could the BinaryTypeSpecifier hold a slice rather than cloning the vector? From the implementation I think it could borrow the options though I'm not 100% sure.

src/grammar.lalrpop
```diff
@@ -579,6 +583,76 @@ Todo: UntypedExpr = {
...
+    <s:@L> "<<" <elems:Comma<BinSegment>> ">>" <e:@L> => {
+        UntypedExpr::Bin {
+            location: location(s, e),
+            elems: elems,
+        }
+    }
+}
+
+BinSegment: UntypedExprBinSegment = {
+    <s:@L> <v:SimpleExpr> <options:("::" <HyphensNoTrailing<BinSegmentOption>>)?>  <e:@L> =>
```
lpil, line 596: Could we use `:` rather than `::`? To match type annotations for values.

src/typ.rs
```diff
@@ -3901,3 +4228,39 @@ pub fn tuple(elems: Vec<Arc<Type>>) -> Arc<Type> {
 pub fn fn_(args: Vec<Arc<Type>>, retrn: Arc<Type>) -> Arc<Type> {
     Arc::new(Type::Fn { retrn, args })
 }
+
+pub fn binary() -> Arc<Type> {
```
lpil, line 4232: We'll also want to insert the type constructor into the default type environment so that it can be referenced in user defined types: https://github.com/gleam-lang/gleam/blob/3e96b7714e862c36f93c38243ee748e6337d70a2/src/typ.rs#L443-L454

src/binary.rs
tomwhatmore: It does, which I think is consistent with both erlang and elixir.

src/format.rs
tomwhatmore: Good question. I don't remember to be honest, it must have refused to compile for reasons I couldn't be bothered to properly figure out at the time, hence the quick fix and the TODOs. It will be sorted before this becomes a real PR :)

src/format.rs
tomwhatmore: Same

src/format.rs
lpil: Thanks!

src/format/tests.rs
tomwhatmore: I need to move the new asserts down here, I've got them in a separate test at the moment because it makes them easier to run on their own. At some point I'd like to try and split these mega-tests up into smaller, better documented ones, but that's a task for another day.

src/binary.rs
lpil: TIL! Thank you

src/format/tests.rs
lpil: Makes sense. I'd be happy to have them as a different test function if you'd like to split the others out too at a later date.

src/grammar.lalrpop
tomwhatmore: I renamed the types about 3 times and it's not completely cleaned up, thought I would see what you thought before doing a final pass.

The Erlang docs calls `<<1>>` a Bin which consists of a bunch of Segments. Elixir seems to go for bitstring and segment. They agree on what the names of the types in the segments are (binary and bitstring). I think bitstring is probably better because it properly encapsulates nesting, e.g.`<<1::1, <<1::1>>::bitstring>>` is in turn a bitstring. Arguments against, would probably have to rename the segments to be consistent, and BitstringSegmentOption<TypedExpr> is very long.

A binary is a specific subset of a bitstring, namely one where the number of bits is divisible by 8, so it feels confusing to use it to refer to the `<<x>>` syntax as a whole (I think?)

src/typ.rs
tomwhatmore: I'm fairly sure this could just be a borrow.

src/grammar.lalrpop
tomwhatmore: Certainly. I actually went for `::` specifically to not match normal types, haha. I can see the arguments both ways though, and would be happy with either.

lpil: Perhaps we should ask the wider group?

src/typ.rs
tomwhatmore: Indeed it could

src/format/tests.rs
tomwhatmore: I'll move the format ones back into `module_test()` because there's only 2 asserts anyway. The typ ones I'll leave in their own test as there's quite a few.

src/typ.rs
tomwhatmore: Good point, will do.

src/typ.rs
tomwhatmore: They could be, though I think I would probably make them all a bitstring rather than a binary if we went that way. I'm not sure if they have much use as types outside the bitstring itself. It would certainly be simpler to just have the one.

lpil: After your explanation of the difference I think bitstring makes more sense than binary as the general type. We could add more specialised types later if we think it'll be useful, for now having a smaller API it's my preference.

tomwhatmore: Having typed my explanations of the difference I came to the same conclusion!

src/grammar.lalrpop
tomwhatmore: Can do

tomwhatmore:
Weird cached grammar.rs issue, fixing.

tomwhatmore:
I have addressed all those comments except `..` vs `.`, which can be trivially resolved once we get a consensus on what people would prefer.

Using `Bitstring` consistently seems like the simplest approach for now, both in terms on naming things consistently and the type system.

tomwhatmore:
I plan to add 2 more commits then I'm happy with it if you are.

1. Update the syntax to `:`
2. Add a couple more tests around infering nested bitstrings

I think that supporting Binary (and maybe the UTFs) as separate types would probably be good but I'm not sure how best to represent them right now. Can always be added later, so no reason to hold up the rest of it.

lpil:
Sounds great, let me know when you need me :)

tomwhatmore:
Think this is good to go now @lpil

CrowdHailer:
I had more thoughts on clever types for bitstring/binary/strings.
I think there is possibly a way to do it but actually the improvement in ergonomics wouldn't be that high.

Certainly writing the following is super easy to follow
```rust
<<a:8, b:16>> = bitstring.from_binary(binary)
<<a:8, b:16>> = bitstring.from_string(string)
```

tomwhatmore:
What were your thoughts? Worth having a note of them even if we don't address the issue right now. I agree that those examples are fine, though I think the other way round is the more annoying one:

```rust
let <<a:binary>> = ...
try b = binary.from_bitstring(a)
```

Would be much nicer to just have a `Binary` from the initial match and not have to deal with a `Result` etc. Possibly one to revisit when we have traits?

CrowdHailer:
I'd agree that is a bit more annoying. 

My thoughts where that I could deal with that annoyance to have the value of bitstring syntax, sooner. Also that the migration from distinct types to something more clever would be backwards compatible.
, It would only add functionality not remove it.

It was more a vote for delivering it separately rather than not to do it.
I'm waiting for a stdlib 0.9.x with binary type so I can put midas on to hex dependencies and do a 0.3.0 release.  I had a look at where I would use binary syntax it would be useful in a few places but conversions like above wouldn't be a deal breaker at all. 

Also there is assert

Issue #615, opened 2020-05-28
Fold definition incorrectly generated in Hex docs?


scripttease:
The Hex docs for list.fold state the following:
```
pub fn fold(
  over list: List(a),
  from initial: b,
  with fun: fn(a, b) -> b,
) -> a
```

Should it be this?
```
pub fn fold(
  over list: List(a),
  from initial: b,
  with fun: fn(a, b) -> b,
) -> b
```

Same for fold-right.

lpil:
Oh dear, that is strange. Looks like a bug in the printer.

tomwhatmore:
Is this fixed? Seems to be generating ok for me on master.

lpil:
I don't remember changing anything. How strange.

I'll do a quick test now.

lpil:
I'm testing it with v0.9.0 of the compiler and the stdlib and it is giving me the incorrect output that @scripttease has above. What versions are you using @tomwhatmore ?

tomwhatmore:
Current master. I just copied the fold function into the hello_world project to give it a quick test, I didn't actually generate the full stdlib docs, so maybe that's the difference? Happy to look into it properly if you want to assign this to me.

lpil:
Thank you Tom!

tomwhatmore:
Actually I think you can just ignore that completely. I just did it again and I'm getting the broken one. Maybe I just misread it, sorry! I will look into it though.

PR #623, opened 2020-06-02
Track type variables in function signatures properly when printing docs

tomwhatmore:
As reported in #615. Stdlib docs will need to be regenerated to close the issue properly.

We should probably have some tests for docs generation!

### DIFF
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -1,4 +1,6 @@
 pub(crate) mod command;
+#[cfg(test)]
+mod tests;
 
 use crate::{
     ast::{Statement, TypedStatement},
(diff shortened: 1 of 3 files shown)


lpil:
We really should! Thanks for this.

Do you think we should add a test for this function in this PR?

tomwhatmore:
Sure, I'll make a new assert macro for docs and add a test for this behaviour.

tomwhatmore:
I've added a new docs test which is currently testing for comments and few different types of function, one of which is the behaviour fixed in the initial commit. It could have quite a few more assertions added to it but I think that's out of scope for this PR.

The plumbing is all in place now though, so it would be pretty easy to add more. Could be some good first issues in there maybe? Or I can just do them if there's no takers.

tomwhatmore:
That's really weird. The code is definitely using those imports, and won't compile without them. I'll see if I can figure it out.

tomwhatmore:
Oh I forgot the `#[cfg(test)]`

lpil:
Classic move 😁

Issue #610, opened 2020-05-26
Formatter: Wrap unqualified imports


lpil:
Have the formatter render these expressions like so:

```haskell
import my/cool/module.{
  One, Two, Three, Four, Five, Six, Seven, Eight, Nine, Ten, Eleven, Twelve,
  Thirteen, Fifteen, Sixteen,
}
```

This will require extending the printing algebra with a form like Elixir's flex_break

https://github.com/elixir-lang/elixir/blob/d1559660740c85d94e3cff942ff0b6c8f7d31818/lib/elixir/lib/inspect/algebra.ex#L680-L712

oneness:
@lpil I can take a stab at this. Any pointer where to focus?

lpil:
For sure. First we'll need to work in the pretty module: https://github.com/gleam-lang/gleam/blob/master/src/pretty.rs

The Document enum will need to be expanded to support this new extension which can be found in the Elixir implemention above

https://github.com/gleam-lang/gleam/blob/3cecc58691af74a1b9e1bdc7c9bd42020a7a9052/src/pretty.rs#L83-L111

Then we will want to update the formatter to use this new form when rendering the unqualified imports: https://github.com/gleam-lang/gleam/blob/3cecc58691af74a1b9e1bdc7c9bd42020a7a9052/src/format.rs#L209-L216

oneness:
Thanks for the context, which should be enough to get me digging into.  So feel free to assign it to me so others know to avoid duplicate efforts.

lpil:
Wonderful, thank you!

PR #633, opened 2020-06-05
Wrap unqualified imports

oneness:
Addresses #610 

Introduces flex break to allow vector of Document to break or unbreak
based on the set limit (currently 80).  

At its core, this is just about adding line breaks where ast of
UnqualifiedImports are more than the current limit. I think I followed
the intended design of decoupling parser specific types from
formatting where the main abstraction is the Document enum
type (Please correct me if I am wrong). This is my first Rust program
and could be done better if we only had to traverse the ast just
once (had to iter on it by cloning). I'd appreciate if you provide any
feedback on how to improve it.

Other than that, I think it is pretty generic enough to be
reused. Again I could be totally wrong and any feedback are welcome.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -207,13 +207,23 @@ impl<'a> Formatter<'a> {
                     nil()
                 } else {
                     ".{".to_doc()
-                        .append(flex_break(
-                            unqualified
-                                .iter()
-                                .map(|e| e.clone().to_doc())
-                                .intersperse(", ".to_doc())
-                                .collect(),
-                        ))
+                        .append(
+                            concat(
+                                unqualified
+                                    .iter()
+                                    .map(|e| {
+                                        break_(&e.clone().name, &e.clone().name).append(
+                                            if let Some(name) = &e.as_name {
+                                                format!(" as {}", name).to_doc()
+                                            } else {
+                                                nil()
+                                            },
+                                        )
+                                    })
+                                    .intersperse(delim(",")),
+                            )
+                            .flex_group(),
+                        )
                         .append("}")
                 })
                 .append(if let Some(name) = as_name {
--- a/src/format.rs
+++ b/src/format.rs
@@ -212,13 +212,8 @@ impl<'a> Formatter<'a> {
                                 unqualified
                                     .iter()
                                     .map(|e| {
-                                        break_(&e.clone().name, &e.clone().name).append(
-                                            if let Some(name) = &e.as_name {
-                                                format!(" as {}", name).to_doc()
-                                            } else {
-                                                nil()
-                                            },
-                                        )
+                                        break_(&e.clone().name, &e.clone().name)
+                                            .append(import_as(&e.clone().as_name))
                                     })
                                     .intersperse(delim(",")),
                             )
@@ -226,11 +221,7 @@ impl<'a> Formatter<'a> {
                         )
                         .append("}")
                 })
-                .append(if let Some(name) = as_name {
-                    format!(" as {}", name).to_doc()
-                } else {
-                    nil()
-                }),
+                .append(import_as(&as_name)),
         }
     }
 
@@ -964,12 +955,16 @@ fn pub_(public: bool) -> Document {
     }
 }
 
+fn import_as(as_: &Option<String>) -> Document {
+    match as_ {
+        None => nil(),
+        Some(s) => " as ".to_doc().append(s.clone()),
+    }
+}
+
 impl Documentable for &UnqualifiedImport {
     fn to_doc(self) -> Document {
-        self.name.clone().to_doc().append(match &self.as_name {
-            None => nil(),
-            Some(s) => " as ".to_doc().append(s.clone()),
-        })
+        self.name.clone().to_doc().append(import_as(&self.as_name))
     }
 }
 
--- a/src/format.rs
+++ b/src/format.rs
@@ -207,11 +207,12 @@ impl<'a> Formatter<'a> {
                     nil()
                 } else {
                     ".{".to_doc()
-                        .append(concat(
+                        .append(flex_break(
                             unqualified
                                 .iter()
                                 .map(|e| e.clone().to_doc())
-                                .intersperse(", ".to_doc()),
+                                .intersperse(", ".to_doc())
+                                .collect(),
                         ))
                         .append("}")
                 })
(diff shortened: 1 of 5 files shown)


src/format/tests.rs
```diff
@@ -25,6 +25,8 @@ fn module_test() {
     assert_format!("import one/two/three as free\n");
     assert_format!("import one/two/three.{thunk} as free\n");
     assert_format!("import one/two/three.{thunk as funky} as free\n");
+    assert_format!("import my/cool/module.{\nOne, Two, Three, Four, Five, Six, Seven, Eight, Nine, Ten, Eleven, Twelve, \nThirteen, Fifteen, Sixteen\n}\n");
```
lpil, line 28: Could we have some indentation after the break? Currently it is formatted like this:

```rust
import my/cool/module.{
One, Two, Three, Four, Five, Six, Seven, Eight, Nine, Ten, Eleven, Twelve, 
Thirteen, Fifteen, Sixteen
}
```

When I think it should be formatted like this:

```rust
import my/cool/module.{
  One, Two, Three, Four, Five, Six, Seven, Eight, Nine, Ten, Eleven, Twelve, 
  Thirteen, Fifteen, Sixteen,
}
```

There is also some trailing whitespace on the second line.

oneness:
Your feedback other than the trailing whitespace bug is addressed and have a look to let me know if that is the right direction taken.  Thanks.

src/format.rs
```diff
@@ -207,13 +207,23 @@ impl<'a> Formatter<'a> {
...
-                        ))
+                        .append(
+                            concat(
+                                unqualified
+                                    .iter()
+                                    .map(|e| {
+                                        break_(&e.clone().name, &e.clone().name).append(
+                                            if let Some(name) = &e.as_name {
+                                                format!(" as {}", name).to_doc()
+                                            } else {
```
lpil, line 218: I think this will render a break between the name of the field being imported and the `as` as `break_` can insert a newline after it.  Could we not have the previous logic and intersperse a `delim(",")`?

src/format/tests.rs
oneness: The fix is checked in.

src/format.rs
oneness: I refactored the code a bit to make it more readable.  Import as is not a `Break Document` so new line will not be inserted.  Please let me know what you think.  Will address any concern/feedback you have.  Thanks.

oneness:
> Looking good!
> 
> I've found an example that get formatted incorrectly (it changes the name of the imported type)
> 
> ```rust
> import gleam/result.{
>   Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa,
>   Abcde, End
> }
> ```
 
Thanks. Nice example. It caught couple of bugs:
1.  Should remove last trailing space char not just any chars. 
2. The fit logic should have checked for `broken` length since it is what is being padded. 

So your example will be formatted like this as it should be: 
```
import gleam/result.{
  Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, Abcde,
  End
}
```
Since `Aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa, Abcde,` is 78 chars long, which fits within the 80 chars limit.  

I fixed both bug and added above to the test cases in the latest commit.

src/format.rs
```diff
@@ -207,19 +207,21 @@ impl<'a> Formatter<'a> {
...
-                                .map(|e| e.clone().to_doc())
-                                .intersperse(", ".to_doc()),
-                        ))
+                        .append(
+                            concat(
+                                unqualified
+                                    .iter()
+                                    .map(|e| {
+                                        break_(&e.clone().name, &e.clone().name)
+                                            .append(import_as(&e.clone().as_name))
```
lpil, line 216: We can reduce the amount of memory allocation like so:
```suggestion
                                        e.name
                                            .to_doc()
                                            .append(break_("", ""))
                                            .append(import_as(&e.as_name))
```

src/pretty.rs
```diff
@@ -209,6 +248,18 @@ fn fmt(b: &mut String, limit: isize, mut width: isize, mut docs: Vector<(isize,
                     docs[0] = (indent, Mode::Broken, (*doc).clone());
                 }
             }
+
+            Document::FlexGroup(doc) => {
+                if !fits(limit - width, vector![(indent, Mode::Flex, (*doc).clone())]) {
+                    docs.push_front((
+                        indent,
+                        Mode::Flex,
+                        line().nest(2).append((*doc).clone()).append(line()),
```
lpil, line 257: The document should not have hard coded indents in it, instead we should apply the nested when constructing the algebra.

With hard coded nesting this won't work for Erlang, which uses 4 space indentation, or for plain text, which doesn't indent at all.

src/pretty.rs
```diff
@@ -182,6 +209,18 @@ fn fmt(b: &mut String, limit: isize, mut width: isize, mut docs: Vector<(isize,
                         b.push_str(" ".repeat(indent as usize).as_str());
                         indent as isize
                     }
+                    Mode::Flex => {
+                        if limit - width < broken.len() as isize {
+                            remove_ws_end(b, 1);
+                            b.push_str("\n");
+                            b.push_str(" ".repeat(2).as_str());
```
lpil, line 216: Same indentation situation here, it should respect the current indent level rather than hard coding 2.

src/pretty.rs
```diff
@@ -182,6 +209,18 @@ fn fmt(b: &mut String, limit: isize, mut width: isize, mut docs: Vector<(isize,
                         b.push_str(" ".repeat(indent as usize).as_str());
                         indent as isize
                     }
+                    Mode::Flex => {
+                        if limit - width < broken.len() as isize {
+                            remove_ws_end(b, 1);
```
lpil, line 214: This doesn't seem right to me. Why do we have trailing whitespace? The Document algebra shouldn't know anything about the content it contains, the formatting rules should be entirely in the structure of the algebra and not conditional on whether it contains content. I think the best approach here is to not construct a Document that renders trailing whitesapce, as it currently does on master.

src/format.rs
lpil: Thanks :)

PR #636, opened 2020-06-05
More bit string improvements

### LINKED ISSUE #628: Rename Bitstring `integer` annotation to `int` to match the Int type

### LINKED ISSUE #634: Should gleam allow the reuse of a variable in a pattern?
This example is from the Elixir docs, but the equivalent works in Erlang too. 

```elixir
<<name_size::size(8), name::binary-size(name_size), " the ", species::binary>> = <<5, "Frank the Walrus">>
```
name_size is bound and then used in the next segment of the same pattern. This is currently impossible in Gleam, you get a `DuplicateVarInPattern` error. I actually added the DuplicateVarInPattern error, but I must admit that I don't fully understand the reasoning for it. Is this behaviour something we:

1. Deliberately don't want to support
2. Would like support to but it's impossible with the current type system
3. Would like to support and we just haven't done it yet

tomwhatmore:
Closes #634 
Closes #628

Additionally makes int the default type for all segments in bit strings, matching Elixir and Erlang.

### DIFF
--- a/test/core_language/test/bit_string_test.gleam
+++ b/test/core_language/test/bit_string_test.gleam
@@ -9,7 +9,7 @@ fn integer_fn() {
 // Valid values
 
 pub fn function_as_value_test() {
-  let <<a>> = <<integer_fn():integer>>
+  let <<a>> = <<integer_fn():int>>
 
   should.equal(a, 1)
 }
--- a/test/core_language/test/bit_string_test.gleam
+++ b/test/core_language/test/bit_string_test.gleam
@@ -1,21 +1,38 @@
 import should
 
 // Helpers
-
 fn integer_fn() {
   1
 }
 
 // Valid values
-
 pub fn function_as_value_test() {
   let <<a>> = <<integer_fn():int>>
 
   should.equal(a, 1)
 }
 
-// Strings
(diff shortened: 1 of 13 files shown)


test/core_language/test/bit_string_test.gleam
```diff
@@ -9,11 +9,26 @@ fn integer_fn() {
...
+pub fn integer_to_binary_test() {
+  let <<a, rest:binary>> = <<1, 17, 42:16>>
+
+  should.equal(rest, <<17, 0, 42>>)
+}
+
+// Sizes
+
+pub fn size_variable_from_match_test() {
+  let <<name_size:8, name:binary-size(name_size), " the ":utf8, species:binary>> = <<5, "Frank the Walrus":utf8>>
```
lpil, line 26: Love this example!

Has the formatter been run on this file? Seems like a long line.

tomwhatmore: Ah, it has not. Let me fix that!

tomwhatmore: Turns out we hadn't formatted the core_language project for a while (ever?) so there's some bonus clean up in a new commit

PR #640, opened 2020-06-07
Fix 611 [Pt 1]

scileo:
The following changes were made:
  - some members of `Document` where changed: `String` was replaced by `Cow<'a, str>`,
  - a copy was removed in the implementation of `Documentable` for `&str`.

These two little changes triggered a lot of errors which are now fixed.


scileo:
Thanks for the time you took to review my changes. I did my very best and tried to change code as conservatively as possible.

I understand your concerns about the huge amount of complexity added by this PR. So far, I have noticed three patterns that could be fixed with the use of a procedural macro. They are:
  - destructuring structs wrapped in `Cow` forces us to `match` over the two `Cow` variants (see [an example here](https://github.com/scileo/gleam/blob/81fcc066a99614f486e7472ef9341ace0fb1bc9e/src/erl.rs#L767)),
  - `match`ing on enums wrapped in `Cow` nearly doubles the amount of match arms, resulting in a lot of duplicate code (see [here](https://github.com/scileo/gleam/blob/81fcc066a99614f486e7472ef9341ace0fb1bc9e/src/erl.rs#L960)),
  - creating an iterator from a `Cow<T>` where `T` implements `IntoIterator` forces us to match over the two `Cow` variants, resulting again in a lot of duplication (see [here](https://github.com/scileo/gleam/blob/81fcc066a99614f486e7472ef9341ace0fb1bc9e/src/erl.rs#L246)).

For these three patterns, I have a solution which I am implementing on a crate I entitled [`butcher`](https://github.com/scileo/butcher). The fix for patterns 1 and 2 will remove a lot of boilerplate at no runtime cost. The fix for pattern 3 will introduce a `match` at each iteration step.

There is also some complexity that can't (as far as I know) be simplified. For example, many lifetimes have to be written manually now. They are not hard to understand, but definitely add some complexity.

This list may not be complete. Don't hesitate to report every pattern introduced here that you want to disappear, so I can investigate to find a fix.

I'll work on `butcher` to make it usable, and then remove as much duplicate as possible.

Issue #577, opened 2020-05-17
Module level constants


sporto:
It would be great to have top level constants in a module that can be used for pattern matching.

For example, I want to parse a string into a custom type.

```rust
const one = "one"
const two = "two"

type Enum {
  One
  Two
}

fn enum_to_code(enum: Enum) -> String {
  case enum {
    One -> one
    Two -> two
  }
}

fn code_to_enum(code: String) -> Result(Enum, String) {
  case code {
    one -> Ok(One)
    two -> Ok(Two)
    _ -> Err("Unknown " ++ code)
  }
}
```

This cannot be achieved with functions

```
  // invalid
  case code {
    one() -> Ok(One)
```

At the moment I would need to repeat the strings in two places:

```
fn enum_to_code...
  case enum {
    One -> "one"
   ...

fn code_to_enum...
  case code {
    "one" -> Ok(One)
  ...
```

Having these constant would allow to only have one source of truth.

lpil:
We would only permit literals to be used as constants, no function calls etc. 

One question I have is what Erlang should be generated for this.

I think it would be good to create a zero arity function for the definition.

Where it is used we could inline the constant, or call the function. I'm not sure which is better.

gamebox:
I would assume you would either inline the constant or generate a macro for the constant and use that in place of it

lpil:
For Gleam use inlining would work, but we'll need something else for Erlang + Elixir users. A macro is good but would only be usable from within the same module.

OvermindDL1:
> We would only permit literals to be used as constants, no function calls etc.

No constant expressions?  ;-)

But yeah, probably makes sense to just replace it inline in code, or use an erlang macro to do the same thing.

lpil:
Not for the first iteration, but later that would be a great addition! A little complicated as we don't have a way to evaluate Gleam code within the Rust compiler, but we can cross that bridge later.

PR #643, opened 2020-06-08
Module constants

thehabbos007:
Draft PR for #577 

Constants can only be simple literals: ints, floats and strings in this draft.
Constants can be marked as public (I have yet to tackle what this will imply) and they are compiled to erl functions for now.

```rust
pub const hello = 5
const again = 7
```
```erlang
hello() -> 5.
again() -> 7.
```
I thought it was neat to have exported constants on top of private constants, is that sensible?
We could have private consts inline and public ones expose a function.

Any pointers/feedback is very much appreciated. This needs defintely more testing too :)

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -192,6 +192,15 @@ pub enum Statement<T, Expr> {
         as_name: Option<String>,
         unqualified: Vec<UnqualifiedImport>,
     },
+
+    ModuleConstant {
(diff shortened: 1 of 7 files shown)


thehabbos007:
Yes, that sounds good. I did restrict the values to a LiteralExpr in the grammar, so will also add the new AST :)

PR #657, opened 2020-06-12
Bit string size and unit option fixes

### LINKED ISSUE #656: Bit string syntax can render invalid Erlang
```rust
<<var:size(2 * 2)>>
```

Renders `<<Var:2 * 2>>`, which is not valid.

tomwhatmore:
Closes #656 

Size options in bit string construction with anything other than Int literals or vars are now wrapped in parentheses.

Additionally, valid unit options now work properly. I'm going to do some more work on them to provide better error messages if you try and use invalid values, but that will be a separate PR.

### DIFF
--- a/test/core_language/test/bit_string_test.gleam
+++ b/test/core_language/test/bit_string_test.gleam
@@ -32,6 +32,13 @@ pub fn size_variable_from_match_test() {
   should.equal(species, <<"Walrus":utf8>>)
 }
 
+pub fn sizes_with_expressions() {
+  let a = 1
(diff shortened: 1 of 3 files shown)


test/core_language/test/bit_string_test.gleam
```diff
@@ -32,6 +32,13 @@ pub fn size_variable_from_match_test() {
   should.equal(species, <<"Walrus":utf8>>)
 }
 
+pub fn sizes_with_expressions() {
```
lpil, line 35: I think this needs a name with _test at the end

PR #663, opened 2020-06-14
Module constants

thehabbos007:
PR for #577 

Constants can only be simple literals: ints, floats and strings, and are inlined at gleam compile time.

```rust
pub const hello = 5
pub const pi = 3.14
const again = "world"
```

### DIFF
--- a/src/erl/tests.rs
+++ b/src/erl/tests.rs
@@ -823,18 +823,6 @@ go() ->
         },
         name: vec!["funny".to_string()],
         statements: vec![
-            Statement::ModuleConstant {
-                doc: None,
-                location: Default::default(),
-                name: "test".to_string(),
-                public: true,
-                typ: crate::typ::int(),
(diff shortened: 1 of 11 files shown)


src/erl/tests.rs
```diff
@@ -2413,6 +2471,37 @@ x() ->
...
     A = <<(x())/integer>>,
     A.
+"#,
+    );
+
+    assert_erl!(
+        r#"
+pub const string_value = "constant value"
+pub const float_value = 3.14
+pub const int_value = 42
```
lpil, line 2481: Love the choice of values! :)

lpil:
There are conflicts that mean I can't rebase the changes in.

Would you like to rebase this branch to be up to date with master? Alternatively I can squash rebase this in, but then we lose the commit history.

thehabbos007:
I can get to do that later today :)

Issue #671, opened 2020-06-22
Allow type annotations in module constants


lpil:
```rust
const my_constant: String = "Hello"
```

lpil:
@thehabbos007 Are you working on the types part of this now?

thehabbos007:
Yes :)

lpil:
Thanks!

PR #673, opened 2020-06-23
Addresses https://github.com/gleam-lang/gleam/issues/671

oneness:


thehabbos007:
Hello! I looked into constant type annotations last night, and seems like we did it almost the exact same way. I'm wondering if we should unify the annotation type and the value of the constant statement itself here
https://github.com/gleam-lang/gleam/blob/248c925719a0e30cc784d5875b9dc9d5b680d3e0/src/typ.rs#L3435
Just so we can get proper type mismatch errors.

https://github.com/thehabbos007/gleam/blob/const_annotations/src/typ.rs#L2310-L2315
I did this within the infer_const in my version :)

oneness:
Good point. I think I missed to add the piece where we do not need to do
type inference for annotated types.  `infer_const` fn is intended to be
used for untyped expr to infer its type. Since it already have type
annotations, it should be treated as TypedConstValue and inference is not
necessary.  Will look into when I have some free time.  But feel free to
add to yours or mine PR to the missing piece.
Kasim Tuman

==========================================

Experience is more about knowing what should be avoided

than what should be done. -- Kasim
============================================

src/format.rs
```diff
@@ -332,14 +335,19 @@ impl<'a> Formatter<'a> {
...
-            .append(match &arg.annotation {
-                Some(a) => ": ".to_doc().append(self.type_ast(a)),
-                None => nil(),
-            })
+            .append(self.annotation_(": ", &arg.annotation))
             .group();
         commented(doc, comments)
     }
 
+    fn annotation_(&mut self, prefix: &str, annotation: &Option<TypeAst>) -> Document {
```
lpil, line 343: Could we hard-code the prefix to `: `? When would it be something else?

lpil: Ah right I get it now! Thanks

PR #677, opened 2020-06-24
Module constants annotation type checking and docs generation

### LINKED ISSUE #665: List module constants in generated HTML docs

### LINKED ISSUE #671: Allow type annotations in module constants

thehabbos007:
Closes #671 
~~Had the code for docs generation ready too, but with the addition of tuple and list constants I ran into some issues, so only submitting the type checking.~~

Closes #665 too now. Example in the comment below

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -239,7 +239,7 @@ impl<'a> Formatter<'a> {
         }
     }
 
-    fn const_expr(&mut self, value: &ConstValue<()>) -> Document {
+    pub fn const_expr<T>(&mut self, value: &ConstValue<T>) -> Document {
         match value {
             ConstValue::Int { value, .. } | ConstValue::Float { value, .. } => {
                 value.clone().to_doc()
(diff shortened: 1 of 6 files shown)


src/format.rs
```diff
@@ -239,7 +239,7 @@ impl<'a> Formatter<'a> {
         }
     }
 
-    fn const_expr(&mut self, value: &ConstValue<()>) -> Document {
+    pub fn const_expr<T>(&mut self, value: &ConstValue<T>) -> Document {
```
lpil, line 242: Does this need to be public and generic?

src/format.rs
```diff
@@ -261,6 +261,23 @@ impl<'a> Formatter<'a> {
...
+        name: &str,
+        value: &TypedConstValue,
+    ) -> Document {
+        let mut printer = typ::pretty::Printer::new();
+
+        pub_(public)
+            .append("const ")
+            .append(name.to_string())
+            .append(": ")
+            .append(printer.print(&value.typ().as_ref()))
```
lpil, line 276: Can we remove the `&` here as we're calling `.as_ref`?

src/typ.rs
```diff
@@ -2308,7 +2321,7 @@ impl<'a> Typer<'a> {
         Ok(ConstValue::List {
             elements,
             location,
-            typ,
+            typ: list(typ),
```
lpil, line 2324: 💜

src/format.rs
thehabbos007: I believe so, since `docs_const_expr` is used in docs and an error is thrown about associated functions since `docs_const_expr` uses `const_expr`. The value given from `docs_const_expr` would also be typed. I'll see if i can change it a little :)

lpil: Ah yes! Makes sense. We can make it private though?

thehabbos007: Okay, at least i could remove pub is that good enough?

lpil:
💜 Thank you!

PR #679, opened 2020-06-25
Fix bug when reassigning variable names in alternative patterns

tomwhatmore:
The bodies of the duplicated clauses created by using alternative patterns were incorrectly incrementing the variable names if you reassigned them. For example:

```rust
pub fn test() {
  let test = 1

  case 1 {
    1 | 2 -> { let test = test + 1; test }
  }
}
```

would output

```erlang
test() ->
    Test = 1,
    case 1 of
        1 ->
            Test1 = Test + 1,
            Test1;

        2 ->
            Test2 = Test1 + 1,
            Test2
    end.
```

This commit changes it to use the same variable names in each clause.

However, it does so by cloning the whole env everytime you render a clause, regardless of whether you're even using alternative patterns at all. I'm not quite how big a deal this is in practice but it could certainly be improved. Thought it would be worth checking if you reckon this general approach is the right solution before spending time optimising it though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 - Error message diagnostic code previews for type errors when using the the
   pipe operator have been made more accurate.
 - Added support for list literals in clause guards.
+- Fixed bug when reassigning a variable inside a case clause with alternative
+  patterns.
 
 ## v0.9.1 - 2020-06-12
 
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -726,8 +726,7 @@ fn clause(clause: &TypedClause, env: &mut Env) -> Document {
         ..
     } = clause;
 
-    let mut pattern_already_printed = false;
-    let env_clone = &mut env.clone();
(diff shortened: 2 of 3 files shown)


lpil:
Good find. I don't think I did the best job here.

I'm wondering if it would be better to clone the document generated for the `then` rather than generating it multiple times. What do you think?

If we are to clone the vars (which could be the way to go) then I would rather clone just the fields needed to be reset rather than the whole Env. Currently it's quite cheap to clone (as we're using `im` HashMaps) but I'd like to not start accidentally cloning large objects at a later date if new fields are added.

tomwhatmore:
Cloning the generated doc sounds like the best plan here for sure. I'll give it a go.

tomwhatmore:
Done. This should be much better than rebuilding that bit of the document for every alternative pattern. Two improvements for the price of one!

tomwhatmore:
Updated changelog and rebased.

lpil:
Thank you! Excellent as always

PR #681, opened 2020-06-26
Todo labels

### LINKED ISSUE #678: Add a label to todo.
Record of a conversation that was had on IRC

https://matrix.to/#/!xBiwxFiVfVDbdvbbBu:matrix.org/$6eeTnz9NwRgBHWZSrQr7B0t0ry2HKkITR3PZduhAkK8?via=matrix.org&via=mozilla.org

```rust
case receive(Infinity)  {
  Ok(message) -> continue
  Error(Nil) -> todo("We should have waited forever")
}
```

tomwhatmore:
Closes #678

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Added support for list literals in clause guards.
 - Fixed bug when reassigning a variable inside a case clause with alternative
   patterns.
+- Todos can now take an optional label.
 
 ## v0.9.1 - 2020-06-12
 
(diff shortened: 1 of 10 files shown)


lpil:
Could we add the label to the Erlang error that gets thrown? Would make it easier to determine which todo was hit.

tomwhatmore:
That was quick!

lpil:
What else am I supposed to do during lock down? :P

Issue #646, opened 2020-06-10
Don't exit on ctrl+c when running `gleam shell`


lpil:
`ctrl-c` is used by the child Erlang shell, we don't want to exit when it is sent by the user.

PR #684, opened 2020-06-27
Don't exit on ctrl+c when running `gleam shell`

oneness:
Addresses:  https://github.com/gleam-lang/gleam/issues/646

**Note:**  It does bring in couple of dependencies such as nix and winapi, which BTW is fantastic libs.

### DIFF
--- a/src/shell.rs
+++ b/src/shell.rs
@@ -1,11 +1,10 @@
 use crate::{
     build::{self, project_root::ProjectRoot},
-    error::Error,
+    error::{Error, GleamExpect},
     file,
 };
 use std::path::PathBuf;
 use std::process::Command;
-extern crate ctrlc;
 
 pub fn command(root_string: String) -> Result<(), Error> {
     let root_path = PathBuf::from(root_string);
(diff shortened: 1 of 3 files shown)


src/shell.rs
```diff
@@ -5,8 +5,8 @@ use crate::{
 };
 use std::path::PathBuf;
 use std::process::Command;
+extern crate ctrlc;
```
lpil, line 8: We can remove this as it's not needed in Rust 2018

src/shell.rs
```diff
@@ -15,6 +15,12 @@ pub fn command(root_string: String) -> Result<(), Error> {
     // Build project
     build::main(config, root_path)?;
 
+    // Don't exit on ctrl+c as it is used by child erlang shell
+    ctrlc::set_handler(move || {
+        println!("\nIf you would like to exit this shell, please use Ctrl+g to switch to user switch prompt and type `q` to quit");
```
lpil, line 20: Let's stick with the default Erlang/Elixir and have just the erl shell text print here, no need to print anything else.

src/shell.rs
```diff
@@ -15,6 +15,12 @@ pub fn command(root_string: String) -> Result<(), Error> {
...
+    })
+    .expect("Error setting Ctrl-C handler");
```
lpil, line 22: Let's use `.gleam_expect` here as it provides a nicer error. You'll need to import `crate::error::GleamExpect`.

PR #689, opened 2020-06-28
Fix unused return value warning in nightly

shritesh:
On Rust nightly, I get the following error:
Screenshot from 2020-06-28 10-20-16

The PR fixes the warning


lpil:
We don't support nightly so I'm reluctant to merge this change.

What does mem::replace return on this nightly version? The warning implies we should be handling something.

shritesh:
`std::mem::replace` always returns the previous dest value which is an `Option<String>` here. IIUC,  in nightly (therefore future Rust versions), the `unused_must_use` lint enforces more strictly that the `Option` be used.

PR #692, opened 2020-06-28
Ignore files from .gitignore during format

### LINKED ISSUE #542: Ignore gitignored files when formatting a directory
If we run `gleam format` in a project it should not format the contents of the `_build` directory (or any other gitignored directories).

If there is no gitignore we should continue without filtering.

This library looks useful: https://github.com/nathankleyn/gitignore.rs

shritesh:
Closes #542

This is the most readable I could make the code given `gitignore`'s API. Suggestions welcome

### DIFF
--- a/src/format/command.rs
+++ b/src/format/command.rs
@@ -96,10 +96,7 @@ pub fn read_and_format_paths(files: Vec<String>) -> Result<Vec<Formatted>, Error
         })?;
 
         if path.is_dir() {
-            let mut gitignore_file = path.clone();
-            gitignore_file.push(".gitignore");
-
-            if let Ok(ignore) = gitignore::File::new(&gitignore_file) {
+            if let Ok(ignore) = gitignore::File::new(&path.join(".gitignore")) {
                 for path in file::gleam_files(&path)
                     .filter(|f| {
                         let excluded = ignore.is_excluded(f).unwrap_or(false);
--- a/src/format/command.rs
+++ b/src/format/command.rs
@@ -96,8 +96,23 @@ pub fn read_and_format_paths(files: Vec<String>) -> Result<Vec<Formatted>, Error
         })?;
 
         if path.is_dir() {
-            for path in file::gleam_files(&path).into_iter() {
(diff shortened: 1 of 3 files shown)


src/format/command.rs
```diff
@@ -96,8 +96,23 @@ pub fn read_and_format_paths(files: Vec<String>) -> Result<Vec<Formatted>, Error
         })?;
 
         if path.is_dir() {
-            for path in file::gleam_files(&path).into_iter() {
-                formatted_files.push(format_file(path)?);
+            let mut gitignore_file = path.clone();
+            gitignore_file.push(".gitignore");
```
lpil, line 100: Can we use `join` rather than `clone` + `push`?

src/format/command.rs
```diff
@@ -96,8 +96,23 @@ pub fn read_and_format_paths(files: Vec<String>) -> Result<Vec<Formatted>, Error
...
+                        !excluded
+                    })
+                    .into_iter()
+                {
+                    formatted_files.push(format_file(path)?);
+                }
+            } else {
+                for path in file::gleam_files(&path).into_iter() {
+                    formatted_files.push(format_file(path)?);
+                }
```
lpil, line 115: Looks good! Perhaps we could reshuffle it a bit like this? (pseudocode)

```rust
let ignore = gitignore::File::new(&gitignore_file)
let not_ignored = |f| match ignore {
    Ok(ignore) -> !ignore.is_excluded(f).unwrap_or(false),
    Err(_) -> true,
};

for path in file::gleam_files(&path)
    .filter(not_ignored)
    .into_iter()
{
    formatted_files.push(format_file(path)?);
}
```

src/format/command.rs
shritesh: Definitely! I didn't know .join was a thing!!

shritesh:
Closing because of https://github.com/nathankleyn/gitignore.rs/issues/15

lpil:
Thank you, sorry for putting you down a bad path there!

shritesh:
No problem!!

PR #693, opened 2020-06-28
This solves issue #542

namoscagnm:


lpil:
Thank you so much for this. I'm sorry but it seems Shritesh and you have clashed and picked up the same ticket. I'm going to go with Shritesh's as they opened https://github.com/gleam-lang/gleam/pull/692 first, in future I'll try and make sure tickets get assigned to avoid this happening again.

PR #694, opened 2020-06-28
Ignore files from .gitignore during format

### LINKED ISSUE #542: Ignore gitignored files when formatting a directory

shritesh:
Closes #542 using the `ignore` crate.

I have followed the pattern from `gleam_files` but the ignore crate comes with a lot of functionality (see [WalkBuilder](https://docs.rs/ignore/0.4.16/ignore/struct.WalkBuilder.html) and [types](https://docs.rs/ignore/0.4.16/ignore/types/index.html)) that can delete a lot of code from the `file` module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Public module level constants generate documentation.
 - The formatter style has been improved to wrap and sort imports.
 - The formatter now permits comments at the end of module function bodies.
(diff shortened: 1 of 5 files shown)


shritesh:
Done!

lpil:
Thank you!

PR #698, opened 2020-06-29
Add index.html to end of module doc path

shritesh:
Currently, module doc links show a directory listing page when viewed locally in the browser without using a web server.

### DIFF
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -65,7 +65,7 @@ pub fn generate_html(
         .clone()
         .map(|m| {
             let name = m.name.join("/");
-            let path = [&name, "/"].concat();
+            let path = [&name, "/index.html"].concat();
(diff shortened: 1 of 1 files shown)


lpil:
Thanks for this, but this is by design. We want to use "pretty" URLs that do not include the `/index.html` part and will be optimising for production usage from a web server rather than usage with a browser's offline file explorer.

shritesh:
I hope you reconsider this in the future. Having to go to HexDocs across the internet or spin up an HTTP server (using another language's package) just to view documentation properly is not a great development experience. ex_doc, rustdoc and many other modern language toolings prominently showcase "works great offline" as a feature.

lpil:
I would be open to adding a web server to the compiler to serve the docs, that would solve the problem

Issue #477, opened 2020-04-22
Add Gleam to Linguist


gamebox:
Looks like we can just add Gleam to Linguist.  I’ll link an example PR.  You can assign as well since I’m not going to do much more on incremental until v0.8 ships

https://github.com/github/linguist/pull/4759

lpil:
I like this idea a lot but I'm not sure we qualify yet. [Here](https://github.com/github/linguist/blob/master/CONTRIBUTING.md#adding-an-extension-to-a-language) it say we should have 1000s of code examples and be able to demonstrate real world code on GitHub.

brightly-salty:
Just an FYI: We need "hundreds of repositories". We only need thousands of code examples if the extension conflicts with one already in Linguist. I think this might be worthwhile to pursue, especially as Linguist lets PRs linger if they are not ready yet.

lpil:
Wonderful news! Would someone like to take the lead here?

brightly-salty:
I would in general, but I'm not sure if I know enough about gleam to create the grammar for it. I'd be happy to submit the Linguist PR if we could find/create the grammar file.

lpil:
That would be great @brightly-salty ! I'll help out if any issues arise.

I believe we can use the grammar from the vscode plugin https://github.com/gleam-lang/vscode-gleam/blob/main/syntaxes/gleam.tmLanguage.json

lpil:
@brightly-salty After doing some digging it seems we have about 70 repos on GitHub. This is not enough yet, so let's focus on other things + getting more Gleam code on GitHub.

Aloso:
I think there are enough repos using Gleam now. I searched GitHub and found a lot of repos solving Advent of Code 2021 in Gleam.

So I downloaded the first 1000 file entries that have ".gleam" in their path, filter by paths that end with ".gleam" and group by repository, which told me that the files are from **95** different repositories. And I was only able to download 1000 file entries due to rate limits; according to the GitHub API there are over 11,000.

Aloso:
@lpil @gamebox I filed a PR: https://github.com/github/linguist/pull/5706

michallepicki:
Current PR for tracking: https://github.com/github/linguist/pull/5688

lpil:
It's finally here!!!

PR #705, opened 2020-06-30
fixed semicolon in readme

dym-sh:
9BB3E6B0-50B5-4480-A96A-E326F63E7275

ಠ_ಠ

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ It looks like this:
 pub type Tree(value) {
   Leaf(value)
   Branch(Tree(value), Tree(value))
-};
+}
 
 pub fn any(tree: Tree(a), check: fn(a) -> Bool) -> Bool {
   case tree {


lpil:
Looks like this broke the fake syntax highlighting. Doh :( 

I'll put it back until we have #477

dym-sh:
but it works for mdBook: https://github.com/gleam-lang/website/blob/master/src/README.md

so the problem is in Linguist's rust parser?

lpil:
That's right, GitHub doesn't yet support Gleam

dym-sh:
i mean the rust parsing/highlighting stops once it encounters "semicolon expected"
but the mdBook in docs also has `rust` for the same code and highlights entire block regardless

lpil:
Yes, that's right. I'm not sure why they do that

Issue #34, opened 2018-10-11
Full exhaustiveness checking of assignments and case expressions


lpil:
edit: See comments below for current status.

lpil:
I was lucky enough to see Simon Peyton Jones talk about how this is now implemented in the Haskell compiler at Code Mesh LDN 2019 (talk title: REVISITING PATTERN MATCH OVERLAP CHECKS IN HASKELL).

The approach sounded good but the paper has not been released yet. Either work off the slides or see if the paper is out when we get around to implementing this.

I believe the same talk was given (possibly in more detail) at the Haskell eXchange 2019.

OvermindDL1:
Can also see the OCaml compilers code for it, it's quite clean.  :-)

michallepicki:
Here is P. Sestoft: ML pattern match compilation and partial evaluation: http://dotat.at/tmp/match.pdf

michallepicki:
Paper mentioned in SPJ's talk: https://www.microsoft.com/en-us/research/wp-content/uploads/2016/08/gadtpm-acm.pdf

michallepicki:
SPJ's talk got published: https://www.youtube.com/watch?v=SWO5OzSxD6Y

CrowdHailer:
From the list of papers I guess this is not a good issue to start with.
Is there anything I can help with. is the SPJ approach the one that is accepted as the "right" approach for this issue?

CrowdHailer:
From the list of papers I guess this is not a good issue to start with.
Is there anything I can help with. is the SPJ approach the one that is accepted as the "right" approach for this issue?

lpil:
> is the SPJ approach the one that is accepted as the "right" approach for this issue?

I didn't know how to evalute the options so I cornered SPJ at a conference and he told me I should use that one. Good enough for me!

QuinnWilton:
I've been reading the SPJ paper tonight, and I think there's a few simplifications that apply to Gleam.

1) Since Gleam is strict, we don't need to worry about handling divergent sets or strictness constraints at all
2) Since Gleam doesn't support GADTs, we don't need to track type equalities (in rules `CCONVAR` and `UCONVAR`, see 4.3)
3) The satisfiability oracle can ignore all constraints and the algorithm remain sound (section 6.1)

By ignoring (3) to start, we lose exhaustiveness checking of guards, but it becomes possible to add support later without changing the core algorithm. This lets us avoid also having to write a type-constraint solver for the initial version.

[...]

PR #706, opened 2020-07-01
Make let binding with a type that has multiple constructors an error

### LINKED ISSUE #537: Using let with a type that has multiple constructors should be an error
```rust
let Ok(x) = result
```
should throw an error and tell you to use assert instead.

tomwhatmore:
Closes #537

I was originally going to store a vec of all constructor names on the `TypeConstructor`, but it seems to be a massive faff getting the relevant one from the starting point of a `Pattern::Constructor`, and also meant that types like Int and String had these empty constructor lists on them too which seemed redundant.

Instead I've added a vec of other constructor names to `ValueConstructorVariant::Record`, which is much easier to access, and means only the custom record types have to deal with this extra list. The downside is that this means the list of alternatives is duplicated across each constructor, rather than being stored once. I think you'd need to define a type with a completely impractical number of constructors for this to make any real difference, but seemed worth mentioning. What do you think?

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -592,6 +592,7 @@ pub enum BindingKind {
     Let,
     Assert,
     Try,
+    Clause,
 }
 
 // BitStrings
--- a/src/error.rs
+++ b/src/error.rs
@@ -1176,6 +1176,33 @@ signed, unsigned, big, little, native, size, unit",
                     )
                     .unwrap();
                 }
+
+                NonExhaustiveLet {
+                    location,
+                    constructor,
+                    unhandled_constructors,
+                } => {
+                    let diagnostic = Diagnostic {
+                        title: "Non-exhaustive let binding".to_string(),
+                        label: "used here".to_string(),
+                        file: path.to_str().unwrap().to_string(),
+                        src: src.to_string(),
+                        location: location.clone(),
+                    };
+                    write(buffer, diagnostic, Severity::Error);
+                    writeln!(
+                        buffer,
+                        "You are handling the {} constructor, but this could also be one of:
+{}
+
+Either use a case expression to handle all possibilities, or use assert instead
+of let if you are sure you don't want to handle the others. This will cause a
+runtime error if they occur.",
+                        constructor,
+                        unhandled_constructors.join(", "),
+                    )
+                    .unwrap();
+                }
             },
 
             Error::Parse { path, src, error } => {
--- a/src/error.rs
+++ b/src/error.rs
@@ -1177,7 +1177,7 @@ signed, unsigned, big, little, native, size, unit",
                     .unwrap();
                 }
 
-                NonExhaustiveLet {
+                NonExhaustiveBinding {
                     location,
                     constructor,
                     unhandled_constructors,
--- a/src/error.rs
+++ b/src/error.rs
@@ -1180,26 +1180,35 @@ signed, unsigned, big, little, native, size, unit",
                 NonExhaustiveBinding {
                     location,
                     constructor,
+                    kind,
                     unhandled_constructors,
                 } => {
                     let diagnostic = Diagnostic {
-                        title: "Non-exhaustive let binding".to_string(),
+                        title: "Non-exhaustive binding".to_string(),
                         label: "used here".to_string(),
                         file: path.to_str().unwrap().to_string(),
                         src: src.to_string(),
                         location: location.clone(),
                     };
                     write(buffer, diagnostic, Severity::Error);
+
+                    let suggestion = match kind {
+                        crate::ast::BindingKind::Let => "Either use a case expression to handle all possibilities, or use assert instead
+of let if you are sure you don't want to handle the others. This will cause a
+runtime error if they occur.",
+                        crate::ast::BindingKind::Try => "You will need to use a case expression to handle all possibilities.",
+                        _ => "Additionally, you have found a compiler bug. Sorry! Please submit a bug report on GitHub."
+                    };
+
                     writeln!(
                         buffer,
                         "You are handling the {} constructor, but this could also be one of:
 {}
 
-Either use a case expression to handle all possibilities, or use assert instead
-of let if you are sure you don't want to handle the others. This will cause a
-runtime error if they occur.",
+{}",
                         constructor,
                         unhandled_constructors.join(", "),
+                        suggestion
                     )
                     .unwrap();
                 }
--- a/src/typ.rs
+++ b/src/typ.rs
@@ -293,13 +293,16 @@ pub enum ValueConstructorVariant {
     Record {
         name: String,
         field_map: Option<FieldMap>,
+        other_constructor_names: Vec<String>,
     },
 }
 
 impl ValueConstructorVariant {
     fn to_module_value_constructor(&self) -> ModuleValueConstructor {
         match self {
-            ValueConstructorVariant::Record { name, field_map } => ModuleValueConstructor::Record {
+            ValueConstructorVariant::Record {
+                name, field_map, ..
+            } => ModuleValueConstructor::Record {
                 name: name.clone(),
                 arity: field_map.as_ref().map_or(0, |fm| fm.arity),
             },
@@ -428,6 +431,7 @@ impl<'a> Typer<'a> {
             ValueConstructorVariant::Record {
                 name: "True".to_string(),
                 field_map: None,
+                other_constructor_names: vec!["False".to_string()],
             },
             bool(),
         );
@@ -436,6 +440,7 @@ impl<'a> Typer<'a> {
             ValueConstructorVariant::Record {
                 name: "False".to_string(),
                 field_map: None,
+                other_constructor_names: vec!["True".to_string()],
             },
             bool(),
         );
@@ -512,6 +517,7 @@ impl<'a> Typer<'a> {
             ValueConstructorVariant::Record {
                 name: "Nil".to_string(),
                 field_map: None,
+                other_constructor_names: vec![],
             },
             nil(),
         );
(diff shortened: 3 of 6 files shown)


src/typ.rs
```diff
@@ -1523,6 +1531,29 @@ impl<'a> Typer<'a> {
                 .map_err(|e| convert_unify_error(e, then.try_binding_location()))?;
         }
 
+        // Ensure exhaustiveness of value if this is a `let` binding
+        if kind == BindingKind::Let {
```
lpil, line 1535: We would want the same for `try` too, so `!= BindingKind::Assert` perhaps?

```rust
try "partial pattern" = Ok("no match")
```

src/typ.rs
```diff
@@ -2783,6 +2814,12 @@ pub enum Error {
         location: SrcSpan,
         label: String,
     },
+
+    NonExhaustiveLet {
```
lpil, line 2818: `NonExhaustiveBinding`?

src/error.rs
```diff
@@ -1176,6 +1176,33 @@ signed, unsigned, big, little, native, size, unit",
...
+                        location: location.clone(),
+                    };
+                    write(buffer, diagnostic, Severity::Error);
+                    writeln!(
+                        buffer,
+                        "You are handling the {} constructor, but this could also be one of:
+{}
+
+Either use a case expression to handle all possibilities, or use assert instead
+of let if you are sure you don't want to handle the others. This will cause a
```
lpil, line 1199: Could be a `try` rather than a `let`

src/typ.rs
tomwhatmore: Wouldn't a try binding already have been type checked as being a result and be implicitly handling both the Ok and Error constructors?

src/typ.rs
tomwhatmore: Sure

tomwhatmore:
A test for the nested version is an excellent idea, will add one.

src/typ.rs
lpil: Yes but the contained value might not be matched, as with that string pattern in the example above

tomwhatmore: Ah yes, got you.

src/error.rs
tomwhatmore: Might be worth having two different error messages then, one for let and one for try, as we can't really suggest using assert if you're currently using try.

lpil: Oh yes, good point, I didn't think of that.

tomwhatmore: Any preference between implementing two different errors altogether or just one with an extra binding type field and using that to show two different bits of text?

tomwhatmore: Went with the latter

tomwhatmore:
Added a test for nested pattern constructors, and did some refactoring to make it pass. The logic now lives inside the `PatternTyper` itself rather than the `infer_let()`, which I think makes more sense actually, especially when more exhaustiveness checking logic is added. This requires the `PatternTyper` to know what kind of binding it's being used in, so I added a new `BindingKind::Clause` variant for when you are using it in a case clause.

I've also renamed the error to not be let specific, and added different suggestion text depending on whether it was a try or let binding.

src/ast.rs
```diff
@@ -592,6 +592,7 @@ pub enum BindingKind {
     Let,
     Assert,
     Try,
+    Clause,
```
lpil, line 595: This doesn't seem right to me. It's now possible to construct a binding with a `clause` kind, which would render invalid Gleam in the formatter. I think as much as possible we could not have data structures that permit invalid states.

tomwhatmore: That seems reasonable. I'm trying to use it for two purposes I guess, and it would probably be better to keep them separate. Are you happy with the general idea of putting this logic in the `PatternTyper` with the extra binding_kind field? If so I could make a new enum for this purpose which would leave BindingKind unchanged.

```rust
// Not sure about the name yet!
enum PatternUsage {
    Binding(BindingKind)
    Clause
}
```

lpil: https://github.com/gleam-lang/gleam/blob/d7b2fb66ecfb8c4fa7cbf32b1143f81bc9320568/src/typ.rs#L1630-L1637

This code here makes me think that it's not the right place. The pattern typer is being constructed to handle a single clause rather than the entire case expression, so if later this code was upgrades to do full case expressions we would have to remove it from there and place it at some higher level.

I'm not sure where it should go, but I think it should go somewhere that _could_ be the home of full exhaustiveness checking in future, at least or `let` and `try`.

tomwhatmore: The issue is that moving this code to outside the `PatternTyper` (back to where I had it originally in the `infer_let()` for example) is that it won't handle nested constructors properly. I think it makes sense to be checking the exhaustiveness as you infer the Pattern, otherwise you'll just have to go back through it all again afterwards. Putting it back into `infer_let()` would also not help at all with the hypothetical future full exhaustiveness checking, as the let version is really just a single clause version of the generalised case one.

Possibly a solution would be to build up a data structure that represents the coverage state as we recursively infer the pattern elements, and then return that alongside the final `TypedPattern` ast. A separate bit of code could then be called from `infer_let()` and `infer_case()` with that data to actually decide if the coverage state is exhaustive or not. This might even start looking like a bit like the Drop Your Guards algorithm from that paper @QuinnWilton posted in the original exhaustiveness issue #34. Or maybe I just think that because I spent a chunk of this morning trying (and mostly failing) to get my head round said paper.

Not sure that helps us much in the short term though! I'll see if anything clever suddenly occurs to me.

lpil: Let me know when you have an idea of what you would like to do (including continuing with it in the pattern typer) and I'll do a full review. Thanks!

lpil:
Any further thoughts?  :)

tomwhatmore:
Yes, lots. The problem is turning them into working code!

My latest scheme is trying to implement the first, easiest part of the Lower Your Guards (LYG) algorithm from the paper in #34. It's the part where you desugar arbitrarily complex clauses into guard trees, which are pretty much lists of simple clause guards, applied sequentially. This guard tree would be created in the pattern typer as it unified and then returned to the caller, which can decide what to do with this information. This means the pattern typer doesn't need to know anything about what kind of binding is going on, which would be much nicer than the code above. As I say, apparently this is the easy bit of the paper, and also the one SPJ spends most of the time on in his CodeMesh talk, so I'm pretty sure I can implement it. I'm currently trying to convince the borrow checker to agree.

At that point the LYG algorithm transforms the guard tree into an uncovered set and an annotated tree, the first of which allows you to check for exhaustiveness, and the latter for redundant clauses. This portion of the paper is currently a bit much for my feeble CS skills and I don't see myself being able to implement it any time soon (though I'm not completely giving up yet). However, I think I can put together a much simpler version of the uncovered set by simply iterating over the clause guard tree nodes and checking if each one can fail to match with the approach already in this PR.

The end result would be a bit over-engineered for the specific functionality described in #537, but it would build the first step of the full pattern coverage checking and then actually use it, so adding the rest of the coverage checking would be an upgrade rather than a rewrite. It would also address #711 with a small extension:

(Let || Try) && any guard tree node can fail to match -> Error
assert && none of the guard tree nodes can fail to match -> Warning

I think I'll either have a working first version to take a look at or alternatively have admitted defeat by the end of the weekend.

lpil:
Wow, you're really going all out! 😁

Fantastic, thank you. Looking forward to seeing what you come up with no matter which route we end up going down

QuinnWilton:
> At that point the LYG algorithm transforms the guard tree into an uncovered set and an annotated tree, the first of which allows you to check for exhaustiveness, and the latter for redundant clauses. This portion of the paper is currently a bit much for my feeble CS skills and I don't see myself being able to implement it any time soon (though I'm not completely giving up yet). However, I think I can put together a much simpler version of the uncovered set by simply iterating over the clause guard tree nodes and checking if each one can fail to match with the approach already in this PR.

If you're interested in chatting about this at any point, I'm happy to try to work through understanding things together with you in IRC or on a phone call. I've read through the paper, but never got around to trying to implement it because I kept getting distracted by easier things that offered instant gratification.

At the very least, I'm happy to help review whatever you come up with if you're like another pair of eyes on it :)

tomwhatmore:
Thanks @QuinnWilton, let's discuss it on IRC and see if we can get anywhere.

I spent the weekend having a nice time outdoors and didn't do any Gleam work, so don't have as much progress as I had hoped, but there are some developments to report. I have now implemented the first 2 sub-algorithms from LYG for a subset of Gleam (RecordConstructors and Vars only so far!) Specifically:

1. Desugaring clauses to the simple guard tree syntax (fancy G in the paper)
2. Using them to build a refinement type that describes the uncovered set (fancy U in the paper)

So this Gleam code

```rust
pub type T {
    A
    B
    C
}

pub type Maybe(a) {
  Nothing
  Just(a)
};

pub fn main(x: Maybe(T)) {
    case x {
        Just(A) -> 1
    }
}
```

produces

```
Guard Tree: Just b ← a, A ← b, -> 1
Refinement type: ⟨x: Maybe A | a ≉ Just ∨ (Just b ← a ∧ (b ≉ A ∨ (A ← b ∧ (✗))))⟩ 
```

This probably won't mean much unless you've read the paper, but the important bit is that it successfully recreates one of its examples (which is why it's so Haskell-y). This works with both bindings and cases, and with any custom types.

As the paper points out, with some degree of understatement, "this is not very helpful to report to the user". In order to actually tell the user they aren't handling `Nothing`, `Just A`, and `Just C`, we need to generate some "concrete inhabitants" using sub-algorithm fancy G. Fancy G is ominously descibed as doing "the real work" and features a dazzling array of new and inpenetrable symbols. I'm goign to put that off for the moment and make sure the first two bits work with the rest of Gleam first.

lpil:
Hey gang! I'm gunna close this as it has gone a bit stale but please feel free to reopen any time you wish. Thanks!

PR #708, opened 2020-07-01
Bitstring pattern match correctly writes discarded variables.

### LINKED ISSUE #704: Discard in bit string pattern renders invalid Erlang
```rust
case x {
 <<_:utf8, rest:binary>> -> True
  _ -> False
} 
```

The `_` doesn't get rendered in the Erlang.

```rust
case x of
 <</utf8, Rest/binary>> -> true;
  _ -> false
end
```

namoscagnm:
This fixes #704

### DIFF
--- a/src/erl/tests.rs
+++ b/src/erl/tests.rs
@@ -549,7 +549,56 @@ main(Args) ->
     A1.
 "#,
     );
+    // https://github.com/gleam-lang/gleam/issues/704
 
+    assert_erl!(
+        r#"
+pub fn bitstring_discard(x: String) -> Bool {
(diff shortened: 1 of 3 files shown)


src/erl/tests.rs
```diff
@@ -549,7 +549,56 @@ main(Args) ->
...
+
+bitstring_discard(X) ->
+    case X of
+        <<_Discardme/utf8, Rest/binary>> ->
+            true;
+
+        _ ->
+            false
+    end.
+"#,
```
lpil, line 600: I like the tests here! Looks good!

PR #713, opened 2020-07-04
Eunit escript

### LINKED ISSUE #648: `gleam eunit` runs eunit
We have a basic version of this implemented already, further improvements are needed:

- [x] Exit 1 if tests fail
- [ ] Pretty progress output
- [ ] Pretty results output

Will want to implement Erlang code in an escript or similar to achieve this, similar to how the Erlang compiling escript works.

thehabbos007:
Closes #648 

Instead of using the `unite_compact` library, I went with `eunit_formatters`. This is closer to the one used in rebar.
Switching to the other library made license handling a lot easier, and it was contained in a single file :)

Failing tests |  Successful tests
:-------------------------:|:-------------------------:
image  |  image

Library uses Apache v2, just like Gleam. In order to comply with the license of the library, I wrote the changes I made to the code, and included the original license. 
We're not hiding any of the modified code either, so we should be covered. The end user evens gets a full copy with license when they run `gleam eunit` the file is compiled at each execution and not shipped as a binary!

### DIFF
--- a/src/eunit.rs
+++ b/src/eunit.rs
@@ -1,10 +1,18 @@
 use crate::{
     build::{self, project_root::ProjectRoot, Origin},
     error::Error,
-    file,
+    file::{self, OutputFile},
+    //file,
 };
 use itertools::Itertools;
 use std::{path::PathBuf, process::Command};
+
+#[derive(Debug)]
+struct EunitFile {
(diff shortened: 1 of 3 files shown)


lpil:
No idea why the build failed, I've run it again

PR #715, opened 2020-07-08
Typer now returns an error if tuple arity doesn't match

### LINKED ISSUE #714: Compiled erlang tuple has wrong amount of elements
Version:

```sh
❯ gleam --version
gleam 0.10.0-dev
```

(found same behavior w/ latest `0.11.0-dev` built from source)

I ran into an issue where it seems that a tuple in the generated erlang code has an incorrect amount of elements.

Code being referenced is pinned at [this commit](https://github.com/MainShayne233/xenon/commit/85eb827752d99f6a0c3e80067992473a8b06b525).

Here is my function:

```rust
fn do_parse(doc: RawDoc) -> Result(XmlElement, String) {
  case doc {
    tuple(label, _, _, _, _, _, _, _, _, _, _, _) -> parse_xml_element(doc)
    other -> {
      io.debug(other)
      Error("Unhandled node type")
    }
  }
}
```

The generated  erlang in `gen` looks like:

```erlang
do_parse(Doc) ->
    case Doc of
        {Label, _} ->
            parse_xml_element(Doc);

        Other ->
            gleam@io:debug(Other),
            {error, <<"Unhandled node type"/utf8>>}
    end.
```

As you can see, the `{Label, _}` tuple is a 2-element tuple, but the tuple in the gleam code has 12 elements.

tomwhatmore:
Closes #714

The code given in the issue shouldn't actually compile at all! Apparently we weren't checking the arity of tuples in patterns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Gleam can now compile Gleam projects without an external build tool.
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
+- Fixed a bug where the compiler failed to return an error when type checking
+  a tuple with the wrong arity in a pattern.
 
 # v0.10.0 - 2020-07-01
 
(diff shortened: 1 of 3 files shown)


tomwhatmore:
Changelog updated

lpil:
<3

PR #722, opened 2020-07-14
Support updating records using a new spread syntax

QuinnWilton:
For example:

```
import should
import record_update

pub type Person {
  Person(
    name: String,
    age: Int,
    country: String
  )
}

pub fn record_update_test() {
  let past = Person("Quinn", 27, "Canada")
  let present = Person(..past, country: "USA", age: past.age + 1)

  should.equal(present, Person("Quinn", 28, "USA"))

  let module_box = record_update.Box(5)
  let updated = record_update.Box(..module_box, value: 6)

  should.equal(updated, record_update.Box(6))
}
```

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -328,6 +328,27 @@ pub struct CallArg<A> {
     pub value: A,
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct RecordUpdateSpread {
+    pub name: String,
+    pub location: SrcSpan,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct UntypedRecordUpdateArg {
+    pub label: String,
+    pub location: SrcSpan,
+    pub value: UntypedExpr,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct TypedRecordUpdateArg {
+    pub label: String,
+    pub location: SrcSpan,
+    pub value: TypedExpr,
+    pub index: usize,
+}
+
 pub type MultiPattern<PatternConstructor, Type> = Vec<Pattern<PatternConstructor, Type>>;
 
 pub type UntypedMultiPattern = MultiPattern<(), ()>;
--- a/src/error.rs
+++ b/src/error.rs
@@ -1179,8 +1179,24 @@ signed, unsigned, big, little, native, size, unit",
                     )
                     .unwrap();
                 }
-            },
 
+                RecordUpdateInvalidConstructor { location } => {
+                    let diagnostic = Diagnostic {
+                        title: "Invalid record constructor".to_string(),
+                        label: "This is not a record constructor".to_string(),
+                        file: path.to_str().unwrap().to_string(),
+                        src: src.to_string(),
+                        location: location.clone(),
+                    };
+                    write(buffer, diagnostic, Severity::Error);
+
+                    writeln!(
+                        buffer,
+                        "You are attempting to update a record using a value that is not a record constructor",
+                    )
+                    .unwrap();
+                }
+            },
             Error::Parse { path, src, error } => {
                 use lalrpop_util::ParseError::*;
 
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -4,11 +4,12 @@ use std::str::FromStr;
 
 use crate::ast::{
     UntypedExpr, UntypedArg, Arg, UntypedModule, Module, UntypedStatement, Statement, TypeAst,
-    UntypedPattern, BinOp, Clause, UntypedClause, RecordConstructor, Pattern, CallArg,
+    UntypedPattern, BinOp, Clause, UntypedClause, RecordConstructor, Pattern, RecordUpdateSpread, UntypedRecordUpdateArg, CallArg,
     ExternalFnArg, ArgNames, UnqualifiedImport, UntypedClauseGuard, ClauseGuard, BindingKind,
     BitStringSegment, BitStringSegmentOption, Constant, UntypedConstant,
 };
 use crate::parser::*;
+use lalrpop_util::ParseError;
 
 grammar;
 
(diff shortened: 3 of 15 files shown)


src/ast.rs
```diff
@@ -328,6 +328,27 @@ pub struct CallArg<A> {
...
 }
 
+#[derive(Debug, PartialEq, Clone)]
+pub struct RecordUpdateSpread {
+    pub name: String,
+    pub location: SrcSpan,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct UntypedRecordUpdateArg<A> {
```
QuinnWilton, line 338: I can remove these type parameters now, and just inline `UntypedExpr` and `TypedExpr`

src/error.rs
```diff
@@ -1155,8 +1155,25 @@ signed, unsigned, big, little, native, size, unit",
                     )
                     .unwrap();
                 }
-            },
 
+                RecordUpdateInvalidConstructor { location, name } => {
```
QuinnWilton, line 1159: As far as I can tell, there isn't actually any possible way of reaching this error with the current grammar, but I added a new error just in case. If you'd prefer panicking, or you have other ideas, I'm open to those too.

src/grammar.lalrpop
```diff
@@ -561,6 +562,45 @@ ArgNames: ArgNames = {
     <label:VarName> <name:DiscardName> => ArgNames::LabelledDiscard { label, name },
 }
 
+RecordUpdate: UntypedExpr = {
+    // <s:@L> <m:VarName> "#" <n:UpName> <args:RecordUpdateArgs> <e:@L> => UntypedExpr::RecordUpdate {
```
QuinnWilton, line 566: Right now I don't support updating records that were imported from another module. In theory this production rule should allow me to handle them, but it makes the grammar ambiguous. We should be able to rearrange the grammar to make things work, but I'm not sure what the best approach is for doing that.

Do you have any ideas?

src/grammar.lalrpop
```diff
@@ -561,6 +562,45 @@ ArgNames: ArgNames = {
...
+    //     location: location(s, e),
+    //     module: Some(m),
+    //     name: n,
+    //     spread: args.0,
+    //     args: args.1,
+    // },
+
+    <s:@L> "#" <n:UpName> <args:RecordUpdateArgs> <e:@L> => UntypedExpr::RecordUpdate {
```
QuinnWilton, line 574: Similarly, I'd prefer not having the `"#"` in front, but without it, the grammar is ambiguous because we'd need 2 tokens of lookahead to determine whether a spread is present, and disambiguate record updates, record construction, etc.

This is again a case that we should be able to handle, but I wasn't sure what the best way of doing so is.

src/grammar.lalrpop
```diff
@@ -561,6 +562,45 @@ ArgNames: ArgNames = {
...
+}
+
+RecordUpdateArgs: (RecordUpdateSpread, Vec<UntypedRecordUpdateArg<UntypedExpr>>) = {
+    "(" <spread:RecordUpdateSpread> <args:Comma<RecordUpdateArg>> ")" => {
+        (spread, args)
+    }
+}
+
+RecordUpdateSpread: RecordUpdateSpread = {
+    <s:@L> ".." <name:VarName> ","? <e:@L> => RecordUpdateSpread {
```
QuinnWilton, line 590: Are these the best spots for the location tokens? I wonder if this rule would be more accurate:

```
 ".." <s:@L> <name:VarName> <e:@L> ","?
```

src/typ.rs
```diff
@@ -2278,6 +2288,94 @@ impl<'a> Typer<'a> {
         })
     }
 
+    fn infer_record_update(
```
QuinnWilton, line 2291: There's a lot going on in this function, and I'd still like to get some comments describing what I'm doing. It may also be worthwhile to flatten things a little bit, if you have any ideas on doing so nicely :)

QuinnWilton:
Oh, I also need to update the changelog. I'll do that tomorrow / after everything else gets fixed.

src/grammar.lalrpop
tomwhatmore: I tried a bit of inlining trickery to see if I could get this working but unfortunately got nowhere. Not very helpful, sorry!

lpil: Something like this perhaps? https://github.com/gleam-lang/gleam/commit/2d63872cc988b9e007638cbde495fe93a5b91535

QuinnWilton: I got it working with a variant of this, thank you!

QuinnWilton:
Just an update on this:
- The syntax I wanted is working now
- Record updates work on imported record types

The last bit, which I'm fiddling with now, is to get type inference working when dealing with polymorphic types. For example, I'm currently unable to detect that this should fail type checking:

```
pub type Box(a) {
    Box(value: a)
};

pub fn update_box(box: Box(a), value: b) {
    Box(..box, value: value)
};
```

The issue is that I'm not sure how to access the type of `box.value` here. My current version checks against the type of `value` in `Box`, which is a generic, and so unifies without error. Here though, I need to unify with the instantiated type for `value` that `box.value` inhabits. The best way I can think of for doing that is:

1) For each argument to the record update, look up the field with that name in the record constructor's field map
2) Look up the associated argument in the record constructor's argument list
3) If it is a `Type::App`, we are done, and can unify against that
4) Otherwise, it is a generic variable, with some ID
5) Iterate through the return type for the record constructor to find the index of the argument with the same ID
6) Grab the argument of the spread with the same index
7) Instantiate that argument
8) Unify the resulting type with the argument from the record update

This sounds really messy and error-prone to me, and I feel like there has to be a better way of doing this that I'm missing. I also haven't finished implementing it yet, so I'm not sure if it's correct. Do you or @tomwhatmore have a better idea?

I'll keep poking away at that tonight, but may have questions tomorrow about how to unify things properly.

QuinnWilton:
Alright, I just pushed up a version that fully works, as far as I can tell. I think this is ready for review now :)

src/error.rs
```diff
@@ -1155,8 +1155,24 @@ signed, unsigned, big, little, native, size, unit",
...
+                        label: "This is not a record constructor".to_string(),
+                        file: path.to_str().unwrap().to_string(),
+                        src: src.to_string(),
+                        location: location.clone(),
+                    };
+                    write(buffer, diagnostic, Severity::Error);
+
+                    writeln!(
+                        buffer,
+                        "You are attempting to update a record using a value that is not a record constructor",
```
QuinnWilton, line 1171: I'm not a fan of this error message. Do you have any suggestions? This error is returned if something else is used with the record update syntax, like a normal function call, or a variable.

src/typ.rs
```diff
@@ -2290,6 +2310,96 @@ impl<'a> Typer<'a> {
...
+        &mut self,
+        constructor: UntypedExpr,
+        spread: RecordUpdateSpread,
+        args: Vec<UntypedRecordUpdateArg>,
+        location: SrcSpan,
+    ) -> Result<TypedExpr, Error> {
+        let constructor = self.infer(constructor)?;
+
+        let (module, name) = match constructor {
```
QuinnWilton, line 2322: This block here could in theory be extracted out to a helper. It's very similar to `get_field_map`

src/grammar.lalrpop
QuinnWilton: The optional comma here is a little tricky, because it allows code like this:

```
let p = Person(..person age: 20)
```

I have an idea for resolving that though.

lpil:
```rust
pub type Box(a) {
    Box(value: a)
};

pub fn update_box(box: Box(a), value: b) {
    Box(..box, value: value)
};
```

Why is it that this should fail? Wouldn't it return `Box(b)`?

QuinnWilton:
> ```rust
> pub type Box(a) {
>     Box(value: a)
> };
> 
> pub fn update_box(box: Box(a), value: b) {
>     Box(..box, value: value)
> };
> ```
> 
> Why is it that this should fail? Wouldn't it return `Box(b)`?

Hm that's a good point. I think there's a decision to be made there. I was thinking that a record update always returns the same type you started with, so that update should return a `Box(a)`, and require that `value` is an `a`.

This case is really simple, but if there were two fields parameterized by the same variable then we would need to ensure we change both fields in the same update.

Which do you think is more intuitive / useful?

lpil:
I think we want to be able to change the parameterised type. For example, we may have a my_record_based_container.map function, and being able to use the update syntax when implementing this would be very useful.

Why is it that both fields would need to be given? This code seems sensible:

```
pub type Box(a, b) {
    Box(value: a, second: b)
};

pub fn update_box(box: Box(a, b), value: c) -> Box(c, b) {
    Box(..box, value: value)
};
```

lpil:
I think we want to be able to change the parameterised type. For example, we may have a my_record_based_container.map function, and being able to use the update syntax when implementing this would be very useful.

Why is it that both fields would need to be given? This code seems sensible:

```
pub type Box(a, b) {
    Box(value: a, second: b)
};

pub fn update_box(box: Box(a, b), value: c) -> Box(c, b) {
    Box(..box, value: value)
};
```

QuinnWilton:
> I think we want to be able to change the parameterised type. For example, we may have a my_record_based_container.map function, and being able to use the update syntax when implementing this would be very useful.
> 
> Why is it that both fields would need to be given? This code seems sensible:
> 
> ```rust
> pub type Box(a, b) {
>     Box(value: a, second: b)
> };
> 
> pub fn update_box(box: Box(a, b), value: c) -> Box(c, b) {
>     Box(..box, value: value)
> };
> ```
> 
> The alternative of having to pull out the other values to insert them again seems awkward.

We chatted about it in IRC, but I just wanted to echo the conversation here to preserve it for the future.

The above sample is actually fine, and what I was talking about is more like this:

```
pub type Box(a) {
    Box(left: a, right: a)
};

pub fn update_box(box: Box(a), value: b) -> Box(b) {
    Box(..box, left: value) // Needs to be Box(..box, left: value, right: value)
};
```

Here, the code should not typecheck, because `left` and `right` both need to be the same type, but the resulting value ends up as `Box(left: b, right: a)`. This is likely possible to detect, but the code as it's written now is not smart enough to handle this case.

I agree with you that this would be a useful improve to make. That being said, do you think that the feature as it exists in this PR is useful enough on it's own, to warrant support the additional behaviour as an extension later? Since that extension would be a weakening of the type system, I don't think it would end up being a backwards incompatible change, as long as you're happy with the syntax I have now.

If not, I'm happy to continue thinking about the problem, to see whether I can come up with a better approach!

QuinnWilton:
I just added some tests for updating records on types with multiple variants, and those aren't quite working right. I'll have to investigate after work today.

QuinnWilton:
I opened #735 after debugging the problem mentioned in my last comment.

This is the test I would like to run:

```gleam
    assert_infer!(
        "
        pub type Either(a, b) {
            Left(value: a)
            Right(value: b)
        };

        pub fn update_left(e: Either(a, b), value: a) {
            case e {
                Left(_) as l -> Left(..l, value: value)
                _ -> e
            }
        };",
        vec![
            ("Either", "fn(a, b) -> Either(a, b)"),
            ("update_left", "fn(Either(a, b), a) -> Either(a, b)")
        ]
    );
```

But it fails with the following error:

```
thread 'typ::tests::infer_module_test' panicked at 'should successfully infer: UnknownField { location: SrcSpan { start: 219, end: 231 }, typ: App { public: true, module: [], name: "Either", args: [Var { typ: RefCell { value: Generic { id: 10 } } }, Var { typ: RefCell { value: Generic { id: 11 } } }] }, label: "value", fields: [] }', src/typ/tests.rs:1971:5
```

The issue is that `l` is an `Either(a, b)`, and so doesn't have any accessors. I don't think the current type system allows any way around this, so I'm not sure if this is something I should be handling in this PR or not.

QuinnWilton:
@lpil I've rebased this against `main` and fixed the conflicts. It should be good for review now :)

PR #736, opened 2020-07-17
Issue 729

### LINKED ISSUE #729: Only show labels that have not been already supplied in the error message for `typ::Error::UnknownLabel`
We could add a `supplied: Vec<String>` field to the `UnknownLabel` error and then use that to remove already supplied labels.

tomwhatmore:
Closes #729

I was originally going to do the "simple version", track the supplied labels and leave the everything else the same. However, it occurred to me that it would be annoying to remove labels supplied by the user before the unknown one, but continue suggesting any they supplied after. That meant continuing through the whole vec even if it finds an unknown label, which also raised the possibility of then finding additional unknown labels.

To deal with all this I've ended up refactoring `UnknownLabel` to be `UnknownLabels`, and it takes a vec of unknown labels with their locations, a vec of supplied valid labels, and the vec of all valid options. These all get printed out with the `MultiLineDiagnostic` I conveniently added last week. Like this:
 
Screenshot 2020-07-17 at 19 02 31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
 - Field access now works before the custom type is defined.
+- The error message returned by the compiler when the user tries to use unknown
+  labelled arguments now handles multiple labels at once, and does not suggest
+  labels they have already supplied.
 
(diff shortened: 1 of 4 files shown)


tomwhatmore:
Thanks, top tip! Should be sorted now.

lpil:
Thanks!

PR #743, opened 2020-08-05
Partially fix #611 with Cow

### LINKED ISSUE #611: Support &str in the printer algebra
Currently the `Document` enum structure can only contain Strings, so any time we have a &str we need to allocate a String and copy the &str into it.

Add a new enum variant for Document which can hold a &str, allowing us to remove lots of cloning from the Erlang printer and Gleam formatter.

This will require adding an explicit lifetime to the Document enum.

scileo:
This PR attempts to eliminate some useless copies. It introduced a lot of lifetime constraints which I tried to resolve in the most conservative way. In a lot of situation, this resulted in heavy use of `Cow`, which introduced type errors. That's why this pull request is very big. I was unfortunately unable to split it into smaller PR.

Feel free to post a comment if you see a situation in which a modification was not necessary.

The `butcher` crate has been used to remove most of the duplicate introduced in [a previous attempt][pr_640] to fix this issue

This partially fix #611. There is still some work to do, such as removing as much calls to `unbutcher` as possible.

[pr_640]: https://github.com/gleam-lang/gleam/pull/640


scileo:
I'm a bit worried by the fact that CI fails for MacOS, but not for other targets. Also the provided error seems to be `lalrpop`-related, but the PR does not change anything related to parsing.

What am I missing?

thehabbos007:
> I'm a bit worried by the fact that CI fails for MacOS, but not for other targets.

This has happened to me before too on the MacOS CI. It was fixed when Louis reran the actions. Looks like an open issue on GitHub's end :)

scileo:
> > I'm a bit worried by the fact that CI fails for MacOS, but not for other targets.
> 
> This has happened to me before too on the MacOS CI. It was fixed when Louis reran the actions. Looks like an open issue on GitHub's end :)

Thanks for your comment @thehabbos007! Fixing merge conflicts triggered CI, which successfully pass :)

scileo:
Hi,

To be honest, I share your opinion in the fact that this PR has way to much impact in the codebase. This is not *directly* caused by the `Butcher` derive macro. The goal of this macro is to remove the duplication you pointed out [in my previous attempt][last-issue-comment] to fix issue #611. For example, the big match expression in the `expr` function contained a lot of duplicated patterns [in my previous attempt][fix-611-1-match] which are removed thanks to the `Butcher` derive macro [in this attempt][fix-611-2-match].

Anyway, I do think that other developers are not wiling to learn the butcher crate works.

After some reflection, I am pretty sure I misunderstood what should be done to address such issue, and that I went way too deep in the codebase to fix it (hence the huge amount of changes). As such, I'll come back in a few days in the issue thread to ask more questions about what should be done.

This issue was opened a long time ago, and there I still have not properly fixed it. I do thank you for your infinite patience here.

[last-issue-comment]: https://github.com/gleam-lang/gleam/pull/640#pullrequestreview-426139235
[fix-611-1-match]: https://github.com/gleam-lang/gleam/pull/640/files#diff-b4aa8175cec75ec8465b3cd33ee70d10R1417
[fix-611-2-match]: https://github.com/gleam-lang/gleam/pull/743/files#diff-b4aa8175cec75ec8465b3cd33ee70d10L1042

lpil:
> This issue was opened a long time ago, and there I still have not properly fixed it. I do thank you for your infinite patience here.

Please do not feel bad in any way!

I don't expect any particular contribution to Gleam from anyone and I do not think I am entitled to anyone's time. I am thankful for your contributions (even if we haven't figured this specific problem out yet) and I have learnt quite a few things from reading your Rust code :) 

Work on whatever is fun, that is the most important thing for open source work in my opinion :)

PR #764, opened 2020-08-17
fix conflict in variable name generation

### LINKED ISSUE #761: Conflict in variable name generation
I haven't been able to find a minimal example, but this block [here](https://github.com/QuinnWilton/gleam-chip8/blob/master/src/chip8/emulator.gleam#L368) is generating the following Erlang:

```erlang
            X1 = chip8@registers:read(erlang:element(3, Emulator), Vx23),
            X0 = X1 div 100,
            X1 = (X1 rem 100) div 10,
            X2 = X1 rem 10,
```

Since `X1` gets reused, I get a match error at runtime. I've temporarily worked around this by changing the first variable to `n` instead.

mrgleam:
Close #761 

append "@" when local variable name duplicate.

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -77,7 +77,7 @@ impl<'a> Env<'a> {
                 name.to_camel_case().to_doc()
             }
             Some(0) => name.to_camel_case().to_doc(),
-            Some(n) => name.to_camel_case().to_doc().append(*n),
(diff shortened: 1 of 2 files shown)


QuinnWilton:
Thank you! I was needing to work around this in my application, so this fix is much appreciated :)

PR #771, opened 2020-08-26
Bit strings now allow string variables as segment values

### LINKED ISSUE #770: Argument error with binaries in bit strings
I'm getting an argument error with the following code.

```rust
pub fn demo_test() {
  let x = "x"
  let y = <<x:utf8>>
  todo
}
```

I'm working on the assumption that all runtime errors  count as a bug?

tomwhatmore:
Closes #770 

The printer is now outputting a `binary` type specifier instead of `utf8` when the segment value is a string variable rather than a literal. Fixes the reported issue, let me know if you can think of any edge cases that seem worth explicitly having a test for.

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -321,7 +321,7 @@ fn const_segment(
         _ => None,
     };
 
-    bit_string_segment(document, options, size, unit, false, env)
+    bit_string_segment(document, options, size, unit, true, env)
 }
 
 fn expr_segment(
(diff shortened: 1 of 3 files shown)


test/core_language/test/bit_string_test.gleam
```diff
@@ -74,3 +74,11 @@ pub fn codepoint_conversion_test() {
 
   should.equal(snake_int, 128013)
 }
+
+pub fn string_variable_test() {
+  let x = "x"
+  let y = <<x:utf8, "ß↑e̊":utf8>>
+  let <<z:8, _:binary>> = y
+
+  should.equal(z, 120) // x is unicode codepoint 120
```
lpil, line 83: How about this sort of thing?

```rust
let x = tuple("ok")
<<x.0, "etc":utf8>>
```

```rust
let x = MyRecord(value: "ok")
<<x.value, "etc":utf8>>
```
```rust
let f = fn() { "ok" }
<<f(), "etc":utf8>>
```

src/erl.rs
```diff
@@ -321,23 +321,30 @@ fn const_segment(
...
-        | TypedExpr::Float { .. }
-        | TypedExpr::Var { .. }
-        | TypedExpr::BitString { .. } => expr(value, env),
+        TypedExpr::Int { .. } | TypedExpr::Float { .. } | TypedExpr::BitString { .. } => {
+            expr(value, env)
+        }
+
+        // Return value as normal, but set flag
+        TypedExpr::Var { .. } => {
+            value_is_a_variable = true;
```
lpil, line 345: I think we may be interested in whether it is a literal rather than when it a variable as I think the same problem may occur for function calls, tuple indexing and field accessing.

lpil (commented):
Look good! I've left a note inline about non-variable cases

src/erl.rs
tomwhatmore: Oh yeah, of course. Will do.

lpil: Thank you

tomwhatmore:
Updated it to support those other cases

lpil (approved):
Sweet! Thank you!

PR #782, opened 2020-09-04
Refactor CI/Release workflow

isacsund:
This PR includes a few notable changes to the GitHub Actions workflow.

Releases will now be created automatically whenever new version tags are pushed. I didn't implement the check to see if the binary works after the artifacts have been uploaded as I think it's too late to find out if they work or not when they are already uploaded. Hopefully this change will make the failing OS X compression disappear, otherwise I'll have to take a deeper look at the root cause.

The `build` jobs are now unified as one job with a matrix. Most steps are common to all 3 platforms so we don't need to have 3 separate jobs anymore. This change will make it easier to extend the build process and avoid code duplication.
_Note_: I have only added testing on Rust stable for now. My idea was to extend this to Rust nightly and possibly more platform architectures in the future, depending on the need.

I'll mark this as draft as I still have not done any testing.


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,106 @@
...
+      uses: actions/cache@v2
+      with:
+        path: target
+        key: ${{ runner.os }}-cargo-test-${{ matrix.toolchain }}-${{ hashFiles('**/Cargo.lock') }}
+
+    - name: Build binary
+      run: cargo build --release
+      env:
+        CARGO_INCREMENTAL: 0
+        RUSTFLAGS: "-C debuginfo=0 -D warnings"
```
lpil, line 43: What do these environment variables do?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - 'v*'
+jobs:
+  create-release:
+    name: create-release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create artifacts directory
+        run: mkdir artifacts
+
+      - name: Create GitHub release
```
lpil, line 14: How do we create pre-release versions with this workflow?

lpil:
This looks great!

I've left a couple questions inline.

Is there a repo where I can see the release workflow having run?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          release_upload_url="$(cat artifacts/release-upload-url)"
+          echo "::set-env name=RELEASE_UPLOAD_URL::$release_upload_url"
+          echo "release upload url: $RELEASE_UPLOAD_URL"
+          release_version="$(cat artifacts/release-version)"
+          echo "::set-env name=RELEASE_VERSION::$release_version"
+          echo "release version: $RELEASE_VERSION"
+
+      - name: Build release binary
+        run: cargo build --release
```
lpil, line 78: On Linux we need to use the Dockerfile to build otherwise the resulting binary will be dynamically linked

isacsund:
Thanks for the comments! I've had a busy weekend but I will try to look at this tomorrow or on Wednesday!

lpil:
Fab, thank you.

No need to rush :)

.github/workflows/ci.yaml
isacsund: I've removed these. My idea was to reduce the disk space used on the runner but now when I think about it I doubt we will be using that much space.

.github/workflows/release.yaml
isacsund: I'm gonna experiment with this a bit. Wouldn't this be the same as building with [MUSL support](https://doc.rust-lang.org/edition-guide/rust-2018/platform-and-target-support/musl-support-for-fully-static-binaries.html) (please correct me if I'm wrong here)?

isacsund:
Sorry for the lack of updates, I've been quite busy lately.
I've made a repository [here](https://github.com/isacsund/release-test) to test the release workflow. To address your previous question about pre-releases, you are still able to edit the release (e.g. tick the pre-release checkbox) after the action has created the release. 
I did add `draft: true` to the release action so you can have a final look at the release before you mark it as non-draft. I don't know if this is something you prefer to have or not so let me know what you think.

I believe `cross` is the way to go for building statically linked binaries, however, `ekidd/rust-musl-builder:stable` didn't work inside GitHub actions so I decided to use `clux/muslrust` instead.

.github/workflows/release.yaml
lpil: I'm not sure, I didn't dig much into how the image worked

.github/workflows/release.yaml
```diff
@@ -0,0 +1,87 @@
...
+    steps:
+      - name: Create GitHub release
+        id: release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          draft: true
```
lpil, line 21: Can we have this be based on whether the tag is a pre-release? I think we can have a step that introspects the tag and then sets the variable depending on the format.

If I don't want to have to edit the job definition as well as create a tag, it should be one or the other I think, and I would prefer just to create a tag.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,87 @@
...
+            cp "target/${{ matrix.target }}/release/gleam.exe" "$staging/"
+            7z a "$staging.zip" "$staging"
+            echo "::set-env name=ASSET::$staging.zip"
+          else
+            cp "target/${{ matrix.target }}/release/gleam" "$staging/"
+            tar -czvf "$staging.tar.gz" "$staging"
+            echo "::set-env name=ASSET::$staging.tar.gz"
+          fi
+
+      - name: Upload release archive
```
lpil, line 79: This looks great! Are these binaries somewhere so I can test before we merge?

lpil (commented):
Really liking this! I've left some inline comments

.github/workflows/release.yaml
isacsund: Hmm... I think I explained this poorly. What I meant was that every release will be marked as draft (only those with write permissions will see the release). To make the release public you'll just have to go to _Releases_ and edit the newly created release and click on _publish_.

There's no job definition to edit here, it still works by just creating the tag. The only difference from how you currently do releases is that now your release will be already created with all the assets, so you can have a look at the built binaries before you publish the release.

.github/workflows/release.yaml
isacsund: Yeah, I created a new tag and published the [release](https://github.com/isacsund/gleam/releases) in my fork.

isacsund:
I'll go ahead and squash/cleanup the commits tomorrow.

isacsund:
Sorry for the delay once again... I cleaned up the commits and I believe it's now good to go.
If you want to try the release workflow yourself before merging I can give you access to my fork so you can test it there.

.github/workflows/release.yaml
lpil: Oh that's really nice, thank you!

lpil (approved):
Brilliant work, thank you so much!

PR #783, opened 2020-09-07
Disallow the utf type specifiers in bit string matching

### LINKED ISSUE #744: uft8 and uft8_codepoint both match as a utf8_codepoint type
I might be wrong in my understanding of the expected behaviour but I believe the following is a bug.

I think this code should compile
```rust
  let <<"ab":utf8, rest:utf8>> = "abcd"
  should.equal("cd", rest)
```

it however fails because the type of rest is `UtfCodepoint`.
My assumption is if the match was to say `rest:utf8_codpoint` explicitly then the type of rest should be UtfCodepoint and this would be the correct compilaton failure. However I think that `rest:utf8` should return rest with a type of string.

tomwhatmore:
Closes #744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Gleam can now compile Gleam projects without an external build tool.
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
+- The utf8, utf16, and utf32 type specifiers are now only available in bit
+  string construction, matching must be done with the codepoint versions.
 
 ## v0.11.2 - 2020-09-01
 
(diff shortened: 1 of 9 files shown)


tomwhatmore:
Not sure why the build is failing on mac os, it just says `can't find crate for 'serde_derive' which 'lalrpop' depends on`. Any ideas @lpil ?

lpil (approved):
Thank you.

I don't understand what's happening with the OSX build but it has started doing this a lot lately. I found a similar looking issue in the Rust tracker and it suggested there was a bug in BSD tar of all things!

tomwhatmore:
That's annoying! Thanks for the info, I have to admit that bug in BSD tar would have been quite far down my list of guesses.

lpil:
I find it baffling. I'm interested in seeing whether #782 helps

PR #784, opened 2020-09-08
Define the type of the closure

MaSven:
Rust could not infere the type of the closure here. So we need to define
it otherwise rust could not compile it.


lpil:
This shouldn't be required, what version of Rust are you using? Is it a recent stable?

MaSven:
I'm running 1.44.1. You can see the error here [Build](https://build.opensuse.org/package/live_build_log/home:MaSven/gleam/openSUSE_Tumbleweed/x86_64). I'm trying to provide rpm packages for opensuse TW.

lpil (approved):
OK thank you. :)

Issue #786, opened 2020-09-11
discarded variables cannot contain numbers


CrowdHailer:
```rust
let a1 = 1
let _a = 1
let _a1 = 1
```

The first two are valid variable names, the last one is not. I think it should be

```
error: Syntax error
   ┌─ ./test/plum_mail/authentication_test.gleam:68:9
   │
68 │   let _a1 = 1
   │         ^ Unexpected token

```

lpil:
Yes you're right, thank you

PR #787, opened 2020-09-11
change regex for discard variable to include numbers

CrowdHailer:
This works to address #786 

I have yet to work out where tests should live,

### DIFF
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -1094,7 +1094,7 @@ UpName: String = {
 }
 
 DiscardName: String = {
-    <v:r"_[A-Za-z_]*"> => v.to_string()
+    <v:r"_[a-z][0-9a-z_]*"> => v.to_string()
 }
 
 AnyName: String = {
--- a/src/grammar.lalrpop
+++ b/src/grammar.lalrpop
@@ -1094,7 +1094,7 @@ UpName: String = {
 }
 
 DiscardName: String = {
-    <v:r"_[a-z][0-9a-z_]*"> => v.to_string()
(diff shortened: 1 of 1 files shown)


CrowdHailer:
Hmm, I can't work out why the changes I have made would result in the test failures I'm seeing. I will have to park this for now.

src/grammar.lalrpop
```diff
@@ -1094,7 +1094,7 @@ UpName: String = {
 }
 
 DiscardName: String = {
-    <v:r"_[A-Za-z_]*"> => v.to_string()
+    <v:r"_[a-z][0-9a-z_]*"> => v.to_string()
```
michallepicki, line 1097: Does this regex work for `_` alone not followed by anything? Maybe it should be this way or something like that:
```suggestion
    <v:r"_[a-z]*[0-9a-z_]*"> => v.to_string()
```

lpil: I think that one would allow `_0`, which I think we want to avoid.

Maybe this?

```
<v:r"_([a-z][0-9a-z_]*)*"> => v.to_string()
```

Tested here -> https://rustexp.lpil.uk/

CrowdHailer: I thought `_([a-z][0-9a-z_]*)?` might be simpler.

Neither seems to work with a discard pattern of just `_` (on your rustexp page)

michallepicki:
The failing test has a capitalized letter in the binding name (`_Discardme`)

src/grammar.lalrpop
lpil: Ah yes! That's much better

lpil:
I'll merge and fix the test, thank you!

lpil (approved):

PR #793, opened 2020-09-30
fix issue 788 : make the discard patterns always render _

### LINKED ISSUE #788: Discarded variables don't increment in ouputed erlang
```rust
let _r = 1
let _r = 2
```

This code runs with this failure.

`** (MatchError) no match of right hand side value: 1`

I think  it's making erlang code that looks like the following.
```erl
_R = 1
_R = 2
```

mrgleam:
Close #788 

make the discard patterns always render _


lpil (approved):
Thank you! Great stuff

PR #800, opened 2020-10-11
add `gleam init` command

Takashiidobe:
I basically copy pasted the `gleam new` command to make it work for `gleam init`. 

This command works fine if you pass it a valid directory, however it doesn't work for the current path, `.` or `./`, although i'll need to add tests to this PR.

I'm not sure of a good way to get the current working directory to pass it back to gleam init, and would like some help with that if possible.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -108,6 +108,23 @@ enum Command {
     #[structopt(name = "docs", about = "Render HTML documentation for a project")]
     Docs(Docs),
 
+    #[structopt(name = "init", about = "Create a new project in an existing directory")]
+    Init {
+        #[structopt(long = "description", help = "description of the project")]
+        description: Option<String>,
+
+        #[structopt(help = "location of the project root")]
+        project_root: Option<String>,
+
+        #[structopt(
+            long = "template",
+            possible_values = &new::Template::VARIANTS,
+            case_insensitive = true,
+            default_value = "lib"
+        )]
+        template: new::Template,
+    },
+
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -108,6 +108,23 @@ enum Command {
     #[structopt(name = "docs", about = "Render HTML documentation for a project")]
     Docs(Docs),
 
+    #[structopt(name = "init", about = "Create a new project in an existing directory")]
+    Init {
```
lpil, line 112: I think we'll also want to optionally take the name of the project because we write this `gleam.toml` and the user may want not to use the name of the directory.

```toml
name = "the_project_name"
```

src/new.rs
```diff
@@ -13,6 +12,90 @@ pub enum Template {
     App,
 }
 
+pub fn init(
```
lpil, line 15: We could remove the duplication by having the `new` function call the `init` function

```rust
init(
  template,
  description,
  name,
  path,
  version,
)
```

lpil (commented):
A good start, thank you. I've left some inline comments.

Currently the `new` command only has manual testing, I would be happy for this to also be the case for `init`. Alternatively we can add a new step to the CI test workflow that creates a new project with `init` and compiles it with `rebar3 eunit`.

lpil:
Hello! I'm gunna close this as it has gone a bit stale but please feel free to reopen any time you wish. Thanks!

PR #829, opened 2020-11-03
Discriminate between record access and module access for shadowed names

### LINKED ISSUE #807: Help compiler discriminate between module.function and variable.field when variable name shadows module name
```gleam
import power.{Power}

pub fn do_something(power: Power) -> Int {
  power.to_int(power)
}
```
`gleam build` gives the following error message:
```
error: Unknown field
  ┌─ /path/whatever.gleam:4:8
  │
4 │   power.to_int(power)
  │        ^^^^^^^ This field does not exist
```

@lpil commented on Discord: 

> In this case the variable is shadowing the module. This is probably one of the most frequent issues in Gleam
>
> In this case the compiler has all the information needed to be able to tell that power means the module (it knows it can't be the variable because it has type Power, which doesn't have that field.
> 
> ```
> if type_is_known && !does_not_have_field {
>   check_if_module_access()
> } else {
>   type_error()
> }
> ```
> We do something like this with the pipe operator already.

QuinnWilton:
Closes #807

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Gleam can now compile Gleam projects without an external build tool.
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
+- The compiler can now discriminate between record access and module access
+  for shadowed names
 
 ## v0.12.0 - 2020-10-31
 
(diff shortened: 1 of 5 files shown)


lpil (approved):
Very clean! Thank you! This will avoid some annoyance in future.

PR #830, opened 2020-11-03
Bring back .content max-width in rendered HTML documentation

### LINKED ISSUE #826: Rendered HTML documentation is now too wide and difficult to read
It seems we lost a max-width so any text is rendered extremely wide, making it difficult to read.

image

Here is the commit that introduced the problem: 

https://github.com/gleam-lang/gleam/commit/b36aa93b83474d112fda5adcb2c34eeb6bfbc284#diff-fd95152bde957f33fe95e62d56c9bdf1af8c94300949c525bf5d442416cdb666L24

michallepicki:
Before:
image

After:
image

Closes #826

### DIFF
--- a/templates/index.css
+++ b/templates/index.css
@@ -21,6 +21,7 @@
   --accent: var(--pink);
 
   /* Sizes */
+  --content-width: 680px;
   --header-height: 60px;
   --sidebar-width: 240px;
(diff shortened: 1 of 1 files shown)


lpil (approved):
Fab, thank you!

Issue #180, opened 2019-05-17
Implement language server protocol


lpil:
To enable IDE like features in all editors.

gamebox:
I will take this on after #157 - which I'm happy to help with.

gamebox:
You can assign this to me as well

lpil:
I'm afraid as this is a core feature I need to maintain I'm going to be taking the lead on this one so that I have full understanding of how it works. Not knowing how part of the compiler works puts me in an tricky position for future work and making informed decisions about it's development.

gamebox:
Oh, ok. 👍

lpil:
Having said that it would be good to see any work you have done. I'm sure we can use it so long as I'm able to get a good understanding of how LSP and our implementation works.

lpil:
https://rust-analyzer.github.io/blog/2020/07/20/three-architectures-for-responsive-ide.html

adituv:
I'm looking at making a start on implementing this.  My current plan is to use the [lsp-server](https://github.com/rust-analyzer/lsp-server) crate as a base, then build on top of that.

The rust-analyzer tool, which builds on top of lsp-server, mentions a [major performance degradation](https://github.com/rust-analyzer/rust-analyzer/blob/28b1cd32b3682f172d9caa862dbf1bcdbf53234b/crates/rust-analyzer/src/main_loop.rs#L31) on Windows, so we may want to implement a similar workaround.

lpil:
That's a very interesting looking crate! Thanks for the recommendation.

I'd like to suggest implementing formatting first in the LSP as that's probably the simplest thing we can do in the compiler at present. It's important that it's clear how to work on and maintain this feature (for me especially) so it would be good to do this in small incremental pieces if possible.

natdm:
This is something I'd love to help with when the first iteration is done. What's the status?

lpil:
Hello! No work has been done at present. I am interested in implementing this using just an event loop, IO, and Rust's serde JSON rather than a generic library (at least to start).

I think that implementing formatting via LSP could be a good place to start.

natdm:
Another thing to consider is treesitter support, further on down the road. Will give you tons of functionality and syntax highlighting just by having the grammar and queries (neovim only).

oblitum:
[...]

PR #832, opened 2020-11-06
Add language server support (#180)

adituv:
This PR adds the `language-server` command to the gleam tool (see issue #180)

The language server itself only has very minimal functionality but should be a good basis for future work.  However, I will not be continuing this code for a while, so I am submitting it here so that others can contribute.  It perhaps should have a separate branch while more functionality is added.

Currently, the language server supports:
* Basic functionality: initialization, shutdown, exit
* Text synchronisation notifications: didOpen, didChange, didClose
* Document formatting

At a minimum, I believe it should also have these features before being merged to main:
* Parse errors returned when applicable on didChange and formatting (publishDiagnostics)
* Syntax highlighting (documentColor)

Beyond this, it needs more testing.  This probably includes live integration testing via a language client in an editor such as VS Code, as well as further unit testing.

There is one known bug with the underlying library that could affect this in future - [tower-lsp#239](https://github.com/ebkalderon/tower-lsp/issues/239).  However, as the current implementation only supports standard input, this is bypassed by closing the input stream.

### DIFF
--- a/src/language_server.rs
+++ b/src/language_server.rs
@@ -1,77 +1,62 @@
 pub(crate) mod format;
 
-use crate::error::Error::LSPError;
-use crate::language_server::format::format_doc;
+use crate::language_server::format::format;
 
-use lsp_types::{
-    request::Formatting, InitializeParams, ServerCapabilities,
-};
+use tokio::runtime::Runtime;
 
-use lsp_server::{Connection, Message, Request, RequestId, Response};
+use tower_lsp::jsonrpc::{ Error, ErrorCode, Result };
+use tower_lsp::lsp_types::*;
+use tower_lsp::{Client, LanguageServer, LspService, Server};
 
-#[allow(unused_variables)]
-pub fn command(project_root: String) -> Result<(), crate::error::Error> {
-    match run_language_server() {
-        Ok(()) => Ok(()),
-        Err(e) => Err(LSPError(e.to_string())),
+#[derive(Debug)]
+struct ServerBackend {
+    client: Client,
+}
+
+#[tower_lsp::async_trait]
+impl LanguageServer for ServerBackend {
+    async fn initialize(&self, _params: InitializeParams) -> Result<InitializeResult> {
+        let mut result = InitializeResult::default();
+        result.capabilities.document_formatting_provider = Some(true);
+        Ok(result)
     }
-}
+    async fn formatting(&self, params: DocumentFormattingParams) -> Result<Option<Vec<TextEdit>>> {
+        let doc_uri = params.text_document.uri;
 
-fn run_language_server() -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
-    tracing::info!("Starting language server");
+        if(doc_uri.scheme() != "file") {
+            return Err(Error::new(ErrorCode::InvalidParams));
+        }
 
-    let (connection, io_threads) = Connection::stdio();
+        let doc_contents = match std::fs::read_to_string(doc_uri.path()) {
+            Ok(contents) => contents,
+            Err(io_error) => return Err(Error { code: ErrorCode::InternalError, message: io_error.to_string(), data: None }),
+        };
 
-    let capabilities = serde_json::to_value(&ServerCapabilities::default()).unwrap();
-    let init_params_json = connection.initialize(capabilities)?;
-    let initialization_params = serde_json::from_value(init_params_json)?;
-
-    lsp_loop(&connection, &initialization_params)?;
-
-    io_threads.join()?;
-
-    tracing::info!("Shutting down language server");
-
-    Ok(())
-}
-
-#[allow(unused_variables)]
-fn lsp_loop(connection: &Connection, params: &InitializeParams) -> Result<(), Box<dyn std::error::Error + Sync + Send>> {
-    for message in &connection.receiver {
-        match message {
-            Message::Request(request) => {
-                tracing::info!("Received request: {:?}", request);
-                if connection.handle_shutdown(&request)? {
-                    return Ok(())
-                }
-
-                match cast::<Formatting>(request) {
-                    Ok((id, params)) => {
-                        let result = format_doc(params.text_document, params.options);
-                        let result = serde_json::to_value(&result).unwrap();
-                        let response = Response { id, result: Some(result), error: None };
-                        connection.sender.send(Message::Response(response))?;
-                        continue;
-                    }
-                    Err(req) => req,
-                };
-            }
-            Message::Response(response) => {
-                tracing::info!("Received response: {:?}", response);
-            }
-            Message::Notification(notification) => {
-                tracing::info!("Received notification: {:?}", notification);
-            }
+        match format(doc_contents) {
+            Ok(x) => Ok(Some(x)),
+            Err(s) => Err(Error { code: ErrorCode::ParseError, message: s, data: None }),
         }
     }
-
-    Ok(())
+    async fn shutdown(&self) -> Result<()> {
+        Ok(())
+    }
 }
 
-fn cast<R>(req: Request) -> Result<(RequestId, R::Params), Request>
-where
-    R: lsp_types::request::Request,
-    R::Params: serde::de::DeserializeOwned,
-{
-    req.extract(R::METHOD)
+pub fn command() -> std::result::Result<(), crate::error::Error> {
+
+    let mut rt = Runtime::new().unwrap();
+
+    let stdin = tokio::io::stdin();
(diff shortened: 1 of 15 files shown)


lpil:
Hello! Just a quick note to say that I'm looking forward to reviewing this but I've been super busy at work. Thanks for your patience

src/language_server.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// Runs the language server with the given input and output streams.
+// Returns true if the server shutdown safely before exiting, otherwise false.
+fn run_server<I, O>(stdin: I, stdout: O) -> std::io::Result<bool>
+where
+    I: AsyncRead + Unpin,
+    O: AsyncWrite,
+{
+    let mut rt = Runtime::new().unwrap();
```
lpil, line 101: We never use `unwrap` or `expect` in this codebase as it violates the type system while providing a poor experience in the event of a crash due to the lack of type safety.

Ideally we would write code that is always type safe, but if that's not possible use the `GleamExpect` trait's `.gleam_expect` method which prints a helpful error message for the user about the compiler bug.

src/language_server/document.rs
```diff
@@ -0,0 +1,213 @@
...
+                apply_content_change(&contents, change)
+            });
+    }
+
+    pub fn contents(&self) -> &str {
+        &self.contents
+    }
+}
+
+fn apply_content_change(contents: &str, change: &TextDocumentContentChangeEvent) -> String {
```
lpil, line 32: Nice!

src/language_server/document.rs
```diff
@@ -0,0 +1,213 @@
...
+                if unchanged_tail.is_empty() { "" } else { "\n" },
+                unchanged_tail
+            )
+        }
+        None => change.text.clone(),
+    }
+}
+
+#[cfg(test)]
+mod test {
```
lpil, line 62: Love the thorough tests

src/language_server/vfs.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        Ok(VFS {
+            overlays: RwLock::new(HashMap::new()),
+            http_client: client,
+        })
+    }
+
+    pub fn get_document_contents(&self, uri: &Url) -> io::Result<String> {
+        {
+            let hm = self.overlays.read().unwrap();
```
lpil, line 33: Let's remove the `unwrap`s in this file. 

I think they will trigger if a mutex is poisoned, which is a compiler bug rather than a user error so we can use `.gleam_expect`.

src/language_server/vfs.rs
```diff
@@ -0,0 +1,153 @@
...
+                let document = &mut *doc_lock.write().unwrap();
+
+                f(document);
+                Some(())
+            }
+            None => None,
+        }
+    }
+
+    #[allow(dead_code)]
```
lpil, line 62: Dead code?

test/language_server/run_langserv.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
lpil, line 1: I really like these tests! Sadly Gleam aims to be totally cross platform so bash isn't an ideal scripting language. Could we convert these to run using NodeJS or similar? That was windows users can run the tests and we can run these on our Windows CI build.

lpil (commented):
This is wonderful work! Thank you so much!

I've left some comments inline largely around error handling and documentation. Let me know your thoughts and if there's stuff I can help with.

src/language_server/vfs.rs
```diff
@@ -0,0 +1,153 @@
...
+use tower_lsp::lsp_types::Url;
+
+#[derive(Debug)]
+pub struct VFS {
+    overlays: RwLock<HashMap<Url, RwLock<Document>>>,
+
+    http_client: Client,
+}
+
+impl VFS {
```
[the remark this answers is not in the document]
adituv: VFS is the conventional for Virtual File System.  We could use the full name though, of course.

src/language_server/vfs.rs
```diff
@@ -0,0 +1,153 @@
+use std::collections::HashMap;
+use std::io;
+use std::io::ErrorKind;
+use std::sync::RwLock;
+
+use super::document::Document;
+
+use reqwest::blocking::Client;
```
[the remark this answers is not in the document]
adituv: I did not realise this would block the event loop.  I was working on a threaded mental model so thought that blocking client would be fine.  Async client would be preferable in this case.

src/language_server/vfs.rs
```diff
@@ -0,0 +1,153 @@
...
+use tower_lsp::lsp_types::Url;
+
+#[derive(Debug)]
+pub struct VFS {
+    overlays: RwLock<HashMap<Url, RwLock<Document>>>,
```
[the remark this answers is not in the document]
adituv: Overlays are a virtual file system concept that let you stack file systems on top of each other.  This being left in as-is is an unfortunate refactoring mistake from the more complicated VFS model I had in my first draft.

src/language_server.rs
```diff
@@ -0,0 +1,140 @@
...
+        let mut result = InitializeResult::default();
+        result.capabilities.document_formatting_provider = Some(true);
+        Ok(result)
+    }
+    async fn did_open(&self, params: DidOpenTextDocumentParams) {
+        self.vfs
+            .create_document(&params.text_document.uri, &params.text_document.text);
+    }
+    async fn did_change(&self, params: DidChangeTextDocumentParams) {
+        // TODO: validate versioned changes
```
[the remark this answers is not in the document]
adituv: I assume this refers to the TODO comment?

Documents are synchronised in LSP by sending versioned change events, roughly comprising a document version number and a diff to get from version (n-1) to n.  This comment means to check that we are applying the diff to the right version, and to error if we are not.

Unfortunately, the LSP doesn't provide a good way of forcing resynchronisation from the server side - instead it recommends for the server to crash and for the client to restart.

src/language_server.rs
```diff
@@ -0,0 +1,140 @@
...
+            Ok(x) => Ok(Some(x)),
+            Err(s) => Err(Error {
+                code: ErrorCode::ParseError,
+                message: s,
+                data: None,
+            }),
+        }
+    }
+    async fn shutdown(&self) -> Result<()> {
+        if let Ok(ref mut did_shutdown_ref) = self.did_shutdown.try_write() {
```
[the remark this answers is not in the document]
adituv: Yes, we probably could.  I can't remember why I chose to go with try_write here and write seems to make more sense.

src/language_server/document.rs
```diff
@@ -0,0 +1,213 @@
+use tower_lsp::lsp_types::TextDocumentContentChangeEvent;
+
+#[derive(Debug)]
+pub struct Document {
```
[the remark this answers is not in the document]
adituv: Sure.  A Document represents a source code file in LSP.  However, a document itself is "representationless" in LSP, instead being only the file contents as a string.  We need to track version information for text synchronisation, hence the separate structure.

In a multilingual language server, it would need to be extended with what the document's language is (e.g. gleam, toml).

src/main.rs
```diff
@@ -197,6 +201,8 @@ enum Docs {
 fn main() {
     initialise_logger();
 
+    let mut exit_code = 0;
```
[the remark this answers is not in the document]
adituv: Sure.  The thought behind this was that the language server needs to return code 1 if it receives an exit notification before a shutdown notification, which is not an error the end-user needs to be aware of, as it's a consequence of a poorly-coded client.

From the user's point of view, a server that has exited normally has not errored, so to my mind it shouldn't display an error there.  This change was for minimal impact, to avoid rewriting the error code to allow suppressing showing errors to the user in certain cases.

src/language_server/vfs.rs
adituv: Yes.  This function is no longer used with the current refactoring of the main language_server file, but I was intending to use it as part of checking document versions in the didChange notification handling.

If we are adding the language-server without that improvement then this function can be removed along with its annotation.

test/language_server/run_langserv.sh
adituv: These tests being left in were a mistake!  The intention was to replace them wholly with tests in rust using [tower-test](https://docs.rs/tower-test/0.3.0/tower_test/index.html)

adituv:
Thanks for your review.  I'm not currently working on this, but I am happy for someone else to take ownership of this and improve it based on your review.

lpil:
I'm going to close this due to lack of activity. Please feel free to reopen! Thank you

PR #838, opened 2020-11-08
Treat lines with only spaces as empty

### LINKED ISSUE #833: Formatter: consistent handling of empty lines
in 0.12.0, if you run `gleam format` on this: 
```gleam
fn something() {
  let x = 1

  let y = 2
  x + y
}
```
The result can differ. 

If there are any spaces in the empty line, the formatter will remove the whole line, like this:
```gleam
fn something() {
  let x = 1
  let y = 2
  x + y
}
```
On the other hand, if the empty line is just a linebreak, the formatter will preserve it. This seems a little unexpected, and it might be better to handle both cases the same way.

(If I get a vote, I would prefer it if the formatter would delete the unnecessary spaces but preserve the linebreak - personally I find it easier to read code if I can space it out using a single linebreak between sections.)

michaeljones:
I imagine this could use a test too but I wanted to see if it seemed acceptable. Manually tested with the code snippet #833.

----

We change the empty-line identifying logic to treat lines with only
spaces as also being empty. Previously they would not register as being
empty and then would be lost in the formatting as they have no
meaningful content.

As multiple empty lines are collapsed together this means that multiple
lines with nothing but spaces will also be collapsed together.

Closes #833

### DIFF
--- a/src/format/tests.rs
+++ b/src/format/tests.rs
@@ -2623,6 +2623,20 @@ fn empty_lines() {
 "
     );
 
+    // Lines with only spaces are treated as empty
+    assert_format_rewrite!(
+        "pub fn main() {
+  let x = 1\n    \n  x
+}
+",
+        "pub fn main() {
+  let x = 1
+
+  x
+}
+"
+    );
+
     assert_format!(
         "pub fn main() {
   let inc = fn(a) { a + 1 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - The compiler can now discriminate between record access and module access
   for shadowed names
 - Check new project names against Erlang standard library modules to avoid clashes.
+- Formatter changed to treat lines with only spaces as empty.
 
 ## v0.12.0 - 2020-10-31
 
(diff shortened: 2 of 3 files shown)


michaeljones:
For what it is worth, I'm never that impressed when I come up with code patterns that involve "set boolean to false and set it to true somewhere in a loop". Maybe there is a cleaner way to do it.

michaeljones:
Rebased to add a failing test in a separate first commit to illustrate the issue.

src/format/tests.rs
```diff
@@ -2623,6 +2623,22 @@ fn empty_lines() {
 "
     );
 
+    // Lines with only spaces are treated as empty
+    assert_format_rewrite!(
+        "pub fn main() {
+  let x = 1
+
```
lpil, line 2630: Can we rewrite this as a line with `\n` at the end? An empty line with just spaces is hard to read and may be erased by some editors.

lpil (commented):
Looks good to me, thank you.

RE the variable: It's not functional, but that's fine! Rust can be very procedural if that's useful, and I think the functional alternative would be harder to understand.

I've left a small note inline.

src/format/tests.rs
michaeljones: Good point. I've rewritten and rebased. I wasn't sure the best way. I guess it could all be collapsed to one line but I've just collapsed the lines around it.

lpil (approved):
Wonderful, thank you

PR #839, opened 2020-11-09
Improve table styling in docs

michaeljones:
For tables created in markdown. Previously it was hard to see the
separate columns clearly due to the lack of spacing and borders.

This styling is heavily based on Github's table rendering but uses the
pink-white colouring for the alternate row background.

Screenshot 2020-11-09 at 17 06 21

---

I worry about the top level `table` styles. It might be worth wrapping some kind of `rendered-markdown` class around all rendered markdown so that some of these styles can be scoped to just markdown areas. I'm not sure what the impacts of this change might be more generally. Are there other tables in the docs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
   for shadowed names
 - Check new project names against Erlang standard library modules to avoid clashes.
 - Formatter changed to treat lines with only spaces as empty.
+- Improve clarity of HTML table display in documentation.
 
 ## v0.12.0 - 2020-10-31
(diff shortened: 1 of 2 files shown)


michaeljones:
I can see that the "rendered-markdown" class does exist and is used but I don't think it is used to wrap the contents of the README so it wouldn't work for the "gleam http" case.

lpil (approved):
I think applying this styling globally is a good way to go. Thanks for this, I really like the design you've gone for here.

PR #840, opened 2020-11-10
Approach for generating JS code from Gleam Source

CrowdHailer:
This is one approach that turns Gleam language to JS language. **It does not include any runtime assumptions or have any analogue to processes/OTP**

- [ ] Design JS terms that will be generated for Gleam language elements, e.g. record/tuple/list
- [ ] Write codegen module to replace https://github.com/gleam-lang/gleam/blob/main/src/erl.rs
- [ ] Translate codegen tests to JS output.
- [ ] introduce CLI switch for output format.

This is a speculative PR, will remain in draft a while probably. Happy to move to my own repo, if this is considered a distraction from the more important issues/PRs here

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -48,6 +48,12 @@ pub fn generate_erlang(analysed: &[Analysed]) -> Vec<OutputFile> {
             path: gen_dir.join(format!("{}.erl", erl_module_name)),
             text: module(ast),
         });
+
+        files.push(OutputFile {
+            // TODO js module names instead
+            path: gen_dir.join(format!("{}.js", erl_module_name)),
+            text: js_module(ast),
+        });
     }
 
     files
@@ -137,6 +143,22 @@ pub fn record_definition(name: &str, fields: &[&str]) -> String {
     buffer
 }
 
+
+pub fn js_module(module: &TypedModule) -> String {
+    // Not sure why statements need a module_name
+    let module_name = module.name.as_slice();
+    let statements = concat(
+        module
+        .statements
+        .iter()
+        .flat_map(|s| js_statement(s, module_name))
+        .intersperse(lines(2)),
+    );
+    statements
+    .append(line())
+    .format(80)
+}
+
 pub fn module(module: &TypedModule) -> String {
     let module_name = module.name.as_slice();
     let exports = concat(
@@ -189,6 +211,41 @@ pub fn module(module: &TypedModule) -> String {
         .append(statements)
         .append(line())
         .format(80)
+}
+
+
+fn js_statement(statement: &TypedStatement, module: &[String]) -> Option<Document> {
+    let mut env = Env::new(module);
+
+    match statement {
+        Statement::CustomType { .. } => None,
+        Statement::Fn {
+            args, name, body, public, ..
+        } => {
+            let fun_args = wrap_args(args.iter().map(|a| match &a.names {
+                ArgNames::Discard { .. } | ArgNames::LabelledDiscard { .. } => "_".to_doc(),
+                ArgNames::Named { name } | ArgNames::NamedLabelled { name, .. } => {
+                    name.clone().to_doc()
+                }
+            }));
+
+            let prefix = if *public { "export function " } else { "function "};
+            let print = prefix.to_doc()
+                .append(name.to_string())
+                .append(fun_args)
+                .append(" {")
+                .append(line().append(js_expr(body, &mut env)).nest(INDENT).group())
+                .append(line())
+                .append("}");
+
+            // Some("".to_doc().append(fun_args).append(name.clone().to_doc()))
+            Some(print)
+        },
+        s => {
+            println!("{:?}", s);
+            unimplemented!()
+        }
+    }
 }
 
 fn statement(statement: &TypedStatement, module: &[String]) -> Option<Document> {
@@ -1077,6 +1134,17 @@ fn maybe_block_expr(expression: &TypedExpr, env: &mut Env<'_>) -> Document {
     match &expression {
         TypedExpr::Seq { .. } | TypedExpr::Let { .. } => begin_end(expr(expression, env)),
         _ => expr(expression, env),
+    }
+}
+
+fn js_expr(expression: &TypedExpr, _env: &mut Env<'_>) -> Document {
+    match expression {
+
+        TypedExpr::Todo { label, .. } => {
+            "throw ".to_doc()
+            .append(label.clone().unwrap_or("todo".to_string()).to_doc().surround("\"", "\""))
+        },
+        _ => unimplemented!()
     }
 }
 
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -566,6 +566,45 @@ fn bin_op(name: &BinOp, left: &TypedExpr, right: &TypedExpr, env: &mut Env<'_>) 
         .append(right_expr)
 }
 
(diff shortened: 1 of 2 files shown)


src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+JS has no way to check if something is a float or integer reliably.
+`Number.isInteger(1.0) // => false`
+
+### Let bindings
+JS same.
+
+### Lists
+
+*JS lists are Arrays, so they will not have the same performance charachteristics.*
+However they match very neatly for syntax
```
lpil, line 36: I think we may need to look into implementing immutable lists in some fashion, otherwise all our recursive algorithms will become very slow.

I'm interested to see what Bucklescript and Purecript do here.

src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+let d = [4, ...b]
+```
+
+Note the first line is a match error if list length is less than 1.
+
+### Tuple
+
+Probably best covered by a JS array, which can be heterogeneous.
+
+**Is it a problem that the dynamic library can't tell the difference between tuple and list in this case?**
```
lpil, line 56: It could accept both I think, but we can worry about libraries after there's a clear design for core.

src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+
+### Case Expressions
+
+JS `switch` statement uses strict equality and is not expressive enough.
```
lpil, line 60: This gets even more complex when we consider nested patterns. There's some papers out there on implementing this efficiently, but a naive version with linear rather than logarithmic dispatch time would be OK. That's the norm in JavaScript anyway.

We can also look into what bucklescript does here.

src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+```rs
+try int_a = parse_int(a)
+```
+
+```js
+let {ok: int_a, error} = parse_int(a)
+if (error !== undefined) throw "Assertion Error"
+```
+
+### Binary syntax
```
lpil, line 158: We can probably skip bit strings to start and focus on the more common features.

lpil (commented):
A good start! Thanks for sharing this.

I like your approach RE runtime assumptions and OTP. I think this is the way to go, keeping the core language agnostic and runtime specific features in libraries.

Other things to think about:
- JavaScript doesn't have tail call optimisation. We will need to implement this.
- How do we have different code for different platforms? Some kind of conditional compilation will be required in places.

src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+```rs
+try int_a = parse_int(a)
+try int_b = parse_int(b)
+try int_c = parse_int(c)
+Ok(int_a + int_b + int_c)
+```
+
+```js
+let {ok: int_a, error} = parse_int(a)
+if (error !== undefined) return {error}
```
lpil, line 139: try is an expression rather than being function scoped so this won't work. Currently we rewrite it to a case expression but we could alternatively use return with a immediately invoked function wrapper.

src/erl.rs
```diff
@@ -137,6 +143,22 @@ pub fn record_definition(name: &str, fields: &[&str]) -> String {
     buffer
 }
 
+
+pub fn js_module(module: &TypedModule) -> String {
```
lpil, line 147: Could we put these in a different module? `javascript.rs` perhaps?

CrowdHailer: definitely, I'm really just poking around now. I though I might be able to use some of the helpers like formatting args, but I actually think that is of minimal benefit now

src/js/notes.md
CrowdHailer: I think that's probably a good idea. Although for version one I am very happy with slow but working

src/js/notes.md
CrowdHailer: Good point, I intended to have a function wrapper, same as for the case examples here https://github.com/gleam-lang/gleam/pull/840/files#diff-31b52eb625476ae2f6e51371fd97e4aa0fae93c75770e29f69b18023a26c6e48R114

QuinnWilton:
> * JavaScript doesn't have tail call optimisation. We will need to implement this.

One way I've done this before (okay, in C, not Javascript), is by using a trampoline. Reginald Braithwaite has a good article on their use in JS, to avoid blowing the stack with tail calls.

It shouldn't be too difficult of a technique for a first pass.

https://raganwald.com/2013/03/28/trampolines-in-javascript.html

lpil:
That's what I was thinking too 👍

Not entirely sure how we detect when a function is recursive but I'm sure that's very solvable

src/js/notes.md
```diff
@@ -0,0 +1,179 @@
...
+if (a === undefined) throw "match error"
+let c = [4, ...b]
+let d = [4, ...b]
+```
+
+Note the first line is a match error if list length is less than 1.
+
+### Tuple
+
+Probably best covered by a JS array, which can be heterogeneous.
```
CrowdHailer, line 54: Note that typescript uses JS arrays as tuples. https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple

lpil:
I've introduced the concept of modular code generation here: https://github.com/gleam-lang/gleam/pull/853

This will help us integrate this feature in future :)

lpil:
I'm going to close this due to lack of activity. Please feel free to reopen! It could be good to capture any learnings in an issue.

Thank you

CrowdHailer:
Further thoughts.

## Async Await vs Promises.

erlang doesn't make async or blocking functions explicit. However JavaScript and many other languages do.

**The question is: is there any way to make use of an async await approach in a Gleam -> JS program?**

### Just use promises

```rust
browser.fetch("https://example.com")
|> promise.then(fn(response) {
  browser.json_body(response)
})
|> promise.then(fn(data) {
  // work with the json
})
```

### await

```rust
try response = promise.await(browser.fetch("https://example.com"))
try data = promise.await(browser.json_body(response))
```

This feels like it nearly works, it would be possible to define a `promise.await` function that awaited a promise and then returned the result, or error if promise is rejected. However any function that used promise.await would have to be marked as async it'self.

There is a horrible hack to make every function async but I doubt that is an acceptble tradeoff

## Serverless

In some discussions, serverless seems like a very promising area to try out gleam adoption with a new audience. 
Gleam could be tried on a single function, there is low risk if the generated JavaScript is readable and JS is a huge audience.
I suspect even if targeting web assembly you might want this.

lpil:
I am very interested in serverless Gleam. As you said functions as a service could be a good low risk way for people to use Gleam, and JavaScript would be a quick and practical route to this.

I have also thought a little bit about a sync/await. At least to start I am reluctant to add any kind of concurrency primitive or syntax to the core language. This is because right now concurrency in Gleam clean is quite platform specific.￼￼ if we were to add more support for a particular style of concurrency I would probably be suggesting we go down the Erlang route￼, which poses a problem for JavaScript. We could make every function a sink as I've said, but this has performance implications and means that it would be difficult to use Gleam from JavaScript code, which contradicts some of the goals of the language.￼￼

CrowdHailer:
> I have also thought a little bit about a sync/await. At least to start I am reluctant to add any kind of concurrency primitive or syntax to the core language. 

That's fair enough. I'm not sure I liked the suggestion either.
I guess the nice thing is that you do have a fallback by using promises. and with types you can't confuse a value with a promise for a value, which was my main failure mode.
Also in serverless you often don't do any async, because there's no filesystem etc. 
With those two points can probably make progress and simply not tackle in the near term

PR #843, opened 2020-11-11
Add a 'source' link to the docs

### LINKED ISSUE #842: Link from hex docs back to repository
It would be great to have a link in the published docs for a package back to the repository to make it easier to explore and jump around between the code and the docs. I would start with just one link in the header or sidebar navigation but I guess this could also be extended to linking from types & functions back to the source code definitions.

It could work like the Rust docs:

image

Which I guess is driven by the `repository` field in the `Cargo.toml` file. We might consider having a similar field in the `gleam.toml` file.

I imagine the source code linking would be more complex and might involve tying the current build of the docs to a particular git tag or similar commit reference in the repository. A first pass might only work for Github (and other major hosting sites) where the URL format for referencing a line in a file on a tag is known. 

Elixir docs provide a `</>` link next to each function which links to the source code: 

https://hexdocs.pm/phoenix_html/Phoenix.HTML.Form.html#label/1

michaeljones:
Provided that the project `gleam.toml` file provides a `repository` field that points to the source repository's webpage.

This provides a 'Source' link in the top right of the page at the end of the header. It is currently pink on pink which is no very accessible. We could change it to black but I wanted to validate the approach first.

Given that we're expecting the `repository` field to be an http(s) link it might make sense to use `repository_http` or `repository_web` or something. In case we'd like to have a ssh link or git protocol link as well. That said, Cargo seems to have used just "repository": https://doc.rust-lang.org/cargo/reference/manifest.html#the-repository-field

I have no idea what I'm doing with strings and options and refs and everything. Hopefully there is a better way to do that.

Closes #842

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Check new project names against Erlang standard library modules to avoid clashes.
 - Formatter changed to treat lines with only spaces as empty.
 - Improve clarity of HTML table display in documentation.
+- Provide example 'links' section in new gleam.toml files.
 
 ## v0.12.0 - 2020-10-31
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -35,7 +35,10 @@ impl Default for BuildTool {
 
 #[derive(Deserialize, Default, Debug, PartialEq)]
 pub struct Docs {
+    #[serde(default)]
     pub pages: Vec<DocsPage>,
+    #[serde(default)]
+    pub links: Vec<DocsLink>,
 }
 
 #[derive(Deserialize, Debug, PartialEq, Clone)]
@@ -43,6 +46,12 @@ pub struct DocsPage {
     pub title: String,
     pub path: String,
     pub source: PathBuf,
+}
+
+#[derive(Deserialize, Debug, PartialEq, Clone)]
+pub struct DocsLink {
+    pub title: String,
+    pub href: String,
 }
 
 pub fn read_project_config(root: impl AsRef<Path>) -> Result<PackageConfig, Error> {
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -56,7 +56,15 @@ pub fn generate_html(
         })
         .collect::<Vec<_>>();
 
-    let links = &[];
+    let links = &project_config
+        .docs
+        .links
+        .iter()
+        .map(|doc_link| Link {
+            name: doc_link.title.clone(),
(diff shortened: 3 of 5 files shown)


lpil (commented):
This looks good but I think we want to go for a slightly different design (see the associated issue).

I think you've got the right approach to implementation so this can likely be adapted.

michaeljones:
I've updated this PR to match the suggested approach in #842. The end result looks like:

image

if I add:
```
[docs]
pages = []
links = [
  { title = "Github", href = "https://github.com/gleam-lang/http" }
]
```
To the `gleam/http` `gleam.toml` file.

It doesn't seem happy if I have `links` but not `pages`. I'm not sure if that should be handled by making both fields use the `Option` type?

lpil:
Sweet, thanks @michaeljones 

Could you explain a little more about what you mean by it's not happy? It should be OK with both fields being missing as a missing vector defaults to an empty vector.

michaeljones:
If I remove the `pages` entry with this implementation then I get this error:

```
$ gleam docs build

error: File IO failure

    ./gleam.toml

    missing field `pages` for key `docs` at line 4 column 1
```
I was a bit confused by it too. I can only assume that it can default the whole structure if nothing is there but struggles to default just one field for reasons that I don't understand.

lpil:
Oh! Maybe that behaviour I mentioned has to be turned on. Could you try adding the `#[serde(default)]` annotation above the `pages` and `links` fields in that struct?

michaeljones:
Yeah, that seems to have done it. Thanks!

michaeljones:
Just adding the commented out block in the new gleam.toml file now.

michaeljones:
Added the example links block to new `gleam.toml` files and updated the changelog appropriately in each commit.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 - Check new project names against Erlang standard library modules to avoid clashes.
 - Formatter changed to treat lines with only spaces as empty.
 - Improve clarity of HTML table display in documentation.
+- Fix use & display of gleam.toml links in documention.
```
lpil, line 13: Could you remove this entry? Just an internal thing so no need to expose it to the users' changelog

michaeljones: Sure

lpil (approved):
Brilliant, thank you!

PR #845, opened 2020-11-14
Link docs to source lines

michaeljones:
This is an attempt to explore having individual documentation entries link back to their source code in Github. Issues so far:

- ~~Reconstructing the module file path and~~ re-reading the source code. I've avoided adding more fields to structs to make them available in the docs code out of fear that it would impact the compilation itself but we can save work but passing references to the filename and source string through to this point.
- Not sure how to get the main Github URL to the doc page. Could look through the "links" in the `gleam.toml` and act if there is one called "GitHub" that references a "github.com" URL but perhaps we might benefit from moving to a `repository` entry in `gleam.toml` after all. **Edit**: I've experimented with this in a commit.
- ~~It might be sensible to only do this for release builds and so perhaps require that some kind of flags are passed to the `gleam docs build` command that suggest what tag to target on the Github side. It doesn't make sense to try if there are uncommitted changes in the repo or for general untagged releases, I imagine.~~ I'm using the `version` entry from the `gleam.toml` file.
- Should ideally work for other common git hosting solutions, eg. Gitlab, BitBucket, Gitea, SourceHut? Possibly can be delayed through due to the current size & spread of the Gleam community.
- It only does it for functions at the moment but that should be easy to extend to other entries.
- The `SrcSpan` `start` and `end` for functions seem to cover the function head and not the body so linking the start & end through to GitHub is slightly less satisfying than it could be as it only highlights the first line or two and not the whole body.

As ever, I don't really know what I'm doing with Rust with the language itself or the APIs so I'm sure there is lots to improve.

### DIFF
--- a/src/build/package_compilation_tests.rs
+++ b/src/build/package_compilation_tests.rs
@@ -40,7 +40,7 @@ fn package_app_file(modules: &[&str]) -> OutputFile {
     OutputFile {
         text: format!(
             r#"{{application, the_package, [
-    {{vsn, "1.1.0"}}
+    {{vsn, "1.1.0"}},
     {{applications, []}},
     {{description, "the description"}},
     {{modules, [{}]}},
@@ -1254,7 +1254,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
-    {vsn, "1.0.0"}
+    {vsn, "1.0.0"},
     {applications, []},
     {description, ""},
     {modules, []},
@@ -1274,7 +1274,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
-    {vsn, "1.3.5"}
+    {vsn, "1.3.5"},
     {applications, []},
     {description, ""},
     {modules, []},
@@ -1294,7 +1294,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
-    {vsn, "1.0.0"}
+    {vsn, "1.0.0"},
     {applications, []},
     {description, "Very exciting"},
     {modules, []},
@@ -1322,7 +1322,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
-    {vsn, "1.0.0"}
+    {vsn, "1.0.0"},
     {applications, [gleam_otp,
                     gleam_stdlib,
                     midas,
--- a/src/build/package_compilation_tests.rs
+++ b/src/build/package_compilation_tests.rs
@@ -6,7 +6,7 @@ use crate::{
         project_root::ProjectRoot,
         Origin,
     },
-    config::{BuildTool, Docs, PackageConfig},
+    config::{BuildTool, Docs, PackageConfig, Repository},
     erl, typ,
 };
 use std::{path::PathBuf, sync::Arc};
@@ -17,8 +17,9 @@ macro_rules! assert_erlang_compile {
         let config = PackageConfig {
             dependencies: HashMap::new(),
             description: "the description".to_string(),
-            version: Some("1.1.0".to_string()),
+            version: "1.1.0".to_string(),
             name: "the_package".to_string(),
+            repository: Repository::None,
             docs: Default::default(),
             otp_start_module: None,
             tool: BuildTool::Gleam,
@@ -1239,8 +1240,9 @@ fn config_compilation_test() {
         PackageConfig {
             dependencies: HashMap::new(),
             description: "".to_string(),
-            version: None,
+            version: "1.0.0".to_string(),
             name: "the_package".to_string(),
+            repository: Repository::None,
             docs: Default::default(),
             otp_start_module: None,
             tool: BuildTool::Gleam,
@@ -1252,6 +1254,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
+    {vsn, "1.0.0"}
     {applications, []},
     {description, ""},
     {modules, []},
@@ -1265,7 +1268,7 @@ fn config_compilation_test() {
 
     // Version is included if given
     let mut config = make_config();
-    config.version = Some("1.3.5".to_string());
+    config.version = "1.3.5".to_string();
     assert_config_compile!(
         config,
         vec![],
@@ -1291,6 +1294,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
+    {vsn, "1.0.0"}
     {applications, []},
     {description, "Very exciting"},
     {modules, []},
@@ -1318,6 +1322,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
+    {vsn, "1.0.0"}
     {applications, [gleam_otp,
                     gleam_stdlib,
                     midas,
--- a/src/config.rs
+++ b/src/config.rs
@@ -38,7 +38,7 @@ impl Default for BuildTool {
 #[derive(Deserialize, Debug, PartialEq)]
 #[serde(tag = "type")]
 pub enum Repository {
-    GitHub { url: String },
+    GitHub { user: String, repo: String },
     Other { url: String },
     None,
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -18,6 +18,8 @@ pub struct PackageConfig {
     pub dependencies: HashMap<String, String>,
     #[serde(default)]
     pub otp_start_module: Option<String>,
+    #[serde(default)]
+    pub repository: Repository,
 }
 
 #[derive(Deserialize, Debug, PartialEq)]
@@ -30,6 +32,20 @@ pub enum BuildTool {
 impl Default for BuildTool {
     fn default() -> Self {
         Self::Other
+    }
+}
+
+#[derive(Deserialize, Debug, PartialEq)]
+#[serde(tag = "type")]
+pub enum Repository {
+    GitHub { url: String },
+    Other { url: String },
+    None,
+}
+
+impl Default for Repository {
+    fn default() -> Self {
+        Self::None
     }
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -38,7 +38,9 @@ impl Default for BuildTool {
 #[serde(tag = "type")]
 pub enum Repository {
     GitHub { user: String, repo: String },
-    Other { url: String },
+    GitLab { user: String, repo: String },
+    BitBucket { user: String, repo: String },
+    Custom { url: String },
     None,
 }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -6,8 +6,7 @@ use std::path::{Path, PathBuf};
 #[derive(Deserialize, Debug, PartialEq, Default)]
 pub struct PackageConfig {
     pub name: String,
-    #[serde(default)]
-    pub version: Option<String>,
+    pub version: String,
     #[serde(default)]
     pub description: String,
     #[serde(default)]
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -3,8 +3,8 @@ pub(crate) mod command;
 mod tests;
 
 use crate::{
-    ast::{Statement, TypedStatement},
-    config::{DocsPage, PackageConfig},
+    ast::{SrcSpan, Statement, TypedStatement},
+    config::{DocsPage, PackageConfig, Repository},
     error::{Error, GleamExpect},
     format,
     fs::OutputFile,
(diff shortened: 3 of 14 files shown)


michaeljones:
Updated to improved styling. Though the CSS is more verbose than I'd like.

michaeljones:
I've added a few more commits trying to tackle varies issues but I'm happy to squash them all together at some point as they don't tell a meaningful story at this point. I'm keeping them separate so that we can drop them individually more easily if anything seems just plain wrong and because I'm concerned about rewriting commits for anyone that has already looked at some.

src/config.rs
```diff
@@ -33,6 +35,20 @@ impl Default for BuildTool {
     }
 }
 
+#[derive(Deserialize, Debug, PartialEq)]
+#[serde(tag = "type")]
+pub enum Repository {
```
lpil, line 40: What does the toml look like for this? Something like this?

```toml
name = "lib"
repository = { type = "github", url = "https://github.com/one/two" }
```

If it's GitHub we probably don't need a `url` but a `user` and `repo`?

```toml
name = "lib"
repository = { type = "github", user = "gleam-lang", repo = "gleam" }
```
```

File: src/docs.rs [hunk 2]
```diff
@@ -153,7 +171,81 @@ pub fn generate_html(
...
+                version: version.clone(),
+            })
+        }
+        // If we either don't have a version or aren't dealing with a GitHub repository then we
+        // can't generate source code links so we generate empty strings for the urls
+        _ => Box::new(EmptySourceLinkGenerator {}),
+    }
+}
+
+trait SourceLinkGenerator {
```
lpil, line 204: Nitpick: Let's call this `SourceLinks` and move all related code to `src/docs/source_links.rs`

src/docs.rs
```diff
@@ -153,7 +171,81 @@ pub fn generate_html(
...
+// Creates empty strings for the urls which are then ignored by the template
+struct EmptySourceLinkGenerator {}
+
+impl SourceLinkGenerator for EmptySourceLinkGenerator {
+    fn url(&self, _location: &SrcSpan) -> String {
+        String::new()
+    }
+}
+
+// Creates links to the GitHbu repository with the #LN-LM anchor syntax for linking directly to a
```
lpil, line 217:
```suggestion
// Creates links to the GitHub repository with the #LN-LM anchor syntax for linking directly to a
```

src/main.rs
```diff
@@ -200,7 +200,12 @@ fn main() {
     let result = match Command::from_args() {
         Command::Build { project_root } => command_build(project_root),
 
-        Command::Docs(Docs::Build { project_root, to }) => docs::command::build(project_root, to),
+        Command::Docs(Docs::Build { project_root, to }) => PathBuf::from(&project_root)
+            .canonicalize()
+            .map_err(|_| Error::UnableToFindProjectRoot {
```
lpil, line 205: When can this error happen? And why does the path need to be canonicalized?

src/docs.rs
```diff
@@ -153,7 +171,81 @@ pub fn generate_html(
...
+) -> Box<dyn SourceLinkGenerator> {
+    match (&project_config.version, &project_config.repository) {
+        (Some(version), Repository::GitHub { url }) => {
+            let relative_path = module_path
+                .strip_prefix(&project_root)
+                .map(|path| path.to_str())
+                .unwrap_or_default()
+                .unwrap_or_default();
+
+            let src = std::fs::read_to_string(&module_path).unwrap_or_default();
```
lpil, line 188: Rather than read the file again I think we probably have the source code already in memory at a higher level, we can pass it down into this function.

src/docs.rs
```diff
@@ -112,10 +125,15 @@ pub fn generate_html(
             modules: modules_links.as_slice(),
             project_name: &project_config.name,
             page_title: &format!("{} - {}", name, project_config.name),
-            module_name: name,
+            module_name: name.clone(),
             project_version: "", // TODO
```
lpil, line 129: If we have a version can we fill this?

src/docs.rs
```diff
@@ -153,7 +171,81 @@ pub fn generate_html(
...
 }
 
-fn function(statement: &TypedStatement) -> Option<Function<'_>> {
+fn build_source_link_generator(
+    project_root: impl AsRef<Path>,
+    project_config: &PackageConfig,
+    module_path: &PathBuf,
+    name: &String,
+) -> Box<dyn SourceLinkGenerator> {
+    match (&project_config.version, &project_config.repository) {
```
lpil, line 180: We have a problem here in that there's nothing that guarentees that the version in `gleam.toml` matches the version in `.app`. This will be very error prone.

Perhaps we should make `version` a required field and then add some code that checks `.app` to see if it is the same version as `gleam.toml`. We don't have a parser for `.app` yet so can just check it's a substring to start.

lpil (commented):
This is absolutely fantastic work! And your Rust is good too :) 

I've left a bunch of comments, let me know what you think. Thank you

src/docs.rs
```diff
@@ -153,7 +171,81 @@ pub fn generate_html(
...
+            })
+        }
+        // If we either don't have a version or aren't dealing with a GitHub repository then we
+        // can't generate source code links so we generate empty strings for the urls
+        _ => Box::new(EmptySourceLinkGenerator {}),
+    }
+}
+
+trait SourceLinkGenerator {
+    fn url(&self, location: &SrcSpan) -> String;
```
lpil, line 205: Let's use `Option<String>` so `None` is used to represent an absent URL rather than `""`

src/config.rs
michaeljones: Yeah, the top one is the current state. I agree your suggest looks better and we can certainly aim for that. I guess the only downside is if we want to account for enterprise deployments of GitHub. In some ways the "GitHub" here is to indicate the URL format for linking to lines rather than to say it has to be one `github.com`. That said, your one looks nicer :)

src/docs.rs
michaeljones: Sounds good.

src/main.rs
michaeljones: No essential perhaps but we hit an error if provide a non-existent project directory as the first argument to `gleam docs build`:

```
$ gleam docs build non-existent-directory

error: Invalid project root

We were unable to find the project root:

  non-existent-directory
```
Without this we get an error when it tries to read `non-existent-directory/gleam.toml` which is relatively readable though not the clearest.

The motivation was this part:

```
            let relative_path = module_path
                .strip_prefix(&project_root)
```
Without canonicalising the `project_root`, that value is just `.` and so the `strip_prefix` doesn't work in the way that I hoped. That said, there might be a better approach to deriving a relative path than this.

The canonicalising does have side effects, eg. at the end of the `gleam docs build` output, it prints the absolute path to the docs folder instead of just `./gen/docs`.

src/docs.rs
michaeljones: I wasn't sure if that was ideal for other compiler operations but if you're happy with that, I can try to look into it.

src/docs.rs
michaeljones: @michallepicki has done that on his branch. I can do the same or just leave it for his side.

michallepicki: Should probably look differently than currently on my branch if version will be mandatory

src/docs.rs
michaeljones: Can certainly try to do that. It also means that the app version, gleam version and git tag should probably all match. `gleam/http` uses a `v` prefix for the tag but not the app file version.

src/docs.rs
michaeljones: Makes sense. The choice was driven by what you can do in the template syntax but I guess we can convert from `Option<String>` to `String` when preparing the `Function` struct in the docs file. I agree that `Option<String>` is a much truer representation of the data.

src/docs.rs
michaeljones: I've done my best here. Passing the src as a string along on the `Analysed` struct. I do end up cloning it though as I'm not sure about the lifetimes where I'm using it. Maybe it can be avoided though.

src/docs.rs
michaeljones: I've made the 'version' required and so I just pass it in here by reference. We'll see how it all looks together.

src/docs.rs
michaeljones: I've attempted this. Happy to have feedback. I guess it isn't great as I'm using the `?` operator and returning a result to early exit from the calling function. Probably could be better.

michaeljones:
@lpil - no rush but I think this is ready for another round of feedback. Definitely some stuff there that could be better. Interested to know your thoughts though. 

The version check did catch that the `gleam-lang/http` `app.src` file has 1.3.0 despite having a git tag of 1.5.0. Not that it checks the git tag but I put the git tag value in the `gleam.toml` file.

src/project.rs
```diff
@@ -213,3 +226,35 @@ pub fn collect_source(
...
+
+fn check_app_file_version_matches(
+    root: impl AsRef<Path>,
+    project_config: &PackageConfig,
+) -> Result<(), Error> {
+    let mut app_src_path = root.as_ref().to_path_buf();
+    app_src_path.push("src");
+    app_src_path.push(format!("{}.app.src", &project_config.name));
+
+    let contents = std::fs::read_to_string(&app_src_path)
```
lpil, line 238: Could we make this function return `Ok(())` if the .app file isn't present? That will be the case for mix users.

src/project.rs
```diff
@@ -213,3 +226,35 @@ pub fn collect_source(
...
+            action: FileIOAction::Read,
+            kind: FileKind::File,
+            err: Some(err.to_string()),
+            path: app_src_path.clone(),
+        })?
+        // Remove all new lines so we can regex easily across the content
+        .replace("\n", "");
+
+    // Search for the vsn line & extract the text in double quotes
+    let re = Regex::new("\\{ *vsn *, *\"([^\"]*)\" *\\}").unwrap();
```
lpil, line 249: Can we use `.gleam_expect` here instead of `unwrap`? It produces a better error message to the user.

src/project.rs
```diff
@@ -213,3 +226,35 @@ pub fn collect_source(
...
+    let caps = re.captures(&contents).unwrap();
+
+    if caps.get(1).unwrap().as_str() == project_config.version {
```
lpil, line 252: Let's make this function return OK if it cannot determine the version number, it is a best-attempt. We don't want to ever call `unwrap` and crash the program.

lpil (approved):
Loving how this is going! And that's for your patience again, Busy with the new job and all!

I've left a couple small comments. I am happy with the history however you want so long as it is relatively clean. :)

lpil (changes requested):
Oops, wrong button. See inline comments

michaeljones:
Thanks, I'll address them. I've also added an extra commit for handling GitLab and BitBucket urls. I couldn't find a good abstraction for all of them so there is a lot of repetition. Also, I'm not sure if BitBucket is actively hosting public projects any more but I guess it still exists and some people might use it.

src/project.rs
michaeljones: Done.

src/project.rs
michaeljones: Done.

src/project.rs
michaeljones: Done - yeah, I still don't have much head wrapped around error handling in Rust. I guess I should try to remember that plain `unwrap` is going to panic and shouldn't be used. I'll try to keep it in mind.

I've changed the whole thing to chained calls. I hope it isn't going too far. I suspect the function could be refactored to be more testable if needed but it seems to work.

lpil (approved):
Wonderful, thank you!

src/project.rs
lpil: Looks great! Really digging your functional style here.

src/build/package_compilation_tests.rs
```diff
@@ -1252,6 +1254,7 @@ fn config_compilation_test() {
         vec![],
         vec![OutputFile {
             text: r#"{application, the_package, [
+    {vsn, "1.0.0"}
```
michaeljones, line 1257: @lpil - I am a bit confused why this doesn't need a comma after it?

lpil: well spotted, it needs a comma. Not sure why it is missing

michaeljones: The tests are passing with it and there are other examples in the file that have never had a comma. Maybe make a new ticket for it? I tried to look into it but I couldn't figure out how that string is assembled.

michaeljones: Found it. I'll try to fix it up.

michaeljones: Done. I think.

michaeljones:
Would you like me to squash the commits? Or is there anything else outstanding before merging? I realise I haven't added any tests. I also realise you're busy with your new job and balancing other gleam projects. So no rush. Just curious that it was approved but not merged (though I've been happy to make other commits here.)

lpil:
I was waiting for CI to pass and then got distracted :)

Thanks Michael! This is fantastic work!

michaeljones:
Great, thank you :) So happy to be able to contribute.

PR #846, opened 2020-11-14
Add a basic version selector to rendered documentation

### LINKED ISSUE #844: Add a versions dropdown to package documentation
The way the versions dropdown is implemented on Elixir hexdocs pages, is based around a configurable javascript file. This file currently contains a `versionNodes` global array value that contains a list of version names with links to their documentation. (I tried to push an idea of exposing hex.pm "Links" section there as well but it wasn't heard at the time, maybe it could be proposed once again).

The javascript config file is being generated by default on hexdocs for all packages, not just Elixir ones. Thus, Gleam documentation could use it as well! Example: https://hexdocs.pm/gleam_stdlib/docs_config.js

Right now, I don't know how it could look like. Something simple, similar to the ex_doc dropdown, but placed on the top pink bar right to the package name?

PR that added the versions dropdown to ex_doc (implementation might have changed slightly since then): https://github.com/elixir-lang/ex_doc/pull/970
PR that added the js file generation to hexdocs: https://github.com/hexpm/hexdocs/pull/9

michallepicki:
Adds dropdown to switch to a different package version. Closes #844

### DIFF
--- a/templates/documentation_layout.html
+++ b/templates/documentation_layout.html
@@ -5,11 +5,15 @@
     <meta name="viewport" content="width=device-width" />
     <title>{{ page_title }}</title>
     <link rel="stylesheet" href="{{ unnest }}/index.css" type="text/css" />
+    <!-- The docs_config.js file is provided by HexDocs and shared
+         between multiple versions of the same package. -->
+    <script src="{{ unnest }}/docs_config.js"></script>
   </head>
   <body>
     <header class="page-header">
       <h2>
-        {{ project_name }}<!-- - {{ project_version }} -->
+        {{ project_name }}
+        <span id="project-version"> - v{{ project_version }} </span>
       </h2>
     </header>
 
@@ -77,5 +81,29 @@
       <div class="pink">rights</div>
       <div class="blue">✨</div>
     </footer>
+
+    <script>
+      if (typeof versionNodes !== 'undefined') {
+        const versionElement = document.getElementById('project-version')
+        const currentVersion = "v{{ project_version }}"
+        if (!versionNodes.find(element => element.version === currentVersion)) {
+          versionNodes.unshift({ version: currentVersion, url: '#' })
+        }
+        const versionsDropdown = `
+          <form autocomplete="off">
+            <select onchange="window.location.href = this.value">
+              ${versionNodes.map(element =>
+                `
+                <option value="${element.url}" ${element.version === currentVersion ? "selected disabled" : ""}>
+                  ${element.version}
+                </option>
+                `
+              ).join('')}
+            </select>
+          </form>
+          `
+        versionElement.innerHTML = versionsDropdown
+      }
+    </script>
   </body>
 </html>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Records defined in other modules can now be used in module constants.
 - Documentation can link from functions, types & constants to their source
   code definitions on popular project hosting sites.
+- Documentation hosted on HexDocs now has a version selector.
 
 ## v0.12.1 - 2020-11-15
 
--- a/templates/index.css
+++ b/templates/index.css
@@ -138,6 +138,26 @@ p code {
   margin: 0;
 }
 
+/* Version selector */
+
+#project-version form {
+  display: inline;
+}
+
+#project-version select {
+  cursor: pointer;
+  border: none;
+  appearance: none;
+  outline: 0px transparent;
+  background-color: transparent;
+  padding: 0.6rem 0.6rem;
+  font-size: 0.9rem;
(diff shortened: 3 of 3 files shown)


michallepicki:
This is how it looks right now:
image

With the `docs_config.js` path it will work out of the box on hexdocs. The path should probably be configurable.

michaeljones:
Very cool to see! :)

templates/documentation_layout.html
```diff
@@ -5,11 +5,13 @@
     <meta name="viewport" content="width=device-width" />
     <title>{{ page_title }}</title>
     <link rel="stylesheet" href="{{ unnest }}/index.css" type="text/css" />
+    <script src="{{ unnest }}/docs_config.js"></script>
```
lpil, line 8: Does this file get automatically added by hexdocs? If so can we add a comment here to explain that.

lpil (commented):
Looks good! I left a comment inline.

michallepicki:
@lpil What do you think about the project version in the toml version file? Should it be required? There is a duplication with the .app file, and I don't know what rebar3_hex uses - I think it would be good for the library maintainer to configure current version in one spot

michallepicki:
Alternatively, I can not use the current package version at all, have an empty span in place and only fill it in with a dropdown if the `versionNodes` from `docs_config.js` is loaded. To detect the current version to make it selected and visible, current page URL could be used (find the version which URL is the beginning of current page's URL).

lpil:
In Michael's pull request I was suggesting that it was made mandatory and a check is added to ensure that the version in `gleam.toml` matches the one in `.app`. We don't have a parser for that file though so it would probably have to be something simple like checking for a substring to start.

What do you think?

michallepicki:
I like the idea! I will try to get something working, I don't know much Rust but I always wanted to learn more about it

michallepicki:
If @michaeljones gets around to it sooner then that's fine as well and I can update this PR accordingly.

michaeljones:
I'm not working on anything yet. Might have a bit of time today but there
are other things to address on my PR too so happy for someone else to
tackle that!

michallepicki:
One more thing I'll note here so that I won't forget: ExDoc currently also tries to redirect to the same page. They do that by splitting the current URL by slashes, taking just the last (file) part and appending it to chosen version URL. Then they do an AJAX call first, and if it succeeds, they navigate to that page. If it fails, they navigate to just the version URL. [code](https://github.com/elixir-lang/ex_doc/blob/2cc44d503287e31581657513463c68a172a95916/assets/js/versions.js#L44)

I did not include this here because Elixir has a flat documentation file structure and Gleam has files nested in subdirectories. Also I think it's kind of a hack. But it works so maybe if I don't come up with anything better, I will include something similar here.

lpil:
Thanks for the info @michallepicki . I agree it's not clear exactly what the best approach for that is yet. Let's hold off for now and tackle that later.

templates/documentation_layout.html
```diff
@@ -5,11 +5,15 @@
...
+    <!-- The docs_config.js file is provided by HexDocs and shared
+         between multiple versions of the same package. -->
+    <script src="{{ unnest }}/docs_config.js"></script>
   </head>
   <body>
     <header class="page-header">
       <h2>
-        {{ project_name }}<!-- - {{ project_version }} -->
+        {{ project_name }}
+        <span id="project-version"> - v{{ project_version }} </span>
```
michallepicki, line 16: Not sure about that, do we want to show the version now whenever the `docs_config.js` wasn't loaded?

michallepicki:
Updated, should be ready for review now! To test this, after generating docs, I was downloading the `docs_config.js` file from HexDocs and saving it in the generated docs directory. Clicking a link for a chosen version in the dropdown should open this version's docs on HexDocs.

michallepicki:
I think I was too quick, looks bad on chrome:
image

I'll find some time this week to test on various browsers. But the versions switcher should be functional now (even if ugly on some browsers).

michallepicki:
The appearance on Chrome should now be better. I used an svg background to style the dropdown because with the `::after` pseudo-element, I could not find a good absolute positioning that worked on both Chrome and Firefox.

michallepicki:
Published something temporarily to https://michallepicki.github.io/otp/ for testing

templates/documentation_layout.html
lpil: I think this is good. We can always improve it later :)

lpil (approved):
Very nice! This'll be a big usability improvement going forward. Thanks!

michallepicki:
Unfortunately even after removing the svg rotation I still see a double arrow on Android:
image

I'll try to find a fix and submit a followup PR

lpil:
Good find. Thank you

PR #847, opened 2020-11-14
Reword docs command help strings

michaeljones:
To remove the 'ah' and to avoid ending a sentence on a 'to'. 

These are perhaps subjective changes. Feel free to ignore :)


lpil (approved):
Thanks!

I wonder what "ah" was

PR #849, opened 2020-11-18
Variable names now only have 1st letter capitalized in erlang

### LINKED ISSUE #848: Trailing underscores in variable names get compiled away causing crash
Sorry for the bad title, didn't know what to write.

```
$ gleam --version
gleam 0.12.1
```

Variables with trailing underscores (`name_`) are allowed in gleam, but the underscores get compiled away making them indistinguishable from the same variable name without trailing underscores in erlang.

This Gleam:
```rust
import gleam/string

pub fn this_will_crash(name_: String) -> String {
  let name = string.concat(["hi ", name_])
  name
}
```

Gets compiled to:
```erlang
-module(gleam_crash).
-compile(no_auto_import).

-export([this_will_crash/1]).

% Notice that all the variables are named Name (no underscore)
this_will_crash(Name) ->
    Name = gleam@string:concat([<<"hi "/utf8>>, Name]),
    Name.
```

Which will crash when run, as erlang doesn't allow variable re-binding
and it thinks this is a pattern match against an existing value:
```
Failures:

  1) gleam_crash_test:crash_test/0: module 'gleam_crash_test'
     Failure/Error: {error,
                        {badmatch,<<"hi greg">>},
                        [{gleam_crash,this_will_crash,1,
                             [{file,
                                  "/Users/a/Trash/gleam_crash/gen/src/gleam_crash.erl"},
                              {line,7}]},
                         {gleam_crash_test,crash_test,0,
                             [{file,
                                  "/Users/a/Trash/gleam_crash/gen/test/gleam_crash_test.erl"},
                              {line,8}]},
                         ...
```

itsgreggreg:
Previously gleam was converting variable names to camel case when compiling from gleam to erlang. This causes ambiguity with certain valid gleam names on conversion to erlang names, see below for examples. With this PR, only the 1st letter of the variable name is capitalized when converting to erlang. This should be fine as valid gleam variable names are a subset of valid erlang variable names. Incidentally the old algorithm was also removing valid gleam and erlang characters like `@`, treating them as word breaks.

#### Before
| gleam | erlang|
|-|-|
|name_|Name|
|name__| Name|
|name|Name|
|one_1| One1|
|one1|One1|
|one_two_three|OneTwoThree|
|gleam@example|GleamExample|

#### Now
| gleam | erlang|
|-|-|
|name_|Name_|
|name__| Name__|
|name|Name|
|one_1| One_1|
|one1|One1|
|one_two_three|One_two_three|
|gleam@example|Gleam@example|

fixes #848

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Variable names now only have 1st letter capitalized when converted to erlang.
 - Gleam can now compile Gleam projects without an external build tool.
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -11,7 +11,7 @@ use crate::{
         ModuleValueConstructor, PatternConstructor, Type, ValueConstructor, ValueConstructorVariant,
     },
 };
-use heck::{CamelCase, SnakeCase};
+use heck::SnakeCase;
 use itertools::Itertools;
 use std::char;
 use std::default::Default;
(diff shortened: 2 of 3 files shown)


lpil (approved):
Very nice, I like how clean this solution is. Thank you

PR #852, opened 2020-11-20
Prevent double arrow from showing up in documentation version selector on Android

michallepicki:
Before:
image

After:
image


lpil (approved):
CSS magic! Thanks

Issue #647, opened 2020-06-10
Package compiler API for use by rebar and mix


lpil:
```shell
gleam compile-package \
  --name='my_package' \
  --out='dir/to/write/to' \
  --src='dir/containing/app/code' \
  --test='dir/containing/test/code' \
  --lib='dir/containing/precompiled/gleam/library/output'
```

This API compiles a single Gleam package (i.e. an Erlang application / hex package). Rebar/other would call this API for each package, starting at the leaves of the dep tree and working up to the project root.

Once all deps are compiled then only the top level package needs to be compiled on future builds as deps don't change.

The `test` flag can be omitted for deps and non-test builds.

## TODO

- [x] Remove `gleam.toml` requirement
- [x] Package compiler
- [x] Enable writing of binary files
- [x] cap'n proto schema
- [ ] Serialise metadata to `modulename.gleam_module` cap'n proto files
  - [x] Typevar id handling 
- [ ] Deserialise metadata from `module-name.gleam_module` cap'n proto files
   - [ ] Typevar id handling
- [x] Command line command
  - [x] Accepts multiple lib directories
  - [ ] Reads `.gleam_module` files from `lib` directories
  - [x] Reads source code from `src` and `test`
  - [x] Runs `PackageCompiler`
  - [x] Writes Erlang to `out`
  - [ ] Writes metadata to `out`

lpil:
https://github.com/gleam-lang/gleam/pull/887

PR #853, opened 2020-11-22
CodeGenerator trait

lpil:


lpil:
Part of https://github.com/gleam-lang/gleam/issues/647 as it means code generation can be configured

PR #856, opened 2020-11-24
Fix focusing sidebar toggle on Safari

### LINKED ISSUE #855: Hamburger menu doesn't open on mobile safari
When I visit the hexdocs for a gleam library on mobile safari, the hamburger button on the upper left doesn't respond to clicks. This means I'm stuck at the initial readme page and unable to navigate the rest of the docs for a library.

michallepicki:
closes #855

This is just a quick fix to allow the menu to be opened on mobile Safari. This (probably) does not fix all aspects of the sidebar menu being broken on mobile Safari. I don't have an iOS device but from my testing, once opened the menu will not be easy to close. I think this is progress, users can refresh when stuck inside the menu. But a better fix would still be welcome - maybe it's possible to sprinkle some more attributes to make mobile Safari happy, or maybe the menu will have to be re-done using a checkbox and sibling CSS styles (or javascript, IMO if necessary).

### DIFF
--- a/templates/documentation_layout.html
+++ b/templates/documentation_layout.html
@@ -19,7 +19,7 @@
 
     <div class="page">
       <nav class="sidebar">
-        <a class="sidebar-toggle" href="#">
+        <a class="sidebar-toggle" href="#" tabindex="0">
           <svg
             aria-hidden="true"
             focusable="false"


lpil (approved):
Thank you! Let's give it a go :)

PR #859, opened 2020-11-25
Lowercase the repository type

michaeljones:
So that the config can have "github" and "bitbucket" instead of "GitHub"
and "BitBucket" as the latter feels less in the coding aesthetic.

(Again, very subjective.)


lpil (approved):
Much better, ta

PR #860, opened 2020-11-26
gleam format sorts imports alphabetically

itsgreggreg:
Not sure if this is something anyone but me wants but it was a low hanging fruit so I went for it.

`gleam format` will now sort imports alphabetically.

this:
```rust
// comment for b_module
import b_module
// two lines of
// comments for c_module
import c_module
// comment for a_module
import a_module
```

formats to:
```rust
// comment for a_module
import a_module
// comment for b_module
import b_module
// two lines of
// comments for c_module
import c_module
```

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -85,11 +85,11 @@ impl<'a> Formatter<'a> {
         for statement in module.statements.iter() {
             let start = statement.location().start;
             match statement {
-                Statement::Import { .. } => {
+                Statement::Import { module, .. } => {
                     has_imports = true;
                     let comments = self.pop_comments(start);
                     let statement = self.statement(statement);
-                    imports.push(commented(statement, comments))
+                    imports.push((module, commented(statement, comments)))
                 }
 
                 _other => {
@@ -101,7 +101,8 @@ impl<'a> Formatter<'a> {
             }
         }
 
-        let imports = concat(imports.into_iter().intersperse(line()));
+        imports.sort_by(|(mod_l, _), (mod_r, _)| mod_l.cmp(mod_r));
+        let imports = concat(imports.into_iter().map(|(_, doc)| doc).intersperse(line()));
         let declarations = concat(declarations.into_iter().intersperse(lines(2)));
 
         let sep = if has_imports && has_declarations {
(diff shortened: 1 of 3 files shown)


lpil (commented):
Looks good!

@CrowdHailer I remember you having concerns about import formatting. How does this look to you?

src/format.rs
```diff
@@ -85,11 +85,11 @@ impl<'a> Formatter<'a> {
         for statement in module.statements.iter() {
             let start = statement.location().start;
             match statement {
-                Statement::Import { .. } => {
+                Statement::Import { module, .. } => {
                     has_imports = true;
                     let comments = self.pop_comments(start);
                     let statement = self.statement(statement);
-                    imports.push(commented(statement, comments))
+                    imports.push((module.join(""), commented(statement, comments)))
```
lpil, line 92: Slices implement Ord (https://doc.rust-lang.org/std/slice/index.html#trait-implementations-1) so I think we can avoid joining the vec here, removing those allocations.

itsgreggreg: Did, not know that. Sweet!

CrowdHailer:
I wanted to be able to group imports somehow. we though separated by comments was a possible approach
e.g.
```
c
a
d

// comment
b
```

becomes
```
a
c
d

//coment
b
```

itsgreggreg:
I think you'd also need some kind of grouping directive. A new line or a comment doesn't indicate a group of any kind imo and I often write sloppily knowing the formatter is going to clean it up on save.

What would import groups be for?

lpil:
A newline with a comment below will be sufficient I think. That seems intentional enough.

CrowdHailer:
Suggestion, would we want the same functionality for large Structs. e.g.
```rs
let Request(
  host:
  scheme:
)
```

same for destructuring.

lpil:
We have to be careful about changing field order because the value expressions could have side effects, so the semantics of the program could change by reordering them.  Destructuring could work, though I'm not sure how I would want them rearranged yet

CrowdHailer:
Fair point, also thought this was issue not PR, my mistake. If interesting it should be a separate issue.

itsgreggreg:
Feel free to close this if it's not in line with where y'all wanna take it. I won't be offended!

lpil:
Thanks @itsgreggreg , sorry this has got a bit complicated.

PR #861, opened 2020-11-27
Mark source_url contents as safe

michaeljones:
Otherwise they are escaped by the template engine and whilst firefox and maybe other browsers still handles the output, the urls look broken in the source code.

Sorry I failed to notice this before. I guess it also raises questions if anything in a gleam module name or project name can be unsafe for html but I image we don't allow modules called `javascript: fetch("./secure/details")` (or whatever scripting hacks look like.)

### DIFF
--- a/templates/documentation_module.html
+++ b/templates/documentation_module.html
@@ -48,7 +48,7 @@
         </a>
       </h2>
       {% if !typ.source_url.is_empty() %}
-      <a href="{{ typ.source_url }}">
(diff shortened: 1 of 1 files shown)


lpil (approved):
Good catch! Hexdoc is an isolated domain, so I think this is safe even if the user finds a way to inject something naughty into the docs. Heck, they could always edit the HTML before uploading it so it doesn't really matter what we do here.

PR #881, opened 2020-12-13
WIP: Erlang typespec generation

michaeljones:
Submitting this for initial feedback. I'm fairly happy with some aspects of the state of the code but it is currently incomplete in so far as it fails to compile the standard library without warnings or errors. 

There are also definitely issues:
- The `type_to_erl` function is probably the wrong approach given that it is special casing things.
- My use of the `pretty` module could be a lot better
- It currently fails on the stdlib because it write out a `-type` for the Direction type in StringBuilder but that isn't used in the generated Erlang output so we get an warning/error.
- It doesn't take `opaque` into account though hopefully that isn't too big a deal.
- There are times when I'm just `to_snake_case` type names and not trying to make sure they are available in the scope. Seems like I'm missing something. Maybe not. 
- It feels like there are other issue lurking underneath
 
I guess I'm hoping to figure out if this exploration has unearthed issues that make it seem like just too much effort to carry on or if perhaps it seems hopeful enough to continue. Maybe a more experienced eye will see troubles in the depths or simple solutions for what seem like problematic hacks in the current approach.

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -1,5 +1,6 @@
 #[cfg(test)]
 mod tests;
+mod typespec;
 
 use crate::{
     ast::*,
@@ -37,16 +38,19 @@ pub fn generate_erlang(analysed: &[Analysed]) -> Vec<OutputFile> {
             .join(origin.dir_name());
         let erl_module_name = name.join("@");
 
+        let mut record_filenames: Vec<String> = Vec::new();
+
         for (name, text) in records(ast).into_iter() {
-            files.push(OutputFile {
-                path: gen_dir.join(format!("{}_{}.hrl", erl_module_name, name)),
-                text,
-            })
+            let path = gen_dir.join(format!("{}_{}.hrl", erl_module_name, name));
+            if let Some(filename) = path.to_str() {
+                record_filenames.push(filename.to_string());
+            }
+            files.push(OutputFile { path, text })
         }
 
         files.push(OutputFile {
             path: gen_dir.join(format!("{}.erl", erl_module_name)),
-            text: module(ast),
+            text: module(ast, &record_filenames),
         });
     }
 
@@ -102,6 +106,8 @@ pub fn records(module: &TypedModule) -> Vec<(&str, String)> {
             _ => &[],
         })
         .filter(|constructor| !constructor.args.is_empty())
+        // Extract record field names and filter out any constructors that have unnamed fields as
+        // they don't qualify as records
         .flat_map(|constructor| {
             let mut fields = Vec::with_capacity(constructor.args.len());
             for (label, ..) in constructor.args.iter() {
@@ -117,12 +123,7 @@ pub fn records(module: &TypedModule) -> Vec<(&str, String)> {
 }
 
 pub fn record_definition(name: &str, fields: &[&str]) -> String {
-    let name = &name.to_snake_case();
-    let escaped_name = if is_erlang_reserved_word(name) {
-        format!("'{}'", name)
-    } else {
-        name.to_string()
-    };
+    let escaped_name = typespec::type_name_to_atom(name);
     use std::fmt::Write;
     let mut buffer = format!("-record({}, {{", escaped_name);
     for field in fields.iter().intersperse(&", ") {
@@ -137,7 +138,7 @@ pub fn record_definition(name: &str, fields: &[&str]) -> String {
     buffer
 }
 
-pub fn module(module: &TypedModule) -> String {
+pub fn module(module: &TypedModule, import_filenames: &[String]) -> String {
     let module_name = module.name.as_slice();
     let exports = concat(
         module
@@ -164,6 +165,8 @@ pub fn module(module: &TypedModule) -> String {
             .intersperse(", ".to_doc()),
     );
 
+    let type_exports = typespec::type_exports(&module);
+
     let statements = concat(
         module
             .statements
@@ -172,10 +175,19 @@ pub fn module(module: &TypedModule) -> String {
             .intersperse(lines(2)),
     );
 
+    let imports = concat(
+        import_filenames
+            .iter()
+            .map(|filename| format!("-include(\"{}\").", filename).to_doc())
+            .intersperse(line()),
+    );
+
     format!("-module({}).", module_name.join("@"))
         .to_doc()
         .append(line())
         .append("-compile(no_auto_import).")
+        .append(lines(2))
+        .append(imports)
         .append(lines(2))
         .append(if exports == nil() {
             nil()
@@ -183,6 +195,15 @@ pub fn module(module: &TypedModule) -> String {
             "-export(["
                 .to_doc()
                 .append(exports)
+                .append("]).")
+                .append(lines(2))
+        })
+        .append(if type_exports == nil() {
+            nil()
+        } else {
+            "-export_type(["
+                .to_doc()
+                .append(type_exports)
                 .append("]).")
                 .append(lines(2))
         })
@@ -194,14 +215,27 @@ pub fn module(module: &TypedModule) -> String {
 fn statement(statement: &TypedStatement, module: &[String]) -> Option<Document> {
     match statement {
         Statement::TypeAlias { .. } => None,
-        Statement::CustomType { .. } => None,
         Statement::Import { .. } => None,
         Statement::ExternalType { .. } => None,
         Statement::ModuleConstant { .. } => None,
+        Statement::CustomType {
+            name,
+            parameters,
+            constructors,
+            ..
+        } => Some(typespec::type_(&name, &parameters, &constructors)),
 
         Statement::Fn {
-            args, name, body, ..
-        } => Some(mod_fun(name.as_ref(), args.as_slice(), body, module)),
+            args,
+            name,
+            body,
+            return_type,
+            ..
+        } => Some(
+            typespec::fun_spec(name.as_ref(), args.as_slice(), &*return_type)
+                .append(line())
+                .append(mod_fun(name.as_ref(), args.as_slice(), body, module)),
+        ),
 
         Statement::ExternalFn { public: false, .. } => None,
         Statement::ExternalFn {
@@ -209,13 +243,18 @@ fn statement(statement: &TypedStatement, module: &[String]) -> Option<Document> 
             module,
             args,
             name,
+            return_type,
             ..
-        } => Some(external_fun(
-            name.as_ref(),
-            module.as_ref(),
-            fun.as_ref(),
-            args.len(),
-        )),
+        } => Some(
+            typespec::external_fun_spec(name.as_ref(), args.as_slice(), &*return_type)
+                .append(line())
+                .append(external_fun(
+                    name.as_ref(),
+                    module.as_ref(),
+                    fun.as_ref(),
+                    args.len(),
+                )),
+        ),
     }
 }
 
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -219,7 +219,7 @@ pub fn module(module: &TypedModule, import_filenames: &[String]) -> String {
         .format(80)
 }
 
-fn statement(statement: &TypedStatement, module: &[String]) -> Option<Document> {
+fn statement(statement: &TypedStatement, current_module: &[String]) -> Option<Document> {
     match statement {
         Statement::TypeAlias { .. } => None,
         Statement::Import { .. } => None,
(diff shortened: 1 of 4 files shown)


michaeljones:
I've also not updated the tests which I guess would make this easier to assess. I can do so if needed. For the moment, here's the top of `gleam@queue.erl` to give a flavour:

```erlang
-module(gleam@queue).
-compile(no_auto_import).

-include("gen/src/gleam@queue_Queue.hrl").

-export([new/0, from_list/1, to_list/1, is_empty/1, length/1, push_back/2, push_front/2, pop_back/1, pop_front/1, reverse/1, is_logically_equal/3, is_equal/2]).

-export_type([queue/1]).

-type queue(Element) :: #queue{in :: list(Element), out :: list(Element)}.

-spec new() -> queue.
new() ->
    {queue, [], []}.

-spec from_list(list) -> queue.
from_list(List) ->
    {queue, List, []}.

-spec to_list(queue) -> list.
to_list(Queue) ->
    gleam@list:append(
        erlang:element(3, Queue),
        gleam@list:reverse(erlang:element(2, Queue))
    ).
```

src/erl.rs
```diff
@@ -1,8 +1,14 @@
...
+mod literals;
+pub mod names;
 #[cfg(test)]
 mod tests;
+mod typespec;
 
 use crate::{
     ast::*,
+    erl::format::wrap_args,
+    erl::literals::atom,
```
lpil, line 11: If possible I'd prefer to have refactoring the erl module into multiple modules to be a different PR. This one is gunna be pretty large just with the feature so to make it easier to review and merge it'd be good to try to avoid including other changes.

src/erl.rs
```diff
@@ -15,11 +21,10 @@ use heck::SnakeCase;
 use itertools::Itertools;
 use std::char;
 use std::default::Default;
+use std::path::Path;
 use std::sync::Arc;
 
-const INDENT: isize = 4;
-
-pub fn generate_erlang(analysed: &[Analysed]) -> Vec<OutputFile> {
+pub fn generate_erlang(analysed: &[Analysed], project_path: impl AsRef<Path>) -> Vec<OutputFile> {
```
lpil, line 27: This function is being removed soon (possibly the next release) as it's part of the original build system which is rather messy.

The new place to look for code generation is the `codegen` modules. Currently the codegen for record headers and modules are two different code generators, so it'll be difficult to use one from the other like this.

The reason for them being split is that different tools may want the headers in different directories, or not at all in the case of the Gleam compiler up until now.

I think it would be easier to show tuples rather than the record sugar. We will need to implement the tuple syntax for records without labelled fields so we could use them always rather than implement records additionally and then having logic to select between the two.

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+            .flat_map(|s| match s {
+                Statement::CustomType {
+                    public: true,
+                    name,
+                    constructors,
+                    ..
+                } => {
+                    // We only count the params that we can find in the constructors as we can only
+                    // put those ones into the type definition as otherwise we'll have unused type
+                    // param warnings
```
lpil, line 21: I'm not sure what this means?

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+pub fn type_name_to_atom(name: &str) -> String {
+    let name = &name.to_snake_case();
+    if names::is_erlang_reserved_word(name) {
+        format!("'{}'", name)
+    } else {
+        name.to_string()
+    }
+}
+
+/* Generate an erlang '-type' line for this custom type. There are three options for entries:
```
lpil, line 42: Nice docs! Rust doc comments start with `///`, so let's add that so it shows up in rendered docs and IDEs

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+
+fn is_record(constructor: &RecordConstructor) -> bool {
+    for (label, ..) in constructor.args.iter() {
+        match label {
+            Some(_) => return true,
+            None => return false,
+        }
+    }
+
+    false
```
lpil, line 121: Maybe this?

```suggestion
    constructor.args.iter().all(|(label, ..)| label.is_some())
```

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+        .append(".")
+}
+
+fn spec_args(args: &[TypedArg]) -> Document {
+    wrap_args(args.iter().map(|a| typ_to_erl(&*a.typ)))
+}
+
+fn typ_to_erl(typ: &Type) -> Document {
+    match typ {
+        Type::App { name, .. } => name.to_snake_case().to_doc(),
```
lpil, line 157: We will need to namespace this type if it comes from another module

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+        // limited possible type parameter names and if we come across one that isn't 'allowed'
+        // then we're better just outputting 'any' instead. This prevents unknown or unused
+        // type parameter warnings from erlang
+        {
+            allowed_params
+                .map(|slice| {
+                    if slice.contains(name) {
+                        name.to_camel_case()
+                    } else {
+                        "any".to_string()
```
lpil, line 198: How does this happen? Shouldn't all type variables be declared beforehand?

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+                .to_doc()
+        }
+        TypeAst::Constructor { name, args, .. } => type_to_erl(name, &args, allowed_params),
+        _ => "term".to_doc(),
+    }
+}
+
+fn type_to_erl(name: &str, args: &[TypeAst], allowed_params: Option<&[String]>) -> Document {
+    match name {
+        "String" => "binary()".to_doc(),
```
lpil, line 211: We're going to have to generate Erlang types from Gleam types as there's not enough information in the un-typed AST.

For example this code will cause the generation of incorrect Erlang

```rust
pub type Bool {
  Maybe
}

File: src/erl/typespec.rs [hunk 9]
```diff
@@ -0,0 +1,263 @@
...
+fn extract_params_from_constructors(constructors: &[RecordConstructor]) -> Vec<String> {
+    let mut set = std::collections::HashSet::new();
+    constructors
+        .iter()
+        .flat_map(|constructor| {
+            constructor
+                .args
+                .iter()
+                .flat_map(|(_, ast, ..)| extract_params_from_ast(&ast))
+                .collect::<Vec<String>>()
```
lpil, line 241: Would be nice to allocate fewer vectors. Perhaps we could pass a mutable reference to a vector into this function.

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+}
+
+/* Extraction functions - There are various places where we want to make sure that we're only
+   outputting type parameters if we know they are going to be used in order to avoid unused type
+   parameter warnings. Consequently we have these functions to help us walk the type trees and find
+   what we need.
+*/
+fn extract_params_from_constructors(constructors: &[RecordConstructor]) -> Vec<String> {
+    let mut set = std::collections::HashSet::new();
```
lpil, line 233: What is this set for? Should it be returned?

src/erl/typespec.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+fn extract_params_from_ast(ast: &TypeAst) -> Vec<String> {
+    match ast {
+        TypeAst::Var { name, .. } => vec![name.to_camel_case()],
+        TypeAst::Constructor { args, .. } => args
+            .iter()
+            .flat_map(|ast| extract_params_from_ast(&ast))
+            .collect::<Vec<String>>(),
+        _ => Vec::new(),
```
lpil, line 254: As with the other, would be nice to allocate less here

lpil (commented):
Wow you've been busy! Looks like a good start. I've left some comments inline

src/erl.rs
michaeljones: I completely agree. I didn't want to be insubordinate :) I assumed that Rust didn't support cyclic imports with `erl` importing `erl::typespec` and `erl::typespec` importing `erl`. I didn't check though. If Rust supports them then I'll role them back into `erl` happily enough.

michaeljones:
Thanks for the review! I'll try to work through some of the suggestions this week. They all seem very reasonable though my understanding of the bigger picture isn't great so I might need some more detail in places.

src/erl.rs
lpil: Oh I see! I think it probably doesn't allow that? Maybe it does! Let's find out

lpil:
Thank you @michaeljones !

src/erl/typespec.rs
michaeljones: It is a bit confusing and needs a comment but it was my attempt at a deduplicating the vector without sorting/reordering (which the standard `vec.dedup` method requires.) It is used in the `.filter(|name| set.insert(name.clone()))` below. It seems that `HashSet.insert` returns `true` if the item was not in the set and `false` if it was so I think it works as a filter predicate. There might well be better ways to approach it though.

michaeljones: Hmm, thinking about this I suspect I'm trying to be careful with the wrong set of params. This 'collection' stage (if it is even needed) doesn't need to be order preserved. I'll try to explain my reasoning for all this with examples when I get a bit of time. It is either required due to certain complexities or possibly just a stop gap to cover an incomplete type conversion process.

src/erl.rs
michaeljones: It seems that Rust doesn't mind having the cross dependencies which weirds me out a bit but it means we can leave everything where it was.

src/erl/typespec.rs
michaeljones: Nice! Though `.all()` returns true for an empty iterator so we might need:
```rust
    (!constructor.args.is_empty()) && constructor.args.iter().all(|(label, ..)| label.is_some())
```
I'll change it to that.

src/erl/typespec.rs
michaeljones: Cool!

src/erl/typespec.rs
michaeljones: Good question, the clearest example would be:
```
pub external fn from(a) -> Dynamic =
  "gleam_stdlib" "identity"
```
If we generate:
```erlang
-spec from(A) -> dynamic.
```
Then we get an error saying that A is unused I believe. Instead we have to say:
```erlang
-spec from(any) -> dynamic.
```
And we can figure out which type parameters to change to `any`s by inspecting (in this case) the return type and figuring out how many type parameters are used in it and replacing any others with `any` in the spec. 

I came across this with types too. Possibly just due to lazy type generation. I don't have code that handled function types for example so if a type parameter is used in a function type in a type definition then it isn't going to be output so we don't want to include it initially. I realise that "laziness" is not a good excuse here but it was done in order to get something imperfect out to see if the direction made sense rather than trying to work through all the possible cases.

src/erl/typespec.rs
michaeljones: See the comment above about generating for `dynamic.from`. If the variable isn't used in the return type then it is considered 'unbound' and is better represented with an `any`. There are some legitimate places for this and some less legitimate places.

src/erl.rs
michaeljones: I think I follow this and can come around to resolving it at some point. This relies on records really just being tagged tuples so we can output the description of the tagged tuple directly?

src/erl/typespec.rs
michaeljones: I've attempted to do this now though I don't know how good an approach it is.

src/erl/typespec.rs
michaeljones: I can see that's an issue. I'm not sure exactly what the answer is though.

src/erl/typespec.rs
michaeljones: Certainly understand this is less than desirable. Happy to come back to it if we get the branch into an otherwise functioning state.

src/erl/typespec.rs
michaeljones: Actually for the use of it in `typespec::type_`, we want to be deduplicated and order preserved, I think. Might be better ways to handle it all though.

michaeljones:
I've added a few more commits. One of them demonstrates an attempt to find all the used types in a module before generating the erlang output. This is for situations like the `Direction` types in `String` and `StringBuilder` in the stdlib which are only used private external functions and so don't appear in the final generated erlang. 

Unfortunately, I'm not sure how best to go about extracting that information. I've shown an approach but I'm not sure of the best way to apply it to `ExternalFnArg` values which don't expose a `Type` like `TypedArg` does. Perhaps it is enough to use the `TypeAst` though I'm not sure exactly how the `Constructor` entry in `TypeAst` maps to a regular type? 

I would welcome input & guidance on the general approach when you have time.

michaeljones:
I've run out of energy for this for the moment. I hope it looks like a good base to take forward either by myself or someone else at some point but I think I need to build up more knowledge of the types and how they are represented before I'll be able to confidently proceed with this. I'll close it for the moment and we can see where we get to.

lpil:
Thanks @michaeljones ! This was a monster undertaking and the quality of the work is great. It will make a good foundation for future work.

I will link this to the ticket for people to find.

Issue #837, opened 2020-11-07
Code cannot refer to consts that are defined later in the file


edkelly303:
In Gleam 0.12.0:
```gleam
fn something() -> String {
  hello
}

const hello: String = "Hello"
```
Gives an error:
```
error: Unknown variable
  ┌─ /whatever/src/gleam_bugs.gleam:2:3
  │
2 │   hello
  │   ^^^^^ did you mean `Nil`?

The name `hello` is not in scope here.
```
Honestly, for readability's sake, I think I probably *should* be defining my constants before I use them - but given we can now write functions and types in an arbitrary order, perhaps the same should be true of constants?

One option might be to issue be a compiler warning, rather than an error, if you use a constant before it has been defined?

lpil:
This one will be trickier as we need to know the values of constants before they are used in other constants. We can however ensure they can be used in any functions.

PR #886, opened 2020-12-24
Infer constants before other statements

itsgreggreg:
Not sure this is the preferred way to get this done. Happy to change it, or feel free to close this if it's off base!

This:
```rust
fn x() {
  a
}

fn z() {
  b + c
}

const a = 1
const b = 2
const c = 3
```

now compiles to this:
```erlang
x() ->
    1.

z() ->
    2 + 3.
```
fixes: #837


lpil (approved):
Wonderful, thank you! And merry christmas!

Issue #876, opened 2020-12-08
Warning about unused literals


jakobht:
The parser will parse `2-1` as `2; -1`.
I understand that it might be hard to change the formatter to parse it as `2 - 1` so I suggest to add a warning about `2` being unused.

lpil:
Good idea! Let's emit this warning

brightly-salty:
Maybe this should be closed? Seems like both PR#891 and PR#893 have been merged

lpil:
The 1-1 issue has been solved but unused literals need to emit a warning still.

To implement this we would need code similar to this code for discarded results: 

https://github.com/gleam-lang/gleam/blob/9b895b80f2c0dcf3544cd6f17c237895c32a9a72/src/typ/expr.rs#L341-L355

The main difference would be that we would be that we would have a check along the lines of `if first.is_literal() { return Err(Error::UnusedLiteral { .. }) }`

brightly-salty:
Would that be in the same infer_seq method? If so, I think I would be willing to attempt this.

lpil:
Yes, that would be the right place to do it. A contribution would be wonderful, thank you

PR #891, opened 2020-12-28
Alternative lexer and parser

itsgreggreg:
💝 Happy Holidays all!

I'm posting this here as it seems like a low noise / low attention way to show and tell.
Marking it as draft as it seems rather presumptuous not to, but it's essentially done and ready for review.

My intention here is really just to show off something I've been working on, not trying to cram it into the codebase if its not wanted!

## What
✨🤖✨
 A new, handwritten, complete lexer and parser for Gleam. It uses the same input
grammar and outputs the same AST so nothing else has changed.

I've tested it on the standard lib, gleam_todo, and a couple other smaller projects.

All tests passing.

## How to check it out
I've hooked it up to a new command called `compile` that works exactly like `build` 
with the only change being that it uses the alt lexer/parser and calls no lalrpop code.

So to check it out and compare:
* clone the repo: `git clone https://github.com/itsgreggreg/gleam/`
* checkout the branch: `cd gleam && git checkout lex_and_parse`
* compile for release: `cargo build --release`

Then you can switch between this parser and the lalrpop parser:
* this parser: `..path_to_repo/target/release/gleam compile`
* lalrpop: `..path_to_repo/target/release/gleam build`

The formatter is also hooked up to use this parser so you can check that out too.
This parser shaves a little off the start up time so it's quite snappy.

### For Me
* I had some time off and I like writing parsers
* To understand Gleam better

### For Gleam
* 🚨 **Much** better error messages (See below)
* 💁‍♀️Easier to add helpful lex and parse error messages in nuanced situations
* 🚀 Gleam project compilation is about 45% faster
* 🏇 Startup time is a little faster making formatting and compiling small projects snappier.
* 🤖 No more needing to blank out comments, the lexer skips them
  * (moving doc comment parsing to the lexer/parser would prob speed the compiler up another 5-10%)
* 🧞 Adding support for things like: "2-1" parsed as "2 - 1" shouln't be too hard: https://github.com/gleam-lang/gleam/issues/876

## What's left to do
* Perhaps testing large projects for 1-1 compilation with lalrpop parser
* Perhaps more perf testing

## What could improve
* Could make the lexer/parser assemble doc comments
* The code size can probably be reduced by either coming up with new combinators or recognizing more syntax patterns that can be parsed together
* Some algorithms could look nicer
* Perhaps some macros could clean up a lot of code
* Impl-ing some functions on `Spanned` could shorten some of the matching
* The code can probably be more rust-y
* The lexer could probably be more concise
* The lexer could emit nesting errors where they occur

## Output Testing
Not exactly comprehensive but I've run it against the standard library and my own projects and found that there are no differences in the generated erlang.

## Perf Testing
Not very scientific but I alternate running `time gleam compile` and `time gleam build`. I've found the new parser to be fairly consistently ~ 45% faster.
I've only run it on my 2015 MBP.

## Other Notes
The lexer is driven by the parser. It only consumes enough of a src file to stay a few characters ahead of what the parser needs to do its job. This means that if there is a parse error, only a little more than is needed to reach that point is ever read.

The lexer is consumed on parsing and strings become owned by the AST, no needless cloning.

The parser module may look long at ~2.5k lines (about twice the lalrpop grammar)
but that's nothing compared to the lalrpop generated parser at ~72k lines. 🤠

It's more comparable to stage 1 of the zig parser: https://github.com/ziglang/zig/blob/master/src/stage1/parser.cpp
Or just the expr portion of the rust parser: https://github.com/rust-lang/rust/blob/master/compiler/rustc_parse/src/parser/expr.rs

### Using ( ) as expression grouping

Before:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:29:17
   │
29 │     let b = 1 + (4 / 2)
   │                 ^ Unexpected token

Expected one of "<<", "[", "[]", "case", "fn(", "todo", "todo(", "tuple", "{", r#"\"(([^\\\\\"]|\\\\.)*)\""#, r#"-?[0-9](_*[0-9])*\\.[0-9]*"#, r#"-[0-9](_*[0-9])*"#, r#"0b[0-1](_*[0-1])*"#, r#"0o[0-7](_*[0-7])*"#, r#"0x[0-9A-Fa-f](_*[0-9A-Fa-f])*"#, r#"[0-9](_*[0-9])*"#, r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax errorr
   ┌─ /Users/a/parser_test/src/parser_test.gleam:29:17
   │
29 │     let b = 1 + (4 / 2)
   │                 ^ This paren cannot be understood here.

Hint: To group expressions in gleam use "{" and "}"
```

### Trying to import a reserved word

Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/one.gleam:1:8
  │
1 │ import todo/mod
  │        ^^^^ Unexpected token

Expected one of r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/one.gleam:1:8
  │
1 │ import todo/mod
  │        ^^^^ This is a reserved word.

Hint: I was expecting to see a name here.
See: https://gleam.run/book/tour/reserved-words
```

### Using a type name in place of a module

Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:8
  │
1 │ import Other
  │        ^^^^^ Unexpected token

Expected one of r#"[a-z][0-9a-z_]*"#
```

Now: 
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:8
  │
1 │ import Other
  │        ^^^^^ I'm expecting a lowercase name here.

Hint: Variable and module names start with a lowercase letter, and can contain a-z, 0-9, or _.
```

### Extra separator anywhere
Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:23
  │
1 │ const cc: tuple(a, b, ,) = 1
  │                       ^ Unexpected token

Expected one of ")", "fn(", "tuple", r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#, r#"_([a-z][0-9a-z_]*)?"#
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:23
  │
1 │ const cc: tuple(a, b, ,) = 1
  │                       ^ This is an extra separator.

Hint: Try removing it?
```

Or:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:13:38
   │
13 │ pub external fn a(name: String, Int, ,) -> Bool =
   │                                      ^ This is an extra delimiter.

Hint: Try removing it?
```
etc.

### Function in module constant type

Before:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:10:49
   │
10 │ const c: fn() -> String = fn(){ "hi" }
   │                           ^^^ Unexpected token

Expected one of ")", "<<", "[", "]", "tuple", r#"\"(([^\\\\\"]|\\\\.)*)\""#, r#"-?[0-9](_*[0-9])*\\.[0-9]*"#, r#"-[0-9](_*[0-9])*"#, r#"0b[0-1](_*[0-1])*"#, r#"0o[0-7](_*[0-7])*"#, r#"0x[0-9A-Fa-f](_*[0-9A-Fa-f])*"#, r#"[0-9](_*[0-9])*"#, r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:9
  │
1 │ const c:fn() -> String = fn(){ "Hi" }
  │         ^^ Functions are not allowed in module constants.

See: https://gleam.run/book/tour/constants
```

Or:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:9
  │
1 │ const c = fn(){ "Hi" }
  │           ^^ Functions are not allowed in module constants.

See: https://gleam.run/book/tour/constants
```

### ListNil used in const value

Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:22
  │
1 │ const c: List(Int) = []
  │                      ^^ Unexpected token

Expected one of "<<", "[", "tuple", r#"\"(([^\\\\\"]|\\\\.)*)\""#, r#"-?[0-9](_*[0-9])*\\.[0-9]*"#, r#"-[0-9](_*[0-9])*"#, r#"0b[0-1](_*[0-1])*"#, r#"0o[0-7](_*[0-7])*"#, r#"0x[0-9A-Fa-f](_*[0-9A-Fa-f])*"#, r#"[0-9](_*[0-9])*"#, r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:22
  │
1 │ const c: List(Int) = []
  │                      ^^ Empty list is not allowed here.
```

### Custom type with no constructors

Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:6:13
  │
6 │ pub type A {}
  │             ^ Unexpected token

Expected one of r#"[A-Z][0-9A-Za-z]*"#
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/parser_test.gleam:6:10
  │
6 │ pub type A {}
  │          ^ Custom types must have at least 1 constructor.

See: https://gleam.run/book/tour/custom-types
```

### Unterminated strings

Before:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:16:11
   │
16 │   let a = "hello
   │           ^ Unknown token

I don't know what this character means. Is it a typo?
```

Now:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:16:11
   │
16 │   let a = "hello
   │           ^ The string starting here was left open.
```

### Operator with a naked right side

Before:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:17:1
   │
17 │ }
   │ ^ Unexpected token

Expected one of "<<", "[", "[]", "case", "fn(", "todo", "todo(", "tuple", "{", r#"\"(([^\\\\\"]|\\\\.)*)\""#, r#"-?[0-9](_*[0-9])*\\.[0-9]*"#, r#"-[0-9](_*[0-9])*"#, r#"0b[0-1](_*[0-1])*"#, r#"0o[0-7](_*[0-7])*"#, r#"0x[0-9A-Fa-f](_*[0-9A-Fa-f])*"#, r#"[0-9](_*[0-9])*"#, r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:16:17
   │
16 │     1 + 10 / 20 *
   │                 ^ This operator has no value on its right side.

Hint: Remove it or put a value after it.
```

### Empty expression group

Before:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:29:14
   │
29 │     let b = {}
   │              ^ Unexpected token

Expected one of "<<", "[", "[]", "assert", "case", "fn(", "let", "todo", "todo(", "try", "tuple", "{", r#"\"(([^\\\\\"]|\\\\.)*)\""#, r#"-?[0-9](_*[0-9])*\\.[0-9]*"#, r#"-[0-9](_*[0-9])*"#, r#"0b[0-1](_*[0-1])*"#, r#"0o[0-7](_*[0-7])*"#, r#"0x[0-9A-Fa-f](_*[0-9A-Fa-f])*"#, r#"[0-9](_*[0-9])*"#, r#"[A-Z][0-9A-Za-z]*"#, r#"[a-z][0-9a-z_]*"#
```

Now:
```
error: Syntax error
   ┌─ /Users/a/parser_test/src/parser_test.gleam:29:13
   │
29 │     let b = {}
   │             ^^ There must be an expression in here.

Hint: Put an expression in there or remove the brackets.
```

### Opaque type alias
Before:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/one.gleam:1:19
  │
1 │ pub opaque type A = String
  │                   ^ Unexpected token

Expected one of "{"
```

Now:
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/one.gleam:1:1
  │
1 │ pub opaque type A = String
  │ ^^^^^^^^^^^^^^^^^ Type Aliases cannot be opaque

See: https://gleam.run/book/tour/type-aliases
```

### Compiler crash on tuple access with underscore
Before:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit }', /Users/runner/work/gleam/gleam/target/x86_64-apple-darwin/release/build/gleam-420d74bfb76de550/out/grammar.rs:54753:33
```

Now (before also crashes on a valid index):
```

error: Out of bounds tuple index
  ┌─ /Users/a/parser_test/src/one.gleam:4:13
  │
4 │   tuple(1,2).1_000
  │             ^^^^^^ this index is too large

The index being accessed for this tuple is 1000, but this tuple has
2 elements so the highest valid index is 1.
```

### Unknown Import highlight (fixes just the underline)

Before: 
```
error: Unknown import
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:8
  │
1 │ import bother as b
  │        ^^^^^^^ did you mean `other`?
```

Now:
```
error: Unknown import
  ┌─ /Users/a/parser_test/src/parser_test.gleam:1:8
  │
1 │ import bother as b
  │        ^^^^^^ did you mean `other`?
```


lpil (commented):
This is an absolutely beautiful piece of work and I am so impressed.

Normally I get a bit fluffed when someone does a big bit of work without discussion first as normally it's difficult to accept it, but the quality is excellent here, and I have been want to rewrite the parser for some time.

I am still in the process of reading all the code, but so far I can't fault it at all and most likely I would like to merge this in unless I find any issues in the rest of my review.

## Nitpicks:

It would be nice to have a comment at the top that gave a quick overview of the algorithm (especially if there is any references the viewer can read), as well as detailing the naming convention. i.e. `parse_*` vs `maybe_*` vs `expect_*`.

I think the formatter has got confused in the error.rs file due to overly long lines, and as such isn't formatting the file any more. I can fix this later.

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+        handle_op(None, &mut opstack, &mut estack, &do_reduce_expression)
+    }
+
+    // examples:
+    //   1
+    //   "one"
+    //   True
+    //   fn() { "hi" }
+    //   unit().unit().unit()
+    //   A(.., tuple(1))
```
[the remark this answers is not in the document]
itsgreggreg: Whoops a double error. That example is supposed to be for a pattern and the `..` is in the wrong places.  Should be `A(tuple(1), ..)`.

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+
+    // Expect a Name else a token dependent helpful error
+    fn expect_name(&mut self) -> Result<(usize, String, usize), ParseError> {
+        let t = self.next_tok();
+        match t {
+            Some((start, tok, end)) => {
+                if let Tok::Name { name } = tok {
+                    Ok((start, name, end))
+                } else if let Tok::UpName { .. } = tok {
+                    parse_error(ParseErrorType::IncorrectName, SrcSpan { start, end })
```
lpil, line 2103: This is really nice!

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+                    return parse_error(ParseErrorType::ExtraSeparator, SrcSpan { start, end });
+                }
+            }
+        }
+
+        Ok(results)
+    }
+
+    // If next token is a Name, consume it and return relevant info, otherwise, return none
+    // don't error in here. Result type is for combinator ease
```
lpil, line 2185: " Result type is for combinator ease"

When is this useful? From what I can see this is often called with `?` which always unwraps the `Ok`. What am I missing?

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+    }
+}
+// Simple-Op-Parser, perform reduction for expression
+fn do_reduce_expression(op: Spanned, estack: &mut Vec<UntypedExpr>) {
+    match (estack.pop(), estack.pop()) {
+        (Some(er), Some(el)) => {
+            let new_e = expr_op_reduction(op, el, er);
+            estack.push(new_e);
+        }
+        _ => panic!(parser_panic_text("Tried to reduce withot 2 expressions")),
```
lpil, line 2361: Can we swap these panics for `gleam::error::fatal_compiler_bug`? It prints a nice error message to the user.

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+}
+
+fn bit_string_const_int(value: String, start: usize, end: usize) -> UntypedConstant {
+    Constant::Int {
+        location: SrcSpan { start, end },
+        value,
+    }
+}
+
+// Question: parse error here?
```
lpil, line 2526: Yes I think so. Removing `Invalid` is on my TODO list

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+
+//
+// Misc Helpers
+//
+
+// When expecting a token and we get something else, this lookup is used
+// to show the user what we were looking for
+// could move this to a Display impl?
+fn what_was_expected(t: &Tok) -> String {
+    let todo = format!("TODO {:?}", t);
```
lpil, line 2568: TODO?

src/parse.rs
```diff
@@ -0,0 +1,2602 @@
...
+    // Parse Expressions
+    //
+
+    // expmples:
+    //   unit
+    //   unit op unit
+    //   unit op unit pipe unit(call)
+    //   unit op unit pipe unit(call) pipe unit(call)
+    fn parse_expression(&mut self) -> Result<Option<UntypedExpr>, ParseError> {
+        // uses the simple operator parser algorithm
```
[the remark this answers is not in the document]
itsgreggreg: Realizing I called it Simple Operator Parser in the code but it's actually called "Simple Precedence Parser".
Here's a somewhat dense explanation on wikipedia https://en.wikipedia.org/wiki/Simple_precedence_parser

Essentially, it's a shift-reduce, bottom-up parser with the only goal of correctly applying precedence.
I use 2 stacks instead of the 1 because I think it's easier to reason about and type.
A stack for storing operators that have been seen but not yet reduced `opstack`
A stack for storing operands (expressions) `estack` which may be bare values like "1" or "0.5" or sub expressions like "1 * 5"

While reading the input stream: 
if you see an operand put it at the top of the estack
if you see and operator compare it to to the top of the opstack and do some stuff
when you get to the end of the input try to reduce the rest of the operators and expressions

Happy to go over the algorithm in more detail if you want.
Lemme know what kind of documentation / examples we want to put into the source!

src/parse/lexer.rs
```diff
@@ -0,0 +1,725 @@
...
+use crate::parse::token::Tok;
+use std::char;
+
+// ideas:
+// * look up 2 / 3 char tokens like keywords
+//   * perf test
+// * track nesting per type and lex error on invalid nesting
+//   * useful?
+
+pub struct Lexer<T: Iterator<Item = (usize, char)>> {
```
lpil, line 11: Does this work with unicode? In the hacky comments stripper we iterated across unicode grapheme clusters

src/parse/lexer.rs
```diff
@@ -0,0 +1,725 @@
+use crate::parse::error::{LexicalError, LexicalErrorType};
+use crate::parse::token::Tok;
+use std::char;
+
+// ideas:
```
lpil, line 5: Is this comment still relevent?

src/parse.rs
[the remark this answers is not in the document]
itsgreggreg: This is actually a great video explanation: https://www.youtube.com/watch?v=n5UWAaw_byw&t=648s

src/parse/error.rs
```diff
@@ -0,0 +1,73 @@
...
+
+impl LexicalError {
+    pub fn to_parse_error_info(&self) -> (&str, Vec<String>) {
+        match self.error {
+            LexicalErrorType::UnexpectedStringEnd => {
+                ("The string starting here was left open.", vec![])
+            }
+            LexicalErrorType::NestingLeftOpen => (
+                "A bracket of some kind was left open.",
+                vec!["Hint: Close it :)".to_string()],
```
lpil, line 61:
```suggestion
                vec!["Hint: You may need to add a closing bracket.".to_string()],
```

src/parse/lexer.rs
```diff
@@ -0,0 +1,725 @@
...
+}
+
+impl<T> Lexer<T>
+where
+    T: Iterator<Item = (usize, char)>,
+{
+    pub fn new(input: T) -> Self {
+        let mut lxr = Lexer {
+            chars: input,
+            nesting: 0,
```
lpil, line 111: Why is it that nesting is tracked in the lexer rather than the parser?

lpil (commented):
It is possible to swap out the ` &dyn Fn` use for `impl Fn`? From what I can see it looks like it doesn't need to be dynamic, and removing the dynamic dispatch would make it a tiny bit faster.

src/parse.rs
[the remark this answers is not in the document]
lpil: That's great, thank you. Some of this info in the comment would be fab!

src/parse.rs
itsgreggreg: the `maybe_` functions that get passed to `series_of` return a `Result<Option<>>` rather than just an `Option` as `series_of` requires it. `maybe_binding_start` and `maybe_one` return simply an `Option` as they never get passed to `series_of`. Happy to re-work this. What do you think?

src/parse.rs
itsgreggreg: Now:
```
error: Syntax error
  ┌─ /Users/a/Trash/parser_test/src/one.gleam:6:7
  │
6 │   <<a:invalid>>
  │       ^^^^^^^ This is not a valid BitString segment option.

Hint: Valid BitString segment options are:
binary, int, float, bit_string, utf8, utf16, utf32, utf8_codepoint, utf16_codepoint
utf32_codepoint, signed, unsigned, big, little, native, size, unit
See: https://gleam.run/book/tour/bit-strings
```

src/parse/lexer.rs
itsgreggreg: Yeah, the lexer does not need to recognize any grapheme longer than a single utf-8 byte and so if it encounters the start of a character that is longer than a byte and is:
In comment mode, it will keep consuming bytes til newline
In string mode, it will keep consuming bytes til closing quote
Not in string or comment mode, it will see it as an unrecognized token.

If it is seen as an unrecognized token Codespan is smart enough to see that the byte pointed to is a multi-byte char, and will correctly highlight it. Like this (looks better in the terminal):
```
error: Syntax error
  ┌─ /Users/a/parser_test/src/one.gleam:6:22
  │
6 │   let a = "🤠🐄 howdy" 🤠🐄
  │                        ^^ I can't figure out what to do with this character.

Hint: Is it a typo?
```

The comment blanker needs no know how many bytes a grapheme takes up because it pushes spaces (" ") to a buffer for each "character" it consumes. When it consumes a multibyte grapheme it would end up pushing the wrong number of spaces to the buffer and file locations would be off.

I actually pushed that change a few months ago. It probably could have been solved better back then, I just noticed that emoji in comments caused error messages to be off and tried switching to graphemes and it worked. :)

src/parse/lexer.rs
itsgreggreg: Good question. The lexer is so straight forward I'm not positive "look up 2 / 3 char tokens like keywords" would buy much but would pull the code size down considerably if it worked.

"track nesting per type and lex error on invalid nesting" Is about things like:

```
fn a() {
  ...
  )
}
```
and
```
fn a() {
  ...
  (
}
```

The lexer could keep track of where nestings were opened and closed rather easily on a stack. Every time a nesting is opened Its type and location are added to the stack, every time a nesting is closed it must complement the top of the stack. We could then emit a syntax error that pointed to the offending opening, closing, or both.

I did not fully implement this as I wasn't sure it was useful, though the lexer currently checks that there is at least _some_ nesting to close when you try, and makes sure that you don't end the module with any nestings open.

src/parse/error.rs
itsgreggreg: This could be either a paren, curly, or square bracket. Do you think the language "bracket" is general enough?

src/parse/lexer.rs
itsgreggreg: My idea was that since in the grammar you must always close inner before outer nestings, it would be easier to point out the error precisely in the lexer. See this response for a little more detail on use: https://github.com/gleam-lang/gleam/pull/891#discussion_r549553214

It could be done in the Parser, you could scan tokens as they come out of the lexer in `next_tok` and accomplish the same thing. I think my mind felt it was easier at the lexing phase. Probably just a working thought thing.

src/parse.rs
itsgreggreg: moved to a Display impl for Tok!

src/parse.rs
lpil: Oh yes. I think I would slightly prefer to have them wrapped at the call site.

src/parse.rs
lpil: Much better

src/parse/lexer.rs
lpil: Nice, thanks

src/parse/lexer.rs
lpil: I could see the error messages produced by stack being useful, but I'm not sure about the current bracket counting system as it seems a lot more limited.￼ Perhaps we could remove the bracket counting and add the stack approach in future if we deem that to be worth investing the time in?￼

src/parse/error.rs
lpil: That's a good point. Maybe we just don't have a hint for this one.￼

lpil:
Looking fab! Is there more you would like to do before this is ready? Perhaps remove the old parser?

itsgreggreg:
Good question, totally up to you.
The biggest thing left to do is to unhook lalrpop and hookup this parser everywhere then removing the `compile` command.  Happy to do that with your go ahead. I don't personally have any more functionality I'd like to add in this PR but of course very open to getting done what others want to get done. I'm in no rush.
Is this the sort of thing that should be mentioned in the changelog?

lpil:
If you could remove lalrpop that would be great!

I think it's worth noting in the changelog as the error messages and performance improvement are things that impact users.

src/error.rs
```diff
@@ -1312,6 +1319,150 @@ When matching you need to use the `{}_codepoint` specifier instead.",
...
+                    ParseErrorType::NotConstType => (
+                        "This type is not allowed in module constants.",
+                        vec!["See: https://gleam.run/book/tour/constants".to_string()]
+                    ),
+                    ParseErrorType::NoExpression=> (
+                        "There must be an expression in here.",
+                        vec!["Hint: Put an expression in there or remove the brackets.".to_string()]
+                    ),
+                    ParseErrorType::NoValueAfterEqual => (
+                        "I was expectin to see a value after this equals sign.",
```
sneako, line 1407:
```suggestion
                        "I was expecting to see a value after this equals sign.",
```

itsgreggreg:
@lpil  lalrpop removed in this commit: https://github.com/gleam-lang/gleam/pull/891/commits/2a1c8f29d6344fb2223f788e38b6cc2a02640634

This leaves build.rs/main() as a function with an empty body: https://github.com/gleam-lang/gleam/pull/891/commits/2a1c8f29d6344fb2223f788e38b6cc2a02640634#diff-d0d98998092552a1d3259338c2c71e118a5b8343dd4703c0c7f552ada7f9cb42L1

Anything in particular you want to do with that, or just leave it for now?

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@
 - Fixed a bug where the formatter would incorrectly remove blocks around some
   binary operators.
 - Constants can now be defined after they are used in functions
+- Error messages have been improoved.
+- Compilation times have been improoved.
```
michallepicki, line 22:
```suggestion
- Error messages have been improved.
- Compilation times have been improved.
```

lpil:
Looks great @itsgreggreg ! Leave the build.rs as it is as I uncomment that capn proto build when working on persisting metadata

CHANGELOG.md
itsgreggreg: lol, thank you!

itsgreggreg:
Cool, you prefer I squash these commits down to a single commit or leave it as is?

lpil:
As is is fine by me.

lpil (approved):
Thank you so much @itsgreggreg ! Fantastic work.

I'll merge this once CI has finished.

itsgreggreg:
Absolutely, thank you, and it's been a pleasure!

PR #893, opened 2020-12-30
Lex 1-1 as 1 - 1

itsgreggreg:
This partially covers the issues described in https://github.com/gleam-lang/gleam/issues/876

Namely this will now lex `1-1` as `1 - 1` and `a-1` as `a - 1` like so:
gleam:
```rust
fn a() {
  let a = 1-1
  let b = a-1
  b
}
```
erlang:
```erlang
a() ->
    A = 1 - 1,
    B = A - 1,
    B.
```

With float literal:
```
error: Type mismatch
  ┌─ /Users/a/Trash/parser_test/src/one.gleam:4:11
  │
4 │   let c = 1.0-1.0
  │           ^^^

Expected type:

    Int

Found type:

    Float
```
With float in variable:
```
error: Type mismatch
  ┌─ /Users/a/Trash/parser_test/src/one.gleam:5:3
  │
5 │   c-1.0
  │   ^

Expected type:

    Int

Found type:

    Float
```

Because this is done at the lexer level, the formater will also work accordingly and format this:
```rust
fn a() {
  let a = 1-1
  let b = a-1
  let c = 1.0
  c-1.0
  b
}
```
to this:
```rust
fn a() {
  let a = 1 - 1
  let b = a - 1
  let c = 1.0
  c - 1.0
  b
}
```

What this doesn't cover:
* anywhere else a `-` is seen with no whitespace before it, like `five()-1`
* things that still result in a term that is constructed that is never used like `1 -1`, that will result in an erlang compilation warning like `gen/test/ints_test.erl:16: a term is constructed, but never used `

Questions:
Is this the right place to put the tests for this?
How important is it that cases like `five()-1` be lexed as `five() - 1`
Is there some case I'm not thinking of where `name-1` should have a different behavior? Since dashes aren't allowed in names I don't think so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Constants can now be defined after they are used in functions
 - The parser has been rewitten from scratch, dramatically improving error 
   messages and compilation times.
+- `1-1` and `a-1` are now parsed as `1 - 1` and `a - 1`
 
 ## v0.12.1 - 2020-11-15
 
--- a/src/parse/lexer.rs
+++ b/src/parse/lexer.rs
@@ -129,17 +129,26 @@ where
     fn consume_normal(&mut self) -> Result<(), LexicalError> {
         // Check if we have some character:
         if let Some(c) = self.chr0 {
+            let mut check_for_minus = false;
             if self.is_upname_start(c) {
                 let name = self.lex_upname()?;
                 self.emit(name)
             } else if self.is_name_start(c) {
+                check_for_minus = true;
                 let name = self.lex_name()?;
                 self.emit(name);
             } else if self.is_number_start(c, self.chr1) {
+                check_for_minus = true;
                 let num = self.lex_number()?;
                 self.emit(num);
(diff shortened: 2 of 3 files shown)


itsgreggreg:
Thinking about this a little bit more, it _could_ be possible to solve this in the parser rather than in the lexer during operator precedence parsing. Essentially if we have 2 exprs next to each other, check if the right one is a negative number and if it is, split it into a `-` and a positive number. Thus `expr negativeNumber` parses to `expr minus positiveNumber`.

The only downside there, depending on your point of view, would be that it wouldn't be able to tell if there was whitespace between the two exprs so `1 -1` and `1-1` would both be parsed as `1 - 1`. Perhaps that's a good thing. Perhaps we could emit a warning. Thoughts?

lpil (approved):
Very nice, thank you! Would you mind also updating the changelog?

lpil (approved):
Thank you! Happy new year too!

PR #896, opened 2021-01-02
Lex comments and resolve on demand

itsgreggreg:
`Parse::parse_module()` now outputs both an `UntypedModule` and a `ModuleExtra`. 
This `ModuleExtra` is all of the non source stuff we want to keep around for things
like formatting and doc generation. There is nothing new in there just comments
and blank line markers.

Comments are lexed as their start and end position only. This means that if they
are not needed we don't have to worry about the lifetime of the source file and
if they are needed then we can slice the source file on demand.

Comments are now only attached to modules when needed for formatting or doc
generation. `SourceTree::insert()` no longer results in module or doc comments
getting attached to the parsed module meaning commands that don't need that
like `build` don't have to pay the cost. As a result of that and the removal of
the comment blanker bulding is slightly faster. About 5% on my machine.

Feel free to send this PR in a different direction if this isn't what is desired. I wont 
be offended!

More info below on line attached comments.


src/format.rs
```diff
@@ -14,17 +14,42 @@ use std::sync::Arc;
...
+
+    let mut formatter = Formatter::with_comments(&intermediate);
+    Ok(pretty_module(&module, &mut formatter))
+}
+
+struct Intermediate<'a> {
+    comments: Vec<Comment<'a>>,
+    doc_comments: Vec<Comment<'a>>,
+    module_comments: Vec<Comment<'a>>,
+    empty_lines: &'a [usize],
```
itsgreggreg, line 45: This `Intermediate` struct is so I don't have to change the internals of `Formatter` which seems like it would have made this PR noisy and I didn't read `Formatter` enough to know if that was the right direction. Thoughts?

src/new.rs
```diff
@@ -521,7 +521,7 @@ fn validate_name(name: String) -> Result<String, Error> {
             name,
             reason: InvalidProjectNameReason::ErlangStandardLibraryModule,
         })
-    } else if crate::parser::is_gleam_reserved_word(name.as_str()) {
+    } else if crate::parse::lexer::str_to_keyword(name.as_str()).is_some() {
```
itsgreggreg, line 524: Don't gotta maintain a list of reserved words anymore, the lexer has to know about them. so much for all those PRs lol.

lpil: Hey this is really nice, we won't make that mistake again

src/format.rs
lpil: Yeah looks good. Can always optimise later if needed

src/parser.rs
```diff
@@ -1,537 +0,0 @@
-use crate::ast::*;
```
lpil, line 1: Very happy to see the back of this module!

lpil (approved):
Very nice!! Thank you

PR #898, opened 2021-01-02
Light weight source linker

itsgreggreg:
When I was going through the code for moving comment parsing to the lexer I noticed this code could be cleaned up a little.

### DIFF
--- a/src/docs/source_links.rs
+++ b/src/docs/source_links.rs
@@ -3,156 +3,87 @@ use crate::{
     config::{PackageConfig, Repository},
     project::Analysed,
 };
-use codespan_reporting::files::Files;
 use std::path::{Path, PathBuf};
 
-pub trait SourceLinks {
-    fn url(&self, location: &SrcSpan) -> Option<String>;
+pub struct SourceLinker {
+    line_starts: Vec<usize>,
+    url_pattern: Option<(String, String)>,
 }
 
-// Returns None for the urls which we can ignore in the template
-struct NoSourceLinks {}
+impl SourceLinker {
+    pub fn new(
+        project_root: impl AsRef<Path>,
+        project_config: &PackageConfig,
+        module: &Analysed,
+    ) -> Self {
+        let path_in_repo = get_path_in_repo(project_root, &module.path);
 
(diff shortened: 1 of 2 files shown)


itsgreggreg:
The diff's a little janky here, might be easier to look at the source directly https://github.com/gleam-lang/gleam/blob/f150e1623056ebec7951e4f79ffe3671e789fd1f/src/docs/source_links.rs

itsgreggreg:
The diff here is kinda janky. The gist is the SrcLinker only needs to know about line indexes and what the url parts look like. So it calculates that up front if it can and doesn't carry anything else with it. No more `Box`, `dyn`, `Trait` `codespan` or `clone`. about .5 the code size.

src/docs/source_links.rs
```diff
@@ -3,156 +3,87 @@ use crate::{
...
-    let end_line = codespan_file
-        .line_index((), location.end)
-        .unwrap_or_default()
-        + 1;
-
-    (start_line, end_line)
+// get all the line start byte offsets for a str
+fn line_starts(src: &str) -> Vec<usize> {
+    std::iter::once(0)
+        .chain(src.match_indices('\n').map(|(i, _)| i + 1))
```
lpil, line 71: Does this work when non `\n` line endings are used?

lpil (commented):
Hey clever. I like the format string bit, much simpler.

src/docs/source_links.rs
itsgreggreg: Oh good question, if the line endings are `\n`(posix) or `\r\n`(win) it should be fine. Will not if they are just `\r`(vintage).

It's copy-pasta from codespan: https://github.com/brendanzab/codespan/blob/master/codespan-reporting/src/files.rs#L251

lpil: That's probably ok I think? What do you reckon?

itsgreggreg: I think it's probably fine, kinda deferring judgement to codespan.

lpil: Let's go then :)

lpil (approved):

Issue #899, opened 2021-01-02
Formatter rendering non-decimal ints incorrectly


lpil:
It is using Erlang syntax instead of Gleam syntax. We should add some tests to cover this too.

PR #900, opened 2021-01-03
Fixed bad integer lexing

itsgreggreg:
Oops. 🐞

Fixes: #899 

Also added a lex error on binary and octal digits out of radix.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -67,6 +67,8 @@ use lexer::{LexResult, Spanned};
 use std::cmp::Ordering;
 use std::str::FromStr;
 use token::Tok;
+#[cfg(test)]
+mod tests;
 
 //
 // Public Interface
@@ -85,7 +87,9 @@ pub fn parse_module(src: &str) -> Result<(UntypedModule, ModuleExtra), ParseErro
 pub fn parse_expression_sequence(src: &str) -> Result<UntypedExpr, ParseError> {
     let lex = lexer::make_tokenizer(src);
     let mut parser = Parser::new(lex);
-    if let Some((e, _)) = parser.parse_expression_seq()? {
+    let expr = parser.parse_expression_seq();
+    let expr = parser.ensure_no_errors(expr)?;
+    if let Some((e, _)) = expr {
         Ok(e)
     } else {
         parse_error(ParseErrorType::ExpectedExpr, SrcSpan { start: 0, end: 0 })
@@ -135,10 +139,10 @@ where
     // place and instead we collect LexErrors in `self.lex_errors` and attempt to continue parsing.
     // Once parsing has returned we want to surface an error in the order:
     // 1) LexError, 2) ParseError, 3) More Tokens Left
-    fn ensure_no_errors(
+    fn ensure_no_errors<A>(
         &mut self,
-        statements: Result<Vec<UntypedStatement>, ParseError>,
-    ) -> Result<Vec<UntypedStatement>, ParseError> {
+        parse_result: Result<A, ParseError>,
+    ) -> Result<A, ParseError> {
         self.lex_errors.reverse();
         if let Some(error) = self.lex_errors.pop() {
             // Lex errors first
@@ -150,9 +154,9 @@ where
                     end: location,
                 },
             )
-        } else if statements.is_err() {
+        } else if parse_result.is_err() {
             // Then parse errors
-            statements
+            parse_result
         } else if let Some((start, _, end)) = self.next_tok() {
             // there are still more tokens
             let expected = vec!["An import, const, type, or function.".to_string()];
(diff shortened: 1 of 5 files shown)


src/parse/lexer.rs
```diff
@@ -506,14 +506,18 @@ where
...
+        if radix < 16 && Lexer::<T>::is_digit_of_radix(self.chr0, 16) {
+            Err(LexicalError {
+                error: LexicalErrorType::DigitOutOfRadix,
+                location: self.get_pos(),
+            })
+        } else {
+            let value = format!("{}{}", prefix, num);
+            let end_pos = self.get_pos();
+            Ok((start_pos, Tok::Int { value }, end_pos))
+        }
```
itsgreggreg, line 520: Brain fart, I new I had to add back the prefix but added the erlang prefix rather than the gleam

src/parse.rs
```diff
@@ -85,7 +87,9 @@ pub fn parse_module(src: &str) -> Result<(UntypedModule, ModuleExtra), ParseErro
 pub fn parse_expression_sequence(src: &str) -> Result<UntypedExpr, ParseError> {
     let lex = lexer::make_tokenizer(src);
     let mut parser = Parser::new(lex);
-    if let Some((e, _)) = parser.parse_expression_seq()? {
+    let expr = parser.parse_expression_seq();
+    let expr = parser.ensure_no_errors(expr)?;
+    if let Some((e, _)) = expr {
```
itsgreggreg, line 92: pass parse_expression_sequence through the same error process as parse_module so we can get at LexicalErrors

src/parse/tests.rs
```diff
@@ -0,0 +1,39 @@
...
+            error: ParseErrorType::LexError {
+                error: LexicalError {
+                    error: LexicalErrorType::DigitOutOfRadix,
+                    location: 9,
+                }
+            },
+            location: SrcSpan { start: 9, end: 9 },
+        }
+    );
+}
```
itsgreggreg, line 39: Added a LexError on a digit out of radix, happy to remove it if we don't really need it. Without it `0b012` gets lexed as `0b01; 2`.

src/parse/lexer.rs
```diff
@@ -550,6 +554,7 @@ where
             if let Some(c) = self.take_number(radix) {
                 value_text.push(c);
             } else if self.chr0 == Some('_') && Lexer::<T>::is_digit_of_radix(self.chr1, radix) {
+                value_text.push('_');
```
itsgreggreg, line 557: lol oops.

src/parse.rs
```diff
@@ -135,10 +139,7 @@ where
     // place and instead we collect LexErrors in `self.lex_errors` and attempt to continue parsing.
     // Once parsing has returned we want to surface an error in the order:
     // 1) LexError, 2) ParseError, 3) More Tokens Left
-    fn ensure_no_errors(
-        &mut self,
-        statements: Result<Vec<UntypedStatement>, ParseError>,
-    ) -> Result<Vec<UntypedStatement>, ParseError> {
+    fn ensure_no_errors<A>(&mut self, statements: Result<A, ParseError>) -> Result<A, ParseError> {
```
lpil, line 142: I think we may need to now inject the error message below? It talks about statements but it could be an expr that is expected

src/parse/lexer.rs
lpil: I think that should be a lex error, having it become an underscore would probably result in a more confusing error.

src/parse/tests.rs
lpil: I think this is a good error to raise, it'll be clearer than the alternative as to what the problem is.

Combining the errors sounds good but I'm happy either way.

lpil (approved):
Thanks for fixing this, saved me a bunch of time.

Do you want to add those bits before I merge?

src/parse.rs
itsgreggreg: Ah needed to change the var name from `statements` to `parse_result`

itsgreggreg:
Added a lex error on trailing underscores in numbers. I'll take a look at combining the error types in another PR, would make this one too noisy and take too long prob.

src/parse.rs
```diff
@@ -150,9 +154,9 @@ where
...
                 },
             )
-        } else if statements.is_err() {
+        } else if parse_result.is_err() {
             // Then parse errors
-            statements
+            parse_result
         } else if let Some((start, _, end)) = self.next_tok() {
             // there are still more tokens
             let expected = vec!["An import, const, type, or function.".to_string()];
```
lpil, line 162: This is the error message I meant- this will be incorrect if there is an end-of-file during a non-statement sequence, right?

itsgreggreg: oh whoops sorry, it's probably fine for now, `parse_expression_sequence` is only available in test and we might want to change things when making that public for some reason.

lpil: Ah cool! Ta

Issue #632, opened 2020-06-04
Generate erlang typespecs for use with dialyzer


jamesmacaulay:
I'd like to propose a modification to the compiler to generate Erlang typespecs for any types defined within Gleam. I'm using Gleam from within an Elixir project and I've come to really appreciate having Dialyzer integration within my editor for my Elixir code. Having generated typespecs would allow that experience to extend automatically to code that calls Gleam functions, and would also allow re-exporting types from Elixir or Erlang modules that "wrap" Gleam modules.

For example, the following Gleam code:

```gleam
pub type Tree(value) {
  Leaf(value)
  Node(Tree(value), Tree(value))
};

pub fn any(tree: Tree(a), check: fn(a) -> Bool) -> Bool {
  case tree {
    Leaf(i) -> check(i)
    Node(left, right) -> any(left, check) || any(right, check)
  }
}

pub fn has_even_leaf(tree: Tree(Int)) -> Bool {
  any(tree, fn(i) {
    i % 2 == 0
  })
}
```

...could generate the following types and function specifications along with the function implementations (unless I've got these wrong—let me know):

```erlang
-type tree(Value) :: {leaf, Value} | {node, tree(Value), tree(Value)}.

-spec any(Tree :: tree(A), Check :: fun((A) -> boolean())) -> boolean().

-spec has_even_leaf(Tree :: tree(integer()) -> boolean().
```

lpil:
A great suggestion! Let's definitely do this one

michaeljones:
I'd be happy to try to take a look at this. @lpil would you mind giving an overview or sharing your thoughts on what bits of the code base are involved and anything else you might normally share in such a situation :) Not rush though. I've got a busy week.

lpil:
I would possibly make a new module called `src/erl/typespec.rs' and start with a function that can turn a `Type` into a `pretty::Document` representing the typespec in a similar way to how the `erl` module does.

One thing to keep in mind is that once this is added to the Erlang code generation all of the integration tests (that's most of them) will break because the format of the generated Erlang will have changed. Perhaps there is an argument in favour of making typespec generation optional, but I am not sure.

michaeljones:
I've been playing around with this but I lack the familiarity with Erlang & typespecs to be confident with it. I would appreciate a bit of help with a few things:

[...]

PR #912, opened 2021-01-09
generate erlang typespecs

itsgreggreg:
Draft code for generating Erlang typespecs. Looking for feedback on anything in general but also specifically:
* Hows the code structure look?
* How's the usage of `pretty::` functions look?
* How do the generated type specs look?
* Anyone willing to give this a try on a project, run the dializer and report back?
* Does gleam want a compiler flag to turn these on and off?

Tests aren't passing because I haven't factored the typespecs in yet.

<hr>

#### Example

This Gleam:
```rust
pub type A {
  A
  B(String)
  C(tuple(Int, Float))
  D(Result(Int, Float))
  F(fn(String) -> Int)
  E(BitString)
}

fn a() {
  C(tuple(1, 1.0))
}

fn aa() {
  B
}

pub opaque type X(a) {
  X(a)
}

fn x(a) {
  X(a)
}

fn xx() {
  X
}

fn r() {
  case True {
    True -> Ok(tuple("a", "long", "typespec"))
    False -> Error(False)
  }
}

pub external type MyExtType

pub external fn z() -> MyExtType =
  "a" "b"

pub external fn y() -> A =
  "a" "b"

fn ex(a: MyExtType) {
  tuple(a)
}
```

results in this erlang:
```erlang
-module(one).
-compile(no_auto_import).

-export([z/0]).

-export_type([a/0, x/1, my_ext_type/0]).

-type a() :: a |
    {b, unicode:unicode_binary()} |
    {c, {integer(), float()}} |
    {d, {ok, integer()} | {error, float()}} |
    {f, fun((unicode:unicode_binary()) -> integer())} |
    {e, bitstring()}.
-opaque x(A) :: {x, A}.
-type my_ext_type() :: any()

-spec a() -> one:a().
a() ->
    {c, {1, 1.0}}.

-spec aa() -> fun((unicode:unicode_binary()) -> one:a()).
aa() ->
    fun(A) -> {b, A} end.

-spec x(any()) -> one:x(any()).
x(A) ->
    {x, A}.

-spec xx() -> fun((any()) -> one:x(any())).
xx() ->
    fun(A) -> {x, A} end.

-spec r() -> {ok,
              {unicode:unicode_binary(),
               unicode:unicode_binary(),
               unicode:unicode_binary()}} |
    {error, boolean()}.
r() ->
    case true of
        true ->
            {ok, {<<"a"/utf8>>, <<"long"/utf8>>, <<"typespec"/utf8>>}};

        false ->
            {error, false}
    end.

-spec z() -> my_ext_type().
    z() ->
    a:b().

-spec y() -> a().
    y() ->
    a:b().

-spec ex(one:my_ext_type()) -> {one:my_ext_type()}.
ex(A) ->
    {A}.
```

The only erlang errors on compilation are of unused functions... so that's good.

maybe closes: #632


src/erl.rs
```diff
@@ -142,30 +143,108 @@ pub fn record_definition(name: &str, fields: &[&str]) -> String {
...
+                                name.append(", ").append(concat(
+                                    c.args
+                                        .iter()
+                                        .map(|a| a.1.to_erlang_type_def())
+                                        .intersperse(", ".to_doc()),
+                                ))
+                            };
+                            "{".to_doc().append(constructor).append("}")
+                        })
+                        .intersperse(" | ".to_doc()),
```
lpil, line 200: It would be nice to wrap and indent definitions if they get too long. Doesn't have to be super pretty (the generated Erlang is a bit ugly) but something like this would be good:

```erl
-type my_type() :: the_first_atom |
  the_second_atom |
  the_third_atom |
  the_forth_atom |
  the_fifth_atom.
```

Something like this perhaps:

```rust
concat(constructors
    .iter()
    .map(constructor_type_definition)
    .intersperse(break_("| ", " | ")))
    .nest(INDENT)
    .group()
```

Something similar for the type tuples would also be good. I think there is a tuple function that would work.

michallepicki:
Gleam:
> pub type A {
>   A

Erlang spec:
> -type a() :: {a} |

Is that right? From the [Gleam for Erlang users cheatsheet](https://gleam.run/cheatsheets/gleam-for-erlang-users/#atoms)

> Any value in a type definition in Gleam that does not have any arguments is an atom in Erlang.

src/erl.rs
```diff
@@ -222,10 +305,33 @@ fn statement(statement: &TypedStatement, module: &[String]) -> Option<Document>
...
+    );
+    let spec = "-spec "
+        .to_doc()
+        .append(name)
+        .append("(")
+        .append(args_spec)
+        .append(") -> ")
+        .append(return_spec)
+        .append(".")
+        .append(line());
```
lpil, line 332: Would be nice to wrap and intent this when long but I dont' mind as much as the custom type definitions.

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
             | "zip"
     )
 }
+
+impl TypeAst {
+    pub fn to_erlang_type_def(&self) -> Document {
+        // reference : https://erlang.org/doc/reference_manual/typespec.html
+        match self {
+            TypeAst::Var { name, .. } | TypeAst::Hole { name, .. } => variable_name(name).to_doc(),
+            TypeAst::Tuple { elems, .. } => "{"
```
lpil, line 1495: I think there is a `tuple` function that can be used here. It'll handle wrapping and indentation and such.

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+                    if args.is_nil() {
+                        t.append("()")
+                    } else {
+                        t.append("(").append(args).append(")")
+                    }
+                } else {
+                    match name.as_str() {
+                        // TODO: more builtins ? Result?
+                        "Int" => "integer()".to_doc(),
+                        "String" => "string()".to_doc(),
```
lpil, line 1532: This would be `unicode:unicode_binary()` as we use utf8 encoded binary strings, while Erlang uses character lists.

https://erlang.org/doc/man/unicode.html#type-unicode_binary

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+            },
+            Self::App {
+                name, module, args, ..
+            } => {
+                if module.is_empty() {
+                    // built-ins
+                    // TODO: more builtins?
+                    match name.as_ref() {
+                        "Int" => "integer()".to_doc(),
+                        "String" => "binary()".to_doc(),
```
lpil, line 1576: `unicode:unicode_binary()`

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+            Self::Tuple {
+                elems, //Vec<Arc<Type>>,
+            } => {
+                let elems = concat(
+                    elems
+                        .iter()
+                        .map(|e| e.to_erlang_type_spec())
+                        .intersperse(", ".to_doc()),
+                );
+                "{".to_doc().append(elems).append("}")
```
lpil, line 1644: Can use the tuple function here.

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+                args,  //: Vec<Arc<Type>>,
+                retrn, //: Arc<Type>,
+            } => {
+                let args = concat(
+                    args.iter()
+                        .map(|a| a.to_erlang_type_spec())
+                        .intersperse(", ".to_doc()),
+                );
+                let retrn = retrn.to_erlang_type_spec();
+                "fun(("
```
lpil, line 1627: Double `(`?

src/erl.rs
```diff
@@ -1380,3 +1486,163 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+                    .to_doc()
+                    .append(args)
+                    .append(") -> ")
+                    .append(retrn)
+                    .append(")")
```
lpil, line 1632: What's this?

src/erl.rs
```diff
@@ -142,30 +143,108 @@ pub fn record_definition(name: &str, fields: &[&str]) -> String {
...
+            .append("]).")
+            .append(lines(2))
+    };
+
+    let type_exports = if type_exports.is_empty() {
+        nil()
+    } else {
+        "-export_type(["
+            .to_doc()
+            .append(concat(type_exports.into_iter().intersperse(", ".to_doc())))
```
lpil, line 238: Let's wrap these with a flex break

https://github.com/gleam-lang/gleam/blob/eb669e955be6b930379e236fe0f1872ae7844c39/src/format.rs#L240-L250

It would be nice to also do this for the regular exports but I understand if you would rather not add that to this PR.

lpil (commented):
Well spotted @michallepicki, custom types with no properties should be atoms.

> Hows the code structure look?

Looks great! I might be tempted to extract a function here and there but that's about it. The `erl` module was one of the first and it's pretty messy compared to a lot of the others.

> How's the usage of pretty:: functions look?

Good, but ideally we would make use of the `pretty` module's ability to wrap long lines by using `break_`, `nest` and `group`. The generated Erlang is a bit ugly but we aim for it to be more or less human looking.

> How do the generated type specs look?

Beyond the issue that michallepicki spotted and my inline notes they look good to me.

> Anyone willing to give this a try on a project, run the dializer and report back?

I'm not a dialyzer user so I'm unsure.

@CrowdHailer @QuinnWilton  is this something you can help with? I'm not sure who uses dialyzer.

> Does gleam want a compiler flag to turn these on and off?

Unsure. It would make it easier to integrate without having to update all the existing tests.

There's not really much in the way of performance cost to rendering these so maybe there's not much value to the user in allowing them to be turned off, so long as we don't make any mistakes that would break builds.

src/erl.rs
itsgreggreg: gave it my best go, passing `args_spec` through `wrap_args` and `return_spec` is going through `to_erlang_type_spec`?  I'm not 100% sure what I'm doing here. :)

src/erl.rs
itsgreggreg: happy to, not entirely sure what I'm doing here. :) I'll have to play around with the `pretty::` stuff, it's fairly abstract to me atm.

src/erl.rs
itsgreggreg: see above

src/erl.rs
itsgreggreg: afaict, type specs for functions look like:
```
       fun()                    %% any function
       | fun((...) -> Type)     %% any arity, returning Type
       | fun(() -> Type)
       | fun((TList) -> Type)
```
https://erlang.org/doc/reference_manual/typespec.html"

so for this code it looks like `fun((` `args` `) -> ` &nbsp;`return` `)`

itsgreggreg:
Pushed up some changes.  Updated the before / after above in the PR description.
Got the document breaking in some places but not sure if I'm doing it right / how to do it right. Hooked up external functions and types. External types are always `:: any()` as we don't know anything about them at compile time. A peculiarity with external functions is their `retrn` `TypeAst` doesn't have a `module` for it's `Constructor` so the type spec is unqualified whereas other uses of module defined types either custom or external are qualified.

src/erl.rs
lpil: Oh wow I've gotten how to Erlang

src/erl.rs
lpil: It is super abstract.

The basic idea is that any place there is a `break_` the printer may decide to render a newline after the contained string if the line is too long. It uses one string for when rending on 2 lines, the other for when on 1 line.

When a line is too long all breaks within a `group` render as a newline. Bit hard to explain but it can be good to add a group to each thing you would split by itself when manually formatting. i.e. put a group around each list, each function call, etc.

lpil:
> A peculiarity with external functions is their retrn TypeAst doesn't have a module for it's Constructor so the type spec is unqualified whereas other uses of module defined types either custom or external are qualified in non-external functions.

Wait, what? That sounds strange. Does this not work then?

```rust
import some_other_module

external fn get() -> some_other_module.ExportedType = "one" "two"
```

src/erl.rs
```diff
@@ -1492,52 +1517,52 @@ impl TypeAst {
...
+                                    tuple(vec!["error".to_doc(), arg_err].into_iter()),
+                                ]
+                                .into_iter()
+                                .intersperse(break_(" |", " | ")),
+                            )
+                            .nest(INDENT)
+                            .group()
+                        }
+                        // QUESTION: what should we do here?
+                        // I think after builtins this would be external types?
```
lpil, line 1561: I think this should be unreachable. Let's call `fatal_compiler_bug` as something has gone very wrong and we would like to know about it.

src/erl.rs
```diff
@@ -1558,12 +1583,10 @@ impl TypeAst {
 
 impl Type {
     pub fn to_erlang_type_spec(&self) -> Document {
-        tracing::trace!("spec_for_type: {:?}", &self);
         match self {
             Self::Var { typ } => match &*Arc::as_ref(typ).borrow() {
-                TypeVar::Generic { .. } => "any()".to_doc(),
+                TypeVar::Generic { .. } | TypeVar::Unbound { .. } => "any()".to_doc(),
```
lpil, line 1588: Erlang's type specs support parametric polymorphism so we'll want to handle that. Adding it later would be a breaking change.

```erl
-type result(A, B) :: {ok, A} | {error, B}.
```

This is trickier as we'll need to have some context when rendering types so that the type variables are rendered correctly across all the types of a definition.

src/erl.rs
lpil: If you're up for a little reading the paper linked from the pretty module is pretty readable as far as papers go

src/erl.rs
itsgreggreg: Could you think of some example code that might cause this to emit the wrong type? Currently parametric types are working:
```rust
pub type A(a, b, c, d) {
  A(a)
  AB(a, b)
  DAB(d, a, b)
  BDACA(b, d, a, c, a)
}
```
erlang:
```
-type a(A, B, C, D) :: {a, A} |
    {ab, A, B} |
    {dab, D, A, B} |
    {bdaca, B, D, A, C, A}.
```

src/erl.rs
itsgreggreg: This code is actually reached when an external function returns an external type. This is part of bullet 3 in my post above. Looking deeper 🤿.

itsgreggreg:
> > A peculiarity with external functions is their retrn TypeAst doesn't have a module for it's Constructor so the type spec is unqualified whereas other uses of module defined types either custom or external are qualified in non-external functions.
> 
> Wait, what? That sounds strange. Does this not work then?
> 
> ```rust
> import some_other_module
> 
> external fn get() -> some_other_module.ExportedType = "one" "two"
> ```

Looks like incorrect type is generated, comment highlighting it in b.erl:

```rust
// a.gleam
pub external type X

pub external fn a() -> X =
  "x" "y"


// b.gleam
import a.{X}

pub external fn b() -> X =
  "x" "y"

pub external fn bb() -> a.X =
  "x" "y"

```

```erlang
% a.erl
-export([a/0]).
-export_type([x/0]).

-type x() :: any()

-spec a() -> x().
    a() ->
    x:y().

% b.erl
-export([b/0, bb/0]).

%            ↓↓↓↓ this should be `a.x()`
-spec b() -> x().
    b() ->
    x:y().

-spec bb() -> a.x().
    bb() ->
    x:y().
```

Here's some debug info from handling the return types of the external functions in `b.gleam`

`b()`:
` TRACE to_erlang_type_def of:Constructor { location: SrcSpan { start: 37, end: 38 }, module: None, name: "X", args: [] }`

`bb()`:
` TRACE to_erlang_type_def of:Constructor { location: SrcSpan { start: 76, end: 79 }, module: Some("a"), name: "X", args: [] }`

notice the return for `b()` has 'None in its `module` field.

itsgreggreg:
Question:

Type _Constructors_ are of course translated `to_snake_case` per the Gleam documentation.
How does Gleam want Type _Names_ to be translated? 

Currently I am running them through `to_snake_case` as well, but we could run them through `atom` which would result in them being quoted.

to_snake_case: Result -> result(), MyType -> my_type(), a.B -> a.b()
atom: Result -> 'Result'(), MyType -> 'MyType'(), a.B -> a.'B'()

itsgreggreg:
Uh oh, looks like we're both in `erl.rs` and I've got some conflicts. Should I hold off?

src/erl.rs
lpil: Oh sorry, I have misread and misunderstood the code in that case. You're 2 steps ahead of me!

lpil:
> Looks like incorrect type is generated, comment highlighting it in b.erl:

Oh dear! I'll look into this.

> How does Gleam want Type Names to be translated?

Snake case please. That matches the Elixir and Erlang conventions best.

> Uh oh, looks like we're both in erl.rs and I've got some conflicts. Should I hold off?

Sorry about that, I added a lifetime to Document which was pretty noisy. I am finished now and won't be touching erl.rs any time soon beyond any bug fixes.

src/erl.rs
lpil: This is a bug. I'll try and find some time to investigate tomorrow.

src/erl.rs
itsgreggreg: I think I mostly got it. Great paper and interesting algorithm / data structure. Got carried away writing some code to better understand it, https://hexdocs.pm/pretty_print/pretty_print/ 🦩

lpil: Ooh very nice!

lpil:
>>    Looks like incorrect type is generated, comment highlighting it in b.erl:
>
> Oh dear! I'll look into this.

Ah yes I've figured it out. A TypeAst only has a module qualifier if it is qualified in the source, so no qualifier means the current scope. When hydrating a type AST the current module is used any time there is no qualifier.

```rust
// src/my_module.gleam

pub external type SomeType // this is my_module.SomeType
```

itsgreggreg:
> > > Looks like incorrect type is generated, comment highlighting it in b.erl:
> > 
> > 
> > Oh dear! I'll look into this.
> 
> Ah yes I've figured it out. A TypeAst only has a module qualifier if it is qualified in the source, so no qualifier means the current scope. When hydrating a type AST the current module is used any time there is no qualifier.
> 
> ```rust
> // src/my_module.gleam
> 
> pub external type SomeType // this is my_module.SomeType
> ```

The type is actually coming from a different module. If you look at function `bb()` it's return signature is `a.x()` . In the type signature for `b()` we have no way of knowing what module `x()` is coming from, and the type `x()` is not in the current module.

lpil:
> The type is actually coming from a different module. If you look at function bb() it's return signature is a.x() . In the type signature for b() we have no way of knowing what module x() is coming from, and the type x() is not in the current module.

Aye, the type ast only contains information written by the user. When hydrating types (and now printing them) we have to keep track of what types are in scope, including the prelude, imports, and type definitions to know where they are from.

https://github.com/gleam-lang/gleam/blob/babfe3bac9930be910fa8dc4ef585dd4e7f289b4/src/typ/hydrator.rs#L104-L107

itsgreggreg:
Gotcha, I think I'm getting it. I think essentially we don't want to render a `TypeAst` to an erlang type spec or def as `TypeAst`s don't have enough information. If that is the case then I think there are two options, pass the environment in and resolve types on demand, or resolve the types before getting to `erl::generate_erlang`. I'm not exactly sure how to move forward. Thoughts?

itsgreggreg:
ping :)

lpil:
Thanks for the reminder, I missed that last message.

The environment struct is mutable so I don't think we can pass it on demand. We could however change the external function statements to be annotated with the argument and return types in the typed-ast version. Then we can print the types rather than the TypeAst.

itsgreggreg:
Ok I think I have everything working except for `TypeVar`s . It seems `TypeVar`s don't have their user given name attached to them https://github.com/gleam-lang/gleam/blob/main/src/typ.rs#L277 and since they can be deeply nested in other types I'm not sure how to proceed.

itsgreggreg:
thinking about it a little more, since `TypeVar`s have an `id`, and for the type spec in erlang we don't really care what name the user gave the type var, we just care that they match in args and return types, it might be possible to map from id to string where 0 is A and 26 is AA and so forth. I'm not positive the `id`'s work that way just a hunch?

itsgreggreg:
Ok I think I got it all working, just gotta fix the tests and write a couple.

itsgreggreg:
Never mind, now a `CustomType`'s parameters don't match up with the type vars in its constructors as the parameters are from the source and the constructor's type vars are from the environment.

itsgreggreg:
All good, I _think_ I figured it out. Back to clean up and tests. Sorry for the chatter on this PR, been learning a lot about the codebase! :)

itsgreggreg:
Feature wise I think this is done. 🚀 Been writing and fixing tests and running the Erlang compiler and things seem to work and make sense. All that's left from a work perspective (other than responding to code review stuff) is finishing up the tests. I'm happy to do that, but there are quite a few and I wanted to make sure this is something gleam wants and I'm moving in the right direction before I tackle it!

In addition to generating the erlang types; external functions, their args, and custom types now have their type information attached during typing.

You can see some examples of how the tests are being changed  in the files changed, just adding the type related stuff.

A thumbs up or a go ahead and I'll finish the tests and move the PR outta draft mode!

src/erl.rs
```diff
@@ -1321,31 +1448,52 @@ fn incrementing_args_list(arity: usize) -> String {
...
     }
 }
 
+// When rendering a type variable to an erlang type spec we need all type variables with the
+// same id to end up with the same name in the generated erlang.
+// This function maps ids to A-Z then AA-AZ, but then AAA - AAZ, rather than BA-BZ etc.
+// This effectively gives us 52 type vars per file until they start looking weird.
+// This is because to properly cycle through the alphabet we have to convert the usize to base 26.
+// If we need more we can always modulus one more time and that'll give us 26*26 good looking vars.
+// Alternatively we could have the id itself after a character like, `A0` and `A24`.
```
lpil, line 1490: Thanks for this excellent comment. I think I would like to raise the number of pretty looking type vars past 52

src/erl.rs
```diff
@@ -1443,3 +1591,87 @@ pub fn is_erlang_standard_library_module(name: &str) -> bool {
...
+                TypeVar::Generic { id, .. } | TypeVar::Unbound { id, .. } => id_to_type_var(*id),
+                TypeVar::Link { typ } => Type::to_erlang_type_spec(typ),
+            },
+
+            Self::App {
+                name, module, args, ..
+            } => {
+                if module.is_empty() {
+                    // built-ins
+                    // TODO: more builtins?
```
lpil, line 1608: Can drop this comment now.

lpil (commented):
Absolutely gorgeous work! Please do continue, I'm very happy with the approach here. Thanks!

itsgreggreg:
All tests updated. I think this is "done".

In fixing more tests I uncovered another interesting difference between the type systems in Gleam and Erlang. Erlang doesn't allow unbound type vars anywhere while Gleam does. There are 2 different places this matters and 2 different solutions.

### Function types
In gleam this is fine: `fn(a: A) -> Int` or even this `fn() { Ok(1) }` both of which have a type with an unbound variable, the first in the args and the 2nd in the return. Interestingly the second's type isn't expressible in gleam syntax, but that's besides the point.

In order to translate this to Erlang, any type variable that has only one use is replaced with `any()`. This means that:

`fn(a: A) -> Int` goes to `-type fn((any()) -> integer()).` and `fn() { Ok(1)}` goes to `-type fn(() -> {ok, integer()} | {error | any()}).`

I think this is what people expect and makes sense from a value perspective.

### Custom Type Declarations
Gleam allows a custom type to have type variables that go unused in it's constructors. This is sometimes called a "phantom type".  Erlang doesn't allow this and we must figure out some workaround. The way things are currently implemented we cannot simply remove them because if a user used the type in another module we wouldn't know if the type was a phantom type or not and we wouldn't be able to know if we should remove the type var from the resulting Erlang. The work around in this PR is to "use" the type var in a "phantom" variant.

For example:
```
// This type has an unused type var "a"
pub type A(a) {
 A
}

// naively it would compile to this Erlang
-type a(A) :: a.

// which the erlang compiler would error on. 
// the solution in this PR produces this Erlang

-type a(A) :: a | {gleam_phantom, A}.
```

similarly more unused type vars go into the gleam_phantom variant like so:
```
// this gleam
type A(a,b,c,d) { A }
// compiles to
-type a(A,B,C,D) :: a | {gleam_phantom, A, B, C, D}.
```

This "solves" the usage problem and makes the Erlang compiler happy, but introduces an unreachable variant. I don't really know what the best thing is to do here. I think it is probably to remove the phantom variables on declaration and usage but gleam would need more information passed to the `erl` module to get that done.

itsgreggreg:
Lol, nevermind, core language tests to the rescue. Erlang doesn't like if you define a type and don't export it or use it in a function signature. 🤔

lpil:
That phantom type situation is quite tricky! Thanks for the write up. It sounds like you might have the best solution but I wonder if there is something else that can be done here.

> but introduces an unreachable variant

What does this mean from the user's point of view? Does it emit a warning or is it just that you could incorrectly pass in the pretend record and dialyzer would think that's ok?

> The easy solution here is to export all types in erlang regardless of whether they are public in Gleam or not. I think this is probably fine though I'm not 100% sure.

Let's go with this for now but I think we have enough information in the compiler to identify unused types so we can potentially handle that later.

itsgreggreg:
> What does this mean from the user's point of view? Does it emit a warning or is it just that you could incorrectly pass in the pretend record and dialyzer would think that's ok?

No warning, the Erlang compiler is satisfied. I think all it means is that Erlang users of functions that return Gleam phantom types will see that they might get a `{gleam_phantom, A}` tuple if they inspect the type. I think this is probably fine as erlang typing is success typing and users will just have to know that the `{gleam_phantom` value will never actually be created. And that yeah you could create the phantom tuple, return it, and the dializer would see that as correct.

lpil (approved):
Wonderful! This is really great! Thanks for picking up such a big chunk of working and sticking with it even with the unexpected issues.

lpil:
Last thing: Would you mind rebasing on main and adding an entry to the changelog? Thank you

CHANGELOG.md
```diff
@@ -79,7 +80,9 @@
   from Hex.
 
 ## v0.12.0-rc3 - 2020-10-24
+
 ## v0.12.0-rc2 - 2020-10-24
+
 ## v0.12.0-rc1 - 2020-10-24
```
itsgreggreg, line 86: ah whoops, these extra additions are from prettier formatting the markdown file, want me to remove them?

lpil: Nah looks good

lpil (approved):
Thank you!

PR #919, opened 2021-01-14
Warn on unused literals

### LINKED ISSUE #876: Warning about unused literals

brightly-salty:
Fixes #876 

Warns on sequential statements where the first expression is a simple constant.
Adds a warning "UnusedLiteral"
Also slightly refactors infer_seq to make it more idiomatic

This is my first Gleam PR so let me know if there are any issues!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -489,7 +489,7 @@ dependencies = [
 
 [[package]]
 name = "gleam"
-version = "0.13.0"
+version = "0.13.1"
 dependencies = [
  "askama",
  "atty",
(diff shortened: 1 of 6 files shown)


lpil:
This looks great!

I think we'll need to make a new function similar to `is_simple_constant` for this as `is_simple_constant` doesn't include lists or records.

It would also be good to include some tests. I think there is a warning test macro in the `typ/tests.rs` module.

lpil (approved):
Wonderful! Thank you

PR #923, opened 2021-01-14
Update dependencies

brightly-salty:
Updates all dependencies except for the following:

Does not upgrade capnc as I am not knowledgeable enough to update the schema.capnc file.
Does not upgrade codespan and codespan-reporting because of https://github.com/brendanzab/codespan/issues/309
Does not upgrade bytes because the newest release of hexpm depends on an older version.


lpil (approved):
Thank you!

PR #929, opened 2021-01-16
setup ci for docker release

CrowdHailer:


.github/workflows/release.yaml
```diff
@@ -86,3 +86,43 @@ jobs:
...
           asset_content_type: application/gzip
+  publish-docker:
+    name: publish-docker
+    runs-on: ubuntu-latest
+    steps: 
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Authenticate with github container registry
+        run: echo "${{ secrets.CR_PAT }}" | docker login https://ghcr.io -u ${{ github.actor }} --password-stdin
```
CrowdHailer, line 97: @lpil You will need to create a Personal Access Token and set it as a repository secret

https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token

.github/workflows/release.yaml
```diff
@@ -86,3 +86,43 @@ jobs:
...
+          IMAGE_ID_BASE=ghcr.io/${{ github.repository }}/gleam
+
+          # Change all uppercase to lowercase
+          IMAGE_ID_BASE=$(echo $IMAGE_ID_BASE | tr '[A-Z]' '[a-z]')
+
+          # Strip git ref prefix from version
+          VERSION=$(echo "${{ github.ref }}" | sed -e 's,.*/\(.*\),\1,')
+
+          # Strip "v" prefix from tag name
+          [[ "${{ github.ref }}" == "refs/tags/"* ]] && VERSION=$(echo $VERSION | sed -e 's/^v//')
```
CrowdHailer, line 116: A bunch of this code just turns a tag of `v0.x.y` into a docker tag of `0.x.y`, which is the convention for docker images.

.github/workflows/release.yaml
```diff
@@ -86,3 +86,43 @@ jobs:
...
+
+          # Use Docker `latest` tag convention
+          [ "$VERSION" == "master" ] && VERSION=latest
```
CrowdHailer, line 119: This isn't used right now, but if this workflow was ever to run on pushes that where untagged. Then it would create an image tagged latest.

For now I wouldn't worry that we're not creating latest tagged images

Dockerfile.elixir
```diff
@@ -0,0 +1,16 @@
+FROM rust:1.49.0 AS build
```
CrowdHailer, line 1: There is no problem that the build step is present in both Dockerfile.elixir and Dockerfile.erlang.
Layer caching is smart enough to do the work only once.

Dockerfile.elixir
```diff
@@ -0,0 +1,16 @@
...
+
+COPY . /usr/src/gleam-src
+
+RUN set -xe \
+        && cd /usr/src/gleam-src \
+        && make install \
+        && rm -rf /usr/src/gleam-src
+
+# Gleam fixed to latest Elixir at time of release
+FROM elixir:1.11.3
```
CrowdHailer, line 11: I think the best thing to do here is to have the elixir version that we are building upon fixed. It will be part of the release process to upgrade this value if we want.

I think having this step is explicit, because it might be confusing to use elixir:latest as if `gleam:0.13.2` is on `elixir:1.11.3` it makes sense, in my mind, to stay on `elixir:1.11.x` until the next moving gleam up to `0.14.0`

Dockerfile.elixir
lpil: Nice! I might change this though to use the statically linked binary created by the release so there's no compilation.

lpil (approved):
Thank you!

PR #931, opened 2021-01-17
WIP: Add newtypes

### LINKED ISSUE #637: Opaque unboxed types
Support creation of type wrappers that function like opaque types to other modules, require explicit boxing and unboxing within the module, and at runtime all boxing is erased so there is no runtime overhead.

i.e. Cents wraps an Int. At runtime it is just an Int, no record wrapper.

This is called a newtype in Haskell.

What should the syntax be?

brightly-salty:
Fixes #637 

Currently a Work In Progress, as there are probably numerous problems with the current implementation

Adds support for newtypes via the "inline" type keyword (placed after "opaque" if both are used). Newtypes must contain only one constructor. Instances of this constructor are replaced with the inner type. Type aliases cannot be newtypes.


src/parse.rs
```diff
@@ -2139,6 +2162,17 @@ where
         }
     }
 
+    // Expect one of multiple possible tokens, advances the token stream
+    fn expect_one_of(&mut self, toks: Vec<Tok>) -> Result<(Tok, (usize, usize)), ParseError> {
```
lpil, line 2166: Let's have this function take a slice rather than a vector to avoid taking ownership and cloning the vector here outside of the error path.

src/parse/error.rs
```diff
@@ -38,15 +38,17 @@ pub enum ParseErrorType {
...
     NoValueAfterEqual, // = <something other than a value>
-    NotConstType,      // :fn(), name, _  are not valid const types
-    OpNakedRight,      // Operator with no value to the right
-    OpaqueTypeAlias,   // Type aliases cannot be opaque
-    TooManyArgHoles,   // a function call can have at most 1 arg hole
+    NotConstType, // :fn(), name, _  are not valid const types
+    OpNakedRight, // Operator with no value to the right
+    OpaqueTypeAlias, // Type aliases cannot be opaque
+    InlineTypeAlias, // Type aliases cannot be inline
+    TooManyArgHoles, // a function call can have at most 1 arg hole
```
lpil, line 51: Can we undo this whitespace change? It's inconsistent now.

src/parse/lexer.rs
```diff
@@ -27,6 +27,7 @@ pub fn str_to_keyword(word: &str) -> Option<Tok> {
         "import" => Some(Tok::Import),
         "let" => Some(Tok::Let),
         "opaque" => Some(Tok::Opaque),
+        "inline" => Some(Tok::Inline),
```
lpil, line 30: I was wondering, is `inlined type` better than `inline type`?

src/typ.rs
```diff
@@ -293,6 +293,7 @@ pub struct TypeConstructor {
     pub module: Vec<String>,
     pub parameters: Vec<Arc<Type>>,
     pub typ: Arc<Type>,
+    pub inline_to: Option<String>,
```
lpil, line 296: Rather than adding this `inline_to` field let's reuse the existing constructor functionality. We will need to have a new constructor variant type that will generate the right Erlang, but otherwise it can be the same as normal record constructors.

lpil (commented):
A great start, thanks. The code looks good. I've some comments inline.

src/parse/lexer.rs
brightly-salty: To my ears "inlined type" sounds like it's already inlined in some way while "inline type" sounds like it will be inlined wherever it's used. But neither is obviously wrong or right. It's up to you.

lpil: Sounds good, let's keep it with "inline" then. Thank you

brightly-salty:
Okay, I've switched it to a value constructor variant. I'm not familiar with Erlang, so I might need guidance with where inline value constructors will differ from record value constructors in the generated Erlang.

lpil:
Hey @brightly-salty ! I sent you some messages in Discord the other day. Was that clear enough or would more be good? Let me know. Thanks

brightly-salty:
Hey @lpil I think I still need a little more guidance on how the inline value constructor variant should differ from the record value constructor variant. Thanks!

lpil:
Certainly!

https://github.com/gleam-lang/gleam/blob/23a53909a520defd0e5214098e5a7bbfc04af18e/src/typ.rs#L227-L231

I think we don't need any additional information to render an inline type so we can remove the fields from this variant and just have it be `Inline,`.

In the `erl` module in the `doc_args_call` function we will want to pattern match on the `ValueConstructorVariant::Inline` case and then render the first element from `args`.

Here's an example for the Record variant at present:

https://github.com/gleam-lang/gleam/blob/3fa57e3dbc5461a01e63da392a5ab2a0989aebb8/src/erl.rs#L1053-L1065

It would be something like this I think:

```rust
        TypedExpr::ModuleSelect {
            constructor: ModuleValueConstructor::Inline,
            ..
        }
        | TypedExpr::Var {
            constructor:
                ValueConstructor {
                    variant: ValueConstructorVariant::Inline,
                    ..
                },
            ..
        } => args.get(0).gleam_expect("No arguments for inline type call"),
```

And we'd also need some tests to ensure it has the right behaviour.

brightly-salty:
All right. I think we're in the home stretch. I just need to know what to put in the `todo!()` sections (`erl::var`,  `erl::expr`, `ExprTyper::infer_const`, and `PatternTyper::unify` @lpil. After that I'll create the tests

src/erl.rs
```diff
@@ -767,6 +767,8 @@ fn var<'a>(name: &'a str, constructor: &'a ValueConstructor, env: &mut Env<'_>)
             _ => atom(record_name.to_snake_case()),
         },
 
+        ValueConstructorVariant::Inline => todo!(),
+
         ValueConstructorVariant::LocalVariable => env.local_var_name(name.to_string()),
```
lpil, line 772: This can be the same as `LocalVariable` I think. `LocalVariable` is the next clause in this match expression so we can do this:
```suggestion
        ValueConstructorVariant::Inline | ValueConstructorVariant::LocalVariable => {
            env.local_var_name(name.to_string())
        }
```

src/erl.rs
```diff
@@ -1064,6 +1066,22 @@ fn docs_args_call<'a>(
...
+            constructor:
+                ValueConstructor {
+                    variant: ValueConstructorVariant::Inline,
+                    ..
+                },
+            ..
+        } => args
+            .get(0)
+            .gleam_expect("No arguments for inline type call")
+            .clone(),
```
lpil, line 1083: Oh that clone is annoying! Sorry I've given you some not super good advice before when I told you to do this.

Something like this would avoid cloning:

```suggestion
        } => {
            if args.is_empty() {
                crate::error::fatal_compiler_bug("No arguments for inline type call")
            } else {
                args.swap_remove(0)
            }
        }
```

src/erl.rs
```diff
@@ -1215,6 +1233,11 @@ fn expr<'a>(expression: &'a TypedExpr, env: &mut Env<'_>) -> Document<'a> {
             ..
         } => const_inline(literal, env),
 
+        TypedExpr::ModuleSelect {
+            constructor: ModuleValueConstructor::Inline,
+            ..
+        } => todo!(),
```
lpil, line 1239: I think we'd render the identity function here.

```suggestion
        } => "fun(A) -> A end".to_doc()
```

src/erl.rs
```diff
@@ -767,6 +767,8 @@ fn var<'a>(name: &'a str, constructor: &'a ValueConstructor, env: &mut Env<'_>)
             _ => atom(record_name.to_snake_case()),
         },
 
+        ValueConstructorVariant::Inline => todo!(),
```
lpil, line 770: I think we'd render the identity function here.

```suggestion
        } => "fun(A) -> A end".to_doc()
```

src/typ/expr.rs
```diff
@@ -1523,6 +1524,8 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
                 let tag = match &constructor.variant {
                     ValueConstructorVariant::Record { name, .. } => name.clone(),
 
+                    ValueConstructorVariant::Inline => todo!(),
```
lpil, line 1527: Let's return a new error here, one saying that inline types cannot be used in clause guards. We can do the same for the other TODO in this function.

Later we add support for this if it needed, but it would be a bigger job.

src/typ/pattern.rs
```diff
@@ -399,6 +399,7 @@ impl<'a, 'b, 'c> PatternTyper<'a, 'b, 'c> {
                     ValueConstructorVariant::Record { ref name, .. } => {
                         PatternConstructor::Record { name: name.clone() }
                     }
+                    ValueConstructorVariant::Inline => todo!(),
```
lpil, line 402: Here I think we would need to add a new `PatternConstructor` variant called `Inline` to match the value constructor.

src/typ/prelude.rs
```diff
@@ -1,7 +1,6 @@
 use super::{Environment, Type, TypeConstructor, ValueConstructorVariant};
 use crate::error::GleamExpect;
 use std::sync::Arc;
-
```
lpil, line 4: Could you revert the changes in this file please? Seems all the blank lines have been removed.

brightly-salty: I just ran `cargo fmt`... Is that not a part of the CI check?

brightly-salty: You could add a rustfmt.toml if you have specific preferences about the way it's formatted.

lpil: It is part of the CI check and by default it will keep these lines. The blank lines have been kept in the other files so I'm not sure this is the result of cargo fmt?

src/erl.rs
```diff
@@ -652,6 +652,12 @@ fn pattern<'a>(p: &'a TypedPattern, env: &mut Env<'_>) -> Document<'a> {
             ..
         } => tag_tuple_pattern(name, args, env),
 
+        Pattern::Constructor {
+            args: _args,
+            constructor: PatternConstructor::Inline,
+            ..
+        } => todo!(),
```
lpil, line 659: Here we can check to see if the args has 1 element and render that arg as a pattern if so, and exit using `fatal_compiler_bug` if not.

brightly-salty: Just using `pattern(args[0].value, env)`?

lpil: Might need to use `swap_remove` to get it to pass the borrow checker, but something like that, yes. We would need to check the length first to ensure we get a nice error message in the event of a bug rather than an out of bounce panic.

src/metadata/module_builder.rs
```diff
@@ -148,6 +148,8 @@ impl<'a> ModuleBuilder<'a> {
                 self.build_optional_field_map(builder.init_field_map(), field_map);
             }
 
+            ValueConstructorVariant::Inline => {}
```
lpil, line 151:
```suggestion
            // TODO: serialize
            ValueConstructorVariant::Inline => {}
```

We will need to handle this for the feature I'm working on currently. I can handle this after merging.

lpil (commented):
Looking good! Would you like to write some tests that validate the correct Erlang is generated?

lpil:
Hi @brightly-salty , are you still working on this?

brightly-salty:
Hi @lpil. I am in general, but I haven't lately because I got a little discouraged when the parsing failed. If you want to finish it up (you have write access to this branch/PR as a member of gleam-lang) feel free, but otherwise I'll get back on it in a week or two.

lpil:
I'm going to close this as we don't have the tests yet. Please do re-open if you'd like to continue, thank you

Issue #858, opened 2020-11-25
Automatically generate a link to repository in docs if available


michaeljones:
We now support a `repository` entry in the `gleam.toml` file. We would like to generate an entry in the `links` list for the documentation if the `repository` value is specified.

So using this type:

https://github.com/gleam-lang/gleam/blob/4b97df352b1c706ef11d09499a55ba38fb1784a1/src/config.rs#L39-L45

To generate an entry in the `links` vector:

https://github.com/gleam-lang/gleam/blob/4b97df352b1c706ef11d09499a55ba38fb1784a1/src/docs.rs#L68

The `repository` value is available on `project_config`.

PR #934, opened 2021-01-19
feat(docs): add automatically generate a link to repository in docs if available

mrgleam:
The PR add automatically generate a link to repository in docs if available for issue #858

Screen Shot 2564-01-19 at 20 23 01

Screen Shot 2564-01-19 at 20 22 49

### DIFF
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -78,15 +78,10 @@ pub fn generate_html(
         })
         .collect::<Vec<_>>();
 
-    let doc_links = &project_config
-        .docs
-        .links
-        .iter()
-        .map(|doc_link| Link {
-            name: doc_link.title.clone(),
-            path: doc_link.href.clone(),
-        })
-        .collect::<Vec<Link>>()[..];
+    let doc_links = project_config.docs.links.iter().map(|doc_link| Link {
+        name: doc_link.title.to_string(),
+        path: doc_link.href.to_string(),
+    });
 
     let repo_link = match &project_config.repository {
         Repository::GitHub { repo, .. }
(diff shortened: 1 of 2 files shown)


src/docs.rs
```diff
@@ -88,6 +88,30 @@ pub fn generate_html(
...
+            path: url.clone(),
+        }),
+        Repository::None => None,
+    }
+    .into_iter()
+    .collect::<Vec<Link>>();
+
+    let links = &doc_links
+        .iter()
+        .chain(repo_link.iter())
```
lpil, line 109: I think if we call `into_iter` rather than `iter` we remove the `map` call that clones the link,

lpil (commented):
Brilliant, thank you! I've left one inline note.

src/docs.rs
mrgleam: I try to change line 108 from ` iter()` to` into_iter()` and remove the `map`

I got error `a value of type std::vec::Vec<docs::Link> cannot be built from an iterator over elements of type &docs::Link`

Can I use `cloned()` for remove the `map`? 

```
    let links = &doc_links
        .iter()
        .chain(repo_link.iter())
        .cloned()
        .collect::<Vec<Link>>()
```

lpil: Sounds good! I was surprised that it didn't work but this isn't performance sensitive code so who cares 👍

mrgleam: Thank you @lpil for your recommend.
I try again in this morning  and it work. I can use` into_iter()` when I remove  reference in `doc_links`

This is ok?


Screen Shot 2564-01-20 at 09 58 28

lpil: That looks great! I think we can also drop the `collect` on `doc_links` and then remove the `into_iter` when it is referenced on line 106

mrgleam: Thank you @lpil 

It has been done :)

lpil (approved):
Sweet! Thank you

PR #938, opened 2021-01-23
Docs: Syntax Highlighting

### LINKED ISSUE #808: Syntax highlighting in generated gleam docs
Code blocks in documentation would be clearer if they were highlighted

rvcas:
I'm opening this up as a draft because it's not fully done yet and I would like some feedback on whether or not this is the correct approach. We might want to do the css similar to the website.

closes #808

### DIFF
--- a/templates/documentation_layout.html
+++ b/templates/documentation_layout.html
@@ -8,7 +8,7 @@
     <!-- The docs_config.js file is provided by HexDocs and shared
          between multiple versions of the same package. -->
     <script src="{{ unnest }}/docs_config.js"></script>
-    <link rel="stylesheet" href="https://unpkg.com/@highlightjs/cdn-assets@10.5.0/styles/default.min.css">
+    <link rel="stylesheet" href="https://unpkg.com/@highlightjs/cdn-assets@10.5.0/styles/atom-one-light.min.css">
     <script src="https://unpkg.com/@highlightjs/cdn-assets@10.5.0/highlight.min.js"></script>
     <script src="https://unpkg.com/@highlightjs/cdn-assets@10.5.0/languages/erlang.min.js"></script>
     <script src="{{ unnest }}/highlightjs-gleam.js"></script>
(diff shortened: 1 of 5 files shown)


templates/documentation_layout.html
```diff
@@ -8,6 +8,8 @@
     <!-- The docs_config.js file is provided by HexDocs and shared
          between multiple versions of the same package. -->
     <script src="{{ unnest }}/docs_config.js"></script>
+    <link rel="stylesheet" href="//unpkg.com/@highlightjs/cdn-assets@10.5.0/styles/default.min.css">
+    <script src="//unpkg.com/@highlightjs/cdn-assets@10.5.0/highlight.min.js"></script>
```
lpil, line 12: Oh this is cool!

I had a play around with this URL and I think we can pull the Gleam plugin from the CDN too

https://unpkg.com/highlightjs-gleam@0.2.2/dist/gleam.min.js

rvcas: I'll try it now

rvcas:
* The padding around the `pre` blocks weren't getting along with the background color of the code blocks after the highlight.
* The CDN link https://unpkg.com/highlightjs-gleam@0.2.2/dist/gleam.min.js seems to return a node module so we might have to vendor it for now
* Instead of defaulting all code blocks to gleam, we could let people use ` ```erlang` as well, which is already included for the rebar.config info that usually shows up in READMEs


Screen Shot 2021-01-23 at 4 33 51 PM
Screen Shot 2021-01-23 at 4 34 06 PM
Screen Shot 2021-01-23 at 4 34 21 PM

rvcas:
atom-one-light:
Screen Shot 2021-01-23 at 5 24 11 PM
Screen Shot 2021-01-23 at 5 24 21 PM

lpil:
That looks really nice! Did you say that code blocks in doc comments with no language specified get highlighted as Gleam?

rvcas:
yes they do

lpil (approved):
Amazing, thank you!

Issue #902, opened 2021-01-04
"\^" at end of string causes erlang to see the string as unclosed


itsgreggreg:
Erlang interprets escaped caret (`\^`) in a string as the start of a control code escape. According to erlang's docs, `\^a..\^z` and `\^A..\^Z` are valid control code escapes and get replaced with their ascii control code equivalent. 

Erlang docs: http://erlang.org/doc/reference_manual/data_types.html#escape-sequences
Escape sequences (column starting with `^@`): https://en.wikipedia.org/wiki/ASCII#Control_code_chart

So in erlang terms this means:
```
<<"\^@"/utf8>> == <<0>>
<<"\^a"/utf8>> == <<1>>
-- and so forth

-- and very unfortunately
<<"\^""/utf8>> == <<2>>
```

In Gleam terms this means:
```
"\^I" == "\t"
"\^J" == "\n"

// and very unfortunately
"\^"" is a syntax error in gleam

// and also very unfortunately 
"\^" causes the string to be un-closed in erlang
```

There are a few interesting things here.
1) Erlang's docs specify that a-z and A-Z are valid, they don't say anything about other characters. `\^"` being interpreted as <<2>> is redundant with `\^b`. Perhaps this is a bug in erlang. probably not :(

2) this passes:
```rust
  <<0:8>>
  |> bit_string.to_string()
  |> result.unwrap("")
  |> should.equal("\^@")
```

3) If you for some reason end a string with `\^`, so "any text \\^", Gleam will compile it to Erlang that won't compile. This is because of the previously mentioned `\^"` gets interpreted as `<<2>>` and erlang sees the string as still open.

Considering this is probably not something erlang will "fix" what should be done here?

lpil:
Yikes. That is not fun. How did you find that one?

Seems like we could replace `\^` with `\\^` when rendering Erlang unless I'm missing something. This is the first time I'm learning about these escape codes.

itsgreggreg:
I was escaping strings destined for a regex and didn't properly escape a caret. I wrote `string.replace("^", "\^")` and it generated the offending Erlang.

Unfortunately it's not a simple as replace `\^` with `\\^` as you have to backwards count slashes. Doing that on `\\^`, which is properly escaped, would produce `\\\^` which is not properly escaped. 

This is also complicated by gleams desire to have multiple targets.

Javascript has these escapes:
[...]

PR #942, opened 2021-01-24
Minimal string escapes

itsgreggreg:
Changes Gleam to only support a base set of string escapes. `\r`, `\n`,`\t`, `\\` and `\"`.

Realized that with these minimal string escapes, since they are supported as-is by essentially any target, the work can all be done in the lexer.

This PR treats any other escape as the character itself, so `\x` compiles to `x` as will any character other than the escapes above.

Another valid strategy would be to auto escape slashes so `\x` would compile to `\\x`. I chose the strategy in this PR as it's the same way erlang and javascript handle escapes that aren't recognized and it means that having a slash show up in the output is always consistently two slashes.

Since Gleam is pre 1.0 it's probably fine but it should be noted this is a backwards incompatible change. If anyone is relying on other erlang escape sequences they will no longer work.

Fixes: #902

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Automatically generate a link to repository in docs if available.
 - Gleam now generates Erlang typespecs.
 - Code in HTML documentation is has highlighted syntax.
+- Gleam now only supports `\r`, `\n`, `\t`, `\"`, and `\\` string escapes.
 
 ## v0.13.2 - 2020-01-14
 
--- a/src/erl/tests.rs
+++ b/src/erl/tests.rs
@@ -2393,3 +2393,22 @@ fn build_in_erlang_type_escaping() {
 "
     );
 }
+
+#[test]
+fn allowed_string_escapes() {
+    assert_erl!(
+        r#"fn a() { "\n" "\r" "\t" "\\" "\"" "\\^" }"#,
+        r#"-module(the_app).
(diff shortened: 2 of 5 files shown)


itsgreggreg:
We could also syntax error on unsupported escapes. This would give users a helpful error message if they tried to use an escape Gleam didn't support and Gleam could add escapes in the future without breaking existing code. Maybe this is the best option? 🤔

lpil:
> We could also syntax error on unsupported escapes. This would give users a helpful error message if they tried to use an escape Gleam didn't support and Gleam could add escapes in the future without breaking existing code. Maybe this is the best option? 🤔

I really like this plan! With a minimal set of escapes it will be good to have a way to add new ones without it being a hard to find breaking change.

itsgreggreg:
Updated it to syntax error on an unsupported escape. Here's what the error looks like:
```

error: Syntax error
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:1:20
  │
1 │ pub const = "hello \ there"
  │                    ^ This is an unescaped backslash.

Hint: Add another backslash before it.
See: https://gleam.run/book/tour/strings.html#escape-sequences
```

note, that section does not currently exist on the strings page in the tour.

lpil (approved):
Nice! Thank you.

Would you be able to also make a note of this behaviour in the tour? https://gleam.run/book/tour/strings.html#escape-sequences

PR #943, opened 2021-01-25
Stricter clippy lints

brightly-salty:
Large PR but mostly many small changes rather than big changes

Fixes all current clippy lints
Removes duplicate lints
Removes explicitly warning/denying on lints that are warn-by-default or deny-by-default
Adds warn/deny for clippy::pedantic
Fixes all new clippy lints
Makes implicit truncation and wrapping explicit

Let me know if any of the changes don't sit right or don't make sense.

### DIFF
--- a/src/bit_string.rs
+++ b/src/bit_string.rs
@@ -12,6 +12,7 @@ pub struct BinaryTypeSpecifier<T> {
 }
 
 impl<T> BinaryTypeSpecifier<T> {
+    #[allow(clippy::too_many_lines)]
     pub fn new(options: &[BitStringSegmentOption<T>], must_have_size: bool) -> Result<Self, Error>
     where
         T: Clone,
@@ -136,16 +137,14 @@ impl<T> BinaryTypeSpecifier<T> {
         match self.typ {
             Some(BitStringSegmentOption::Integer { .. }) => Some(crate::typ::int()),
             Some(BitStringSegmentOption::Float { .. }) => Some(crate::typ::float()),
-            Some(BitStringSegmentOption::Binary { .. }) => Some(crate::typ::bit_string()),
-            Some(BitStringSegmentOption::BitString { .. }) => Some(crate::typ::bit_string()),
-            Some(BitStringSegmentOption::UTF8 { .. }) => Some(crate::typ::string()),
-            Some(BitStringSegmentOption::UTF16 { .. }) => Some(crate::typ::string()),
-            Some(BitStringSegmentOption::UTF32 { .. }) => Some(crate::typ::string()),
-            Some(BitStringSegmentOption::UTF8Codepoint { .. }) => Some(crate::typ::utf_codepoint()),
-            Some(BitStringSegmentOption::UTF16Codepoint { .. }) => {
-                Some(crate::typ::utf_codepoint())
-            }
-            Some(BitStringSegmentOption::UTF32Codepoint { .. }) => {
+            Some(BitStringSegmentOption::Binary { .. })
+            | Some(BitStringSegmentOption::BitString { .. }) => Some(crate::typ::bit_string()),
+            Some(BitStringSegmentOption::UTF8 { .. })
+            | Some(BitStringSegmentOption::UTF16 { .. })
+            | Some(BitStringSegmentOption::UTF32 { .. }) => Some(crate::typ::string()),
+            Some(BitStringSegmentOption::UTF8Codepoint { .. })
+            | Some(BitStringSegmentOption::UTF16Codepoint { .. })
+            | Some(BitStringSegmentOption::UTF32Codepoint { .. }) => {
                 Some(crate::typ::utf_codepoint())
             }
             _ => None,
@@ -193,10 +192,11 @@ pub enum Error {
 
 impl<A> BitStringSegmentOption<A> {
     pub fn unit_is_allowed(&self) -> bool {
-        !matches!(self,
+        !matches!(
+            self,
             BitStringSegmentOption::UTF8 { .. }
-            | BitStringSegmentOption::UTF16 { .. }
-            | BitStringSegmentOption::UTF32 { .. }
+                | BitStringSegmentOption::UTF16 { .. }
+                | BitStringSegmentOption::UTF32 { .. }
         )
     }
 }
--- a/src/bit_string.rs
+++ b/src/bit_string.rs
@@ -12,7 +12,6 @@ pub struct BinaryTypeSpecifier<T> {
 }
 
 impl<T> BinaryTypeSpecifier<T> {
-    #[allow(clippy::too_many_lines)]
     pub fn new(options: &[BitStringSegmentOption<T>], must_have_size: bool) -> Result<Self, Error>
     where
         T: Clone,
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -140,7 +140,7 @@ pub fn generate_html(
         let name = module.name.join("/");
 
         // Read module src & create line number lookup structure
-        let source_links = SourceLinker::new(&project_root, project_config, &module);
+        let source_links = SourceLinker::new(&project_root, project_config, module);
 
         let template = ModuleTemplate {
             unnest: module.name.iter().map(|_| "..").intersperse("/").collect(),
@@ -157,7 +157,7 @@ pub fn generate_html(
                     .ast
                     .statements
                     .iter()
-                    .flat_map(|statement| function(&source_links, &statement))
+                    .flat_map(|statement| function(&source_links, statement))
                     .collect();
                 f.sort();
                 f
@@ -167,7 +167,7 @@ pub fn generate_html(
                     .ast
                     .statements
                     .iter()
-                    .flat_map(|statement| type_(&source_links, &statement))
+                    .flat_map(|statement| type_(&source_links, statement))
                     .collect();
                 t.sort();
                 t
@@ -177,7 +177,7 @@ pub fn generate_html(
                     .ast
                     .statements
                     .iter()
-                    .flat_map(|statement| constant(&source_links, &statement))
+                    .flat_map(|statement| constant(&source_links, statement))
                     .collect();
                 c.sort();
                 c
@@ -253,6 +253,7 @@ fn markdown_documentation(doc: &Option<String>) -> String {
     }
 }
 
+#[allow(clippy::integer_division)]
 fn render_markdown(text: &str) -> String {
     let mut s = String::with_capacity(text.len() * 3 / 2);
     let p = pulldown_cmark::Parser::new_ext(&*text, pulldown_cmark::Options::all());
@@ -468,7 +469,7 @@ fn check_app_file_version_matches(
                 // Error if we've found the version and it doesn't match
                 Err(Error::VersionDoesNotMatch {
                     toml_ver: project_config.version.clone(),
-                    app_ver: version.to_string(),
+                    app_ver: version,
                 })
             }
         })
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -8,7 +8,7 @@ use crate::{
     config::{DocsPage, PackageConfig, Repository},
     docs::source_links::SourceLinker,
     error::{Error, GleamExpect},
-    format,
+    format::Formatter,
     fs::OutputFile,
     pretty,
     project::{self, Analysed, ModuleOrigin},
@@ -214,7 +214,7 @@ fn function<'a>(
     source_links: &SourceLinker,
     statement: &'a TypedStatement,
 ) -> Option<Function<'a>> {
-    let mut formatter = format::Formatter::new();
+    let mut formatter = Formatter::new();
     match statement {
         Statement::ExternalFn {
             public: true,
@@ -242,12 +242,7 @@ fn function<'a>(
         } => Some(Function {
             name,
             documentation: markdown_documentation(doc),
-            signature: print(format::Formatter::docs_fn_signature(
-                true,
-                name,
-                args,
-                &ret.clone(),
-            )),
+            signature: print(Formatter::docs_fn_signature(true, name, args, &ret.clone())),
             source_url: source_links.url(location),
         }),
 
@@ -262,16 +257,19 @@ fn markdown_documentation(doc: &Option<String>) -> String {
     }
 }
 
-#[allow(clippy::integer_division)]
 fn render_markdown(text: &str) -> String {
-    let mut s = String::with_capacity(text.len() * 3 / 2);
+    let mut s = String::with_capacity(
+        (3 * text.len())
+            .checked_div(2)
+            .gleam_expect("Somehow 2 turned out to be 0"),
+    );
     let p = pulldown_cmark::Parser::new_ext(&*text, pulldown_cmark::Options::all());
     pulldown_cmark::html::push_html(&mut s, p);
     s
 }
 
 fn type_<'a>(source_links: &SourceLinker, statement: &'a TypedStatement) -> Option<Type<'a>> {
-    let mut formatter = format::Formatter::new();
+    let mut formatter = Formatter::new();
     match statement {
         Statement::ExternalType {
             public: true,
@@ -282,7 +280,7 @@ fn type_<'a>(source_links: &SourceLinker, statement: &'a TypedStatement) -> Opti
             ..
         } => Some(Type {
             name,
-            definition: print(format::Formatter::external_type(true, name.as_str(), args)),
+            definition: print(Formatter::external_type(true, name.as_str(), args)),
             documentation: markdown_documentation(doc),
             constructors: vec![],
             source_url: source_links.url(location),
@@ -359,7 +357,7 @@ fn constant<'a>(
     source_links: &SourceLinker,
     statement: &'a TypedStatement,
 ) -> Option<Constant<'a>> {
-    let mut formatter = format::Formatter::new();
+    let mut formatter = Formatter::new();
     match statement {
         Statement::ModuleConstant {
             public: true,
--- a/src/docs.rs
+++ b/src/docs.rs
@@ -146,8 +146,7 @@ pub fn generate_html(
         let source_links = SourceLinker::new(&project_root, project_config, module);
 
         let template = ModuleTemplate {
-            unnest: itertools::Itertools::intersperse(module.name.iter().map(|_| ".."), "/")
-                .collect(),
+            unnest: module.name.iter().map(|_| "..").intersperse("/").collect(),
             links: &links,
             pages: &pages,
             documentation: render_markdown(module.ast.documentation.iter().join("\n").as_str()),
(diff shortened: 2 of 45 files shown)


src/docs/tests.rs
```diff
@@ -41,19 +43,19 @@ pub fn complicated_fun(
...
-        version: Default::default(),
-        repository: Default::default(),
-        description: Default::default(),
-        dependencies: Default::default(),
+        docs: crate::config::Docs::default(),
+        tool: crate::config::BuildTool::default(),
+        version: String::default(),
+        repository: crate::config::Repository::default(),
+        description: String::default(),
+        dependencies: std::collections::HashMap::default(),
```
lpil, line 51: This seems worse, what is the reason for this one? The Rust type checker can easily infer this information.

src/docs/tests.rs
```diff
@@ -41,19 +43,19 @@ pub fn complicated_fun(
...
         otp_start_module: None,
     };
 
-    let mut analysed = crate::project::analysed(vec![input]).expect("Compilation failed");
+    let mut analysed = crate::project::analysed(&[input]).gleam_expect("Compilation failed");
     analysed
         .iter_mut()
-        .for_each(|a| a.attach_doc_and_module_comments());
+        .for_each(crate::project::Analysed::attach_doc_and_module_comments);
```
lpil, line 58: This seems a lot more verbose, let's not do this one.

src/erl.rs
```diff
@@ -33,15 +39,15 @@ pub fn generate_erlang(analysed: &[Analysed]) -> Vec<OutputFile> {
...
             .join(project::OUTPUT_DIR_NAME)
             .join(origin.dir_name());
         let erl_module_name = name.join("@");
 
-        for (name, text) in records(ast).into_iter() {
+        for (name, text) in &records(ast) {
             files.push(OutputFile {
                 path: gen_dir.join(format!("{}_{}.hrl", erl_module_name, name)),
-                text,
+                text: text.clone(),
```
lpil, line 50: Let's undo this change to the loop, is it slower now as it has to clone the text each time.

src/bit_string.rs
```diff
@@ -12,6 +12,7 @@ pub struct BinaryTypeSpecifier<T> {
 }
 
 impl<T> BinaryTypeSpecifier<T> {
+    #[allow(clippy::too_many_lines)]
```
lpil, line 15: Let's not enable this lint globally, I don't think it is useful

src/docs.rs
```diff
@@ -253,6 +263,7 @@ fn markdown_documentation(doc: &Option<String>) -> String {
     }
 }
 
+#[allow(clippy::integer_division)]
```
lpil, line 266: What's the recommended way to do this without adding this attribute? I would prefer not to add these exception attributes

src/docs.rs
```diff
@@ -272,7 +283,7 @@ fn type_<'a>(source_links: &SourceLinker, statement: &'a TypedStatement) -> Opti
             ..
         } => Some(Type {
             name,
-            definition: print(formatter.external_type(true, name.as_str(), args)),
+            definition: print(format::Formatter::external_type(true, name.as_str(), args)),
```
lpil, line 286: Let's import Formatter and use it without the qualification here.

src/typ/error.rs
```diff
@@ -498,6 +500,7 @@ impl UnifyError {
         self.with_unify_error_situation(UnifyErrorSituation::ReturnAnnotationMismatch)
     }
 
+    #[allow(clippy::wrong_self_convention)]
```
lpil, line 503: Let's rename this method.

src/typ/hydrator.rs
```diff
@@ -28,15 +31,21 @@ pub struct ScopeResetData {
     created_type_variable_ids: im::HashSet<usize>,
 }
 
-impl Hydrator {
-    pub fn new() -> Self {
+impl Default for Hydrator {
+    fn default() -> Self {
```
lpil, line 35: Why is this done?

src/typ/expr.rs
```diff
@@ -69,17 +78,17 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
     ///
     pub fn infer(&mut self, expr: UntypedExpr) -> Result<TypedExpr, Error> {
         match expr {
-            UntypedExpr::ListNil { location, .. } => self.infer_nil(location),
+            UntypedExpr::ListNil { location, .. } => Ok(self.infer_nil(location)),
```
lpil, line 81: If these functions don't return a result they should not be called infer_ any more. `make_nil` perhaps. Same for the others.

src/typ/expr.rs
```diff
@@ -384,6 +393,7 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
         })
     }
 
+    #[allow(clippy::map_unwrap_or)]
```
lpil, line 396: What does this mean?

src/typ/expr.rs
```diff
@@ -745,10 +746,10 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
 
         let return_type = self.new_unbound_var(self.environment.level);
 
-        for subject in subjects.into_iter() {
+        for subject in subjects {
```
lpil, line 749: Let's revert this, it is slower as it clones the subject wastefully.

src/typ/expr.rs
```diff
@@ -757,8 +758,8 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
             subject_types.push(subject_type);
         }
 
-        for clause in clauses.into_iter() {
-            let typed_clause = self.infer_clause(clause, &subject_types)?;
+        for clause in clauses {
+            let typed_clause = self.infer_clause(clause.clone(), &subject_types)?;
```
lpil, line 762: Let's revert this, it is slower.

src/typ/expr.rs
```diff
@@ -1655,13 +1668,13 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
...
-        untyped_elements: Vec<UntypedConstant>,
+        untyped_elements: &[UntypedConstant],
         location: SrcSpan,
     ) -> Result<TypedConstant, Error> {
         let mut elements = Vec::with_capacity(untyped_elements.len());
 
-        for element in untyped_elements.into_iter() {
-            let element = self.infer_const(&None, element)?;
+        for element in untyped_elements {
+            let element = self.infer_const(&None, element.clone())?;
```
lpil, line 1677: Same again, this is slower as it clones the ast.

lpil (commented):
Hello!

I like a lot the removal of expect and unwrap, the shorter lint list in  `main`, as well as the change from `to_*` to `into_*` when `self` is consumed. Let's keep those.

I especially don't like forced use of `clone`, `if`, trait name qualification, and limits on function length. The other ones I don't understand the value of yet. If there's not a clear motivation I don't think we should include the lints.

This PR is very big and out of the blue! It if difficult to review this much code and I would prefer this to be done in multiple smaller PRs. Ideally an issue would have been created first to discuss which lints we want enabled.

src/docs/tests.rs
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#default_trait_access
> It’s more clear to the reader to use the name of the type whose default is being gotten than the generic Default.

src/docs/tests.rs
brightly-salty: We could import "Analyzed" if you want... The purpose is to remove the closure when a function will do.

src/erl.rs
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#explicit_into_iter_loop
This was the reason... If you want me to change it, I'll add a global allow for this lint or allow for this function.

src/bit_string.rs
brightly-salty: I agree that it is annoying. I'll add it to the global allow because it comes as part of clippy::pedantic which is generally useful.

src/docs.rs
brightly-salty: Let me fix it... That was before I implemented the truncate!() macro

src/erl.rs
```diff
@@ -209,26 +220,26 @@ pub fn module(module: &TypedModule, writer: &mut impl Utf8Writer) -> Result<()>
...
 
         Statement::Fn {
             args, name, body, ..
         } => Some(mod_fun(name.as_ref(), args.as_slice(), body, module)),
 
-        Statement::ExternalFn { public: false, .. } => None,
         Statement::ExternalFn {
             fun,
-            module,
+            module: external_module,
```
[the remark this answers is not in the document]
brightly-salty: This was caused by a desire to avoid shadowing: https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

src/erl.rs
```diff
@@ -77,28 +83,31 @@ struct Env<'a> {
...
     }
 
-    pub fn next_local_var_name<'a>(&mut self, name: String) -> Document<'a> {
-        let next = self.erl_function_scope_vars.get(&name).map_or(0, |i| i + 1);
-        let _ = self.erl_function_scope_vars.insert(name.clone(), next);
-        let _ = self.current_scope_vars.insert(name.clone(), next);
+    pub fn next_local_var_name<'a>(&mut self, name: &str) -> Document<'a> {
+        let next = self
+            .erl_function_scope_vars
+            .get(&name.to_owned())
```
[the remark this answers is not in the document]
brightly-salty: This was caused by a sequence of https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value and then https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg. If you want, I can revert these changes and either allow them on a per-function basis or globally.

src/erl.rs
```diff
@@ -77,28 +83,31 @@ struct Env<'a> {
...
+            current_scope_vars: im::HashMap::default(),
+            erl_function_scope_vars: im::HashMap::default(),
             module,
         }
     }
 
-    pub fn local_var_name<'a>(&mut self, name: String) -> Document<'a> {
-        match self.current_scope_vars.get(&name) {
+    pub fn local_var_name<'a>(&mut self, name: &str) -> Document<'a> {
+        match self.current_scope_vars.get(&name.to_owned()) {
```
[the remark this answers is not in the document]
brightly-salty: This was caused by a sequence of https://rust-lang.github.io/rust-clippy/master/index.html#needless_pass_by_value and then https://rust-lang.github.io/rust-clippy/master/index.html#ptr_arg. If you want, I can revert these changes and either allow them on a per-function basis or globally.

src/erl.rs
```diff
@@ -77,28 +83,31 @@ struct Env<'a> {
 impl<'env> Env<'env> {
     pub fn new(module: &'env [String]) -> Self {
         Self {
-            current_scope_vars: Default::default(),
-            erl_function_scope_vars: Default::default(),
+            current_scope_vars: im::HashMap::default(),
```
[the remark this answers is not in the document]
brightly-salty: This was caused by https://rust-lang.github.io/rust-clippy/master/index.html#default_trait_access. If you want to, I can allow this lint on a per-function basis or globally.

src/typ/expr.rs
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#map_unwrap_or It worked in most cases, but was a false positive on this particular function.

src/typ/expr.rs
brightly-salty: Okay, I'll work on it.

src/erl.rs
```diff
@@ -767,28 +774,28 @@ fn var<'a>(name: &'a str, constructor: &'a ValueConstructor, env: &mut Env<'_>)
...
+            .into_doc()
             .append(atom(name.to_string()))
             .append("/")
             .append(*arity),
 
         ValueConstructorVariant::ModuleFn {
             arity,
             module,
-            name,
+            name: fn_name,
```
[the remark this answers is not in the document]
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

src/erl.rs
```diff
@@ -767,28 +774,28 @@ fn var<'a>(name: &'a str, constructor: &'a ValueConstructor, env: &mut Env<'_>)
...
             ..
         } => "fun "
-            .to_doc()
+            .into_doc()
             .append(module_name_join(module))
             .append(":")
-            .append(atom(name.to_string()))
+            .append(atom(fn_name.to_string()))
```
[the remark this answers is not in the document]
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

src/erl.rs
```diff
@@ -1102,7 +1102,7 @@ fn docs_args_call<'a>(
             ..
         } => {
             if let TypedExpr::Call {
-                fun,
+                fun: call_fun,
```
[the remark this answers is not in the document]
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

src/erl.rs
```diff
@@ -1116,7 +1116,7 @@ fn docs_args_call<'a>(
                         e => merged_args.push(expr(e, env)),
                     }
                 }
-                docs_args_call(fun, merged_args, env)
+                docs_args_call(call_fun, merged_args, env)
```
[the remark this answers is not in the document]
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#shadow_unrelated

src/error.rs
```diff
@@ -15,18 +15,23 @@ use termcolor::Buffer;
 pub type Src = String;
 pub type Name = String;
 
+#[allow(clippy::expect_used)]
```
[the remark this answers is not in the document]
brightly-salty: This function (fatal_compiler_bug), is used by gleam_expect, so I used regular expect so that we wouldn't have an infinite loop if there was a repeated bug in Buffer.

src/error.rs
```diff
@@ -239,13 +246,14 @@ fn did_you_mean(name: &str, options: &mut Vec<String>, alt: &'static str) -> Str
 }
 
 impl Error {
+    #[allow(clippy::unwrap_used, clippy::too_many_lines)]
```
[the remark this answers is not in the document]
brightly-salty: I'll work on fixing it. There are a lot of unwraps used in this function.

src/error.rs
```diff
@@ -1578,6 +1722,7 @@ fn std_io_error_kind_text(kind: &std::io::ErrorKind) -> String {
     }
 }
 
+#[allow(clippy::non_ascii_literal)]
```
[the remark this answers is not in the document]
brightly-salty: I'm allowing it globally now.

src/main.rs
```diff
@@ -43,11 +29,11 @@
     clippy::unimplemented,
     clippy::expect_used,
     clippy::unwrap_used,
-    clippy::ok_expect,
     clippy::integer_division,
     clippy::indexing_slicing,
     clippy::mem_forget
 )]
+#![allow(clippy::module_name_repetitions, clippy::non_ascii_literal)]
```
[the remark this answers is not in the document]
brightly-salty: You're right, I'm removing it from that one module.

src/main.rs
```diff
@@ -208,6 +199,7 @@ pub struct CompilePackage {
 }
 
 impl CompilePackage {
+    #[must_use]
```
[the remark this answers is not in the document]
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate
This prevents us from accidentally calling CompilePackage::new (or whichever function is under the attribute) and forgetting to use the result. Clippy lints (with the lint enabled) on pure functions to add `#[must_use]` to them.

src/metadata/module_builder.rs
```diff
@@ -1,18 +1,22 @@
...
-    },
-    fs::Writer,
-    schema_capnp::*,
-    typ::{
-        self, FieldMap, Type, TypeConstructor, TypeVar, ValueConstructor, ValueConstructorVariant,
-    },
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::convert::TryFrom;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
brightly-salty: Okay, will do.

src/typ/hydrator.rs
brightly-salty: https://rust-lang.github.io/rust-clippy/master/index.html#new_without_default

src/docs/tests.rs
lpil: The closure is optimised awayby the compiler, there no performance benefit so all this does is make it harder to write code that uses traits.

src/erl.rs
lpil: This seems to be a buggy lint as the suggested alternative is slower. Let's not enable this one

src/erl.rs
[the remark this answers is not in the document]
lpil: Let's disable it globally.

src/erl.rs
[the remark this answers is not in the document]
lpil: Even with that change we don't need to allocate a string to pass a reference to `get`, we can use the original reference. Here we can drop the `&` and the `.to_owned`

src/erl.rs
```diff
@@ -131,15 +142,15 @@ pub fn records(module: &TypedModule) -> Vec<(&str, String)> {
...
     let name = &name.to_snake_case();
     let escaped_name = if is_erlang_reserved_word(name) {
         format!("'{}'", name)
     } else {
         name.to_string()
     };
-    use std::fmt::Write;
     let mut buffer = format!("-record({}, {{", escaped_name);
-    for field in fields.iter().intersperse(&", ") {
+    for field in Itertools::intersperse(fields.iter(), &", ") {
```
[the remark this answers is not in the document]
lpil: Let's disable this lint globally. Rust adding an intersperse later is something we can deal with when it happens 👍

src/erl.rs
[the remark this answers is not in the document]
lpil: I'm cool with shadowing, let's disable this lint globally.

src/typ/expr.rs
lpil: If it is giving false positives let's disable this globally as it's not ready to be used.

src/typ/error.rs
```diff
@@ -237,6 +237,7 @@ pub enum Warning {
 }
 
 #[derive(Debug, PartialEq)]
+#[allow(clippy::pub_enum_variant_names)]
```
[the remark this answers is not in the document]
brightly-salty: https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/io/ObjectOutputStream.java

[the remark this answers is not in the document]
brightly-salty: I can change it if you want but just allowed for now.

lpil:
I'm going to close this one due to inactivity. If there are some changes you would like to continue with please open a new PR. Thank you!

PR #951, opened 2021-01-27
Canonicalize path on docs publish

### LINKED ISSUE #950: Source links are altered when published to hexdocs
Everything between the github repository root and the hash gets removed, though it is correct on the local copy. This doesn't seem to be the case with `gleam_stdlib`, so I'm not sure about steps to reproduce, other than to link the example for now.

https://hexdocs.pm/shellout/shellout/

Local copy generated with `gleam docs publish --version 0.1.0`
https://gist.github.com/tynanbe/7f5fa408bc33f562361f3ab30665d276

itsgreggreg:
Maybe fixes #950 

I didn't test this, it just looks like the problem and at the least might point in the right direction.

### DIFF
--- a/src/docs/command.rs
+++ b/src/docs/command.rs
@@ -5,7 +5,7 @@ use crate::{
 };
 use bytes::Bytes;
 use hexpm::Client;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Good idea, let's see if this helps.

PR #953, opened 2021-01-29
Adding more checks to bit string segment options

### LINKED ISSUE #949: Some ways bitstrings can crash or not compile correctly
These are all in declaration not patterns. I haven't investigated what will cause invalid patterns.

# Fixed
these are notes while i work on this

## Compiles but not what intended

An Expression in unit
```gleam
  <<3:size(1 + 1)-unit(1 + 1)>>
// Renders as
    <<3:(1 + 1)>>.
```

## Runtime Crash

Function that evaluates to a negative number
```
  let a = fn(b) { b - 4 }
  <<3:size(a(3))>>

// ** exception error: bad argument
```

## Compiles in Gleam, not in Erlang

A variable in unit 
```
let a = 4
<<3:size(2)-unit(a)>>
// Warning: variable 'A' is unused
// renders as 
//    A = 4,
//    <<3:2>>. 
```

A negative size
```gleam
<<3:size(-1)>>
<<3:-1>>
// syntax error before: '-'
```

A negative number in variable
```gleam
let a = -1
<<3:size(a)>>
// Warning: binary construction will fail because of a type mismatch
```

Any use of `unit` without an explicit size:
```gleam
<<10:unit(4)>>
// a bit unit size must not be specified unless a size is specified too
```

Size or unit given with utfx option
```
  <<"3":8-utf32>>
// neither size nor unit must be given for segments of type utf8/utf16/utf32
```

Size or unit given with utfx_codepoint
```
  assert Ok(c) = string.utf_codepoint(97)
  <<c:10-utf8_codepoint>>
// neither size nor unit must be given for segments of type utf8/utf16/utf32
```

### Erlang warning will crash

Bad size for a `float` type
```
  <<3.0:8-float>>
// Warning: bad float bit size
```

Using `binary` option on a non 8 bit divisible segment
```
  <<<<1:1>>:binary>>
// Warning: binary construction will fail with a 'badarg' exception (size 1 cannot be evenly divided by unit 8)
```

## Compiler panics
```
pub fn bts() {
  let <<<<1>>:bit_string>> = <<<<1>>:bit_string>>
  1
}

### PR DESCRIPTION
**Pull request by itsgreggreg**
Fixes #949 
(the linked documentation PR) fixes: #709

### Size option now has a guard on non int literals and is guaranteed to be non-negative.
Previously things like this could crash at runtime:
```rust
let a = -1
<<1:size(a)>>
let a = fn() { -1 }
<<1:size(a())>>
```

now the generated erlang has a "guard" in the form of `lists:max([_, 0])`
```erlang
A = -1,
<<1:(lists:max([(A), 0]))>>,
A@1 = fun() -> -1 end,
<<1:(lists:max([(A@1()), 0]))>>.
```

Integer literals are checked at compile time:
`<<1:-1>>` becomes `<<1:0>>` and `<<1:1>>` stays `<<1:1>>`

### Unit now must be an integer literal >= 1 and <= 256
Previously it allowed many erroneous values either ignoring them or crashing at runtime. Checked at compile time.

```
error: Syntax error
  ┌─ /Users/a/parser_test/src/a.gleam:2:20
  │
2 │   <<1:size(1)-unit(0)>>
  │                    ^ This is not a valid BitString unit value.

Hint: unit must be an integer literal >= 1 and <= 256
See: https://gleam.run/book/tour/bit-strings
```

a non int value is an unexpected token error.

### BitString patterns cannot be nested
This used to be a compiler panic, now syntax error

```
error: Syntax error
  ┌─ /Users/a/parser_test/src/a.gleam:2:9
  │
2 │   let <<<<1>>:bit_string>> = <<<<1>>:bit_string>>
  │         ^^^^^ BitString patterns cannot be nested.

See: https://gleam.run/book/tour/patterns
```

### Options that have no effect are disallowed
Hopefully helps people learn more about BitStrings

```
error: BitString Segment Error
  ┌─ /Users/a/parser_test/src/a.gleam:2:7
  │
2 │   <<1:signed>>
  │       ^^^^^^ This option is only allowed in BitString patterns.

Hint: This option has no effect in BitString values.
See: https://gleam.run/book/tour/bit-strings.html
```

### Binary is disallowed in values 
Gleam has no `Binary` type and the `bit_string` or `utfX` options cover existing types

```
error: BitString Segment Error
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:2:10
  │
2 │   <<"hi":binary>>
  │          ^^^^^^ This option is only allowed in BitString patterns.

Hint: This option has no effect in BitString values.
See: https://gleam.run/book/tour/bit-strings.html
```

### DIFF
--- a/src/erl/tests.rs
+++ b/src/erl/tests.rs
@@ -1727,10 +1727,8 @@ x() ->
 -spec main() -> bitstring().
 main() ->
     A = -1,
-    B = <<A:(lists:max([(A
-          * 2), 0]))/unit:2,
-          A:(lists:max([(3
-          + x()), 0]))/unit:1>>,
+    B = <<A:(lists:max([(A * 2), 0]))/unit:2,
+          A:(lists:max([(3 + x()), 0]))/unit:1>>,
     B.
 "#,
     );
--- a/src/erl/tests.rs
+++ b/src/erl/tests.rs
@@ -1687,9 +1687,9 @@ main() ->
         r#"fn main() {
   let a = 1
   let simple = <<1, a>>
-  let complex = <<4:int-unsigned-big, 5.0:little-float, 6:native-int-signed>>
+  let complex = <<4:int-big, 5.0:little-float, 6:native-int>>
   let <<7:2, 8:size(3), b:binary-size(4)>> = <<1>>
-  let <<c:unit(1), d:binary-size(2)-unit(2)>> = <<1>>
+  let <<c:8-unit(1), d:binary-size(2)-unit(2)>> = <<1>>
 
   simple
 }
@@ -1701,11 +1701,9 @@ main() ->
 main() ->
     A = 1,
     Simple = <<1, A>>,
-    Complex = <<4/integer-unsigned-big,
-                5.0/little-float,
-                6/native-integer-signed>>,
+    Complex = <<4/integer-big, 5.0/little-float, 6/native-integer>>,
     <<7:2, 8:3, B:4/binary>> = <<1>>,
-    <<C/unit:1, D:2/binary-unit:2>> = <<1>>,
+    <<C:8/unit:1, D:2/binary-unit:2>> = <<1>>,
     Simple.
 "#,
     );
@@ -1713,7 +1711,7 @@ main() ->
     assert_erl!(
         r#"fn x() { 2 }
 fn main() {
-  let a = 1
+  let a = -1
   let b = <<a:unit(2)-size(a * 2), a:size(3 + x())-unit(1)>>
 
   b
@@ -1728,8 +1726,11 @@ x() ->
 
 -spec main() -> bitstring().
 main() ->
-    A = 1,
-    B = <<A:(A * 2)/unit:2, A:(3 + x())/unit:1>>,
+    A = -1,
+    B = <<A:(lists:max([(A
+          * 2), 0]))/unit:2,
+          A:(lists:max([(3
+          + x()), 0]))/unit:1>>,
     B.
 "#,
     );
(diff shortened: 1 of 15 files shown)


src/parse/error.rs
```diff
@@ -36,18 +36,20 @@ pub enum ParseErrorType {
...
-    OpNakedRight,      // Operator with no value to the right
-    OpaqueTypeAlias,   // Type aliases cannot be opaque
-    TooManyArgHoles,   // a function call can have at most 1 arg hole
+    NotConstType, // :fn(), name, _  are not valid const types
+    OpNakedRight, // Operator with no value to the right
+    OpaqueTypeAlias, // Type aliases cannot be opaque
+    TooManyArgHoles, // a function call can have at most 1 arg hole
     UnexpectedEOF,
     UnexpectedReservedWord, // reserved word used when a name was expected
     UnexpectedToken { expected: Vec<String> },
```
itsgreggreg, line 55: Rust changed the comment formatting in here when I added a variant. This is why you don't auto align comments. :)

lpil: Alas!

lpil:
I'll wait until you're happy before reviewing. Thanks Greg!!

itsgreggreg:
companion documentation PR: https://github.com/gleam-lang/website/pull/85

src/erl/tests.rs
```diff
@@ -1728,8 +1726,11 @@ x() ->
 
 -spec main() -> bitstring().
 main() ->
-    A = 1,
-    B = <<A:(A * 2)/unit:2, A:(3 + x())/unit:1>>,
+    A = -1,
+    B = <<A:(lists:max([(A
+          * 2), 0]))/unit:2,
+          A:(lists:max([(3
+          + x()), 0]))/unit:1>>,
```
lpil, line 1733: This indentation seems a little strange. Could we have it break like this?

```erlang
    B = <<A:(lists:max([(A * 2), 0]))/unit:2,
          A:(lists:max([(3 + x()), 0]))/unit:1>>,
```

lpil (commented):
Looks great, though wow GitHub made a mess of that diff.

I've left one comment, let me know what you think.

itsgreggreg:
> Looks great, though wow GitHub made a mess of that diff

Yeah I probably should have created a new file and deleted the old one. The diff here probably doesn't help at all.

lpil (approved):
Fab, thank you!

PR #960, opened 2021-02-01
Better errors for malformed names (fixes #956)

### LINKED ISSUE #956: Provide a better warning when rejecting camelCase
camelCase variables are forbidden by the compiler to eliminate bikeshedding over naming conventions, but the current error doesn't communicate that at all:

```
error: Syntax error
    ┌─ /.../thing..gleam:115:18
    │
115 │ pub fn overlappedBy(compare: fn (a, a) -> Order, a: Interval(a), b: Interval(a)) -> Bool {
    │                  ^^ I was not expecting this.

Expected one of: "("
```

It'd be good if the error explained why this is a syntax error, if you're coming from most other languages this will be quite confusing otherwise. I suggest:

```
error: Syntax error
    ┌─ /.../thing..gleam:115:18
    │
115 │ pub fn overlappedBy(compare: fn (a, a) -> Order, a: Interval(a), b: Interval(a)) -> Bool {
    │                  ^^ I was not expecting this.

Gleam only allows lowercase letters and underscores in variable and type names. Try:

    overlappedby
    overlapped_by
```

Maybe with a little hint explaining its to avoid arguments over naming styles.

samontea:
This patch adds a new LexicalErrorType (`LexicalErrorType::CamelCaseName`) and throws it if the lexer encounters a "camel case" name. 

To implement this change it also allows LexicalErrors to pass `SrcSpan`'s as locations and it removes the `Copy`  train from the traits of `LexicalErrorType` and `LexicalError`. 

This fixes #956.

### DIFF
--- a/src/parse/error.rs
+++ b/src/parse/error.rs
@@ -1,4 +1,5 @@
 use crate::ast::SrcSpan;
+use heck::CamelCase;
 use heck::SnakeCase;
 
 #[derive(Debug, PartialEq, Clone)]
@@ -17,6 +18,7 @@ pub enum LexicalErrorType {
     UnrecognizedToken { tok: char },
     BadName { name: String },
     BadDiscardName { name: String },
+    BadUpname { name: String },
 }
 
 #[derive(Debug, PartialEq)]
@@ -96,6 +98,14 @@ impl LexicalError {
                 vec![
                     "Hint: Discard names start with _ and contain a-z, 0-9, or _.".to_string(),
                     format!("Try: _{}", name.to_snake_case()),
+                ],
+            ),
+            LexicalErrorType::BadUpname { name } => (
+                "This is not a valid upname.",
+                vec![
+                    "Hint: Upnames start with an uppercase letter and contain".to_string(),
+                    "only lowercase letters, numbers, and uppercase letters.".to_string(),
+                    format!("Try: {}", name.to_camel_case()),
                 ],
             ),
         }
--- a/src/parse/error.rs
+++ b/src/parse/error.rs
@@ -3,7 +3,7 @@ use crate::ast::SrcSpan;
 #[derive(Debug, PartialEq, Copy, Clone)]
 pub struct LexicalError {
     pub error: LexicalErrorType,
-    pub location: usize,
+    pub location: SrcSpan,
 }
 
 #[derive(Debug, PartialEq, Copy, Clone)]
--- a/src/parse/error.rs
+++ b/src/parse/error.rs
@@ -15,7 +15,7 @@ pub enum LexicalErrorType {
     RadixIntNoValue,       // 0x, 0b, 0o without a value
     UnexpectedStringEnd,   // Unterminated string literal
     UnrecognizedToken { tok: char },
-    CamelCaseName { name: String },
+    BadName { name: String },
 }
 
 #[derive(Debug, PartialEq)]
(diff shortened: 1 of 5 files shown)


itsgreggreg:
This is essentially exactly what I would have done, excellent work! A few little things:

Could we change the error `CamelCaseName` to `BadName`

Could we change the error text to be the same text as this one: https://github.com/gleam-lang/gleam/blob/main/src/error.rs#L1382

Could we add an error for DiscardNames called `BadDiscardName` with the hint of:
`Hint: Discard names start with _ and can contain a-z, 0-9, or _.`
calling `to_camel_case()` on discard names might erase the leading underscore, I'm not sure but something to look out for.


I know this is different than what is mentioned in #956, sorry for the run-around, I honestly just thought I would pick it up and didn't do enough research before making a suggestion!

src/parse/lexer.rs
```diff
@@ -452,6 +462,22 @@ where
         while self.is_name_continuation() {
             name.push(self.next_char().unwrap());
         }
+
+        // Throw a lexical error if a name is in "camel case".
```
itsgreggreg, line 466: nitpick, `// Return an error if an uppercase letter is used`

itsgreggreg:
Thinking about it a tiny bit more, this is going to stop lexing after a capital letter but before an underscore. So a name like this will not fully lex: `time_UTC_ms` it will only lex `time_UTC`, that's probably fine for showing the user what to fix but might look a little sloppy, we may need a function like this:

```rust
    fn is_name_error_continuation(&self) -> bool {
        self.chr0
            .map(|c| matches!(c, '0'..='9' | 'a'..='z' | 'A'..='Z' | '_' ))
            .unwrap_or(false)
    }

```
notice the underscore in there. Are there any other characters we want to collect and show as an error? `-`, `$`, `?`

src/parse/error.rs
```diff
@@ -78,6 +80,13 @@ impl LexicalError {
                 "I can't figure out what to do with this character.",
                 vec!["Hint: Is it a typo?".to_string()],
             ),
+	        LexicalErrorType::CamelCaseName { name } => (
+		        "This is not a valid name.",
+		        vec![
+			        "Hint: In Gleam names must start with a lowercase letter and contain only lowercase letters, numbers, and '_'.".to_string(),
```
lpil, line 86: Can we wrap this message onto two lines? Around 70 chars

```suggestion
			        "Hint: In Gleam names must start with a lowercase letter and contain".to_string(),
			        "only lowercase letters, numbers, and '_'.".to_string(),
```

lpil (commented):
Really nice work! I'm happy with this and have left one little note inline.

Greg's feedback looks good to me as well, though would be happy to leave the `BadDiscardName` for another PR/for someone else if you would prefer not to do this.

Thanks!

src/parse/error.rs
samontea: I'm unsure if it's still too long for you because I shortened the message per @\itsgreggreg's request. Let me know if you still think it should be shortened!

samontea:
I think I followed through on all the requested changes? I authored these commits kind of late so I may have misread or mistyped. Let me know if anything looks wrong!

I also added better error messages for malformed upnames (Upnames?) as some "extra credit". 😛

itsgreggreg:
lgtm!
I haven't taken a look at the but happy to help with the conflicts if you need!

lpil (approved):
Brilliant, thank you! This will be a nice user experience improvement.

Would you mind added a note to the CHANGELOG.md file and rebasing on main to resolve those conflicts? Then we can merge it in. Thank you

samontea:
Just rebased & mad the additions to the change log. ✅

lpil (approved):
Thank you very much!

PR #961, opened 2021-02-03
Lots of warning on unused things

itsgreggreg:
Creates warnings for essentially unused anything in a Module. Just does simple, scope-based usage counting so it's not exhaustive. Can't check mutually unused, etc.

Here's a couple files I've been using to tests this: https://gist.github.com/itsgreggreg/3afef44cd1a3e5b2ef9fa9810a5361b2

Happy to fix the tests and get this working in a direction that is good for Gleam. If this is headed in the wrong direction also all good, happy to close it. Just kind a went on a whim for getting a v1 of something like this working.

The gist is, for every environment scope a parallel `entity_usages` scope is created. Usages are initialized as statements and patterns are typed. Usages are increased as expressions are evaluated. When a scope is popped any of its entities with a usage count of 0 are turned into a warning.


## Examples
```                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
warning: Unused Variable                                                                                                                                                                                                                                                                                                                                                
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:19:11                                                                                                                                                                                                                                                                                                                      
   │                                                                                                                                                                                                                                                                                                                                                                    
19 │     tuple(e, 2) -> 5                                                                                                                                                                                                                                                                                                                                               
   │           ^ This variable is never used.                                                                                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                                                                                                        
Hint: you can ignore it with an underscore: `_e`.                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                        
warning: Unused Variable                                                                                                                                                                                                                                                                                                                                                
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:17:7                                                                                                                                                                                                                                                                                                                       
   │                                                                                                                                                                                                                                                                                                                                                                    
17 │   let d = 5                                                                                                                                                                                                                                                                                                                                                        
   │       ^ This variable is never used.                                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                                                                        
Hint: you can ignore it with an underscore: `_d`.                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                        
warning: Unused Variable                                                                                                                                                                                                                                                                                                                                                
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:16:6                                                                                                                                                                                                                                                                                                                       
   │                                                                                                                                                                                                                                                                                                                                                                    
16 │ fn b(c) {                                                                                                                                                                                                                                                                                                                                                          
   │      ^ This variable is never used.                                                                                                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                        
Hint: you can ignore it with an underscore: `_c`.                                                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                                                                        
warning: Unused Private Type Constructor                                                  
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:4:3                                           
  │                                                                                       
4 │   E(Int)                                                                              
  │   ^^^^^^ This private type constructor is never used.                                 

warning: Unused Imported Value                                                            
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:1:17                                          
  │                                                                                       
1 │ import b.{A, B, b_const, b_fun as b_fun_use}                                          
  │                 ^^^^^^^ This imported value is never used.                            

warning: Unused Private Function                                                          
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:10:1                                         
   │                                                                                      
10 │ fn a() {                                                                             
   │ ^^^^^^ This private function is never used.                                          

warning: Unused Private Function                                                          
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:23:1                                         
   │                                                                                      
23 │ ╭ external fn f() -> String =                                                        
24 │ │   "a" "b"                                                                          
   │ ╰─────────^ This private function is never used.                                     

warning: Unused Imported Value                                                            
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:1:26                                          
  │                                                                                       
1 │ import b.{A, B, b_const, b_fun as b_fun_use}                                          
  │                          ^^^^^ This imported value is never used.                     

warning: Unused Imported Item                                                             
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:1:11                                          
  │                                                                                       
1 │ import b.{A, B, b_const, b_fun as b_fun_use}                                          
  │           ^ This imported type constructor is never used.                             

warning: Unused Imported Type                                                             
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:1:14                                          
  │                                                                                       
1 │ import b.{A, B, b_const, b_fun as b_fun_use}                                          
  │              ^ This imported type is never used.                                      

warning: Unused Private Type                                                              
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:3:1                                           
  │                                                                                       
3 │ type D {                                                                              
  │ ^^^^^^ This private type is never used.                                               

warning: Unused Private Type Constructor                                                  
  ┌─ /Users/a/Trash/parser_test/src/a.gleam:5:3                                           
  │                                                                                       
5 │   C                                                                                   
  │   ^ This private type constructor is never used.                                      

warning: Unused Private Function                                                          
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:16:1                                         
   │                                                                                      
16 │ fn b(c) {                                                                            
   │ ^^^^^^^ This private function is never used.                                         

```

Also warns on variables overwritten in same scope with no usages:
```rust
fn a() {
  let b = 1
  let b = 2
  b
}
```


Warns with
```
warning: Unused Variable                                                                                                                                                 
   ┌─ /Users/a/Trash/parser_test/src/a.gleam:11:7                                                                                                                        
   │                                                                                                                                                                     
11 │   let b = 1                                                                                                                                                        
   │       ^ This variable is never used.                                                                                                                               
                                                                                                                                                                         
```

### DIFF
--- a/src/typ/environment.rs
+++ b/src/typ/environment.rs
@@ -1,6 +1,5 @@
 use super::*;
 use std::collections::HashMap;
-use std::collections::VecDeque;
 
 #[derive(Debug)]
 pub struct Environment<'a, 'b> {
@@ -30,17 +29,18 @@ pub struct Environment<'a, 'b> {
     // can safely be generalised.
     pub ungeneralised_functions: HashSet<String>,
 
-    // value_usages is a stack of value scopes. When a value is used
-    // it crawls up the scope stack for a value with that name and
-    // increments its usage.
-    pub value_usages: VecDeque<HashMap<String, (ValueKind, SrcSpan, usize)>>,
+    // entity_usages is a stack of scopes. When an entity is created it is
+    // added to the top scope. When an entity is used we crawl down the scope
+    // stack for an entity with that name and increments its usage.
+    pub entity_usages: Vec<HashMap<String, (EntityKind, SrcSpan, usize)>>,
 }
 
-// For Keeping track of value usages and knowing which error to display.
+/// For Keeping track of entity usages and knowing which error to display.
 #[derive(Debug, Clone)]
-pub enum ValueKind {
+pub enum EntityKind {
     PrivateConstant,
-    PrivateConstructor(String),
+    // String here is the type constructors type name
+    PrivateTypeConstructor(String),
     PrivateFunction,
     ImportedConstructor,
     ImportedType,
@@ -57,8 +57,8 @@ impl<'a, 'b> Environment<'a, 'b> {
         importable_modules: &'a HashMap<String, (Origin, Module)>,
         warnings: &'a mut Vec<Warning>,
     ) -> Self {
-        let mut value_usages = VecDeque::new();
-        value_usages.push_front(HashMap::new());
+        let mut entity_usages = vec![];
+        entity_usages.push(HashMap::new());
         let typer = Self {
             uid,
             level: 1,
@@ -71,7 +71,7 @@ impl<'a, 'b> Environment<'a, 'b> {
             importable_modules,
             current_module,
             warnings,
-            value_usages,
+            entity_usages,
         };
         register_prelude(typer)
     }
@@ -98,15 +98,18 @@ impl<'a, 'b> Environment<'a, 'b> {
 
     pub fn open_new_scope(&mut self) -> ScopeResetData {
         let local_values = self.local_values.clone();
-        self.value_usages.push_front(HashMap::new());
+        self.entity_usages.push(HashMap::new());
         self.level += 1;
         ScopeResetData { local_values }
     }
 
     pub fn close_scope(&mut self, data: ScopeResetData) {
         self.level -= 1;
-        let unused = self.value_usages.pop_front().unwrap();
-        self.handle_unused_local_values(unused);
+        let unused = self
+            .entity_usages
+            .pop()
+            .gleam_expect("There was no top entity scope.");
+        self.handle_unused(unused);
         self.local_values = data.local_values;
     }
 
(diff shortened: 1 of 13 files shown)


src/typ/error.rs
```diff
@@ -183,19 +183,58 @@ pub enum Error {
...
+
+    UnusedPrivateFunction {
+        location: SrcSpan,
+        name: String,
+    },
+
+    UnusedVariable {
+        location: SrcSpan,
+        name: String,
+    },
```
itsgreggreg, line 237: most of this is reformatting because I added `imported: bool` to `UnusedConstructor` and `UnusedType`

src/typ/expr.rs
```diff
@@ -1404,21 +1404,23 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
...
+                            location: *location,
+                            name: name.to_string(),
+                            variables: self
+                                .environment
+                                .local_values
+                                .keys()
+                                .map(|t| t.to_string())
+                                .collect(),
+                        }),
+                }?;
```
itsgreggreg, line 1423: Had to rearrange this because `get_module_const` now needs an `&mut self` instead of just an `&self` in order to record the const usage

src/typ/environment.rs
```diff
@@ -127,12 +140,13 @@ impl<'a, 'b> Environment<'a, 'b> {
         name: String,
         variant: ValueConstructorVariant,
         typ: Arc<Type>,
+        origin: SrcSpan,
     ) {
         let _ = self.local_values.insert(
             name,
             ValueConstructor {
                 public: false,
-                origin: Default::default(), // TODO: use the real one
```
itsgreggreg, line 135: The random location lines here and there are for removing this `TODO`. The initial implementation I went for used this function as an entry point so I had to have the correct origin. The current implementation does not use this function but I kept the changes related to that `TODO`. Happy to take them out if it crowds this PR or if another solution is desired.

lpil: Sounds good to me!

src/typ/environment.rs
```diff
@@ -21,24 +22,32 @@ pub struct Environment<'a, 'b> {
...
 
     // Functions that have not yet been inferred then generalised.
     // We use this to determine whether functions that call this one
     // can safely be generalised.
     pub ungeneralised_functions: HashSet<String>,
+
+    // value_usages is a stack of value scopes. When a value is used
+    // it crawls up the scope stack for a value with that name and
+    // increments its usage.
+    pub value_usages: VecDeque<HashMap<String, (ValueKind, SrcSpan, usize)>>,
```
lpil, line 36: Nit: I think this is only accessed from the front, so we can use a `Vec` and push/pop from the back.

src/typ/environment.rs
```diff
@@ -21,24 +22,32 @@ pub struct Environment<'a, 'b> {
...
+}
+
+// For Keeping track of value usages and knowing which error to display.
+#[derive(Debug, Clone)]
+pub enum ValueKind {
```
lpil, line 41: I wonder if there's a better name than `ValueKind` as it is for both types and values.

- `Named`
- `Defined`
- `VariableKind`
- ???

Same goes for the `value_usages`.

src/typ/environment.rs
```diff
@@ -467,21 +482,113 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+
+    fn increment_usage(&mut self, name: &str) {
+        let mut val_found = false;
+        let mut cascade = None;
+        for scope in self.value_usages.iter_mut() {
+            if !val_found {
+                match scope.get_mut(name) {
+                    Some((ValueKind::PrivateConstructor(type_name), _, usages)) => {
+                        val_found = true;
+                        cascade = Some(type_name.clone());
```
lpil, line 579: Could we add comments or more names to this to make it clearer?

My first questions were:

- Is PrivateConstructor a type or value constructor?
- What is `type_name`?
- Why do we decend into the next level for this one variant?

src/warning.rs
```diff
@@ -112,35 +112,104 @@ be an update. Remove the update or remove fields that need to be copied.")
                     .unwrap();
                 }
 
-                Warning::UnusedType { location, name } => {
+                Warning::UnusedType {
+                    location, imported, ..
+                } => {
+                    let title = if *imported {
```
lpil, line 118: Can we have only the first word capitalised in these warning messages? To match the others

src/typ/environment.rs
```diff
@@ -467,21 +482,113 @@ impl<'a, 'b> Environment<'a, 'b> {
...
-        for (name, location) in self.unused_private_mixed_constructors.drain() {
-            self.warnings
-                .push(Warning::UnusedConstructor { name, location })
+    }
+
+    pub fn init_value_usage(&mut self, name: String, kind: ValueKind, location: SrcSpan) {
+        match self
+            .value_usages
+            .front_mut()
+            .unwrap()
```
lpil, line 550: Let's call `.gleam_expect` here do there's a nice message if it crashes due to a fatal bug.

src/typ/environment.rs
```diff
@@ -467,21 +482,113 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+            .insert(name.to_string(), (kind, location, 0))
+        {
+            Some((kind, location, 0)) => {
+                // a value was overwritten in the top most scope without being used
```
lpil, line 554: Cool!

src/typ/environment.rs
```diff
@@ -467,21 +482,113 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+        self.increment_usage(name);
     }
 
     pub fn value_used(&mut self, name: &str) {
-        let _ = self.unused_private_mixed_constructors.remove(name);
+        self.increment_usage(name);
+    }
+
+    fn increment_usage(&mut self, name: &str) {
+        let mut val_found = false;
```
lpil, line 572: In this function we can use `break` to stop iteration when we find the value rather than continuing but skipping the body with an `if`

lpil (commented):
Looking good! A marked improvement on the previous version.

I've left some notes inline.

src/typ/environment.rs
itsgreggreg: Hows `EntityKind` and `entity_usages` sound?

lpil: That's great!

itsgreggreg:
Marking this ready for review as the tests are passing and I've added a few. Happy to keep pushing this in whatever direction is good for Gleam.

src/typ/environment.rs
```diff
@@ -466,22 +484,112 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+    }
+
+    /// Increments an entity's usage in the current or nearest enclosing scope
+    pub fn increment_usage(&mut self, name: &str) {
+        let mut cascade = None;
+        for scope in self.entity_usages.iter_mut().rev() {
+            match scope.get_mut(name) {
+                Some((EntityKind::PrivateTypeConstructor(type_name), _, usages)) => {
+                    // If a type constructor is used, we consider
+                    // its type also used
```
lpil, line 514: Sorry I'm not understanding this. Could you give an example of the code that this handles? Thank you

src/typ/environment.rs
```diff
@@ -466,22 +484,112 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+        let unused = self.entity_usages.pop().unwrap();
+        self.handle_unused(unused);
     }
 
-    pub fn value_used(&mut self, name: &str) {
-        let _ = self.unused_private_mixed_constructors.remove(name);
+    fn handle_unused(&mut self, unused: HashMap<String, (EntityKind, SrcSpan, usize)>) {
+        for val in unused {
+            match val {
+                (name, (EntityKind::ImportedType, location, 0)) => {
```
lpil, line 542: Why is it that we track the number of times use instead of deleting the entry from the hashset if it is used? I can only see the `0` case handled, so do we care about the others?

lpil (commented):
Looks great! I've left two questions. They're more for my understanding than problems :)

src/typ/environment.rs
itsgreggreg: It cascades the usage of a type constructor to a usage of it's type.
Otherwise this code would leave the private type as unused:
```rust
type A {
  B
}
pub fn a() {
  let b = B
  case b {
    B -> 1
  }
}
```

src/typ/environment.rs
itsgreggreg: Because if we removed the entity from the scope on usage, the next usage would result in a variable in a wider scope getting counted as used. Essentially this has to do with value name shadowing, for example:
```
const a = 1

fn b(a) {
// | argument gets counted as used here and removed
// |   | const gets counted as used here and removed
   a + a
}
```

itsgreggreg: That being said, a count is probably not necessary, a boolean is probably good enough.

src/typ/environment.rs
lpil: Ah yes, thank you

src/typ/environment.rs
lpil: Ah yes, that makes sense. Thank you

lpil (approved):
Thank you, great stuff as per usual. Mind adding an entry to the changelog?

lpil:
Thank you!

PR #965, opened 2021-02-04
Dark mode and other documentation improvements

tynanbe:
**Highlights**

- Dark mode with toggle and local storage (default off)
- Pre-wrap with toggle and local storage (default off)
- Mimimal, extensible options framework
- Gleam.js module
- SVG icon library
- Cross-browser compatibility (tested on recent Chromium, Firefox)
- Mobile compatibility (tested on Android 7, iOS 12, 13)
- Minor usability and style tweaks

**Demo**

https://hexdocs.pm/shellout/


lpil (commented):
Wow, brilliant work! This is really nice functionality and the styling decisions you've made look gorgeous.

Could we see what it looks like with a darker shadow around the code snippets in dark mode? It stands out quite a bit on my screen (though less so on my ipad).

It took me and the person I tested with some time to work out what the wrapping button did, and I'm not sure it's worth having overall. Is this something found elsewhere? I couldn't find any examples, and I think I would prefer it to always scroll.

Is there a reason for some of the JS being in the HTML and some in a file?

Thank you

tynanbe:
> Wow, brilliant work! This is really nice functionality and the styling decisions you've made look gorgeous.

Thank you!

> Could we see what it looks like with a darker shadow around the code snippets in dark mode? It stands out quite a bit on my screen (though less so on my ipad).

See 08ac825

> It took me and the person I tested with some time to work out what the wrapping button did, and I'm not sure it's worth having overall. Is this something found elsewhere? I couldn't find any examples, and I think I would prefer it to always scroll.

I was certain I'd seen at least one documentation generator with this feature, but I can't remember or find one now, so I will just convey some thoughts on the matter.

**Pros**

- One can find this feature in many text editors/IDEs, if not any popular doc generators.
- Mobile performance can be severely impacted by large, horizontally scrollable content areas, in my experience — e.g. viewing source code on GitHub via a mobile device can easily freeze the browser for several minutes, repetitiously as the user continues scrolling in any direction.
- Selection of portions of text where horizontal scrolling is required mid-selection can be frustrating, even near impossible in one go, on Mobile devices.

**Cons**

- Some contemporary browsers hide scrollbars entirely, particularly on mobile devices. That said, the styling changes coming with this pull request make it much more obvious when a snippet is overflowing its container (horizontal padding on `<code>` instead of `<pre>`).
- This feature comprises a tiny amount of code, and could be easily removed.

**Other**

- This feature is mostly useful for mobile browsers. If kept, it might be prudent to hide the toggle until the layout collapses.
- In my tests on iOS 12 and 13, I was unable to convince the browser to do anything but `white-space: pre-wrap`-like behavior. Some searching suggested this is a known issue with Safari in general.

Ultimately it's not my call of course, and I'm not really attached to it either way, but it can be rather nice when casually scrolling down a page, that the reader does not need to stop and horizontally scroll every snippet, particularly if the user can understand the code either way. In summary there are valid reasons to have this feature, but disable it by default.

> Is there a reason for some of the JS being in the HTML and some in a file?

The JS after the opening `<body>` tag initializes the user's selected options before any content is rendered, e.g. avoiding a flash of content with the wrong theme. The JS at the end of the body was moved to a separate file for caching purposes. More of the JS in the template could probably be moved as well.

The early options loading and related local storage logic is mostly just adapted from mdBook, as seen on https://gleam.run/book/tour/

Currently there is one bug I overlooked and need to fix, whereby a device that communicates a preference for dark mode will default to `theme-dark`, but the toggle works improperly thereafter.

Finally, it would be ideal if the Gleam compiler version were passed to the template, so static asset links can use it in query strings, ensuring browsers will automatically get fresh, uncached css/js as needed, and preventing confusing broken states arising from running newer html with older, cached css/js. However, I am not experienced enough with Rust to handle that side of the equation; help would be appreciated :)

michallepicki:
Overall superb work

I don't find the "tag" icon to be intuitive, if I didn't know there is a version selector underneath I wouldn't even try clicking on that area...

itsgreggreg:
Beautiful work. Love that the core colors work in both light and dark mode. Also agree that maybe the tag icon needs a little down arrow next to it. Here's my other thoughts for what they're worth:
Screen Shot 2021-02-05 at 11 26 00 AM

tynanbe:
Thanks for the feedback! I've updated the branch to address some of these and other issues.

The current state of the branch is reflected in the demo:
https://hexdocs.pm/shellout/shellout/

Items of note:

- Changed tag icon to history icon — functionality should be intuitive now.
- Made star bullet points slightly smaller — since type constructors are an important feature that sets Gleam apart from its BEAM peers, it seems prudent to give them a little love and distinguish them from other lists that might be in the doc.
- Switched order of settings toggles, theme is now first.
- Made "prewrap" toggle show only on small screens — this option toggles wrapping long source code lines (discussed above).
- Updated toggle tooltips for clarity.
- Created 2 custom Gleam icons to replace module name's pink underline and add to section headings, making them easier to identify/locate (and for fun/aesthetics), especially when scrolling through heavier docs.
- Fixed the default dark mode bug, theme selection now shows proper icon when device conveys dark mode preference.
- Refactored some JS to fix dark mode bug, consolidate settings code (reduce maintenance overhead), and for consistency.
- Handled long project, module, and member titles.

<details>
<summary>A summary screenshot</summary>

gleam-docs-xs
</details>

> Expanding on "might be better up here?". Just thinking about where to put it so that it stays in place regardless of content size. Navigation elements and page controls getting pushed down the page based on content length isn't my fav.

Hopefully the summary screenshot conveys why the header is a less-than-ideal location for the settings. Currently the best locations are top/bottom of content/sidebar, in my opinion. Since the current options are likely "set and forget" for most people, they probably don't need a prominent position; however I might suggest the top left corner of the content area otherwise.

All features have been re-tested and work well enough for me. The last thing needed for this to be production-ready is passing the Gleam compiler version string to templates and adding query strings to local static assets.

itsgreggreg:
The jagged header underline is a bit much for my eyes, I think no underlines at all works well.

The text is too big in the mobile screenshot. The top line and h1 should probably be able to fit at least ~30 characters across on the minimum supported screen size. That's a job for another PR but imo it doesn't indicate that page controls have to go at the bottom of a table of contents. Imo while browsing code on mobile should of course be possible, optimizing for larger screens should be the focus of the docs as they are going to be most useful while developing which is typically done on a larger device.

Re the bullets. It's currently confusing to me that the type is listed with its constructors in a code block and then the constructors are listed individually below that. (It only really makes sense when there are doc comments for a constructor.) Having the stars there makes it seem like there's another thing is going on. Web users have pre-trained cognitive associations with common page icons and elements. The star outline has a few pre-trained associations and list item is unfortunately not one. Regular bullets (or maybe no bullets) indicate to me it's just a list. IMO calling specific attention to type constructors in library docs isn't going to help people learn gleam or distinguish gleam from other languages on the beam. That being said I do think that Gleam doc pages should have a link in the footer that says something like "a Gleam library" with a link explaining what Gleam is and how to run it. I'm sure lots of people are going to get to Gleam libraries from Hex and not know what they've stumbled upon. I just don't think stars as bullets are going to help with that. Maybe the little faded stars icon you've put on the bottom right of the page can be expanded to say "A Gleam Library" or something.

Imo the history icon still isn't clear enough. I personally didn't recognize it as a history icon. It just looks like a clock with an arrow to me. Maybe undo? People are going to be navigating both backward and forward in versions with the version selector. I think a downward pointing triangle is the way to go.

I recognize I sound like a buzz kill here, all the fun stuff is being critiqued, but doc pages are going to get a lot of use by people of all skill levels and simplicity, usability, and intuition are going to factor heavily into their usefulness. What makes elixir's or rust's docs so world class is not their design quirkiness but their usefulness and imo that's where Gleam should focus it's efforts as well. The dark mode is great and looks awesome, the rest of the changes are a bit too much in my opinion.

lpil:
> Made "prewrap" toggle show only on small screens — this option toggles wrapping long source code lines (discussed above).

That's a nice approach. I dig it.

> I think a downward pointing triangle is the way to go.

I think this could be clearest. Does the icon pack have a suitable arrow icon?

> The last thing needed for this to be production-ready is passing the Gleam compiler version string to templates and adding query strings to local static assets.

I'm happy for this to be another PR if you want.

I think the stars look better now they're a bit smaller, but don't have any strong feelings on them either way. I think they and the underlines look nice, but am interested in what others think.

image

I think this underline could do with being longer to match the word "Functions"

tynanbe:
I'm sure everyone here agrees about the importance and goals of good documentation. I also doubt anyone has ever refused to use a language because its docs were too stylish :smile:

Primarily I set out to address the issue that bright white backgrounds hurt my eyes, especially given increased screen time. I could have stopped there and kept my userstyles to myself, but opened this PR to try and share something with the community and simultaneously make the docs a bit more accessible. As work progressed I tried to give the dark and light themes a unified design language, and ended up fixing/tweaking various broken things that got in the way. For the most part I changed very little from the existing templates where modification was unnecessary.

In the present day I'd reckon that many of us consume developer docs on mobile devices, as a lot of development time is spent thinking and researching available tools/libraries to begin with. The mobile/desktop paradigm has become increasingly blurred, making it apparent that we ought really to care about the usability and ease of access of our content from most any device — how can the screen space be used efficiently, yet cohesively across devices.

As for aesthetics, differing opinions are a given. Mostly we should consider how design decisions impact understanding and effective utilization of the content. Developers can also appreciate turn-key polished, professional documentation, no design skills needed, whether it suits their personal stylistic preferences or not. The design decisions I've made are not intended to be "quirky"; rather to achieve the aforementioned and to help define an ecosystem, and perhaps even to help attract attention or generate interest. In that regard, a link to the Gleam homepage, or docs page, would be a great addition.

Finally I did not set a mobile font size, nor did I create the footer, I just worked with what was there. I see the point about star bullets being assumed interactive elements, and agree that the constructor subsection in general could be re-examined. For select boxes, the implementation is browser-dependant, and does not always behave like a dropdown menu, making a down arrow or triangle less intuitive/appropriate, but probably still understandable. For the sketchy underlines, would it be generally preferable if the color were less intense — I disagree that the section headers are better with no visual demarcation?

lpil:
This is looking really good! There is still a few things which we may decide that we want to change stylistically but I don't want that to get in the way of merging in this work. I'm going to merge this in, and then we can get feedback and iterate in later pull requests￼￼ if need be.

Love the improvements here! A Real big difference

Thank you￼￼

lpil (approved):

Issue #967, opened 2021-02-05
Warnings as errors flag


itsgreggreg:
A flag to turn compile time warnings into compile time errors.

That something people are interested in?

Ex:
`gleam build --warnings-as-errors`

PR #969, opened 2021-02-07
Warnings as errors compiler build flag

itsgreggreg:
Adds a `--warnings-as-errors` flag to the `build` command. Which optionally turns compile time warnings into compile time errors.

Closes https://github.com/gleam-lang/gleam/issues/967

Side question: Is there a reason why the `build` command is hidden from the `help` message?

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -160,6 +160,10 @@ pub enum Error {
     MetadataDecodeError {
         error: Option<String>,
     },
+
+    GenericFileDiagnostics {
+        diagnostics: Vec<(Diagnostic, String)>,
+    },
 }
 
 impl From<capnp::Error> for Error {
@@ -1562,6 +1566,15 @@ but it cannot be found.",
                 };
 
                 write_project(buffer, diagnostic);
+            }
+
+            Error::GenericFileDiagnostics { diagnostics } => {
+                for (diagnostic, extra) in diagnostics {
+                    write(buffer, diagnostic.clone(), Severity::Error);
+                    if !extra.is_empty() {
+                        writeln!(buffer, "{}\n", extra).expect("error pretty buffer write");
+                    }
+                }
             }
         }
     }
--- a/src/error.rs
+++ b/src/error.rs
@@ -161,8 +161,8 @@ pub enum Error {
         error: Option<String>,
     },
 
-    GenericFileDiagnostics {
-        diagnostics: Vec<(Diagnostic, String)>,
+    ForbiddenWarnings {
+        count: usize,
     },
 }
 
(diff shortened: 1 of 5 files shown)


GentileFulvio (commented):
I noticed that there aren't any tests for the added functionality ? I could be wrong though.

Sorry for this bold comment but I was looking at the issue and noticed that a PR was created and wondered how one would add this functionality to the code.

itsgreggreg:
@GentileFulvio good lookin out. I'm not too sure how to test this. The existing tests are for functionality deeper than `main.rs`. Maybe Gleam needs public interface tests?

lpil:
We could take an approach like the core language tests and add a CI step that runs on a project in `test` with some warnings in it and assert that it returns non-zero.

We could probably do some some form of CLI level integration testing, but it's not been hugely cost effective to do this yet.

lpil:
> Side question: Is there a reason why the build command is hidden from the help message?

People keep getting confused by it, thinking it is a replacement for rebar3. It's going to be deprecated in a few versions time and be replaced with the new `gleam compile-package` command and the new Gleam build tool.

src/error.rs
```diff
@@ -160,6 +160,10 @@ pub enum Error {
     MetadataDecodeError {
         error: Option<String>,
     },
+
+    GenericFileDiagnostics {
```
lpil, line 164: I think if we pre-render the warnings and put them here we won't be able to have multiple formats for errors later (i.e. LSP, JSON, GitHub Actions diagnostic thingies). The preciseness of the warnings previously worked well for this.

Perhaps an approach like this?

```rust
    let any_warnings = warning::print_all(analysed.as_slice());
    if warnings_as_errors && any_warnings {
        return Err(Error::ForbiddenWarnings);
    }
```
```
error: Warnings detected

The warnings above were found when compiling this project and the `--warnings-as-errors`
flag has been set so we can't continue compilation. Fix the warnings and try again.
```

GentileFulvio:
> We could take an approach like the core language tests and add a CI step that runs on a project in `test` with some warnings in it and assert that it returns non-zero.
> 
> We could probably do some some form of CLI level integration testing, but it's not been hugely cost effective to do this yet.

And testing the `as_errors` function in `src/warning.rs` is not possible as well I guess ? Because of the `&[crate::project::Analysed]` type right ?

lpil:
We could construct that type and test it if we wanted it. It's just about deciding what we want to test. Currently we don't have any tests around the format that errors get printed with

lpil (approved):
Very nice, thank you

lpil:
I've rebased these onto main (had to fix a merge conflict). Thanks!

PR #972, opened 2021-02-08
Version strings for local static assets

tynanbe:
Following up on #965, added the compiler's version string to local static assets for cache busting purposes.
Updated the changelog.


lpil (approved):
Oh clever, nice!

PR #973, opened 2021-02-09
add hex plugin to library projects

CrowdHailer:


lpil:
What does this enable?

CrowdHailer:
It is needed to publish a package to hex

lpil:
Is that right? How come I haven't needed that for anything I have
published? For example the stdlib

CrowdHailer:
Maybe we're publishing different ways?
Maybe your local setup has it installed globally?

lpil:
The Hex docs say this should be added globally, so I think that's the way to go https://hex.pm/docs/rebar3_publish

PR #982, opened 2021-02-19
Always use forward slashes in docs source paths

### LINKED ISSUE #954: Docs source links path separators depend on the OS rendered on
On windows we get urls like this: `https://github.com/gleam-lang/stdlib/blob/v0.13.0/src\gleam\map.gleam#L18-L18`

GitHub corrects for this thankfully, but we should update this to always render correctly. Likely a matter of building the string from the path ourselves rather than relying on Rust's path to string code.

evuez:
This should fix #954.

First PR to the repo and haven't written any Rust in a while, sorry if this is a mess... It seems to work, but not sure if I'm forgetting some cases!

### DIFF
--- a/src/docs/source_links.rs
+++ b/src/docs/source_links.rs
@@ -4,7 +4,7 @@ use crate::{
     line_numbers::LineNumbers,
     project::Analysed,
 };
-use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


lpil (approved):
Wonderful, thank you

PR #991, opened 2021-03-02
Allow numbers in project name

CrowdHailer:
This change also prevents `_` being the first character in a project name, which is what the error message already says.


lpil (approved):
Much better, thanks!

Issue #985, opened 2021-02-21
Improve error messages for string concat


inoas:
In some languages that either/and/or feature any of this: auto-casting, weak-types, operator overloading, `+` may be used to concat strings.

Similar to the warning that `+` is for integer addition (where `+.` is for float addition) the gleam compiler could emit a warning when `+` is used to attempt to concat strings and reference the correct operator to be used.

lpil:
Good idea! Thank you.

For the int based operators we suggest the appropriate operator, but we don't have a string concat operator. What could we suggest instead? `string.concat` perhaps. We would need to note that this is a library function rather than being in Gleam core.

PR #992, opened 2021-03-06
Error message for string concat

syukronrm:
Related issue #985 

Produce hint when trying to use `+` and `+.` operator for string:
```
error: Type mismatch
  ┌─ /home/me/projects/gleam/experiments/test_str_concat/src/test_str_concat.gleam:4:9
  │
4 │   100 + "100"
  │         ^^^^^

The + operator expects arguments of this type:

    Int

But this argument has this type:

    String

Hint: String can be concatted using `string.concat` from stdlib


===> Unable to compile Gleam project
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -777,9 +777,8 @@ But this argument has this type:
                         given = printer.pretty_print(given, 4),
                     )
                     .unwrap();
-                    if let Some((alt, t)) = hint_alternative_operator(&op, given.as_ref()) {
-                        writeln!(buf, "Hint: the {} operator can be used with {}s\n", alt, t)
-                            .unwrap();
+                    if let Some(t) = hint_alternative_operator(&op, given.as_ref()) {
+                        writeln!(buf, "Hint: {}\n", t).unwrap();
                     }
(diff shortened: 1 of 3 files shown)


src/error.rs
```diff
@@ -1672,26 +1671,37 @@ fn import_cycle(buffer: &mut Buffer, modules: &[String]) {
...
         _ => None,
     }
 }
+
+fn hint_numeric_message(alt: &str, type_: &str) -> String {
+    format!("the {} operator can be used with {}s\n", alt, type_)
+}
+
+fn hint_string_message() -> String {
+    "String can be concatted using `string.concat` from stdlib".to_string()
```
lpil, line 1706:
```suggestion
    "Strings can be joined using the `append` or `concat` functions from
the `gleam/string` module".to_string()
```

lpil (commented):
Looks great! Would you mind updating the changelog as well? Thank you

syukronrm:
Updated!

lpil (approved):
Wonderful! Thank you

PR #994, opened 2021-03-12
Fix grammar in InvalidProjectName error message

wezm:
I tried to create a project with the name `hello-gleam` and received this error:

```
% gleam new hello-gleam

error: Invalid project name

We were not able to create your project as `hello-gleam` is
does not have the correct format. Project names must start
with a lowercase letter and may only contain lowercase letters,
numbers and underscores.

Please try again with a different project name.
```

I stumbled over "We were not able to create your project as `hello-gleam` **is does** not have the correct format.". This PR fixes the grammar in the message.


lpil (approved):
Much better, thank you

PR #1008, opened 2021-04-03
Fix table alternative row colour in dark mode

michaeljones:
I don't know if this is the best colour choice but it seems better than
'pink-white' which is the colour at the moment.

We don't have many tables in the documentation so this might have been
missed on the initial introduction of dark mode.

Before:
image

After:
image


lpil (approved):
Oh dear! Well spotted, this is a big improvement

Issue #1005, opened 2020-10-29
Different tuple syntax


sporto:
I use tuples all the time. Having to write `tuple(a, b)` feels verbose.
It would be great to have a string literal for tuples. e.g. `(a, b)` or something that will be coherent with the rest of Gleam.

Thanks

lpil:
Let's go with `#()`!

PR #1011, opened 2021-04-06
Add alternative `#(x, y, ...)` tuple syntax

tranzystorek-io:
Adds a new tuple syntax, as per #1005

Retains backwards compatibility for now (two possible syntax variants are accepted)


lpil (approved):
Very nice!

PR #1012, opened 2021-04-07
Refactor some clippy lints

tranzystorek-io:


src/type_.rs
```diff
@@ -1536,7 +1533,7 @@ pub fn register_import(
 
             // Determine local alias of imported module
             let module_name = match &as_name {
-                None => module[module.len() - 1].clone(),
+                None => module.last().unwrap().clone(),
```
lpil, line 1536: Let's use `.gleam_expect` here.

Actually it would be good to disallow use of unwrap if possible.

lpil (commented):
Nice! Thank you

Can we remove the use of `bool::then`? I think this is unclear to people less familiar with Rust.

src/type_.rs
tranzystorek-io: Done

tranzystorek-io:
Removed `bool::then` as requested ✅

lpil (approved):
Thank you

PR #1013, opened 2021-04-07
Add version updates to release checklist

michaeljones:
So that various other projects can be kept up to date with new releases.

I hope this is helpful. I've not made a PR for the Dockerfile yet but came across it the other day. 

An alternative approach to this would be to convert it to an issue template and that might allow for a markdown checklist so you can tick off the items as you go. I like checklists. Might not be worth the gain though :)


lpil (approved):
Thank you

PR #1014, opened 2021-04-07
Do a general refactor

tranzystorek-io:
Highlights:

- Simplify `for` loops (redundant `.into_iter`s, etc.)
- Remove `.as_slice()`, `as_ref()`, `as_str()` where deref coercion applies
- Use free functions from `itertools` for readability

PR #1015, opened 2021-04-08
Refactor Environment::handle_unused

tranzystorek-io:
Didn't make it in time to add one more change to #1014 😅

Eliminates boilerplate `self.warnings.push(...)` code


lpil (approved):
Oooh very slick. I dig it

PR #1025, opened 2021-04-10
Add more refactoring

tranzystorek-io:
Highlights:

- Replace the `&*variable` deref pattern with verbose `.deref()` methods, or remove if unnecessary
- `erl::record()`: rewrite iterator adaptors to be more readable/simpler
- `Environment::increment_usage()`: simplify the cascading logic to use the `while let` loop
- use `Itertools`' convenience wrappers, `map_ok` and `try_collect`

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -132,26 +132,25 @@ pub fn records(module: &TypedModule) -> Vec<(&str, String)> {
     module
         .statements
         .iter()
-        .flat_map(|s| match s {
+        .filter_map(|s| match s {
             Statement::CustomType {
                 public: true,
                 constructors,
                 ..
-            } => &constructors[..],
-            _ => &[],
+            } => Some(constructors),
+            _ => None,
         })
+        .flatten()
         .filter(|constructor| !constructor.arguments.is_empty())
-        .flat_map(|constructor| {
-            let mut fields = Vec::with_capacity(constructor.arguments.len());
-            for RecordConstructorArg { label, .. } in &constructor.arguments {
-                match label {
-                    Some(s) => fields.push(&**s),
-                    None => return None,
-                }
-            }
-            Some((&constructor.name, fields))
+        .filter_map(|constructor| {
+            constructor
+                .arguments
+                .iter()
+                .map(|RecordConstructorArg { label, .. }| label.as_deref())
+                .collect::<Option<Vec<_>>>()
+                .map(|fields| (constructor.name.as_str(), fields))
         })
-        .map(|(name, fields)| (name.as_ref(), record_definition(name, &fields[..])))
+        .map(|(name, fields)| (name, record_definition(name, &fields)))
         .collect()
 }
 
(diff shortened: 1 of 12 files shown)


src/type_/environment.rs
```diff
@@ -512,28 +511,24 @@ impl<'a, 'b> Environment<'a, 'b> {
...
-                Some(val) => {
-                    val.2 = true;
-                    break;
+        let mut next_name = name.to_string();
+
+        while let Some((kind, _, used)) = self
+            .entity_usages
+            .iter_mut()
+            .rev()
+            .find_map(|entity| entity.get_mut(&next_name))
```
lpil, line 520: This isn't an entity, it is a scope.
```suggestion
            .find_map(|scope| scope.get_mut(&next_name))
```

src/type_/environment.rs
```diff
@@ -512,28 +511,24 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+        {
+            *used = true;
+
+            match kind {
+                // if a type constructor is used, we consider its type also used
```
lpil, line 525:
```suggestion
                // If a type constructor is used, we consider its type also used
```

src/type_/environment.rs
```diff
@@ -512,28 +511,24 @@ impl<'a, 'b> Environment<'a, 'b> {
...
-                    if type_name != name {
-                        cascade = Some(type_name.clone());
-                    }
-                    *usages = true;
-                    break;
-                }
-                Some(val) => {
-                    val.2 = true;
-                    break;
+        let mut next_name = name.to_string();
```
lpil, line 514: Let's call this `name` as it's not the next name when the loop is evaluated.
```suggestion
        let mut name = name.to_string();
```

lpil (commented):
Very nice! I've left a few small comments

src/type_/environment.rs
tranzystorek-io: ✅

src/type_/environment.rs
tranzystorek-io: ✅

src/type_/environment.rs
tranzystorek-io: ✅

src/erl.rs
```diff
@@ -580,7 +579,7 @@ fn expr_segment<'a>(
     let size = |expression: &'a TypedExpr, env: &mut Env<'a>| match expression {
         TypedExpr::Int { value, .. } => {
             let v = value.replace("_", "");
-            let v = u64::from_str(&v).unwrap_or(0);
+            let v = u64::from_str(&v).unwrap_or_default();
```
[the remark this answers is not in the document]
tranzystorek-io: ✅

src/erl.rs
```diff
@@ -492,15 +491,15 @@ fn atom<'a>(value: String) -> Document<'a> {
...
-        // Escape because of keyword collision
-        value if is_erlang_reserved_word(value) => Document::String(format!("'{}'", value)),
-
-        // No need to escape
-        _ if RE.is_match(&value) => Document::String(value),
-
-        // Escape because of characters contained
-        _ => Document::String(value).surround("'", "'"),
+    if is_erlang_reserved_word(&value) {
+        // escape because of keyword collision
```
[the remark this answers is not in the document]
tranzystorek-io: ✅

CrowdHailer:
I'm copying as much of erl.rs as possible. are there any big changes I should be aware of in this PR?

lpil:
@CrowdHailer it's polishing really, nothing essential to port across.

lpil (approved):
Nice!

PR #1027, opened 2021-04-11
Adding JavaScript as a Gleam compilation target.

CrowdHailer:
## Items to complete.

- [ ] Integration with compiler 
- [ ] Compiler warnings for unsupported features.
- [ ] reinstate warnings as errors flag.
- ... more to be added later.

There are several things which I am **not** tackling at the moment. Including
- BitString syntax 
- Selective code gen for external functions/types depending on target. It is sufficient for prototyping having packages without external compile well on both targets but for packages with external fn integration to only work on one target 
- Tail call optimisation or implementing immutable lists. Probably both of these are very useful but I want to treat them as an optimisation on top of this first effort.

*Note: I have confused myself on the PR's but this one is opened as a draft and is here to stay.*

### Integration with compiler.

This is the main thing I would like feedback on first.

I don't fully understand the differentiation between project_compiler and package_compiler. At this point I have integrated with package_compiler. To generate JavaScript files I have added a `--target` flag to the cli any value except js will result in the erlang code gen. i.e.
```
gleam compile-package --src tmp --out tmp/gen --name tessa --target js
```

1. Is package_compiler the correct integration point?
2. Is target the correct name for the flag, alternatives platform or backend?
3. Is there an error type that exists for marking features as unsupported, it would be good to get a nice compiler error in the case of someone compiling to JS a program that uses the bitstring syntax.


src/main.rs
```diff
@@ -1,4 +1,4 @@
-#![deny(warnings)]
+// #![deny(warnings)]
```
CrowdHailer, line 1: Obviously this change will be undone once the PR is taken out of draft.

src/main.rs
```diff
@@ -194,9 +195,21 @@ pub struct NewOptions {
...
 }
 
+// arg_enum! {
+//     #[derive(Debug)]
+//     pub enum Target {
+//         Erlang,
+//         JavaScript
+//     }
+// }
+// https://stackoverflow.com/questions/54687403/how-can-i-use-enums-in-structopt
```
CrowdHailer, line 205: I could not work out what Rust wanted from me here. 
Does it make sense to specify an enum for the target or we could just pass around strings as currently implemented?

tranzystorek-io: I think an enum is clean :) If we want a "js" option, we're probably better off implementing `FromStr` manually than using `arg_enum!`

lpil: Here's an existing example of how we do this with strum https://github.com/gleam-lang/gleam/blob/main/src/new.rs#L16-L23

src/js/tests.rs
```diff
@@ -0,0 +1,540 @@
...
+#[test]
+fn string_literals() {
+    assert_js!(
+        r#"
+fn go() {
+    "Hello, Gleam!"
+}
+"#,
+r#"function go() {
+    return "Hello, Gleam!"
```
lpil, line 35: Let's use semicolons, otherwise we'd need to be clever and detect when code is being generated for which semicolons are mandatory.

src/js.rs
```diff
@@ -0,0 +1,568 @@
...
+    type_::{
+        ModuleValueConstructor, PatternConstructor, Type, TypeVar, ValueConstructor,
+        ValueConstructorVariant,
+    },
+    Result,
+};
+use itertools::Itertools;
+use std::sync::Arc;
+
+const INDENT: isize = 4;
```
lpil, line 20: Can we use 2? That seems to be common in JS

src/js.rs
```diff
@@ -0,0 +1,568 @@
...
+            collect_cons(tail, elements)
+        }
+        // TODO is it possible to write improper lists in Gleam
+        other => Some(other)
+    }
+}
+
+fn tuple<'a>(elems: impl Iterator<Item = Document<'a>>) -> Document<'a> {
+    concat(Itertools::intersperse(elems, break_(",", ", ")))
+        .nest_current()
```
lpil, line 289: Can we indent normally rather than nest at the current level? I don't think JS uses that style of indentation.

lpil:
> Is there an error type that exists for marking features as unsupported, it would be good to get a nice compiler error in the case of someone compiling to JS a program that uses the bitstring syntax.

For this we'll need to add a new error variant and have the JS codegen functions return an `Err(TheNewError)` if those features are used.

src/js.rs
CrowdHailer: Might do this at the end. Rather a pain as the test.rs file is automatically indenting to 4 as I move stuff around

src/js/tests.rs
CrowdHailer: Done,

src/main.rs
```diff
@@ -194,9 +195,27 @@ pub struct NewOptions {
...
+    JavaScript
+}
+
 #[derive(StructOpt, Debug)]
 #[structopt(flatten)]
 pub struct CompilePackage {
+    #[structopt(
+        help = "The compilation target for the generated project", 
+        long = "target", 
+        possible_values = &Target::VARIANTS,
```
tranzystorek-io, line 214: `possible_values` is a clap method, StructOpt should deal with it automatically, so this line is not needed

src/js.rs
```diff
@@ -0,0 +1,761 @@
...
+            ..
+        } => {
+            // println!("args: {:?}", args);
+            let fields = match field_map {
+                Some(FieldMap{fields, ..}) => {
+                    // fields.into_iter()
+                    // TODO Can I guarantee the order here?
+                    
+                    let t = fields
+                    .iter();
```
CrowdHailer, line 366: How can I get a vector of map keys ordered by the value they contain?

The `fields` value here looks like `{"name": 0, "age": 1}`
I want the ordered names so I can zip with the ordered args and generate the code `{type: "Cat", name: "example", age: 1}`

tranzystorek-io:
```rs
fields.iter().sorted_by(|(_, &v)| v).map(|(name, _)| name)
```

is roughly what you need

CrowdHailer: Well it's sounds obvious when you put it like that.  Thank you.

lpil: Just in case it's not obvious, not all arguments are guaranteed to have labels so positional and labelled arguments will need to be merged.

Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+        && make install \
+        && rm -rf /usr/src/gleam-src
+
+FROM node:15.14.0
+
+COPY --from=build /usr/local/cargo/bin/gleam /bin
+RUN gleam --version
+
+WORKDIR /opt/app
+CMD ["bash"]
```
CrowdHailer, line 16: This is just a Dockerfile that lives in my repo while I'm developing it. I'm sharing it here in case it's helpful for anyone debugging the errors I get.

Currently I'm running
```
gleam compile-package --src . --out gen --name tessa --target javascript
// then
sirv . --host 0.0.0.0 --dev
```

@lpil is there a watch option for `gleam compile-package` (or quick hack) you could think of so I can just change my gleam code and refresh the page and get the updated JS?

lpil: I use `watchexec` so it's `watchexec -- gleam compile-package`, or whatever command you want to run

https://github.com/watchexec/watchexec

src/js/tests.rs
```diff
@@ -0,0 +1,751 @@
...
+macro_rules! assert_js {
+    ($src:expr, $erl:expr $(,)?) => {
+        // println!("\n\n\n{}\n", $src);
+        let (mut ast, _) = crate::parse::parse_module($src).expect("syntax error");
+        ast.name = vec!["the_app".to_string()];
+        let mut modules = std::collections::HashMap::new();
+        // Probably a nicer way to do types values accessors
+        modules.insert(
+            "rocket_ship".to_string(), 
+            (Origin::Src, type_::Module { name: vec!["rocket_ship".to_string()], types: std::collections::HashMap::new(), values: std::collections::HashMap::new(), accessors: std::collections::HashMap::new()})
```
CrowdHailer, line 14: @tranzystorek-io do you have any clever idea for adding module content here.

I use the module rocket_ship below to test imports but I can't yet test, the following. The module has no functions and building it by hand looks a pain.
```
import rocket_ship.{launch}
```

tranzystorek-io: If it's only about making a hashmap literal, then the [maplit](https://docs.rs/maplit/1.0.2/maplit/) crate is the solution

CrowdHailer: Thanks, I had a look at that. In the end I co-opted the parse/compile machinery and generate a dummy module from a string.

lpil:
Gunna close this in favour of the other PRs.

PR #1028, opened 2021-04-11
Clean up dependencies

tranzystorek-io:
Include `strum` and `serde` macros implicitly with the `derive` feature


lpil (approved):
Thank you

PR #1052, opened 2021-04-16
Add the option of compiling to JS, actual implementation a no-op

CrowdHailer:


src/js.rs
```diff
@@ -0,0 +1,14 @@
...
+};
+
+pub fn module(
+    _module: &TypedModule,
+    _line_numbers: &LineNumbers,
+    _writer: &mut impl Utf8Writer,
+) -> Result<()> {
+    Err(UnsupportedFeature {
+        target: crate::Target::JavaScript,
+        feature: "Any feature! ".to_string(),
```
CrowdHailer, line 12: Obviously, this message will start saying something more useful from the very next PR

CrowdHailer:
@lpil is this what I think the scope of the first PR should be, there's no associated issue so far.

This allows the following
```
$gleam compile-package --src tmp --out temp/gen --name tessa --target javascript

error: Unsupported feature for compilation target

Any feature!  is not supported for javascript compilation
```

lpil (approved):
Thank you!

PR #1053, opened 2021-04-16
Rendering string output for JS

CrowdHailer:


src/js/statements.rs
```diff
@@ -0,0 +1,29 @@
...
+        })),
+        Statement::ExternalType { .. } => None,
+        Statement::ModuleConstant { .. } => Some(Err(UnsupportedFeature {
+            target: Target::JavaScript,
+            feature: "Adding a module constant ".to_string(),
+        })),
+        Statement::Fn { .. } => unimplemented!("TODO this PR"),
+        Statement::ExternalFn { .. } => Some(Err(UnsupportedFeature {
+            target: Target::JavaScript,
+            feature: "Using an external function ".to_string(),
```
lpil, line 26:
```suggestion
            feature: "Using an external function".to_string(),
```

src/js/statements.rs
```diff
@@ -0,0 +1,29 @@
...
+        Statement::TypeAlias { .. } => None,
+        Statement::CustomType { .. } => None,
+        Statement::Import { .. } => Some(Err(UnsupportedFeature {
+            target: Target::JavaScript,
+            feature: "Importing modules ".to_string(),
+        })),
+        Statement::ExternalType { .. } => None,
+        Statement::ModuleConstant { .. } => Some(Err(UnsupportedFeature {
+            target: Target::JavaScript,
+            feature: "Adding a module constant ".to_string(),
```
lpil, line 21:
```suggestion
            feature: "Adding a module constant".to_string(),
```

src/js/statements.rs
```diff
@@ -0,0 +1,29 @@
...
+    statement: &'a TypedStatement,
+    _module: &'a [String],
+    _line_numbers: &'a LineNumbers,
+) -> Option<Result<Document<'a>>> {
+    match statement {
+        Statement::TypeAlias { .. } => None,
+        Statement::CustomType { .. } => None,
+        Statement::Import { .. } => Some(Err(UnsupportedFeature {
+            target: Target::JavaScript,
+            feature: "Importing modules ".to_string(),
```
lpil, line 16:
```suggestion
            feature: "Importing modules".to_string(),
```

src/js/statements.rs
```diff
@@ -0,0 +1,145 @@
...
+
+fn unsupported<M: ToString, T>(label: M) -> Result<T> {
+    Err(UnsupportedFeature {
+        target: Target::JavaScript,
+        feature: label.to_string(),
+    })
+}
+
+pub fn statement<'a>(
+    _current_module: &'a [String],
```
CrowdHailer, line 17: Some of these unused variables are here because the function signatures match the erl.rs version and I think they get used in imminent PR's

src/js/tests.rs
```diff
@@ -0,0 +1,28 @@
+mod strings;
+
+#[macro_export]
+macro_rules! assert_js {
```
CrowdHailer, line 4: This is "mostly" copy pasted from erl.rs I think that's probably best than finding a way to reuse the logic but switch on target somehow.

src/js/tests/strings.rs
```diff
@@ -0,0 +1,16 @@
+use super::super::*;
+use crate::assert_js;
```
CrowdHailer, line 2: I'm not sure exposing the macro on the crate is the neatest. But I was struggling to access the macro from a sub module and this was a solution that works.

CrowdHailer:
@lpil this is ready for review. I've made this PR about strings, and that is the test case I have created.

Note that to make this work I needed to implement functions, but it is my intention to write tests that cover all the different ways to call a function and args vs named args in a separate "functions" PR

src/js.rs
```diff
@@ -1,14 +1,30 @@
...
-    ast::*, error::Error::UnsupportedFeature, fs::Utf8Writer, line_numbers::LineNumbers, Result,
-};
+mod statements;
+#[cfg(test)]
+mod tests;
+
+use crate::{ast::*, fs::Utf8Writer, line_numbers::LineNumbers, pretty::*, Result};
+use itertools::Itertools;
+
+const INDENT: isize = 4;
```
lpil, line 8: Let's make this 2 as it seems to be more common.
```suggestion
const INDENT: isize = 2;
```

src/js/statements.rs
```diff
@@ -0,0 +1,145 @@
...
+    _line_numbers: &'a LineNumbers,
+) -> Result<Document<'a>> {
+    let env = Env {
+        return_last: &true,
+        semicolon: &true,
+    };
+    let body = expr(body, &env)?;
+    Ok(if &true == public { "export " } else { "" }
+        .to_doc()
+        .append("function ")
```
lpil, line 64:
```suggestion
    Ok(if *public { "export function " } else { "function " }
        .to_doc()
```

src/js/statements.rs
```diff
@@ -0,0 +1,145 @@
...
+    }?;
+    Ok(match expression {
+        TypedExpr::Seq { .. } | TypedExpr::Assignment { .. } => rendered,
+        _ => match env.return_last {
+            true => "return ",
+            _ => "",
+        }
+        .to_doc()
+        .append(rendered)
+        .append(match env.semicolon {
```
lpil, line 136: I think rather than considering `;` part of the expression I would prefer to have it be part of the JS generation for `Seq`, `Assignment` and `Fn` AST nodes. That way we don't need to mutate this state and check each time.

src/js/tests/strings.rs
lpil, line 2: Can we just have a `tests.rs` module to match the other modules?

lpil:
Looking really good so far! Few small things in the comments

src/js/tests/strings.rs
CrowdHailer: I'd rather break it up. I got to 800+ lines on #1027 and I'm sure I was quite a long way short of done.

I wasn't able to work through the erl tests, because there are different edge cases you need to consider and things like named arguments behave very differently. Currently I'm breaking it up into files per language feature, matching the chapters of https://gleam.run/book/tour/

src/js/statements.rs
CrowdHailer: I tried that, it might be possible, but I struggled to make it work.
Because JS doesn't do blocks properly I rendered Gleam blocks to comma expression in JS, at which point you have let expressions that end in a comma.

src/js/tests/strings.rs
lpil: OK 👍

src/js/statements.rs
lpil: Cool I'll see if I can work something out.

lpil (approved):
Thank you! Exciting to have some of this on the main branch!

CrowdHailer:
It very much is,

PR #1055, opened 2021-04-16
implement number literals

CrowdHailer:


src/javascript/expression.rs
```diff
@@ -113,10 +115,66 @@ impl Generator {
...
+            println!("name: {:?}", name);
+            unimplemented!("binop")
+        }
+    }
+}
+
+fn print_bin_op<'a>(left: &'a TypedExpr, right: &'a TypedExpr, op: &str) -> Output<'a> {
+    let left_expr = match left {
+        TypedExpr::BinOp { .. } => Ok(Generator::compile(left)?.surround("(", ")")),
+        _ => Generator::compile(left),
```
CrowdHailer, line 163: @lpil is there some other way to use the `expression` function that you were intending?
I want to re use it here because `left` (and `right`) are terms are `TypedExpr` however using `Generator::compile` now mean all my tests are failing with a return on each line.

CrowdHailer:
Handling div by zero.

I think that ternary expressions are the cleanest way to do this.
```rust
a / b # Int
p /q # floats
```
```js
var gleam$tmp = b
gleam$tmp === 0 ? 0 : Math.floor(a / gleam$tmp)

var gleam$tmp = q
gleam$tmp === 0 ? 0 : p / gleam$tmp
```
tmp variable making sure expression b & q are only evaluated once.

src/javascript/expression.rs
lpil: The same generator is intended to be used for the entire module (as it can carry any state we need such as whether it is in the tail postion). This is failing because the state is being discarded and reset with each new generator so it is back to being in the tail position state.

These functions should be methods on the generator, like in the formatter with the printer.

i.e.

```rust
fn bin_op<'a>(&mut self, left: &'a TypedExpr, right: &'a TypedExpr, op: &str) -> Output<'a> {
    let left_expr = match left {
        TypedExpr::BinOp { .. } => Ok(self.expression(left)?.surround("(", ")")),
// ...
```

src/javascript/expression.rs
```diff
@@ -113,10 +115,66 @@ impl Generator {
...
+        BinOp::LtEqInt | BinOp::LtEqFloat => print_bin_op(left, right, "<="),
+        // https://dmitripavlutin.com/how-to-compare-objects-in-javascript/
+        // BinOp::Eq => "=:=",
+        // BinOp::NotEq => "/=",
+        BinOp::GtInt | BinOp::GtFloat => print_bin_op(left, right, ">"),
+        BinOp::GtEqInt | BinOp::GtEqFloat => print_bin_op(left, right, ">="),
+        BinOp::AddInt | BinOp::AddFloat => print_bin_op(left, right, "+"),
+        BinOp::SubInt | BinOp::SubFloat => print_bin_op(left, right, "-"),
+        BinOp::MultInt | BinOp::MultFloat => print_bin_op(left, right, "*"),
+        BinOp::DivInt => Ok(Document::String("Math.floor".to_string())
```
lpil, line 149: This doesn't need to be a string, we can use a `str` and avoid the extra heap allocation

src/javascript/expression.rs
```diff
@@ -113,10 +115,66 @@ impl Generator {
     }
 }
 
+fn int(value: &str) -> Document<'_> {
+    value.to_doc()
```
lpil, line 119: Let's fail if this is a binary, octal, or hex number. Also let's use the binary operator to hint that it's an int

```rust
let x = 1
```
```js
var x = 1 | 0;
```

lpil:
> Handling div by zero.
> 
> 
> 
> I think that ternary expressions are the cleanest way to do this.
> 
> ```rust
> 
> a / b # Int
> 
> p /q # floats
> 
> ```
> 
> ```js
> 
> var gleam$tmp = b
> 
> gleam$tmp === 0 ? 0 : Math.floor(a / gleam$tmp)
> 
> 
> 
> var gleam$tmp = q
> 
> gleam$tmp === 0 ? 0 : p / gleam$tmp
> 
> ```
> 
> tmp variable making sure expression b & q are only evaluated once.

I've just discovered that we can do this with || to avoid the variable

```JavaScript
1 / 0 || 0
```

src/javascript/expression.rs
CrowdHailer: Ah yes, I think there is to_doc() function that works

src/javascript/expression.rs
CrowdHailer: The  binary, octal, or hex all work in JavaScript.
```
0o12 === 10
true
0xF === 15
true
0b11 === 3
true
```

I would argue against providing the integer hint. I was following the reasoning that we should as much as possible generate the JS someone would write. It's not idiomatic handwritten JS to have those hints. Also I thought those hints where useful while Web Assembly was being developed, but if optimising for web assembly wouldn't we just compile to web assembly.

lpil: Wow that's cool! I have no idea you could do that in JS.

It's not super common, but it will avoid any bugs coming from not having real ints, and it will generate faster code than leaving it off. Microsoft even recommend adding it to hand-written JS https://docs.microsoft.com/en-us/previous-versions/windows/apps/hh781219(v=win.10)?redirectedfrom=MSDN#use-integer-arithmetic-where-possible

src/javascript/expression.rs
CrowdHailer: Ok, I think I follow. Getting a bit OO'y with all this use of `self` not really an approach I think about any more but looks like it makes sense in rust land.

Bit busy this week so might not get round to it till the weekend.

lpil: Fab, thank you

CrowdHailer:
> I've just discovered that we can do this with || to avoid the variable

@lpil I think you mean with a single `|`
For me in firefox double `|` returns infinity.

src/javascript/expression.rs
CrowdHailer: Added `| 0` for division it also handles the infinity so no explicit check needed.

lpil:
Oh yes, that's right. I'm not sure if there's a similar trick for floats

CrowdHailer:
Ready for review, had more time than I thought.

src/javascript/tests/numbers.rs
```diff
@@ -0,0 +1,118 @@
...
+    2.0 >=. 1.0 // => True
+    2.0 <=. 1.0 // => False
+}
+"#,
+        r#""use strict";
+
+function go() {
+  1.0 + 1.4;
+  5.0 - 1.5;
+  (function(x) { return isFinite(x) ? x: 0})(5.0 / 2.0);
```
CrowdHailer, line 109: This was the smallest way I could think of handling infinities and floats. At least the expression is only printed once.

lpil: Makes sense, thanks

lpil (approved):
Fab, thank you!

PR #1057, opened 2021-04-16
Add general refactoring

tranzystorek-io:
Highlights:

- `erl::variable_name`: make the function a bit more readable
- `error::did_you_mean`: simplify the function and its signature
- `pretty::args_to_gleam_doc`: decrease code nesting for clarity
- general clean-up around matching, `Option`s etc.

### DIFF
--- a/src/ast/typed.rs
+++ b/src/ast/typed.rs
@@ -143,10 +143,11 @@ impl TypedExpr {
         lazy_static! {
             static ref NON_ZERO: Regex = Regex::new(r"[1-9]").unwrap();
         }
-        match self {
-            Self::Int { value, .. } | Self::Float { value, .. } => NON_ZERO.is_match(value),
-            _ => false,
-        }
+
+        matches!(
+            self,
+            Self::Int{ value, .. } | Self::Float { value, .. } if NON_ZERO.is_match(value)
+        )
     }
 
     pub fn location(&self) -> SrcSpan {
--- a/src/error.rs
+++ b/src/error.rs
@@ -255,21 +255,14 @@ impl FileKind {
     }
 }
 
-fn did_you_mean(name: &str, options: &mut Vec<String>, alt: &'static str) -> String {
-    // Remove magic variables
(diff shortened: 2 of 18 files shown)


src/ast/typed.rs
```diff
@@ -143,10 +143,11 @@ impl TypedExpr {
...
             static ref NON_ZERO: Regex = Regex::new(r"[1-9]").unwrap();
         }
-        match self {
-            Self::Int { value, .. } | Self::Float { value, .. } => NON_ZERO.is_match(value),
-            _ => false,
-        }
+
+        matches!(
+            self,
+            Self::Int{ value, .. } | Self::Float { value, .. } if NON_ZERO.is_match(value)
```
lpil, line 149: TIL you can use guards like this

src/error.rs
```diff
@@ -207,9 +207,7 @@ pub enum StandardIoAction {
 
 impl StandardIoAction {
     fn text(&self) -> &'static str {
-        match self {
-            StandardIoAction::Read => "read from",
-        }
+        "read from"
```
lpil, line 210: Undo this please so that when new variants are added we get a type error and the code gets correctly updated.

src/error.rs
```diff
@@ -255,21 +253,14 @@ impl FileKind {
...
-    });
-
-    match options.get(0) {
-        Some(option) => format!("did you mean `{}`?", option),
-        None => alt.to_string(),
-    }
+    options
+        .iter()
+        .filter(|&option| option != crate::ast::CAPTURE_VARIABLE)
+        .min_by_key(|option| strsim::levenshtein(option, name))
```
lpil, line 261: Very nice

lpil (commented):
Very nice! Thank you. There's one bit I would like reverted to ensure I don't miss updates in future, but otherwise great stuff

src/error.rs
tranzystorek-io: ✅

lpil (approved):
Thank you

PR #1059, opened 2021-04-20
implement external functions

CrowdHailer:
tackles #1049

### DIFF
--- a/src/javascript.rs
+++ b/src/javascript.rs
@@ -132,15 +132,10 @@ fn external_function<'a, T>(
     };
     let arguments = wrap_args(arguments.iter().enumerate().map(|a| match a {
         (index, ExternalFnArg { label, .. }) => {
-            let arg_name = label.clone().unwrap_or(format!("arg{}", index));
-            Document::String(arg_name)
+            label.as_ref().map(Documentable::to_doc).unwrap_or_else(|| Document::String(format!("arg{}", index)));
         }
     }));
-    let body = module
-        .to_doc()
-        .append(".")
-        .append(fun)
-        .append(arguments.clone());
+    let body = docvec!["return ", module, ".", fun, arguments.clone()];
     Ok(docvec![
         head,
         name,
--- a/src/javascript.rs
+++ b/src/javascript.rs
@@ -179,14 +179,12 @@ impl<'a> Generator<'a> {
         } else {
             "function "
         };
-        let arguments = wrap_args(arguments.iter().enumerate().map(|a| match a {
-            (index, ExternalFnArg { label, .. }) => {
-                let arg_name = label.clone().unwrap_or(format!("arg{}", index));
-                Document::String(arg_name)
-                // label
-                //     .as_ref()
-                //     .map(Documentable::to_doc)
-                //     .unwrap_or_else(|| Document::String(format!("arg{}", index)))
+        let arguments = wrap_args(arguments.iter().enumerate().map(|a| {
+            match a {
+                (index, ExternalFnArg { label, .. }) => label
+                    .as_ref()
+                    .map(|l| l.as_str().to_doc())
+                    .unwrap_or_else(|| Document::String(format!("arg{}", index))),
             }
         }));
         let body = docvec!["return ", module, ".", fun, arguments.clone()];
--- a/src/javascript.rs
+++ b/src/javascript.rs
@@ -132,7 +132,12 @@ fn external_function<'a, T>(
     };
     let arguments = wrap_args(arguments.iter().enumerate().map(|a| match a {
         (index, ExternalFnArg { label, .. }) => {
-            label.as_ref().map(Documentable::to_doc).unwrap_or_else(|| Document::String(format!("arg{}", index)));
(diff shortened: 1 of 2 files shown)


src/javascript.rs
```diff
@@ -111,6 +118,40 @@ fn module_function<'a>(
...
+    arguments: &'a [ExternalFnArg<T>],
+    module: &'a str,
+    fun: &'a str,
+) -> Output<'a> {
+    let head = if public {
+        "export function "
+    } else {
+        "function "
+    };
+    let arguments = wrap_args(arguments.iter().enumerate().map(|a| match a {
```
CrowdHailer, line 133: I tried to look for a clever way to reuse some of the code in `module_function` but in the end just keeping it separate seemed clearest.

src/javascript/tests.rs
```diff
@@ -174,3 +174,30 @@ const e = [
...
+
+function foo() {
+  document.foo()
+}
+
+export function bar(arg0, arg1) {
+  document.bar(arg0, arg1)
+}
+
+function baz(x, y) {
```
CrowdHailer, line 198: I use the name of the arguments if they are given, to make the output nicer.

lpil: Very nice

src/javascript.rs
```diff
@@ -111,6 +118,40 @@ fn module_function<'a>(
...
+        (index, ExternalFnArg { label, .. }) => {
+            let arg_name = label.clone().unwrap_or(format!("arg{}", index));
+            Document::String(arg_name)
+        }
```
lpil, line 137: I think we can remove the cloning of the name here

```suggestion
        (index, ExternalFnArg { label, .. }) => {
            label.as_ref().map(Documentable::to_doc).unwrap_or_else(|| Document::String(format!("arg{}", index)));
        }
```

src/javascript.rs
```diff
@@ -111,6 +118,40 @@ fn module_function<'a>(
...
+    }));
+    let body = module
+        .to_doc()
+        .append(".")
+        .append(fun)
+        .append(arguments.clone());
```
lpil, line 143: Need a `return` too. Pesky statement based language.
```suggestion
    let body = docvec!["return ", module, ".", fun, arguments.clone()];
```

src/javascript.rs
lpil: I like this aye

lpil (commented):
Looks good! Could you rebase on main? I changed the structure a tiny bit to hold some extra state

src/javascript.rs
CrowdHailer: Have added and got.

```
the trait `pretty::Documentable<'_>` is not implemented for `&std::string::String`
```

src/javascript.rs
```diff
@@ -119,6 +126,41 @@ impl<'a> Generator<'a> {
...
+        let head = if public {
+            "export function "
+        } else {
+            "function "
+        };
+        let arguments = wrap_args(arguments.iter().enumerate().map(|a| match a {
+            (index, ExternalFnArg { label, .. }) => {
+                let arg_name = label.clone().unwrap_or(format!("arg{}", index));
+                Document::String(arg_name)
+                // label
```
CrowdHailer, line 147: This is the code, from you suggestion, moved after merging main. I wasn't sure how to use it so have left my original code as it was functioning.

src/javascript/tests.rs
```diff
@@ -181,3 +181,30 @@ const e = [
...
+pub external type Thing
+external fn foo() -> Thing = "document" "foo"
+pub external fn bar(Int, String) -> Thing = "document" "bar"
+external fn baz(x: Int, y: String) -> Thing = "document" "baz"
+ 
+"#,
+        r#""use strict";
+
+function foo() {
+  return document.foo()
```
lpil, line 198: Oh, I just had a thought. We're probably going to have to `import { foo } from "document";` ?

CrowdHailer: No, don't think so. My example project I had working in the browser didn't require any of that.
I guess bundlers might need separate thinking about, but in those cases it's probably a setting in the bundler to flag what globals are referenced.

lpil: Let's do this in a later PR

lpil (approved):

src/javascript.rs
lpil: OK, will need to turn the string into a str first
```rust
label.as_ref().map(|l| l.as_str().to_doc()).unwrap_or_else(|| Document::String(format!("arg{}", index)));
```

src/javascript.rs
lpil: This would be the new definition of `arg_name`

lpil:
Could you resolve the merge conflicts? Seems to be disabled for me

src/javascript/tests.rs
lpil: Actually I just had an idea how how externals could work that may work well and means we don't need to do this

CrowdHailer:
There seems to be a problem with the option to allow you to edit as I am using the midas-framework org. https://github.com/isaacs/github/issues/1681 
Will do it myself now

Issue #1045, opened 2021-04-15
JS: Operators


lpil:
For expressions and clause guards

- [x] Greater than `>`
- [x] Greater than `>.`
- [x] Greater or equal `>=`
- [x] Greater or equal `>=.`
- [x] Less than `<`
- [x] Less than `<.`
- [x] Less or equal `>=`
- [x] Less or equal `>=.`
- [ ] Boolean and `&&`
- [ ] Boolean or `||`
- [x] Add `+`
- [x] Add `+.`
- [x] Subtract `-`
- [x] Subtract `-.`
- [x] Multiply `*`
- [x] Multiply `*.`
- [x] Divide `/`
- [x] Divide `/.`
- [x] Modulo `%`

CrowdHailer:
All numerical comparisons are tackled in #1055

CrowdHailer:
This Issue should be closed as the boolean operators where added in #1060 

https://github.com/gleam-lang/gleam/pull/1060/files#diff-9e23c200211e3fa575acdbf7ecade7ce9a76c2bd2b5d6e2c227961f81c4b7248R44

Issue #1040, opened 2021-04-15
JS: Compilation of prelude types `Nil` and `Bool`


lpil:
- [x] Expressions
- [x] Constants
- [x] Patterns

To do this we will need to check the type to see if the constructor belongs to the prelude.

```rust
True
False
Nil
```
```js
true
false
undefined
```

PR #1060, opened 2021-04-20
compilation of True/False/Nil

CrowdHailer:
tackles #1040

### DIFF
--- a/src/javascript/expression.rs
+++ b/src/javascript/expression.rs
@@ -1,5 +1,9 @@
 use super::*;
-use crate::{ast::*, pretty::*};
+use crate::{
+    ast::*,
+    pretty::*,
+    type_::{ValueConstructor, ValueConstructorVariant},
+};
 
 #[derive(Debug)]
 pub struct Generator {
@@ -37,7 +41,9 @@ impl Generator {
             TypedExpr::RecordAccess { .. } => unsupported("Custom Record"),
             TypedExpr::RecordUpdate { .. } => unsupported("Function"),
 
-            TypedExpr::Var { .. } => unsupported("Referencing variables"),
+            TypedExpr::Var {
+                name, constructor, ..
+            } => self.variable(name, constructor),
             TypedExpr::Seq { first, then, .. } => self.sequence(first, then),
             TypedExpr::Assignment { .. } => unsupported("Assigning variables"),
 
@@ -100,6 +106,15 @@ impl Generator {
             break_("", " "),
             "})()",
         ))
+    }
+
+    fn variable<'a>(&mut self, _name: &'a str, constructor: &'a ValueConstructor) -> Output<'a> {
+        match &constructor.variant {
+            ValueConstructorVariant::Record { name, .. } if name == "True" => Ok("true".to_doc()),
+            ValueConstructorVariant::Record { name, .. } if name == "False" => Ok("false".to_doc()),
+            ValueConstructorVariant::Record { name, .. } if name == "Nil" => Ok("null".to_doc()),
+            _ => unsupported("Referencing variables"),
+        }
     }
 
     fn sequence<'a>(&mut self, first: &'a TypedExpr, then: &'a TypedExpr) -> Output<'a> {
--- a/src/javascript/expression.rs
+++ b/src/javascript/expression.rs
@@ -121,7 +121,7 @@ impl<'module> Generator<'module> {
                 Ok("false".to_doc())
             }
             ValueConstructorVariant::Record { .. } if constructor.type_.is_nil() => {
-                Ok("null".to_doc())
+                Ok("undefined".to_doc())
             }
             _ => unsupported("Referencing variables"),
         }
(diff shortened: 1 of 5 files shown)


src/javascript/expression.rs
```diff
@@ -172,6 +187,9 @@ pub fn constant_expression<'a, T, Y>(expression: &'a Constant<T, Y>) -> Output<'
         Constant::String { value, .. } => Ok(string(&value.as_str())),
         Constant::Tuple { elements, .. } => array(elements.iter().map(|e| constant_expression(&e))),
         Constant::List { .. } => unsupported("List as constant"),
+        Constant::Record { name, .. } if name == "True" => Ok("true".to_doc()),
```
CrowdHailer, line 190: Is checking just the name enough? I was looking that the module field in this `Constant::Record` struct and it was marked as None so I couldn't see anything that would flag this as Prelude.

Is it even possible to define a Custom type with varients True/False/Nil?

lpil: It is possible to use those names with custom types, so we need to check that the type, which is stored on the record's `typ` field.

`Type` has methods like `is_int`, we can add an `is_bool` and `is_nil` to match.

src/javascript/tests/booleans.rs
```diff
@@ -0,0 +1,41 @@
...
+    False
+    Nil
+}
+"#,
+        r#""use strict";
+
+function go() {
+  true;
+  false;
+  return null;
```
lpil, line 18: I think `undefined` might be a closer fit as it's what is returned by JavaScript non-returning expressions.

CrowdHailer: I think semantically `null` is closer. undefined is meant to be more due to JS's ability to declare a variable before using it.
Personally I often handwrite JS (or typescript) with null so that I know undefined means I've fogotten an execution path.

Tbh, I suspect there will a large number of developers who could argue either way.

src/javascript/expression.rs
CrowdHailer: How come in the ast module there is a single constant module. which has typed untyped variants. https://github.com/gleam-lang/gleam/blob/main/src/ast/constant.rs#L4-L5

Whereas for expressions there are separate data structures for the typed, untyped variants?

CrowdHailer: Also it looks like the details of an `is_false` function does simply check the name and that the module defining it is none. https://github.com/gleam-lang/gleam/blob/main/src/type_.rs#L93

Still neater to put it on the Type so will do that.

src/javascript/tests/booleans.rs
lpil: I think `null` is typically used to say "this field exists but it is empty". i.e. `{ name: null }`. This is how we use `None`.

`undefined` is used when something returns nothing (`let x = non_returning_function()`) or when accessing something that does not exist `{ "ok": 1 }.missing_field`, which is how we use `Nil`.

src/javascript/expression.rs
lpil: > How come in the ast module there is a single constant module. which has typed untyped variants. https://github.com/gleam-lang/gleam/blob/main/src/ast/constant.rs#L4-L5
> Whereas for expressions there are separate data structures for the typed, untyped variants?

This is because the typed and untyped AST have a different structure, as more has been learnt about the code during type analysis. This isn't the case for constants (at least not to an extent we can't represent it with generics) so it has one parameterised AST.

src/javascript/tests/booleans.rs
CrowdHailer: > `undefined` is used when something returns nothing

Not in my experience. In fact in typescript I see a lot of deliberate use of `null` so that you use the checker to easily find bugs where you have forgotten a return or to handle the fallback case.

As mentioned not a hill I want to die on. I can disagree and commit. And if you'd rather have undefined, I'll make it so.

src/javascript/expression.rs
CrowdHailer: Thanks for the explanation. I think PR is done apart from null vs undefined.

CrowdHailer:
I've added the binary boolean operators to this PR as well which should close #1045

CrowdHailer:
@lpil bit lost on the conflict.

What was the reasoning for introducing the $divide function? Is there a way to track these changes rather than jumping them in. I believe that most of the things you've done are probably improvements but feel a bit like I'm playing catchup.

src/javascript/tests/booleans.rs
lpil: I think we can avoid the need for the runtime error handling behaviour through the type system, it'll never let you return `Nil` erroneously. Let's go with `undefined`.

CrowdHailer: Done

lpil:
> What was the reasoning for introducing the $divide function?

I think it makes the code a little clearer/normal looking and it removes

lpil (approved):
Wonderful stuff, thank you

lpil:
As for keeping up with changes, unsure. Sorry I'm used to this being mostly me. Do you have any suggestions?

CrowdHailer:
It's not a big problem, I'm just keen to get as much of the work covered in my original PR in ASAP. And assumed you wouldn't touch much until you got more time. Finally Rust type warnings did make handling a merge conflict quicker than I expected.

Tbh I would just suggest code refactors become an issue before a PR that would mean I could just not open anything more my side if I knew they were coming. That said I wouldn't want to slow you down and as mentioned they were less of an issue than I was worried they might be.

Issue #1041, opened 2021-04-15
JS: Anonymous functions


lpil:
Not including capture syntax

CrowdHailer:
Is not including capture syntax because capture syntax is just syntactic sugar and doesn't need any special printing? This was the conclusion I reached on my previous PR
https://github.com/gleam-lang/gleam/pull/1027/files#diff-05eb347e16ba4d984db0b9f737a000a1a2845d0e8131b93f0b2ba0e8c6e83478R517-R542

Note that what does need to happen is to allow use `gleam$capture` in JS rather than `gleam@capture` because `@` is not allowed in variable names.

lpil:
Yup that's right, it gets expanded before this stage

lpil:
> Note that what does need to happen is to allow use gleam$capture in JS rather than gleam@capture because @ is not allowed in variable names.

Well spotted. I'm not super happy with this magic string, perhaps there is a better option here.

CrowdHailer:
Could you have an enum with to_js and to_erl

```rust
enum HiddenVar {
  Tmp
  Capture
}

to_js -> "gleam$".append(name)
to_erl -> "gleam@.append(name)
```
That's very pseudo cody but some such.

Issue #1034, opened 2021-04-15
Basic JS code generation


lpil:
- [x] Expressions
  - [x] Ints (with bitwise operator hints to make then ints rather than floats)
  - [x] Floats
  - [x] Strings
  - [x] Function calls
- [x] Constants
  - [x] Ints
  - [x] Floats
  - [x] Strings
- [x] Module level functions

lpil:
One for @CrowdHailer I think!

CrowdHailer:
@lpil what's the reason for this separation? constants seem very simple compared too (and unrelated too) module function calls.

lpil:
@CrowdHailer Best effort guesses from me! You know more than I do at the moment so if you think there are PRs that split the work into chunks that make more sense for you please do that and we can squidge the issues to match.

CrowdHailer:
#1053 covers string expressions

lpil:
I've added expression sequences

CrowdHailer:
I think you've change the logic a bit. By wrapping everything in immediately invoking functions so that a seqence can always end with a `return expression`. I was avoiding immediatly invoking functions so that you could use return for the try syntax. Not too much of a problem I assume but will need to think a bit differently.

lpil:
`try` is expression based rather than function based, so I don't believe it's possible to use `return` with it unless you wrap it in an iife. If this code where to return at the `try` it would have the wrong result (Error(_) rather than 1).

```rust
let x = {
  try y = z
  y
}
1
```

tclain:
would be happy to help in js codegen as I write js/ts for a living.

Any pointers to help me navigate the codebase and insert this codegen relative code ?

lpil:
Hey @tclain ! It would be great to have you help out, thank you.

Would you be interested in any particular feature compilations? Maybe lists?

CrowdHailer:
@lpil I've added constants for ints + floats to #1055. As they are literals there are no `| 0` hints required.
That PR should then mean all the constants terms in this issue and expressions except for function calls are complete.

lpil:
> As they are literals there are no | 0 hints required.

Int literals are floats unless the hint is added, V8 etc won't apply the optimisation. I'll add it quickly.

> That PR should then mean all the constants terms in this issue and expressions except for function calls are complete.

🎉

CrowdHailer:
Is this one done now?

lpil:
Yes! Thank you

PR #1062, opened 2021-04-22
Calling Functions

CrowdHailer:
This finishes off calling functions in #1034  and declaring/using anonymous functions in #1041

### DIFF
--- a/src/javascript/expression.rs
+++ b/src/javascript/expression.rs
@@ -35,7 +35,7 @@ impl<'module> Generator<'module> {
 
             TypedExpr::Case { .. } => unsupported("Case"),
 
-            TypedExpr::Call { .. } => unsupported("Function"),
+            TypedExpr::Call { fun, args, .. } => self.call(fun, args),
             TypedExpr::Fn { .. } => unsupported("Function"),
 
             TypedExpr::RecordAccess { .. } => unsupported("Custom Record"),
@@ -124,6 +124,8 @@ impl<'module> Generator<'module> {
                 Ok("undefined".to_doc())
             }
             ValueConstructorVariant::LocalVariable => Ok(name.to_doc()),
+            // Whats the difference between name and modfn name
+            ValueConstructorVariant::ModuleFn {..} => Ok(name.to_doc()),
             _ => unsupported("Referencing variables"),
         }
     }
@@ -138,6 +140,14 @@ impl<'module> Generator<'module> {
         self.not_in_tail_position(|gen| {
             array(elements.iter().map(|element| gen.wrap_expression(element)))
         })
+    }
+
+    fn call<'a>(&mut self, fun: &'a TypedExpr, arguments: &'a [CallArg<TypedExpr>]) -> Output<'a> {
+        let fun = self.not_in_tail_position(|gen| gen.expression(fun))?;
+        let arguments = self.not_in_tail_position(|gen| {
+            call_arguments(arguments.iter().map(|element| gen.wrap_expression(&element.value)))
+        })?;
+        Ok(docvec![fun, arguments])
     }
 
     // TODO: handle precedence rules
@@ -225,6 +235,19 @@ fn array<'a, Elements: Iterator<Item = Output<'a>>>(elements: Elements) -> Outpu
         docvec![break_("", ""), elements].nest(INDENT),
         break_(",", ""),
         "]"
+    ]
+    .group())
+}
+
+fn call_arguments<'a, Elements: Iterator<Item = Output<'a>>>(elements: Elements) -> Output<'a> {
+    let elements = Itertools::intersperse(elements, Ok(break_(",", ", ")))
+        .collect::<Result<Vec<_>, _>>()?
+        .to_doc();
+    Ok(docvec![
+        "(",
+        docvec![break_("", ""), elements].nest(INDENT),
+        break_(",", ""),
+        ")"
     ]
     .group())
 }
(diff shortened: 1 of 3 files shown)


src/javascript/expression.rs
```diff
@@ -139,6 +141,32 @@ impl<'module> Generator<'module> {
         })
     }
 
+    fn call<'a>(&mut self, fun: &'a TypedExpr, arguments: &'a [CallArg<TypedExpr>]) -> Output<'a> {
+        let fun = self.not_in_tail_position(|gen| gen.expression(fun))?;
+        let arguments = self.not_in_tail_position(|gen| {
+            call_arguments(
```
lpil, line 147: Can use the existing wrap_args function

```suggestion
            wrap_args(
```

src/javascript/tests/functions.rs
```diff
@@ -0,0 +1,56 @@
...
+export function add(x, y) {
+  return x + y;
+}
+"#
+    );
+}
+
+#[test]
+fn calling_functions() {
+    assert_js!(
```
lpil, line 21: Can we have some tests that cover the formatting? Would like to see tests that show show the code is split over multiple lines due to

- argument names too long
- function names too long
- nested function names too long
- function body too long
- a sequence of expressions in the function body

lpil (commented):
Very nice, thanks. Left a couple comments

src/javascript/tests/functions.rs
CrowdHailer: Not sure what you mean by nested function names two long

lpil: Sorry, not very clear. Kinda like this:

```rust
pub fn main() {
  aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa(
    bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb(
      1111111111111,
      2222222222222222,
    )
  )
}

File: src/javascript/expression.rs (hunk 1 above)
Comment (CrowdHailer) [reply src/javascript/expression.rs:147]: wrap_args returns Document not Output, i.e. it doesn handle errors, that's because the arguments for a function definition are always renderable. however a function call can have any expression so rendering it has to handle the error cases where an unsupported expression is used.

We could use the error handling version for all but that difference was why I kept them separate

Comment (lpil) [reply src/javascript/expression.rs:152]: Oh yes, thank you that makes sense

**Review by lpil — APPROVED** [review]
Thank you

File: src/javascript/tests/functions.rs [hunk 3]
```diff
@@ -0,0 +1,153 @@
...
+"#
+    );
+
+    assert_js!(
+        r#"
+pub fn math(x, y) {
+x + y
+x - y
+2 * x
+}"#,
```
lpil, line 143: I think this should be an anonymous function? This PR doesn't handle top level functions
```suggestion
pub fn math(x, y) {
  fn() {
    x + y
    x - y
    2 * x
  }
}"#,
```

lpil (changes requested):
See above comment, thank you

src/javascript/tests/functions.rs
CrowdHailer: Fair point, my mistake rushing.

I've fixed it. I like the result with the anonymous closing bracket on the same line in the generated JS, but just want to confirm is that the behaviour you were hoping for.

src/javascript/expression.rs
```diff
@@ -144,6 +146,31 @@ impl<'module> Generator<'module> {
...
+    fn fun<'a>(&mut self, arguments: &'a [TypedArg], body: &'a TypedExpr) -> Output<'a> {
+        let tail = self.tail_position;
+        self.tail_position = true;
+        let result = self.expression(body);
+        self.tail_position = tail;
+        Ok(docvec!(
+            docvec!(fun_args(arguments), " => {", break_("", " "), result?)
+                .nest(INDENT)
+                .group(),
+            break_("", " "),
```
lpil, line 170: I think this break should be inside the group` so that the `}` is rendered on its own line

CrowdHailer: :thinking:

src/javascript/expression.rs
```diff
@@ -144,6 +146,36 @@ impl<'module> Generator<'module> {
...
+        self.tail_position = true;
+        let result = self.expression(body);
+        self.tail_position = tail;
+        Ok(docvec!(
+            docvec!(
+                fun_args(arguments),
+                " => {",
+                break_("", " "),
+                result?,
+                break_("", " ")
```
lpil, line 172: It needs to be inside the group but not inside the nest, otherwise the `}` has one level too much indentation

src/javascript/tests/functions.rs
```diff
@@ -0,0 +1,157 @@
...
+  }
+}"#,
+        r#""use strict";
+
+export function math(x, y) {
+  return () => {
+    x + y;
+    x - y;
+    return 2 * x;
+    };
```
lpil, line 153:
```suggestion
  };
```

src/javascript/expression.rs
CrowdHailer: Thanks for the patience, this formatting is causing a lot of head scratching.

lpil (approved):
Thank you

src/javascript/expression.rs
lpil: I totally understand, it took me forever to learn all the tricks here. It's not obvious at all

PR #1066, opened 2021-04-22
object equality functions

### LINKED ISSUE #1050: JS: Equality operators
- [ ] Equal `==`
- [ ] Not equal `!=`

Gleam uses structural equality. How should this be implemented in JS? Perhaps we inline a recursive equality function into all modules.

CrowdHailer:
Uses the same approach as the function added for integer division.

Uses deep equality function described here. https://dmitripavlutin.com/how-to-compare-objects-in-javascript/
I have modified it so it takes any value as an argument rather than expecting objects as the arguments to the function.

Should close #1050

### DIFF
--- a/src/javascript.rs
+++ b/src/javascript.rs
@@ -137,7 +137,10 @@ impl<'a> Generator<'a> {
         args: &'a [TypedArg],
         body: &'a TypedExpr,
     ) -> Output<'a> {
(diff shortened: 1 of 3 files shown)


lpil (approved):
This is great! Thanks!

I think we should start something similar to the `test/core_language` tests for JS. We could rename that one to `erlang_test` and make a `javascript_test` that is similar.

I'll make an issue for it

PR #1070, opened 2021-04-25
And Imports and calling module functions.

CrowdHailer:


src/javascript/tests.rs
```diff
@@ -228,6 +258,33 @@ function baz(x, y) {
...
+// }
+"#,
+        r#""use strict";
+
+import * as rocket_ship from "./rocket_ship.js";
+
+import * as foo from "./rocket_ship.js";
+
+import * as rocket_ship from "./rocket_ship.js";
+const {launch: boom_time, fuel} = rocket_ship;
```
CrowdHailer, line 280: decided it was nicer to use the short JS syntax if no alias was given for the name. i.e. render `fuel` rather than `fuel: fuel`. Add's a smidge of complexity to the render code that could be argued is unnecessary.

src/javascript/expression.rs
```diff
@@ -235,6 +240,20 @@ impl<'module> Generator<'module> {
             .append(" ")
             .append(right))
     }
+
+    fn module_select<'a>(
+        &mut self,
+        module_name: &'a Vec<String>,
+        label: &'a String,
+        constructor: &'a ModuleValueConstructor,
+    ) -> Output<'a> {
```
CrowdHailer, line 249: If this never fails I will change from `Output` to `Document`

src/javascript/tests.rs
```diff
@@ -3,6 +3,34 @@ mod functions;
...
+    let src = r#"
+pub const c = 299_792_458
+
+pub fn launch() {
+  Ok("launched")
+}
+pub fn fuel(amount: Int) {
+  Ok("fueled")
+}
+  "#;
```
lpil, line 16: Would it be possible to instead create a macro that allows us to compile multiple modules? I would prefer this because then all of the data for the test can be viewable in one place rather than having to to know that there is some implicitly added module somewhere else in the codebase. I think this would make it much clearer what is happening in each test, and make it easier to create the required state in other tests.

This is what we currently do for the Erlang testing, though admittedly this code is not within the erl tests module

src/javascript/tests.rs
lpil: Nice, I like it.

Could we have spaces inside the {}? Could be a good use case for an object helper function that does the formatting and intending for any time we need an object.

CrowdHailer: Is that a standard formatting. I guess there are a lot of opinions in JS but I always write my code blocks with a space and objects without. i.e.
```js
() => { return {x: 1} }
```

CrowdHailer: Where else would the helper be used. I try to not create a new abstraction till I have a couple (ideally 3 places to use it)

src/javascript/tests.rs
CrowdHailer: I was getting a bit lost in rust macro land, so I stuck with this solution as it was working, certainly not very elegant.
Do you have some pointers for how you'd like it to work. some example erl code that does similar things.

src/javascript/tests.rs
lpil: > Is that a standard formatting. 

Yes I think so. Or at least, it's what prettier, airbnb, and standard do

src/javascript/tests.rs
lpil: Maybe slap a TODO on this and I'll tackle this one after merge. No point wasting your time

CrowdHailer: I'll add the TODO, I can always look myself later if I want to go deeper on rust macros

src/javascript/tests.rs
CrowdHailer: I did my own research and got to the same conclusion.

I've got a wrap_object helper in this PR. https://github.com/gleam-lang/gleam/pull/1071/files#diff-e0fbde2af80ff8ed47119a170f2c53099fef19d489f2b8ccec4ef28105278b3eR443

If we merge that one first I will refactor this PR to use that helper.

src/javascript.rs
```diff
@@ -121,6 +126,39 @@ impl<'a> Generator<'a> {
...
+    fn import(
+        &mut self,
+        module: &'a Vec<String>,
+        as_name: &'a Option<String>,
+        unqualified: &'a Vec<UnqualifiedImport>,
+    ) -> Document<'a> {
+        let module_name = as_name
+            .as_ref()
+            .map(|n| n.as_str().to_doc())
+            .unwrap_or_else(|| Document::String(module.join("_")));
```
lpil, line 138: I think we should only use the last section, as we do in Gleam. If we use the full path it is possible to have a collision between two modules like so:

```gleam
import one/two
import one_two
```
```js
import * as one_two from "./one/two";
import * as one_two from "./one_two";
```

src/javascript.rs
```diff
@@ -121,6 +126,39 @@ impl<'a> Generator<'a> {
...
+        let path = Document::String(module.join("/")).surround("\"./", ".js\"");
```
lpil, line 139: I think this will generate invalid code for modules that are nested when imports.

i.e. If `one/two` imports `three/four` then the path will resolve to `one/two/three/four`

src/javascript.rs
```diff
@@ -121,6 +126,39 @@ impl<'a> Generator<'a> {
...
+            _ => {
+                let matches = unqualified.into_iter().map(|i| match &i.as_name {
+                    Some(aliased_name) => docvec![i.name, ": ", aliased_name,],
+                    None => i.name.to_doc(),
+                });
+                docvec![
+                    import_line,
+                    line(),
+                    "const ",
+                    concat(Itertools::intersperse(matches, break_(",", ", "))).surround("{", "}"),
```
lpil, line 153: I think this would be a good place to use an object helper to ensure that it formats correctly. At the moment this one renders like so:

```js
const {
one,
two,
three,
four,
five,
six,
} = module;
```
And one long import will cause every on in the file to be rendered on multiple lines.

lpil: Ah I see this is in the other PR, thanks

src/javascript.rs
CrowdHailer: Agreed and Fixed

src/javascript/tests.rs
CrowdHailer: Updated to use object helper, has spaces.

src/javascript.rs
CrowdHailer: Uses helper

> Tests for the formatting of long lines would be good

Haven't added tests as all the long line formatting is checking in the custom records tests.

src/javascript/tests.rs
CrowdHailer: Comment added

src/javascript.rs
CrowdHailer: Updated to now correctly generate path `../three/four`. I'm leaving all package imports to handle as a single thing relating to issue #1072 

For now I think this works for all cases of handling imports from code generated in this project.

(Note I also think this will work for gleam packages dependend on for now because they all become generated modules in the gen dir)

CrowdHailer:
@lpil ready for another review I think

lpil (approved):
Nice, thank you

Issue #1073, opened 2021-04-26
JS: implement a JavaScript error for `todo`'s in Gleam source code.


CrowdHailer:

CrowdHailer:
Close as I think this is done

PR #1071, opened 2021-04-25
Custom records

CrowdHailer:
Adding custom records tackles. #1043 #1044

PR #1074, opened 2021-04-26
render a TODO error

CrowdHailer:


src/javascript/expression.rs
```diff
@@ -235,6 +243,34 @@ impl<'module> Generator<'module> {
...
+        let line = location.start;
+
+        println!("{:?}", self);
+        docvec![
+            "throw Object.assign",
+            wrap_args(
+                vec![
+                    docvec!["new Error", wrap_args(std::iter::once(string(message)))],
+                    // TODO use the object wrapper in other PR
+                    "{}".to_doc(),
```
CrowdHailer, line 264: Use the object wrapper in PR #1071

src/javascript/expression.rs
```diff
@@ -326,6 +334,34 @@ impl<'module> Generator<'module> {
             .append(" ")
             .append(right))
     }
+
+    fn todo<'a>(&mut self, message: &'a Option<String>, location: &'a SrcSpan) -> Document<'a> {
+        self.tail_position = false;
```
CrowdHailer, line 339: I set tail position false here because we are permanently not in the tail position as from here the function will not return.
i.e. we don't want to render
```
return throw error
```
just
```
throw error
```

src/javascript/expression.rs
```diff
@@ -326,6 +334,34 @@ impl<'module> Generator<'module> {
...
+        let message = message
+            .as_ref()
+            .map(|s| s.as_str())
+            .unwrap_or_else(|| "This has not yet been implemented");
+        let module_name = Document::String(self.module_name.join("_"));
+        let line = location.start;
+
+        println!("{:?}", self);
+        docvec![
+            "throw Object.assign",
```
CrowdHailer, line 350: I use `Object.assign` directly here rather than another function call (like `$gleam_error`) to not add any more items to the stacktrace.

CrowdHailer: From this post https://xjamundx.medium.com/custom-javascript-errors-in-es6-aa891b173f87

> One of the common ways to handle this is to append errors with custom properties (which is used heavily in node.js itself.)
> This is a very reliable approach, but has two downsides:
> 
> 1. It can be a lot of lines of code (1 line per extra property)
> 2. Your error instance can’t have any custom methods

I do this here rather than create a custom error, because (1.) is not much of an issue, it's not many lines and they're generated anyway. (2.) I don't have a need for any custom methods on the error. 

To create a custom error would need to be done once per project, so that anyone catching the error could use `instanceof`. To implement that would need us to import a single gleam helpers module which we currently don't have.

So yeah this approach because the downsides aren't very big and a single custom error would be more complexity than it's worth to add.

src/javascript/tests.rs
```diff
@@ -275,3 +275,40 @@ function $isObject(object) {
...
+fn go() {
+    todo
+}
+"#,
+        r#""use strict";
+
+function go() {
+  throw Object.assign(
+    new Error("This has not yet been implemented"),
+    { gleam_error: "todo", module: "the_app", function: "go", line: 15 }
```
CrowdHailer, line 292: Not bothered with any long names in the formatting test as this uses `wrap_args` and `wrap_object` that are already well tested for splitting over multiple lines properly.

CrowdHailer:
This should close #1073

src/javascript/expression.rs
```diff
@@ -326,6 +334,43 @@ impl<'module> Generator<'module> {
...
+                vec![
+                    docvec!["new Error", wrap_args(std::iter::once(string(message)))],
+                    wrap_object(
+                        vec![
+                            ("gleam_error".to_doc(), Some(string(gleam_error))),
+                            ("module".to_doc(), Some(module_name.surround("\"", "\""))),
+                            (
+                                "function".to_doc(),
+                                Some(
+                                    Document::String(self.function_name.to_string())
```
lpil, line 361: Not `string(self.function_name)`?

src/javascript/expression.rs
```diff
@@ -326,6 +334,43 @@ impl<'module> Generator<'module> {
...
+        self.tail_position = false;
+        let gleam_error = "todo";
+        let message = message
+            .as_ref()
+            .map(|s| s.as_str())
+            .unwrap_or_else(|| "This has not yet been implemented");
+        let module_name = Document::String(self.module_name.join("_"));
+        let line = location.start;
+
+        println!("{:?}", self);
```
lpil, line 348:
```suggestion
```

src/javascript/expression.rs
```diff
@@ -326,6 +334,43 @@ impl<'module> Generator<'module> {
...
+                            ("gleam_error".to_doc(), Some(string(gleam_error))),
+                            ("module".to_doc(), Some(module_name.surround("\"", "\""))),
+                            (
+                                "function".to_doc(),
+                                Some(
+                                    Document::String(self.function_name.to_string())
+                                        .surround("\"", "\"")
+                                )
+                            ),
+                            ("line".to_doc(), Some(Document::String(format!("{}", line)))),
```
lpil, line 365:
```suggestion
                            ("line".to_doc(), Some(line.to_doc()),
```

src/javascript/expression.rs
```diff
@@ -326,6 +334,43 @@ impl<'module> Generator<'module> {
...
+
+    fn todo<'a>(&mut self, message: &'a Option<String>, location: &'a SrcSpan) -> Document<'a> {
+        self.tail_position = false;
+        let gleam_error = "todo";
+        let message = message
+            .as_ref()
+            .map(|s| s.as_str())
+            .unwrap_or_else(|| "This has not yet been implemented");
+        let module_name = Document::String(self.module_name.join("_"));
+        let line = location.start;
```
lpil, line 346: This isn't a line number, it's a byte offset. We'll need to convert it like this

https://github.com/gleam-lang/gleam/blob/5719757fd20e1324f492fcba466e5aa9e5d2f549/src/erl.rs#L1360

src/javascript/expression.rs
lpil: Makes sense, thank you. I would be good to have the extra properties get printed in the console but I can't find a way to do that in JavaScript without generating lots of extra classes.

src/javascript/tests.rs
lpil: Makes sense, ta

lpil (commented):
I like the use of Object.assign with an error. Very nice

src/javascript/expression.rs
CrowdHailer: Added TODO with comments

lpil (approved):
Wonderful, thank you

Issue #1044, opened 2021-04-15
JS: Custom types with fields


lpil:
- [x] Expressions
  - [x] local
  - [x] imported
  - [x] imported an unqualified
- [x] Patterns
  - [x] local
  - [x] imported
  - [x] imported an unqualified
- [x] Constants
  - [x] local
  - [x] imported
  - [x] imported an unqualified

```rust
pub fn main() {
  Thing(1)
  Named(1)
  Named(thingy: 2)
}

pub type Box {
  Thing(Int)
  Named(thingy: Int)
}
```
```js

export function main() {
  { type: "Thing", 0: 1 };
  { type: "Named", thingy: 1 };
  return { type: "Named", thingy: 2 };
}
```
follows on from https://github.com/gleam-lang/gleam/issues/1043

CrowdHailer:
Today I learnt this was valid Gleam.
```rust
type Ip{
    Ip(String)
}

fn go() {
    let x = Ip
    x("1.2.3.4")
}
```

An interesting case that I'm not really sure how to handle.  Because if we consider the generation of a record something we do with literals. i.e we haven't made a constructor function for Ip. Then the line `let x = ` does nothing at all.

In fact it get's even more confusing. consider:
```rust
type Point{
    Point(Int, Int)
}

fn go() {
    let x = Point(1, _)
    x(2)
}
```

Because capturing is handled by the gleam compiler before the printing process it will try to render functions where there is no constructor function.

lpil:
This is the purpose of the value constructor variant enum, it contains the information as to what kind of constructor a function is so that even when used as a variable we know how to render the appropriate code. There's some examples in the Erlang generation code today.

PR #1079, opened 2021-04-29
Constant custom types with named fields are not rendered properly.

CrowdHailer:
Did not get to adding constants with fields to the tests, this adds that test. 

This should be fixable  by using the `construct_record` helper that takes a field_map. 

https://github.com/gleam-lang/gleam/blob/main/src/javascript/expression.rs#L428
https://github.com/gleam-lang/gleam/blob/main/src/javascript/expression.rs#L472-L482

As far as I could see however the constant_expression for Records doesn't have a field map so that would need to be part of the fix. 

This needs to be fixed before #1044 is considered resolved.


lpil (approved):
I agree with your reasoning, this a good trade off here.

Thank you

PR #1080, opened 2021-05-01
Add error message clause for inexhaustive case expression

shunjilin:
- Add error message clause for inexhaustive case expression
- Update tests for error clause

I tried to use `docvec!` to structure the output just like #977 .

It appears that the generated case clauses should be separated by newlines;
I tried to insert a newline before the error clause (not reflected in this PR yet) but if I do it within `docvec!` it would insert a newline with an indent.

Should I split it up into two `docvec!` and insert a newline with `append`?

Or perhaps there is a nicer way to do the refactor with `docvec!` that I am missing, any advice would be appreciated


lpil (commented):
Wow that was quick!

I like what you've done here, though I think the code that renders the extra clause could be moved into the clauses function here.

https://github.com/gleam-lang/gleam/blob/982053c73cde1fc1eb58d278f48378cf92c6977f/src/erl.rs#L1147-L1152

The iterator returned by `map` can be extended with the new clause using `.chain(std::iter::once(the_new_clause_doc))`

RE docvec and append, they do the same thing, just different interfaces for the save document vector data structure

shunjilin:
> 
> 
> Wow that was quick!
> 
> I like what you've done here, though I think the code that renders the extra clause could be moved into the clauses function here.
> 
> https://github.com/gleam-lang/gleam/blob/982053c73cde1fc1eb58d278f48378cf92c6977f/src/erl.rs#L1147-L1152
> 
> The iterator returned by `map` can be extended with the new clause using `.chain(std::iter::once(the_new_clause_doc))`
> 
> RE docvec and append, they do the same thing, just different interfaces for the save document vector data structure

got it, thank you! yes that makes sense, I will take a closer look at the `clauses` function

src/erl.rs
```diff
@@ -1163,14 +1187,14 @@ fn case<'a>(subjects: &'a [TypedExpr], cs: &'a [TypedClause], env: &mut Env<'a>)
...
-        .append("end")
-        .group()
+    docvec![
+        "case ",
+        subjects_doc,
+        " of",
+        docvec![line(), clauses(cs, env, location)].nest(INDENT),
+        line(),
+        "end",
+    ]
```
lpil, line 1197: I think we lost a `group` here?

shunjilin: > 
> 
> I think we lost a `group` here?

Sorry about that, fixed!
Just curious, I believe the `group()` function is used to create a `Group` data type that would be used to format the indents? I wonder why it did not make a difference in the unit tests when I left it out 🤔 

Also, CI seems to require me to run cargo fmt on `src/project/tests.rs` on the entire file, even in the places where I have not made changes, hope that is ok

lpil: No worries, easy to do if the tests don't catch it. It might not be needed now I think about it, but let's keep it in case there's a hole in the tests. We can remove it later in another PR if we want.

Formatting sounds good, thanks

lpil (approved):
Wonderful!

One last thing, would you mind updating the changelog file? Thank you

shunjilin:
> 
> 
> Wonderful!
> 
> One last thing, would you mind updating the changelog file? Thank you

got it! will do.

But before that, it seems like I have run into a problem in the `test-projects` step of the CI.

It seems the erlang compiler is able to check for redundant case clauses, for example:

```erlang
-spec max(integer(), integer()) -> integer().
max(A, B) ->
    case A > B of
        true ->
            A;

        false ->
            B;

        Gleam@Case ->
            erlang:error(#{gleam_error => 'case',
                           message => <<"Case pattern match failed"/utf8>>,
                           value => Gleam@Case,
                           module => <<"gleam/int"/utf8>>,
                           function => <<"max"/utf8>>,
                           line => 139})
    end.
```
this gives the error `this clause cannot match because a previous clause always matches`

I can think of using `try/catch` instead of `case` as a solution, but am not sure of the performance implications. Any advice would be appreciated

lpil:
Oh damn, this is quite a problem.

Try catch could work, though I'm not sure how we would ensure that we catch the correct error as there may be one thrown from deeper within one of the clauses.

Did you have some particular Erlang code in mind for the try/catch approach?

shunjilin:
> 
> 
> Oh damn, this is quite a problem.
> 
> Try catch could work, though I'm not sure how we would ensure that we catch the correct error as there may be one thrown from deeper within one of the clauses.
> 
> Did you have some particular Erlang code in mind for the try/catch approach?

Ah, thats true, I did not account for the nested clauses, let me think it over and get back to you

lpil:
Thank you. I am sorry I've accidentally given you this hard problem! I didn't realise how tricky it was

shunjilin:
> 
> 
> Thank you. I am sorry I've accidentally given you this hard problem! I didn't realise how tricky it was

No problem at all! Happy to help. Will be away for two days on a trip but hope to solve this by the weekend. Please don't hesitate to take over or assign to someone else if this is blocking progress!

lpil:
Enjoy your trip! No rush

shunjilin:
I've updated to use an outer `try` / `catch` to convert `case_clause` error to the gleam error. This should work for nested gleam `case` statements, the gleam error should reflect the failure of the deepest `case`.

As for other keywords that generate `case`, such as `assert` and `try`, they need to handle the error locally and convert to the gleam specific error, otherwise it might get caught by this `try` / `catch`. At the moment I cannot find where this might be a problem, but my knowledge is limited, please advice!

Also any advice on refactoring, style etc would be appreciated.

lpil:
I'm afraid that version isn't going to work because it'll catch errors thrown by inexhaustive cases in Erlang Elixir etc as well. Here's an example:

```erlang
-module(one).
-export([one/0]).

one() ->
  case 1 of
    2 -> nil
  end.
```
```rust
pub fn main() {
  case one() {
    Nil -> Nil
  }
}

external fn one() -> Nil = "one" "one"
```

Here the error is thrown by the Erlang module, but it'll be incorrectly recorded as being from the Gleam

shunjilin:
> 
> 
> I'm afraid that version isn't going to work because it'll catch errors thrown by inexhaustive cases in Erlang Elixir etc as well. Here's an example:
> 
> ```erlang
> -module(one).
> -export([one/0]).
> 
> one() ->
>   case 1 of
>     2 -> nil
>   end.
> ```
> 
> ```rust
> pub fn main() {
>   case one() {
>     Nil -> Nil
>   }
> }
> 
> external fn one() -> Nil = "one" "one"
> ```
> 
> Here the error is thrown by the Erlang module, but it'll be incorrectly recorded as being from the Gleam

Ah I see, thank you.

What if we try pattern matching the stacktrace to make sure the first entry matches the module and function/arity the `case` statement of interest is in? Do you think that would suffice?

PR #1084, opened 2021-05-02
Init command implementation

chespinoza:
This PR implements an `init` command that takes as argument the root folder where the project files are going to be generated, it asks for the project name, which is going to be used in the gleam config files.

```
➜  hello git:(init-command) ✗ ./gleam init .
Enter a name for your project: my_project

Your Gleam project my_project has been successfully created.
The project can be compiled and tested by running these commands:


    rebar3 eunit

```

Therefore if you want to create a project called `my_super_project` in another folder, for instance in `my_projects/first_project` you just have to:

```
➜ gleam init my_projects/first_project
```

And you'll get all the required project files into  `my_projects/first_project` but you'r project internally will be called `my_super_project`

Related issue: https://github.com/gleam-lang/suggestions/issues/100

### DIFF
--- a/src/new.rs
+++ b/src/new.rs
@@ -529,7 +529,9 @@ pub fn hello_world_test() {{
 pub fn create(options: &mut NewOptions, version: &'static str, init_mode: bool) -> Result<()> {
     if init_mode {
         print!("Enter a name for your project: ");
-        std::io::stdout().flush().expect("Error while flushing stdin");
+        std::io::stdout()
+            .flush()
+            .expect("Error while flushing stdin");
         options.project_root = Some(options.name.to_string());
         options.name = String::new();
         let _ = std::io::stdin()
@@ -574,7 +576,6 @@ The project can be compiled and tested by running these commands:
     );
     Ok(())
 }
-
 
 fn write(path: PathBuf, contents: &str) -> Result<()> {
     let mut f = File::create(&path).map_err(|err| Error::FileIo {
--- a/src/new.rs
+++ b/src/new.rs
@@ -526,15 +526,38 @@ pub fn hello_world_test() {{
     }
 }
 
-pub fn create(options: NewOptions, version: &'static str) -> Result<()> {
+pub fn create(options: &mut NewOptions, version: &'static str, init_mode: bool) -> Result<()> {
+    if init_mode {
+        print!("Enter a name for your project: ");
+        std::io::stdout().flush().expect("Error while flushing stdin");
+        options.project_root = Some(options.name.to_string());
+        options.name = String::new();
+        let _ = std::io::stdin()
+            .read_line(&mut options.name)
+            .expect("Unable to read project name");
+        options.name = options.name.trim().to_string();
+    } else {
+        validate_root_folder(&options.name)?;
+    }
     validate_name(&options.name)?;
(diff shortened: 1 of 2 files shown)


src/new.rs
```diff
@@ -526,27 +526,52 @@ pub fn hello_world_test() {{
...
+        _ => format!("cd {}", creator.options.name).to_string(),
+    };
 
     println!(
         "
 Your Gleam project {} has been successfully created.
 The project can be compiled and tested by running these commands:
 
-    cd {}
+    {}
```
lpil, line 570: Could we include this following newline in the `cd_folder` so that if there's no path to `cd` into then there isn't a blank line

src/new.rs
```diff
@@ -526,27 +526,52 @@ pub fn hello_world_test() {{
...
+        },
+        version,
+    );
     creator.run()?;
 
     let test_command = match &creator.options.template {
         Template::Lib | Template::App | Template::Escript => "rebar3 eunit",
         Template::GleamLib => "gleam eunit",
     };
+    let cd_folder = match init_mode {
```
lpil, line 560: I think we would still want to include the `cd` if  in `init` mode but if the target directory isn't the current directory

lpil (commented):
Very nice!

Could we make it use the name of the directory rather than having an interactive asking of the project name? Currently we only use interactivity for requesting passwords.

lpil:
Closes https://github.com/gleam-lang/gleam/issues/1088

chespinoza:
> Very nice!
> 
> Could we make it use the name of the directory rather than having an interactive asking of the project name? Currently we only use interactivity for requesting passwords.
Hi @lpil,
It was the first idea I had, but then I remembered discord comments about people using non-standard naming for the root project folder, which probably make sense for particular cases so that's why I separated the project name from the project folder name, I could refactor and use the root folder name as the project name, the PR would be simpler but it'd force users to follow the naming standard for root folders making the tooling less flexible, do we want that?
Enabling the `init` command to initialise a folder whatever their name is, seemed more pragmatic and aligned as an answer to the necessity of having a `init` command to me.

lpil:
We can follow the same logic as `new` and permit the user to supply an alternative name if the directory name is not acceptable.

lpil:
Actually thinking about this now I'm not sure I understand why we need both `init` and `new`? I think I've got confused somewhere. 

Why do we not use `new`?

image

src/new.rs
```diff
@@ -553,22 +553,23 @@ pub fn create(options: &mut NewOptions, version: &'static str, init_mode: bool)
...
+    let cd_folder = if options.project_root.as_ref() == Some(&".".to_string()) {
+        "".to_string()
+    } else {
+        format!("\tcd {}\n", creator.options.name).to_string()
+    };
+
     let test_command = match &creator.options.template {
-        Template::Lib | Template::App | Template::Escript => "rebar3 eunit",
+        Template::Lib | Template::App | Template::Escript => "\trebar3 eunit",
         Template::GleamLib => "gleam eunit",
```
lpil, line 564: I think we'll also need indentation here

chespinoza:
This will be addressed on new PR => https://github.com/gleam-lang/gleam/pull/1101

Issue #1075, opened 2021-04-27
JS: Lists


CrowdHailer:
- [ ] patterns
  - [x] fixed length
  - [x] spread tail
- [x] expressions
  - [x] fixed length
  - [x] spread tail
  - [x] multi-line expresions as elements `[{1; 2}]`
- [x] constants
  - ~~[ ] multi-line expresions as elements `[{1; 2}]`~~

lpil:
Lists in Gleam are immutable data structure and make use of structural sharing to negate the time and memory costs of conventional immutability.

# Code generation

- Gleam -> JavaScript
- `[]` -> `[]`
- `[1]` -> `[1, []]`
- `[1, 2]` -> `[1, [2, []]]`

CrowdHailer:
Are multi line expressions even possible in constants, wouldn't that require sequences or some such?

lpil:
They're not possible currently

CrowdHailer:
Will take them off the checklist then. Also I marked expression spread as done. I think it's done, the test is here. https://github.com/gleam-lang/gleam/blob/main/src/javascript/tests/lists.rs#L12

lpil:
Ah, nice one! Thank you

PR #1095, opened 2021-05-06
Add expression and constant Lists

CrowdHailer:
Issue #1075


src/javascript/tests/lists.rs
```diff
@@ -0,0 +1,84 @@
...
+    [11111111111111111111111111111111111111111111, 11111111111111111111111111111111111111111111111111111111111111111111111111111111]
+}
+"#,
+        r#""use strict";
+
+function go() {
+  [111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111,
+    []];
+  return [11111111111111111111111111111111111111111111,
+    [11111111111111111111111111111111111111111111111111111111111111111111111111111111,
```
CrowdHailer, line 42: I'm reasonably happy with this formatting. gets a bit messy with multiline elements but I think it was always going too

src/javascript/expression.rs
```diff
@@ -464,6 +472,26 @@ fn array<'a, Elements: Iterator<Item = Output<'a>>>(elements: Elements) -> Outpu
     .group())
 }
 
+fn wrap_list<'a, E>(
+    elements: &'a [E],
+    inner: Document<'a>,
+    mut renderer: impl FnMut(&'a E) -> Output<'a>,
```
CrowdHailer, line 478: needs to implement FnMut because the closure in the expression has a reference to self.

src/javascript/tests/lists.rs
lpil: Can we go for this kind of formatting?

```js
return [
  11111111111111111111111111111111111111111111,
  [
    11111111111111111111111111111111111111111111111111111111111111111111111111111111
    []
  ],
]
```

I think that's a bit more "javascripty". We use this formatting for calls so can copy the structure of the document from there.

CrowdHailer: I can try, I did go for something like this but each new element had another indent call so my first version wasn't working.

At this point I'm not sure any approach is very javascripty, and I liked the compactness of this version but I can give a crack at it.

lpil: Actually yeah that's a good point. Let's use this, at least for now

lpil (approved):
Thank you

PR #1099, opened 2021-05-07
Add let bindings. 

CrowdHailer:
Very much draft status at the moment. Though comments on the approach welcome

Still todo:

- [x] Add integers and increment for variable names. Should I extract the variable name counter from erlang and reuse or copy paste?
- [x] Do matching for lists. Is there a smart way to could the length of a list (not array)
- [ ] ~~There is no default `bad_match` behaviour in JS, should assert and let raise the same error if the term doesn't match~~(Later PR, just throw bad match for now)
- [ ]  ~~is it possible to also add case to this PR, perhaps with guards unsupported.~~ (later PR)
- [x] Fix formatting for long if clauses


lpil:
> Add integers and increment for variable names. Should I extract the variable name counter from erlang and reuse or copy paste?

Copy paste sounds good to me

> Do matching for lists. Is there a smart way to could the length of a list (not array)

It has to be traversed as in Erlang unfortunately. The optional chaining operator may be useful here? Unsure.

```javascript
let list_without_first_three_elements = list?.[1]?.[1]?.[1]
```

CrowdHailer:
I didn't know that `?` operator was native in JS. interesting I can't use that you check the length because the list might contain nil's which we've modelled as undefined.

src/javascript/expression.rs
```diff
@@ -418,6 +448,139 @@ impl<'module> Generator<'module> {
...
+            constructor: PatternConstructor::Record { name },
+            ..
+        } if name == "True" => {
+            checks.push(equality(&path, "true".to_doc()));
+            Ok(())
+        }
+        Pattern::Constructor {
+            constructor: PatternConstructor::Record { name },
+            ..
+        } if name == "False" => {
```
CrowdHailer, line 557: @lpil as far as I can see the PatternConstructor::Record is untyped the only varient is a record and the only field is name, therefore as the AST is now there is no way to check that the type's module is nil and therefore this could be a user type.

What I want to do is to implement `is_bool` `is_nil` somehow here, same as we use for expressions. Not sure which bit should be changed to add this.

src/type_/pattern.rs
```diff
@@ -426,7 +426,10 @@ impl<'a, 'b, 'c> PatternTyper<'a, 'b, 'c> {
                 let constructor_typ = cons.type_.clone();
                 let constructor = match cons.variant {
                     ValueConstructorVariant::Record { ref name, .. } => {
-                        PatternConstructor::Record { name: name.clone() }
+                        PatternConstructor::Record {
+                            name: name.clone(),
+                            field_map: cons.field_map().map(|f| f.clone()),
```
CrowdHailer, line 431: Uses clone because name already does.

src/type_.rs
```diff
@@ -277,7 +277,10 @@ pub struct Module {
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum PatternConstructor {
-    Record { name: String },
+    Record {
+        name: String,
+        field_map: Option<FieldMap>,
```
CrowdHailer, line 282: Uses FieldMap, not &FieldMap to match what is already done in value constructor.

CrowdHailer:
Marked as ready for review, I expect to make quite a few changes but I think it's going to at least stay with this approach.

src/javascript/expression.rs
```diff
@@ -201,7 +230,11 @@ impl<'module> Generator<'module> {
         for (i, expression) in expressions.iter().enumerate() {
             if i + 1 < count {
                 documents.push(self.not_in_tail_position(|gen| gen.expression(expression))?);
-                documents.push(";".to_doc());
+                match expression {
+                    // Assignment generates multiple lines of JS and so handles it's own ;
```
lpil, line 234: I think the assignment shouldn't generate a trailing semicolon, as with other expressions

src/javascript/expression.rs
```diff
@@ -210,6 +243,215 @@ impl<'module> Generator<'module> {
...
+                let _ = elems
+                    .into_iter()
+                    .enumerate()
+                    .map(|x| {
+                        let (index, pattern) = x;
+                        let mut path = path.clone();
+                        path.push(Document::String(format!("{}", index)));
+                        self.traverse_pattern(pattern, &mut path, checks, assignments)
+                    })
+                    .collect::<Result<Vec<()>, Error>>()?;
```
lpil, line 390: Are we building vectors and discarding them? We could use a loop and avoid the work of building the vectors in this function

src/javascript/tests/booleans.rs
```diff
@@ -59,3 +59,30 @@ function go() {
...
+    let Nil = y
+}
+"#,
+        r#""use strict";
+
+function go(x, y) {
+  let gleam$tmp = x;
+  if (!(gleam$tmp === true)) throw new Error("Bad match");
+  
+  let gleam$tmp = x;
```
lpil, line 79: This is invalid JS, we need to use a different name per assignment otherwise it will crash on attempting to redefine it.

src/javascript/tests/lists.rs
```diff
@@ -82,3 +82,49 @@ const b = [1, [2, [3, []]]];
...
+    let [1, 2] = x
+    let [_, #(3, b)] = y
+    let [head, ..tail] = y
+}
+"#,
+        r#""use strict";
+
+function go(x, y) {
+  let gleam$tmp = x;
+  if (!(gleam$tmp?.length === 0)) throw new Error("Bad match");
```
lpil, line 102: Can shorten this using arrays truthy falseyness

```suggestion
  if (!(!gleam$tmp)) throw new Error("Bad match");
```

src/javascript/tests/lists.rs
```diff
@@ -82,3 +82,49 @@ const b = [1, [2, [3, []]]];
...
+  
+  let gleam$tmp = y;
+  if (!(
+    gleam$tmp?.[1]?.[1]?.length === 0 &&
+    gleam$tmp[1][0][0] === 3
+  )) throw new Error("Bad match");
+  let b = gleam$tmp[1][0][1];
+  
+  let gleam$tmp = y;
+  if (!(gleam$tmp?.[1]?.length !== undefined)) throw new Error("Bad match");
```
lpil, line 123: Can do the same here too.

```suggestion
  if (!(gleam$tmp)) throw new Error("Bad match");
```

src/javascript/expression.rs
```diff
@@ -210,6 +243,215 @@ impl<'module> Generator<'module> {
...
+                            }
+                        }
+                        self.traverse_pattern(&arg.value, &mut path, checks, assignments)
+                    })
+                    .collect::<Result<Vec<()>, Error>>()?;
+                Ok(())
+            }
+
+            Pattern::VarUsage { .. } | Pattern::BitString { .. } => {
+                unimplemented!("Custom type matching not supported in JS backend")
```
lpil, line 449:
```suggestion
                unimplemented!("BitString matching not supported in JS backend")
```

src/javascript/expression.rs
```diff
@@ -210,6 +243,215 @@ impl<'module> Generator<'module> {
...
+                Ok(())
+            }
+            Pattern::Tuple { elems, .. } => {
+                // We don't check the length, because type system means it's a tuple
+                let _ = elems
+                    .into_iter()
+                    .enumerate()
+                    .map(|x| {
+                        let (index, pattern) = x;
+                        let mut path = path.clone();
```
lpil, line 386: This gets cloned a lot, might be a good place to use a mutable data structure? Could be one for a later PR

src/javascript/tests/custom_types.rs
```diff
@@ -142,6 +162,26 @@ function update(cat) {
...
+  let x$1 = gleam$tmp["name"];
+  
+  let gleam$tmp = cat;
+  if (!(
+    gleam$tmp["type"] === "Cat" &&
+    gleam$tmp["cuteness"] === 4
+  )) throw new Error("Bad match");
+  let x$2 = gleam$tmp["name"];
+  
+  return x$2;
```
lpil, line 183: Can we have this in another text rather than making the examples bigger? I find it makes the test failure output a lot easier to work with if you have small tests that check one thing rather than large ones that check multiple things

src/javascript/tests/numbers.rs
```diff
@@ -148,3 +148,22 @@ function $divide(a, b) {
...
+fn go(x) {
+  let 4 = x
+}
+"#,
+        r#""use strict";
+
+function go(x) {
+  let gleam$tmp = x;
+  if (!(gleam$tmp === 4)) throw new Error("Bad match");
+
```
lpil, line 165:
```suggestion
```

src/javascript/tests/lists.rs
CrowdHailer: No unfortunately not, the possible return values we have are `0`, `2` and `undefined`. 0 is also falsey but not a case we want to match on. similar to above where we want to match on 0 but not undefined.

src/javascript/expression.rs
lpil: Actually I'm not sure that would be possible

src/javascript/tests/lists.rs
CrowdHailer: It can trip into being an undefined that's passed down through the `?` if the list is shorter than needed for the match.

if you list is `[]` then `list?.[1]` is undefined and `undefined?.length` is again undefined.

src/javascript/tests/booleans.rs
CrowdHailer: Fixed this by introducing a `tmp_var` that is incremented using the next_local_variable_function.
This passes a `Document<'a>` around that get's cloned quite a lot. I think this is correct because it potentially is rendered into the output document multiple times.

src/javascript/tests/custom_types.rs
CrowdHailer: Fair enough, I was reusing the definition of the Cat type but your right it had got a bit cumbersome

src/javascript/tests/numbers.rs
lpil: I think that sounds good but we shouldn't have an empty line at the end of blocks

CrowdHailer: Interesting, I'm not sure how to fix that because the expression code just adds a line to all the lines it generates. is there a way to strip out a line after it's in a document block?

src/javascript/tests/lists.rs
lpil: Ah yes! That makes perfect sense. Thank you.

CrowdHailer:
Think ready to go now we add empty lines only in not tail position. I have a case PR nearly ready off this one too, :-)

src/javascript/expression.rs
```diff
@@ -210,6 +239,233 @@ impl<'module> Generator<'module> {
...
+                        }
+                        self.traverse_pattern(
+                            &arg.value,
+                            tmp_var.clone(),
+                            &mut path,
+                            checks,
+                            assignments,
+                        )
+                    })
+                    .collect::<Result<Vec<()>, Error>>()?;
```
lpil, line 458: Can we remove the vector building here? It is discarded immediately. A for loop will work as alternative.

CrowdHailer: Done

lpil (approved):
Thank you!

CrowdHailer:
Woop :heart: :purple_heart:

PR #1101, opened 2021-05-08
New command project naming behaviour

chespinoza:
This PR changes the `new` command behaviour to require the `project_root` argument and makes `name` a long optional argument, by doing so it enables to use folder names which doesn't follow the guidelines for a project name.

```bash
USAGE:
    gleam new <project-root> --description <description> --name <name> --template <template>
```

Therefore:

- If no `--name` parameter is used, the command will use the project root folder name as the project name
- if the project root is the current folder (.), the project name will be derived from the current folder name

Examples:

The project is created into a folder called `new`, no `--name` is specified therefore the project name is inferred from the current folder name: `new` 

```bash
./gleam new .

Your Gleam project new has been successfully created.
The project can be compiled and tested by running these commands:

    cd .
    rebar3 eunit
```

The project_root specifies a folder with a non-standard name, but a project name is supplied: `my_super_project`

```bash
./gleam new myfolder/my-project --name my_super_project

Your Gleam project my_super_project has been successfully created.
The project can be compiled and tested by running these commands:

    cd myfolder/my-project
    rebar3 eunit
```

The most basic example, just providing a simple `project_root` argument, therefore a folder called `my_project` will be created and the project name will be the same.  

```bash
./gleam new my_project

Your Gleam project my_project has been successfully created.
The project can be compiled and tested by running these commands:

    cd my_project
    rebar3 eunit
```

There is no project name specified, therefore the name will be inferred from the folder name: `my_ultra_project` 

```bash
./gleam new myfolder/my_ultra_project

Your Gleam project my_ultra_project has been successfully created.
The project can be compiled and tested by running these commands:

    cd myfolder/my_ultra_project
    rebar3 eunit
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -176,7 +176,7 @@ pub struct NewOptions {
     #[structopt(help = "location of the project root")]
     pub project_root: String,
 
-    #[structopt(help = "name of the project", default_value = "")]
+    #[structopt(long, help = "name of the project", default_value = "")]
     pub name: String,
 
     #[structopt(
--- a/src/main.rs
+++ b/src/main.rs
@@ -173,7 +173,10 @@ enum Command {
 #[derive(StructOpt, Debug)]
 #[structopt(flatten)]
 pub struct NewOptions {
-    #[structopt(help = "name of the project")]
+    #[structopt(help = "location of the project root")]
+    pub project_root: String,
+
+    #[structopt(help = "name of the project", default_value = "")]
     pub name: String,
 
     #[structopt(
@@ -182,9 +185,6 @@ pub struct NewOptions {
         default_value = "A Gleam project"
     )]
     pub description: String,
-
-    #[structopt(help = "location of the project root")]
-    pub project_root: Option<String>,
 
     #[structopt(
         long = "template",
--- a/src/main.rs
+++ b/src/main.rs
@@ -170,14 +170,14 @@ enum Command {
     CompilePackage(CompilePackage),
 }
 
-#[derive(StructOpt, Debug)]
+#[derive(StructOpt, Debug, Clone)]
 #[structopt(flatten)]
 pub struct NewOptions {
     #[structopt(help = "location of the project root")]
     pub project_root: String,
 
-    #[structopt(long, help = "name of the project", default_value = "")]
-    pub name: String,
+    #[structopt(long, help = "name of the project")]
+    pub name: Option<String>,
 
     #[structopt(
         long = "description",
--- a/src/new.rs
+++ b/src/new.rs
@@ -547,7 +547,7 @@ pub fn create(options: NewOptions, version: &'static str) -> Result<()> {
     let cd_folder = if options.project_root == *"." {
         "".to_string()
     } else {
-        format!("\tcd {}\n", creator.options.name)
+        format!("\tcd {}\n", creator.options.project_root)
     };
 
     let test_command = match &creator.options.template {
--- a/src/new.rs
+++ b/src/new.rs
@@ -3,6 +3,7 @@ use crate::{
     NewOptions, Result,
 };
 use serde::{Deserialize, Serialize};
+use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::{Path, PathBuf};
@@ -35,10 +36,7 @@ pub struct Creator {
 
 impl Creator {
     fn new(options: NewOptions, gleam_version: &'static str) -> Self {
-        let root = match &options.project_root {
-            Some(root) => PathBuf::from(root),
-            None => PathBuf::from(&options.name),
-        };
+        let root = PathBuf::from(&options.project_root);
         let src = root.join("src");
         let test = root.join("test");
         let github = root.join(".github");
@@ -527,9 +525,23 @@ pub fn hello_world_test() {{
 }
 
 pub fn create(options: NewOptions, version: &'static str) -> Result<()> {
-    validate_name(&options.name)?;
-    validate_root_folder(&options.name)?;
-    let creator = Creator::new(options, version);
+    let name: String = if options.name.is_empty() {
+        get_foldername(&options.project_root).gleam_expect("unable to get folder name")
+    } else {
+        options.name
+    };
+    validate_name(&name)?;
+    validate_root_folder(&name)?;
+    let creator = Creator::new(
+        NewOptions {
+            name,
+            description: options.description.clone(),
+            project_root: options.project_root.clone(),
+            template: options.template,
+        },
+        version,
+    );
+
     creator.run()?;
 
     let test_command = match &creator.options.template {
--- a/src/new.rs
+++ b/src/new.rs
@@ -72,7 +72,7 @@ impl Creator {
             }
 
             Template::App => {
-                crate::fs::mkdir(&self.src.join(&self.options.name))?;
+                crate::fs::mkdir(&self.src.join(self.options.name.clone().unwrap()));
                 self.gitignore()?;
                 self.github_ci()?;
                 self.app_readme()?;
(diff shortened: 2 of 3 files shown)


src/main.rs
```diff
@@ -174,7 +174,10 @@ enum Command {
 #[derive(StructOpt, Debug)]
 #[structopt(flatten)]
 pub struct NewOptions {
-    #[structopt(help = "name of the project")]
+    #[structopt(help = "location of the project root")]
+    pub project_root: String,
+
+    #[structopt(long, help = "name of the project", default_value = "")]
```
lpil, line 180: Let's use an optional type to represent the missing value rather than an empty string.
```suggestion
    #[structopt(long, help = "name of the project")]
    pub name: Option<String>
```

src/new.rs
```diff
@@ -527,11 +525,31 @@ pub fn hello_world_test() {{
 }
 
 pub fn create(options: NewOptions, version: &'static str) -> Result<()> {
-    validate_name(&options.name)?;
-    validate_root_folder(&options.name)?;
-    let creator = Creator::new(options, version);
+    let name: String = if options.name.is_empty() {
+        get_foldername(&options.project_root).gleam_expect("unable to get folder name")
```
lpil, line 529: `gleam_expect` is to be used in situtations when something has got horribly an unexpectedly wrong, it indicates a major compiler bug.

Here I think we can expect issues to arise when determining the project name, so instead lets return an error.

src/new.rs
```diff
@@ -527,11 +525,31 @@ pub fn hello_world_test() {{
...
+    validate_root_folder(&name)?;
+    let creator = Creator::new(
+        NewOptions {
+            name,
+            description: options.description.clone(),
+            project_root: options.project_root.clone(),
+            template: options.template,
+        },
+        version,
+    );
```
lpil, line 543: I think we could mutate the options to avoid creating a new one.

```suggestion
    options.name = name;
    let creator = Creator::new(options, version);
```

We will need to mark the options argument as `mut`.

src/new.rs
```diff
@@ -527,11 +525,31 @@ pub fn hello_world_test() {{
...
+
     creator.run()?;
 
+    let cd_folder = if options.project_root == *"." {
+        "".to_string()
+    } else {
+        format!("\tcd {}\n", creator.options.name)
```
lpil, line 550:
```suggestion
        format!("\tcd {}\n", creator.options.project_root)
```

lpil (commented):
This is looking very nice. I especially like the detailed write up in the description and the new command line API. Thank you!

I've left some notes inline.

src/new.rs
```diff
@@ -613,3 +630,21 @@ fn validate_name(name: &str) -> Result<(), Error> {
...
+                .and_then(|x| x.to_str())
+                .map(ToString::to_string)?,
+        ),
+        _ => Some(
+            Path::new(path)
+                .file_name()
+                .and_then(|x| x.to_str())
+                .map(ToString::to_string)?,
+        ),
+    }
```
tranzystorek-io, line 649: No need to use the `?` operator and then wrap result into `Some`, `map` already returns an Option

src/new.rs
```diff
@@ -520,18 +528,28 @@ pub fn hello_world_test() {{
...
+    if options.name.is_none() {
+        options.name = get_foldername(&options.project_root);
+    };
+    validate_name(&options.name.clone().unwrap())?;
+    validate_root_folder(&options.name.clone().unwrap())?;
+    let creator = Creator::new(options.clone(), version);
+
     creator.run()?;
 
+    let cd_folder = if options.project_root == *"." {
```
tranzystorek-io, line 547: The `*` is unnecessary

src/new.rs
```diff
@@ -74,7 +72,7 @@ impl Creator {
             }
 
             Template::App => {
-                crate::fs::mkdir(&self.src.join(&self.options.name))?;
+                crate::fs::mkdir(&self.src.join(self.options.name.clone().unwrap()));
```
tranzystorek-io, line 75: In all these you can avoid cloning by doing `self.options.name.as_ref().unwrap()` instead

lpil: And let's set a default in case the name was not specified. Calling unwrap will crash the program if there is no name. We can use the name of the target directory as a default and return and error if the directory name is not appropriate

src/new.rs
```diff
@@ -520,18 +528,28 @@ pub fn hello_world_test() {{
...
     }
 }
 
-pub fn create(options: NewOptions, version: &'static str) -> Result<()> {
-    validate_name(&options.name)?;
-    validate_root_folder(&options.name)?;
-    let creator = Creator::new(options, version);
+pub fn create(mut options: NewOptions, version: &'static str) -> Result<()> {
+    if options.name.is_none() {
+        options.name = get_foldername(&options.project_root);
```
lpil, line 539: Because we are storing name to be used as an optional it needs to be cloned and unwrapped at each time it is used. I don't mind the cleaning so much because this code only runs once, but I wish to avoid the use of unwrap because it is unsafe and could potentially crash. While we know that the name does exist at this point there is no way of verifying that using the type system when using unwrap￼.

Rather than determining the name to use and then storing it as an option let's store it just as a string. This way we know that it always exists when we need it in the other functions.￼￼

src/new.rs
```diff
@@ -613,3 +630,21 @@ fn validate_name(name: &str) -> Result<(), Error> {
         Ok(())
     }
 }
+
+fn get_foldername(path: &str) -> Option<String> {
```
lpil, line 634:
```suggestion
fn get_foldername(path: &str) -> Result<String, Error> {
```

I think we want this function to return a result￼￼￼. This way if the function is able to determine what the name should be then we have just a string which we can safely use. If it is unable to determine what the name is that we should use it returns in error and the error can be presented to the user to say they have supplied incorrect details￼.

src/new.rs
chespinoza: > Rather than determining the name to use and then storing it as an option let's store it just as a string. This way we know that it always exists when we need it in the other functions.￼￼
Hi @lpil,
I'm not sure what you mean by that, the point of this PR is making the name of the project optional so it could be inferred from the filesystem if needed, also decoupling the name of the project from the base folder making possible to have folder names like `my-project`, the easy way I found was using a default value ("") for `name` but you didn't agree with that, would you mind to explain yourself a bit more about what you are suggesting now, please? :)

src/new.rs
chespinoza: got something like this working, do you think it'd be ok?

```rust
fn get_foldername(path: &str) -> Result<String, Error> {
    match path {
        "." => env::current_dir()
            .gleam_expect("invalid folder")
            .file_name()
            .and_then(|x| x.to_str())
            .map(ToString::to_string)
            .ok_or(Error::UnableToFindProjectRoot{path: path.to_string()}),
        _ => Path::new(path)
            .file_name()
            .and_then(|x| x.to_str())
            .map(ToString::to_string)
            .ok_or(Error::UnableToFindProjectRoot{path: path.to_string()}),
    }
}
```

src/new.rs
chespinoza: there is a comment from @lpil about avoid using `unwrap` I'll come back to this later, thanks for your suggestion.

src/new.rs
lpil: Yes, that's right. We want the name to be optional, falling back to the name of the project root directory if the name is not given.

Here's some pseudocode

```rust
let name = if let Some(name) = config.name {
  name.to_string()
} else {
  name_from
};
validate_name(&name)?;
let creator = Creator::new(options.clone(), name, version);
creator.run()?;
```

It is the same approach as using an empty string but it is clearer because the optional type indicates that this value may not be present, rather than having one string value be handled differently.

lpil (approved):
Perfect, thank you

lpil:
It looks like the building is failing due to a linter warning. Would you mind taking a look at that? Thank you

lpil (approved):
Wonderful work! Thank you!

Sorry about the CI failure, I broke the main branch.

PR #1102, opened 2021-05-11
Add compiler error for duplicate module constants

### LINKED ISSUE #1100: Dissallow redeclarations of duplicate constants
I think the following code is meaningless.
```rust
pub const x = 1
pub const x = 2
```

Even if the constant was not public is there any reason to be able to declare the same constant twice?

shunjilin:
close #1100 

A naive solution to preventing duplicate module constants, using existing solutions for duplicate functions and types as a reference.

As mentioned in #1100, I was wondering if it would be better to consolidate the logic with the existing functions as they seem to be pretty similar?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Gleam can now run eunit without an external build tool.
 - Gleam can now run an Erlang shell without an external build tool.
 - Projects without rebar3 can be generated using the `gleam-lib` template.
+- A compile time error is now raised when multiple module level constants with the same name are defined.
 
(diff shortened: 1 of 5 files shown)


lpil (approved):
Perfect, thank you! I'm happy with the duplication, I don't think it causes a problem.

Would you mind updating the CHANGELOG? Thank you

shunjilin:
> 
> 
> Perfect, thank you! I'm happy with the duplication, I don't think it causes a problem.
> 
> Would you mind updating the CHANGELOG? Thank you

Thank you, done!

lpil:
Thank you ✨

PR #1104, opened 2021-05-12
use an always valid capture variable

CrowdHailer:
I started working on a solution to have a generic Gleam variable to pass around the ast.
something like
```rust
pub struct GleamVariable<'a>(&'a str);

pub const CAPTURE_VARIABLE: GleamVariable<'static> = GleamVariable("capture_variable");
```

This was actually introduced quite a lot of pain. i.e. https://github.com/gleam-lang/gleam/blob/main/src/parse.rs#L2736

Either the name of an arg needs to have a type that is an enum of String & GleamVariable. Doable but hits a lot of places.
Or, and I think even worse, the parse function needs to know if it's parsing for JS or erlang and generates a string variable name that is appropriate for each env.

In the end I realised that this simple fix would do everything we need, I can't imagine there are many programs with variables `gleam_something` although I will accept it's a possibility. and in the future maybe you would forbid variable names to start `gleam_`

### DIFF
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -10,7 +10,7 @@ pub use self::constant::{Constant, TypedConstant, UntypedConstant};
 use crate::type_::{self, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor};
 use std::sync::Arc;
 
-pub const CAPTURE_VARIABLE: &str = "gleam_capture_variable";
+pub const CAPTURE_VARIABLE: &str = "_gleam_capture";
 
 pub trait HasLocation {
     fn location(&self) -> SrcSpan;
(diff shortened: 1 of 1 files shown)


lpil:
The way that the function caputre syntax works at the moment is such a hack. I want to introduce an AST node for function captures at some point, but that's a bunch of work.

Is the problem here that the magic string used is valid Erlang but not valid JS? If so let's just switch to a variable name that can be used in both but cannot be written in Gleam, such as `_capture`.

CrowdHailer:
Is the problem here that the magic string used is valid Erlang but not valid JS?
That is exactly the problem. I chose `gleam_capture_variable` to replace `gleam@capture_variable` 

`@` is invalid in JS var names

lpil:
`gleam_capture_variable` is a valid Gleam variable name though, what do you think of `_capture`?

lpil (approved):
Thanks

PR #1107, opened 2021-05-14
fix bug where erlang keyword used in type constructor is not quoted

### LINKED ISSUE #1106: Syntax error when naming a constructor using an erlang keyword
Hi. I had this

```
type Stage {
  Start
  InProgress(Int, String)
  End(result: Float)
}
````

When trying to compile the code I get
```
===> Compiling gleam_codec
===> Compiling gen/test/codec_test.erl failed
gen/test/codec_test.erl:48: syntax error before: 'end'
```

Generated erlang:

```
    C = codec:map1(
        <<"End"/utf8>>,
        fun(A) -> {end, A} end,
        codec:field(<<"result"/utf8>>, codec:float())
    ),
```

If I rename `End` to `EEnd` the error goes away.

Thanks

shunjilin:
close #1106 

Added a test for the case provided by the issue, but let me know if a simpler test is sufficient


lpil (approved):
Great work, thank you very much!

Issue #1098, opened 2021-05-07
`Unexpected label` when argument is not used.


sporto:
Consider a function like

```
fn greet(name name: String, title title: String) {
   name
}

greet(name: "Sam", title: "Mr")
```

Title is not used in the greet function, so the compiler suggest to use `_title`.

```
fn greet(name name: String, title _title: String) {
   name
}
```

However after doing that the compiler fails with:

```
           title: "Mr",
    │             ^^^^^^^^^ Unexpected label

You have already supplied all the labelled arguments that this constructor accepts.
```

I think that the caller should not concern itself with the fact that `title` is not used.

lpil:
Ah yes, this does seem like a bug. Thanks for the report.

shunjilin:
I believe this may be closed by #1111  (sorry, forgot to link the PR)

lpil:
Thanks again!

PR #1111, opened 2021-05-17
Fix unexpected label when argument not used

shunjilin:
#1098

The issue for #1098 seems to be due to `field_map` not registering labelled function arguments with discarded argument.

This is a simple fix which includes discarded labelled arguments into `field_map`, so that it does not get classified as an `unknown_label` in `FieldMap::reorder`.

I am not sure if this is the right approach, and where would be an appropriate place to put a test for this, any advice would be appreciated


lpil:
Ah very nice find, thank you. Would you be able to add a test replicating the problem? I wouldn't want to cause a regression by mistake

shunjilin:
> 
> 
> Ah very nice find, thank you. Would you be able to add a test replicating the problem? I wouldn't want to cause a regression by mistake

Added a test, but not sure if it is in the appropriate location.
Let me know if it needs to be somewhere else!

lpil (approved):
Thank you

PR #1114, opened 2021-05-17
Escape Erlang keywords in Gleam type names

### LINKED ISSUE #1108: Syntax error when naming a type an erlang keyword
Ran into a different error while checking #1106

Naming a type constructor a reserved keyword leads to invalid erlang code generated:

e.g. gleam code:
```
pub type End {
  Test(String)
}
```

generates:
```erl
-module(module_name).
-compile(no_auto_import).

-export_type([end/0]).

-type end() :: {test, binary()}.
```

error `syntax error before: 'end'` on lines 4,6

~Not sure if there is a straightforward solution for this as functions cannot be quoted unlike atoms~

Edit: my mistake, it seems like the type declaration can be quoted, see #1114

matheus-consoli:
Hi!

In this PR, we refactor the `atom` function, separating the creation of a `Document::String` from the actual escape logic in a new function.
Using this new function, we can escape all reserved keywords when creating a type.
I also added to the reserved list some words based on the erlang [documentation](http://erlang.org/documentation/doc-5.8/doc/reference_manual/introduction.html) and created a test case using all of them.

This should fix  #1108 ;)

### DIFF
--- a/src/erl.rs
+++ b/src/erl.rs
@@ -486,6 +486,10 @@ fn fun_spec<'a>(
 }
 
 fn atom<'a>(value: String) -> Document<'a> {
+    Document::String(escape_atom(value))
+}
+
+fn escape_atom(value: String) -> String {
     use regex::Regex;
     lazy_static! {
         static ref RE: Regex = Regex::new(r"^[a-z][a-z0-9_@]*$").gleam_expect("atom RE regex");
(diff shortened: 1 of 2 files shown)


lpil (approved):
Very clean, thank you. I love the thorough test there.

lpil:
OK CI seems to be broken, a GitHub outage. I'll merge this. Thank you

PR #1115, opened 2021-05-18
Do not use recursion in JS $equal

### LINKED ISSUE #1113: JS: Alter `$equal` function to use constant stack space
Currently it recurses, growing the stack with each nested record. Update it to push values to check into an array and pop them in a while loop until there is no more.

sabiwara:
Hello!
This is my first time contribution, I tried to give this issue a shot.
Some note: in order to avoid expensive operations (push/pop/loop on keys) for objects that are `===` (same pointer if object), they never make it to the array.

Closes #1113

### DIFF
--- a/src/javascript.rs
+++ b/src/javascript.rs
@@ -11,27 +11,36 @@ const INDENT: isize = 2;
 const DEEP_EQUAL: &str = "
 
 function $equal(x, y) {
-  if ($isObject(x) && $isObject(y)) {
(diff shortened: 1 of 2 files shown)


lpil (commented):
Wonderful, thank you! This is perfect

lpil (approved):

lpil:
Something is wrong with CI. I'm sure this is working, I'll merge it. Thanks!

PR #1120, opened 2021-05-23
Differentiate prelude constructor and custom type constructors

### LINKED ISSUE #1112: JS: Bool and Nil should only be special cased if we know that they are the prelude ones
Searach for the URL of this issue to find the problem in the codebase.

We will need to add `Type` to the `Pattern::Constructor` variant and check it when rendering patterns in JS.

shunjilin:
closes #1112 

Add `type_` field to `Pattern` constructor to allow js to differentiate between prelude constructor and custom type constructor

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1040,6 +1040,7 @@ where
             name,
             with_spread,
             constructor: (),
+            type_: (),
         })
     }
 
--- a/src/ast.rs
+++ b/src/ast.rs
@@ -649,8 +649,7 @@ pub enum Pattern<Constructor, Type> {
         module: Option<String>,
         constructor: Constructor,
         with_spread: bool,
-        // TODO: https://github.com/gleam-lang/gleam/issues/1112
-        // type_: Type,
+        type_: Type,
     },
 
     Tuple {
--- a/src/javascript/pattern.rs
+++ b/src/javascript/pattern.rs
@@ -1,5 +1,5 @@
 use super::*;
-use crate::type_::{FieldMap, PatternConstructor};
+use crate::type_::{FieldMap, PatternConstructor, Type};
 
 static ASSIGNMENT_VAR: &str = "$";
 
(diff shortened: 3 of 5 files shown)


src/parse.rs
```diff
@@ -1040,6 +1040,7 @@ where
             name,
             with_spread,
             constructor: (),
+            type_: (),
```
shunjilin, line 1043: Looking at the signature I assumed this is untyped, not sure if putting `()` here is correct

lpil (commented):
Looks great!

As a bit of a nit-pick can we keep difference match clauses for true, false, and nil that isn't with the other record clause? I find it clearer if it's split out.

We could use something like this:

```rust
            Pattern::Constructor {
                constructor: PatternConstructor::Record { type_, .. },
                ..
            } if type_.is_nil() => {
                self.booly_check(false);
                Ok(())
            }
```

shunjilin:
> 
> 
> Looks great!
> 
> As a bit of a nit-pick can we keep difference match clauses for true, false, and nil that isn't with the other record clause? I find it clearer if it's split out.
> 
> We could use something like this:
> 
> ```rust
>             Pattern::Constructor {
>                 constructor: PatternConstructor::Record { type_, .. },
>                 ..
>             } if type_.is_nil() => {
>                 self.booly_check(false);
>                 Ok(())
>             }
> ```

Thank you! Fixed, and looks cleaner indeed

> 
> 
> Looks great!
> 
> As a bit of a nit-pick can we keep difference match clauses for true, false, and nil that isn't with the other record clause? I find it clearer if it's split out.
> 
> We could use something like this:
> 
> ```rust
>             Pattern::Constructor {
>                 constructor: PatternConstructor::Record { type_, .. },
>                 ..
>             } if type_.is_nil() => {
>                 self.booly_check(false);
>                 Ok(())
>             }
> ```

Thank you! Fixed, and looks cleaner indeed

lpil (approved):
Wonderful work! Thank you! ✨

lpil:
CI is broken because of a mistake I made, can ignore that

Issue #1103, opened 2021-05-12
Typed records in Erlang header files


varnerac:
Gleam currently generates `.hrl` files with record definitions. It'd be nice if we added type information to those generated record definitions.

For example

```erl
-record(person, {name :: binary()}).
```

vs.

```erl
-record(person, {name}).
```

lpil:
A great idea! Let's do this

jcc333:
I would love to take a crack at this; am I correct in thinking it'd mostly live over in `records`/`record_definition` within src/erl.rs?

lpil:
That would be great, thank you @jcc333 . Yes I think that's right

PR #1121, opened 2021-05-23
Draft for 'Typed records in erlang header files'

jcc333:
Allegedly closes [#1103](https://github.com/gleam-lang/gleam/issues/1103).

A couple of notes:

1. Running on fedora in `test/language`, `make` yields the following eventual error for this branch as well as for `main`. Reading between the lines in some other issues, it seems like `case` may just not be there in the JS target yet, in which case, apologies for the redundant report.
```
cargo run -- compile-package --name language --src src --out target-javascript --target javascript
    Finished dev [unoptimized + debuginfo] target(s) in 0.10s
     Running `/home/jclemer/gleam/target/debug/gleam compile-package --name language --src src --out target-javascript --target javascript`

error: Unsupported feature for compilation target

```
2. There's some regretable/verbose test set-up in src/erl/tests.rs

### DIFF
--- a/src/build/package_compilation_tests.rs
+++ b/src/build/package_compilation_tests.rs
@@ -653,7 +653,7 @@ call_thing() ->
             },
             OutputFile {
                 path: PathBuf::from("_build/default/lib/the_package/src/one_Point.hrl"),
-                text: "-record(point, {x, y}).
+                text: "-record(point, {x :: integer(), y :: integer()}).
 "
                 .to_string(),
             },
@@ -1027,7 +1027,7 @@ funky() ->
             },
             OutputFile {
                 path: PathBuf::from("_build/default/lib/the_package/src/one_Person.hrl"),
-                text: "-record(person, {name, age}).
+                text: "-record(person, {name :: binary(), age :: integer()}).
 "
                 .to_string(),
             },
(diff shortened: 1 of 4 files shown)


src/docs.rs
```diff
@@ -17,7 +17,7 @@ use askama::Template;
 use itertools::Itertools;
 use std::path::Path;
 
-const MAX_COLUMNS: isize = 65;
+pub const MAX_COLUMNS: isize = 65;
```
jcc333, line 20: This seems like it might be a poor change to make; I'd be happy to take an alternate route

lpil: I think the use of a constant is good, but for the other Erlang we're using 80.

https://github.com/gleam-lang/gleam/blob/6d2ce262451a268739c2e44c1f42a6df825c57d5/src/erl.rs#L401

Let's extract that into a constant and share it between these two places? Keeping the docs one private.

lpil (commented):
Very nice! I'm surprised by how little code this is as well.

I think we may need to run `cargo fmt` to make CI happy.

We could improve this to have good formatting of long lines by converting the string building part of `record_definition` to use the printing algebra, though this could be a later PR.

jcc333:
- I rewrote history to take out the use of `docs::MAX_COLUMNS`
- "converting the string building part of record_definition to use the printing algebra" I may need a hint as to what that would entail - I'm not entirely sure I follow

lpil:
> "converting the string building part of record_definition to use the printing algebra" I may need a hint as to what that would entail - I'm not entirely sure I follow

Yes it's a bit unclear! Let's merge this in and if you're interested in adding full formatting support let me know and I'll explain more.

lpil:
Oh, and would you mind updating the changelog file? Thanks

lpil:
I'm gunna merge this. Thank you!!

PR #1127, opened 2021-05-25
build node image

CrowdHailer:
Same as Elixir/erlang images.

node as far as I can see has the same convention for labelling slim/alpine images.


lpil (approved):
Thank you

PR #1130, opened 2021-05-26
Revert "Draft for 'Typed records in erlang header files'"

jcc333:
Reverts gleam-lang/gleam#1121

@lpil I just wanted to double-check a couple of things to make sure this didn't introduce a bug:
- one of the CI steps was failing (the app template step) and I wanted to check that that wasn't expected in non-main branches or something like that
- I also haven't updated the changelog yet
If all is well in the execution I'm happy to submit a follow-up for the changelog and close the revert-PR. I just didn't want to leave this merged if it shouldn't be for app-template-reasons. I was struggling with figuring out where that was going wrong, so if it's as simple as "it won't work but that's alright", then that's great.

### DIFF
--- a/src/build/package_compilation_tests.rs
+++ b/src/build/package_compilation_tests.rs
@@ -653,7 +653,7 @@ call_thing() ->
             },
             OutputFile {
                 path: PathBuf::from("_build/default/lib/the_package/src/one_Point.hrl"),
-                text: "-record(point, {x :: integer(), y :: integer()}).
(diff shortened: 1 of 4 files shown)


lpil:
The problem with CI was an unrelated problem on the main branch that was my fault. Your work was perfect, thank you! I have updated the changelog

lpil:
Oh sorry, I have just spotted the unbound variable error on the latest CI build. I’ll investigate.

PR #1147, opened 2021-06-05
Generate JS for alternative patterns in case expressions

### LINKED ISSUE #1116: JS: Alternative patterns in case expressions
```rust
case x {
  1 | 2 -> 0
  _ -> 1
}
```

JohnDoneth:
Resolves #1116.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -1097,7 +1097,6 @@ impl<'comments> Formatter<'comments> {
     }
 
     fn tuple_index<'a>(&mut self, tuple: &'a UntypedExpr, index: u64) -> Document<'a> {
-        dbg!(tuple);
         match tuple {
             UntypedExpr::TupleIndex { .. } => self.expr(tuple).surround("{", "}"),
             _ => self.expr(tuple),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Gleam now supports alternative patterns in case expressions for the JavaScript target.
 - Gleam compiler now is catching values which are being imported more than once 
   in an unqualified fashion, this will raise a compile time error. 
 - Gleam can now compile Gleam projects without an external build tool.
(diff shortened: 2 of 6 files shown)


src/format.rs
```diff
@@ -1097,7 +1097,6 @@ impl<'comments> Formatter<'comments> {
     }
 
     fn tuple_index<'a>(&mut self, tuple: &'a UntypedExpr, index: u64) -> Document<'a> {
-        dbg!(tuple);
```
lpil, line 1100: Oops!

lpil:
Thank you! The code looks good.

With this approach I was wondering how guards and assignments would play with it.

```rust
case xs {
  [x] | [_, x] if x == 1 -> 1
  _ -> 0
}
```

Does this generate a guard that checks both `x`s?

```rust
case xs {
  [x] | [_, x] -> x
  _ -> 0
}
```

Does this return the right `x` in all situations?

My initial thought was that we would need to generate an `if (...) {...` per alternative, as we do in Erlang, but perhaps I'm wrong. It would be good to see some tests for these situations.

JohnDoneth:
Forgetting to properly support guards and assignments was definitely an oversight. Thank you for the feedback! 

> My initial thought was that we would need to generate an if (...) {... per alternative, as we do in Erlang, but perhaps I'm wrong.

I switched to this method as it seemed like the simplest way to properly support assignments.

> It would be good to see some tests for these situations.

Done! I added your examples as tests.

lpil (approved):
This is really nice! Your code is very clear. Thank you for those tests, they help me understand what is happening.

Would you be interested in adding some integration tests? Here's an example of some existing ones. https://github.com/gleam-lang/gleam/blob/23a1e9effc825fdfe68e0941b8715ef9bda0a257/test/language/src/main.gleam#L311
They can be ran with the command `make test` or `cd test/language && make`

And could you add an entry to the changelog file? Thank you

lpil (approved):
Excellent work! Thank you

lpil:
I have rebased these commits onto main manually as there was a merge conflict. Thank you

PR #1149, opened 2021-06-06
WIP: unify generic variables

### LINKED ISSUE #1096: Provide a hint when using a generic function without a return type annotation *before* it is defined.
Given the following snippet, we will get a compile error on **line 9**.
```rust
1 | import gleam/map.{Map}
2 | 
3 | pub fn insert (map: Map(a, b), key: a, value: b) -> Map(a, b) {
4 |    delete(map, key)
5 |        |> map.insert(key, value)
6 | }
7 |
8 | /// Note the missing return type annotation
9 | pub fn delete (map: Map(x, y), key: x) {
10|    map.delete(map, key)
11| }
```
This is because the compiler cannot unify the inferred type of `delete(map, key)` with the more general inferred type of delete's definition. 

This is the compiler being well-behaved and operating correctly, but might present a bit of a shock to those coming from Haskell/Elm/Purescript/... where order of definition isn't significant to the compiler. I stumbled across this when I _unintentionally_ left a return type annotation off a function and might've spent a long time working out what was up had I not noticed relatively shortly.

Adding a hint to the compile error in scenarios like these might be enough of a nudge to say "Hey, maybe you forgot some type annotations?" Something like:
```
error: Type mismatch
   ┌─ /.../minimal_map_example.gleam:10:5
   │
10 │     map.delete(map, key)
   │     ^^^^^^^^^^^^^^^^^^^^

annotation of this function.

Expected type:

    Map(a, b)

Found type:

    Map(c, d)

Hint: It looks like you used this function before I reached it's
definition. That's OK, but sometimes I need extra information
in those cases, did you leave out type annotations? See if
adding them resolves this issue.
```

shunjilin:
Attempt to close #1096

It seems like currently gleam does not attempt to unify unbound generic variables against each other, resulting in the compiler being unable to compile certain cases without explicit type annotation, see #1096 for more detail.

The attempted fix is to unify generic variables against each other, by assigning them to the same id.

This has caused some of the tests involving generics to break, but I believe they are false negatives. Instead of inferring `CouldNotUnify` for these tests, with this change the compiler only gives out actual type error only where the generics are instantiated with incompatible types.

Examples taken from the tests (I have commented out the tests for the time being):

<details>
<summary>Example 1</summary>
<p>
```gleam
pub type Box(a) {
  Box(value: a)
}

pub fn go(box1: Box(a), box2: Box(b)) {
  let _: Box(a) = box2
  let _: Box(b) = box1
  5
}

pub fn no_error() {
  go(Box(1), Box(2)) // a and b are of type Int, no error
}

pub fn error() {
  go(Box(1), Box("2") // Error: Expected type: Box(Int), Found type: Box(String)
}
```
</p>
</details>

<details>
<summary>Example 2</summary>
<p>
```gleam
pub type Box(a) {
  Box(value: a, i: Int)
}

pub fn update_box(box: Box(a), value: b) {
  Box(..box, value: value)
}

pub fn no_error() {
  update_box(Box(1.0, i: 1), 2.0) // a and b are of type Float, no error
}

pub fn error() {
  update_box(Box(1.0 i:1), 2) // Error: Expected type: Float, Found type: Int
}
```
</p>
</details>

If this approach is ok I would have to update and add new tests. I could be completely wrong any advice would be appreciated.

### DIFF
--- a/src/type_.rs
+++ b/src/type_.rs
@@ -330,6 +330,10 @@ impl TypeVar {
         matches!(self, Self::Unbound { .. })
     }
 
+    pub fn is_generic(&self) -> bool {
+        matches!(self, Self::Generic { .. })
+    }
+
     pub fn is_nil(&self) -> bool {
         match self {
             Self::Link { type_ } => type_.is_nil(),
--- a/src/type_/environment.rs
+++ b/src/type_/environment.rs
@@ -388,7 +388,7 @@ impl<'a, 'b> Environment<'a, 'b> {
 
                 TypeVar::Generic { id } => {
                     if let Type::Var { type_: typ } = t2.deref() {
-                        if typ.borrow().is_unbound() {
+                        if typ.borrow().is_unbound() || typ.borrow().is_generic() {
                             *typ.borrow_mut() = TypeVar::Generic { id: *id };
                             return Ok(());
                         }
--- a/src/type_/tests.rs
+++ b/src/type_/tests.rs
@@ -2795,40 +2795,40 @@ fn main() {
     );
 
     // Type variables are shared between function annotations and let annotations within their body
-    assert_module_error!(
-        "
-        pub type Box(a) {
-            Box(value: a)
-        }
-        pub fn go(box1: Box(a), box2: Box(b)) {
-            let _: Box(a) = box2
-            let _: Box(b) = box1
-            5
-        }",
-        Error::CouldNotUnify {
-            situation: None,
-            location: SrcSpan {
-                start: 139,
-                end: 143
-            },
-            expected: Arc::new(Type::App {
-                public: true,
-                module: vec!["my_module".to_string()],
-                name: "Box".to_string(),
-                args: vec![Arc::new(Type::Var {
-                    type_: Arc::new(RefCell::new(TypeVar::Generic { id: 8 })),
-                })]
-            }),
-            given: Arc::new(Type::App {
-                public: true,
-                module: vec!["my_module".to_string()],
-                name: "Box".to_string(),
-                args: vec![Arc::new(Type::Var {
-                    type_: Arc::new(RefCell::new(TypeVar::Generic { id: 10 })),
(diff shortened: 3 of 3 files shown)


lpil:
Hello! I think I might be misunderstanding this but it seems this code makes distinct generic type variables unifiable, which would change the meaning of the type system.

Today in the sig `fn(a, b) -> Int` the `a` is a shorthand for "for every type from the set of all possible types", which explains why you  cannot unify `a` and `b`- if `b` can be unified with `a` it means that `b` cannot be any type from the set of all types, instead it has to be the specific type of `a`. 

With this change we seem to lose the ability to say that generic types are distint from each other, but do not gain anything in exchange because we were already able to use the same generic type multiple times by specifying the same name.

With the generalisation issue linked it isn't limited to generic types, the problem can also arise with concrete types, so this would not solve that one either I'm afraid.

shunjilin:
Thank you very much for the explanation! Yes you are right, looking at the examples above, the compiler should not allow us to write such functions as they can possibly be wrong (if `a` and `b`) are distinct.

Let me dig a little deeper as to how to address the issue in #1096:
``rust
> 1 | import gleam/map.{Map}
> 2 | 
> 3 | pub fn insert (map: Map(a, b), key: a, value: b) -> Map(a, b) {
> 4 |    delete(map, key)
> 5 |        |> map.insert(key, value)
> 6 | }
> 7 |
> 8 | /// Note the missing return type annotation
> 9 | pub fn delete (map: Map(x, y), key: x) {
> 10|    map.delete(map, key)
> 11| }
> ```

where the error is:
```
> Expected type:
> 
>     Map(a, b)
> 
> Found type:
> 
>     Map(c, d)
```

lpil:
Perhaps instead of typing top to bottom we could attempt to re-order the functions by the way they are used. Would mean we traverse the AST twice, which a shame for performance, but perhaps it is OK.

PR #1155, opened 2021-06-13
Refactor slightly how iterators are used

tranzystorek-io:
As usual, some refactors around iterators 😄


lpil:
I'm unsure why CI has failed. I'll try running it again.

lpil (approved):
Let's see what happens on main

Issue #1156, opened 2021-06-15
Urgent: Fix CI release automation


lpil:
The current one we are using is failing as the tooling we use has since broken. **This means we cannot release v0.16**. Some detail: https://github.com/actions-rs/cargo/issues/183

It also uses inconsistent naming for the archives of each version.

Upgrade it to fix these problems.

This config from Embark could be used as a base, though any solution would be acceptable.
https://github.com/EmbarkStudios/cargo-deny/blob/e0f644df92aedf57dd1476254eb4942864731e19/.github/workflows/ci.yaml#L128-L207

As a bonus we could also enable aarch64 compilation for OSX and Linux.

We would need to test this in a new repo before merging in here.

PR #1157, opened 2021-06-16
Remove cross and add a new linux-amd64-musl workflow

tranzystorek-io:
Addresses #1156

Removes `cross` based building for musl binaries. Instead, builds gleam separately for the musl target inside the clux/muslrust container that is preconfigured to build rust projects statically with musl.

Also now the muslrust image is stable, instead of nightly.


lpil (approved):
Thank you!

PR #1159, opened 2021-06-17
Refactor for new Rust version

tranzystorek-io:
Highlights:

- nested or-patterns, yay! 😃
- import variants for selected enums that were too verbose to use as `LongEnumName::Variant`

### DIFF
--- a/src/bit_string.rs
+++ b/src/bit_string.rs
@@ -1,5 +1,4 @@
-use crate::ast::BitStringSegmentOption::{self, *};
-use crate::ast::SrcSpan;
+use crate::ast::{BitStringSegmentOption, SrcSpan};
 use crate::type_::Type;
 use std::sync::Arc;
 
@@ -40,6 +39,8 @@ impl<T> SegmentOptionCategories<'_, T> {
     }
 
     fn segment_type(&self) -> Arc<Type> {
+        use BitStringSegmentOption::*;
+
         match self.typ {
             Some(Int { .. }) => crate::type_::int(),
             Some(Float { .. }) => crate::type_::float(),
@@ -64,6 +65,8 @@ fn type_options<T>(
     value_mode: bool,
     must_have_size: bool,
 ) -> Result<Arc<Type>, Error> {
+    use BitStringSegmentOption::*;
+
     let mut categories = SegmentOptionCategories::new();
     // Basic category checking
     for option in input_options {
@@ -245,6 +248,8 @@ fn type_options<T>(
 }
 
 fn is_unicode<T>(opt: &BitStringSegmentOption<T>) -> bool {
+    use BitStringSegmentOption::*;
+
     matches!(
         opt,
         Utf8 { .. }
--- a/src/bit_string.rs
+++ b/src/bit_string.rs
@@ -1,4 +1,5 @@
-use crate::ast::{BitStringSegmentOption, SrcSpan};
+use crate::ast::BitStringSegmentOption::{self, *};
+use crate::ast::SrcSpan;
 use crate::type_::Type;
 use std::sync::Arc;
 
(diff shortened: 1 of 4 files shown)


tranzystorek-io:
This might have to wait a short while until stable muslrust has been built with Rust 1.53 😅

tranzystorek-io:
Update: action needs to be taken on the clux/muslrust repo side: clux/muslrust#78

lpil:
Hahah wow, immediately another musl problem!

src/bit_string.rs
```diff
@@ -40,18 +41,18 @@ impl<T> SegmentOptionCategories<'_, T> {
...
-            Some(BitStringSegmentOption::Float { .. }) => crate::type_::float(),
-            Some(BitStringSegmentOption::Binary { .. }) => crate::type_::bit_string(),
-            Some(BitStringSegmentOption::BitString { .. }) => crate::type_::bit_string(),
-            Some(BitStringSegmentOption::Utf8 { .. }) => crate::type_::string(),
-            Some(BitStringSegmentOption::Utf16 { .. }) => crate::type_::string(),
-            Some(BitStringSegmentOption::Utf32 { .. }) => crate::type_::string(),
-            Some(BitStringSegmentOption::Utf8Codepoint { .. }) => crate::type_::utf_codepoint(),
-            Some(BitStringSegmentOption::Utf16Codepoint { .. }) => crate::type_::utf_codepoint(),
-            Some(BitStringSegmentOption::Utf32Codepoint { .. }) => crate::type_::utf_codepoint(),
+            Some(Int { .. }) => crate::type_::int(),
```
lpil, line 44: Could we move the `use` into this function? I would prefer to have unqualified imports be as restricted as possible to make it clear where things are defined.

src/parse/lexer.rs
```diff
@@ -653,10 +653,7 @@ where
     // Test if a digit is of a certain radix.
     fn is_digit_of_radix(c: Option<char>, radix: u32) -> bool {
         match radix {
-            2 => matches!(c, Some('0'..='1')),
-            8 => matches!(c, Some('0'..='7')),
-            10 => matches!(c, Some('0'..='9')),
-            16 => matches!(c, Some('0'..='9') | Some('a'..='f') | Some('A'..='F')),
+            2 | 8 | 10 | 16 => matches!(c, Some(c) if c.is_digit(radix)),
```
lpil, line 656: Can we do something like this?

```rust
c.map(|c| c.is_digit(radix)).unwrap_or(false)
```

I found the matches a bit confusing with the guard

tranzystorek-io:
> 
> 
> Hahah wow, immediately another musl problem!

Well tbh this is only problematic with 1.53-specific features, so I'm kind of to blame for my hot blood 😝

src/parse/lexer.rs
tranzystorek-io: My initial idea was:
```rust
c.filter(|c| c.is_digit(radix)).is_some()
```
it has both a `filter` and a `is_some` that indicate more readably that it's a check

src/bit_string.rs
tranzystorek-io: My aim was mostly to shorten the matches further below that use or-patterns, because otherwise rustfmt does weird things like
```rust
Struct(
     field: Some(
         op
         @
         LongEnum::VariantA { .. }
         | LongEnum::VariantB { .. }
))
```
so maybe I could move the use only there?

lpil: I'm happy with it being unqualified, just would like the `use Enum::*` to be inside the module rather than at the top level
https://github.com/gleam-lang/gleam/pull/1159/files#diff-259329068baeb3eee63257f180ea5aff88a8ed160bb3ed2d14dc10fae02d7b1bR1

lpil (approved):

PR #1160, opened 2021-06-18
first implementation of try with a subpattern

CrowdHailer:
tackles #1137


src/javascript/expression.rs
```diff
@@ -347,7 +347,17 @@ impl<'module> Generator<'module> {
             // method to take a Document as a value and pass the subject
             // document into that also rather than letting the pattern generator
             // determine what it should be.
-            _ => return unsupported("Patterns in try"),
+            // adapting the `assignment` function would probably use `new_with_document`
```
CrowdHailer, line 350: Extra comments added here about cleaning up this code at a later stage.
Don't think it needs to be cleaned as part of this PR because this PR is small ish and sucessfully adds a feature.

lpil (approved):
Hey, that wasn't as much work as I thought it would be. Thank you

Issue #1140, opened 2021-06-02
Split compiler into core crate and command line crate


lpil:
To enable use of the compiler as a library compiled to WASM in the browser.

PR #1163, opened 2021-06-18
Add architecture guide

tranzystorek-io:
Proposed rough architecture guide, based on https://matklad.github.io/2021/02/06/ARCHITECTURE.md.html

Should be useful for new contributors, as well as for #1140

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -1,0 +1,44 @@
+# Architecture
+
+This document describes the high-level architecture of the Gleam transpiler.
+
+## Overview
+
+On the highest level, a `gleam` binary receives Gleam source code as input
+and produces one of its available backend formats:
+
+- Erlang source code (not Erlang Abstract Format, as e.g. Elixir does)
+- Javascript source code
+
+There is currently no formal specification of Gleam source,
+apart from the programmer-oriented [Language tour](https://gleam.run/book/tour/index.html).
+
+The code transpilation is for the most part oriented around the properties
+provided by Erlang's BEAM machine, including:
+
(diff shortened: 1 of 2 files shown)


tranzystorek-io:
@lpil I'll try to fill out the code map section, but I'll definitely need some proofreading.

My main goal is to write this guide in a way that makes redesign for #1140 a bit less painful. Especially, defining which modules are part of an API boundary (or if there is no clear one) would make e.g. creating a wasm playground much easier.

architecture.md
```diff
@@ -0,0 +1,44 @@
...
+## Overview
+
+On the highest level, a `gleam` binary receives Gleam source code as input
+and produces one of its available backend formats:
+
+- Erlang source code (not Erlang Abstract Format, as e.g. Elixir does)
+- Javascript source code
+
+There is currently no formal specification of Gleam source,
+apart from the programmer-oriented [Language tour](https://gleam.run/book/tour/index.html).
```
lpil, line 14: Can also note the integration tests in `test/language`. I would like to cover most the language runtime behaviour there.

architecture.md
```diff
@@ -0,0 +1,44 @@
...
+
+The code transpilation is for the most part oriented around the properties
+provided by Erlang's BEAM machine, including:
```
lpil, line 17: While it could be good to talk about this stuff I think it would be better suited to some kind of design documentation rather than architecture documentation.

lpil (approved):
I'm gunna merge this so I can add bits :)

PR #1164, opened 2021-06-19
Add basic info about parser design to docs

tranzystorek-io:


docs/architecture.md
```diff
@@ -71,9 +71,27 @@ Pay attention especially to the following sections:
 - **Architecture Invariant**: important note about crucial assumptions about the code/behavior
 - **API Boundary**: an exposed interface that requires careful design, e.g. because it has internal users to support
 
+### `ast`
+
+Types that serve as the Gleam AST, both in its *untyped* and *typed* form.
+Each tree describes a single Gleam module at its root.
+
+Untyped trees contain untyped expressions only.
+This form purely describes a sequence of characters.
```
lpil, line 80: What does this mean?

docs/architecture.md
```diff
@@ -71,9 +71,27 @@ Pay attention especially to the following sections:
...
+
+Typed trees contain typed expressions together with module metadata.
```
lpil, line 82:
```suggestion
Typed trees contain typed expressions together with metadata such as the types of expressions and information on how to generate code for each value constructor used.
```

docs/architecture.md
```diff
@@ -71,9 +71,27 @@ Pay attention especially to the following sections:
...
 ### `parse`
 
-TODO
+The parser receives source code for a single Gleam module (as `&str`, as opposed to a stream)
+and produces an untyped AST together with `ModuleExtra`
+containing various types of comments plus empty lines.
+
+Currently this means that parsing is tightly coupled with the AST structure.
+This might prove inflexible when creating tools such as a Gleam LSP.
+
```
lpil, line 93:
```suggestion
```

I think speculation about future features belongs in the issues and discussions for those features.

docs/architecture.md
tranzystorek-io: A bit clumsy attempt to say that this AST is free from semantic context, e.g. no additiona information about types, names, etc. If it's too unclear or redundant, we can remove this.

docs/architecture.md
tranzystorek-io: Agreed, my intention was to give an example where this coupling falls short when working with code.

tranzystorek-io: Replaced the LSP remark with an API boundary note about high-level operations

lpil (approved):
Thank you

PR #1166, opened 2021-06-19
Import aliased custom type constructors

### LINKED ISSUE #1126: Permit custom type constructors to be aliased when imported
```rust
import some/module.{Name as TheAlias}
```

We must ensure that the name is uppercase and that functions and consts are always lowercase when imported, so we'll need to validate this too.

matheus-consoli:
Hi! This should close #1126, allowing aliased imports of custom types :)

In this PR, we split the matching arms of `parse_unqualified_imports` into two branches:
    - the first, deals with patterns that start with a lowercase name, and should allow only lowercase aliases;
    - the second, deals with patterns that start with Uppercase names, here's is where we rely on the previous conditions to ensure that only types will be allowed to be imported with an uppercase name, as the definition is already previously verified.

The logic follows that a `Token::UpName` could be imported as, and only as, a `Token::UpName`; and that a `Token::Name` expects to be aliased as another `Token::Name`.

I added a test to ensure that this is working on the happy path, but I didn't figure out how to write a fail test (i.e. when a functional is imported with an uppercase name -- but I manually tested that this does not work).

> Also, note that this may be a breaking change, as some tests failed by types that were being imported with a lowercase name

### DIFF
--- a/src/build/package_compilation_tests.rs
+++ b/src/build/package_compilation_tests.rs
@@ -829,7 +829,7 @@ make() ->
                 origin: Origin::Src,
                 path: PathBuf::from("/src/two.gleam"),
                 name: "two".to_string(),
-                code: "import one.{Empty as e, id as i} fn make() { i(e) }".to_string(),
+                code: "import one.{Empty as E, id as i} fn make() { i(E) }".to_string(),
             },
         ],
         Ok(vec![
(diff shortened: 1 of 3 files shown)


lpil (approved):
Very nice, thank you! And thanks for waiting for me to get round to this one, I have had a busy couple of weeks.

Issue #1169, opened 2021-06-21
Importing a module and a function with the same name causes a runtime error [JS]


pd-andy:
Given an import like:
```js
import dom/svg.{svg}
```
We produce the js:
```
import * as svg from "./dom/svg.js";
const { svg } = svg;
```
It seems this binding does nothing (and if it did, it would shadow the import) because when the compiled code calls:
```js
  return svg(...)
```
I get the runtime error `svg is not a function`.

Maybe we could do something like capitalising imports or prepending `$`, something that you can't write in gleam code, to avoid these name collisions:
```js
// gleam
import dom/svg.{svg}

// js
import * as Svg from "..."
const { svg } = Svg
```

lpil:
Nice find, thank you

PR #1178, opened 2021-07-01
Camel case JS module names

lpil:
Does this look like a suitable solution for https://github.com/gleam-lang/gleam/issues/1169 to you @pd-andy ?


pd-andy:
Yeah looks excellent, exactly how I'd have imagined it. FWIW this is called `PascalCase` not `CamelCase`, that's a bit nit picky but it might confuse someone down the line.

lpil:
Yeah I think you're right, but the library calls it camel so let's go with that I guess.

lpil:
Thank you!

PR #1211, opened 2021-07-16
add a docker option to the new project command

CrowdHailer:
Maybe there's not much future in docker-compose.

However following perfection is the enemy of progress argument, this option is here now, and I will use it on my fork quite happily if not merged.
I don't have erlang rebar on my machines. I do have docker so I will now do the following.

```sh
gleam new foo --docker
cd foo
docker-compose run foo bash
rebar3 eunit.
```

I don't know the trade-offs but more people have docker than erlang. and installing erlang/rebar has seemed to be a blocker for some new gleam users.

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -158,6 +158,13 @@ pub struct NewOptions {
         default_value = "lib"
     )]
     pub template: new::Template,
+
+    #[structopt(
+        long,
+        possible_values = &new::Docker::VARIANTS,
+        case_insensitive = true,
+    )]
+    pub docker: Option<Option<new::Docker>>,
 }
 
 #[derive(StructOpt, Debug)]
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -20,7 +20,8 @@
     clippy::inefficient_to_string,
     clippy::linkedlist,
     clippy::macro_use_imports,
-    clippy::option_option,
+    // Used for optional cli args that take optional values
+    // clippy::option_option,
     clippy::verbose_file_reads,
     clippy::unnested_or_patterns,
     rust_2018_idioms,
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -25,6 +25,14 @@ pub enum Template {
     Escript,
 }
 
+#[derive(Debug, Serialize, Deserialize, Display, EnumString, EnumVariantNames, Clone, Copy)]
+#[strum(serialize_all = "kebab_case")]
+pub enum Docker {
+    Default,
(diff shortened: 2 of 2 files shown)


compiler-cli/src/main.rs
```diff
@@ -158,6 +158,13 @@ pub struct NewOptions {
         default_value = "lib"
     )]
     pub template: new::Template,
+
+    #[structopt(
+        long,
+        possible_values = &new::Docker::VARIANTS,
+        case_insensitive = true,
+    )]
+    pub docker: Option<Option<new::Docker>>,
```
CrowdHailer, line 167: Option Option allows the command to be used as `--docker` without any argument or `--docker default/slim/alpine`

compiler-cli/src/new.rs
```diff
@@ -526,6 +542,44 @@ pub fn hello_world_test() {{
...
+                    Docker::Alpine => "-alpine",
+                };
+                write(
+                    self.root.join("docker-compose.yml"),
+                    &format!(
+                        r#"version: "3"
+    
+services:
+  {name}:
+    image: ghcr.io/gleam-lang/gleam:v{version}-erlang{docker_type}
```
CrowdHailer, line 561: Note this will not work for people building gleam from master because the version is set to `x.y.z-dev` and there is no corresponding docker image.

I think this is an ok trade-off because people willing to work of main are probably ok changing this value.

CrowdHailer:
Also totally agree with the point about having libraries be able to specify there own templates. Love it.

But I guess that would come after the tooling we the is already in milestones so not a near term thing.

compiler-cli/src/new.rs
CrowdHailer: Second note this always assumes you want a  erlang version because projects generated by a gleam new are rebar3 projects. 

There might be a future where you generate new mix/npm projects in which case when adding those you would want to choose the approprate image. However because of native gleam tooling there might not be a future where we generate mix/npm projects so for now assuming erlang seems sensible

CrowdHailer:
Moved from discord for posterity

> Though I don't want a docker project in the compiler core, would need to be some kind of plug in etc

Fair enough, there is already code for releasing docker images in this Repo.

> I intend to only have gleam based projects built into the compiler, even rebar will be removed in future

Like the direction, I consider this feature something that would also be removed when possible and as other things mature, like the tooling for using project templates that are not defined in the core.

lpil:
Is there a useful tool this could be moved into today? Hmm

CrowdHailer:
Not that is immediately obvious to me.

lpil:
I'm going to close this as I'm not sure how to move forward with this one currently, but let's revisit this in future.

PR #1213, opened 2021-07-17
Recusive variables are not escaped.

CrowdHailer:
I also had a look at changing things here https://github.com/gleam-lang/gleam/blob/a0aa0d1d65c3e186215f9c4558ad93e36efcb038/compiler-core/src/javascript/expression.rs#L61

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -593,7 +593,10 @@ impl<'module> Generator<'module> {
                     }
                     // Create an assignment for each variable created by the function arguments
                     if let Some(name) = argument {
-                        docs.push(Document::String(format!("{} = ", name)));
+                        // TODO escape here
+                        // docs.push(Document::String(format!("{}!! = ", name)));
+                        docs.push(maybe_escape_identifier(name));
+                        docs.push(" = ".to_doc());
                     }
                     // Render the value given to the function. Even if it is not
                     // assigned we still render it because the expression may
--- a/compiler-core/src/javascript/tests/functions.rs
+++ b/compiler-core/src/javascript/tests/functions.rs
@@ -465,3 +465,21 @@ export function main(_, _1, _2) {
 "#
     );
 }
+
+#[test]
+fn keyword_in_recursive_function() {
+    assert_js!(
+        r#"pub fn main(with: Int) -> Nil {
+  main(with - 1)
+}
+"#,
+        r#""use strict";
+
+export function main(with$) {
+  while (true) {
+    with$ = with$ - 1
+  }
+}
+"#
+    );
+}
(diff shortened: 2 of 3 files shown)


compiler-core/src/javascript/tests/functions.rs
```diff
@@ -465,3 +465,21 @@ export function main(_, _1, _2) {
...
+    assert_js!(
+        r#"pub fn main(with: Int) -> Nil {
+  main(with - 1)
+}
+"#,
+        r#""use strict";
+
+export function main(with$) {
+  while (true) {
+    with$ = with$ - 1
```
CrowdHailer, line 480: Currently this will produce the line
```
with = with$ - 1
```

compiler-core/src/javascript/expression.rs
```diff
@@ -593,7 +593,10 @@ impl<'module> Generator<'module> {
                     }
                     // Create an assignment for each variable created by the function arguments
                     if let Some(name) = argument {
-                        docs.push(Document::String(format!("{} = ", name)));
+                        // TODO escape here
+                        // docs.push(Document::String(format!("{}!! = ", name)));
+                        docs.push(maybe_escape_identifier(name));
```
CrowdHailer, line 598: This seemed like a simple fix. But I got many lifetime errors

```
error[E0495]: cannot infer an appropriate lifetime due to conflicting requirements
   --> compiler-core/src/javascript/expression.rs:588:50
    |
588 |                     .zip(self.function_arguments.clone())
    |                                                  ^^^^^
    |
note: first, the lifetime cannot outlive the lifetime `'module` as defined on the impl at 34:6...
   --> compiler-core/src/javascript/expression.rs:34:6
    |
34  | impl<'module> Generator<'module> {
    |      ^^^^^^^
note: ...so that the types are compatible
   --> compiler-core/src/javascript/expression.rs:588:50
    |
588 |                     .zip(self.function_arguments.clone())
    |                                                  ^^^^^
    = note: expected `&Vec<std::option::Option<&str>>`
               found `&Vec<std::option::Option<&'module str>>`
note: but, the lifetime must be valid for the lifetime `'a` as defined on the method body at 536:27...
   --> compiler-core/src/javascript/expression.rs:536:27
    |
536 |     fn call_with_doc_args<'a>(
    |                           ^^
note: ...so that the expression is assignable
   --> compiler-core/src/javascript/expression.rs:607:17
    |
607 |                 Ok(docs.to_doc())
    |                 ^^^^^^^^^^^^^^^^^
    = note: expected `std::result::Result<pretty::Document<'a>, _>`
               found `std::result::Result<pretty::Document<'_>, _>`

```

lpil: oh how strange. I'll investigate

lpil: This is a fiddly one to do with lifettimes so I've just duplicated some code and cloned it.

PR #1215, opened 2021-07-20
Refactor iterator usage

tranzystorek-io:
Highlights:

- Make functions accept `IntoIterator` where it makes sense
- Use arrays instead of chaining `std::iter::once`, or instead of temporary vecs
- Remove redundant `into_iter()` calls


lpil (approved):
Hey that's really nice. I like the use of arrays. Thanks!

Issue #1216, opened 2021-07-21
Remove deprecated tuple syntax


lpil:
```rust
tuple(1, 2)
```

tranzystorek-io:
Do we want to do hard deprecate of that syntax? Or just some sort of soft warning for now?

lpil:
Let's just remove it. The current version can be used to upgrade.

aflatter:
Hello, I've just started working on this and hope to make a PR soon. 👋

PR #1229, opened 2021-08-01
Remove old tuple syntax

aflatter:
Hi there,

I've been following Gleam for some time and this week-end I noticed that it's implemented in Rust and there's a couple of issues that are approachable enough to jump right in. So here's my attempt at #1216.


lpil (approved):
Brilliant, thank you very much!

Sorry for the slighty delay, I'm kicking off my flat renovation and it's a whole bunch of faff

aflatter:
> Brilliant, thank you very much!
> 
> Sorry for the slighty delay, I'm kicking off my flat renovation and it's a real time sink

Perfectly fine, thanks for merging and have fun redecorating!

PR #1232, opened 2021-08-07
Remove 'use strict' from generated js

### LINKED ISSUE #1225: Remove "use strict"; from generated JS modules
ES modules are always strict, so it is not needed.

michaeljones:
As noted in #1225, it is not required as ES modules are always strict.

Tests updated with something like:

    grep -R -l 'use strict' . | xargs sed -i -z 's/"use strict";\n\n//g'

And some manual adjustments.

Apparently '-z' causes sed to more or less treat the whole file as one
line so you can use it to work across new lines.

Closes #1225

### DIFF
--- a/compiler-core/src/javascript.rs
+++ b/compiler-core/src/javascript.rs
@@ -87,12 +87,11 @@ impl<'a> Generator<'a> {
     }
 
     pub fn compile(&mut self) -> Output<'a> {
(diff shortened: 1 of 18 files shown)


michaeljones:
I'm guessing this is fine but I've little familiarity with the code base at this point so this was largely done on grep results :)

lpil (approved):
Oh nice! Your shell-script-fu is stronger than mine

PR #1242, opened 2021-08-14
🐛 Fixed bug where `equal` was far too optimistic.

pd-andy:
Things we can now handle properly:
- Structural equality for object-like things (objects, arrays, classes).
- Value equality for idiots that do `new Number(10)` somehow.
- Structural equality for all `TypedArray`s.
- Fancy `Date` equality that we stole from ReScript.

---

> 🚨 **Note**: This is using `Object.prototype.constructor` and **not** `Object.prototype.constructor.name`. This means equality will break if the prelude is included in every Gleam module individually. I can change it back to `.name` if that's the case.

### DIFF
--- a/compiler-core/templates/prelude.js
+++ b/compiler-core/templates/prelude.js
@@ -174,7 +174,7 @@ function dateEqual(a, b) {
 }
 
 function typedArrayEqual(a, b) {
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thanks !

PR #1247, opened 2021-08-16
JS: fix codegen of case subjects (fixes #1237)

### LINKED ISSUE #1237: Pipe in case statement generates invalid JS
```rs
fn repro() {
  case 1 |> io.debug() {
    1 -> Nil
  }
}
```
I do this quite a lot when debugging, but in JS land the subject of a case is not properly wrapped.
Generated js is as follows
```js
function repro() {
  let $ = let _pipe = 1;
  Io.debug(_pipe);
  if ($ === 1) {
    return undefined;
  } else {
    throw new Error("Bad match");
  }
}

```

utkarshkukreti:
Hi! Not 100% sure if this is the right way, but the output looks correct to me. Please let me know what you think!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,8 @@
   Erlang.
 - Fixed a bug where discard variables used in assert assignments would generate
   invalid Erlang code.
+- Fixed a bug where some expressions as case subjects would generate invalid
+  JavaScript code.
 
 ## v0.16.1 - 2021-06-21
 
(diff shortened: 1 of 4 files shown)


lpil (approved):
That is perfect, thank you very much.

Would you mind also updating the CHANGELOG file?

utkarshkukreti:
@lpil done!

lpil:
Thank you!

Sorry about the delay, I have been very busy.

PR #1261, opened 2021-08-31
Add support for map equality check

UnsolvedCypher:
Currently, map comparisons do not work properly. This change is necessary for the tests in gleam-lang/stdlib#232 to pass.
I'm not at all familiar with how the compiler works so this may not be the best way to do it, please let me know if there are any changes needed :)

### DIFF
--- a/compiler-core/templates/prelude.js
+++ b/compiler-core/templates/prelude.js
@@ -235,11 +235,19 @@ export function isEqual(x, y) {
       !structurallyCompatibleObjects(a, b) ||
       unequalDates(a, b) ||
       unequalBuffers(a, b) ||
-      unequalArrays(a, b);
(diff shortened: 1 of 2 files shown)


lpil:
Would you mind adding some tests in here? https://github.com/gleam-lang/gleam/blob/main/test/javascript_prelude/main.js#L164

Could be good to have the same for `WeakMap` too

UnsolvedCypher:
I've included the tests. I'm not sure that structural equality is even possible with WeakMap because there is no way to get the keys of a WeakMap. However, I added support for Set comparisons instead.

lpil (approved):
Wonderful, thank you very much!

Ah yes, I didn't realise you couldn't get the keys for a weakmap.

PR #1333, opened 2021-11-11
Add new refactorings

tranzystorek-io:
Higlights:

- Another round of `Deref` simplifications
- Initialize `HashMap`s with `[...].into()`
- Initialize wrapped `&Path`s instead of `Pathbuf`s where applicable
- Rust 2021: `[...].into_iter()` finally works!


compiler-cli/src/dependencies.rs
```diff
@@ -114,13 +114,12 @@ impl Manifest {
         let toml = toml::to_vec(&self).expect("manifest.toml serialization");
         let mut file = fs::writer(&path)?;
         file.write(
-            "# This file was generated by Gleam
+            b"# This file was generated by Gleam
```
lpil, line 117: TIL

lpil (approved):
Forever making the compiler that bit nicer! Thank you

Issue #1353, opened 2021-11-27
Automatically include link to GitHub etc in Hex package links if we know the repo


lpil:
If the `gleam.toml` includes repo config

```toml
repository = { type = "github", user = "gleam-lang", repo = "erlang" }
```

Then we should include a link to the repo automatically

https://github.com/gleam-lang/gleam/blob/cbe11776c228874efcf7d8fcfa39303d2d7f3b78/compiler-cli/src/publish.rs#L237-L276

So it shows up here

![image](https://user-images.githubusercontent.com/6134406/143680174-2c74c400-200b-4f7e-84fe-0eddfb49042d.png)

PR #1354, opened 2021-11-27
Validate Link before publishing

TristanCacqueray:
This change decodes link href to prevent using invalid Uri.

Related #1353

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use flate2::{write::GzEncoder, Compression};
-use gleam_core::{config::Link, config::PackageConfig, hex, Error, Result};
+use gleam_core::{config::PackageConfig, hex, Error, Result};
 use hexpm::version::{Range, Version};
 use itertools::Itertools;
 use sha2::Digest;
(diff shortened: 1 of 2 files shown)


TristanCacqueray:
This is probably not very good rust as I'm not very familiar with the language but that seems to work as expected. In particular I'm not quite sure why I had to map again L156 to create the `&str` link name, when trying to do it in the parse_link resulted in all sort of weird errors :-)

Would it make sense to push the validation in the core de-serialization by changing the original Link struct instead?

lpil:
Hello! Thanks for this, a good improvement for sure.

> Would it make sense to push the validation in the core de-serialization by changing the original Link struct instead?

That's a nice idea! Let's go with that one.

TristanCacqueray:
Good, that's easier like that. I added extra checks to ensure the host and scheme are defined, otherwise `blah` is valid according to `http::Uri`. Though perhaps relative path are valid, in that case we could use http-serde instead.

lpil (approved):
Very nice! Really clean. Thank you

lpil:
Happy with the commit history 👍

TristanCacqueray:
Great, thank you!

Issue #1346, opened 2021-11-23
`gleam deps list`


lpil:
A command that prints all dependencies

tommyschaefer:
Hi there! I'd love to give this a go if possible 😄

If so, do you have a particular formatting in mind? My initial thought was something along the lines of:

```
From gleam.toml:

    *  gleam_stdlib (~> 0.18)                                    
    *  gleam_otp (~> 0.3)
```

But I also took a look at other dependency list formats:

Ruby `gem list`:

```

*** LOCAL GEMS ***

abbrev (default: 0.1.0)
base64 (default: 0.1.0)
...
```

Elixir `mix deps`:

```
* bunt 0.2.0 (Hex package) (mix)
  locked at 0.2.0 (bunt) 7af5c7e0
  ok
* credo 1.6.1 (Hex package) (mix)
  locked at 1.6.1 (credo) 698607fb
  ok
* dialyxir 1.1.0 (Hex package) (mix)
  locked at 1.1.0 (dialyxir) 07ea8e49
  ok
...
```

Node `npm ls`:

```
assets@ /Users/tom/code/waffle_foundry/assets
├── @babel/core@7.13.15
├── @babel/preset-env@7.13.15
├── @tailwindcss/forms@0.3.2
...
```

Rust `cargo tree --depth 1`:

```
...

gleam-wasm v0.18.0-rc2 (/Users/tom/code/gleam/compiler-wasm)
├── gleam-core v0.18.0-rc2 (/Users/tom/code/gleam/compiler-core) (*)
└── wasm-bindgen v0.2.78
[dev-dependencies]
└── wasm-bindgen-test v0.3.28
```

Of these, I quite like Rust's output, and think that might serve as helpful inspiration 😄 What are your thoughts?

lpil:
I think a flat list of all locked packaged would be good. For now we could show the name and version, but perhaps we could also show the licence in future.

I quite like the first format you've suggested. I think the Elixir one is a bit noisy.

Later on we could have a `gleam deps tree` for the nested format that shows why a package is selected, showing multiple entries for packages depended on multiple times.

tommyschaefer:
That sounds great! I’ll give it a try and submit a PR! 😄

PR #1355, opened 2021-11-27
Add deps list command

tommyschaefer:
Hi there! This is my first pass at fixing #1346 😄 

Based on our conversation in the issue, I was targeting this format:

```
From gleam.toml:

    *  gleam_stdlib (~> 0.18)                                    
    *  gleam_otp (~> 0.3)
```

While implementing it, I ran into a couple of questions

- Should we distinguish between development dependencies?
- Should we show the version range specified in the config, or the concrete version in the manifest?

I wasn't sure what the best answer to these questions were, but I decided to start with a label for development dependencies and version ranges. When running `gleam deps list` from the test project, here's the current output:

```
project deps (from gleam.toml):

    * gleam_stdlib (~> 0.18)
    * gleam_otp (~> 0.3)
    * first_gleam_publish_package (> 0.0.0) (dev)
    * accept (~> 0.3) (dev)
    * gleeunit (~> 0.1) (dev)
```

I'm totally happy to make any changes you think would be best, though! 😄 Thank you so much for your time!

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -25,8 +25,10 @@ use crate::{
 };
 
 pub fn list() -> Result<()> {
+    let runtime = tokio::runtime::Runtime::new().expect("Unable to start Tokio async runtime");
+
     let config = crate::config::root_config()?;
-    let manifest = read_manifest_from_disc()?;
+    let (_, manifest) = get_manifest(runtime.handle().clone(), Mode::Dev, &config)?;
     list_manifest_packages(std::io::stdout(), config.name, manifest)
 }
 
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -24,6 +24,26 @@ use crate::{
     http::HttpClient,
 };
 
+pub fn list() -> Result<()> {
+    let config = crate::config::root_config()?;
+
+    println!(
+        "{} deps (from {}):\n",
+        config.name,
+        paths::root_config().display().to_string(),
+    );
+
+    for (name, version) in config.dependencies {
+        println!("    * {} ({})", name, version);
+    }
+
+    for (name, version) in config.dev_dependencies {
+        println!("    * {} ({}) (dev)", name, version);
+    }
+
+    Ok(())
+}
+
 pub fn download(new_package: Option<(&str, bool)>) -> Result<Manifest> {
     let span = tracing::info_span!("download_deps");
     let _enter = span.enter();
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -54,7 +54,52 @@ fn list_manifest_packages<W: std::io::Write>(
 
 #[test]
 fn list_manifest_format() {
-    assert_eq!("Hello", "there")
+    let mut buffer = vec![];
+    let manifest = Manifest {
+        requirements: HashMap::new(),
+        packages: vec![
+            ManifestPackage {
+                name: "root".to_string(),
+                version: Version::parse("1.0.0").unwrap(),
+                build_tools: ["gleam".into()].into(),
+                otp_app: None,
+                requirements: vec![],
+                source: ManifestPackageSource::Hex {
+                    outer_checksum: Base16Checksum(vec![1, 2, 3, 4]),
+                },
+            },
+            ManifestPackage {
+                name: "aaa".to_string(),
+                version: Version::new(0, 4, 2),
+                build_tools: ["rebar3".into(), "make".into()].into(),
+                otp_app: Some("aaa_app".into()),
+                requirements: vec!["zzz".into(), "gleam_stdlib".into()],
+                source: ManifestPackageSource::Hex {
+                    outer_checksum: Base16Checksum(vec![3, 22]),
+                },
+            },
+            ManifestPackage {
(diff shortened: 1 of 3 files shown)


lpil:
Looks good so far!

My thoughts:

I would like to show all the deps, even the ones that are not directly depended upon. We can get this from the manifest.toml file.

Doing this would make telling whether it's a development dependency or not more challenging, but we can omit that for now.

It would be nice to have a test for the output, so perhaps we can write it to a string or buffer first, similar to how we do for converting the manifest to toml

tommyschaefer:
> I would like to show all the deps, even the ones that are not directly depended upon. We can get this from the manifest.toml file.

Ah, yes! Makes sense to me!

> Doing this would make telling whether it's a development dependency or not more challenging, but we can omit that for now.

:+1:

> It would be nice to have a test for the output, so perhaps we can write it to a string or buffer first, similar to how we do for converting the manifest to toml

Oh cool! I’ll use that as a guide and give this a go.

Thank you so much for your feedback!!

tommyschaefer:
Hi there! Apologies in advance, I'm quite new to Rust and may have gone about things in the wrong way! Happy to make any changes you think would be best! 😄 

I attempted to follow the [manifest writing example](https://github.com/gleam-lang/gleam/blob/235eb17ff473cef461cdca0478302e23ba610aea/compiler-core/src/project/manifest.rs#L20) and [its test](https://github.com/gleam-lang/gleam/blob/235eb17ff473cef461cdca0478302e23ba610aea/compiler-core/src/project/manifest.rs#L94) as closely as I could. I ran into a few snags, though, and am not sure if I went about things in the correct way. Here are some of the snags I ran into, and what I did to solve them:

- I started out by having `list()` take a buffer that implemented `std::fmt::Write`, mirroring the manifest writing function, but had trouble figuring out how to get `Stdout` to work with this. I opted to try using `std::io::Write` instead, since `Stdout` implements that trait already.

- From here, I looked into using the `WrappedWriter` to handle error conversion, but realized I didn't have a path to provide, since I was wanting to write to standard out. I noticed there was a `StandardIo` error, but no `Write` `StandardIoAction`. I decided to add a `Write` action and map any `std::io::Error`s to this.

- From there, I wanted to add a test of the output. I was originally wanting to test the `list()` function directly, but was struggling to do so without passing in a manifest from `main`, which got pretty ugly pretty quickly. I decided to try pulling out a helper function that does the majority of the work, test that, and leave `list()` itself untested.

Again, I'm very new to Rust, so I know any / all of these could have been not-so-great decisions 😂 Totally happy to go back to the drawing board!!

I had a few other odds and ends questions, but I'll add those as comments on specific lines of code 😄 

Thank you so much for your time!

compiler-cli/src/dependencies.rs
```diff
@@ -24,6 +24,84 @@ use crate::{
     http::HttpClient,
 };
 
+pub fn list() -> Result<()> {
+    let config = crate::config::root_config()?;
+    let manifest = read_manifest_from_disc()?;
```
tommyschaefer, line 29: I chose to use this function, but realized there's also [`get_manifest`](https://github.com/gleam-lang/gleam/blob/235eb17ff473cef461cdca0478302e23ba610aea/compiler-cli/src/dependencies.rs#L318-L342). If I understand correctly, `read_manifest_from_disc` only reads the manifest from the toml file, whereas `get_manifest` reads and resolves the manifest versions. I wasn't sure which would be desired for this command, so went with the simpler of the two to start with. Which (or neither 🤷 😄) do you think would be best?

compiler-cli/src/dependencies.rs
```diff
@@ -24,6 +24,84 @@ use crate::{
...
+    list_manifest_packages(std::io::stdout(), config.name, manifest)
```
tommyschaefer, line 30: I noticed there's also [`cli::stdout_buffer_writer`](https://github.com/gleam-lang/gleam/blob/235eb17ff473cef461cdca0478302e23ba610aea/compiler-cli/src/cli.rs#L134-L137). Would it be best to write to that instead? 😄

lpil: I think this is good as we're not using any colourful printing features

compiler-cli/src/dependencies.rs
lpil: I think if we are to read it from the disc we would want to create a friendly error message for the situation in which the manifest is absent, so the user knows how to process.

Having said that, is there any scenario in which the user would want to list the deps but not permit them to be resolved beforehand? I think it's likely OK to use the other function so that thet can be resolved and this command can succeed if run before dep resolution.

tommyschaefer: Ah, yes! That makes sense to me. I hadn't considered what would happen if the user didn't have a manifest yet. I'll update things to use the other function 😄 Thank you!

tommyschaefer:
I've updated things to use `get_manifest` now 😄 One question, when dependencies need to be resolved, "Resolving dependencies" gets printed out and is indented a bit:

image

Is that okay?

lpil:
Yeah that's good :)

I'll just run the test now

lpil (approved):
Thank you!

tommyschaefer:
Thank you so much for your help and feedback on this!! This was such a blast to work on! 😄

lpil:
Ah thank you, that means a lot! 💜

PR #1366, opened 2021-12-04
[pretty] Simplify fits

ddfisher:
This PR simplifies the `fits` function. The reference implementation in _Strictly Pretty_ (which this mirrors) is actually overly complex: `indent` and `mode` are unused and can be removed without changing behavior.

**I know this PR is both out of the blue and unimportant -- totally understand if it's not something you want to spend time reviewing.**

### Why `indent` is Unused
`indent` is tracked and passed around, but is never used in a position which can affect the return value of `fits`.
### Why `mode` is Unused
`fits` is only called with `Mode::Unbroken` (and there's no point in calling it with `Mode::Broken` -- at that point you've already decided to split the rest of the line) and can only ever stay the same or transition to `Mode::Unbroken`. I.e. `mode` will _always_ be `Mode::Unbroken`.

### DIFF
--- a/compiler-core/src/pretty.rs
+++ b/compiler-core/src/pretty.rs
@@ -139,13 +139,13 @@ enum Mode {
     Unbroken,
 }
 
-fn fits(mut limit: isize, mut docs: im::Vector<(isize, Mode, Document<'_>)>) -> bool {
+fn fits(mut limit: isize, mut docs: im::Vector<Document<'_>>) -> bool {
     loop {
         if limit < 0 {
             return false;
         };
 
-        let (indent, mode, document) = match docs.pop_front() {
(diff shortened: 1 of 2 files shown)


lpil:
Oh wow, you've blown my mind! It is so obvious after your detailed and helpful explanation, yet I've not noticed this potential improvement for years. Cool stuff.

> I know this PR is both out of the blue and unimportant -- totally understand if it's not something you want to spend time reviewing.

You're very kind, but I am excited to get contributions like these. Thank you very much

+ Welcome to Gleam! :)

ddfisher:
Yay! I'm glad this was a positive contribution!  Thanks for the warm welcome. ^_^

PR #1370, opened 2021-12-05
Fix: Moved licence to top level

TakshakRamteke:
- [x] moved to license to top level
- [x] removed the license in individual compiler's


TakshakRamteke:
amm.. hey i was wonder if we should add a like this to the README : 
```
## License
All the code for this project is under Apache License 2.0
``` 
i didn't add it in because i was a little confused if i should do it or not

lpil (approved):
Thank you!

I think this is OK without the README addition

PR #1374, opened 2021-12-06
Add subcommand for generating shell completion scripts

tranzystorek-io:
Since clap gives those basically for free 😄

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- `gleam` learned a new subcommand `completion` that generates shell completion scripts.
+
 ## v0.18.0 - 2021-12-06
 
 - New projects now include `gleeunit`.
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -87,7 +87,10 @@ use std::{
     io::Write,
     path::{Path, PathBuf},
 };
-use structopt::{clap::{AppSettings, Shell}, StructOpt};
+use structopt::{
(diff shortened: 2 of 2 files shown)


lpil:
Very cool! How do we use it?

tranzystorek-io:
@lpil depends on script, but basically for example `gleam completion zsh > <completion_dir>/_gleam`, usually package maintainers put them in global completion paths like `/usr/share/bash-completion/completions/` or `usr/share/zsh/site-functions/`

tranzystorek-io:
A user typically configures a local completion directory and adds it to an env var like `FPATH` in zsh

tranzystorek-io:
Ugh, seems `AppSettings::Hidden` doesn't hide commands from the completion generator

obraz

lpil:
Hmm, that is a bit annoying. It could be OK if we remove the print-config command.

Perhaps newer versions of clap might hide them? It seems odd they are not hidden.

tranzystorek-io:
> Perhaps newer versions of clap might hide them?

Did a quick check on clap 3.0.0-beta.5 for my own project, hidden apps still show up in scripts 🙁

lpil:
It looks like it will be fixed in a future release? https://github.com/clap-rs/clap/issues/2541

tranzystorek-io:
From the discussion it seems the status is back to [#1335](https://github.com/clap-rs/clap/issues/2541#issuecomment-991113676)

lpil:
Ah, that is a shame

lpil:
I'm going to close this as we're a bit stuck, but please reopen once we can hide commands.

PR #1377, opened 2021-12-07
Apple Silicon support in releases

moogle19:
I added a target for the `aarch64-apple-darwin` toolchain, to cross compile an Apple Silicon version on the default macOS runner, since currently there aren't no Apple Silicon runners on GitHub.

I tested the release build in my fork (https://github.com/moogle19/gleam/releases/tag/untagged-c219bd7e74291f9abbcb) and the binary locally on my M1 Mac and it seems to work just fine.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,13 +26,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: [macos, windows-64bit]
+        build: [macos, macos-arm, windows-64bit]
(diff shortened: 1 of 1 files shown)


lpil:
Wow! I had no idea it would be that straightforward. Brilliant stuff, thank you.

moogle19:
@lpil Yeah,the only downside is, that it is currently not possible to run tests on arm64.
Is it possible to trigger the release action manually to add an arm64 binary to the current release?

lpil:
That's a good idea. I can build it manually and add it

PR #1388, opened 2021-12-12
Fix arm64 builds for macOS

### LINKED ISSUE #1387: OSX ARM64 release build fails
https://github.com/gleam-lang/gleam/runs/4497851708?check_suite_focus=true

Tagging @moogle19 who may know more

moogle19:
Fixes #1387.

I changed the macOS version to `macos-11` for the arm64 builds.

For whatever reason my test builds used macOS 11 although `macos-latest` was defined as version, which should default to macOS 10.15. The cross compilation only works in macOS >= 11.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -36,7 +36,7 @@ jobs:
             os: macos-latest
             target: x86_64-apple-darwin
           - build: macos-arm64
-            os: macos-latest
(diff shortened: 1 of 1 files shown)


lpil (approved):
Good find, thank you.

I think GitHub rolls out version upgrades in stages, just for an extra bit of randomness for CI builds 🙄

Issue #1381, opened 2021-12-09
Update Erlang tests to use snapshots + have multiple directories


lpil:
Update the Erlang tests to the same standard as the JS tests.

1. Split big test functions into one per assertion
2. Split the big test module into a module per type of thing being tested
3. Use snapshots rather than hard-coding the expected Erlang.

This can be done in multiple PRs to make it easier.

michaeljones:
I can have a look at this. For some reason I sometimes like tasks like this. For a bit anyway :)

lpil:
Wonderful!

Don't feel it needs to be one massive PR. If you do just some of it that would be great still.

lpil:
Is there more you'd like to do on this @michaeljones ? I am super happy with what you've done

michaeljones:
Happy to have been of help. It depends on whether you're happy with the remaining tests in https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/erlang/tests.rs staying there? I've split out the more obviously group-able ones but it gets harder and harder to decide with the ones left. 

I guess I'm happy to stop there but you might want to take it further.

lpil:
I think snapshots are where the majority of the value come from, so I'm happy. Thank you!

PR #1391, opened 2021-12-17
Split out erlang record tests

michaeljones:
To a separate file, to separate test functions and to use insta snapshots rather than manually specified expectations.

For  #1381

Merge if you feel like it or I can keep adding to it for a bit. Depends on how you like to handle these things. I'd appreciate feedback when you have time so I don't go too far in the wrong direction.

These tests are different to the others that use the `assert_erl` macros. I don't know if you want them ported over (if that is even possible) or if you're happy to keep them separate with direct use of the insta::assert_snapshot approach.


michaeljones:
I'll leave it there for a moment in case I'm doing something wrong. I don't want to repeat a bad pattern a few dozen more times :)

lpil (approved):

PR #1396, opened 2021-12-19
feature(cli): New `clean` subcommand.

### LINKED ISSUE #1389: Gleam 'clean' command
It would be handy and nice to have a `gleam clean` command.

* Remove artefacts from the build directory that gleam has generated in the past.
* more fine-grained options may be useful.

For reference, see
- https://doc.rust-lang.org/cargo/commands/cargo-clean.html
- https://www.scala-sbt.org/1.x/docs/Running.html

vic:
`gleam clean` will simply remove the `build` directory.

Is there any other directory outside of `build` that we should also remove?

Closes #1389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- You can now remove build artifacts using the new `gleam clean` command.
 - The `compile-package` can now generate `package.app` files and compile source
   modules to `.beam` bytecode files.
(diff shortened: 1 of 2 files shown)


lpil (approved):
Brilliant, thank you.

Would you mind also updating the Changelog file?

vic:
@lpil done :)

lpil (approved):
Fab, thank you

PR #1401, opened 2021-12-19
Automatic changelog for each release.

vic:
Using [conventional commit messages](https://www.conventionalcommits.org/en/v1.0.0/) and
the [release-please-action](https://github.com/google-github-actions/release-please-action) action
will help us keep an updated Changelog with proper references to issues
and PRs.

The `release-please` action keeps an updated open PR whenever new
branches are merged into main. Once we merge each _release PR_ a new tag
will be generated.

For an example of conventional commits turned into a Changelog, see
[release-please-action/CHANGELOG.md](https://github.com/google-github-actions/release-please-action/blob/main/CHANGELOG.md)

### DIFF
--- a/.github/workflows/release-please.yaml
+++ b/.github/workflows/release-please.yaml
@@ -1,0 +1,16 @@
+name: release-please
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: GoogleCloudPlatform/release-please-action@v2
+        name: create release
+        with:
+          release-type: rust
+          bump-minor-pre-major: true 


vic:
I see we already run some actions on each release, @lpil this file https://github.com/gleam-lang/gleam/blob/main/.github/workflows/release.yaml#L60 just builds on release-mode and pushes executable artifacts to github right ? 

The `rust` release-please action configured on this PR would push crates or if we don't want that we could change the release type to `simple` for it to do nothing remote.

lpil:
Thanks for this but I much prefer to manually write the changelog. I find generated changelogs are much less clear to read than ones made by humans.

vic:
Nice, I'm closing this one.

PR #1402, opened 2021-12-19
Dont allow lowercase boolean variables in patterns

### LINKED ISSUE #1358: Pattern match using lowercase boolean value
This code compiles and prints "oops"

```gleam
  case 42 > 42 {
    true -> io.println("oops")
    false -> io.println("ok")
  }
```

The erlang compiler does warn about `Warning: variable 'True' is unused`.

I guess gleam would fail when it will check for exhaustiveness, but perhaps the compiler could also emit an user friendly hint or warning for such simple mis-use?

vic:
Inside patterns, when people use `true` and `false` variables
they probaly meant to use Gleam's `True` and `False`.

```gleam
case 42 > 42 {
     true -> "Nop"
     false -> "Yes"
}
```

The previous code generates an error like:

Screen Shot 2021-12-19 at 12 23 09

Closes #1358

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1698,6 +1698,10 @@ Try a different name for this module.",
                         "This is a reserved word.",
                         vec!["Hint: I was expecting to see a name here.".to_string(), "See: https://gleam.run/book/tour/reserved-words".to_string()]
                     ),
+                    ParseErrorType::LowcaseBooleanPattern => (
+                        "Did you want a Bool instead of a variable?",
+                        vec!["Hint: In Gleam boolean literals are True and False".to_string(), "See: https://gleam.run/book/tour/bools.html".to_string()]
+                    ),
                     ParseErrorType::UnexpectedToken { expected } => {
                         let mut messages = expected.clone();
                         if let Some(s) = messages.first_mut() {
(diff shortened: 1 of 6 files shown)


vic:
Looks like some tests actually used `true` and `false` in let patterns.

On this branch, those names could be used anywhere except for pattern variables,
`fn true()` is valid function name, and `fn foo(true: String)` would still be a valid function parameter name.


@lpil any pointers on how to proceed with this PR ?

compiler-core/src/error.rs
```diff
@@ -1698,6 +1698,10 @@ Try a different name for this module.",
                         "This is a reserved word.",
                         vec!["Hint: I was expecting to see a name here.".to_string(), "See: https://gleam.run/book/tour/reserved-words".to_string()]
                     ),
+                    ParseErrorType::LowcaseBooleanPattern => (
+                        "Looks like you wanted a Bool value instead of a variable here?.",
```
lpil, line 1702:
```suggestion
                        "Looks like you wanted a Bool value instead of a variable here?",
```

lpil (commented):
Hi @vic ! The code looks grand to me.

Perhaps we could rename those variables in the tests to `true_` and `false_`

vic:
@lpil thanks for your input :). I guess this PR is ready now.

lpil (approved):
Wonderful! Thanks again, really good stuff.

Issue #1277, opened 2021-10-01
Nightly `gleam` binary compilation


cognivore:
# Why?

Seemingly, early adopters want to use nightly exclusively because most them also contribute to the tech.
Having nightlies would benefit them, as well as augment CI experience.

# What?

 - [ ] Support all the currently supported release targets
 - [ ] Host nightlies somewhere
 - [ ] Host an archive of nightlies for quick experiments
 - [ ] Enable `nightly` and, perhaps, versions from the archive to be used from `setup-gleam` github hook

# How?

 - [ ] Learn how current release management is performed
 - [ ] Figure out Rust cross-compilation
 - [ ] Determine the level of scrutiny in artefact testing that is acceptable for nightlies
 - [ ] Crowd-source or otherwise obtain access to hardware, perhaps virtual, needed to build and host the artefacts

HarryET:
Hey, I'm happy to have a look at this if it's still somthing that Gleam needs/wants?

lpil:
Assigned you!

PR #1405, opened 2021-12-20
Add nix flake environment.

vic:
This branch adds a nix-flake for Gleam.

This will allow people using the [nix](https://nixos.org) package manager to build and run any commit of Gleam. Also allows contributors using nix to have a development environment with
all dependencies loaded and ready to `hack && cargo build`

See the guide on the Nix.md file.


vic:
A few years ago I did an [asdf-plugin for gleam](https://github.com/asdf-community/asdf-gleam). In hopes it helped more people to get a running gleam fast and try it out. 

Since about a year I've been using nix for all my dependency needs instead of asdf. So here's my contribution again in hopes that more people can try gleam right away.

For example, my PR branch is runnable like this:

Screen Shot 2021-12-20 at 16 21 12

vic:
Perhaps this could be the solution of #1277 for people using Nix.

lpil:
While I think this is a great thing to have, I'm afraid I can't accept this as I'm not able to maintain it as I don't know enough about Nix. This will need to be maintained by somebody else in a different repo, as with the asdf plugin.

vic:
I'll be moving this flake to a repo of mine.

Is it ok to mention how to use my external repo to get a gleam environment on this repo documentation? Maybe just a line or two so that people using nix can quickly get a full environment.

lpil:
Yes I think that's a really good idea! Let's get it in there.

Issue #1256, opened 2021-08-18
Generic type errors don't use the right name


UnsolvedCypher:
Currently, type errors which involve generics call the generics `a`, `b`, etc regardless of what the generic type is actually called by the user, which can result in confusing errors. For example:
```
pub fn multi_result(x: some_name) {
  io.print(x)
}
```

Results in
```
error: Type mismatch
  ┌─ /project/src/gleam_aoc.gleam:7:12
  │
7 │   io.print(x)
  │            ^

Expected type:

    String

Found type:

    a
```

I think it would be ideal to call the type `some_name` in the error message rather than just `a`.

lpil:
Could be a good improvement to the clarity of error messages. Thank you

vic:
I'd like to take this one.

lpil:
Thank you

PR #1407, opened 2021-12-21
Display user annotated names on unification type error

vic:


vic:
@lpil this is my *very work in progress* (aka, *I don't know what I'm doing*) try to show user annotated types on errors.

I have at least a couple of things I belive you can guide me (since I'm not sure it's the proper way to solve this issue)

1) Passing around the type annotation so it's available on the error formatting code.
   
   I added a lot of `println!` statements to get an idea of where things are being done. So far, I figured out where the unification error happens.
   I [added a test](https://github.com/gleam-lang/gleam/pull/1407/files#diff-3961af61fd82b1db19f362844082421752dfc50808eab9ef5e16317e4764661aR18) using the same code example as in #1256 

   But it seems the unification error is between the expected `String` type an Generic-type that comes from a phase named [preregistered fn types](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_.rs#L939)

   Is that correct? shouldn't the unification be between String and the user's annotated type `TypeAst` converted to a `Type` by means of `hydrator..type_from_ast` ?

2) In the mean time, trying to get the annotation to be available to the error value, I added an [optional annotation](https://github.com/gleam-lang/gleam/pull/1407/files#diff-6fd04d13c756a77c08670d59ef562ad305467a91b1c3cda5ccc06bd9a26b86ecR253) to `ValueConstructorVariant.LocalVariable`.

 Again I don't know if this is right, shall it be a `Type` hidrated from the `TypeAst` ?
 This caused to add a None annotation to all LocalVariables being created in other places. 

3)  Now that the annotation was available in the LocalVariable introduced into the environment, [make it available on the error value](https://github.com/gleam-lang/gleam/pull/1407/files#diff-892d21f7b2f39e0c5d4ef6a85f69d96b223f6cd4309106ac900cf9c391abb183R1716).

4) Once available in the error value, show it in the error message.

 Currently I'm just [printing the debug string](https://github.com/gleam-lang/gleam/pull/1407/files#diff-5f934eeb9c5fa397dee18043a9bfce13b4501938ad316bff5bc63954cfd8444eR1026) of the TypeAST, again I believe it should be a proper `Type` since I see functions for pretty printing types but not typeAst.

 

Any feedback and guidance on all of this is more than welcome :)

lpil:
Hi! Thanks for waiting, I've been busy with the the holidays.

>But it seems the unification error is between the expected String type an Generic-type that comes from a phase named preregistered fn types
>
>Is that correct? shouldn't the unification be between String and the user's annotated type TypeAst converted to a Type by means of hydrator..type_from_ast ?

This happens as we make 2 passes through a module to type check it. The first we record all the functions in the environment (along with their type definitions if annotations are present), and the second pass we perform the type inference. This is why we don't use the hydrator on that second pass, it was already used in the first.

If we did it in a single pass you wouldn't be able to refer to functions lower in the module.

> In the mean time, trying to get the annotation to be available to the error value, I added an optional annotation to ValueConstructorVariant.LocalVariable.

That makes sense to me. I don't think we'd want it to be a Type as there's no way to know the name of a type variable, and the same type variable may have different names in different functions.

I think it may not be suffient to store it on the local variable enum variant as it isn't possible to represent more complex types such as `List(something)`, and we would want to get the right type name in other scenarios such as this:

```rust
pub fn main(x: something) -> Int {
  let y = x
  y
}
```

I think we may need to store a map of type ids onto their annotated names. I think we will need to reset this map between each new function as different types can have the same name in each function.

It may be that one of the existing maps in the typing environment is suitable to storing this information.

vic:
@lpil Hope you are having very happy holidays!

Thanks for your input, it was really helpful. 

Today I spent most of the day following the codebase and seeing exactly what you said, the first-pass (`register_import`, `register_types` and `register_values`) so that code can refer to functions/consts/etc defined later on a module and where hydrators are used to keep track of the generic-types later used on unification (`infer_statement`). Now it's clear to me why it's done this way :).

I followed your advice and changed the hydrator `rigid_type_ids` to be a hashmap instead of just a set. Keeping the annotated name for the type var. Turns out all I needed to do after this was to make this `rigid_type_id -> annotated_name` map available to the error pretty printer so that it could lookup for the annotated names instead of generating fresh names for them. 

I added two more additional tests inspired by your comment above.

compiler-core/src/type_/expression.rs
```diff
@@ -1757,14 +1757,22 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
...
+                match return_type {
+                    Some(return_type) => self.unify(return_type, typed_body.type_()).map_err(|e| {
+                        e.return_annotation_mismatch()
+                            .into_error(typed_body.location())
+                    }),
+                    _ => Ok(()),
+                }
+                .map(|_| typed_body)
+            })
+            .map_err(|e| e.with_unify_error_rigid_names(body_rigid_names))?;
```
lpil, line 1775: I found this a bit hard to read, could we go back to using `if let Some`?

To make it easier the `in_new_scope` can be changed to return `Result<(_, _)>` rather than `(Result<_>, _)`, and then we can keep the `?` on the end of that line as before.

```rust
let body = body_typer.infer(body)?;
Ok((body_typer.hydrator.rigid_names(), body))
```

compiler-core/src/type_/hydrator.rs
```diff
@@ -76,7 +76,11 @@ impl Hydrator {
     /// an annotation. As such it should never be instantiated into an unbound
     /// variable.
     pub fn is_rigid(&self, id: &usize) -> bool {
-        self.rigid_type_ids.contains(id)
+        self.rigid_type_names.contains_key(id)
+    }
+
+    pub fn rigid_names(&self) -> im::HashMap<usize, String> {
+        self.rigid_type_names.clone()
```
lpil, line 83: I think we could take ownership here, save the clone (even if it is a cheap one)
```suggestion
    pub fn take_rigid_names(&mut self) -> im::HashMap<usize, String> {
        std::mem::take(&mut self.rigid_type_names)
```

lpil (commented):
This is really cool! It's great to see those error messages.

One this I was wondering, is it possible to get the type printer to show the same name for two types here?

```rust
pub main(x: a, y) {
  x == [y]
}
```

Here `x` would be reported as `a`, but I think we would also generated `a` as the parameter to the `List` type.  I think we'd need to record the already used names and then check that the name that comes from `next_letter` hasn't already been taken before we use it.

compiler-core/src/type_/expression.rs
vic: I'll try to refactor this code to make it more readable.

compiler-core/src/type_/hydrator.rs
lpil: Ah, I must have misread the code. I thought it was deallocated after that method was called, making it safe to remove that hashmap

lpil:
Hi @vic , are you still working on this? Thanks!

vic:
hey @lpil, sorry for the very late update. Just rebased and fixed the non-idiomatic rust, hope it's much more easy to read now. Sorry for the delay.

lpil (approved):
Fantastic work, thank you! I'm impressed by how clean and small your changes are here. I think this will be a big UX improvement.

\+ no worries for the delay, we're all busy :)

vic:
rebased :)

PR #1413, opened 2021-12-24
Implement WebAssembly Compiler

JohnDoneth:
This is the branch that is currently used by the [gleam-playground](https://johndoneth.github.io/gleam-playground/) project.

## Overview

- All of the IO is implemented for types that are used in-place of `ProjectIO` which allows the code to run in the browser. Eventually we could use HTTP in the browser to fetch dependencies which would allow anyone to use any public library they wanted in the browser.
- Because Rust doesn't seem to have a suitable bundler for JavaScript just yet, we pass all of the compiled files for JavaScript over the WebAssembly bridge with Serde where they can be bundled with Rollup. For Erlang, we pass all of the compiled Erlang files. In the case of an error, it is returned as a formatted string.


compiler-core/src/io.rs
```diff
@@ -56,6 +55,63 @@ pub struct OutputFile {
     pub path: PathBuf,
 }
 
+#[allow(missing_debug_implementations)]
+pub struct ReadDir {
+    entries: Box<dyn Iterator<Item = io::Result<DirEntry>>>,
+}
+
+impl ReadDir {
+    pub fn from_entries(entries: Vec<io::Result<DirEntry>>) -> Self {
```
lpil, line 64: Could this function take `Box<dyn Iterator<Item = io::Result<DirEntry>>>`? Seems we build vecs from iterators to turn them back to iterators, that seems like it could be avoided.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "compiler-wasm/static/build/packages/gleam_stdlib"]
+	path = compiler-wasm/static/build/packages/gleam_stdlib
+	url = git@github.com:gleam-lang/stdlib.git
```
lpil, line 3: In adding the stdlib as a dependency of the compiler we've now got a circular dependency cycle. The stdlib depends on the compiler, and the compiler depends upon the stdlib.
I don't think this is the ideal design, and it means we're no longer able to do breaking language changes as there's no stdlib version that will support the new compiler behaviour.

We would also need to update the release build scripts and the container image build scripts.

I think the best option is to have the wasm based compiler work the same as the build tools and have the stdlib be passed in as normal Gleam code.

compiler-wasm/static/build/packages/packages.toml
```diff
@@ -0,0 +1,2 @@
+[packages]
+gleam_stdlib = "0.18.0"
```
lpil, line 2: What's this file for?

compiler-wasm/src/lib.rs
```diff
@@ -1,6 +1,204 @@
...
+
+const PROJECT_NAME: &str = "gleam-wasm";
+
+/// Compile a set of `source_files` into a different set of source files for the
+/// `target` language.
+///
+/// The `main_source` file is required.
+/// The `additional_source_files` are optional.
+pub fn compile(
+    main_source: &str,
```
lpil, line 31: I think the idea of a main source module makes sense in the playground, but not in the compiler. I think we should include it like a normal module in the source files hashmap.

compiler-core/src/io.rs
JohnDoneth: I tried this and unfortunately it leads to lifetime issues with the closures involved with the boxed `dyn Iterator` which leads me to believe we need the allocation to avoid that issue. I was able to unbox the entries field and updated `ReadDir` to work with the `FromIterator` and `IntoIterator` traits which cleans up the code a bit... I think? 2ab5e588

compiler-wasm/static/build/packages/packages.toml
JohnDoneth: I can't recall why I created that file. Removed.

compiler-wasm/src/lib.rs
JohnDoneth: Done :+1:

.gitmodules
JohnDoneth: Done :+1:

lpil:
Is this ready to review @JohnDoneth ?

JohnDoneth:
Yes! @lpil

.gitmodules
michallepicki: @JohnDoneth So the git submodule and test-musl changes can be reverted now? Or is that still used?

JohnDoneth: Yes, thanks, I will revert those changes

michallepicki: Cool! I think it would be good to merge in main or rebase the brach as well since there's a conflict showing up (at least right now for me)

lpil: Is this still needed?

compiler-core/src/pretty.rs
```diff
@@ -155,22 +155,22 @@ fn fits(mut limit: isize, mut docs: im::Vector<&Document<'_>>) -> bool {
 
             Document::ForceBreak => return false,
 
-            Document::Nest(_, doc) => docs.push_front(&doc),
+            Document::Nest(_, doc) => docs.push_front(doc),
```
lpil, line 158: Why has this file changed?

lpil (commented):
Fantastic stuff!

Thank you for waiting, I wanted to get the release out before looking at adding anything new to the compiler.

Would you mind rebasing up to date with main? There's also some changes that I don't understand, I've left comments inline.

compiler-core/src/pretty.rs
JohnDoneth: These were all Clippy suggestions

lpil: Ah! Thank you

.github/workflows/ci.yaml
```diff
@@ -63,21 +63,52 @@ jobs:
 
   test-musl:
     runs-on: ubuntu-latest
-    container: clux/muslrust:stable
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
 
-      - name: Link to predefined musl toolchain
+      - name: Build binary
```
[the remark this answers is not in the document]
JohnDoneth: At one point I needed to change this to work with Docker directly for submodules to work, but since that is no longer necessary these changes can be reverted.

.gitmodules
JohnDoneth: Removed

lpil (approved):
Brilliant! Thank you so much. I'm excited to see where the playground project goes, I think it could be a major asset for Gleam

+ thanks for waiting for me, I have been slow with this PR.

PR #1415, opened 2021-12-24
Allow expression for record in record update syntax

### LINKED ISSUE #1411: Allow expressions for the record to update in record update syntax
Currently the record update syntax only allows an identifier for the record to update. It could be nice if we could use an expression like so:
```
Person(..another_record.person, is_happy: True)
```

J3RN:
Fixes #1411 

Overall the change is very small. The updated parser now expects the item after the spread operator to be any expression instead of only `Name`. This `base` expression had to be wrapped in a `Box` when put into the `RecordUpdateSpread` struct in order to prevent a cyclic dependency; let me know if there's a better way to do this :sweat_smile: 

Added a few spot-checking tests for both Erlang and JavaScript; specifically for the record being a function call and for the record being a field access (as referenced in #1411). Let me know if there's any more you'd like me to add!

### DIFF
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_update.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_update.snap
@@ -4,10 +4,10 @@ expression: "\npub type Person {\n  Person(name: String, age: Int)\n};\npub fn u
 
 ---
 error: Unknown variable
-  ┌─ /src/one/two.gleam:6:10
+  ┌─ /src/one/two.gleam:6:12
   │
 6 │   Person(..person)
-  │          ^^^^^^^^ did you mean `Person`?
+  │            ^^^^^^ did you mean `Person`?
 
 The name `person` is not in scope here.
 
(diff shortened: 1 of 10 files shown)


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_update.snap
```diff
@@ -4,10 +4,10 @@ expression: "\npub type Person {\n  Person(name: String, age: Int)\n};\npub fn u
 
 ---
 error: Unknown variable
-  ┌─ /src/one/two.gleam:6:10
+  ┌─ /src/one/two.gleam:6:12
   │
 6 │   Person(..person)
-  │          ^^^^^^^^ did you mean `Person`?
+  │            ^^^^^^ did you mean `Person`?
```
J3RN, line 10: These error indicators are now limited to the expression being expanded, which I personally consider a feature :grin:

lpil (approved):
Beautiful work! Thank you very much. Would you mind also updating the changelog?

J3RN:
Done!

lpil (approved):

PR #1419, opened 2021-12-29
Attempt to boot the gleam binary after building release

thepratt:
Aim is to round up and tick off some of the CI tagged tickets, with #769 being the first up.

TODO:
- [x] Validate 7zip commands w/ a windows prompt

PR #1420, opened 2021-12-29
Provide sha256 and sha512 checksums for releases

thepratt:
We may have an additional happy accident here @lpil from https://github.com/gleam-lang/gleam/pull/1419#issuecomment-1002348110

> RE release creation, I'm happy to have the draft be created even if it fails. I can delete the draft release no problem, and it's less work than fixing the build which I would have to do anyway.

To merge 3 artifact upload statements into 1, I changed the action. The new action supports glob updating but doesn't allow passing in a URI of an existing release. Releases with the same name(s) will be used if found - this needs verification.

/cc @jinyus

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        build: 
+        build:
           - macos
           - macos-arm64
           - windows-64bit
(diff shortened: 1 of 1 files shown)


thepratt:
https://github.com/softprops/action-gh-release/blob/8a65c813553f4d05769635eb1b70180d25b9b61b/README.md#L188
> When the release info keys (such as name, body, draft, prerelease, etc.) are not explicitly set and there is already an existing release for the tag, the release will retain its original info.

lpil (approved):
Oh that is really nice!

Have you run this job somewhere? I assume it passed

thepratt:
> Oh that is really nice!
> 
> Have you run this job somewhere? I assume it passed

Not the new action, but have ensured all the bash pieces work (at least on linux).

lpil:
OK let's merge it and we'll find out later :)

PR #1423, opened 2021-12-29
Switch remaining erlang tests to snapshots

michaeljones:
Still need to break them out to separate files but this step is a bit
more automatable and still gets us a win.

Further work on #1381


lpil (approved):
You absolute star, this is so much better and will save me a ton of time in future. Thank you!

PR #1427, opened 2022-01-01
Update `parse_custom_type` examples

J3RN:
These examples, e.g.
```gleam
type A(a) { A(String, a: b) }
```
seem to indicate that the type parameter `a` declares a field that must exist in all records/constructors. This may have been true at one point, but this is not how type parameters are used today. I replaced these examples with examples from [the "Generics" section of the Language Tour](https://gleam.run/book/tour/custom-types.html#generics).


lpil (approved):
Much better, thank you

Issue #1403, opened 2021-12-20
Confusing error message when try expression has incorrect error type compared to function annotation


silicium14:
This happens when the error type of an expression used in a `try` does not match the error type of the function's annotation. In this case the compiler produces a type mismatch error and the error is reported to be on the last expression of the function. This is confusing because the last expression can have no relation at all with the `try` expression. It would be more clear to the developer if the the compiler reported the error on the `try` statement.

Gleam version: 0.18.2
Example code:
```gleam
pub fn a_function() -> Result(Bool, Nil) {
  try value = Error("a String when the function annotation says Nil")
  // There could many more lines here
  Ok(True)
}
```

Compiler error:
```
error: Type mismatch
  ┌─ /.../example_test.gleam:3:3
  │
3 │   Ok(True)
  │   ^^^^^^^^

The type of this returned value doesn't match the return type 
annotation of this function.

Expected type:

    Result(Bool, Nil)

Found type:

    Result(Bool, String)
```

lpil:
A good suggestion, thank you.

We will need to think carefully about how we could improve this error message. The last try may be many lines back so it could be unclear to show all that code.

lpil:
Some thoughts from the discord channel

![image](https://user-images.githubusercontent.com/6134406/147857265-4e68e94b-4294-4441-8262-ba484d6e5a19.png)

silicium14:
The issue is when the types of the try expression and the return value are consistent with each other, making for a successfully unified Result type, but this unified Result type does not match the function return type annotation.

Example 1: the unified type is `Result(Bool, String)` but the annotation is `Result(Bool, Nil)`
```
pub fn a_function() -> Result(Bool, Nil) {
  try value = Error("a String when the function annotation says Nil")
  Ok(True)
}
```

Example 2: the unified type is `Result(a, String)` but the annotation is `Result(Nil, Nil)`
```
pub fn a_function() -> Result(Nil, Nil) {
  try _ = Ok("irrelevant")
  Error("a string")
}
```

We can see that the offending part can be the try expression, the return value, or both of them. Even when there is only one offender we can't tell which one it is after the unification of types into one Result type. Knowing this, when encountering this error we could show the entire code from the try expression. If this makes for many lines of code we could trim the content. Here is an example :
```
error: Type mismatch
   ┌─ ./src/example.gleam:59:11
   │
59 │   try _ = Ok("irrelevant")
   │           ^^^^^^^^^^^^^^^
[...]

PR #1433, opened 2022-01-02
Enhance type mismatch errors in the presence of try

silicium14:
Fixes issue https://github.com/gleam-lang/gleam/issues/1403

When a type mismatch occurs in the presence of a try, the whole code from the try expression to the block return value is reported. The display library already handles the trimming when there are many lines.

### DIFF
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__function_return_annotation_mismatch_with_try_nested.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__function_return_annotation_mismatch_with_try_nested.snap
@@ -1,0 +1,29 @@
+---
+source: compiler-core/src/type_/tests/errors.rs
+expression: "fn() -> Result(Nil, Nil) {\n          try _ = {\n            try _ = {\n                try _ = Error(1)\n                Ok(Nil)\n            }\n            Ok(Nil)\n          }\n          Ok(Nil)\n        }"
+
+---
+error: Type mismatch
+  ┌─ /src/one/two.gleam:2:11
+  │  
+2 │ ╭           try _ = {
+3 │ │             try _ = {
+4 │ │                 try _ = Error(1)
+5 │ │                 Ok(Nil)
+  · │
+8 │ │           }
+9 │ │           Ok(Nil)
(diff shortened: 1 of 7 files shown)


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__function_return_annotation_mismatch_with_try_nested.snap
```diff
@@ -0,0 +1,29 @@
...
+  ┌─ /src/one/two.gleam:4:25
+  │  
+4 │                   try _ = Error(1)
+  │ ╭─────────────────────────^
+5 │ │                 Ok(Nil)
+6 │ │             }
+7 │ │             Ok(Nil)
+8 │ │           }
+9 │ │           Ok(Nil)
+  │ ╰─────────────────^
```
lpil, line 16: I think it would be clearer to point to the start of the first try rather than the first value of the inner-most try. Could we do that instead?

lpil (commented):
Really nice work, the changes look very clean. I've left one small request inline

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__function_return_annotation_mismatch_with_try_nested.snap
silicium14: I remove the code to handle nested try and made it to point to the `try` keyword in all cases, is this ok ?

silicium14: I rebased `gleam-lang:main` and force pushed ee04f3a.

lpil (approved):
Brilliant work, thank you very much. This will be much clearer.

P.S. Thank you for waiting for me to get back to you on this, I wanted to get the release out the door and it took a little longer than expected.

Would you mind rebasing again? There is a merge conflict. Thank you 🙏

silicium14:
I rebased, fixed formatting and squashed the two commits into one.

No problem, it's great to have a new release out!

lpil (approved):
Thanks again! This'll help a lot with the user experience

Issue #1438, opened 2022-01-03
Binary checking doesn't work on Windows


lpil:
```
Run if [ "windows-latest" = "windows-latest" ]; then

7-Zip 21.06 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-11-24

Scanning the drive for archives:
1 file, 4420295 bytes (4317 KiB)

Extracting archive: gleam-v0.19.0-rc1-windows-64bit.zip
--
Path = gleam-v0.19.0-rc1-windows-64bit.zip
Type = zip
Physical Size = 4420295

Everything is Ok

Size:       12759040
Compressed: 4420295
D:\a\_temp\b924d4ad-2995-432f-a69e-d174ef93fbd5.sh: line 3: gleam.exe: command not found
```

@thepratt Do you understand what is happening here? I was unable to work this issue out.

thepratt:
The commented out correction you made is correct - we're running windows under bash and not powershell/cmd. Never really had to deploy anything to/from windows.
https://github.com/gleam-lang/gleam/blob/main/.github/workflows/release.yaml#L71-L73
![Screenshot from 2022-01-04 00-59-12](https://user-images.githubusercontent.com/3391281/147996084-ee562d60-97b9-484b-b1c7-7462715cc4ff.png)

PR #1441, opened 2022-01-04
Re-instate window-bash release command

thepratt:
Fixes 1438

See https://github.com/gleam-lang/gleam/issues/1438#issuecomment-1004457258


lpil (approved):
I was so close! Haha!

Thank you

PR #1444, opened 2022-01-04
Move building release and clippy out of test CI jobs to speed them up

### LINKED ISSUE #1443: Do not build with --release mode on CI in jobs that mainly run tests
While working on [adding rust-cache](https://github.com/gleam-lang/gleam/pull/1440) I noticed that the regular [ci workflow](https://github.com/gleam-lang/gleam/blob/main/.github/workflows/ci.yaml) is currently creating a release build, and then running tests.

So all code (and all deps as well!) are compiled twice: once in release cargo profile, and then in test profile (which is much faster, but still...). edit: Actually some deps three times, for the `clippy` linter in dev profile as well.

From what I can tell the Gleam build process is using cargo defaults, so performing a release build vs building the package for tests only differs in profile flags, like optimizations, am I right? Or is `cargo build` "testing" the creation of the executable more than just running `cargo test`?

The release build is mainly useful for the uploaded linux artifact. It could be a separate job parallel to running tests.

Alternatively I think one should be able to run `cargo build --profile test`

michallepicki:
Closes #1443

~Will need a small change to add rust-cache for that new job if #1440 gets merged~   (done)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,12 +42,6 @@ jobs:
         with:
           key: ${{ matrix.target }}
 
-      - name: Run linter
-        uses: actions-rs/cargo@v1
-        with:
-          command: clippy
-          args: --workspace --target ${{ matrix.target }} -- -D warnings
-
       - name: Run tests
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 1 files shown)


michallepicki:
`lint-build` took ~6 minutes now but thanks to cache will be faster next time.

It might need to be configured to be a required job to pass for the repo auto-merge PR stuff

lpil (approved):
Very nice, thank you

Issue #1445, opened 2022-01-04
Validate licences are valid SPDX


lpil:
Part of: https://github.com/erlef/build-and-packaging-wg/issues/34

Error if `gleam.toml` has a non-SPDX licence in it.

Accept non-SPDX licences from Hex as existing Hex packages may not have compliant licence configuration.

J3RN:
I'll take a look at this. My biggest question is whether there's a cleverer way to do this other than simply making a list of all valid license IDs and checking membership. Then again, having a file/database/whatever in this repository seems heavy-handed and calling out to an API or something doubly so.

PR #1450, opened 2022-01-07
Verify that package licenses are valid SPDX expressions

J3RN:
Fixes: #1445 

### To-do
- [ ] Testing. ~~I looked at `package_compilation_tests.rs`, but the relevant macros that we have in there to do boilerplate (`assert_erlang_compile`, `assert_javascript_compile`, etc) don't provide a facility to set licenses. We could add an additional argument, but it wouldn't be relevant to the vast majority of tests in the file. cc @lpil Any guidance here is appreciated~~ I found `make_config()` under `package_compilation_test()` and I think that's what I need.

### Notes

- Gleam's config format supports a _list_ of licenses, which seems ambiguous (are they "or"ed or "and"ed?). [SPDX expressions](https://spdx.github.io/spdx-spec/SPDX-license-expressions/) provide an unambiguous way to specify multiple licenses with explicit `AND`s and `OR`s; we may want to move from our list format to a simple string field with a SPDX expression going forward.

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -130,10 +130,7 @@ pub enum Error {
     InvalidVersionFormat { input: String, error: String },
 
     #[error("invalid license")]
-    InvalidLicense {
-        license: String,
-        reason: spdx::error::Reason,
-    },
+    InvalidLicense { license: String },
 
     #[error("project root already exists")]
     ProjectRootAlreadyExist { path: String },
@@ -400,33 +397,12 @@ numbers and underscores.",
                 write_project(buf, diagnostic);
             }
 
-            Error::InvalidLicense { license, reason } => {
+            Error::InvalidLicense { license } => {
                 let diagnostic = ProjectErrorDiagnostic {
                     title: "Invalid license".to_string(),
                     label: format!(
-                        "The license \"{}\" {}.",
+                        "The license \"{}\" is not a valid SPDX License ID.",
                         license,
-                        match reason {
-                            spdx::error::Reason::UnknownLicense => "is not a SPDX license".to_string(),
-                            spdx::error::Reason::UnknownException => "contains non-SPDX exception".to_string(),
-                            spdx::error::Reason::InvalidCharacters => "contains invalid characters".to_string(),
-                            spdx::error::Reason::UnclosedParens =>
-                                "contains an opening parenthesis without a closing parenthesis".to_string(),
-                            spdx::error::Reason::UnopenedParens =>
-                                "contains a closing parenthesis without an opening parenthesis".to_string(),
-                            spdx::error::Reason::Empty =>
-                                "contains no terms that are valid in SPDX expressions".to_string(),
-                            spdx::error::Reason::Unexpected(expected) => match expected {
-                                [] => "contains an unexpected term".to_string(),
-                                expected => "contains an unexpected term, expected \"".to_string() + &expected.join("\" or \"") + "\""
-                            },
-                            spdx::error::Reason::SeparatedPlus =>
-                                "contains a '+' after whitespace, which is not allowed in SPDX expressions".to_string(),
-                            spdx::error::Reason::UnknownTerm =>
-                                "contains a term that is not valid in SPDX expresssions".to_string(),
-                            spdx::error::Reason::GnuNoPlus =>
-                                "contains a '+' after a GNU license. Please use '-or-later' instead.".to_string(),
-                        }
                     ),
                 };
                 write_project(buf, diagnostic);
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -401,39 +401,31 @@ numbers and underscores.",
             }
 
             Error::InvalidLicense { license, reason } => {
-                let msg;
                 let diagnostic = ProjectErrorDiagnostic {
                     title: "Invalid license".to_string(),
                     label: format!(
                         "The license \"{}\" {}.",
                         license,
                         match reason {
-                            spdx::error::Reason::UnknownLicense => "is not a SPDX license",
-                            spdx::error::Reason::UnknownException => "contains non-SPDX exception",
-                            spdx::error::Reason::InvalidCharacters => "contains invalid characters",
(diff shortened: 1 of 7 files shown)


compiler-core/src/error.rs
```diff
@@ -394,6 +400,46 @@ numbers and underscores.",
                 write_project(buf, diagnostic);
             }
 
+            Error::InvalidLicense { license, reason } => {
+                let msg;
```
J3RN, line 404: Alright, I'll be the first to admit that `msg` is a hack. Essentially on lines 423 and 427, we create new `String`s. These are scoped only to their little block in the match arm, which means that they promptly go out of scope before having an opportunity to do anything useful. What I needed was for these values to live long enough to be able to be passed to `format!`, hence `msg`.

If there's a better way to do this (I'll be there is!), please tell me.

compiler-core/src/error.rs
```diff
@@ -394,6 +400,46 @@ numbers and underscores.",
...
+                            spdx::error::Reason::UnclosedParens =>
+                                "contains an opening parenthesis without a closing parenthesis",
+                            spdx::error::Reason::UnopenedParens =>
+                                "contains a closing parenthesis without an opening parenthesis",
+                            spdx::error::Reason::Empty =>
+                                "contains no terms that are valid in SPDX expressions",
+                            spdx::error::Reason::Unexpected(expected) => match expected.len() {
+                                0 => "contains an unexpected term",
+                                1 => {
+                                    msg = String::from("contains an unexpected term, expected ") + expected.first().unwrap();
```
michaeljones, line 423: I think it is ideal to avoid unwrap if possible? Maybe `expected.first().unwrap_or("[unknown]")` or some similar default value that we can probably rely on never return. Or I guess (though I've not done much of it rust) is that you can match on `expected` as a slice rather than matching on its length and then you can match against an empty slice and a single element slice and then `_` so that in the single element slice situation you have direct access to the element itself. A bit like the first example on this page: https://adventures.michaelfbryan.com/posts/daily/slice-patterns/

michaeljones (commented):
I'm no one in particular but here are some thoughts :)

compiler-core/src/error.rs
J3RN: Ah, right, I originally did the `unwrap` because I thought `expected` was a `Vec` and (IIRC) you can't pattern-match on `Vec`s in a `match`. However I was wrong; as you note, it is a slice.

compiler-core/src/error.rs
J3RN: You're right, I think I was trying too hard to be performance-sensitive in a place where it doesn't really matter.

So far I've opted for `+` (really just an alias for `push_str`) in the match arms, but if people want a `format!` in there, I'm happy to change it :grin: Heck, this whole function could be done with a `fmt::Formatter`, which I considered.

J3RN:
@michaeljones Thanks for the review! I'm appreciative for an eyes on my newbie Rust code!

compiler-core/src/error.rs
michaeljones: Oh, cool. I didn't realise `+` existed for strings. I've been format!-ing everywhere :) (And some `push_str`) Good to know!

lpil:
Looks good! Very clean.

RE the list of liences, I'm not 100% sure how it works, I'd need to check the Hex docs. Originally I had only 1 but they permit multiple, so I've gone with that.

J3RN:
> I'd need to check the Hex docs

What I found with a quick glance was:
> A list of licenses the project is licensed under. This attribute is required. It is recommended to use SPDX License identifier.

[Source](https://hex.pm/docs/publish)

This makes me think "AND", but :shrug:. Might be best for me to go chat with the Hex.pm team about possibly moving to an SPDX expression.

---

> This would work by changing the type for a licence to some wrapper
>
> ```
> pub struct SpdxLicence(pub String)
> ```
>
> And then implementing Serde::Deserialize for it in a way that calls the validation function you've created there.

This seems fine. I think instead of `SpdxLicense` we'd want something named `SpdxExpression` or similar to convey that it handles `and`s, `or`s, and `with <exception>`s.

lpil:
Sounds good to me

J3RN:
Yes, I think I can use [`spdx::license_id`](https://docs.rs/spdx/0.8.1/spdx/fn.license_id.html) to verify that the specified licenses are known SPDX License IDs. I'm resuming work on this today.

J3RN:
> And then implementing Serde::Deserialize for it in a way that calls the validation function you've created there

This is done, and pushed. Sadly it makes the error a bit more verbose, though:
```
error: File IO failure

    gleam.toml

    Foobar is not a valid SPDX License ID for key `licences` at line 15 column 1
```

compiler-core/src/error.rs
```diff
@@ -129,6 +129,9 @@ pub enum Error {
     #[error("{input} is not a valid version. {error}")]
     InvalidVersionFormat { input: String, error: String },
 
+    #[error("invalid license")]
+    InvalidLicense { license: String },
```
lpil, line 133: Is this still used?

lpil (commented):
Looks great! Thank you

Last question, was there an answer from the Hex team on how a non-SPDX licence is expressed? I can't find the issue now.

compiler-core/src/error.rs
J3RN: Good call! This was vestigial.

J3RN:
> Last question, was there an answer from the Hex team on how a non-SPDX license is expressed? I can't find the issue now.

The issue of non-SPDX licenses was not raised in [the ambiguity issue with the Hex.pm team](https://github.com/hexpm/hexpm/issues/1108). I could raise it, but the specter of non-SPDX licenses being unsupported was raised in the [ERLEF Build and Packaging WG minutes from Jan 04](https://github.com/erlef/build-and-packaging-wg/blob/master/minutes/2022-01-04.md):
> Supports common ones and even rare ones, but woN’t (sic) work well for people creating their own licenses

Should probably say "people _or organizations_ creating their own licenses", as I've seen more of the latter than the former :smirk:

PR #1454, opened 2022-01-09
Clean up cargo usage in GitHub actions

michallepicki:
- Use actions-rs/cargo where possible
- Simplify build target paths where possible (I couldn't figure that part out for musl builds)
- Force colored cargo output
- Disable debug symbols for faster dev and test builds (and smaller caches)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,10 +4,16 @@ on:
   push:
     branches:
       - main
+
+env:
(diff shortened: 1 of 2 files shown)


michallepicki:
(I hoped I could figure out the release cache not working but nope, giving up on that one. Doesn't matter much anyway.)

lpil (approved):
Brilliant, thank you

PR #1456, opened 2022-01-11
Add link to hexdocs to gleam new's README.md

michallepicki:


lpil:
Would it be good to add some badges?

michallepicki:
> Would it be good to add some badges?

Looked around and most people use https://shields.io/ ? e.g.

[Module Version](https://hex.pm/packages/propcheck)
[Hex Docs](https://hexdocs.pm/propcheck/)
[Total Download](https://hex.pm/packages/propcheck)

Seems like the module one and total downloads would get red if package is not on hex. The hex docs one is a static, can change color to a different one.

michallepicki:
I think I would add the hex version and hex docs ones, maybe change hex docs color to pink, or maybe add both with pink (can override the color for dynamic ones as well) and don't add the downloads one for now

lpil:
Sounds great to me!

PR #1462, opened 2022-01-15
Add limited exhaustiveness checking for custom types at top level of pattern matches

michallepicki:
Relates to #34

- Add types_constructors mapping to Module
- Add module_types_constructors to Environment
- register declared custom types to constructor names mappings
- error if not all constructors are matched in let expression
- error if not all constructors are matched ~in any of case expression subjects~ when there's only one case expression subject
- encode / decode new module information with Capn' Proto

I'm just learning Rust, I'll appreciate any help with cleaning up the implementation! ❤️


compiler-core/src/type_/environment.rs
```diff
@@ -241,6 +245,45 @@ impl<'a, 'b> Environment<'a, 'b> {
         }
     }
 
+    /// Lookup constructors for type in the current scope.
+    ///
+    pub fn get_constructors_for_type(
+        &self,
+        module_alias: &Option<String>,
+        name: &str,
+    ) -> Result<&Vec<String>, UnknownTypeConstructorError> {
```
michallepicki, line 254: @lpil Can I make this assume success? (I think this will be used always after unifying types and confirming the type is there)

compiler-core/src/type_/expression.rs
```diff
@@ -611,6 +611,39 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
...
+                    // println!("pattern_type: {:?}", pattern_type);
+                    // println!("m: {:?}", m);
+                    // println!("type_name: {:?}", type_name);
+                    if let Ok(constructors) =
+                        self.environment.get_constructors_for_type(m, type_name)
+                    {
+                        // println!("constructors: {:?}", constructors);
+                        let constructors_n = constructors.len();
+                        if constructors_n != 1 {
+                            return Err(Error::IncorrectNumClausePatterns { // TODO_EXH_CHECK add and return a new kind of error here
```
michallepicki, line 636: @lpil I chose this error as a placeholder for now but I'm curious, when is this one returned? (I didn't find any tests that mention it. Actually I didn't find any tests that check that errors get correctly returned...)

michallepicki: Found it, it's for scenarios like this one:
```
case 1, 2 {
  1 -> Nil
}
```

michallepicki: And I found a snapshot test at
`compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__wrong_number_of_subjects.snap`

compiler-core/src/type_/expression.rs
```diff
@@ -704,6 +712,21 @@ impl<'a, 'b, 'c> ExprTyper<'a, 'b, 'c> {
...
+            for c in &typed_clauses {
+                patterns.push(c.pattern[i].clone());
+                for a in &c.alternative_patterns {
+                    patterns.push(a[i].clone());
+                }
+            }
+            if !self.environment.exhaustive(patterns, t) {
+                return Err(Error::NotExhaustivePatternMatch { location });
+            }
+        }
```
michallepicki, line 728: Just realized this doesn't fully work, e.g. doesn't warn for situations like
```
  case b1, b2 {
    True, False -> Nil
    False, True -> Nil
  }
```

I could do some combinatorics to check if all possibilities are covered but maybe for this simple limited solution only do the check if there's one subject? 🤔

michallepicki: Re-watched SPJ's talk linked in the ticket, they store a "refinement types" style (set theoretic) representation of possible values, and when iterating through clauses they refine/narrow down the sets by what matches given clause. That's pretty wild.

lpil: I'd like to implement [this algorithm](https://github.com/gleam-lang/gleam/issues/34#issuecomment-896821619) later which will handle this. The Rust reference should help.

For now it's OK to have an imperfect solution.

compiler-core/src/error.rs
```diff
@@ -1604,6 +1604,23 @@ Try a different name for this module.",
...
+                        label: "".to_string(),
+                        file: path.to_str().unwrap().to_string(),
+                        src: src.to_string(),
+                        location: *location,
+                    };
+                    write(buf, diagnostic, Severity::Error);
+                    wrap_writeln!(
+                        buf,
+                        "This expression does not cover all possibilities.
+Each constructor must have a pattern that can match.",
```
lpil, line 1620: Can we list the unmatches variants here?

compiler-core/src/type_/environment.rs
lpil: I think this function should return a result because the Environment doesn't know anything about when it is being used. We would assert value at present at a higher level of abstraction in the typer.

compiler-core/src/type_/environment.rs
```diff
@@ -583,4 +632,60 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+        // println!("value_typ: {:?}", value_typ);
+        match &*value_typ {
+            Type::App {
+                name: type_name,
+                module: module_vec,
+                ..
+            } => {
+                // TODO_EXH_CHECK this is bad.
+                // But I would prefer to not need to keep track of imported type names to imported constructor names.
+                // (because the imports can be named and usage can be a mix of qualified and imported constructor and type names)
```
lpil, line 651: What does this mean?

compiler-core/src/type_/environment.rs
```diff
@@ -583,4 +632,60 @@ impl<'a, 'b> Environment<'a, 'b> {
...
+                            if covers_constructor(p, c) {
+                                continue 'constructors_loop;
+                            }
+                        }
+                        return false;
+                    }
+                }
+                true
+            }
+            // For now only type constructors are checked, and only their
```
lpil, line 687: Their what? :)

lpil (commented):
Thanks for taking this on, very cool stuff!

I'm finding the code a bit hard to follow as this is a complex problem. Could we make liberal use of comments to indicate intent? I've found err-ing of the side of stating the obvious can be helpful in the type checker as it's something we're generally not so familiar with.

compiler-core/src/type_/expression.rs
michallepicki: I looked at the linked algorithm, which confirmed my suspicions that what I did here for multiple subjects is not that useful. I'll replace this code with a simpler check that takes effect only when there's one subject. I think it'll be less confusing to the user

compiler-core/src/error.rs
michallepicki: I can try!

compiler-core/src/type_/environment.rs
michallepicki: Right, I didn't word this clearly. When implementing this I needed to add a mapping from type to a list of its constructor names to the environment. I observed that the Environment and existing get_type / get_value functions work for local (potentially imported+aliased) names. But to keep things simple, and because in typed expressions types are referred to with original name, I decided not to import this mapping when importing a module. So this mapping is kept in its original module, and I need to reconstruct that module's name to a format that can be found in "importable_modules" in the environment (this is the "bad feeling" part, I didn't like that I needed to do this).

Maybe when printing better error messages, to print the patterns that user missed, this will need to be revisited and reconsidered.

compiler-core/src/error.rs
michallepicki: I think I made it work. it refers to original constructor names even if user aliases them by importing with `as`, but that's probably good for now

michallepicki:
Next I'll work on adding some snapshot tests

michallepicki:
@lpil I implemented capnproto encoding & decoding (hopefully correctly). I added snapshot tests and some comments. I think this is ready for another look!

compiler-core/src/type_/environment.rs
```diff
@@ -587,4 +636,70 @@ impl<'a> Environment<'a> {
...
     }
+
+    /// Checks that the given patterns are exhaustive for given type.
+    /// Currently only performs exhaustiveness checking for custom types,
+    /// only at the top level (without recursing into constructor arguments).
+    pub fn exhaustive(
+        &mut self,
+        patterns: Vec<Pattern<PatternConstructor, Arc<Type>>>,
+        value_typ: Arc<Type>,
+    ) -> Option<Vec<String>> {
```
lpil, line 647: I think the name doesn't line up quite with the return value here. Either we could call it `unmatched_constructors` and keep the same return, or we could call it `exhaustive` or `check_exhaustiveness` and return `Result<(), Vec<String>>`.

 I think the altter latter would probably be more in keeping with the existing codebase. Makes it clearer it's the sad path.

compiler-core/src/type_/environment.rs
```diff
@@ -587,4 +636,70 @@ impl<'a> Environment<'a> {
...
+                name: type_name,
+                module: module_vec,
+                ..
+            } => {
+                let m = if module_vec.is_empty() || module_vec == self.current_module {
+                    None
+                } else {
+                    Some(module_vec.join("/"))
+                };
+                let mut matched_constructors = Vec::new();
```
lpil, line 659: Let's use HashSets for the matched and unmatched constructors? They're unique so it fits a little better, and it means we can avoid the iteration through the matched constructors below when determining which are unmatched.

We could either initially construct the set with the constructors and remove them as we see them, or we could do something similar to the current approach and take the difference of the two sets.

compiler-core/src/type_/expression.rs
```diff
@@ -707,6 +722,40 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+        // Because exhaustiveness checking in presence of multiple subjects is similar
+        // to full exhaustiveness checking of tuples or other nested record patterns,
+        // and we currently only do only limited exhaustiveness checking of custom types
+        // at the top level of patterns, only consider case expressions with one subject.
+        if subjects_count == 1 {
+            if let Some(subject_type) = subject_types.get(0) {
+                let value_typ = collapse_links(subject_type.clone());
+                // Currently guards are not considered in exhaustiveness checking,
+                // so we go through all clauses and pluck out only the patterns.
```
lpil, line 734: We want to consider any pattern with a guard to not match anything as the guard could evaluate to false.

compiler-core/src/type_/expression.rs
```diff
@@ -707,6 +722,40 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 .map_err(|e| e.case_clause_mismatch().into_error(typed_clause.location()))?;
             typed_clauses.push(typed_clause);
         }
+
+        // Because exhaustiveness checking in presence of multiple subjects is similar
+        // to full exhaustiveness checking of tuples or other nested record patterns,
+        // and we currently only do only limited exhaustiveness checking of custom types
+        // at the top level of patterns, only consider case expressions with one subject.
+        if subjects_count == 1 {
```
lpil, line 730: Would it be clearer if this code is extracted into a named function?

Could then convert this check + nesting into a guard then.

```rust
if subjects_count != 1 {
  return Ok(());
}

// rest here...
```

lpil (commented):
Looking really good! I've left some little Rust notes but I'm happy with general design here.

compiler-core/src/type_/environment.rs
michallepicki: Updated!

compiler-core/src/type_/environment.rs
michallepicki: Done, I am not fluent with iterators yet (`.iter()`,`.into_iter()`, `.cloned()` etc) so this part may be worth double-checking.

Also because the HashSet doesn't keep a stable order, I had to do `.sorted()` when converting to a vec when returning unmatched constructors, to make the tests pass reliably

compiler-core/src/type_/expression.rs
michallepicki: Updated

compiler-core/src/type_/expression.rs
michallepicki: I think this is a similar situation to nested pattern matches (when there are patterns present for constructor's arguments). Currently this PR doesn't take them into account and the exhaustiveness check is "happy" when a constructor is at least mentioned. So the check defaults to passing.

You're saying I should do the opposite for guards, and default to not passing? And keep the behavior in presence of nested patterns as it is now?

michallepicki: Here's an example to show why I think this is similar and I would expect both scenarios to behave similarly, even now without full exhaustiveness checking implemented:
```
pub type MyType {
  BoolBox(Bool)
  OtherConstructor
}

pub fn main1(v) {
  case v {
    BoolBox(b) if b == True -> v
    OtherConstructor -> v
  }
}

pub fn main2(v) {
  // vs
  case v {
    BoolBox(True) -> v
    OtherConstructor -> v
  }
}
```

michallepicki: With the previous example forcing the user to add a fallback clause seems natural.
Here's another example to consider:
```
pub type Inner {
  Inner
}

pub type MyType {
  Box(Inner)
  OtherConstructor
}

pub fn main1(v) {
  case v {
    Box(i) if i == Inner -> v
    OtherConstructor -> v
  }
}

// vs
pub fn main2(v) {
  case v {
    Box(Inner) -> v
    OtherConstructor -> v
  }
}
```

lpil: I think any time we see a guard we should consider it unmatched. Ideally we would do the same if we see any non-variable record field but perhaps that is too much complexity to add here.

michallepicki: I'm not super happy about it and I would default to passing the check, instead of nagging the programmer to add fallback clauses when compiler isn't smart enough to see that all possibilities are covered... But maybe if this is supposed to be a check that a programmer can rely on, it will be better in the end, when a proper algorithm is implemented.

I think I can probably make it work the way you want it with a minimal change: when preparing a flattened list of patterns from Case, do not even include patterns from branches that have non empty guards.

michallepicki: Updated!

lpil: I think we should see this as a subset of the future behaviour, so we will want to be consistent with the future handling of guards and with other ML languages. There's no way to know if a guard will be true or not, so we consider the clause not to match anything.

compiler-core/src/type_/environment.rs
```diff
@@ -587,4 +636,64 @@ impl<'a> Environment<'a> {
...
+                        } = pattern
+                        {
+                            pattern = assign_pattern;
+                        }
+
+                        match pattern {
+                            // If the pattern is a Discard or Var, all constructors are covered by it
+                            Pattern::Discard { .. } => return Ok(()),
+                            Pattern::Var { .. } => return Ok(()),
+                            // If the pattern is a constructor, add it to list of matched patterns
```
michallepicki, line 679:
```suggestion
                            // If the pattern is a constructor, remove it from unmatched patterns
```

compiler-core/src/type_/expression.rs
michallepicki: Alright! This is in https://github.com/gleam-lang/gleam/pull/1462/commits/9a40cdd45bd1a5424aae071a5b0472f690419103

compiler-core/src/type_/expression.rs
michallepicki: This was done in https://github.com/gleam-lang/gleam/pull/1462/commits/fe9534deaf0f186ecc27e13223be62a7f939be80

compiler-core/src/type_/environment.rs
michallepicki: This was done in https://github.com/gleam-lang/gleam/pull/1462/commits/c85818160140f801824ce5b1ff7b71468562c0b0

compiler-core/src/type_/environment.rs
michallepicki: This was done in https://github.com/gleam-lang/gleam/pull/1462/commits/f931228a71f2f517a836f33b1a55b21a5c6d2938

lpil (approved):
I've rebased these commits onto main, thank you so much!

michallepicki:
So you prefer to have a linear history of commits on main without any merge commits? I'll try to keep that in mind!

Issue #1434, opened 2022-01-02
Better type mismatch error for an inconsistent try return value


silicium14:
In the following code, the return value `1` is not consistent with the try because it should be of type Result with an error type matching the one of the try value.
```
fn a_function() {
  try _ = Error(Nil)
  1
}
```
This produces the following error:
```
error: Type mismatch
   ┌─ ./src/example.gleam:4:3
   │
10 │   1
   │   ^

Expected type:

    Result(a, Nil)

Found type:

    Int

```
We could enhance this error to explain that the mismatch is caused by the use of an earlier try.
- When the return value is not a Result:
```
error: Type mismatch
  ┌─ ./src/example.gleam:4:3
  │
4 │   1
  │   ^

This returned value has a type incompatible with the previous try expression.
The returned value after a try must be of type Result with the same error type as the try expression.


Expected type:

    Result(a, Nil)

Found type:

    Int

```
- When the return value is a Result with the wrong error type:
```
fn another_function() -> Result(Nil, Int) {
  Ok(Nil)
}
fn a_function() {
  try _ = Error(Nil)
  another_function()
}
```
```
error: Type mismatch
   ┌─ ./src/example.gleam:13:3
   │
13 │   another_function()
   │   ^^^^^^^^^^^^^^^^^^

This returned value has a type incompatible with the previous try expression.
The returned value after a try must be of type Result with the same error type as the try expression.


Expected type:

    Result(Nil, Nil)

Found type:

    Result(Nil, Int)

```

lpil:
Thank you

silicium14:
I think it would be nice to also explain that all the try expressions in a block must have the same error type (you suggested this on the discord channel). This would be helpful in the following case:
```
pub fn a_function() {
  try _ = Error(0)
  try _ = Error("string")
  try _ = Error(False)
  Ok(Nil)
}
```
Adding a sentence to cover this would result in the following message:
```
error: Type mismatch
   ┌─ ./src/inconsistent_try.gleam:33:3
   │
33 │ ╭   try _ = Error(False)
34 │ │   Ok(Nil)
   │ ╰─────────^

This returned value has a type incompatible with the previous try expression.
The returned value after a try must be of type Result. All the try expressions
in a block and the final result value must have the same error type.

Expected type:

    Result(Nil, String)

Found type:

    Result(Nil, Bool)

```
Is this acceptable?

lpil:
[...]

PR #1463, opened 2022-01-15
Enhance type mismatch error message for an inconsistent try

silicium14:
Fixes https://github.com/gleam-lang/gleam/issues/1434


lpil (approved):
This is really nice! I think it will make it much easier to understand what is happening. Thank you

PR #1471, opened 2022-01-19
Show whole pipeline in return type mismatch errors

### LINKED ISSUE #1378: Pipeline error message showing incorrect code
```rust
import gleam/string

fn get_data() -> String {
  ""
  |> string.split(on: ",")
}
```
```
error: Type mismatch
  ┌─ src/thingy.gleam:4:3
  │
4 │   ""
  │   ^^

annotation of this function.

Expected type:

    String

Found type:

    List(String)
```

michallepicki:
Closes #1378

### DIFF
--- a/compiler-core/src/type_/pipe.rs
+++ b/compiler-core/src/type_/pipe.rs
@@ -18,17 +18,24 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
         expressions: Vec1<UntypedExpr>,
     ) -> Result<TypedExpr, Error> {
         let size = expressions.len();
-        let mut expressions = expressions.into_iter();
-        let first = expr_typer.infer(
+        let end = expressions
+            .iter()
+            .last()
             // The vec is non-empty, this indexing can never fail
-            expressions.next().expect("Empty pipeline in typer"),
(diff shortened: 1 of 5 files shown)


compiler-core/src/type_/pipe.rs
```diff
@@ -18,17 +18,24 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
         expressions: Vec1<UntypedExpr>,
     ) -> Result<TypedExpr, Error> {
         let size = expressions.len();
-        let mut expressions = expressions.into_iter();
-        let first = expr_typer.infer(
+        let end = expressions
+            .iter()
+            .last()
```
lpil, line 23: Tiny tiny nit but I think we can call the `.last()` methon on a slice to avoid iterating over the prior expressions.

https://doc.rust-lang.org/std/primitive.slice.html#method.last

lpil (approved):
Thank you

compiler-core/src/type_/pipe.rs
michallepicki: Updated!

lpil:
💜

PR #1476, opened 2022-01-26
Update thread_local to fix cargo deny check

michallepicki:
It failed for me in https://github.com/gleam-lang/gleam/runs/4949255483?check_suite_focus=true , I think it would fail on current `main` or any new branch as well.


lpil (approved):
Thank you

PR #1499, opened 2022-02-23
feat: initialize git repo when creating a new project

wesleimp:


michallepicki:
@wesleimp Does it succeed and just skip the step if there's no git installed in the system?

wesleimp:
> @wesleimp Does it succeed and just skip the step if there's no git installed in the system?

Yes, it does.

lpil (commented):
Great stuff!

Looking at the `git2` project it depends on the libgit C library. Could we instead use the `git` command line? We could check to see if git is installed first and do nothing if it is not.

There's already code for running shell commands so we can reused that.

lpil (commented):
Looking great! Would you mind updating the CHANGELOG file with a `## Unreleased` section at the top?

Thank you

lpil (approved):
Fantastic, thank you!

lpil:
I've rebased these changes onto the main branch, thank you

PR #1522, opened 2022-02-27
fix: make `gleam add` support multiple arguments

megapctr:
Please note I am not very familiar with rust, let me know if this change could be improved.

Enabling "Hide whitespace" should make this change more readable, since a lot of the code was just moved one indent deeper to put it in a loop

### DIFF
--- a/compiler-cli/src/add.rs
+++ b/compiler-cli/src/add.rs
@@ -7,26 +7,12 @@ use gleam_core::{
 
 use crate::{cli, fs};
 
-pub fn command(to_add: String, dev: bool) -> Result<()> {
-    // Insert the new package into the manifest and perform dependency
-    // resolution to determine a suitable version
-    let manifest = crate::dependencies::download(Some((&to_add, dev)))?;
+pub fn command(packages: Vec<String>, dev: bool) -> Result<()> {
+    // Insert the new packages into the manifest and perform dependency
+    // resolution to determine suitable versions
+    let manifest = crate::dependencies::download(Some((packages.to_vec(), dev)))?;
 
-    // Pull the selected version out of the new manifest so we know what it is
-    let version = manifest
-        .packages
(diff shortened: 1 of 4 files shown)


compiler-cli/src/add.rs
```diff
@@ -36,18 +22,34 @@ pub fn command(to_add: String, dev: bool) -> Result<()> {
...
-        toml["dev-dependencies"][&to_add] = toml_edit::value(range);
-    } else {
-        toml["dependencies"][&to_add] = toml_edit::value(range);
-    };
+    // Insert the new deps
+    for package_to_add in packages {
+        // Pull the selected version out of the new manifest so we know what it is
+        let version = &manifest
+            .packages
+            .iter()
```
megapctr, line 30: Rust complained about `manifest.packages` being moved until I changed `.into_iter()` to `.iter()` and borrowed `manifest`, perhaps there is a better solution to this?

compiler-cli/src/add.rs
```diff
@@ -36,18 +22,34 @@ pub fn command(to_add: String, dev: bool) -> Result<()> {
...
+            .find(|package| package.name == *package_to_add)
```
megapctr, line 31: unless the `manifest` has a smart implementation of `find`, this is an `O(n*m)` operation. But since people won't be adding thousands of packages at a time this shouldn't be a problem

compiler-cli/src/add.rs
lpil: This is perfect 👍

compiler-cli/src/add.rs
lpil: Well spotted, but aye it's a tiny amount of data and we run it once, so no problem.

lpil (approved):
Nice work! Very clean. Thank you

Issue #1513, opened 2022-02-26
Debounce compilation of project and run on keypress


lpil:
Use `recv_timeout` to identify periods of idleness suitable for recompilation.

https://docs.rs/crossbeam/latest/crossbeam/channel/struct.Receiver.html#method.recv_timeout

lpil:
Closing this. We can compile on save for now.

PR #1525, opened 2022-02-28
Fix compile errors and cover a few more error cases

wesleimp:


compiler-cli/src/lsp.rs
```diff
@@ -236,6 +239,12 @@ impl LanguageServer {
         Ok(())
     }
 
+    fn did_open(&mut self) -> Result<()> {
+        // Recompile the project to detect any errors
+        let _ = self.compiler.compile()?;
+        Ok(())
+    }
```
lpil, line 246: Do we need to recompile a project when a file is opened? There will be no changes.

compiler-cli/src/main.rs
```diff
@@ -396,11 +395,11 @@ fn command_build(
...
-    diagnostic::write_title(
-        &mut buffer,
-        "Deprecated rebar3 build command",
-        Severity::Warning,
-    );
+    // diagnostic::write_title(
+    //     &mut buffer,
+    //     "Deprecated rebar3 build command",
+    //     Severity::Warning,
+    // );
```
lpil, line 402: We want to keep this

compiler-core/src/diagnostic.rs
```diff
@@ -43,32 +43,32 @@ pub struct Diagnostic {
 impl Diagnostic {
     pub fn write(&self, buffer: &mut Buffer) {
         use std::io::Write;
-        match self.location {
-            Some(location) => self.write_span(location, buffer),
+        match &self.location {
+            Some(location) => self.write_span(location.clone(), buffer),
```
lpil, line 47: Let's have the function take the location by reference rather than cloning it.

compiler-cli/src/lsp.rs
```diff
@@ -371,31 +378,37 @@ fn error_to_diagnostic(error: &Error) -> Option<PublishDiagnosticsParams> {
...
-                diagnostics: vec![diagnostic],
-                version: None,
-            };
+            );
             Some(diagnostic_params)
         }
 
-        Error::Type { .. }
-        | Error::UnknownImport { .. }
+        Error::Type { error, src, path } => {
```
lpil, line 392: Rather than pattern matching on each error variant we want to convert the error to a diagnostic no matter what it is, so we can remove the entire `match` expression

compiler-core/src/error.rs
```diff
@@ -689,14 +689,14 @@ Second: {}",
                         .cloned()
                         .filter(|label| !supplied.contains(label))
                         .collect();
-                    let title = if unknown.len() > 1 {
+                    let _title = if unknown.len() > 1 {
```
lpil, line 692: We want to use this title below

compiler-core/src/type_/error.rs
```diff
@@ -219,6 +220,115 @@ pub enum Error {
     },
 }
 
+impl Error {
+    pub fn to_diagnostic(&self, src: String, path: PathBuf) -> Diagnostic {
```
lpil, line 224: I think this is duplicating the existing to diagnostic code? We already convert all the type errors to diagnostics in the error module

compiler-core/src/warning.rs
```diff
@@ -43,146 +47,158 @@ your program.",
...
-                        location: *location,
-                    },
-                    format!("Hint: you can ignore it with an underscore: `_{}`.", name),
-                ),
+                    "This is a hint".to_string()
+                )
+
+                // TODO: Check all this matches. OldDiagnostic does not exists
+
+                // type_::Warning::ImplicitlyDiscardedResult { location } => (
```
lpil, line 62: Please keep these uncommented, this is the most urgent code to fix.

lpil (commented):
A nice start, I've left some comments

compiler-cli/src/lsp.rs
wesleimp: At the moment, errors/warnings are not shown when the file is opened

compiler-cli/src/main.rs
wesleimp: It seem the function has been removed, so I commented that part

compiler-core/src/error.rs
wesleimp: But it was not being used 🤔

compiler-core/src/type_/error.rs
wesleimp: you're right, I've missed that.

compiler-cli/src/main.rs
lpil: Yes that's intentional. This isn't finished yet so it should not compile. The errors are there to guide development, to ensure nothing is missed.

compiler-cli/src/lsp.rs
lpil: I don't wish to compile the project when a file is opened, and this branch is about diagnostics only. Logic around when we compile the project comes with https://github.com/gleam-lang/gleam/issues/1513

compiler-core/src/warning.rs
```diff
@@ -192,8 +254,8 @@ your program.",
         buffer
             .write_all(b"\n")
             .expect("error pretty buffer write space before");
-        let (diagnostic, extra) = self.to_diagnostic();
-        write_old(buffer, diagnostic, Severity::Warning);
+        let (_diagnostic, extra) = self.to_diagnostic();
+        // write_old(buffer, diagnostic, Severity::Warning);
```
lpil, line 258: Could we put this code back please? So it's clear there is work still to do

wesleimp: Sorry, I forgot to put this back. Done!

compiler-cli/src/lsp.rs
wesleimp: Awesome, so I will remove this

lpil (approved):
Thank you!

PR #1536, opened 2022-03-04
feat: surface warnings in LSP

### LINKED ISSUE #1511: Surface warnings in LSP

wesleimp:
Closes #1511 

image

### DIFF
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -170,7 +170,7 @@ impl LanguageServer {
                     match self.handle_notification(&connection, notification) {
                         Ok(()) => {
                             let mut diagnostics: HashMap<PathBuf, Vec<Diagnostic>> = HashMap::new();
-                            let warnings = self.compiler.project_compiler.warnings();
+                            let warnings = self.compiler.project_compiler.take_warnings();
                             for warn in warnings {
                                 let diagnostic = warn.to_diagnostic();
                                 match to_lsp_diagnostic(diagnostic) {
@@ -448,7 +448,7 @@ fn to_lsp_diagnostic(
 
         Some((path, diagnostic))
     } else {
-        None
+        todo!("Locationless warning for LSP")
     }
 }
 
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -5,7 +5,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
-    path::Path,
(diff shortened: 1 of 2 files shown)


compiler-cli/src/lsp.rs
```diff
@@ -167,14 +167,38 @@ impl LanguageServer {
                 }
 
                 lsp_server::Message::Notification(notification) => {
-                    if let Err(error) = self.handle_notification(&connection, notification) {
-                        match error_to_diagnostic(&error) {
+                    match self.handle_notification(&connection, notification) {
+                        Ok(()) => {
+                            let mut diagnostics: HashMap<PathBuf, Vec<Diagnostic>> = HashMap::new();
```
lpil, line 172: We still want to get warnings even if compilation failed, so let's move this out of the `if let/match` so that it always runs.

compiler-cli/src/lsp.rs
```diff
@@ -390,28 +445,17 @@ fn error_to_diagnostic(error: &Error) -> Option<PublishDiagnosticsParams> {
...
-            uri,
-            diagnostics: vec![diagnostic],
-            version: None,
-        };
 
-        Some(diagnostic_params)
+        Some((path, diagnostic))
     } else {
-        todo!("Locationless diagnostic for LSP")
+        None
```
lpil, line 451: Could you make this a todo please? I need to see if there's a way to surface these other warnings.

compiler-core/src/build/project_compiler.rs
```diff
@@ -96,6 +96,10 @@ where
         self.options.mode
     }
 
+    pub fn warnings(&mut self) -> Vec<Warning> {
```
lpil, line 99: Could you call this `take_warnings` please? To make it clear that it mutates the state of the compiler.

lpil (commented):
Very nice work. I've left some small comments

compiler-cli/src/lsp.rs
wesleimp: Sure. I missed that

compiler-cli/src/lsp.rs
wesleimp: My initial approach was like this, but the compiler does not return warnings if some error occurs. I don't if it's related to the order they occur, I mean, the compiler returns only the warnings before the error.

wesleimp: Done!

lpil (approved):
Thank yoiu

Issue #1477, opened 2022-01-26
Clear build directory before compilation if the compiler version has changed


lpil:
Let's write the Gleam compiler version to the build directory and wipe the directory if it is absent or has a different compiler version.

When running the build command

1. Read a version file from the build directory
2. Check to see if the contents of the file checks the version of the compiler
3. Clear the build directory if it does not match

We can use the existing helper functions for file IO

vstreame:
I'd love it give this a shot. I'll see if I can open a draft PR for this today

lpil:
Thank you!

PR #1547, opened 2022-03-06
Clear build directory if compiler version changes

vstreame:
This updates the project compiler to first look at a `build/gleam_version` file to know what version the artifacts were last built with. If the version has changed (or if the file doesn't yet exist) then we first delete the `build` directory and create a new `gleam_version` file with the updated version. This should ensure that we don't reuse any artifacts from the last version.

This is my first contribution to this project (and to a non-toy rust codebase) so I'm sure I may not have followed all the right conventions. Happy to make any and all modifications!

Fixes https://github.com/gleam-lang/gleam/issues/1477

### DIFF
--- a/compiler-core/src/build/project_compiler.rs
+++ b/compiler-core/src/build/project_compiler.rs
@@ -5,11 +5,13 @@ use crate::{
     },
     codegen::{self, ErlangApp},
     config::PackageConfig,
+    error::{FileIoAction, FileKind},
     io::{CommandExecutor, FileSystemIO, FileSystemWriter},
     metadata, paths,
     project::ManifestPackage,
     type_,
     uid::UniqueIdGenerator,
+    version::COMPILER_VERSION,
     warning, Error, Result, Warning,
 };
 use std::{
@@ -106,6 +108,7 @@ where
 
     /// Returns the compiled information from the root package
     pub fn compile(&mut self) -> Result<Package> {
+        self.check_gleam_version()?;
         self.compile_dependencies()?;
 
         if self.options.perform_codegen {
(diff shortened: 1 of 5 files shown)


compiler-core/src/paths.rs
```diff
@@ -45,6 +45,13 @@ pub fn build_deps_package_test(package: &str) -> PathBuf {
     build_deps_package(package).join("test")
 }
 
+/// A path to a special file that contains the version of gleam that last built
+/// the artifacts. If this file does not match the current version of gleam we
+/// will rebuild from scratch
+pub fn build_version() -> PathBuf {
+    build().join("version")
```
lpil, line 52: Let's call it this, to avoid any potential confusion with project version
```suggestion
pub fn build_gleam_version() -> PathBuf {
    build().join("gleam_version")
```

compiler-core/src/build/project_compiler.rs
```diff
@@ -130,6 +135,33 @@ where
...
+    /// one with confusing or hard to debug states.
+    pub fn check_gleam_version(&self) -> Result<(), Error> {
+        let build_path = paths::build();
+        let version_path = paths::build_version();
+        if self.io.is_file(&version_path) {
+            let version = self.io.read(&version_path)?;
+            if version == self.options.gleam_version {
+                return Ok(())
+            }
+        }
```
lpil, line 150: Could we log a message here?
```suggestion
        }

        tracing::info!("removing_build_state_from_different_gleam_version");

```

lpil (commented):
Great stuff! The code is super clear, and I like your use of comments.

Left a couple tiny notes inline

vstreame:
@lpil Thanks for the feedback! I'll update. Question: As I was doing this I used the env var `CARGO_PKG_VERSION` from `compiler-cli` and `compiler-wasm`. But I realize now I may just be able to use the `CARGO_PKG_VERSION` from the `compiler-core` itself. Would that be better? Or should we change the the CLI/WASM versions change?

HarryET:
Might it be worth regenerating the manifest.toml too?

lpil:
@vstreame Getting it from core sounds good. We can just refer to that constant throughout core and cli.

vstreame:
@HarryET Is the `manifest.toml` file is the equivalent to a lock file in other languages? Wouldn't regenerating it mean to lose your locked versions of your dependencies? Or is the file related to something else?

lpil:
Aye, it's a lock file. We don't need to do anything about that unless the format changes in a future compiler version. If it does we can detect that from the manifest itself so there's nothing to do in this case.

vstreame:
@lpil Okay, used the version of the `compiler-core` itself. That shrunk the PR/commits quite a bit. This change should be good to go now.

vstreame:
@lpil Added line to `CHANGELOG` file

lpil (approved):
Thank you!

lpil:
Hm, I'm not sure why the tests failed. Looks like a problem with CI. I'm going to merge this.

Issue #1539, opened 2022-03-04
`Unknown label` error message is somewhat unclear


TanklesXL:
For example:
```
error: Unknown label
    ┌─ ./src/gleam/otp/task.gleam:176:15
    │
176 │   |> list.map(wit: async_apply(_, list.map(_, f)))
    │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Did you mean `with`?

The other labelled arguments that this constructor accepts are `with`.
```
In this case it would be expected that only the `wit` would be underlined.

Also, one would expect the second part of the error to list the all/ the other labels available (in this case `list.map` doesn't have others so maybe it should just list `with` as the only one with a clearer message wording), maybe similar to the message generated for struct field access. 

For example: 
```
error: Unknown record field
    ┌─ ./src/gleam/otp/task.gleam:140:30
    │
140 │       process.close_channels(task.receive)
    │                              ^^^^^^^^^^^^ Did you mean `receiver`?

The value being accessed has this type:

    Task(a)

It has these fields:

    .owner
    .pid
    .receiver
```

lpil:
Great suggestion, let's improve this.

So something like this?

```
error: Unknown label
    ┌─ ./src/gleam/otp/task.gleam:176:15
    │
176 │   |> list.map(wit: async_apply(_, list.map(_, f)))
    │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Did you mean `with`?

It accepts these labels:

    with
```

Perhaps it could include the type


```
error: Unknown label
    ┌─ ./src/gleam/otp/task.gleam:176:15
    │
176 │   |> list.map(wit: async_apply(_, list.map(_, f)))
    │               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Did you mean `with`?

It accepts these labels:

    with: fn(a) -> b
```

TanklesXL:
I love the idea of having the type in there, could be added to the record access message as well.

PR #1548, opened 2022-03-06
feat(error): Update unknown label formatting

vstreame:
Instead of listing the labels on one (wrapped) line we instead put each label on its own line. This both matches other errors like "Unknown record field" as well as allows room for metadata to be shown in the future. I.E. The types of these labels

Fixes https://github.com/gleam-lang/gleam/issues/1539

I tried going doing the second part to show what type each label has, but it doesn't look like that information is available?
 In `type_/fields.rs` it seems the only thing we have is a `HashMap<String, usize>`. Is there a canonical way to get from the index / string to the corresponding types? If we could do that then we can add the type information to the `UnknownLabels` struct.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 - You can now specify multiple packages when using `gleam add`
 - If the compiler version changes we now rebuild the project from scratch on
   next build command (#1547)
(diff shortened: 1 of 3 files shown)


lpil (approved):
Fab, thank you

Issue #1549, opened 2022-03-06
Fix rebar3 missing error on CI


lpil:
The main branch is currently broken

PR #1550, opened 2022-03-06
 fix(build): Do not destroy downloaded packages

vstreame:
This patch corrects which directory we delete when we detect that the compiler version has changed. Before we were deleting the top level `build` directory. But that was also deleting our downloaded `packages` and never re-downloading them.

This fixes that by leaving the dependencies alone and instead _only_ deleting/recompiling the build targets instead.

Fixes https://github.com/gleam-lang/gleam/issues/1549

### DIFF
--- a/compiler-core/src/build/project_compiler.rs
+++ b/compiler-core/src/build/project_compiler.rs
@@ -138,8 +138,8 @@ where
     /// before continuing. This will ensure that upgrading gleam will not leave
     /// one with confusing or hard to debug states.
     pub fn check_gleam_version(&self) -> Result<(), Error> {
-        let build_path = paths::build();
-        let version_path = paths::build_gleam_version();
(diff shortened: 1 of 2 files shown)


vstreame:
Looks like Rebar is being installed, but the `bin` isn't callable?

```
/home/runner/work/_actions/erlef/setup-beam/v1.11.0/dist/install-rebar3.sh 3.14.3
  Installed rebar3 version follows
  rebar 3.14.3 on Erlang/OTP [23](https://github.com/gleam-lang/gleam/runs/5440943559?check_suite_focus=true#step:3:23) Erts 11.1.8
```

 Looking further

vstreame:
Turns out there was nothing wrong with CI and instead this bug was introduced in https://github.com/gleam-lang/gleam/pull/1547. Sorry about that!

lpil (approved):
Wonderful, thank you

Issue #1532, opened 2022-03-03
Split diagnostic text into error + hints in order to improve presentation in editors


wesleimp:

lpil:
Hello! What does this mean?

wesleimp:
Most of languages send their hints as a diagnostic event, and not as a plain string. e.g

![image](https://user-images.githubusercontent.com/42675056/156775713-7d7cb440-1b7d-4c45-9772-07bcddc1a75a.png)

We could convert the hint string in `Warning` module to a `Diagnostic` and send warning and hint event together

lpil:
Oh I see! For the first iteration I'd like to focus on getting just the basic functionality ready, we can revisit this and refine the diagnostics emitted at a later date.

lpil:
In the discord we've discovered that vscode ruins the formatting of errors in the problems pane, so we will have to do this.

![image](https://user-images.githubusercontent.com/6134406/156847895-9594fa0a-2d43-498a-9adf-2113a40a12d3.png)

We'll have to redesign the `diagnostic` struct to be a bit more precise rather than having a big blob of text.

wesleimp:
I’ll tackle this one

PR #1552, opened 2022-03-07
Surface hints for warnings in LSP

wesleimp:
Splits warnings and hints as two different lsp events

image

### DIFF
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -448,17 +448,18 @@ fn to_lsp_diagnostic(
             Level::Warning => ("Warning", DiagnosticSeverity::WARNING),
         };
 
-        let mut message = format!("{}: {}\n\n", prefix, diagnostic.title);
+        let mut message = format!("{}: {}", prefix, diagnostic.title);
 
         if let Some(label) = location.label.text {
+            message.push_str("\n\n");
             message.push_str(&label);
             if !label.ends_with(['.', '?']) {
                 message.push('.');
             }
-            message.push_str("\n\n");
         }
 
         if !diagnostic.text.is_empty() {
+            message.push_str("\n\n");
             message.push_str(&diagnostic.text);
         }
 
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -11,7 +11,7 @@ use std::{
 use gleam_core::{
     ast::SrcSpan,
     build::{self, Package, ProjectCompiler},
-    diagnostic::Level,
+    diagnostic::{self, Level},
     io::{CommandExecutor, FileSystemIO},
     line_numbers::LineNumbers,
     Error, Result,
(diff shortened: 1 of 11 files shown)


lpil (commented):
Thanks for waiting! I have been busy with my new flat.

This looks good but I think it won't solve [this motivating problem](https://github.com/gleam-lang/gleam/issues/1532#issuecomment-1059563402) where the errors are formatted poorly. That is the priority here.

It may be good to look at what other LSPs do here, they may give us some ideas.

edit: I think I have been unclear with the original issue, sorry about that. I will remake it to be clearer.

compiler-cli/src/lsp.rs
```diff
@@ -225,8 +226,19 @@ impl LanguageServer {
...
+            match to_lsp_diagnostic(diagnostic.clone()) {
+                Some((path, lsp_diagnostic)) => {
+                    self.store_diagnostic(path.clone(), lsp_diagnostic.clone());
+
+                    if let Some(hint) = diagnostic.hint {
+                        let lsp_hint = Diagnostic {
+                            severity: Some(DiagnosticSeverity::HINT),
+                            message: hint,
+                            ..lsp_diagnostic
+                        };
```
lpil, line 238: I think this logic should live within the `store_diagnostic` function

compiler-core/src/diagnostic.rs
```diff
@@ -48,7 +49,13 @@ impl Diagnostic {
             None => self.write_title(buffer),
         };
 
-        writeln!(buffer, "{}", self.text).expect("write_project");
+        let mut text = self.text.clone();
```
lpil, line 52: Let's write to the buffer rather than cloning this string

compiler-core/src/warning.rs
```diff
@@ -50,7 +43,8 @@ your program.",
 
                 type_::Warning::ImplicitlyDiscardedResult { location } => Diagnostic {
                     title: "Unused result value".into(),
-                    text: "Hint: If you are sure you don't need it you can assign it to `_`".into(),
+                    text: "".into(),
+                    hint: Some("If you are sure you don't need it you can assign it to `_`".into()),
```
lpil, line 47: Are there any tests for this error formatting? I think this one has changed the formatting to have 3 empty lines in a row

wesleimp:
> Thanks for waiting! I have been busy with my new flat.
> 
> This looks good but I think it won't solve [this motivating problem](https://github.com/gleam-lang/gleam/issues/1532#issuecomment-1059563402) where the errors are formatted poorly. That is the priority here.
> 
> It may be good to look at what other LSPs do here, they may give us some ideas.
> 
> edit: I think I have been unclear with the original issue, sorry about that. I will remake it to be clearer.

Yeah, I think we're talking about two different issues. This PR splits the hint and warning in separated events. I've linked the wrong issue. Sorry for that, my bad

Maybe I should keep this PR and open up a issue related to this, WDYT?

compiler-cli/src/lsp.rs
wesleimp: I'm not sure, because `stone_diagnostic` just receives the diagnostics and store it. I think this function shouldn't have any split logic.

compiler-core/src/diagnostic.rs
wesleimp: Should I write to the buffer and then also directly write to the buffer the hint?

compiler-core/src/warning.rs
wesleimp: It's not writing 3 lines, but I'll provide the tests anyway, thanks.

wesleimp: I'll provide the tests for this

wesleimp: It's a warning and we doesn't have anything to snapshot warnings. The output is going like this:
image

wesleimp: I've created a snapshot test for warnings. I think now your comment is covered

compiler-cli/src/lsp.rs
lpil: Currently hints are only surfaced for warnings, we need them to always be included in the LSP output but we also don't want to have to copy this bit of code to every place that the `store_diagnostic` function is called. It's not ever correct to do one without the other so they should be the same function.

lpil (commented):
Looking really good, and thank you for the new snapshots. Very useful

lpil (approved):
Thank you!

Issue #1516, opened 2022-02-26
report unused imports


megapctr:
it would be nice if the compiler/lsp warned about unused imports, like it does for unused variables.

Perhaps `gleam format` could remove them automatically.

if it's useful, I collected some of the cases to be handled:
- unused imports
- unused named imports, `import foobar as foo`
- ~unused unqualified imports, `import foo.{Foo}`~ (this is already handled)
- unused named unqualified imports, `import foo.{Foo as Foobar} as foobar`, oh god why
- distinguish variables from imports of the same name, e.g.:
```
import gleam/http/request

fn handler(request) {
  io.println(request.path)
}
```

lpil:
Yup, let's get this one in! I'm surprised we don't cover this with the existing usage detection.

PR #1553, opened 2022-03-08
Report unused module imports

vstreame:
With this patch we will not track the usages of imported modules if they have no unqualified imports (aka are naked). When a module is imported but then never used we will report a warning at the end.

Things not yet known:
- [x] Does this cover modules aliases? (I.E. `import gleam/http as fetch`)
- [ ] Did I cover every usage of module? This I'm very unsure of

Things not covered in this PR:
- [ ] unused named unqualified imports (I.E. `import foo.{Foo as Foobar} as foobar`)
- [ ] distinguish variables from imports of the same name

Related to: https://github.com/gleam-lang/gleam/issues/1516

### DIFF
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -316,8 +316,8 @@ impl<'a> Environment<'a> {
                     })
             }
 
-            Some(module) => {
-                let module = self.imported_modules.get(module).ok_or_else(|| {
+            Some(m) => {
+                let module = self.imported_modules.get(m).ok_or_else(|| {
                     UnknownValueConstructorError::Module {
                         name: name.to_string(),
                         imported_modules: self
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -13,6 +13,7 @@ pub struct Environment<'a> {
     pub imported_names: HashMap<String, SrcSpan>,
     pub importable_modules: &'a im::HashMap<String, Module>,
     pub imported_modules: HashMap<String, Module>,
+    pub unused_modules: HashMap<String, SrcSpan>,
     pub imported_types: HashSet<String>,
 
     /// Values defined in the current function (or the prelude)
@@ -78,6 +79,7 @@ impl<'a> Environment<'a> {
             module_types_constructors: prelude.types_constructors.clone(),
             module_values: HashMap::new(),
             imported_modules: HashMap::new(),
+            unused_modules: HashMap::new(),
             imported_names: HashMap::new(),
             accessors: prelude.accessors.clone(),
             local_values: prelude.values.clone().into(),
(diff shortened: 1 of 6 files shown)


compiler-core/src/type_/environment.rs
```diff
@@ -41,6 +41,7 @@ pub struct Environment<'a> {
     /// entity_usages is a stack of scopes. When an entity is created it is
     /// added to the top scope. When an entity is used we crawl down the scope
     /// stack for an entity with that name and mark it as used.
+    /// NOTE: The bool in the touple here tracks if the entity has been used
```
lpil, line 44:
```suggestion
    /// NOTE: The bool in the tuple here tracks if the entity has been used
```

compiler-core/src/type_/environment.rs
```diff
@@ -55,6 +56,7 @@ pub enum EntityKind {
     ImportedType,
     ImportedTypeAndConstructor,
     ImportedValue,
+    ImportedModule,
```
lpil, line 59: The usage tracking system is fairly complex because it handles scopes, but modules can't be imported in a scoped fashion so we don't need to use it here. Instead we can store a set of imported modules in the typing environment can delete them from there as they are seen. This will be simpler and won't result in false negatives from treating the two namespaces as one.

Later this scoped usage tracking system will be deleted and replaced with a GC style graph traversal which can detect cycles, so I'd prefer not to put more items in there.

compiler-core/src/type_/tests.rs
```diff
@@ -125,6 +127,90 @@ macro_rules! assert_no_warnings {
     };
 }
 
+macro_rules! assert_import_warning {
```
lpil, line 130: Rather than making a new macro let's copy the existing pattern in some of the other macros and make it possible to supply another module that can be imported.

compiler-core/src/type_/tests.rs
```diff
@@ -1426,6 +1512,23 @@ fn unused_variable_warnings_test() {
...
+            location: SrcSpan { start: 7, end: 16 },
+        }
+    );
+
+    // Using imported function
+    assert_no_import_warning!("import gleam/foo; pub fn bar() { foo.foo() }");
+
+    // Using imported type
+    assert_no_import_warning!("import gleam/foo; pub fn bar(a: foo.Baz) { a }");
+}
```
lpil, line 1530: Can we have 1 assertion per test? We're slowly migrating all the tests to that.

Could we get a test that no warnings are emitted if an imported module isn't used but does import unqualified values

lpil (commented):
Looking good. I'd like to change the approach a little here to make it simpler and possibly more precise. I've left details inline

compiler-core/src/type_/environment.rs
ekosz: @lpil Sounds good. I'll update to use that technique instead

vstreame:
@lpil Updated the implementation. Please let me know if this is closer to what you were thinking of!

lpil (approved):
Wonderful! Thank you very much.

Sorry for the delay, I'm in the process of finishing off my home renovation and it gets tricky to fit that in around work and Gleam.

lpil:
Oh it is really cool seeing the warnings in the LSP

vstreame:
@lpil Yesss, glad its working!

PR #1555, opened 2022-03-11
feat(parse): Add hint for multi line clauses

### LINKED ISSUE #1404: Case clause with multiple expressions not obvious for beginners
As a beginner, one might try to put multiple expressions on the right side of a case clause like so
```
pub fn a_function() {
  case True {
    True ->
      let a = 1
      a + 1
    False -> 0
  }
}

```
If we try to compile this we get this error:
```
error: Syntax error
  ┌─ /.../example_test.gleam:6:9
  │
6 │       a + 1
  │         ^ I was not expecting this.

Expected one of: "->"
```
We can fix the code by using braces to turn the multiple expressions into one like so
```
pub fn a_function() {
  case True {
    True -> {
      let a = 1
      a + 1
    }
    False -> 0
  }
}
```
but I think this is not obvious for a beginner, the compiler message does not help and the Language Tour page for Case https://gleam.run/book/tour/case-expressions.html does not show an example of this. This is what I felt when I encountered this. Maybe the compiler could be changed to produce a more helpful message, but I don't know if this is feasible. We could also show this in the Language Tour page.

vstreame:
When we expect a `RArrow` token inside a clause but don't see one, it
may be due to the developer forgetting to wrap their multiple
expressions with curly braces. Currently the developer is left on their
own to figure that out.

This patch adds a `hint` field to the `UnexpectedToken` parse error and
hints that a developer may want to add curly braces if they were trying
to do a multiple expressions in a clause.

This also allows gleam to add future hints to other `UnexpectedToken`
errors in subsequent patches.

Fixes #1404

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
   (#1548)
 - Permit type holes in function arguments and return annotations (#1519)
 - Report unused module imports (#1553)
+- Add parse error hint for multi line clauses without curly braces (#1555)
 
 ## v0.20.1 - 2022-02-24
 
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -152,7 +152,10 @@ where
             // there are still more tokens
             let expected = vec!["An import, const, type, if block, or function.".to_string()];
             return parse_error(
-                ParseErrorType::UnexpectedToken { expected },
+                ParseErrorType::UnexpectedToken {
+                    expected,
+                    hint: None,
+                },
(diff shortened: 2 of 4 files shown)


vstreame:
@lpil What do you think of this wording?

lpil (approved):
Looks grand! Would you mind also updating the changelog?

vstreame:
@lpil Done! Sorry that I keep forgetting to do that

lpil (approved):
Easily done :) Thank you!

PR #1557, opened 2022-03-12
gleam tasks

massivefermion:
Enable the gleam build tool to run user defined tasks.
The user can create gleam files inside `src/tasks` and run them using the gleam build tool. For example if there is a file `src/tasks/welcome.gleam`, the command `gleam welcome` will run that file.
It's possible to divide the tasks into different directories like `src/tasks/hello/world.gleam` and `src/tasks/bye/world.gleam` and run them with `gleam hello world` and `gleam bye world`.
Of course if gleam can't find a task, it will return an error.
I think this can be really helpful and hope you find it helpful too.

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -188,6 +188,9 @@ enum Command {
     /// Run the language server, to be used by editors
     #[clap(name = "lsp", hide = true)]
     LanguageServer,
+
+    #[clap(external_subcommand)]
+    Task(Vec<String>),
 }
 
 #[derive(Args, Debug, Clone)]
@@ -341,6 +344,8 @@ fn main() {
         Command::Clean => clean(),
 
         Command::LanguageServer => lsp::main(),
+
+        Command::Task(arguments) => run::command(arguments, None, run::Which::Task),
     };
 
     match result {
--- a/compiler-cli/src/run.rs
+++ b/compiler-cli/src/run.rs
@@ -12,23 +12,61 @@ use crate::fs::ProjectIO;
 pub enum Which {
     Src,
     Test,
+    Task,
 }
 
 pub fn command(arguments: Vec<String>, target: Option<Target>, which: Which) -> Result<(), Error> {
     let config = crate::config::root_config()?;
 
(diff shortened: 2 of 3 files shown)


zhenfeng-zhu:
I think it is best to explicitly specify in the toml file, instead of the file directory.

lpil:
Hi @massivefermion, thanks for the contribution.

Could you go a little more into the motivation for this feature and what problems it solves?

What does this enable that cannot be done with the main entrypoint today? For example using this pattern: https://github.com/team-balloonicorn/puck/blob/29c482b6805b8627053412c3ebe161e3dba04b1e/src/puck.gleam#L18-L28

massivefermion:
Well, actually I think it was premature P.R. because although you can run tasks in your project right now, what I had in mind was not this. The use case I want is for libraries that need to provide predefined tasks that the user can run any time they want, like phoenix's `mix phx.gen.context`, etc. But after some tests, this does not work if the project is a dependency of another project.

lpil:
Ah yes, it would be fantastic to be able to run commands defined in other packages.
We should have a think about what the API for such a system might be.

In Rust you can define binaries in your config, though it's not quite the same as you install them rather than adding them to the project.

In python you can do `python -m modulename`

With nodejs you can do `npx packagename`

massivefermion:
For the record, I think elixir does it best!

lpil:
Which way do you mean? Elixir has (I think) 6 APIs for doing this

massivefermion:
https://hexdocs.pm/mix/Mix.Task.html

lpil:
Fab. I'll make an issue for the running of commands from other packages

massivefermion:
> Which way do you mean? Elixir + Mix has at least 6 APIs for doing this

Just curious. What are the other 5?

lpil:
Mix tasks, mix aliases, mix archives, mix escripts, mix run, iex mix install, iex -S mix.

lpil:
I'm gunna close this as we have some design work to do, but thank you very much for this! I expect once we have worked out what API we want we can reuse this.

PR #1559, opened 2022-03-19
added ability to specify branch/tag/commit for repo

massivefermion:
not all repositories have version tags which means for those repositories, the links in the gleam documentation for functions, types,... won't work. So I think it would help if the developer has the ability to specify a branch,tag or commit for their repo.

P.S.
I wanted to use `ref` but it's a keyword in rust and I didn't know how to tell `serde` to rename it.

### DIFF
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -429,25 +429,45 @@ pub struct ErlangConfig {
 #[derive(Deserialize, Debug, PartialEq, Clone)]
 #[serde(tag = "type", rename_all = "lowercase")]
 pub enum Repository {
-    GitHub { user: String, repo: String },
-    GitLab { user: String, repo: String },
-    BitBucket { user: String, repo: String },
-    Custom { url: String },
+    GitHub {
+        user: String,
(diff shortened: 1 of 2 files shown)


lpil:
Hi! Thanks for the contribution.

If we allow the programmer to supply the ref for the URL in the documentation it'll need to be passed in as an argument to `gleam publish` rather than be part of the `gleam.toml` configuration. The toml file doesn't change with each commit so it would be always out of date.

massivefermion:
I didn't realize you actually wanted it to be a convention in the community. I think it's better to document that somewhere so people are aware of it.

massivefermion:
I forgot to say, this PR is completely backwards compatible. If `refer` is not specified in the repository section of `gleam.toml`, it just fall backs to the default.

massivefermion:
yeah, you're right, it doesn't make sense for a commit. I don't know why I wrote that. What I had in mind was for a branch or a tag.

lpil:
Closing this due to inactivity. Please feel free to reopen!

PR #1560, opened 2022-03-19
docs: Add module docs for statements and types

vstreame:
Figured after I learned this myself I would add comments so that future developers don't have to code splunk themselves as much to gain an understanding.

I may have gotten some of these wrong! Please correct me where I may have misunderstood what was happening.

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -280,6 +280,16 @@ pub type UntypedStatement = Statement<(), UntypedExpr, (), ()>;
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
+    /// A function definition
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// // Public function
+    /// pub fn bar() -> String { ... }
+    /// // Private function
+    /// fn foo(x: Int) -> Int { ... }
+    /// ```
     Fn {
         end_location: usize,
         location: SrcSpan,
@@ -292,6 +302,14 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// A new name for an existing type
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// pub type Headers =
+    ///   List(#(String, String))
+    /// ```
     TypeAlias {
         location: SrcSpan,
         alias: String,
@@ -302,6 +320,21 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// A newly defined type with one or more constructors.
+    /// Each variant of the custom type can contain different types, so the type is
+    /// the product of the types contained by each variant.
+    ///
+    /// This might be called an algebraic data type (ADT) or tagged union in other
+    /// languages and type systems.
+    ///
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// pub type Cat {
+    ///   Cat(name: String, cuteness: Int)
+    /// }
+    /// ```
     CustomType {
         location: SrcSpan,
         name: String,
@@ -313,6 +346,16 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         typed_parameters: Vec<T>,
     },
 
+    /// Import a function defined outside of Gleam code.
+    /// When compiling to Erlang the function could be implemented in Erlang
+    /// or Elixir, when compiling to JavaScript it might be implemented in
+    /// JavaScript or TypeScript.
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// pub external fn random_float() -> Float = "rand" "uniform"
+    /// ```
     ExternalFn {
         location: SrcSpan,
         public: bool,
@@ -325,6 +368,15 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// Import a type defined in another language.
+    /// Nothing is known about the runtime characteristics of the type, we only
+    /// know that it exists and that we have given it this name.
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// pub external type Queue(a)
+    /// ```
     ExternalType {
         location: SrcSpan,
         public: bool,
@@ -333,6 +385,16 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// Import another Gleam module so the current module can use the types and
+    /// values it defines.
+    ///
+    /// # Example(s)
+    ///
+    /// ```gleam
+    /// import unix/cat
+    /// // Import with alias
+    /// import animal/cat as kitty
+    /// ```
     Import {
         location: SrcSpan,
         module: Vec<String>,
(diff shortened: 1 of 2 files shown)


compiler-core/src/ast.rs
```diff
@@ -302,6 +320,15 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// A named collections of keys and values
```
lpil, line 323:
```suggestion
    /// A newly defined type with one or more constructors.
    /// Each variant of the custom type can contain different types, so the type is
    /// the product of the types contained by each variant.
    ///
    /// This might be called an algebraic data type (ADT) or tagged union in other
    /// languages and type systems.
    ///
```

compiler-core/src/ast.rs
```diff
@@ -313,6 +340,13 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         typed_parameters: Vec<T>,
     },
 
+    /// Function from another language
```
lpil, line 343:
```suggestion
    /// Import a function defined outside of Gleam code.
    /// When compiling to Erlang the function could be implemented in Erlang
    /// or Elixir, when compiling to JavaScript it might be implemented in
    /// JavaScript or TypeScript.
```

compiler-core/src/ast.rs
```diff
@@ -325,6 +359,13 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// Type from another language
```
lpil, line 362:
```suggestion
    /// Import a type defined in another language.
    /// Nothing is known about the runtime characteristics of the type, we only
    /// know that it exists and that we have given it this name.
```

compiler-core/src/ast.rs
```diff
@@ -333,6 +374,15 @@ pub enum Statement<T, Expr, ConstantRecordTag, PackageName> {
         doc: Option<String>,
     },
 
+    /// Import type or function from another gleam module
```
lpil, line 377:
```suggestion
    /// Import another Gleam module so the current module can use the types and
    /// values it defines.
```

compiler-core/src/type_.rs
```diff
@@ -46,25 +46,71 @@ pub trait HasType {
...
+    /// coming from the gleam prelude instead (where Int, Float, String, etc is
+    /// defined). If the `App` takes generic arguments they will be put in the
+    /// `args` field.
+    ///
+    /// # Example
+    ///
+    /// ```gleam
+    /// Int //=> Type::App { name: "Int", args: vec![], module: vec![], public: true }
+    /// String //=> Type::App { name: "String", args: vec![], module: vec![], public: true }
+    /// ```
```
lpil, line 60:
```suggestion
    /// A nominal (named) type such as `Int`, `Float`, or a programmer defined
    /// custom type such as `Person`. The type can take other types as
    /// arguments (aka "generics" or "parametric polymorphism").
    ///
    /// If the type is defined in the Gleam prelude the `module` field will be
    /// empty, otherwise it will contain the name of the module that
    /// defines the type.
```

compiler-core/src/type_.rs
```diff
@@ -46,25 +46,71 @@ pub trait HasType {
...
+    /// ```gleam
+    /// fn(x: Int, y: Int) -> Int { x + y } //=> Type::Fn {
+    /// //   args: vec![
+    /// //     Type::App { name: "Int", args: vec![], module: vec![], public: true },
+    /// //     Type::App { name: "Int", args: vec![], module: vec![], public: true },
+    /// //   ],
+    /// //   retrn: Type::App { name: "Int", args: vec![], module: vec![], public: true },
+    /// // }
+    /// ```
+    ///
```
lpil, line 84: I think it's confusing to make the comparison to the function values here, they are unrelated. It's also nothing to do with anonymous functions specifically, that's purely a matter of the value level language. All functions whether named or anonymous has this function type.

```suggestion
    /// The type of a function. It takes arguments and returns a value.
    ///
```

compiler-core/src/type_.rs
```diff
@@ -46,25 +46,71 @@ pub trait HasType {
...
+    /// ```gleam
+    /// #(Int, String) //=> Type::Tuple {
+    /// //   elems: vec![
+    /// //     Type::App { name: "Int", args: vec![], module: vec![], public: true },
+    /// //     Type::App { name: "String", args: vec![], module: vec![], public: true },
+    /// //   ],
+    /// // }
+    /// ```
+    ///
+    Tuple { elems: Vec<Arc<Type>> },
```
lpil, line 113:
```suggestion
    /// A tuple is an ordered collection of 0 or more values, each of which
    /// can have a different type, so the `tuple` type is the sum of all the
    /// contained types.
    ///
    Tuple { elems: Vec<Arc<Type>> },
```

compiler-core/src/type_.rs
```diff
@@ -46,25 +46,71 @@ pub trait HasType {
...
-        type_: Arc<RefCell<TypeVar>>,
-    },
+    /// A Var is a type variable, internally it holds a `TypeVar`, which could be
+    /// an `Unbound` variable (it is one specific type but we don't know what yet
+    /// in the inference process) a `Generic` variable (it stands in for any
+    /// possible type and cannot be specialised to any one type), or a `Link` (it
+    /// was an unbound variable but we worked out that it is some other type and
+    /// now we point to that one)
+    ///
+    Var { type_: Arc<RefCell<TypeVar>> },
```
lpil, line 97: Better to keep these `TypeVar` docs with the `TypeVar`.
```suggestion
    /// A type variable. See the contained `TypeVar` enum for more information.
    ///
    Var { type_: Arc<RefCell<TypeVar>> },
```

compiler-core/src/type_.rs
```diff
@@ -337,8 +383,44 @@ pub enum PatternConstructor {
...
 pub enum TypeVar {
+    /// Unbound is an unbound variable. It is one specific type but we don't
+    /// know what yet in the inference process. It has an id that we can refer
+    /// to later in the type inference process.
+    ///
+    /// # Example
+    ///
+    /// ```gleam
+    /// [] => Type::App { name: "List", arguments: vec![Type::Var { TypeVar::Unbound }] }
+    /// ```
```
lpil, line 394:
```suggestion
    /// Unbound is an unbound variable. It is one specific type but we don't
    /// know what yet in the inference process. It has a unique id which can be used to
    /// identify if two unbound variable Rust values are the same Gleam type variable
    /// instance or not.
    ///
```

compiler-core/src/type_.rs
```diff
@@ -337,8 +383,44 @@ pub enum PatternConstructor {
...
+    ///
+    /// ```gleam
+    /// let x = []
+    /// // Here x is Type::App { name: "List", arguments: vec![Type::Var { TypeVar::Unbound }] }
+    /// x == [1]
+    /// // Then x becomes Type::App { name: "List", arguments: vec![Type::Var {
+    /// //    TypeVar::Link { Type::App { name: "Int" } }
+    /// // }] }
+    /// ```
+    ///
```
lpil, line 410:
```suggestion
    /// Link is type variable where it was an unbound variable but we worked out
    /// that it is some other type and now we point to that one.
    ///
```

lpil (commented):
Great stuff! I've made some suggestions which I think are more accurate.

I've generally removed the examples that are not Gleam code as I found them to be unclear. If we want to have those kind of examples it would be better to have them as a textual explanation instead.

And Gleam with a capital G please :)

vstreame:
@lpil Merged in all of your suggestions. And will do about capitalizing Gleam in the future

lpil (approved):
Thank you! And thanks for waiting too

PR #1563, opened 2022-03-27
feat(javascript): Output Typescript declaration files

### LINKED ISSUE #1269: Generate typescript type definitions for Gleam modules when compiling to JavaScript
Similar to how we generate Erlang typespecs.

vstreame:
This modifies the Javascript codegen to _also_ emit Typescript type declaration files alongside the existing `.mjs` files. This allows the generated Gleam code to be better used in Typescript environments without losing type safety.

~~@lpil I tried to make sure to copy over from the Javascript tests and tests that looked relevant. But I'm still not sure I'm covering every edge case. Is there a way in Gleam to mark a feature as "experimental" and have a user opt into using it?~~ I've added a commit to allow the Typescript code generation to be toggled on/off via the project TOML file. It defaults to off.

Features I'm unsure about:
- [x] Generics - Didn't see any existing tests for these in JS
- [x] Opaque types - Not sure how to models these in TS. Typescript doesn't have any equivalent unfortunately.

Closes #1269

As an aside, I wanted to record myself building this to possibly get more people interested in Gleam (not saying I succeeded haha). Here are the video links for any future code spelunkers:
Part 1:  https://youtu.be/bEcCyiO1b4k
Part 2: https://youtu.be/t472JiJ1Lnk
Part 3: https://youtu.be/WxHlU8thAMY

### DIFF
--- a/compiler-core/src/javascript/typescript.rs
+++ b/compiler-core/src/javascript/typescript.rs
@@ -1,0 +1,774 @@
+//! This module is responsible for generating TypeScript type declaration files.
+//! This code is run during the code generation phase along side the normal
+//! Javascript code emission. Here we walk through the typed AST and translate
+//! the Gleam statements into their TypeScript equivalent. Unlike the Javascript
+//! code generation, the TypeScript generation only needs to look at the module
+//! statements and not the expressions that may be _inside_ those statements.
+//! This is due to the TypeScript declarations only caring about inputs and outputs
+//! rather than _how_ those outputs are generated.
+//!
+//! ## Links
+//! <https://www.typescriptlang.org/>
+//! <https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html>
+
+use std::{collections::HashMap, ops::Deref, sync::Arc};
+
+use heck::ToUpperCamelCase;
+use itertools::Itertools;
+
+use crate::{
+    ast::{
+        Statement, TypedArg, TypedConstant, TypedExternalFnArg, TypedModule,
+        TypedRecordConstructor, TypedStatement,
+    },
+    docvec,
+    pretty::{break_, Document, Documentable},
+    type_::{Type, TypeVar},
+};
+
+use super::{concat, import::Imports, line, lines, wrap_args, Output, INDENT};
+
+/// The `TypePrinter` contains the logic of how to convert Gleam's type system
+/// into the equivalent TypeScript type strings.
+///
+#[derive(Debug)]
+struct TypePrinter<'a> {
+    tracker: UsageTracker,
+    current_module: &'a [String],
+}
+
+impl<'a> TypePrinter<'a> {
+    fn new(current_module: &'a [String]) -> Self {
+        Self {
+            current_module,
+            tracker: UsageTracker::default(),
+        }
+    }
+
+    /// Converts a Gleam type into a TypeScript type string
+    ///
+    pub fn print(&mut self, type_: &Type) -> Document<'static> {
+        self.do_print(type_, None)
+    }
+
+    /// Helper function for genering a TypeScript type string after collecting
+    /// all of the generics used in a statement
+    ///
+    pub fn print_with_generic_usages(
+        &mut self,
+        type_: &Type,
+        generic_usages: &HashMap<u64, u64>,
+    ) -> Document<'static> {
+        self.do_print(type_, Some(generic_usages))
+    }
+
+    fn do_print(
+        &mut self,
+        type_: &Type,
+        generic_usages: Option<&HashMap<u64, u64>>,
+    ) -> Document<'static> {
+        match type_ {
+            Type::Var { type_: typ } => self.print_var(&typ.borrow(), generic_usages),
+
+            Type::App {
+                name, module, args, ..
+            } if module.is_empty() => self.print_prelude_type(name, args, generic_usages),
+
+            Type::App {
+                name, args, module, ..
+            } => self.print_type_app(name, args, module, generic_usages),
+
+            Type::Fn { args, retrn } => self.print_fn(args, retrn, generic_usages),
+
+            Type::Tuple { elems } => tuple(elems.iter().map(|e| self.do_print(e, generic_usages))),
+        }
+    }
+
+    fn print_var(
+        &mut self,
+        type_: &TypeVar,
+        generic_usages: Option<&HashMap<u64, u64>>,
+    ) -> Document<'static> {
+        match type_ {
+            TypeVar::Generic { id } => match &generic_usages {
+                Some(usages) => match usages.get(id) {
+                    Some(&0) => super::nil(),
+                    Some(&1) => "any".to_doc(),
+                    _ => id_to_type_var(*id),
+                },
+                None => id_to_type_var(*id),
+            },
+            // Shouldn't get here unless something went wrong
+            TypeVar::Unbound { .. } => "any".to_doc(),
+            TypeVar::Link { type_: typ } => self.do_print(typ, generic_usages),
+        }
+    }
+
+    /// Prints a type coming from the Gleam prelude module. These are often the
+    /// low level types the rest of the type system are built up from. If there
+    /// is no built-in TypeScript equivalent, the type is prefixed with "$Gleam."
+    /// and the Gleam prelude namespace will be imported during the code emission.
+    ///
+    fn print_prelude_type(
+        &mut self,
+        name: &str,
+        args: &[Arc<Type>],
+        generic_usages: Option<&HashMap<u64, u64>>,
+    ) -> Document<'static> {
+        match name {
+            "Nil" => "null".to_doc(),
+            "Int" | "Float" => "number".to_doc(),
+            "UtfCodepoint" => {
+                self.tracker.prelude_used = true;
+                "$Gleam.UtfCodepoint".to_doc()
+            }
+            "String" => "string".to_doc(),
+            "Bool" => "boolean".to_doc(),
+            "BitString" => {
+                self.tracker.prelude_used = true;
+                "$Gleam.BitString".to_doc()
+            }
+            "List" => {
+                self.tracker.prelude_used = true;
+                docvec![
+                    "$Gleam.List",
+                    wrap_generic_args(args.iter().map(|x| self.do_print(x, generic_usages)))
+                ]
+            }
+            "Result" => {
+                self.tracker.prelude_used = true;
+                docvec![
+                    "$Gleam.Result",
+                    wrap_generic_args(args.iter().map(|x| self.do_print(x, generic_usages)))
+                ]
+            }
+            // Getting here sholud mean we either forgot a built-in type or there is a
+            // compiler error
+            name => panic!("{} is not a built-in type.", name),
+        }
+    }
+
+    /// Prints a "named" programmer-defined Gleam type into the TypeScript
+    /// equivalent.
+    ///
+    fn print_type_app(
+        &mut self,
+        name: &str,
+        args: &[Arc<Type>],
+        module: &[String],
+        generic_usages: Option<&HashMap<u64, u64>>,
+    ) -> Document<'static> {
+        let name = format!("{}$", ts_safe_type_name(name.to_upper_camel_case()));
+        let name = match module == self.current_module {
+            true => Document::String(name),
+            false => {
+                // If type comes from a seperate module, use that module's nam
+                // as a TypeScript namespace prefix
+                docvec![
+                    Document::String(format!("${}", module_name(module))),
+                    ".",
+                    Document::String(name),
+                ]
+            }
+        };
+        if args.is_empty() {
+            return name;
+        }
+
+        // If the App type takes arguments, pass them in as TypeScript generics
+        docvec![
+            name,
+            wrap_generic_args(args.iter().map(|a| self.do_print(a, generic_usages)))
+        ]
+    }
+
+    /// Prints the TypeScript type for an anonymous function (aka lambda)
+    ///
+    fn print_fn(
+        &mut self,
+        args: &[Arc<Type>],
+        retrn: &Type,
+        generic_usages: Option<&HashMap<u64, u64>>,
+    ) -> Document<'static> {
+        docvec![
+            wrap_args(args.iter().enumerate().map(|(idx, a)| docvec![
+                "x",
+                idx,
+                ": ",
+                self.do_print(a, generic_usages)
+            ])),
+            " => ",
+            self.do_print(retrn, generic_usages)
+        ]
+    }
+
+    /// Allows an outside module to mark the Gleam prelude as "used"
+    ///
+    pub fn set_prelude_used(&mut self) {
+        self.tracker.prelude_used = true
+    }
+
+    /// Returns if the Gleam prelude has been used at all during the process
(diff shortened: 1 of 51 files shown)


compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+        TypedRecordConstructor, TypedStatement,
+    },
+    docvec,
+    pretty::{break_, Document, Documentable},
+    type_::{Type, TypeVar},
+};
+
+use super::{concat, import::Imports, line, lines, wrap_args, Output};
+
+const INDENT: isize = 4;
```
lpil, line 31:
```suggestion
use super::{INDENT, concat, import::Imports, line, lines, wrap_args, Output};
```

Let's use the same as in the JS code

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+    fn new(current_module: &'a [String]) -> Self {
+        Self {
+            current_module,
+            tracker: Rc::new(RefCell::new(UsageTracker::default())),
+        }
+    }
+
+    /// Converts a Gleam type into a Typescript type string
+    ///
+    pub fn print(&self, type_: &Type) -> Document<'static> {
```
lpil, line 52: Could we not store this directly in the TypePrinter and then pass around a mutable reference to that? From what I can see the UsageTracker is not shared elsewhere

```suggestion
#[derive(Debug)]
struct TypePrinter<'a> {
    tracker: UsageTracker,
    current_module: &'a [String],
}

impl<'a> TypePrinter<'a> {
    fn new(current_module: &'a [String]) -> Self {
        Self {
            current_module,
            tracker: UsageTracker::default(),
        }
    }

    /// Converts a Gleam type into a Typescript type string
    ///
    pub fn print(&mut self, type_: &Type) -> Document<'static> {
```

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+    pub fn prelude_used(&self) -> bool {
+        self.tracker.borrow().prelude_used
+    }
+}
+
+/// Prints a Gleam tuple in the Typescript equivalent syntax
+///
+fn tuple<'a>(elems: impl IntoIterator<Item = Document<'a>>) -> Document<'a> {
+    concat(Itertools::intersperse(elems.into_iter(), break_(",", ", ")))
+        .nest_current()
```
lpil, line 177: Can we indent by 1 rather than using `nest_current`? I think that would suit the TypeScript style more.

It would also be good to have a test for how a really long tuple renders, to ensure it is formatted correctly.

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+        module: &[String],
+    ) -> Document<'static> {
+        let name = format!("{}$", ts_safe_type_name(name.to_camel_case()));
+        let name = match module == self.current_module {
+            true => Document::String(name),
+            false => {
+                // If type comes from a seperate module, use that module's nam
+                // as a Typescript namespace prefix
+                let module_name = module
+                    .last()
```
lpil, line 129: What happens if two modules have the same final segment in their name?

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+        };
+        if args.is_empty() {
+            return name;
+        }
+        // If the App type takes arguments, pass them in as Typescript generics
+        let args = concat(Itertools::intersperse(
+            args.iter().map(|a| self.print(a)),
+            ", ".to_doc(),
+        ));
+        docvec![name, "<", args, ">"]
```
lpil, line 143: It would be nice to construct this document so that it wraps nicely if it doesn't fit on one line, similar to how a function call works for JS generation.

Once we have it we could also reuse this nicely indenting document construction for printing Result and List from the stdlib.

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+    ///
+    fn print_fn(&self, args: &[Arc<Type>], retrn: &Type) -> Document<'static> {
+        let args = concat(Itertools::intersperse(
+            args.iter()
+                .enumerate()
+                .map(|(idx, a)| docvec!["x", idx, ": ", self.print(a)]),
+            ", ".to_doc(),
+        ));
+        let retrn = self.print(retrn);
+        docvec!["(", args, ") => ", retrn]
```
lpil, line 156: Same as with type app, it would be nice to have this wrap and indent nicely when the line is too long

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+        match statement {
+            Statement::TypeAlias {
+                alias,
+                public,
+                type_,
+                ..
+            } if *public => vec![self.type_alias(alias.clone(), type_)],
+            Statement::TypeAlias { .. } => vec![],
+
+            // TODO: Do we need to do anything here?
```
lpil, line 333: Yes please! Define it as `any`

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+
+            Statement::Import { .. } => vec![],
+
+            Statement::CustomType {
+                public,
+                constructors,
+                opaque,
+                name,
+                ..
+            } if *public && !*opaque => self.custom_type_definition(name, constructors),
```
lpil, line 344: We would want to define the type but not the constructors if it is opaque and public

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+
+        definitions.push(Ok(docvec![
+            Document::String("export type ".to_string()),
+            Document::String(format!("{}$", name)),
+            Document::String(" = ".to_string()),
+            concat(Itertools::intersperse(
+                constructors.iter().map(|x| x.name.to_doc()),
+                break_("| ", " | "),
+            )),
+            Document::String(";".to_string()),
```
lpil, line 411:
```suggestion
            "export type ",
            name,
            "$ = ",
            concat(Itertools::intersperse(
                constructors.iter().map(|x| x.name.to_doc()),
                break_("| ", " | "),
            )),
            ";",
```

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+                    .label
+                    .as_ref()
+                    .map(|s| super::maybe_escape_identifier_doc(s))
+                    .unwrap_or_else(|| Document::String(format!("x{}", i)));
+                docvec![name, ": ", self.type_printer.print(&arg.type_)]
+            }),
+            break_(",", ", "),
+        ));
+
+        let class_body = docvec![line(), "constructor(", parameters, "): this",].nest(INDENT);
```
lpil, line 441: I think constructors are not supposed to have any return annotation?

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+
+    fn fun_args(&self, args: &'a [TypedArg]) -> Document<'a> {
+        wrap_args(
+            args.iter()
+                .enumerate()
+                .map(|(i, a)| match a.get_variable_name() {
+                    None => {
+                        docvec![
+                            Document::String(format!("x{}", i)),
+                            ": ",
```
lpil, line 454:
```suggestion
                            "x",
                            i
                            ": ",
```

compiler-core/src/javascript/typescript.rs
```diff
@@ -0,0 +1,530 @@
...
+            self.type_printer.print(return_type),
+            ";",
+        ])
+    }
+
+    fn external_fun_args(&self, args: &'a [TypedExternalFnArg]) -> Document<'a> {
+        wrap_args(args.iter().enumerate().map(|(i, a)| match &a.label {
+            None => {
+                docvec![
+                    Document::String(format!("x{}", i)),
```
lpil, line 513:
```suggestion
                    "x", i,
```

lpil (commented):
Beautiful work!

The `$` suffix on types is interesting, I hadn't considered the single namespace in TypeScript. I quite like it, but I wonder if there were any other options you considered? Would putting the word "type" in there in some fashion be informative to the reader?

I may have missed this but could we get a test for:
- functions that return functions
- functions that return unapplied record constructors
- opaque types

Thank you

vstreame:
@lpil Thanks for the kind words! I did consider appending the word "Type" as a suffix instead of "$", but I changed to "$" as I thought that had less of a chance to clash with any identifiers already in gleam. I assume that why you originally added the "$" prefix for imported modules? 

I'll definitely add those tests. Rest of the comments I'll make inline!

compiler-core/src/javascript/typescript.rs
vstreame: Yeah I tried doing that before, but it caused me to have to mark every single function at `mut` as well as every function in the `TypescriptGenerator` as well. But I'll give it another shot for sure!

compiler-core/src/javascript/typescript.rs
vstreame: Yeah, I wondered about that as well. Right now it would cause issues. What I was thinking is respecting the `as_name` field during the import statement, but here when looking at the `App` type_ I only have access to the original module vector rather than its alias. So I think when I collect the import I would need to build up some hash map of `&[String] => String`. But can use you a pointer as a key like that in Rust?

compiler-core/src/javascript/typescript.rs
vstreame: Will look at what Javascript does and move that here as well!

compiler-core/src/javascript/typescript.rs
vstreame: Will add a test

compiler-core/src/javascript/typescript.rs
vstreame: So the idea is that in the Typescript declaration file it would look something like this?

Input:
```gleam
pub opaque type Animal {
  Cat(name: String)
  Dog(name: String)
} 
```

Output:
```typescript
import * as $gleam from "../gleam.d.ts";

class Cat extends $gleam.CustomType {
  constructor(name: string);
  name: string;
}

class Dog extends $gleam.CustomType {
  constructor(name: string);
  name: string;
}

export type Animal$ = Cat | Dog;
```

So the type is "exposed" but no way to import the individual constructors?

compiler-core/src/javascript/typescript.rs
vstreame: I was getting some weird errors from Rust about not using a variable where it was defined without this (I think it shows up in the second video), but I'll try again without these!

compiler-core/src/javascript/typescript.rs
vstreame: Yeah, just looked again and you're totally right. Will also update this to wrap the argument list like the functions

compiler-core/templates/prelude.d.ts
```diff
@@ -1,3 +1,8 @@
+export class CustomType {
+  inspect(): string;
+  withFields(fields: any): this;
```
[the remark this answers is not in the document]
vstreame: Yeah I mentioned in the video I would make this better off-camera. Looks like I forgot to do that :( But I think I can do something here that's even safer than `object` (which allows `null`s)

vstreame:
Okay just pushed up

1. Updated Changelog message
2. `s/Typescript/TypeScript` TIL honestly
3. Added tests for
    * Unapplied record constructor
    * Opaque types
    * External types
    * Returning a function
4. Added typed fields to the records
    Before the generated records looked like
    ```typescript
    class Cat extends $gleam.CustomType {
      constructor(name: string);
    }
    ```
    But I realized that that's slightly wrong. We also need to then type the fields on the class.
    ```typescript
    class Cat extends $gleam.CustomType {
      constructor(name: string);

      name: string;
    }
    ```
5. Using JS's `INDENT`
6. Added a helper called `wrap_generic_args` which works the same as JS's `wrap_args` function. But with `<`/`>` instead of parens
7. Added external types
8. Made opaque records no longer export their constructors
9. Updated `withFields` type signature

vstreame:
Started work on generics. Looks like this will be more complicated than I hoped. I'm borrowing/stealing a bunch of ideas & code from how `erlangs.rs` handles them. Left a failing test to remind myself where I left things. 

Got module and external functions working. Need to do records next

vstreame:
Okay, I added passing tests for generics. I think I covered the common cases? But would love more feedback on situations I may not be handling still.

I also:
1. Fixed the formatting of TS tuples and added a tests for them
2. Added a formatting test for functions

@lpil When you're free next I would love your thoughts on that hash table idea I commented about to resolve the issue of two modules with the same name. As well as make sure I understood and implemented the `opaque` types correctly.

But feels like we're really close here!

compiler-core/src/javascript/typescript.rs
lpil: I think so, but we could also just convert it to a string if that's a problem.

compiler-core/src/javascript/typescript.rs
lpil: If that's valid typescript that would be great. Sorry for taking so long to reply, I missed this one

lpil:
Hi E! What's the status here? Do you need me to take a look at anything? Thank you

vstreame:
@lpil Sorry! I was a bit busy the past couple weeks. This should be good to merge now. But after merge I'll create a ticket to deal with modules who's last segment is the same name.

lpil:
Thanks @vstreame but I would prefer to wait until the issue is fixed before merging. Let me know if you'd like me to pick it up rather than waiting. No rush on my end.

vstreame:
@lpil I just pushed a fix for the issue. Now we construct the module name via joining the module parts rather than just take the last part.

FYI, this is still a bug in the _javascript_ builder as it does not do this yet. So we'll still need to create a ticket to fix that

lpil:
Hi @vstreame !

I think we've still got the possibility of an import collision when using this approach. Here's an example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bbd993cb1adb804e1796137be7d6af06

Rather than using the module name we could use the alias (if present). The compiler will ensure these are unique during compilation so then we'll only need to worry about the prelude, for which we could use a name like `_` or `$`.

vstreame:
@lpil Was that the correct link? I think that is a Rust example instead of a Gleam one.

The reason I didn't use aliases is that the Types do not have access to them. For example [the App type](https://github.com/gleam-lang/gleam/blob/c73b14f0a434127959b4ab8f88ae229fa69ee337/compiler-core/src/type_.rs#L58-L63) only has access to the module array. 

Do you think I should add the alias field to the `App` type?

lpil:
Yes sorry, it was showing how the Rust code could produce a collision.

Ah! I see what you mean about the aliases now! We will have to maintain some state that maps imported modules onto names as without that extra information there is no way to know if the name has been taken already or not. I'm having quite a busy week but later I could do this if you are busy.

Thanks

lpil:
Thank you so much for this @vstreame , and for your patience! This is fantastic work

Issue #1502, opened 2022-02-25
Removing .gleam files seems to not force them to be removed from build / gleam run continues to execute them


inoas:
1. Created a project with `gleam new module_test`
2. Wondered why the main module is not called main` or `app` and just renamed `module_test.gleam` to `main.gleam`
3. Changed code in `main.gleam`
4. Compiled and executed via `gleam run`
5. Changes I did to main.gleam where not reflected in the output of `gleam run`
6. `gleam clean` and then `gleam run` resulted in the compiler error (as expected): `exception error: undefined function module_test:main/0%`

I guess this is similar to elixir/erlang and that I can change the entry point application run by `gleam run` via changes to `gleam.toml`=>`name` but was a bit surprised that the compiler did not delete build artifacts when there are no more corresponding source code files:

`module_test.gleam` was gone (renamed to main.gleam) and gleam run still ran the compiled version out of build (I guess).

lpil:
Ah yes, it would be good to clear outdated files.

PR #1566, opened 2022-03-31
fix(tooling): clear outdated files from the build directory before compilation

mrgleam:
#1502

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -89,7 +89,12 @@ impl gleam_core::io::FileSystemReader for ProjectIO {
                 .filter_map(Result::ok)
                 .filter(|e| e.file_type().is_file())
                 .map(|d| d.into_path())
-                .filter(|p| p.file_name().and_then(OsStr::to_str).map(|f| !f.contains("@@")).unwrap_or(false))
+                .filter(|p| {
+                    p.file_name()
+                        .and_then(OsStr::to_str)
+                        .map(|f| !f.contains("@@"))
+                        .unwrap_or(false)
+                })
                 .filter(|p| p.extension().and_then(OsStr::to_str) == Some("erl"))
         })
     }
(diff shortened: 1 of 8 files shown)


lpil:
Hi @mrgleam ! What is the motivation for this change?

mrgleam:
> Hi @mrgleam ! What is the motivation for this change?

I followed from this comment.

https://github.com/gleam-lang/gleam/issues/1502#issuecomment-1053544130

mrgleam:
> AH I see. I have been unclear, sorry about that. We don't want to delete the entire directory (as this would then require a full compilation again), we instead want to delete compiled files from previous runs that no longer exist for the current one.
> 
> For example, if we compile a project and have a `one.gleam` it'll make `one.erl` and other files. If we delete the `one.gleam` and compile it again it should remove `one.erl` etc too.

Thank you very much. My English is not good.
It's a very clear explanation.

I can get the gleam source files and check with file in the build directory but how about gleam@@compile.erl, gleam@@main.erl and gleam@@main.beam ?

What should I do with  gleam@@compile.erl, gleam@@main.erl and gleam@@main.beam ?

Screen Shot 2565-04-02 at 04 03 18

lpil:
Those files should always remain. I think the ideal solution would track what files are written on each pass, diff those sets and delete the files missing from the latest run.

compiler-cli/src/fs.rs
```diff
@@ -79,6 +79,20 @@ impl gleam_core::io::FileSystemReader for ProjectIO {
                 .collect()
         })
     }
+
+    fn erlang_files(&self, dir: &Path) -> Box<dyn Iterator<Item = PathBuf>> {
```
lpil, line 83: This will not catch any files that are not `.erl` files. For example `.hrl`, `.mjs`, metadata and fingerprint files. Rather than try and find them all after the fact I think we should store on disc a record of all the files written and diff that against the files written in the latest compilation.

mrgleam: Do you mean record of all the files written under function perform_codegen and encode_and_write_metadata right?

Screen Shot 2565-04-05 at 22 09 28

mrgleam:
I open new PR at https://github.com/gleam-lang/gleam/pull/1578

PR #1568, opened 2022-04-03
Bump `heck` and `strum` dependency versions

Veykril:
Quick bump of `heck` and `strum` to reduce duplicated transitive dependencies.

There is one dupe left, `base64 v0.12.3` and `base64 v0.13.0`, but that one can't be removed as it is introduced by `hexpm` and its dependencies.


lpil (approved):
Excellent! Thank you

PR #1569, opened 2022-04-04
feat(codegen): Skip dev-dependencies in `.app` if not root project

vstreame:
Currently we add all dependencies to the generated `.app` folder even if
they are `dev_dependencies` and we're not currently compiling the root
project. This can cause errors as Gleam will be expecting certain Erlang
applications that may not be present.

### DIFF
--- a/compiler-core/src/build/package_compilation_tests.rs
+++ b/compiler-core/src/build/package_compilation_tests.rs
@@ -2,7 +2,7 @@ use super::*;
 use crate::{
     ast::SrcSpan,
     build::{
-        package_compiler::{DependencyMode, PackageCompiler, Source},
+        package_compiler::{PackageCompiler, Source},
         Origin, Target,
     },
     codegen,
@@ -49,9 +49,10 @@ macro_rules! assert_erlang_compile {
             &root,
             &out,
             &lib,
-            Target::Erlang,
+            &TargetCodegenConfiguration::Erlang {
+                app_file: None,
+            },
             ids,
-            DependencyMode::IncludeDev,
             file_writer,
             Some(&mut build_journal),
         );
@@ -109,9 +110,8 @@ macro_rules! assert_javascript_compile {
             &root,
             &out,
             &lib,
-            Target::JavaScript,
+            &TargetCodegenConfiguration::JavaScript,
             ids,
-            DependencyMode::IncludeDev,
             file_writer,
             Some(&mut build_journal),
         );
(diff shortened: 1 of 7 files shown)


vstreame:
Context: https://discord.com/channels/768594524158427167/811254431051808778/960349024932753408

compiler-core/src/build/package_compiler.rs
```diff
@@ -53,6 +54,7 @@ where
         lib: &'a Path,
         target: Target,
         ids: UniqueIdGenerator,
+        is_project_root: bool,
```
lpil, line 57: The PackageCompiler doesn't have a concept of the root, it doesn't work at that level of abstraction, and in some uses of it (like the `compile-package` CLI) there is no such thing as a root.

Instead let's pass in a `Mode`, or set it as a property. This would be clearer than a boolean too.

lpil:
Fab stuff, thanks for the fix. I've left a comment above

vstreame:
@lpil Updated! Let me know how you feel about the names (I'm terrible at naming things)

compiler-core/src/build/package_compilation_tests.rs
```diff
@@ -40,8 +40,16 @@ macro_rules! assert_erlang_compile {
...
-        let mut compiler =
-            PackageCompiler::new(&config, &root, &out, &lib, Target::Erlang, ids, file_writer);
+        let mut compiler = PackageCompiler::new(
+            &config,
+            &root,
+            &out,
+            &lib,
+            Target::Erlang,
+            ids,
+            DependencyMode::IncludeDev,
```
lpil, line 50: Seeing as this only does anything to Erlang generation what do you think about passing it in as part of the target? Something like this perhaps:

```rust
pub enum TargetConfiguration {
    JavaScript,
    Erlang {
        include_dev_dependencies_in_app_file: bool
    }
}
```

Otherwise I would prefer a more specific name and for it to be a property as below. I think this more general name could imply that it does more than just alter the `.app` file generate.

vstreame: @lpil What do you think of expanding the use cases for this enum? Other things I think we could skip if we knew we weren't the root project:

1. Skip downloading dev dependencies 
2. Skip compiling dev dependencies

Unless you think those _should_ happen when not the root project?

lpil: There's no concept of a root package outside of the context of the `ProjectCompiler`, we wouldn't be able to use it in the `PackageCompiler` as it doesn't make sense there. To it all packages are equal and it knows nothing of what order things are to be compiled in.

lpil (approved):
Thank you E!

PR #1578, opened 2022-04-08
fix(tooling): clear outdated files from the build directory after compilation

mrgleam:
fix issues #1502 

and continue from pull request https://github.com/gleam-lang/gleam/pull/1566 (close)


compiler-core/src/build/package_compiler.rs
```diff
@@ -37,6 +37,7 @@ pub struct PackageCompiler<'a, IO> {
     pub copy_native_files: bool,
     pub compile_beam_bytecode: bool,
     pub silence_subprocess_stdout: bool,
+    pub write_build_journal: bool,
```
lpil, line 40: Rather than have a boolean and also pass in the hashset as an argument we could store it here optionally. If present then we are to write, if not then we do not.

```suggestion
    /// If present any package artefacts written by the compiler will be recorded
    /// in this data structure.
    pub build_journal: HashSet<&'a mut PathBuf>,
```

compiler-core/src/build/project_compiler.rs
```diff
@@ -122,6 +125,8 @@ where
         }
         let result = self.compile_root_package();
 
+        self.check_build_journal();
```
lpil, line 128: Let's check for errors here
```suggestion
        self.check_build_journal()?;
```

lpil (commented):
Fantastic stuff! Very clean. I've added a couple suggestions.

Thanks for waiting, I have been a bit busy!

compiler-core/src/build/package_compiler.rs
mrgleam: A- Do you mean this right?
``` 
pub build_journal: &'a mut HashSet<PathBuf>,
```

lpil: Yes, sorry

lpil (commented):
Wonderful! Would you mind also updating the change log file for me please? Thank you.

mrgleam:
> Wonderful! Would you mind also updating the change log file for me please? Thank you.

@lpil Done! Sorry that I keep forgetting to do that

lpil (approved):
Thanks!!

lpil:
I have rebased these changes onto the main branch. Thank you for waiting

PR #1579, opened 2022-04-08
fix int and float operator hints

inoas:


inoas:
Is there anything I can do to fix please CI?

lpil:
Ah! Great catch. Thank you

You'll need to run `cargo insta review` to update the snapshots. https://insta.rs/

I can do this if you'd like

lpil (approved):
Brilliant, thank you!

PR #1586, opened 2022-04-12
Bitstring float

CrowdHailer:


lpil (approved):
Wow, great stuff! Thank you

PR #1588, opened 2022-04-12
feat: Initial build lock

### LINKED ISSUE #1515: Lock build directory during compilation
To avoid LSP and CLI conflicting.

We want to use an approach to locking where if the binary exits then the lock is released.

We will want to print a message to say that we are waiting to acquire a lock on the build directory if we have to wait.

This library looks useful. https://docs.rs/named-lock/0.1.1/named_lock/struct.NamedLock.html

HarryET:
Fixes #1515


compiler-core/src/build/project_compiler.rs
```diff
@@ -68,6 +70,10 @@ where
             .into_iter()
             .map(|p| (p.name.to_string(), p))
             .collect();
+
+        // TODO can this return a result?
+        let lock = NamedLock::create("gleam-compile")?;
```
HarryET, line 75: This cannot be a result, is there a better way todo it? Does this name mean anything?

compiler-core/src/build/project_compiler.rs
```diff
@@ -112,6 +119,11 @@ where
 
     /// Returns the compiled information from the root package
     pub fn compile(&mut self) -> Result<Package> {
+        let lock_guard = self.lock.try_lock();
+        if(lock_guard.is_err()) {
+            return lock_guard // TODO fix return type
```
HarryET, line 124: Is there a better way to return this? Or am I doing something wrong?

HarryET (commented):
Can I have a hand with these issues?

\cc @lpil

compiler-core/src/build/project_compiler.rs
lpil: Use `lock` rather than `try_lock` :)

compiler-core/src/build/project_compiler.rs
lpil: Perhaps replace `?` with `.expect("Could not lock build directory")`

HarryET:
@lpil looks like it passing tests now but the library used might have a few issues, see: https://github.com/gleam-lang/gleam/runs/6004698233

HarryET:
Looks like this might be another suitable option? https://crates.io/crates/fslock, any comments @lpil

HarryET:
Okay, I have tried a few different libraries and I have tried different ways around it. I am now waiting for comment from @lpil as maybe you have some insight?

deny.toml
```diff
@@ -20,7 +20,7 @@ ignore = [
 # https://embarkstudios.github.io/cargo-deny/checks/licenses/cfg.html
 [licenses]
 unlicensed = "deny"
-allow = ["MIT", "Apache-2.0", "MPL-2.0", "ISC", "OpenSSL", "CC0-1.0", "BSD-3-Clause"]
+allow = ["MIT", "Apache-2.0", "MPL-2.0", "ISC", "OpenSSL", "CC0-1.0", "BSD-3-Clause", "BSD-2-Clause"]
```
HarryET, line 23: @lpil any comment on this?

compiler-core/src/build/project_compiler.rs
```diff
@@ -112,6 +115,11 @@ where
 
     /// Returns the compiled information from the root package
     pub fn compile(&mut self) -> Result<Package> {
+        if cfg!(not(target_arch = "wasm32")) {
```
HarryET, line 118: Doesn't look like this does anything might be worth just removing?

HarryET: Never mind this is now in conjunction with a line in `cargo.toml`

deny.toml
lpil: Looks great!

compiler-core/src/build/project_compiler.rs
```diff
@@ -112,6 +114,9 @@ where
 
     /// Returns the compiled information from the root package
     pub fn compile(&mut self) -> Result<Package> {
+        let lock = NamedLock::create("gleam-compile").expect("Could not lock build directory");
+        let _lock_guard = lock.try_lock().expect("Could not lock build directory");
```
lpil, line 118: Can we use `lock`? That will wait for it to unlock rather than crashing the compiler if it can't lock, which is what `expect` does.

compiler-core/src/build/project_compiler.rs
```diff
@@ -112,6 +114,9 @@ where
 
     /// Returns the compiled information from the root package
     pub fn compile(&mut self) -> Result<Package> {
+        let lock = NamedLock::create("gleam-compile").expect("Could not lock build directory");
```
lpil, line 117: Where does this create the file? I would like it in the `build` directory.

compiler-core/src/build/project_compiler.rs
lpil: Would bringing this back help with wasm compilation?

HarryET: Doesn't look like it. Appears like with all the cfg stuff make it so the package doesn't get compiled but it then complains that stuff doesn't exist as using that if doesn't stop it compiling that code as its a runtime check. Is there a way to-do a compile time check @lpil?

compiler-core/src/build/project_compiler.rs
HarryET: Okay!

compiler-core/src/build/project_compiler.rs
lpil: Oh sorry, I see what you mean now. Got it

compiler-core/src/build/project_compiler.rs
lpil: Ah cool, it puts it in `/tmp`. Sounds good :)

PR #1592, opened 2022-04-15
Bitstring match in JS for integer segments

CrowdHailer:


compiler-core/src/javascript/pattern.rs
```diff
@@ -343,12 +343,31 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
...
+                    self.push_int(index);
+                    self.traverse_pattern(subject, &segment.value)?;
+                    self.pop();
+                }
+                self.pop();
+
+                Ok(())
             }
+            Pattern::VarUsage { location, .. } => Err(Error::Unsupported {
+                feature: "Bit string matching".to_string(),
```
CrowdHailer, line 368: This is the same error as this pattern feature had before. maybe it's not so clear now but I'm not even sure how this would happen so potentially will leave improving error message till a bit later in the process.

compiler-core/templates/prelude.js
```diff
@@ -130,6 +130,14 @@ export class BitString {
   get length() {
     return this.buffer.length;
   }
+
+  atLeastLength(desired) {
+    return desired <= this.length;
```
lpil, line 135: Can we inline these functions in the generated code rather than add these methods? I added them for the list classes mostly just because they're too complicated to inline nicely

CrowdHailer: sure, I created those methods only after I spotted them on the list for similarity.

lpil (approved):
Wonderful! Thank you

PR #1600, opened 2022-04-17
fix(lsp): uri to module name fails on Windows

### LINKED ISSUE #1595: LSP: uri to module name fails on Windows

HarryET:
Fixes #1595 by using a built in method on `Uri` to convert it to a `PathBuf`

### DIFF
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -716,7 +716,7 @@ impl LanguageServer {
 }
 
 fn uri_to_module_name(uri: &Url, root: &Path) -> Option<String> {
-    let path = uri.to_file_path().expect("uri to pathbuf");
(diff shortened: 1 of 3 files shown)


compiler-cli/src/lsp.rs
```diff
@@ -715,6 +717,45 @@ impl LanguageServer {
     }
 }
 
+#[cfg(target_os = "windows")]
+fn uri_to_module_name(uri: &Url, root: &Path) -> Option<String> {
+    let mut uri_path = decode(&*uri.path().replace('/', "\\"))
```
lpil, line 722: What does the decode bit do here?

HarryET: For some reason parts of the path are URL encoded, for example the colon is shown as %3A which causes issues.

PR #1601, opened 2022-04-17
fix(ci): Nightly builds broken

HarryET:
Nightly builds were broken as the delete assets action couldn't find any assets due to first run, this PR also adds a check to make sure it only runs when there have been changes within the past 24hrs


lpil (approved):
Very nice!!

PR #1602, opened 2022-04-18
pattern matching for sized, float and binary segments when compiling to JS

CrowdHailer:


compiler-core/src/javascript/pattern.rs
```diff
@@ -346,22 +374,58 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
...
+                            Ok(())
+                        }
+
+                        [Opt::Size { value: size, .. }] => match &**size {
+                            Pattern::Int { value, .. } => {
+                                let start = offset.bytes;
+                                offset.increment(value.parse::<usize>().unwrap() / 8);
+                                let end = offset.bytes;
+
+                                self.push_method(format!("intFromSlice({}, {})", start, end));
```
CrowdHailer, line 394: this needs to be a custom method to handle transforming the slice to an integer. originally I tried just pushing values to the path. i.e. instead of `path.intFromSlice(1,2)` you'd have `path.buffer.slice(1,2)` which wouldn't require any additions to prelude but the return value from slice is still a utf8 array and not yet an int.

lpil: Makes sense. Could we have a new Index variant per method? Better than passing strings around I think. Might want to rename "Index" to something else too...

CrowdHailer: I don't quite understand what you mean here. do you mean `self.push_intFromSlice(start, end)`? is that for performance?

lpil: I don't want to be Passing strings around when we could use a proper type. You get more help from the compiler if not using strings and it is the pattern that is already being used for the other checks.

compiler-core/templates/prelude.js
```diff
@@ -130,6 +130,23 @@ export class BitString {
...
+    return byteArrayToFloat(this.buffer.slice(index, index + 8))
+  }
+
+  intFromSlice(start, end) {
+    return byteArrayToInt(this.buffer.slice(start, end))
+  }
+
+  restFrom(index) {
+    return new BitString(this.buffer.slice(index))
+  }
```
lpil, line 148: Could we get tests for these? https://github.com/gleam-lang/gleam/blob/main/test/javascript_prelude/main.mjs

And type definitions https://github.com/gleam-lang/gleam/blob/main/compiler-core/templates/prelude.d.ts

Can these functions ever raise an exception?

lpil (commented):
Looking really good! Just a few small things missing for the JavaScript prelude additions.

lpil (approved):

PR #1604, opened 2022-04-18
feat: Named accessors with multiple variants

### LINKED ISSUE #935: Generate accessors for structurally compatible multi-variant custom types
As @CrowdHailer put it
> If you have
> ```rust
> type Foo {
>   A(id: Int)
>   B(id: Int)
>}
>```
>
> Should you be able to call?
> ```rust
> let foo = A(3)
> foo.id
>```

HarryET:
Fixes #935 

## Tests Needed
- [x] Accessor cannot be used if type is different
- [x] Accessor cannot be used if label is different
- [x] Accessor can be used in position other than 1st field
- [x] Accessor can be used in position other than 1st field when 1st field of each variant is different
- [x] Test showing Erlang code generator
- [x] Test showing JavaScript code generation

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -283,6 +283,12 @@ impl TypeAst {
         }
     }
 
+    pub fn compare_without_location(&self, other: TypeAst) -> bool {
+        // TODO write this
+        false
+    }
+
+    // TODO delete this
     pub fn clear_location(&self) -> TypeAst {
         match self {
             TypeAst::Constructor {
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -282,6 +282,41 @@ impl TypeAst {
             | TypeAst::Constructor { location, .. } => *location,
         }
     }
+
+    pub fn clear_location(&self) -> TypeAst {
+        match self {
+            TypeAst::Constructor {
+                module,
+                name,
+                arguments,
+                ..
+            } => TypeAst::Constructor {
+                module: module.clone(),
+                name: name.clone(),
+                arguments: arguments.clone(),
+                location: SrcSpan { start: 0, end: 0 },
+            },
+            TypeAst::Fn {
+                arguments, return_, ..
+            } => TypeAst::Fn {
+                arguments: arguments.clone(),
+                return_: return_.clone(),
+                location: SrcSpan { start: 0, end: 0 },
+            },
+            TypeAst::Var { name, .. } => TypeAst::Var {
+                name: name.clone(),
+                location: SrcSpan { start: 0, end: 0 },
+            },
+            TypeAst::Tuple { elems, .. } => TypeAst::Tuple {
+                elems: elems.clone(),
+                location: SrcSpan { start: 0, end: 0 },
+            },
+            TypeAst::Hole { name, .. } => TypeAst::Hole {
+                name: name.clone(),
+                location: SrcSpan { start: 0, end: 0 },
+            },
+        }
+    }
 }
 
 pub type TypedStatement = Statement<Arc<Type>, TypedExpr, String, String>;
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -296,36 +296,47 @@ impl TypeAst {
                 module,
                 name,
                 arguments,
-                ..
+                location: _,
             } => match other {
                 TypeAst::Constructor {
                     module: o_module,
                     name: o_name,
                     arguments: o_arguments,
-                    ..
+                    location: _,
                 } => module == o_module && name == o_name && arguments == o_arguments,
                 _ => false,
             },
             TypeAst::Fn {
-                arguments, return_, ..
+                arguments,
+                return_,
+                location: _,
             } => match other {
                 TypeAst::Fn {
                     arguments: o_arguments,
                     return_: o_return_,
-                    ..
+                    location: _,
                 } => arguments == o_arguments && return_ == o_return_,
                 _ => false,
             },
-            TypeAst::Var { name, .. } => match other {
-                TypeAst::Var { name: o_name, .. } => name == o_name,
+            TypeAst::Var { name, location: _ } => match other {
+                TypeAst::Var {
+                    name: o_name,
+                    location: _,
+                } => name == o_name,
                 _ => false,
             },
-            TypeAst::Tuple { elems, .. } => match other {
-                TypeAst::Tuple { elems: o_elems, .. } => elems == o_elems,
+            TypeAst::Tuple { elems, location: _ } => match other {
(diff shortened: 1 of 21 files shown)


lpil (commented):
Looking great!

We need a few more tests to cover other edge cases:

- Accessor cannot be used if type is different
- Accessor cannot be used if index is different
- Accessor can be used in position other than 1st field
- Accessor can be used in position other than 1st field when 1st field of each variant is different
- Test showing Erlang code generator
- Test showing JavaScript code generation

compiler-core/src/type_.rs
```diff
@@ -1579,18 +1579,34 @@ fn make_type_vars(
...
+
+    let (first_constructor_arg, constructor_args) = constructor_args_data
+        .split_first()
+        .expect("No constructs with args");
+    let args = first_constructor_arg
+        .iter()
+        .filter(|data_type| {
+            constructor_args.iter().all(|set| {
+                set.iter()
+                    .map(|item| (item.label.clone(), item.type_.clone()))
```
lpil, line 1605: I think we can do this without cloning the label

HarryET: Switched to `.as_ref()`

compiler-core/src/type_.rs
```diff
@@ -1579,18 +1579,34 @@ fn make_type_vars(
...
+        .map(|constructor| constructor.arguments.as_slice())
+        .collect_vec();
+
+    let (first_constructor_arg, constructor_args) = constructor_args_data
+        .split_first()
+        .expect("No constructs with args");
+    let args = first_constructor_arg
+        .iter()
+        .filter(|data_type| {
+            constructor_args.iter().all(|set| {
```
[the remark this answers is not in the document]
HarryET: Renamed

compiler-core/src/type_.rs
```diff
@@ -1579,18 +1579,39 @@ fn make_type_vars(
         .try_collect()
 }
 
-fn custom_type_accessors<A>(
+#[allow(unused_qualifications)]
```
lpil, line 1582: We should never disable the linter, best fix this.

HarryET: Without that I get this:
```
warning: unnecessary qualification                                                                                                                                                                                                     
    --> compiler-core\src\type_.rs:1582:37
     |
1582 | fn custom_type_accessors<A: Clone + std::cmp::PartialEq>(
     |                                     ^^^^^^^^^^^^^^^^^^^
     |
note: the lint level is defined here
    --> compiler-core\src\lib.rs:33:5
     |
33   |     unused_qualifications,
     |     ^^^^^^^^^^^^^^^^^^^^^
   ```
  I then remove it to get this on the next build:
  ```
PS B:\Development\gleam-lang> cargo build
   Compiling gleam-core v0.21.0-rc1 (B:\Development\gleam-lang\compiler-core)
error[E0369]: binary operation `==` cannot be applied to type `&ast::RecordConstructorArg<A>`                                                                                                                                          
    --> compiler-core\src\type_.rs:1617:30
     |
1617 |                         item == &clear_data_type
     |                         ---- ^^ ---------------- &ast::RecordConstructorArg<A>
     |                         |
     |                         &ast::RecordConstructorArg<A>
     |
help: consider further restricting this bound
     |
1582 | fn custom_type_accessors<A: Clone + std::cmp::PartialEq>(
     |                                   +++++++++++++++++++++

For more information about this error, try `rustc --explain E0369`.                                                                                                                                                                    
error: could not compile `gleam-core` due to previous error
```

compiler-core/src/ast.rs
```diff
@@ -282,6 +282,41 @@ impl TypeAst {
             | TypeAst::Constructor { location, .. } => *location,
         }
     }
+
+    pub fn clear_location(&self) -> TypeAst {
```
lpil, line 286: This is expensive, we shouldn't be making a new copy of the TypeAst, instead we should have a comparison function

compiler-core/src/type_.rs
lpil: It's asking you to remove `std::cmp::`, that's the qualification. If you `cmd+.` on it in vscode it'll be able to fix it for you

HarryET: Oh sorry, will try now. I'm currently using CLion and it isn't helping with that 😅

compiler-core/src/ast.rs
HarryET: Fixed to use comparison functions

compiler-core/src/ast.rs
```diff
@@ -282,6 +289,47 @@ impl TypeAst {
...
         }
     }
+
+    pub fn compare_without_location(&self, other: &TypeAst) -> bool {
+        match self {
+            TypeAst::Constructor {
+                module,
+                name,
+                arguments,
+                ..
```
lpil, line 299: Can we replace the `..` with explicit `thingy: _`? I want to be alerted to future fields being added.

compiler-core/src/erlang/tests/records.rs
```diff
@@ -93,6 +93,34 @@ pub fn get_name(person: Person) { person.name }
...
+    // In positions other than the 1st field
+    assert_erl!(
+        "
+pub type Person {
+    Teacher(name: String, age: Int, title: String)
+    Student(name: String, age: Int)
+}
+pub fn get_name(person: Person) { person.name }
+pub fn get_age(person: Person) { person.age }"
+    );
```
lpil, line 121: Could we also have a test like this where the first is different? Thanks

```gleam
pub type Person {
    Teacher(name: Nil, age: Int)
    Student(name: String, age: Int)
}
pub fn get_age(person: Person) { person.age }
```

lpil (commented):
Fantastic work, really loving all the tests. I've left 2 small comments

compiler-core/src/erlang/tests/records.rs
HarryET: For both erlang and js?

compiler-core/src/type_.rs
```diff
@@ -1579,28 +1579,50 @@ fn make_type_vars(
...
+                    .collect_vec()
+                    .is_empty()
+            })
+        })
+        .collect::<Vec<_>>();
 
     let mut fields = HashMap::with_capacity(args.len());
     hydrator.disallow_new_type_variables();
-    for (index, RecordConstructorArg { label, ast, .. }) in args.iter().enumerate() {
+    for (_, (index, RecordConstructorArg { label, ast, .. })) in args.iter().enumerate() {
```
lpil, line 1619:
```suggestion
    for (index, RecordConstructorArg { label, ast, .. }) in args.iter() {
```

lpil (approved):
Brilliant!

Issue #1583, opened 2022-04-09
Feature request(tooling): gleam publish --no-prompt


TanklesXL:
It would be nice to have a flag for promptless publishing behaviour such as `--no-prompt` (or similar) option for the publish command to be able to run the command from within CI pipelines. 
For authentication, the command could accept input directly through its arguments or through env vars (maybe `HEXPM_USER` and `HEXPM_PASS`)

lpil:
Great idea!

TanklesXL:
Well that was fast 😅

TanklesXL:
Also, this might be useful for `gleam docs publish` as well? Not sure if this was covered as I can't find the associated changes 😅

lpil:
Sorry, I must have closed this by mistake.

TanklesXL:
@lpil i think you can mark this completed now :)

lpil:
Thank you

PR #1606, opened 2022-04-19
feat(tooling): env vars for hexpm auth & `--yes/-y` flag for gleam publish

TanklesXL:
Starting work on addressing #1583

Current course of action:
- Add env vars check to `ApiKeyCommand` trait default implementation (This will allow other commands that use this behaviour to get the benefit of this work for free)
- Add a `yes` flag to `gleam publish` to avoid the "are you sure" prompt (i'm open to changing the name of the flag and associated variables)

I noticed `gleam docs remove` doesn't use the `ApiKeyCommand` workflow (yet) and has its own auth setup that is basically identical, might be suitable for a refactor later on or as part of this if I have time :smile:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+- Added the ability to replace a release up to one hour after it is published using `gleam publish --replace`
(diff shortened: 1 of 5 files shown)


lpil (commented):
Great stuff! Would you mind also updating the change log file for me please? Thank you

TanklesXL:
@lpil will do! I was also curious as to where this might be documented as this will affect any command than implements that ApiKeyCommand trait

lpil (approved):
Thank you 💖

lpil (approved):
Thank you!

PR #1610, opened 2022-04-27
fix: Named Accessors, edge case with Optional types

HarryET:
Edge case involving optional types, looks like types are not checked correctly


compiler-core/src/javascript/tests/records.rs
```diff
@@ -53,3 +53,18 @@ pub type Person {
...
+    assert_js!(
+        "
+pub type Person {
+    Teacher(name: String, age: Option(Int), title: String)
+    Student(name: String, age: Option(Int))
+}
+pub fn get_name(person: Person) { person.name }
+pub fn get_age(person: Person) { person.age }"
+    );
+}
```
lpil, line 70: End of file newline please.

In vscode this is `"files.insertFinalNewline": true`, good to have it globally as text files on *nix system should always have them.

HarryET: Fixed

lpil (approved):
Thank Harry!

PR #1612, opened 2022-04-29
fix(nightly build): remove timestamp

TanklesXL:
removing the timestamp from the nightly build filenames

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -106,6 +106,7 @@ jobs:
         shell: bash
         run: |
           VERSION=nightly
+          DAY=$(date +"%y%m%d")
 
           if [ "${{ matrix.os }}" = "windows-latest" ]; then
             ARCHIVE="gleam-$VERSION-${{ matrix.build }}.zip"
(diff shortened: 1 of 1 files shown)


HarryET (commented):
Looks good to me, the only thing I suggest if we aren't going to have it in the archive name or as the executable version we should edit the `Upload release archive` to update it's description. See docs here: https://github.com/softprops/action-gh-release#-customizing

Another comment, maybe it's worth keeping the executable version (in cargo.toml) to avoid conflicts, if we don't want date then maybe use short Git SHA?

.github/workflows/release-nightly.yaml
```diff
@@ -71,7 +71,7 @@ jobs:
       - name: Update versions
         shell: bash
         run: |
-          version=0.0.0-nightly-$(date +"%y%m%d")
+          version=0.0.0-nightly
```
lpil, line 74: Please revert to including the date in the version, so people can tell when they got their release. 🙏

TanklesXL: Yeah that's what @HarryET mentioned as well, I was a little overzealous at first when I went about removing the date from everything 😆

TanklesXL: Does this also apply to the change on line 164?

.github/workflows/release-nightly.yaml
```diff
@@ -161,7 +164,7 @@ jobs:
...
```
HarryET, line 167:
```suggestion
          version=0.0.0-nightly-$(date +"%y%m%d")
```
This line needs to be reverted as you suggested in a reply to @lpil's comment

HarryET (changes requested):

lpil (approved):
Nice!

PR #1627, opened 2022-05-12
Add explicit target during build

### LINKED ISSUE #1626: Wrong CPU type in binary
`asdf` fails with:

```
/Users/varnerac/.asdf/lib/commands/command-exec.bash: line 28: /Users/varnerac/.asdf/installs/gleam/0.21.0/bin/gleam: Bad CPU type in executable
/Users/varnerac/.asdf/lib/commands/command-exec.bash: line 28: /Users/varnerac/.asdf/installs/gleam/0.21.0/bin/gleam: Undefined error: 0
```

When I download the 0.21.0 macOS ARM64 release from GitHub, it looks like it's actually x86_64

```
-rwxr-xr-x@  1 varnerac  staff    14M Apr 24 08:58 gleam
-rw-r--r--@  1 varnerac  staff    14M May 11 23:51 gleam-v0.21.0-macos-arm64.tar
λ ~/Downloads/ file gleam
gleam: Mach-O 64-bit executable x86_64
```

moogle19:
This fixes #1626.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -51,7 +51,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: build
-          args: --release
+          args: --release --target ${{ matrix.target }}
 
       - name: Build archive
         shell: bash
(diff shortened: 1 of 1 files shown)


michallepicki:
Sorry about that! I think the nightly release workflow also needs a similar change. Also maybe the cache key should be changed from v1 to v2 (although last I checked the cache didn't seem to be working for the release workflows, I don't know if that changed since)

lpil (approved):
Aha! Well spotted. Thank you for you help

PR #1628, opened 2022-05-14
Proof of concept: Use Deno as a JavaScript runtime

aslilac:
Not really mergable as is, but is a nice PoC and gives us a spot for discussion

- We probably want a way to specify runtime (semi-independent of build target)
- Deno runs in a strict sandbox by default, which I'm disabling here by passing `-A`. This sandboxing is one of Deno's best features imo, so it'd be great to find a way to preserve it.

A theoretical gleam.toml could look like...

```
target = "javascript"
runtime = "deno"
runtime_options = ["--allow-read=."]
```

### DIFF
--- a/compiler-cli/src/run.rs
+++ b/compiler-cli/src/run.rs
@@ -38,7 +38,7 @@ pub fn command(arguments: Vec<String>, target: Option<Target>, which: Which) -> 
     // Run the command
     let status = match target.unwrap_or(config.target) {
         Target::Erlang => run_erlang(&module, arguments),
-        Target::JavaScript => run_javascript(&config, &module, arguments),
+        Target::JavaScript => run_deno(&config, &module, arguments),
     }?;
 
     std::process::exit(status);
@@ -67,13 +67,13 @@ fn run_erlang(module: &str, arguments: Vec<String>) -> Result<i32, Error> {
         args.push(argument);
     }
 
-    ProjectIO::new().exec("erl", &args, &[], None, false)
+    ProjectIO::new().exec("erl", None, &args, &[], None, false)
 }
 
 fn run_javascript(
     config: &PackageConfig,
     module: &str,
-    arguments: Vec<String>,
+    mut arguments: Vec<String>,
 ) -> Result<i32, Error> {
     let mut args = vec![];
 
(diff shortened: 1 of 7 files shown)


compiler-cli/src/run.rs
```diff
@@ -90,9 +90,33 @@ fn run_javascript(
...
+    args.push("-".into());
+    let script = format!(
+        "import('./{}.mjs').then(module => module.main())",
+        module.to_string_lossy()
+    );
+
+    args.append(&mut arguments);
 
-    ProjectIO::new().exec("node", &args, &[], None, false)
+    ProjectIO::new().exec("deno", Some(script.as_bytes()), &args, &[], None, false)
```
lpil, line 121: Can we pass this program as an argument? I imagine in future we will want this to work `cat somefile > gleam run` so we'll need to be able to use stdin from the program running on Deno

lpil:
Sounds like we had roughly the same ideas!

Currently this is valid Gleam config
```toml
[erlang]
application_start_module = "my_app/application"
extra_applications = ["inets", "ssl"]
```

I could imagine us having something like this

```toml
[javascript]
runtime = "deno"

[javascript.deno]
allow_all = false
allow_read = "."
allow_wibble = true
allow_woo = true
```

Where the default for `javacript.deno` is `{ allow_all = true }`

compiler-cli/src/run.rs
aslilac: We could pass it as an argument if we emitted the bit of code as a file (which I think is a good idea anyway), but I couldn’t find the right spot to do so. 

If we go that route, I feel like we should probably only emit it if the appropriate module actually has a `main` function to run, which I have no idea how to check.

aslilac:
I love the config ideas!

One other thing, `gleam run` should take a `—runtime` flag, instead of `—target` (which can be inferred from the runtime)

compiler-cli/src/run.rs
lpil: There's a file generated for Erlang called `erlang@@main` or something along those lines, we could probably copy how that is created.

> I feel like we should probably only emit it if the appropriate module actually has a main function to run, which I have no idea how to check.

Let's always emit it but have an issue to add a nice error if we detect that the main function is missing.

lpil:
> One other thing, gleam run should take a —runtime flag, instead of —target (which can be inferred from the runtime)

That's an interesting idea! I wonder if people might feel the target flag more approachable. We would need to keep `--target` for `gleam build` and `gleam run`

aslilac:
We would definitely need it for `gleam build`, but I worry if accepting `--runtime` and `--target` for `gleam run` could be confusing

lpil:
Yes, it is also my worry that it could be confusing. I think we would probably want to have runtime take precedence over the other.

@inoas that's right, test and run share all of the same options.

inoas:
> Yes, it is also my worry that it could be confusing. I think we would probably want to have runtime take precedence over the other.

`--runtime` should enforce specific targets (erlang never runs on say spidermonkey) anyway and --target may never(?) be needed (exist for legacy reasons and if specified sets --runtime to defaults such as nodejs or beam)

aslilac:
> if specified sets `--runtime` to defaults

It should obey whatever setting you use for `javascript.runtime` in gleam.toml, and *then* use a default :)

lpil:
Closing for now, feel free to reopen!

PR #1639, opened 2022-05-20
feat: preserve empty lines between comments when formatting

### LINKED ISSUE #1621: Formatter: Preserve spaces between comments
As @inoas put it:

1. Idempotency of adding and removing simple comments: If I add and remove `//` in a long test with multiple pipes the formatting I gave it is then removed, as new lines do not get comments between them. I have to manually insert `//` between two outcommented blocks of code to keep the spacing which is tedious. Also new `//` blocks try to glue to the next statement.

```gleam
  value
  |> io.debug

  value
  |> dynamic.from
  |> io.debug

  value
  |> dynamic.from
  |> dynamic.classify
  |> io.debug
```
becomes

```gleam
  // value
  // |> io.debug
  // value
  // |> dynamic.from
  // |> io.debug
  value
  |> dynamic.from
  |> dynamic.classify
  |> io.debug
```

becomes

```gleam
  value
  |> io.debug
  value
  |> dynamic.from
  |> io.debug
  value
  |> dynamic.from
  |> dynamic.classify
  |> io.debug
```

https://github.com/gleam-lang/gleam/discussions/1619#discussioncomment-2708716

superhawk610:
This PR modifies the formatter to keep (at most) one empty line between standard `//` comments. This allows "commenting out" expressions to be idempotent, i.e. comment, format, uncomment will leave the source meaningfully unchanged. For example:

```gleam
value
|> io.debug

value
```

when commented/formatted becomes

```gleam
// value
// |> io.debug

// value
```

and reverts to the original when uncommented/formatted.

Closes #1621.

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -397,8 +397,7 @@ impl<'comments> Formatter<'comments> {
             Some(_) => concat(Itertools::intersperse(
                 comments.map(|c| match c {
                     Some(c) => "///".to_doc().append(Document::String(c.to_string())),
-                    // empty lines have been dropped by pop_doc_comments
-                    None => unreachable!(),
+                    None => unreachable!("empty lines dropped by pop_doc_comments"),
                 }),
                 line(),
             ))
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -1508,35 +1508,33 @@ fn printed_comments<'a, 'comments>(
 
     let mut doc = Vec::new();
     while let Some(c) = comments.next() {
-        if let Some(c) = c {
-            doc.push("//".to_doc().append(Document::String(c.to_string())));
-            match comments.peek() {
-                // next line is a comment
-                Some(Some(_)) => doc.push(line()),
-                // next line is empty
-                Some(None) => {
-                    let _ = comments.next();
-                    match comments.peek() {
-                        Some(_) => doc.push(lines(2)),
-                        None => {
-                            if trailing_newline {
-                                doc.push(force_break());
-                                doc.push(lines(2));
-                            }
+        // There will never be consecutive empty lines (None values),
+        // and whenever we peek a None, we advance past it.
+        let c = c.expect("no consecutive empty lines");
+        doc.push("//".to_doc().append(Document::String(c.to_string())));
+        match comments.peek() {
+            // Next line is a comment
+            Some(Some(_)) => doc.push(line()),
+            // Next line is empty
+            Some(None) => {
+                let _ = comments.next();
+                match comments.peek() {
+                    Some(_) => doc.push(lines(2)),
+                    None => {
+                        if trailing_newline {
+                            doc.push(force_break());
+                            doc.push(lines(2));
                         }
                     }
                 }
-                None => {
-                    if trailing_newline {
-                        doc.push(force_break());
-                        doc.push(line());
-                    }
+            }
+            // We've reached the end, there are no more lines
+            None => {
+                if trailing_newline {
+                    doc.push(force_break());
+                    doc.push(line());
                 }
             }
-        } else {
-            // there will never be consecutive empty lines (None values),
-            // and whenever we peek a None, we advance past it
-            unreachable!()
         }
     }
     Some(concat(doc))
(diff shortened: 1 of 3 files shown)


compiler-core/src/format.rs
```diff
@@ -1487,22 +1500,54 @@ fn list<'a>(elems: Document<'a>, tail: Option<Document<'a>>) -> Document<'a> {
...
+                        doc.push(line());
+                    }
+                }
+            }
+        } else {
+            // there will never be consecutive empty lines (None values),
+            // and whenever we peek a None, we advance past it
+            unreachable!()
+        }
+    }
```
superhawk610, line 1541: This is a bit gnarly, but I wasn't able to simplify any further without introducing whitespace on empty lines. Perhaps someone more familiar with the formatting approach could recommend a better way?

compiler-core/src/format.rs
```diff
@@ -1578,20 +1623,48 @@ where
...
+        .iter()
+        .position(|l| *l > limit)
+        .unwrap_or(empty_lines.len());
+    let popped_comments = comments
+        .get(0..end_comments)
+        .unwrap()
         .iter()
-        .map(|c| c.content);
+        .map(|c| (c.start, Some(c.content)));
+    let popped_empty_lines: Box<dyn Iterator<Item = _>> = if retain_empty_lines {
```
superhawk610, line 1647: Using a `Box<dyn Iterator>` feels very wrong here, but I couldn't figure out another way to use `std::iter::empty()` and make the call to `merge_by` reusable. I'm still pretty fresh with Rust, so there's probably something cleaner I'm missing 🙂

compiler-core/src/format.rs
```diff
@@ -386,7 +395,11 @@ impl<'comments> Formatter<'comments> {
         match comments.peek() {
             None => nil(),
             Some(_) => concat(Itertools::intersperse(
-                comments.map(|c| "///".to_doc().append(Document::String(c.to_string()))),
+                comments.map(|c| match c {
+                    Some(c) => "///".to_doc().append(Document::String(c.to_string())),
+                    // empty lines have been dropped by pop_doc_comments
+                    None => unreachable!(),
```
lpil, line 401:
```suggestion
                    None => unreachable!("Unexpected empty lines found for trailing doc comments"),
```

compiler-core/src/format.rs
lpil: We could replace the `if let Some(c) = c {` with `c.expect("consecutive empty lines")`, that'd remove one level of indentation.

compiler-core/src/format.rs
```diff
@@ -1487,22 +1500,54 @@ fn list<'a>(elems: Document<'a>, tail: Option<Document<'a>>) -> Document<'a> {
...
+                        Some(_) => doc.push(lines(2)),
+                        None => {
+                            if trailing_newline {
+                                doc.push(force_break());
+                                doc.push(lines(2));
+                            }
+                        }
+                    }
+                }
+                None => {
```
lpil, line 1529: Could we get a comment above this branch, to match the others.

compiler-core/src/format.rs
```diff
@@ -1487,22 +1500,54 @@ fn list<'a>(elems: Document<'a>, tail: Option<Document<'a>>) -> Document<'a> {
...
-        comments.map(|c| "//".to_doc().append(Document::String(c.to_string()))),
-        line(),
-    )))
+
+    let mut doc = Vec::new();
+    while let Some(c) = comments.next() {
+        if let Some(c) = c {
+            doc.push("//".to_doc().append(Document::String(c.to_string())));
+            match comments.peek() {
+                // next line is a comment
```
lpil, line 1514:
```suggestion
                // Next line is a comment
```

compiler-core/src/format.rs
```diff
@@ -1487,22 +1500,54 @@ fn list<'a>(elems: Document<'a>, tail: Option<Document<'a>>) -> Document<'a> {
...
+                Some(Some(_)) => doc.push(line()),
+                // next line is empty
```
lpil, line 1516:
```suggestion
                // Next line is empty
```
Just me being fussy :)

compiler-core/src/format.rs
```diff
@@ -1578,20 +1623,48 @@ where
...
-    let popped = comments
-        .get(0..end)
-        .expect("Comments before slicing popped")
+    let end_empty_lines = empty_lines
+        .iter()
+        .position(|l| *l > limit)
+        .unwrap_or(empty_lines.len());
+    let popped_comments = comments
+        .get(0..end_comments)
+        .unwrap()
```
lpil, line 1644: I try to avoid unwrap. Is there a default we could use here?

compiler-core/src/format.rs
```diff
@@ -1578,20 +1623,48 @@ where
...
         .iter()
-        .map(|c| c.content);
+        .map(|c| (c.start, Some(c.content)));
+    let popped_empty_lines: Box<dyn Iterator<Item = _>> = if retain_empty_lines {
+        Box::new(
+            empty_lines
+                .get(0..end_empty_lines)
+                .unwrap()
```
lpil, line 1651: And for this one too

compiler-core/src/format.rs
lpil: Agreed. I'm not sure what is best here.
Could we move the iterator methods out of the `if else` and have something like this

```rust
if retain_empty_lines {
    empty_lines
} else {
   &[] 
}
    .get(.....) etc
```

If that doesn't work maybe a static slice would help
```rust
static EMPTY_USIZE_SLICE: &'static [usize] = &[];
```

lpil (commented):
This looks fantastic! Thank you. I've added some inline notes, nothing major.

Would you mind updating the changelog too?

superhawk610:
Thanks for the code review, that cleaned things up nicely! I updated the changelog too 🌈

PR #1641, opened 2022-05-20
fix: don't output newline at top of docs that only contain comments

### LINKED ISSUE #1640: `gleam format` adds unwanted empty line to docs that contains only comments
The formatter will add an empty line at the top of a document that contains only simple `//` comments.

For example, this

```
// foo
// bar
```

formats to

```

// foo
// bar
```

_Originally posted by @superhawk610 in https://github.com/gleam-lang/gleam/discussions/1619#discussioncomment-2788049_

superhawk610:
This fixes an issue where the formatter would introduce an unwanted newline at the top of documents that only contained simple `//` comments. For example:

```gleam
// foo
// bar
```

This will no longer have a newline added at the top.

This PR introduces two new functions, `Document::is_empty` and `join`. `is_empty` recursively checks all members of a document and returns true when they contain no printable characters; this is used to determine whether or not newlines should be appended/prepended when joining formatted module sections together. `join` is a simple wrapper for `concat(Itertools::intersperse(..), separator`, an idiom I noticed was used pretty frequently around the formatter. I introduced it in a separate commit, so it can be reverted if it's undesirable 😇

Closes #1640.

### DIFF
--- a/compiler-core/src/pretty.rs
+++ b/compiler-core/src/pretty.rs
@@ -16,6 +16,8 @@
 #[cfg(test)]
 mod tests;
 
+use itertools::Itertools;
+
 use crate::{io::Utf8Writer, Result};
 
 #[macro_export]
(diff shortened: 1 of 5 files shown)


compiler-core/src/pretty.rs
```diff
@@ -315,4 +324,20 @@ impl<'a> Document<'a> {
...
+        use Document::*;
+        match self {
+            Line(n) => *n == 0,
+            ForceBreak => true,
+            String(s) => s.is_empty(),
+            Str(s) => s.is_empty(),
+            // assuming `broken` and `unbroken` are equivalent
+            Break { broken, .. } => broken.is_empty(),
+            FlexBreak(d) | Nest(_, d) | NestCurrent(d) | Group(d) => d.is_empty(),
+            Vec(docs) => docs.iter().all(|d| d.is_empty()),
```
lpil, line 340: I think this will return false for an empty ven when we want it to be true. Prob worth having a test for that

lpil (commented):
Great stuff! I like the new helper function a lot

lpil (approved):
AH, I was wrong. Thank you for making those tests to show me.

Fab work! 💜

Issue #1633, opened 2022-05-16
Some doc-strings are not included in the docs


NicklasXYZ:
It seems like doc-strings are not included in the docs if both an erlang and javascript block is included after the function that is being documented. I am not sure this is actually the cause but it seems to be the case when I try to build the docs for the following module:

Docs:
https://hexdocs.pm/gleam_stats/gleam_stats/math.html

Related source code: 
https://github.com/nicklasxyz/gleam_stats/blob/v1.0.0/src/gleam_stats/math.gleam#L611-L611

lpil:
Thank you

NicklasXYZ:
https://github.com/gleam-lang/gleam/pull/1651 seems to have fixed the issue :)

PR #1651, opened 2022-05-31
Sort statements before generating docs

tynanbe:
Fixes cases where doc comments are separated from their associated statements.

This bug can be reproduced most notably with type alias documentation.

Possibly addresses #1633.


lpil (approved):
Thank you!

PR #1659, opened 2022-06-10
CI: Change nightly schedule, retry asset uploads

tynanbe:
Attempts to address the nightly build intermittently failing to upload the Linux archive with error `ECONNRESET`.

- Runs all builds at `00:45`
- Uses ~~a workaround~~ `bash` and the `hub` cli tool to try the upload a maximum of 5 times. ~~(https://github.com/softprops/action-gh-release/issues/187#issuecomment-1028216200)~~

Haven't seen the retry functionality tested, but the build is still able to succeed with this change: ~~https://github.com/tynanbe/gleam/actions/runs/2472937671~~
https://github.com/tynanbe/gleam/actions/runs/2478743818

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -2,7 +2,7 @@ name: release-nightly
 
 on:
   schedule:
-    - cron: '0 0 * * *'
+    - cron: '45 0 * * *'
 
 env:
   CARGO_TERM_COLOR: always
@@ -15,7 +15,7 @@ jobs:
     outputs:
       should_run: ${{ steps.should_run.outputs.should_run }}
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: print latest_commit
         run: echo ${{ github.sha }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/release-nightly.yaml
```diff
@@ -225,3 +227,58 @@ jobs:
             ${{ env.ASSET }}
             ${{ env.ASSET }}.sha256
             ${{ env.ASSET }}.sha512
+      - name: Upload release archive (retry)
```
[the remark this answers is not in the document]
tynanbe: Hmm, not sure. The `name` fields aren't needed, just the `id` fields. That could get rid of a bit of the excess.

tynanbe:
@lpil, updated. I also cleaned up line endings, so make sure to ignore whitespace when viewing the diff.

.github/workflows/release-nightly.yaml
```diff
@@ -1,227 +1,241 @@
...
+            ./gleam.exe --version
+          elif [ "${{ matrix.build }}" = "macos-arm64" ]; then
+            echo We cannot test an ARM binary on a AMD64 runner
+          else
+            tar -xvzf "$ASSET"
+            ./gleam --version
+          fi
+
+      - name: Upload release archive
+        shell: bash
```
lpil, line 142: How does this work if hub hasn't been installed yet?

tynanbe: Hub is included with the standard github images.

lpil (approved):
Thank you!

PR #1670, opened 2022-06-25
fix-todo

### LINKED ISSUE #1669: todo warnings are being reported as on line 1 rather than actual location

inoas:
Closes #1669

### DIFF
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -1335,11 +1335,9 @@ fn unknown_label() {
 #[test]
 fn todo_warning_test() {
     assert_warning!(
-        "fn main() {
-          1 == todo
-        }",
+        "fn main() { 1 == todo }",
         Warning::Todo {
-            location: SrcSpan { start: 27, end: 31 },
+            location: SrcSpan { start: 17, end: 31 },
             typ: Arc::new(Type::Var {
                 type_: Arc::new(RefCell::new(TypeVar::Link { type_: int() })),
             }),
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -1337,7 +1337,7 @@ fn todo_warning_test() {
     assert_warning!(
         "fn main() { 1 == todo }",
         Warning::Todo {
-            location: SrcSpan { start: 17, end: 31 },
+            location: SrcSpan { start: 17, end: 21 },
             typ: Arc::new(Type::Var {
                 type_: Arc::new(RefCell::new(TypeVar::Link { type_: int() })),
             }),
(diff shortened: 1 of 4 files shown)


inoas:
```
// https://github.com/gleam-lang/gleam/issues/1669
#[test]
fn todo_test() {
    // Implicitly discarded Results emit warnings
    assert_warning!(
        "
fn main() {
  todo
}",
        Warning::Todo {
            location: SrcSpan { start: 0, end: 50 }
        }
    );
}
```
=> `missing structure fields: - typ`

inoas:
```
    Todo {
        location: SrcSpan,
        typ: Arc<Type>,
    },

    ImplicitlyDiscardedResult {
        location: SrcSpan,
    },
```

What is this typ(e) doing for Todo?

inoas:
So the change for the test does not test if the output contains a valid reference to the line number...

compiler-core/src/type_/tests.rs
```diff
@@ -1331,12 +1331,15 @@ fn unknown_label() {
...
 #[test]
 fn todo_warning_test() {
     assert_warning!(
-        "fn main() { 1 == todo }",
+        "fn main() {
+          1 == todo
+        }",
         Warning::Todo {
-            location: SrcSpan { start: 17, end: 21 },
+            location: SrcSpan { start: 27, end: 31 },
```
michallepicki, line 1342: I think this file has the typechecker tests, where the location was still correct. It was being lost sometime later, I don't think this test is calling `to_diagnostic` anywhere? Maybe there are some other tests for the LSP

inoas: I have fixed the location with the change here I think https://github.com/gleam-lang/gleam/pull/1670/files#diff-69d4f3e128a4feb5c69856b9c1754c7d6cc0a2be8a9fcaa7e7b5e682cb965a84L33

But the tests do not test for the warning string to reflect the location.

before:

Screen Shot 2022-06-25 at 19 08 06

after:

Screen Shot 2022-06-25 at 19 08 38

lpil: This test is unrelated to the problem here. Could you revert these changes please?

lpil (commented):
Thank you for the fix! Could you also include a snapshot test to ensure that this does not happen again. Thank you

inoas:
> Thank you for the fix! Could you also include a snapshot test to ensure that this does not happen again. Thank you

I want to but don't yet know how to. I would want to test against the warning message's string.

compiler-core/src/type_/tests.rs
inoas: Yes!

lpil (approved):
Brilliant, thank you!

Issue #1671, opened 2022-06-27
Formatter: case guards loses curley braces even if required for manual precedence


inoas:
```gleam
import gleam/io

pub fn main() {
  let is_enabled = False
  let is_admin = True
  let is_confirmed = False

  let test1 = is_enabled && { is_confirmed || is_admin }
  let test2 = is_enabled && is_confirmed || is_admin

  io.debug(test1) // False
  io.debug(test2) // True
 
  let is_enabled = False
  let is_admin = True
  let is_confirmed = False

  case is_enabled, is_admin, is_confirmed {
    // `{` and `}` will be stripped on save for me, on the next line, changing the logic to the one reflected in the next/followup case clause.
    is_enabled, is_admin, is_confirmed if is_enabled && { is_confirmed || is_admin } ->
      io.println("enabled and (admin and/or confirmed)")
    is_enabled, is_admin, is_confirmed if is_enabled && is_confirmed || is_admin ->
      io.println("(enabled and confirmed) and/or admin")
    _, _, _ -> Nil
  }

}
```

inoas:
It seems during let assigns the formatter does the right thing, as it removes `{}` from `let test2 = { is_enabled && is_confirmed } || is_admin`.

lpil:
Oh that is rubbish! Let's get this fixed ASAP

PR #1672, opened 2022-06-27
Formatter: case guards loses curley braces even if required for precedence (just failing test)

inoas:
For now just test cases for https://github.com/gleam-lang/gleam/issues/1671


inoas:
I won’t have time to dig into Rust enough to fix this in forseeable future

PR #1676, opened 2022-06-30
Switch TLS backend to rustls

tranzystorek-io:
Not sure if this needs some additional testing for certs etc.


lpil:
Rad!

What do you think about using rustls-tls-native-roots? I think that would mean using rust-tls but getting cacerts from the host. I like the sound of that as the certs won't become outdated if people use older Gleam versions in future.

tranzystorek-io:
Sure, sounds reasonable!

lpil (approved):
Thank you 💜

Issue #1677, opened 2022-07-01
`gleam update` command to update versions of dependencies


lpil:
Throw away the manifest and re-resolve the deps.

PR #1684, opened 2022-07-10
gleam update command

lpil:


inoas:
Closes https://github.com/gleam-lang/gleam/issues/1677

PR #1687, opened 2022-07-11
Fix: Improve hint when binding or rebinding variables omitting let.

### LINKED ISSUE #1580: Improve hint when binding or rebinding variables omitting `let`
Attempts of binding or rebinding variables without the `let` keyword should print a more obvious error message:

```gleam
pub fn main() {
 foo = 42
}
```

```
error: Syntax error
  ┌─ ./src/main.gleam:2:6
  │
2 │  foo = 42
  │      ^ I was not expecting this.

Expected one of: "}"
```

See: <https://johndoneth.github.io/gleam-playground/?s=A4VwRgBAZgdhC2BDAljAFASggbwFDQHsCIBeCAFgCZcBfIA%3D>

Ref: https://discord.com/channels/768594524158427167/814516237173784607/961748042481041409

Kn99HN:
- Closes #1580

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -71,6 +71,13 @@ and can contain a-z, 0-9, or _.",
 contain a-z, A-Z, or 0-9.",
                 )],
             ),
+            ParseErrorType::InvalidAssignmentStatement => (
+                "This is not a valid assignment statement.",
+                vec![
+                    "Hint: Use let for binding".to_string(),
+                    "See: https://gleam.run/book/tour/let-bindings".to_string(),
+                ],
+            ),
             ParseErrorType::InvalidBitStringSegment => (
                 "This is not a valid BitString segment option.",
                 vec![
@@ -200,6 +207,7 @@ pub enum ParseErrorType {
     ExtraSeparator,          // #(1,,) <- the 2nd comma is an extra separator
     IncorrectName,           // UpName or DiscardName used when Name was expected
     IncorrectUpName,         // Name or DiscardName used when UpName was expected
+    InvalidAssignmentStatement,
     InvalidBitStringSegment, // <<7:hello>> `hello` is an invalid bitstring segment
     InvalidBitStringUnit,    // in <<1:unit(x)>> x must be 1 <= x <= 256
     InvalidTailPattern,      // only name and _name are allowed after ".." in list pattern
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -30,6 +30,7 @@ pub struct ParseError {
 impl ParseError {
     pub fn details(&self) -> (&'static str, Vec<String>) {
         match &self.error {
+            ParseErrorType::ExpectedEqual => ("I was expecting a '=' after this", vec![]),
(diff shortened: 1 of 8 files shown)


inoas:
I also noticed that this error message isn't optimal:

```gleam
pub fn main() {
  let foo
}
```

```
error: Syntax error
  ┌─ ./src/main.gleam:3:1
  │
3 │ }
  │ ^ I was not expecting this.

Expected one of: "="
```

https://johndoneth.github.io/gleam-playground/?s=A4VwRgBAZgdhC2BDAljAFASggbwFAQgBsBTAF2gHsLcBfIA%3D

I got to this when thinking non-gleam:

1. initialize variable
2. rebind


```gleam
// invalid pseudo code

let foo
foo = 5
```

compiler-core/src/parse/tests.rs
```diff
@@ -232,3 +232,15 @@ fn anonymous_function_labeled_arguments() {
         }
     );
 }
+
+#[test]
+fn missing_let_binding() {
+    assert_error!(
+        "let bar = 42;
```
inoas, line 239: AFAIK the semicolons are tolerated but removed by the formatter and shouldn't be there, but I could be wrong.

Kn99HN: Gotcha.

Kn99HN:
> "="

Thanks for the comment. I also just realized that the case when `foo: Int = 32` wouldn't return an error. I'll look more into it

Kn99HN:
> I also noticed that this error message isn't optimal:
> 
> ```gleam
> pub fn main() {
>   let foo
> }
> ```
> 
> ```
> error: Syntax error
>   ┌─ ./src/main.gleam:3:1
>   │
> 3 │ }
>   │ ^ I was not expecting this.
> 
> Expected one of: "="
> ```
> 
> https://johndoneth.github.io/gleam-playground/?s=A4VwRgBAZgdhC2BDAljAFASggbwFAQgBsBTAF2gHsLcBfIA%3D
> 
> I got to this when thinking non-gleam:
> 
>     1. initialize variable
> 
>     2. rebind
> 
> 
> ```gleam
> // invalid pseudo code
> 
> let foo
> foo = 5
> ```

So I think in this case, the error message is not wrong but it is not the most accurate location. So it can be confusing. My approach is to make it more explicit error by relying on pattern.location(). Please let me know your thoughts on this.

compiler-core/src/parse/error.rs
```diff
@@ -211,6 +220,7 @@ pub enum ParseErrorType {
     NoConstructors,            // A type "A {}" must have at least one constructor
     NoCaseClause,              // a case with no clauses
     NoExpression, // between "{" and "}" in expression position, there must be an expression
+    NoLetBinding, // Bindings and rebinds always require let and must always bind to a value.
```
Kn99HN, line 223: Rename "InvalidAssignmentStatement" to "NoLetBinding" for more precise error.

Kn99HN:
> 

Do you want the snapshot test be in the same file as `parse/test.rs`?

lpil:
That would be good, although if there is not already a test macro for parse errors feel free to put it in one of the test modules that does have an error macro.

lpil (approved):
Thank you! Great work!

PR #1688, opened 2022-07-13
fix:  Empty functions are currently syntax errors

### LINKED ISSUE #1584: Empty functions are currently syntax errors
Today, if you write the code

```gleam
pub fn foo() {
}
```

inside of a module you will get the following error back from the Gleam compiler:

```
error: Syntax error
  ┌─ ./src/foo.gleam:6:1
  │
6 │ }
  │ ^ I was not expecting this.

Expected one of: The body of a function
```

I think this can be confusing, especially for new Gleam developers. Instead I think we should do one of two things:

1. Not have this be a Syntax error
2. Continue to be a Syntax error, but with a more informative error message

**Option 1: No longer a syntax error**

If we go down this path I think instead of being a Syntax error we can make this be a function that returns `Nil`. This would def be a special circumstance and be the only time that Gleam isn't WYSIWYG, but I think an empty body `== Nil` would also make sense for developers.

**Option 2: A more informative error message**

If instead we to continue to enforce function bodies cannot be empty, then I would suggest we change the error message a bit. Instead of the current error message, it would be great if we could be a bit more informative and include a suggestion.

 ```
error: Syntax error
  ┌─ ./src/foo.gleam:6:1
  │
5│ pub fn foo() {
6│ }
  │ ^ I was not expecting this.

Function bodies cannot be empty in Gleam. Maybe you would like to add a "todo" instead? Example:

pub fn foo() {
  todo
}
```

Or a message sort of like that with a solution a new developer can copy/paste.

### PR DESCRIPTION
**Pull request by Kn99HN**
- Insert `ToDo` in case of an empty function.
- Closes #1584

### CODE REVIEW

File: compiler-core/src/parse.rs [hunk 1]
```diff
@@ -1270,9 +1271,18 @@ where
...
                 return_annotation,
+            })),
+            None =>  Ok(Some(Statement::Fn {
+                doc: None,
+                location: SrcSpan { start, end },
+                end_position: rbr_e - 1,
+                public,
+                name,
+                arguments: args,
+                body: UntypedExpr::Todo { location: SrcSpan {start, end}, label: None },
```
Kn99HN, line 1282: I wasn't sure what to put for the location so I chose to use `start` and `end` here. Lmk if there are better options.

compiler-core/src/type_/snapshots/gleam_core__type___tests__empty_func_warning_test.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: compiler-core/src/type_/tests.rs
+expression: "pub fn main() { foo() }\n        pub fn foo() { }"
```
inoas, line 3:
```gleam
pub fn main() { }
```

... should probably be sufficient for the testcase + snapshot and should return:

```
pub fn main() {
  todo
}
```
... I guess.

inoas:
Thanks for helping out. I am just a contributor like you, no authority. So plase do not feel discouraged in any way!

compiler-core/src/type_/snapshots/gleam_core__type___tests__empty_func_warning_test.snap
Kn99HN: Thanks, I will make a change.

I'm not sure how to make it return `pub fn main() {
  todo
}` since the `assert_warning` macro returns the source code and we are modifying the AST. Would you mind elaboring on that?

lpil (approved):
Looks fab! Thank you!

PR #1690, opened 2022-07-15
fix:  Formatter: case guards loses curley braces even if required for manual precedence

### LINKED ISSUE #1671: Formatter: case guards loses curley braces even if required for manual precedence

Kn99HN:
- Closes #1671 

### Bug Description
For let binding such as: 
```
let test1 = is_enabled && { is_confirmed || is_admin }
```
The formatter calls `binop` function, which compare the precedence of each sub-expression. Therefore, it doesn't strip braces in the case above.

For case guard 
```
...
case is_enabled, is_admin, is_confirmed {
    is_enabled, is_admin, is_confirmed if is_enabled && { is_confirmed || is_admin } -> ...
```
The formatter calls `clause_guard` function. But the function doesn't compare precedence of each sub-expression. Therefore, it strips braces away.

### Fix

My fix is to add an implementation of `precedence` for `ClauseGuard` struct. Then in `clause_guard` function, we can rely on `operator_side` function, which compares precedence and handles the braces accordingly.

Thank you @inoas for the test cases! Super helpful and clear to understand the bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
   empty lines in a row.
 - Fixed a bug alternative patterns with a clause containing a pipe with a pipe
   after the case expresson could render incorrect Erlang.
+- Fixed a bug where formatter would strip curly braces around case guards due 
+to lack of precedence.
 
 ## v0.22.1 - 2022-06-27
 
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -873,6 +873,28 @@ impl<A, B> ClauseGuard<A, B> {
             | ClauseGuard::LtEqFloat { location, .. } => *location,
         }
     }
+
+    pub fn precedence(&self) -> u8 {
+        match self {
+            ClauseGuard::Or { .. } => 1,
+            ClauseGuard::And { .. } => 2,
+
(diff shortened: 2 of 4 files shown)


lpil (approved):
Fantastic work, thank you so much! Would you mind also adding an entry to the change log file?

lpil (approved):
Thank you!

PR #1694, opened 2022-07-18
Fix changelog

inoas:
Ref: https://github.com/gleam-lang/gleam/pull/1690

@Kn99HN unless I am mistaken, that's the change?


lpil (approved):
Thanks!

PR #1701, opened 2022-07-25
support 'equals' method in JS for object equality

schurhammer:
If the object has a .equals method, use that to determine equality.
See discussion in gleam-lang/stdlib/pull/328

### DIFF
--- a/compiler-core/templates/prelude.js
+++ b/compiler-core/templates/prelude.js
@@ -316,8 +316,11 @@ export function isEqual(x, y) {
       unequalSets(a, b);
     if (unequal) return false;
 
-    if (typeof a.equals === "function" && a.equals(b)) {
-      continue;
+    const proto = Object.getPrototypeOf(a)
+    if (proto !== null && typeof proto.equals === "function") {
+      try {
+        if (a.equals(b)) continue; else return false;
+      } catch {}
     }
 
     let [keys, get] = getters(a);
--- a/compiler-core/templates/prelude.js
+++ b/compiler-core/templates/prelude.js
@@ -316,6 +316,10 @@ export function isEqual(x, y) {
       unequalSets(a, b);
     if (unequal) return false;
 
+    if (typeof a.equals === "function" && a.equals(b)) {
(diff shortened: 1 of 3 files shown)


lpil (commented):
Great stuff!

Could you wrap this in a try/catch? Gleam should never crash.

Perhaps we also want to check that it is on the prototype, so an equals property does not trigger it.

Could we also have some tests here? https://github.com/gleam-lang/gleam/blob/main/test/javascript_prelude/main.mjs

Thanks

compiler-core/templates/prelude.js
```diff
@@ -316,6 +316,10 @@ export function isEqual(x, y) {
       unequalSets(a, b);
     if (unequal) return false;
 
+    if (typeof a.equals === "function" && a.equals(b)) {
+      continue;
```
lpil, line 320: I think we want to return the value of `a.equals(b)` here

schurhammer: Ah yea, I think its if false return, if true continue? Since there could be more values to check on the stack. Does that sound right?

schurhammer:
> Could you wrap this in a try/catch? Gleam should never crash.

Yea sure, I though it was overkill haha. From my testing try catch is wayyy slower though (like 1000x, no kidding), but mostly only if it actually throws something.

> Perhaps we also want to check that it is on the prototype, so an equals property does not trigger it.

We could do that. We don't do it for inspect though, and some people use normal objects instead of prototype (for example see https://github.com/sveltejs/svelte/blob/master/src/runtime/store/index.ts#L108)

compiler-core/templates/prelude.js
hayleigh-dot-dev: Yeah if we return true early then in
```
{ x: { equals () { … } }
, y: “foo”
}
```
we’d never check `y` for equality, seems like the wrong behaviour to me. 

We don’t return true early for identity equality I don’t see why this would be any different!

lpil (approved):

PR #1706, opened 2022-07-31
Display module qualifiers when type errors would be otherwise unclear

### LINKED ISSUE #107: Display module qualifiers when type errors would be otherwise unclear

Kn99HN:
- Closes #107 

### Notes
- I decided to choose this option to avoid changing API call to print to include a `HashSet` of repeated types.


lpil (approved):
Brilliant! Thank you very much.

And thanks for waiting while I catch up with Gleam PRs :)

PR #1710, opened 2022-08-05
Fix root_folder validation

moogle19:
If given a `--name`, `gleam new` would validate the name as `PROJECT_ROOT` and not the actual root, 
e.g. `gleam new --name foo bar`, wouldn't work when a folder `foo` is present in the current directory.

This should fix the validation.


lpil (approved):
Thank you

PR #1711, opened 2022-08-06
Use Prettier to format CSS and JS files

lucasavila00:
Unfortunately, the HTML templates do not format well, so I skipped them.


lpil (changes requested):
Can you remove the CI check please? I don't want to be any dependency on node modules in order to do Gleam development. It would raise the barrier to entry and anything we can do to encourage new contributors is worth doing

lucasavila00:
Sure, will do. It's understandable. When first installing the repository I felt it.

Having a [devcontainer](https://code.visualstudio.com/docs/remote/containers) would help on that by the way.

lucasavila00:
A devcontainer would also support running the project in Github Codespaces.

As an example: https://github.com/ChrisMarinos/FSharpKoans#running-with-github-codespaces

https://github.com/ChrisMarinos/FSharpKoans/blob/fsharp/.devcontainer/devcontainer.json

lpil (approved):
Thank you!

lpil:
A devcontainer sounds cool! We currently have something like this with gitpod but more could be good

PR #1712, opened 2022-08-07
Draft: Add built-in search to generated documentation

lucasavila00:
It works, styling is not finished though. A demo is live https://deploy-gleam-docs.web.app/gleam/regex.html#Options

I wonder how we should do the mobile version of it...

Please pardon the horrible rust code, I hadn't used rust in 5 years.

### DIFF
--- a/compiler-core/src/docs.rs
+++ b/compiler-core/src/docs.rs
@@ -13,6 +13,8 @@ use crate::{
 };
 use askama::Template;
 use itertools::Itertools;
+use serde::Serialize;
+use serde_json::to_string as serde_to_string;
 
 const MAX_COLUMNS: isize = 65;
 const VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -47,6 +49,8 @@ pub fn generate_html(
 
     let mut files = vec![];
 
+    let mut search_indexes = vec![];
+
     let modules_links: Vec<_> = modules
         .clone()
         .map(|m| {
@@ -79,6 +83,15 @@ pub fn generate_html(
             path: PathBuf::from(&page.path),
             text: temp.render().expect("Page template rendering"),
         });
+
+        search_indexes.push(SearchIndex {
+            doc: config.name.to_string(),
+            title: config.name.to_string(),
+            content,
+            // TODO: use full URL
+            url: page.path.to_string(),
+            rel_url: page.path.to_string(),
+        })
     }
 
     // Generate module documentation pages
@@ -94,6 +107,8 @@ pub fn generate_html(
             unnest = ".".to_string();
         }
 
+        let page_title = format!("{} - {}", name, config.name);
+
         let template = ModuleTemplate {
             gleam_version: VERSION,
             unnest,
@@ -102,7 +117,7 @@ pub fn generate_html(
             documentation: render_markdown(&module.ast.documentation.iter().join("\n")),
             modules: &modules_links,
             project_name: &config.name,
-            page_title: &format!("{} - {}", name, config.name),
+            page_title: &page_title,
             module_name: name,
             project_version: &config.version.to_string(),
             functions: module
@@ -134,9 +149,25 @@ pub fn generate_html(
                 .render()
                 .expect("Module documentation template rendering"),
         });
+
+        search_indexes.push(SearchIndex {
+            doc: module.name.to_string(),
+            title: page_title.to_string(),
+            // TOODO: get text content, not HTML
+            content: template.to_string(),
+            // TODO: get full url
+            url: format!("{}.html", module.name),
+            rel_url: format!("{}.html", module.name),
+        })
     }
 
     // Render static assets
+
+    files.push(OutputFile {
+        path: PathBuf::from("search-data.json"),
+        text: serde_to_string(&search_indexes).expect("search index serialization"),
+    });
+
     files.push(OutputFile {
         path: PathBuf::from("index.css"),
         text: std::include_str!("../templates/index.css").to_string(),
@@ -400,4 +431,12 @@ struct ModuleTemplate<'a> {
     types: Vec<Type<'a>>,
     constants: Vec<Constant<'a>>,
     documentation: String,
+}
+#[derive(Serialize, PartialEq, Eq, PartialOrd, Ord, Clone)]
+struct SearchIndex {
+    doc: String,
+    title: String,
+    content: String,
+    url: String,
+    rel_url: String,
 }
--- a/compiler-core/src/docs.rs
+++ b/compiler-core/src/docs.rs
@@ -88,9 +88,9 @@ pub fn generate_html(
             doc: config.name.to_string(),
             title: config.name.to_string(),
             content,
-            // TODO: use full URL
-            url: page.path.to_string(),
-            rel_url: page.path.to_string(),
+            // TODO: get full url
+            url: format!("/{}", page.path),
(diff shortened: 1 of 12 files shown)


inoas:
This is awesome and will help with discover and learning by docs a lot especially while more LSP work is underway.

lpil:
This is really cool! I am loving the demo. 💜

Could we remove the dependency on node modules please? It is nice to have the code formatted but I think currently mandating it adds a bit too much friction to the workflow for new contributors.

There's already quite a lot of things that need to be installed and I want to keep it as easy as possible

lucasavila00:
> Could we remove the dependency on node modules please? It is nice to have the code formatted but I think currently mandating it adds a bit too much friction to the workflow for new contributors.

Will do.

>Is there a way to make it so that string.append comes up first if you search for that or string append?

Yeah. It doesn't work now because there is no string.append anywhere in the codebase.
But we can add {module_name}.{export_name} to the search index, thus making it work.

lucasavila00:
I think it is ready to be reviewed once I fix the lint issues.

There is one TODO left, which I think should be done in another MR:

"properly parse markdown and remove it's syntax (not the replace approach)"

It will require adding a new dependency https://github.com/kivikakk/comrak and a custom printer. Maybe it is better not to do it?

lucasavila00:
Also, I did not document that there is a new setting in gleam.toml, a "homepage" that configures the deployment URL.
It is required if the documentation is not at the root of the domain. It happens when deploying to github pages, for instance.

compiler-core/src/docs.rs
```diff
@@ -196,6 +327,16 @@ fn function<'a>(
     }
 }
 
+fn text_documentation(doc: &Option<String>) -> String {
+    let raw_text = doc
+        .as_ref()
+        .map(|it| it.to_string())
+        .unwrap_or_else(|| "".to_string());
+
+    // TODO: parse markdown properly and extract the text nodes
```
lucasavila00, line 336: I think this MR should stop here and leave this TODO to the future.
As I said earlier, it will require adding a new dependency such as https://github.com/kivikakk/comrak and it will increase the scope of this MR too much.

compiler-core/templates/index.css
```diff
@@ -708,3 +733,343 @@ body.theme-dark .pride .blue {
     width: var(--sidebar-toggle-size);
   }
 }
+
+/* Search */
+
+.search {
```
lucasavila00, line 739: The CSS is pretty much the same as just-the-docs'.
I think it could be changed a bit to be better aligned with the current site styling, however I'd appreciate some help with this part :)

lpil:
Looking really great! ✨ 

On the mobile mode when I enter text and then delete it the search input is de-focused and I have to click it again. Could we keep it focused?

The desktop search input looks a little off to me at the moment. There's not enough margin and too much padding on the left of the input, and the magnifying glass doesn't align with anything.

Perhaps the magnifying glass should go on the right side? Or maybe we could use the mobile design on all viewport sizes.

lucasavila00:
> We cannot do this via config as the documentation can be hosted anywhere, or opened from the file system. All URLs are currently relative and the search would need to work this way too.

It is impossible to create a link to it, if hosted on github pages. A relative link on a page https://lucasavila00.github.io/sql-select-ts/ will delete the project name suffix from the link.

This setting is not mandatory, the default value for it makes the app use relative URLs.
However, if one wants to use github pages to host the site, they will need to use a custom domain or this setting.

lucasavila00:
Perhaps "homepage" setting should be called `baseurl`, as in JTD? https://github.com/just-the-docs/just-the-docs/blob/main/_config.yml#L18

lucasavila00:
> On the mobile mode when I enter text and then delete it the search input is de-focused and I have to click it again. Could we keep it focused?

That's fixed.

> The desktop search input looks a little off to me at the moment. There's not enough margin and too much padding on the left of the input, and the magnifying glass doesn't align with anything.

> Perhaps the magnifying glass should go on the right side? Or maybe we could use the mobile design on all viewport sizes.

I like the idea of using the mobile design on all viewport sizes. Less code to maintain. But I'm not sure if having it be fullscreen would be the best UX :thinking:

inoas:
> Notice that the lack of margin on the input is due to the fact the search input tries to match the width and position of the rendered content.

I'd not be concerned about that too much.

mobile:
- looks great

desktop:

- here long lib and version names will break the current layout and/or shorten the library name with elipsis, i'd flexbox align right a button on the max-width of sidebar+content with the spotlight/search icon and Search LIBNAME and on hit replace the whole header with a search bar

alternatively something like this:

Screen Shot 2022-08-07 at 21 29 34

lpil:
> Notice that the lack of margin on the input is due to the fact the search input tries to match the width and position of the rendered content.

Sorry, I'm not sure what you mean there.

Here's a video that might show what I mean

https://user-images.githubusercontent.com/6134406/183759579-74e72d6d-efa0-43f3-8652-1ea4d73bb913.mov

When not focused there is extra padding to the left of the icon, while it looks correct when focused.

To the left of the input as a whole there is not enough margin to match the earlier gaps in the navbar.

lucasavila00:
> When not focused there is extra padding to the left of the icon, while it looks correct when focused.

Yeah, I'm aware. I was mostly worried about the general design. I'll make this padding go away.

> To the left of the input as a whole there is not enough margin to match the earlier gaps in the navbar.


That's what I meant.There is no "margin", the design uses these "columns". We could increase the padding on the left column of the navbar, but it will make gleam_std name appear with ellipsis.

image


Personally I think the sidebar width should be increased, padding added to the left column of the navbar, the font of `gleam_stdlib` reduced, and the magnifying glass icon static to the left - but I didn't do it at first as it'd be a big re-design of the current site. I'm not good with design and the current one looks very good :p

lpil:
We cannot constrain the first part of the titlebar to the width of the navbar, even on the largest size it is too small to fit a fair normal package name at the moment, such as with the stdlib.

image

I think having the search box aligned to the right or to use the full width like on mobile may be the best options.

lucasavila00:
> We cannot constrain the first part of the titlebar to the width of the navbar, even on the largest size it is too small to fit a fair normal package name at the moment, such as with the stdlib.
> 
> image
> 
> I think having the search box aligned to the right or to use the full width like on mobile may be the best options.

Ok. I think we should remove all desktop styling and have it be un-responsive, mobile all the time.

lucasavila00:
I did not make it use the mobile version, but a different approach to maintain the desktop one.

It's not perfect but might be good enough?

inoas:
Screen Shot 2022-08-11 at 10 28 35

I agree that on desktop the sidebar should be larger, it should also be collapsable, and should probably collapse when searching, that way small width desktop isn't affected.

But personally I am still fine with the change as is, the gain is much larger with the PR as is and responsive layout styling optimization can be done in followup PRs by you, others (or me).

lucasavila00:
> I agree that on desktop the sidebar should be larger, it should also be collapsable, and should probably collapse when searching, that way small width desktop isn't affected.
> 

Thanks for the picture, that collapsing of the name hadn't happened to me.
Collapsing it is a good idea :thinking:

lucasavila00:
By the way, I wouldn't mind if anybody were to take it over from here :)

lpil:
Roger. I will pick this one up to spare you the CSS bikeshedding 👍  Thanks @lucasavila00 !! Love this addition

PR #1715, opened 2022-08-07
Do not permit multiple imports to use the same name

### LINKED ISSUE #1696: Do not permit multiple imports to use the same name
```gleam
import gleam/erlang/process.{...}
import gleam/otp/process.{...}
```

lucasavila00:
Closes #1696


compiler-core/src/type_/tests.rs
```diff
@@ -1678,6 +1678,85 @@ fn unused_imported_module_warnings_test() {
...
+        Origin::Src,
+        "thepackage",
+        &modules,
+        &mut warnings,
+    );
+
+    assert!(infer_result.is_err());
+    assert_eq!(
+        Error::DuplicateImport {
+            location: SrcSpan { start: 33, end: 43 },
```
lucasavila00, line 1752: Is there a way to test this "e2e"? So that we can check that the spans are correct and so on.

lpil: You could create a module error test using the macro in the type tests modules.

https://github.com/gleam-lang/gleam/blob/2e27150eade0bf192753079059eb88bbd48ce29b/compiler-core/src/type_/tests/errors.rs#L1699-L1701

compiler-core/src/type_.rs
```diff
@@ -1920,6 +1920,25 @@ pub fn register_import(
             }
 
             if unqualified.is_empty() {
+                let imported_name = module
```
lpil, line 1923: Do we not need to do this when `unqualified` is not empty?

lucasavila00: Yeah, we need it. Should be fixed.

compiler-core/src/type_/tests/errors.rs
```diff
@@ -105,7 +105,7 @@ macro_rules! assert_error {
 }
 
 macro_rules! assert_with_module_error {
-    ($(($name:expr, $module_src:literal)),+, $src:expr $(,)?) => {
+    (($name:expr, $module_src:literal), $src:expr $(,)?) => {
```
lucasavila00, line 108: I tried fixing the variadic macro but I couldn't. I unrolled it and implemented the new tests.

compiler-core/src/type_/tests.rs
```diff
@@ -1663,6 +1663,24 @@ fn unused_variable_warnings_test() {
     assert_no_warnings!("pub fn a(b) { case b { #(c, _) -> c } }");
 }
 
+#[test]
```
lucasavila00, line 1666: There was a test on the JS suite that showed this was an accepted syntax.
I re-created the test in compiler core, and the need for ImportedNameSrcSpan is because of this.

lpil: Do you think we should force them to merge the two imports into one?

lucasavila00: No. IMHO this is a linter's job.

For instance, I use a linter to prevent me from doing it in an important .py file, but I regularly import stuff like this example in jupyter notebook or in interactive mode.

lpil: I think different rules would likely apply in a future interactive mode so we can put that aside for now. These rules will be only for regular source files.

compiler-core/src/type_/tests/errors.rs
```diff
@@ -1706,3 +1763,45 @@ fn ambiguous_type_error() {
...
+        ),
+        "
+        import foo/sub
+        import foo2/sub.{bar}
+        pub fn main() {
+            sub.bar()
+        }
+        ",
+    );
+}
```
lpil, line 1807: One last thing, could we have a test for these? They should all error

```rust
import one
import two as one
```

```rust
import one as x
import two as x
```

```rust
import one.{fn1} as x
import two.{fn2} as x
```

compiler-core/src/type_/tests.rs
lucasavila00: `make test` fails because it can't compile stdlib https://github.com/gleam-lang/stdlib/blob/main/src/gleam/dynamic.gleam#L3-L4

lucasavila00: Should this syntax be valid? :thinking:

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__same_imports_multiple_times_3.snap
```diff
@@ -0,0 +1,15 @@
...
+source: compiler-core/src/type_/tests/errors.rs
+expression: "\n        import one as x\n        import two as x\n        "
+---
+error: Duplicate import
+  ┌─ /src/one/two.gleam:2:16
+  │
+2 │         import one as x
+  │                ^^^ First imported here
+3 │         import two as x
+  │                ^^^ Reimported here
```
lucasavila00, line 11: It'd be better if the highlight was on the "x", but this needs to be fixed outside of the type checker.

compiler-core/src/type_/tests.rs
lpil: Good find. I've released a new version of the stdlib with that mistake corrected. If you rebase on main you will get the fix

lucasavila00: I did. Now it happens on https://github.com/gleam-lang/stdlib/blob/main/src/gleam/string.gleam#L8-L9

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__same_imports_multiple_times_3.snap
lpil: Good shout, I've fixed this on main. Please rebase and you'll get the changes

lpil:
I'm sorry but the merge commit will need to be removed, the history in this repo is linear to aid with viewing the history. If you rebase on main you'll get the updates and it'll remove the merge commit. Let me know if you would like me to do this

lucasavila00:
>I'm sorry but the merge commit will need to be removed, the history in this repo is linear to aid with viewing the history. If you rebase on main you'll get the updates and it'll remove the merge commit. Let me know if you would like me to do this.

No, that's coo. I'll rebase.

lpil:
Looks like there's more. I'll build this branch locally to find them all.

lpil:
OK, slightly sloppy there by me, I forgot to commit the test file. It should all be ready for you now on main.

lpil (approved):
Brilliant! Thank you very much 💜

PR #1720, opened 2022-08-09
Draft: Improvements to typescript codegen

lucasavila00:


lpil (commented):
Interesting, I wonder what happened here. Perhaps we broke something...

Do you have more you want to do here or should I review thoroughly?

lucasavila00:
> Interesting, I wonder what happened here. Perhaps we broke something...
> 
> Do you have more you want to do here or should I review thoroughly?

The implementation is not really solid, and the changes I made should be reflected in function names and so on.
I created the tests and basically brute forced a fix that did not break anything else, but I don't trust my reasoning enough to say this is an improvement...

Another thing that made it difficult for me was the compile + exec flow of rust, and the rust language itself. I would love if it was possible to build this typescript translation in gleam, not rust :D

lpil:
Me too! Unfortunately I think we are a long way away from being able to write native Gleam code and link it from the compiler

lucasavila00:
> Me too! Unfortunately I think we are a long way away from being able to write native Gleam code and link it from the compiler

A wild thought, but I think there should be a Gleam interpreter written in Rust - this same interpreter would allow const-evaluated functions

```
const fn double(a) {a *2}
const abc = double(123)
```

And then, for some parts of the compiler where performance is not deemed important, as in TS codegen, it'd be written in Gleam.

lpil:
An interpreter would be very handy!

I would consider anything in the compile loop (such as TS generation) to be something where we care about performance a lot so it'd need to be fast in that instance. Could be useful for various tooling commands though.

Another option could be to start a long lived instance of the VM that we can send jobs to over the network.

inoas:
That interpreter could also then be used for a REPL?

lpil:
It could, though we wouldn't be able to do anything that involves FFI

compiler-core/src/javascript/typescript.rs
```diff
@@ -90,11 +90,10 @@ fn collect_generic_usages<'a>(
 fn generic_ids(type_: &Type, ids: &mut HashMap<u64, u64>) {
```
lucasavila00, line 90: This generic_ids function is super hacky.

It would be better if the type information was "calculated" upstream and the typescript file would only print it.

The same type representation would be required if we are to compile to any type-checked language aside from typescript.

lpil:
Is this ready to review? Thanks

compiler-core/src/javascript/typescript.rs
lpil: What do you mean by hacky? Where would you like this calculated instead?

lucasavila00:
> Is this ready to review? Thanks

For now I fear I won't be able to improve the MR, so it is ready for review.

lpil (approved):
Thank you!

PR #1722, opened 2022-08-10
Serialize build modules & Expose function to get serialized modules in wasm

lucasavila00:
After building with `wasm-pack build --target no-modules`

This page can console.log the full build ast.

```html
<html>
  <head>
    <meta content="text/html;charset=utf-8" http-equiv="Content-Type" />
  </head>
  <body>
    <script src="./gleam_wasm.js"></script>

    <script>
      const initialSource = `
pub fn main() {
  42
}`;
      const { init, get_modules, compile } = wasm_bindgen;

      async function run() {
        await wasm_bindgen("./gleam_wasm_bg.wasm");

        const result = init();
        console.log(result);

        const args = {
          target: "javascript",
          sourceFiles: {
            "./src/main.gleam": initialSource,
          },
          dependencies: [],
          mode: "Prod",
        };
        console.error(compile(args));
        console.error(get_modules(args));
      }

      run();
    </script>
  </body>
</html>
``` 

image

I'd like to have a tool like this to help with compiler development https://ts-ast-viewer.com/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2151,6 +2151,9 @@ name = "vec1"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5fc1631c774f0f9570797191e01247cbefde789eebfbf128074cb934115a6133"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "version_check"
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -26,7 +26,7 @@ lazy_static = "1.4.0"
 # Levenshtein string distance for typo suggestions
 strsim = "0.10.0"
 # Data (de)serialisation
-serde = { version = "1.0.130", features = ["derive"] }
+serde = { version = "1.0.130", features = ["derive", "rc"] }
 # Cap'n Proto binary format runtime
 capnp = "0.14.3"
 # Enum trait impl macros
@@ -36,7 +36,7 @@ askama = "0.10.5"
 # Markdown parsing
 pulldown-cmark = { version = "0.8.0", default-features = false }
 # Non-empty vectors
-vec1 = "1.8.0"
+vec1 = { version = "1.8.0", features = ["serde"] }
 # Hex package manager client
 hexpm = "2.0.0"
 # XDG directory locations
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -16,6 +16,7 @@ use std::sync::Arc;
 
 #[cfg(test)]
 use pretty_assertions::assert_eq;
+use serde::Serialize;
 
 pub const TRY_VARIABLE: &str = "_try";
 pub const PIPE_VARIABLE: &str = "_pipe";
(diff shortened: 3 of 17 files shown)


lpil:
What might this be used for? I don't want to make the AST a public interface as it has very frequent breaking changes

lucasavila00:
> I don't want to make the AST a public interface as it has very frequent breaking changes

I get it. The focus of such feature is to allow creation of tools that will help compiler development and interest in general. It should not be treated as a public API by no means - Personally I don't mind re-building the compiler with the patch on the wasm part, but it would be nice if the serialization part was merged.

> What might this be used for?

Some examples that comes to my mind are https://ts-ast-viewer.com/  and https://ts-morph.com/ .

I'm also working on hacking some random stuff with it, like this gleam-written "macro" or "ppx" to auto-build gleam type decoders from serde types

image

image


Super meta, this thing is building itself...

image

lpil:
That's so cool!

If you're only interested in custom type definitions we could perhaps expose an API for that? Otherwise I think it would be frustrating maintaining tools that this proposed API as they would break with every Gleam release.

lucasavila00:
> as they would break with every Gleam release

It's a part of the deal. Typescript exposes this kind of stuff with a big disclaimer, for instance. https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#disclaimer


> If you're only interested in custom type definitions we could perhaps expose an API for that?

Ideally, I'd like to have access to the parser -> ast -> pretty-printed flow. And also the entire type checking data. Basically the output of build module, the output of parser, and the ability to send the parser output (or my own ast) to the pretty printer.

I get the point of breaking changes eventually causing issues. Maybe anything related to this should be locked behind a build flag?

lpil:
We definitely want to have formatting in the playgrounds. At some point in the future we will if that's an official project and add a bunch of features, including that.

I think we could have this AST feature in future but we're just not quite stable enough for it I think. If we include it we are putting it forward for people to use those people would likely become burnt out on trying to keep up with all the different versions of Gleam. If we didn't put it out those people would potentially be working on something else with a more stable foundation which would benefit the community more overall.

Code generation something I'm very interested in for the language so I think we want to put a bunch of consideration into the interface as well, rather than adding something quickly without a good grasp of what the long term implications might be.

lucasavila00:
> We definitely want to have formatting in the playgrounds. At some point in the future we will if that's an official project and add a bunch of features, including that.
> 
> I think we could have this AST feature in future but we're just not quite stable enough for it I think. If we include it we are putting it forward for people to use those people would likely become burnt out on trying to keep up with all the different versions of Gleam. If we didn't put it out those people would potentially be working on something else with a more stable foundation which would benefit the community more overall.
> 
> Code generation something I'm very interested in for the language so I think we want to put a bunch of consideration into the interface as well, rather than adding something quickly without a good grasp of what the long term implications might be.
> 
> I think I would be more in favour of doing what Rust does and exposing the lexer but not the parser. This would be much more stable

Thanks for the explanation. I'm lost in the terminology but I'll do some research on it.

lpil:
Do ask any questions! I've probably not made myself very clear

PR #1725, opened 2022-08-12
LSP: Don't find node on function call, do find node on assignment, find node when de-structuring

lucasavila00:
It's more what the Rust LSP does...

I constantly mis-hover a function argument, and the LSP shows the call.
I constantly try to create a variable and hover it to inspect the type, but the assignment does not has hover.

This MR fixes both.

Is `find_node` used for important stuff or can it be changed to please the LSP UX?

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -703,7 +703,7 @@ impl CallArg<TypedExpr> {
 }
 
 impl CallArg<Pattern<PatternConstructor, Arc<Type>>> {
-    pub fn find_node(&self, byte_index: usize, typ: &Type) -> Option<Located<'_>> {
+    pub fn find_node<'a>(&'a self, byte_index: usize, typ: &'a Type) -> Option<Located<'a>> {
         self.value.find_node(byte_index, typ)
     }
 }
@@ -1061,7 +1061,7 @@ impl<A, B> Pattern<A, B> {
 }
 
 impl Pattern<PatternConstructor, Arc<Type>> {
-    pub fn find_node(&self, byte_index: usize, typ: &Type) -> Option<Located<'_>> {
+    pub fn find_node<'a>(&'a self, byte_index: usize, typ: &'a Type) -> Option<Located<'a>> {
         if !self.location().contains(byte_index) {
             return None;
         }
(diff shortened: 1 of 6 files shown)


lpil:
> Is find_node used for important stuff or can it be changed to please the LSP UX?

It's only used by LSP and is safe to change. The LSP is pretty basic and rough currently :)

compiler-core/src/ast.rs
```diff
@@ -1053,6 +1059,48 @@ impl<A, B> Pattern<A, B> {
...
+        if !self.location().contains(byte_index) {
+            return None;
+        }
+
+        match self {
+            Pattern::Int { .. }
+            | Pattern::Float { .. }
+            | Pattern::String { .. }
+            // TODO: How not to clone the type?
+            | Pattern::Var { .. } => Some(Located::Pattern(self, typ.clone())),
```
lucasavila00, line 1074: How not to clone the type?

compiler-core/src/build.rs
```diff
@@ -185,6 +186,7 @@ impl Module {
 #[derive(Debug, Clone, PartialEq)]
 pub enum Located<'a> {
     Expression(&'a TypedExpr),
+    Pattern(&'a TypedPattern, Type),
```
lucasavila00, line 189: What's the best approach for the lifetime of Type? Should it be a reference?
I ran into issues here...

compiler-core/src/build.rs
```diff
@@ -183,8 +184,9 @@ impl Module {
 }
 
 #[derive(Debug, Clone, PartialEq)]
-pub enum Located<'a> {
```
lucasavila00, line 186: Having this use references creates issues with stuff that are in RefCells.
So far, it created issues with the `Type` of `Pattern(TypedPattern, Type)`.
Is it a big deal having copies here?

lpil: I would prefer not to copy. Where is it that we have the Rc problem? I couldn't see it at a glance

lucasavila00: https://github.com/gleam-lang/gleam/blob/cd3798f68aee1fb094c9768593e42519548fc460/compiler-core/src/ast.rs#L1079-L1091


Specifically, the link inside `TypeVar::Link { type_ }`

PR #1726, opened 2022-08-12
Fix lint warnings from Clippy 1.63.0

lucasavila00:
Fixes this rule https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

It was added today, and the CI job uses the latest version.

That's why https://github.com/gleam-lang/gleam/pull/1725 is red.


lpil (approved):
Thank you

Issue #1738, opened 2022-08-26
Add support for building Elixir Mix deps


lpil:
So that Gleam projects can use Elixir libraries.

Here's some useful info from tsloughter: https://github.com/Supersonido/rebar_mix/issues/29

Tagging @tynanbe. Looks like we didn't have an issue for this!

tynanbe:
Oh, looking at PR #1484, looks like we did have issue #1408.

I've just pushed an update to my old `mix` branch after rebasing on `main` yesterday, and all seems working on initial testing.

lpil:
Ah, well found. Thanks

PR #1741, opened 2022-08-27
Build dependency packages managed with Mix (take 2)

### LINKED ISSUE #1408: Elixir dependency support
Gleam should be able to fetch and build Elixir deps within Gleam projects.

tynanbe:
Closes #1408 and #1738.

Updated from previous PR to use the method shown in https://github.com/Supersonido/rebar_mix/issues/29 for compiling mix packages (thanks @tsloughter).

Includes an improved and more efficient method of loading core Elixir libraries.

### DIFF
--- a/compiler-core/src/build/project_compiler.rs
+++ b/compiler-core/src/build/project_compiler.rs
@@ -345,7 +345,9 @@ where
         let project_dir = paths::build_deps_package(name);
         let mix_build_dir = project_dir.join("_build").join(mix_target);
         // Absolute build path is needed for mix to make accurate symlinks
-        let mix_build_path = std::env::current_dir()
+        let mix_build_path = self
+            .io
+            .current_dir()
             .expect("Project root")
             .join(&mix_build_dir);
         let mix_build_lib_dir = mix_build_dir.join("lib");
@@ -413,17 +415,28 @@ where
     }
 
     fn maybe_link_elixir_libs(&mut self, build_dir: &PathBuf) -> Result<(), Error> {
-        let elixir_libs = ["elixir", "logger", "mix"];
+        // These Elixir core libs will be loaded with the current project
+        // Each should be linked into build/{target}/erlang if:
+        // - It isn't already
+        // - Its ebin dir doesn't exist (e.g. Elixir's path changed)
+        let elixir_libs = ["eex", "elixir", "logger", "mix"];
 
+        // The pathfinder is a file in build/{target}/erlang
+        // It contains the full path for each Elixir core lib we need, new-line delimited
+        // The pathfinder saves us from repeatedly loading Elixir to get this info
         let mut update_links = false;
         let pathfinder_name = "gleam_elixir_paths";
         let pathfinder = build_dir.join(pathfinder_name);
         if !self.io.is_file(&pathfinder) {
+            // The pathfinder must be written
+            // Any existing core lib links will get updated
             update_links = true;
             // TODO: test
             let env = [("TERM", "dumb".into())];
+            // Prepare the libs for Erlang's code:lib_dir function
             let elixir_atoms: Vec<String> =
                 elixir_libs.iter().map(|lib| format!(":{}", lib)).collect();
+            // Use Elixir to find its core lib paths and write the pathfinder file
             let args = [
                 "--eval".into(),
                 format!(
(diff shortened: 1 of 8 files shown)


compiler-core/src/build/project_compiler.rs
```diff
@@ -328,6 +329,156 @@ where
...
+        if target != Target::Erlang {
+            tracing::info!("skipping_mix_build_for_non_erlang_target");
+            return Ok(());
+        }
+
+        let build_dir = paths::build_packages(mode, target);
+        let project_dir = paths::build_deps_package(name);
+        let mix_build_dir = project_dir.join("_build").join(mix_target);
+        // Absolute build path is needed for mix to make accurate symlinks
+        let mix_build_path = std::env::current_dir()
```
lpil, line 348: We'll need to pass in this path from the `compiler-cli` package as the `compiler-core` doesn't know how to perform side effects like interact with the file system. This is so it can be compiled to WASM

Alternatively we could add a `current_dir()` to `self.io`

compiler-core/src/build/project_compiler.rs
```diff
@@ -328,6 +329,156 @@ where
...
+        ];
+        let args = [
+            "-pa".into(),
+            mix_path(&ebins),
+            "-S".into(),
+            "mix".into(),
+            "compile".into(),
+            "--no-deps-check".into(),
+            "--no-load-deps".into(),
+            "--no-protocol-consolidation".into(),
```
lpil, line 389: Do we ever consolidate protocols?

compiler-core/src/build/project_compiler.rs
```diff
@@ -328,6 +329,156 @@ where
...
+        } else {
+            Err(Error::ShellCommand {
+                program: "mix".to_string(),
+                err: None,
+            })
+        }
+    }
+
+    fn maybe_link_elixir_libs(&mut self, build_dir: &PathBuf) -> Result<(), Error> {
+        let elixir_libs = ["elixir", "logger", "mix"];
```
lpil, line 416: I think we probably want eex also?

compiler-core/src/build/project_compiler.rs
```diff
@@ -328,6 +329,156 @@ where
...
+            Ok(())
+        } else {
+            Err(Error::ShellCommand {
+                program: "mix".to_string(),
+                err: None,
+            })
+        }
+    }
+
+    fn maybe_link_elixir_libs(&mut self, build_dir: &PathBuf) -> Result<(), Error> {
```
lpil, line 415: Could we get some comments in this function explaining what it does at each step and why we need to do it? To make it easier for newcomers and also myself in a couple years time!

lpil (commented):
Beautiful job! Really great work, I'm excited to see this.

I've left a couple v small comments inline

lpil:
I got this when building plug with a new version of Elixir

```
louis ~/app (main ?) $ gleam add plug
  Resolving versions
Downloading packages
 Downloaded 6 packages in 0.04s
      Added plug v1.13.6
louis ~/app (main ?) $ gleam build
  Resolving versions
  Compiling gleam_stdlib
  Compiling mime
Shall I install Hex? (if running non-interactively, use "mix local.hex --force") [Yn] ** (Mix) Could not find an SCM for dependency :ex_doc from MIME.Mixfile
error: Shell command failure
```

Can we force these prompts to be accepted?

tynanbe:
> Can we force these prompts to be accepted? We can also leave this for later if it's not easy. The error message has the solution in it so not a huge deal

We could maybe run `mix local.hex --force` when we link in the Elixir core libs?

Then again, not sure there's a safe way to do it from CLI. Would need to test. We probably don't want to eagerly force something like that without any knowledge of the environment.

Perhaps we'd be able to do it via a small Elixir script?

lpil:
> We probably don't want to eagerly force something like that without any knowledge of the environment.

Yup, you're right. Let's leave it as it is

compiler-core/src/build/project_compiler.rs
tynanbe: See #1743

lpil (approved):
Cracking work!! Thank you!

tynanbe:
@lpil, thanks for the invaluable assistance and quick followup~

Issue #1730, opened 2022-08-17
Bundle all JS libraries with generated docs so that they can be used offline


inoas:
When distributing libraries we could/should distribute the full generated documentation, so that using hexdocs.pm is optional / no internet connection is required.

> > Are there plans to ever make the generated docs available locally? If so would it be wise to start bundling docs assets locally such as https://github.com/gleam-lang/gleam/pull/1712/files#diff-35e517b7b2494a23576f20256e6a64bd9b258e2f0e089b2752d9f2eb4ae346e0R294
>
> They are
> gleam docs build
> Oh like the assets
> Yeah good idea! Could you make an issue?

Ref: https://discord.com/channels/768594524158427167/811254431051808778/1009422223712452668

Note: The current generated docs use CDN'ed assets and future docs should ship with local assets.

Further ideas:

- We could also generate a meta docs file that allows searching all locally installed deps' docs and/or also just lists them in a tree to click the doc and get to that.
- Deps that require other deps could als be crosslinked in their HTML documentation.

lpil:
Great suggestion! Thank you

aflatter:
@lpil FWIW I started to play around with this. I added a `templates/index.js` that references the files that are currently loaded via CDN and that is then built using `esbuild` into `templates/dist/index.js` and `templates/dist/index.css`. Is that an approach you'd be okay with given that it requires `npm` and `esbuild` during the build/CI process?

The alternative would be to a) store copies of all assets in the `gleam` repo and manually handle updates (wouldn't recommend) b) run `npm install` and/or `esbuild` and commit the resulting `node_modules` and/or built assets to the `gleam` repo.

lpil:
Fantastic, thank you.

I would like to save the modified versions of each library into the repository and then have a link in the HTML to each one. We definitely do not want to depend upon npm or require any JavaScript build tooling in order to compile Gleam.

inoas:
Thank you so much!

inoas:
I think this is actually completed already, if not reopen please.

PR #1764, opened 2022-09-16
Bundle JS and CSS instead of relying on unpkg CDN

aflatter:
This PR introduces a build step for the assets required to serve the documentation in order to make it usuable without an internet connection (see #1730). The resulting files can be committed to the repo so that you usually don't need NPM to work on `gleam` - unless you work on the documentation pages.

To build:

```
cd compiler-core/templates
npm install
npm run esbuild
git add dist/index.css dist/index.js
```

There's further potential for clean-up:
- We could move all of the docs-related JS into a subdirectory to make clear that it's separate from `prelude.js`
- We could run a formatter like `standard` on the JavaScript
- I think all of the code in `gleam.js` could be merged into `index.js` (or the other way around)


aflatter:
@lpil Here we go 👍 Let me know what you think.

lpil (commented):
Hello! Thanks for this!

Could you convert this to just having the files vendored in the repository, the same as we do for the existing assets? We don't want to depend on npm at all or have another build tool.

aflatter:
> Hello! Thanks for this!
> 
> Could you convert this to just having the files vendored in the repository, the same as we do for the existing assets? We don't want to depend on npm at all or have another build tool.

Hmm, can you expand on your rationale here?

re `npm`: If we just put the minified versions of each library into the directory, it's hard to track the versions and update them as required.

re `esbuild`: Gives us a single `js`/`css` file to include and lets us work with `import` instead of the global scope - making things much easier to understand.

Though I can understand the urge to avoid any frontend-specific tooling, I think committing the resulting bundle to the repo strikes an okay balance: Usually you won't need to install/work with these things, but if you need to, it will make the work easy to understand.

lpil:
It's not hard to keep track of versions, the version numbers commonly go both in the file name and the comment header of the file. It is simpler and easier to download a file than to install a package manager, a bundler, and run both.

Having 1 dependency in the global scope isn't difficult to understand, and continuing to use it that way means we don't need to rewrite any of the existing code.

We are not adding any new build tools to the Gleam compiler. We are actively working to remove other tools that are used in a fashion similar to how this PR works because the update workflow is confusing.

lpil:
Closing this due to inactivity, please feel free to reopen!

Issue #957, opened 2021-01-31
Destructuring pattern matching syntax for strings


lpil:
It would be useful to be able to pattern match on the prefix of a string in the same way that we can for bitstrings.

What could the syntax be? Erlang uses bit string syntax, Elixir uses bit string syntax or the `<>` operator.

Would we also support this syntax for strings construction? If so we may want to ensure the compile optimises multiple concatenations into a single Erlang bit string expression.

itsgreggreg:
You could think of it as Gleam already having the syntax but not typing it correctly.
`<<"hi":utf8, "there":utf8>>` is a valid pattern, expression and constant Utf8 String but gleam types it as a BitString. If the type checker knew that if every segment was a utf8 value then the whole thing was a String then additional syntax might not be needed.

itsgreggreg:
Perhaps adding `:string` as a segment option a la `<<"a":string>>` would help the type checker and alleviate needing `utf8` to be a sort of typeclass.

lpil:
We could use the same syntax but it would be the first instance of overloading in the language. I think we would want to think carefully about how we design and implement this as it has big implications for the language. The big question I think is how does the compiler decide if this is supposed to return a string or a bitstring?

```erlang
<<x:utf8, y:utf8>> 
```

Unless we can answer this question well I think I would prefer another syntax like Elixir has.

itsgreggreg:
A draw back of using pure concatenation is that you can't specify how long a prefix is in pattern matching. The prefix must be an exact value in patterns.  `x <> y` is undecidable in patterns.

Currently `<<"a", "b">>` is disallowed in gleam as segments default to int. perhaps having that default to `:string` and adding a `:string`  and `:grapheme` option could solve this problem. In the parsers I've written in Gleam I usually resort to matching integers off the front of a BitString because there's no low cost, convenient way to stream graphemes off the front of a `String`, something like this could be nice:


```
case some_long_input {
  <<"{", rest:string>> -> ...
  <<"}", rest:string>> -> ...
  <<a:grapheme, rest:string>> -> ...
}
```

where `string` is a variable length `String` if in the tail position, otherwise is an exact string, and `grapheme` is a single utf8 grapheme which is what `utf8` in pattern matches does now though with the BitString type

lpil:
```rust
case some_long_string_input {
  "{" <> rest -> ...
  a <> rest ->...
}
```

With a syntax like this I'm wondering how you express these things:

- `a` is of a certain length
[...]

PR #1768, opened 2022-09-20
<>

lpil:
Closes https://github.com/gleam-lang/gleam/issues/957

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- The `<>` operator can now be used for string concatenation and for string
+  prefix pattern matching.
 - Fixed a bug where TypeScript defintions may have incorrect type parameters.
 
 ## v0.23.0 - 2022-09-15
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1031,11 +1031,24 @@ pub enum Pattern<Constructor, Type> {
 
     Concatenate {
         location: SrcSpan,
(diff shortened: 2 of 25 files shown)


CrowdHailer:
I find the matching on variables pretty strange. or at least a potential for confusion. If you remove the comma section in the example there would be this line
```
greeting <>  name -> name
```
The first being a match/equality check and the second being an assignment seams magical. If later someone added a variable `name` earlier in the scope that would be a pretty confusing change in behaviour

lpil:
@CrowdHailer agreed, this is a discussion we have been having in Discord. Current favourite is to require guards for this.

```gleam
let greeting = "Hello"
case x {
  x <> ", " <> name if x == greeting -> name
  _ -> ...
}
```

tynanbe:
> Current favourite is to require guards for this.

```gleam
let a = "Eh"
let greeting = "Hello"

case phrase {
  // Invalid
  a <> ", " <> name -> name

  // Invalid
  a <> ", " <> name if name == "Joe" -> name

  // Valid?
  a <> ", " <> name if a == greeting -> name

  // Valid
  "#{a}, " <> name -> name

  // Valid
  "#{a}, " <> name if name == "Joe" -> name

  // Valid? Invalid?
  "#{a}, " <> name if a == greeting -> name
}
```

Should guards be allowed at all for `a`?

I'm not sure how much guards reduce confusion should multiple `<>` operators be allowed in a single pattern. One would still have to know that the last `<>` has a special meaning when it's followed by a variable.

PR #1770, opened 2022-09-23
Support compilation of Elixir files in src

### LINKED ISSUE #1766: Support compilation of Elixir files within a Gleam project
We can shell out and call `elixirc`?

We will need to do this last as the Elixir code may need to execute Gleam and Erlang code during compilation. How do we make it avaliable?

tynanbe:
Closes #1766.

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -145,16 +145,30 @@ where
         ];
         // Add the list of modules to compile
         for module in modules {
-            let path = self.out.join("build").join(module).with_extension("erl");
+            let path = self.out.join("build").join(module);
             args.push(path.to_string_lossy().to_string());
             self.add_build_journal(path);
-
-            let beam_path = self.out.join("ebin").join(module).with_extension("beam");
-            self.add_build_journal(beam_path);
         }
+        // Compile Erlang and Elixir modules
+        // Write a temporary journal of compiled Beam files
         let status = self
             .io
             .exec("escript", &args, &[], None, self.silence_subprocess_stdout)?;
+
+        let tmp_journal = self.lib.join("gleam_build_journal.tmp");
+        if self.io.is_file(&tmp_journal) {
+            // Read the temporary journal's lines into a vector
+            let beam_paths: Vec<PathBuf> = self
+                .io
+                .read(&tmp_journal)?
+                .split('\n')
+                .map(PathBuf::from)
+                .collect();
+            for beam_path in beam_paths {
+                self.add_build_journal(beam_path);
+            }
+            self.io.delete_file(&tmp_journal)?;
+        }
 
         if status == 0 {
             Ok(())
@@ -217,6 +231,15 @@ where
             match extension {
                 "mjs" | "js" | "hrl" => (),
                 "erl" => {
+                    let _ = to_compile_modules.insert(relative_path.clone());
+                }
+                "ex" => {
+                    // Elixir core libs must be loaded
+                    maybe_link_elixir_libs(
+                        &self.io,
+                        &self.lib.to_path_buf(),
+                        self.silence_subprocess_stdout,
+                    )?;
                     let _ = to_compile_modules.insert(relative_path.clone());
                 }
                 _ => continue,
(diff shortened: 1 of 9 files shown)


tynanbe:
Just noticed that I'll need to see that `gleam publish` picks up `.ex` files too.

compiler-core/src/build/package_compiler.rs
```diff
@@ -219,6 +233,15 @@ where
                 "erl" => {
                     let _ = to_compile_modules.insert(relative_path.clone());
                 }
+                "ex" => {
+                    // Elixir core libs must be loaded
+                    maybe_link_elixir_libs(
```
lpil, line 238: Why is it that we run this for every Elixir file? It looks like we could run it just once?

compiler-core/templates/gleam@@compile.erl
```diff
@@ -7,14 +7,31 @@
...
     #arguments{out = Out, lib = Lib, modules = Modules} = parse(Args),
+    IsElixirModule = fun(Module) ->
+        filename:extension(Module) =:= ".ex"
+    end,
+    {ElixirModules, ErlangModules} = lists:partition(IsElixirModule, Modules),
     ok = configure_logging(),
     ok = add_lib_to_erlang_path(Lib),
     ok = filelib:ensure_dir([Out, $/]),
+    {ErlangOk, ErlangBeams} = compile_erlang(ErlangModules, Out),
+    {ElixirOk, ElixirBeams} = compile_elixir(ElixirModules, Out),
```
lpil, line 18: I think we don't want to run the Elixir build if the Erlang build has finished, it likely depends on the Erlang modules being present.

compiler-core/templates/gleam@@compile.erl
```diff
@@ -43,20 +60,71 @@ worker_loop(Parent, Out) ->
...
+                    log(Message),
+                    erlang:send(Parent, Message);
+                error ->
                     log({failed, Module}),
                     erlang:send(Parent, failed)
             end,
             worker_loop(Parent, Out)
     end.
 
+compile_elixir(Modules, Out) ->
```
lpil, line 78: Let's add a check to see if Elixir is installed before continuing, the error message willbe cryptic otherwise.

lpil (commented):
Fab stuff! Really nice.

Left a few Qs

lpil:
If you rebase on main the build will be fixed I think

tynanbe:
@lpil, addressed all concerns, cleaned up, and caught some other issues along the way.

Thanks for the review!

compiler-core/templates/gleam@@compile.erl
```diff
@@ -43,20 +66,83 @@ worker_loop(Parent, Out) ->
...
+    case Modules of
+        [] -> {true, []};
+        _ ->
+            log({starting, "compiler.app"}),
+            ok = application:start(compiler),
+            log({starting, "elixir.app"}),
+            case application:start(elixir) of
+                ok -> do_compile_elixir(Modules, Out);
+                _ ->
+                    io:put_chars([Error, $\n]),
```
lpil, line 101: Let's print this to standard error as it is an error

compiler-core/templates/gleam@@compile.erl
```diff
@@ -43,20 +66,83 @@ worker_loop(Parent, Out) ->
...
+            ToBeam = fun(ModuleAtom) ->
+                Beam = filename:join(Out, atom_to_list(ModuleAtom)) ++ ".beam",
+                log({compiled, Beam}),
+                Beam
+            end,
+            {true, lists:map(ToBeam, ModuleAtoms)};
+        {error, Errors, _} when ShouldLog ->
+            ErrorFiles = lists:usort([File || {File, _, _} <- Errors]),
+            Log = fun(File) ->
+                log({failed, binary_to_list(File)})
```
lpil, line 127: Are these errors that we want to show the user, so they can debug the code? If so let's print them to standard error rather than logging them

compiler-core/templates/gleam@@compile.erl
```diff
@@ -43,20 +66,83 @@ worker_loop(Parent, Out) ->
...
+
+do_compile_elixir(Modules, Out) ->
+    ModuleBins = lists:map(fun(Module) ->
+        log({compiling, Module}),
+        list_to_binary(Module)
+    end, Modules),
+    OutBin = list_to_binary(Out),
+    Options = [{dest, OutBin}],
+    ShouldLog = should_log(),
+    % Silence "redefining module" warnings
```
lpil, line 114: Could we document why it is that we get these warnings?

compiler-core/templates/gleam@@compile.erl
```diff
@@ -43,20 +66,83 @@ worker_loop(Parent, Out) ->
...
+    'Elixir.Code':compiler_options([{ignore_module_conflict, true}]),
+    case 'Elixir.Kernel.ParallelCompiler':compile_to_path(ModuleBins, OutBin, Options) of
+        {ok, ModuleAtoms, _} ->
+            ToBeam = fun(ModuleAtom) ->
+                Beam = filename:join(Out, atom_to_list(ModuleAtom)) ++ ".beam",
+                log({compiled, Beam}),
+                Beam
+            end,
+            {true, lists:map(ToBeam, ModuleAtoms)};
+        {error, Errors, _} when ShouldLog ->
```
lpil, line 124: I think we can remove this guard because it already is called within the log function. All other code can always log and not worry about whether or not it should actually be printed

lpil (commented):
Looking great! Just a couple small things and then we are good to merge I think. Thank you for waiting while I was out of the country

compiler-core/templates/gleam@@compile.erl
tynanbe: Added a comment. This bit is just for logging failed filenames, to match the Erlang compiler behaviour.

lpil (approved):
Amazing work! Thank you so much! 💜

PR #1781, opened 2022-10-05
Fix width of images in gleam docs, so the image scales correctly 

ajstrand:
From discord:

>  we need to add a max-width or something to doc content 

> 11:33 PM]tynanbe: I think the best you can do might be using regular html for the img and adding something like a width="630" attribute.
[11:34 PM]lilac: that'll mess with how it's displayed on Github. this needs to be fixed at the doc generator level.

>  but max-width: 100% actually looks fine

I ended up implementing the the `max-width:100%` because `max-width:100vw` makes the image not respect the bounds of the parent container

Here's some test screenshots with an example image:
Screen Shot 2022-10-04 at 8 41 53 PM

Screen Shot 2022-10-04 at 9 04 44 PM

Screen Shot 2022-10-04 at 9 05 01 PM

### DIFF
--- a/compiler-core/templates/index.css
+++ b/compiler-core/templates/index.css
@@ -164,6 +164,10 @@ p code {
   max-width: var(--content-width);
 }
 
+.content img {
+  max-width: 100vw;
+}
+
 /* Page header */
 
 .page-header {
--- a/compiler-core/templates/index.css
+++ b/compiler-core/templates/index.css
@@ -165,7 +165,7 @@ p code {
 }
 
 .content img {
-  max-width: 100vw;
+  max-width: 100%;
 }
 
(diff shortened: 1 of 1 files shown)


inoas:
I think this is a good minimal change. I considered following alternatives:

1. `width: 100%; display block`: breaks things like icons if you want them in there and they have a smaller size such as pngs or svgs-with-set-pixel-sizes
2. `width: auto; max-height: 20rem;' or similar: does not allow to place tall images such as a cheatsheet or a vertical flow diagram.

I've applied the change as this PR suggests it locally in firefox inspector and it looks good.

lpil (approved):
Brilliant! Thank you

PR #1794, opened 2022-10-13
Fix minor typo

patriksvensson:
Just a minor thing I noticed while reading the source code.

The text ` /// A wrapper around std::io::Write that has Gleam's error handling.` 
should be `/// A wrapper around std::io::Read that has Gleam's error handling.`


lpil (approved):
Thank you!

PR #1798, opened 2022-10-16
Correctly record public information

rockwotj:
Module level variables were being recorded as false unconditionally, which is not true. This fixes the issue by introducing a new `insert_module_variable` function that allows specifying if the variable is public/private and using that in the correct places.

### DIFF
--- a/compiler-core/src/ast/tests.rs
+++ b/compiler-core/src/ast/tests.rs
@@ -64,7 +64,7 @@ fn compile_expression(src: &str) -> TypedExpr {
         }),
         module: "mymod".into(),
     };
-    environment.insert_module_variable(
+    environment.insert_variable(
         "Cat".into(),
         variant,
         type_::fn_(vec![type_::string(), type_::int()], cat_type.clone()),
--- a/compiler-core/src/ast/tests.rs
+++ b/compiler-core/src/ast/tests.rs
@@ -64,10 +64,11 @@ fn compile_expression(src: &str) -> TypedExpr {
         }),
         module: "mymod".into(),
     };
-    environment.insert_variable(
+    environment.insert_module_variable(
         "Cat".into(),
         variant,
         type_::fn_(vec![type_::string(), type_::int()], cat_type.clone()),
+        true
     );
 
     environment.insert_accessors(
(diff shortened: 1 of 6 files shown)


lpil:
Hello! Thank you very much for this patch. Did you find an incorrect behaviour due to this? If so it would be good to get a test that replicates the problem so that we can make sure it doesn't happen again. 

I believe that local variables are always private because they never get exposed. There is another data structure for storing module variables so I think we would need to rename this new method to something else.

rockwotj:
> Did you find an incorrect behaviour due to this?

Not yet! I've been playing with the compiler and noticed this.

> If so it would be good to get a test

I haven't been able to find a good place to put a test (as there are no downstream effects AFAIK) but if you have a suggestion always happy to add tests for things.

> I think we would need to rename this new method to something else.

All the other usages were using `LocalVariable` variant so I was able to clean the naming and args up a bit - I ended up renaming the original function. Please let me know if you have something else in mind :)

rockwotj:
> If so it would be good to get a test that replicates the problem so that we can make sure it doesn't happen again.

Here is the fixed test in my upstream version that happens from this patch: https://github.com/rockwotj/native-gleam/commit/b96b25815ba0d57b7a9be35dc6543d16e92b9e23

Those changes aren't yet ready to be upstreamed however, and I don't think the generated JS or Erlang cares about the referenced value's visibility.

lpil:
Oh cool! I'm not familiar with C++, what does the `_private` section in the generated code do there?

rockwotj:
It's the namespace for the generated code. Because of how template inlining works, private functions still need to have external linkage and declared in a header. So to partition the public and private functions/constants the private functions get wrapped into a `_private` namespace. This is a common pattern in C++ ([examples in boost](https://cs.github.com/boostorg/multiprecision?q=org%3Aboostorg%20(%22namespace%20internal%22%20OR%20%22namespace%20impl%22%20OR%20%22namespace%20detail%22))) where there is usually a namespace called `detail`, or `impl` or `internal` that is not part of the public API.

Before this change we assumed all usages of functions are in the private namespace for the module causing compilation errors.

lpil (approved):
Ah! Thank makes sense. Thank you for the explanation 💜

PR #1800, opened 2022-10-16
feat: Only run CI when rust code changed

HarryET:
Only run the Compiler CI when the compiler code has been changed. Saves on CI minutes and means that PRs that don't affect the compiler don't have to wait for the compiler tests.


lpil (approved):
Thank you

lpil:
Hmm actually I'm not sure when else CI would run, there's only one project in here. I'll close for now, CI speed isn't an issue for us atm

Issue #1709, opened 2022-08-05
Syntax to write nested functions in a flat style (promise.then, result.then)


lucasavila00:
Inspired by the discussion https://github.com/gleam-lang/gleam/discussions/1708

I propose a syntax extension that de-sugars:

from

```gleam
import gleam/result
import gleam/javascript/promise

// Todo: improve with + as syntax
let op2 = with result.then as then {
  then x = Ok("x")
  then y = Ok("y")
  io.println(x)
  io.println(y)
  Ok(string.concat([x, y])
}

let op3 = with promise.then as then {
  then x = promise.resolve("x")
  then y = promise.resolve("y")
  io.println(x)
  io.println(y)
  promise.resolve(string.concat([x, y])
}

```

to

``` gleam
let op2 =
  Ok("x")
  |> result.then(fn(x) {
    Ok("y")
    |> result.then(fn(y) {
      io.println(x)
      io.println(y)
      Ok(string.concat([x, y]))
    })
  })

let op3 =
  promise.resolve("x")
  |> promise.then(fn(x) {
    promise.resolve("y")
    |> promise.then(fn(y) {
      io.println(x)
      io.println(y)
      promise.resolve(string.concat([x, y]))
    })
  })
``` 

or to

```gleam
  let op2 =
    result.then(
      Ok("x"),
      fn(x) {
        result.then(
          Ok("y"),
          fn(y) {
            io.println(x)
            io.println(y)
            Ok(string.concat([x, y]))
          },
        )
      },
    )

  let op3 =
    promise.then(
      promise.resolve("x"),
      fn(x) {
        promise.then(
          promise.resolve("y"),
          fn(y) {
            io.println(x)
            io.println(y)
            promise.resolve(string.concat([x, y]))
          },
        )
      },
    )
``` 

It should work with any user defined type, too.

lpil:
Tagging @hayleigh-dot-dev as I think she is very interested in this.

I could see this being a very powerful and convenience feature to have in the language. The main things I would want to investigate and think through carefully before adding something like this are:

## Ease of use

How does this impact the complexity of the language and how easy it is to learn?

It is important that the language remains as approachable as possible, even if that means language features are a little less expressive and powerful in exchange.

## Parsing

How would the passing of the syntax work? In the specific example I think the parser would need to know that then has become a special keyword once the syntax has been defined.

[...]

PR #1812, opened 2022-10-25
Use

lpil:
Part of https://github.com/gleam-lang/gleam/issues/1709

- [x] Implement
- [x] Invalid call error
- [x] No expressions error
- [x] Test error locations are sensible
- [x] Permit discards


compiler-core/src/parse.rs
```diff
@@ -678,6 +683,35 @@ where
         Ok(Some(expr))
     }
 
+    // A `use` expression
+    // use <- function
+    // use <- function()
+    // use <- function(a, b)
+    // use <- module.function(a, b)
+    // use a, b, c <- function(a, b)
+    // use a, b, c, <- function(a, b)
```
inoas, line 692: Can it also be used in pattern matching like let:

> let Cat(name: name, cuteness: _, age: _) = cat

lpil: Nope, at least not with this first version.

PR #1819, opened 2022-10-31
docs generator pass 1

inoas:
## Up to date local static assets

- do not use CDNs so that assets are available offline or if cdn shuts or goes down
- upgrade highlightjs static asset to 10.5 to 11.6 including all the languages we support
- add highlightjs code for javascript so that we fully support all languages gleam tooling supports
- move static assets into subdirs to clarify what the files are used for (e.g. not for templating itself, but as a dependency on the template output)
- add/correct existance tests to `ci.yaml` for project generation
- ~~prepare to add google web font as woff2 only~~
- ~~The local assets up the package size buy approximately 150-160 KiB (including woff2 files and extended latin and cyrillic, do we need cyrillic?)~~


compiler-core/src/docs.rs
```diff
@@ -222,30 +222,85 @@ pub fn generate_html(
         });
     }
 
-    // Render static assets
+    // Render static assets - TODO: do not push these through the templater
```
lpil, line 225: What does this mean?

compiler-core/src/docs.rs
```diff
@@ -222,30 +222,85 @@ pub fn generate_html(
...
+    // files.push(OutputFile {
+    //     path: PathBuf::from("fonts/karla-v23-latin-ext_latin-regular.woff2"),
+    //     text: std::include_str!("../templates/docs-fonts/karla-v23-latin-ext_latin-regular.woff2"),
+    // });
+
+    // files.push(OutputFile {
+    //   path: PathBuf::from("fonts/ubuntu-mono-v15-latin-ext_latin_greek-ext_greek_cyrillic-ext_cyrillic-regular.woff2"),
+    //   text: std::include_str!("../templates/docs-fonts/ubuntu-mono-v15-latin-ext_latin_greek-ext_greek_cyrillic-ext_cyrillic-regular.woff2"),
+    // });
+
```
lpil, line 296:
```suggestion
```

compiler-core/templates/docs-css/index.css
```diff
@@ -1,5 +1,39 @@
...
+@font-face {
+  font-family: 'Karla';
+  font-style: normal;
+  font-weight: 700;
+  font-display: swap;
+  src: local(''),
+       url('../fonts/karla-v23-latin-ext_latin-700.woff2') format('woff2'); // Chrome 26+, Opera 23+, Firefox 39+
+}
+*/
+
```
lpil, line 36:
```suggestion
```

lpil (commented):
Nice! Could you remove the unused files also please

compiler-core/src/docs.rs
inoas: This PR is not yet done entirely.
Currently all `OutputFile` are being pushed through the template parser as part of `generate_html` but:

- we do not need to push static assets through that
- we cannot push static binary assets like woff2 web fonts through that

Also notice: The google web font breaks privacy requirements in the EU

inoas: Here `generate_html` is called:

```rust
pub(crate) fn build_documentation(
    config: &PackageConfig,
    compiled: &mut Package,
) -> Result<Vec<gleam_core::io::OutputFile>, Error> {
    compiled.attach_doc_and_module_comments();
    cli::print_generating_documentation();
    let mut pages = vec![DocsPage {
        title: "README".to_string(),
        path: "index.html".to_string(),
        source: paths::readme(), // TODO: support non markdown READMEs. Or a default if there is none.
    }];
    pages.extend(config.documentation.pages.iter().cloned());
    let outputs = gleam_core::docs::generate_html(config, compiled.modules.as_slice(), &pages);
    Ok(outputs)
}
```

inoas: I will break this PR apart again, and remove the unused files and do the google web fonts thing as a separate PR

inoas:
@lpil should be gtm now

lpil (commented):
Wonderful, thank you! Would you mind updating the changelog file?

inoas:
> Wonderful, thank you! Would you mind updating the changelog file?

Done, I have omitted the non-developer facing changes, I hope that's correct.

inoas:
The failing tests seem unrelated to this PR.

lpil (approved):
Thank you!!

PR #1824, opened 2022-11-02
don't recompile unchanged modules

sdancer:
pr


lpil (commented):
Cool stuff!

Do you think there is a way we can test any of the components here? The build tool is light on unit tests, instead preferring to use integration tests. I'm not sure how we would write an integration test for this- how would we identify that the modules have not been recompiled if they don't have to be?

Can we comment this code heavily? I'm finding it help a lot of maintenence and approachability to overly-communicate intent in this project.

sdancer:
oki, however will go back to this after the lsp autocompletion work, because its providing a lot of familiarity with the same interfaces used here.

sdancer:
```
I'm not sure how we would write an integration test for this- how would we identify that the modules have not been recompiled if they don't have to be?
```
most the effort is to ensure that in any uncertain case modules are recompiled (the other way around) the major problem with this feature would be modules not recompiling when they should

lpil:
Right, how would we test that case also?

michallepicki:
Should this be closed now after https://github.com/gleam-lang/gleam/pull/1910 got merged?

lpil:
Ah yes, thank you both!

PR #1825, opened 2022-11-02
bitstring allow float sizes

sdancer:
change the test, add 16/32/64 so it always work, dynamic size(from variable) is also allowed on erlang


lpil (commented):
Can we have tests in `test/language` to show it working at runtime on Erlang? And a note in the changelog too please

Thanks!

sdancer:
the test that updated on the commit shows it working properly producing working erlang code

PR #1827, opened 2022-11-02
Float size2

sdancer:
updated pr


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
   permits callback using code to be written without indentation.
 - Fixed a bug where the warning for `todo` would not print the type of the code
   to complete.
+- Fixed float-size in bitstring
```
lpil, line 10: Could you explain what this means? Ideally a user can read this and know what new things they can do.

compiler-core/src/bit_string.rs
```diff
@@ -234,15 +234,7 @@ fn type_options<T>(
...
-    // size cannot be used with float
-    match categories {
-        SegmentOptionCategories {
-            typ: Some(Float { .. }),
-            size: Some(opt),
-            ..
-        } => err(ErrorType::FloatWithSize, opt.location()),
-        _ => Ok(categories.segment_type()),
-    }
+    Ok(categories.segment_type())
```
lpil, line 237: Is FloatWithSize still used?

lpil (commented):
Nice work!

Could we get runtime tests too to show it working? https://github.com/gleam-lang/gleam/blob/8970ea76e3daead73ab61dc68b9e1be8b37c9f9f/test/language/src/main.gleam#L1039

Thanks!

compiler-core/src/bit_string.rs
sdancer: should be used in the future to validate static size of floats to 16/32/64 only, not used now

sdancer: actually will add that now(FloatWithSize), lacked enough understanding to add it two days ago, will also add the runtime tests

CHANGELOG.md
lpil: Could you update this please? 🙏

compiler-core/src/bit_string.rs
lpil: Nice, thanks!

lpil (approved):
Thank you!

lpil:
Sorry, could you resolve the merge conflict please?

sdancer:
nice

Issue #1667, opened 2022-06-23
Raise helpful custom error when a user attempts to call functions in constants


aslilac:
Something like...

```gleam
const puppies = map.from_list([
    "August",
    "Dot",
    "Mady",
    "Spot",
    "Toby"
  ] |> list.map(fn (name) { #(name, Puppy(name, good: True, soft: True)) }))
```

...turns into...

<img width="566" alt="Screen Shot 2022-06-22 at 6 24 06 PM" src="https://user-images.githubusercontent.com/418348/175182048-434cef97-c4c1-4597-8791-0eec9d328a89.png">

...which is definitely not what I expected. If you turn it into a `let` and put it in a function body, it works perfectly, so I'd expect the error to be a bit more specific about the fact that this is failing because it's a `const` declaration

lpil:
We can improve this error message for sure. Thanks for the report

isomorpheme:
I'd like to take a shot at fixing this!

lpil:
Brilliant, thank you

Ian-GL:
Hi! Could I give it a try given it's been a bit since there was some activity with the issue? :)

lpil:
Hey @Ian-GL ! That would be fantastic, thank you

PR #1839, opened 2022-11-11
Add error explanation when a constant is declared using a function

Ian-GL:
Changed the parser for the const when it finds a dot to also check for the token after the next name and add the error if that next token is `(`.

Screen Shot 2022-11-10 at 17 18 18

Solves #1667


compiler-core/src/parse.rs
```diff
@@ -2167,14 +2167,27 @@ where
...
-                            constructor: None,
-                            typ: (),
-                        }))
+                        if self.peek_tok1() == Some(&Token::LeftParen) {
+                            let _ = self.next_tok(); // name
+
+                            parse_error(
+                                ParseErrorType::UnexpectedToken {
+                                    expected: vec!["A constant definition".to_string()],
+                                    hint: Some("Functions cannot be called in constant declarations. You can use let instead (and make sure it is inside a scope)".to_string())
```
lpil, line 2176:
```suggestion
                                    hint: Some("Functions can only be called within other functions.".to_string())
```

Perhaps this might be clearer? I wasn't sure what the `let` and scope bits meant.

compiler-core/src/parse.rs
```diff
@@ -2167,14 +2167,27 @@ where
...
-                            module: Some(name),
-                            name: end_name,
-                            constructor: None,
-                            typ: (),
-                        }))
+                        if self.peek_tok1() == Some(&Token::LeftParen) {
+                            let _ = self.next_tok(); // name
+
+                            parse_error(
+                                ParseErrorType::UnexpectedToken {
```
lpil, line 2174: I think a dedicated error would be best here as we want more information, and the `expected` field below isn't quite correct. If we do this we wouldn't need to have the `hint` be part of the error value, it could be added when printing the error.

lpil (commented):
Fab stuff! I've left some notes below. We should also add a snapshot test to ensure the error is formatted as we expect. The `assert_error` macro would work well for this

compiler-core/src/parse.rs
Ian-GL: Yes, I think that's way clearer. I was trying to make a hint related to what was described in the issue, which mentions that if the function is called as `let a =` inside a function it works. But it is better to have this be more generic.

Ian-GL:
> Fab stuff! I've left some notes below. We should also add a snapshot test to ensure the error is formatted as we expect. The `assert_error` macro would work well for this

Hey @lpil , I added a new error but I am having trouble with the test. The snapshot for the test I added does show a different error of the compiler complaining about `const` instead of the actual function call (as can be seen in the committed snapshot).

I was wondering if you could point me into what could be happening or what could I check. When I reproduce it using one of the test packages it works as expected:

Screen Shot 2022-11-11 at 16 21 13

lpil:
Yup that's right. I think it would probably be easiest to put this in the type tests rather than the parse ones.

Ian-GL:
Hi @lpil ! After digging more into how to write a test for this I realized the current macros to assert errors on a module wouldn't be suitable for this case since those expect the module to be parsed and the error to happen in the actual compile to Erlang (if I understand correctly), while in this case the error happens during parsing. So I made a new assertion for expected syntax errors like this one.

I'm quite unsure if this sort of test is fine and if the types test file is the correct place. Very happy to change this to what makes more sense and I hope I did understand correctly how the assert macros are working.

compiler-core/src/parse.rs
```diff
@@ -2167,14 +2167,27 @@ where
...
+
+                            parse_error(
+                                ParseErrorType::UnexpectedFunction,
+                                SrcSpan {
+                                    start,
+                                    end: end + 1,
+                                },
+                            )
+                        } else {
+                            let _ = self.next_tok(); // name
```
lpil, line 2181: This is the same in both branches so let's lift it to before the `if`

lpil (commented):
Oh ok! Well in which case we'll need to have a test macro that gets a parse error and then prints it and make an assertion about the error output using the snapshot macros, like we do with type errors etc.

Let me know if you'd like me to make this macro for you 👍

Ian-GL:
> Oh ok! Well in which case we'll need to have a test macro that gets a parse error and then prints it and make an assertion about the error output using the snapshot macros, like we do with type errors etc.
> 
> Let me know if you'd like me to make this macro for you 👍

Gonna give it a try before I play the "phone the creator of the language" card :)

I think by looking at how the other macros do this I should be able to do it.

compiler-core/src/parse.rs
Ian-GL: I initially did this because if I understand correctly, `next_tok` pops the current token, which would made `peek_tok1` return another token and not the one that could be a LeftParen. I've changed that to instead pop the token and then just match `self.tok0`. https://github.com/gleam-lang/gleam/pull/1839/commits/0982dc4db6eacf6609a1e64cef21d5c0c20f584a

Happy to change this if there's a better way to do it.

lpil:
Ah nice, just need to call this method now for the error to see how it renders:

https://github.com/gleam-lang/gleam/blob/c53aa2dbf7579352c8b1924733a48ee4296f83b6/compiler-core/src/error.rs#L349

Ian-GL:
> Ah nice, just need to call this method now for the error to see how it renders:
> 
> https://github.com/gleam-lang/gleam/blob/c53aa2dbf7579352c8b1924733a48ee4296f83b6/compiler-core/src/error.rs#L349

Hey @lpil , sorry for taking long to address this. I just made the change you asked for so I think this should be good now :)

lpil (approved):
Perfect! Thank you very much 💜 

And no worries about taking time, you are doing free work for the Gleam community and that is super appreciated at any pace

PR #1841, opened 2022-11-12
fix x86_64-unknown-linux-musl test env

inoas:
Demo that it works is here:
https://github.com/inoas/gleam-src/actions/runs/3452674111/jobs/5762734689


lpil (approved):
Amazing, thank you

Issue #1847, opened 2022-11-14
Update CI to upgrade deprecated commands


Ian-GL:
I noticed there are warnings in the recent Actions runs given that there was a change on the command to set variables. The guide to upgrade is [this one](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) and it seems like a simple change.

I would like to take up this task. Opened an issue first to discuss the approach. My plan would be to first change the nightly workflow given it is less complex and if it works without issues, then make a follow up PR with the rest of the workflows.

lpil:
Fab, thank you!

lpil:
Thank you!

PR #1848, opened 2022-11-15
Update nightly-build action and BEAM setup action

Ian-GL:
Upgraded the `set-ouput` deprecated command to use instead the env files as described on the GHA update guide [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/). I'm not sure on the best way to test this but I've tested a similar action with the env file on a personal private repo to be sure it worked.

Also upgraded the BEAM setup action to the latest one which also gets rid of this warning on the action (as can be seen [here](https://github.com/erlef/setup-beam/releases/tag/v1.14.0)).

This work is related to #1847

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -75,7 +75,7 @@ jobs:
           profile: minimal
 
       - name: Install Erlang (non-macos)
-        uses: erlef/setup-beam@v1.13.0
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you! Would you mind also updating the changelog file? Sorry I should have said that before

lpil (approved):
Thanks!

Issue #1834, opened 2022-11-06
Javascript nested sequences 


rockwotj:
I believe nested sequences within JavaScript is broken.

Take this example program (that granted is doing naughty things ignoring the `Result`).

```gleam
pub fn main() -> Result(Int, Nil) {
  let x = 1;
  {
    try b = Error(Nil);
    b
  }
  Ok(x)
}
```

The result of `main` in Erlang is `Ok(1)` but `Error(Nil)` in JavaScript.

Output Erlang:

```erlang
-module(main).
-compile(no_auto_import).

-export([main/0]).

-spec main() -> {ok, integer()} | {error, nil}.
main() ->
    X = 1,
    case {error, nil} of
        {error, _try} -> {error, _try};
        {ok, B} ->
            B
    end,
    {ok, X}.
```

Output JavaScript:

```javascript
import { Ok, Error } from "./gleam.mjs";

export function main() {
  let x = 1;
  let $ = new Error(undefined);
  if (!$.isOk()) return $;
  let b = $[0];

  b;
  return new Ok(x);
}
```

I think sequences in JS need to be unconditionally wrapped in a lambda due to `try`. It maybe that if there are no try statements then the lambda may be omitted, but that seems unnecessary.

lpil:
We are supposed to wrap sequences in a function.

I think the budget is that we are not doing the same wrapping for try, which we should. Should be an easy fix!

rockwotj:
Maybe I am missing something, but this doesn't seem like an easy fix.

Here's a diff that I think should fix things:

```diff
diff --git a/compiler-core/src/javascript/expression.rs b/compiler-core/src/javascript/expression.rs
index 34bbf0b3..169efe8f 100644
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -375,7 +375,14 @@ impl<'module> Generator<'module> {
         let mut documents = Vec::with_capacity(count * 3);
         for (i, expression) in expressions.iter().enumerate() {
             if i + 1 < count {
-                documents.push(self.not_in_tail_position(|gen| gen.expression(expression))?);
+                documents.push(self.not_in_tail_position(|gen| {
+                    // Nested sequences still need to wrapped in iife for `try`.
+                    if matches!(expression, TypedExpr::Sequence { .. } | TypedExpr::Pipeline { .. }) {
+                        gen.immediately_involked_function_expression(expression)
+                    } else {
[...]

PR #1850, opened 2022-11-15
Try block

lpil:
Closes https://github.com/gleam-lang/gleam/issues/1834

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -248,6 +248,17 @@ impl<'module> Generator<'module> {
     }
 
     /// Wrap an expression in an immediately involked function expression if
+    /// required due to using early returns.
+    pub fn top_of_sequence_expression<'a>(&mut self, expression: &'a TypedExpr) -> Output<'a> {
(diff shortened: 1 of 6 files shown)


compiler-core/src/javascript/expression.rs
```diff
@@ -247,6 +247,17 @@ impl<'module> Generator<'module> {
         result
     }
 
+    /// Wrap an expression in an immediately involked function expression if
+    /// required due to using early returns.
+    pub fn top_of_sequence_expression<'a>(&mut self, expression: &'a TypedExpr) -> Output<'a> {
+        match expression {
+            TypedExpr::Sequence { .. } | TypedExpr::Try { .. } => {
```
rockwotj, line 254: Sequence here isn't strictly needed right? There are no other early return situations outside of `Try` right?

rockwotj (approved):

compiler-core/src/javascript/expression.rs
lpil: I think you're right, yes. Thank you!

PR #1851, opened 2022-11-15
Update use of set-output on release-containers workflow

Ian-GL:
`set-output` is also used on the release containers flow. Maybe this could be merged tomorrow after seeing the nightly build works as expected with the new format.

Related to #1847


lpil (approved):
Thank you!

PR #1858, opened 2022-11-19
0.25 RC Docs gen: BugFix highlightJS for Elixir and further langs

inoas:
Before this PR:
- Erlang, Elixir etc would not be highlighted correctly.
- Additional modules would throw warnings as `highlightBlock` is called with unsupported elements if the gleam highlight js file gets loaded before the erlang and elixir ones.

Followup/fix for https://github.com/gleam-lang/gleam/pull/1819

Changes:
- `highlightBlock()` is deprecated in favour of `highlightElement`.
- Calling `hljs.highlightAll();` applies hljs to all languages loaded at that moment, so we don't need to do it specifically for the gleam highlighter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
   the entire project rather than just the top level package.
 - The formatter now adds a 0 to floats ending with `.` (ie 1. => 1.0).
 - New projects require `gleam_stdlib` v0.25.
-- Fixed a regression where Elixir and Erlang markdown code blocks in generated
+- Fixed a regression where Elixir and Erlang Markdown code blocks in generated
   docs would not be highlighted.
 
 ## 0.24.0 - 2022-10-25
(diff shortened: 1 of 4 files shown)


inoas:
Screen Shot 2022-11-19 at 19 12 40
Screen Shot 2022-11-19 at 19 12 37
Screen Shot 2022-11-19 at 19 12 30

@tynanbe Could you confirm that this fixes your issue?

lpil (commented):
The forEach block was making it so that the default language for highlighting was Gleam. How does this get done now?

lpil (approved):
Thank you!

lpil:
Could you update the changelog also? Thanks

inoas:
I've checked 0.24 and this bug was most likely introduced in https://github.com/gleam-lang/gleam/pull/1819 which is between 0.24 and 0.25.

inoas:
I have found a wording for the changelog, I hope that fits for you.

lpil (commented):
Could you add it to a new section please? rc1 has already been released

lpil (approved):
Thank you

PR #1867, opened 2022-11-26
Make the typechecker recover from errors and fill the typed AST

cybergsus:
This PR focuses making sure errors can be propagated while also having
available data for the LSP:

- [x] Introduces a new `Result`-like structure that always has data, `FilledResult`, to the pipeline, which interacts almost seamlessly with `Result`s and is able to gather multiple errors.

- [x] Uses the new `FilledResult` and `FilledResultContext` helper structures to recover missing data (e.g unknown types from bad lookups) while propagating any errors that happened.

- [ ] Passes tests (current status: they don't pass due to error duplication happening *somewhere* in the code)

- [ ] Uses the data returned by `FilledResult` in LSP to be able to provide information even when the types are missing.

### DIFF
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -2,12 +2,12 @@ use std::time::Instant;
 
 use gleam_core::{
     build::{Options, Package, ProjectCompiler},
-    Result,
+    WResult,
 };
 
 use crate::{build_lock::BuildLock, cli, dependencies::UseManifest, fs};
 
-pub fn main(options: Options) -> Result<Package> {
+pub fn main(options: Options) -> WResult<Package> {
     let lock = BuildLock::new()?;
     let manifest = crate::dependencies::download(cli::Reporter::new(), None, UseManifest::Yes)?;
 
--- a/compiler-cli/src/compile_package.rs
+++ b/compiler-cli/src/compile_package.rs
@@ -8,11 +8,11 @@ use gleam_core::{
     metadata, paths,
     type_::Module,
     uid::UniqueIdGenerator,
-    Result,
+    Result, WResult,
 };
 use std::path::Path;
 
-pub fn command(options: CompilePackage) -> Result<()> {
+pub fn command(options: CompilePackage) -> WResult<()> {
     let ids = UniqueIdGenerator::new();
     let mut type_manifests = load_libraries(&ids, &options.libraries_directory)?;
     let mut defined_modules = im::HashMap::new();
--- a/compiler-cli/src/docs.rs
+++ b/compiler-cli/src/docs.rs
@@ -7,7 +7,7 @@ use gleam_core::{
     error::Error,
     hex,
     io::HttpClient as _,
-    paths, Result,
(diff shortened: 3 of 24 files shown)


cybergsus:
I found that switching `Arc<Type>` for `Option<Arc<Type>>` in the typed AST is not a nice solution since it would require a ton of `.unwrap()`s elsewhere in the code. I'm going to experiment with making the typechecker output an optionally typed AST and converting that AST to 'fully typed' (i.e `unwrap` all the types) when passing it to codegen

cybergsus:
After typing for a bit, I find making a new AST a lot of work that I don't know which direction is going to take me. I'm going to continue with this extra 'unknown' type and start producing `UnknownFromError` types when erroring out.

cybergsus:
looks like introducing an unknown type is not really needed, since we already have an `unbound_var` which acts as our unknown type.

cybergsus:
I'm implementing the forwarding of errors while keeping data, here's a sample from `PatternTyper::unify`: ```rs
            Pattern::List {
                location,
                elements,
                tail,
            } => {
                let mut ctx = FilledResultContext::new();

                let typ = if let Some(args) =
                    type_.get_app_args(true, &[], "List", 1, self.environment)
                {
                    args.get(0)
                        .expect("Failed to get type argument of List")
                        .clone()
                } else {
                    let list_contents = self.environment.new_unbound_var();
                    ctx.register_error(Error::CouldNotUnify {
                        location,
                        situation: None,
                        expected: type_.clone(),
                        given: list(list_contents),
                        rigid_type_names: hashmap![],
                    });
                    list_contents
                };

                let elements = elements
                    .into_iter()
                    .map(|element| ctx.slurp_filled(self.unify(element, typ.clone())))
                    .collect();

                let tail =
                    tail.map(|tail| Box::new(ctx.slurp_filled(self.unify(*tail, list(typ)))));

                ctx.finish(Pattern::List {
                    location,
                    elements,
                    tail,
                })
            }
```

cybergsus:
The amount of modification this requires is turning out to be huge. I'll make separate commits that don't compile to make the review easier, and we can decide if some things can be squashed together.

cybergsus:
I have converted quite a bunch of the typechecking functions at this point to recover from errors and still return valid data. But now I have come to a (somewhat) dead end: in `ExprTyper::infer_field_access`, we try to first infer it as a record access and then, if it fails, try to infer the expression as a module access. I could make `infer_record_access` return valid data even if it fails to get a constructor/field map. And I can recover from `infer_module_access` errors. But what data do I return if both fail? The partial record access or the partial module access? I'm thinking of searching in the errors that `infer_record_access` returns for an `UnknownRecordField`, but I'm not sure that this would suffice.

sdancer:
check my PR about partial compilation for the lsp

cybergsus:
> i would recommend u to check my PR about partial compilation for the lsp

Would you mind to point to it?

cybergsus:
Looking at the tests error output it seems like I've introduced duplication of errors on the compiler.

Issue #1879, opened 2022-12-03
Support creating a new Gleam project in the current directory with `gleam new .`


samueldurantes:
Apparently not is possible to create a Gleam project in your current directory.

Example:

![image](https://user-images.githubusercontent.com/44513615/205461410-caff5c07-57a5-49d1-982d-73b5407bfd61.png)

lpil:
Sounds good to me 👍

pascalwengerter:
@lpil would you prefer `gleam new .` or `gleam init` (according to how Rust handles it IIRC) or potentially both/aliasing it somehow?

lpil:
`gleam new .` please 🙏

abs0luty:
I can work on this one. ❤

PR #1881, opened 2022-12-04
feat: allow creates a new gleam project in the current directory

samueldurantes:
Closes: #1879


lpil (commented):
Hi! Thank you for this 💜 

Rather than checking if the directory is `.` let's instead check if any of the file to create already exist. This would allow us to write to other already existing directories too.

Let's also make a test for this, either in the CI config or as an actual Rust test with a temp directory.

lpil:

PR #1884, opened 2022-12-05
feat: add ts file support

### LINKED ISSUE #1835: Support TypeScript file inclusion in builds
Currently TypeScript files aren't included in the build output while JavaScript files are. This would be useful for JavaScript targets such as Deno that can run TypeScript files directly without a build step.

brettkolodny:
Closes #1835 

These changes successfully allow you to have top level `.ts` files in your src directory that are copied over to the build. 

At the moment the only test for this is the fact that running `make javascript` in `test/language` does indeed copy the file over because using the file in a Gleam project would require either a Node build step or Deno support. Is this enough for the feature or should some build step be included?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed a bug where returning an anonymous function from a pipeline and calling
   it immediately without assigning it to a variable would produce invalid Erlang
   code.
+- Added support for top level TypeScript file inclusion in builds.
 
 ## v0.25.0 - 2022-11-24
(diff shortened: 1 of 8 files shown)


lpil:
Heya! We could do something ugly and have the test project look inside the build directory and assert the file is there. Not pretty at all and quite brittle but it'll let us know if it breaks

lpil (approved):
Thank you

lpil:
I have rebased this into main

PR #1885, opened 2022-12-06
feat: Deno JavaScript runtime support

brettkolodny:
Support for Deno as a runtime for the JavaScript target.

* Adds a `target` property to the `javascript` config that can be set to `node` or `deno`, and defaults to `node`.
* Adds a `javascript-deno` section where permissions can be set.
    * allow_all `bool`
    * allow_read `Vec<String>`
    * allow_write `Vec<String>`
    * allow_env `Vec<String>`
    * allow_net `Vec<String>`
    * allow_hrtime `bool`
    * allow_sys `bool`
    * allow-run `Vec<String>`

## Progress

- [x] gleam run
- [x] gleam build
- [x] permissions
- [x] tests
- [x] proper error messages on invalid runtime


brettkolodny:
To get Deno to work with `gleam run` I ended up creating an entry file template like Erlang has based on the discussion in #1628. This isn't _needed_ for node but I currently have node doing the same for consistency.

lpil (commented):
Hey @brettkolodny ! This is looking fab.

Why wouldn't we use `deno eval` the same way we use `node -e`?

brettkolodny:
> Why wouldn't we use `deno eval` the same way we use `node -e`?

@lpil Unfortunately `deno eval` has implicit access to all permissions so using it wouldn't allow users to specify what permissions they want deno to have access to.

lpil:
Wow, I didn't expect that at all. Thanks!

brettkolodny:
I've modified the test in `test/language` to work with Deno and added to to the Makefile to run with Deno. I've also tested that the permissions seem to be working but I don't know if you want that to be automated somehow as well. Outside of that all of the core functionality seems to be done.

compiler-cli/src/run.rs
```diff
@@ -94,3 +99,75 @@ fn run_javascript(
...
+
+        // Allow sys
+        if config.javascript.deno.allow_sys {
+            args.push("--allow-sys".into())
+        }
+
+        // Allow hrtime
+        if config.javascript.deno.allow_sys {
+            args.push("--allow-sys".into())
+        }
```
lpil, line 127: I think this should be --allow-hrtime or such?

compiler-core/src/error.rs
```diff
@@ -2288,6 +2296,36 @@ issue in our tracker: https://github.com/gleam-lang/gleam/issues",
...
             }
+            Error::InvalidRuntime {
+                target,
+                invalid_runtime,
+            } => {
+                let text = format!("Invalid runtime for {} target: {}", target, invalid_runtime);
+
+                let hint = match target {
+                    Target::JavaScript => {
+                        // let valid_runtimes_text = valid_runtimes.join(", ");
```
lpil, line 2307:
```suggestion
```

test/project_javascript/gleam.toml
```diff
@@ -4,6 +4,12 @@ target = "javascript"
 
 [javascript]
 typescript_declarations = true
+runtime = "deno"
```
lpil, line 7: Can we default to node here? Thanks

lpil (commented):
Fantastic work! Thank you

We'll also need Deno added to CI

lpil:
Sorry @brettkolodny , I think the deno step needs to go in the other section of the workflow

brettkolodny:
Ah yeah you were right, I just checked it where I saw node and didn't read around it.

compiler-cli/src/run.rs
```diff
@@ -37,8 +42,17 @@ pub fn command(arguments: Vec<String>, target: Option<Target>, which: Which) ->
...
+            Some(r) => Err(Error::InvalidRuntime {
+                target: Target::Erlang,
+                invalid_runtime: r,
+            }),
+            _ => run_erlang(&module, arguments),
+        },
+        Target::JavaScript => match config.javascript.runtime {
+            Runtime::Deno => run_javascript_deno(&config, arguments),
+            Runtime::Node => run_javascript_node(&config, arguments),
+        },
```
lpil, line 55: I'm getting a little bit confused here and I'm not sure if it's because I am unwell, but it looks like the Erlang we use the runtime that is passed in on the command line wall with JavaScript we look at the config? Is that right? I think we want to pay attention to both.

brettkolodny: Yup you were right, fixed so that it will use the value passed to the CLI before defaulting to the config.

compiler-core/src/error.rs
```diff
@@ -2288,6 +2296,32 @@ issue in our tracker: https://github.com/gleam-lang/gleam/issues",
...
+                let text = format!("Invalid runtime for {} target: {}", target, invalid_runtime);
+
+                let hint = match target {
+                    Target::JavaScript => Some(format!(
+                        "available runtimes for {} are: {:?}",
+                        target,
+                        Runtime::VARIANTS
+                    )),
+                    Target::Erlang => Some(
+                        "You can not set a runtime for erlang. Did you mean to target javascript?"
```
lpil, line 2312:
```suggestion
                        "You can not set a runtime for Erlang. Did you mean to target JavaScript?"
```

compiler-core/src/error.rs
```diff
@@ -2288,6 +2296,32 @@ issue in our tracker: https://github.com/gleam-lang/gleam/issues",
...
             }
+            Error::InvalidRuntime {
+                target,
+                invalid_runtime,
+            } => {
+                let text = format!("Invalid runtime for {} target: {}", target, invalid_runtime);
+
+                let hint = match target {
+                    Target::JavaScript => Some(format!(
+                        "available runtimes for {} are: {:?}",
```
lpil, line 2307: This will print using Rust syntax, let's list the names of the runtimes here.

lpil (approved):
Fab! One little note inline. Would you also mind updating the changelog? Thank you

lpil (approved):
Oops only just saw this. Thank you!

lpil:
I have rebased this onto main, thanks!

Issue #1883, opened 2022-12-04
Formatter: Large numbers could be autoformatted to use `_` for improved community code consistency.


inoas:
So `54678023` => `54_678_023` and `54678023.2231` => `54_678_023.2231`?

lpil:
Sounds good 👍

inoas:
1. FP does not get underscored (not sure if it is valid even).
2. Leave hex, octa and binary ints as the programmer styled them, as discussed

PR #1889, opened 2022-12-10
Formatter: underscored numbers

inoas:
Closes https://github.com/gleam-lang/gleam/issues/1883


inoas:
@aslilac suggested code more like this:
```rust
fn add_underscores(base: &str) -> String {
    // Preserve negativeness
    let negative = base.starts_with('-');
    
    // Only retain digits (so we don't have to deal with existing underscores)
    let mut base = base.to_string();
    base.retain(|ch| ch.is_ascii_digit());
    
    // 1 underscore will be added per every 3 digits, allocate accordingly
    let mut result = String::with_capacity(base.len() * 4 / 3 + 1);
    
    if negative {
        result.push('-');
    }
    
    for (i, ch) in base.chars().enumerate() {
        result.push(ch);
        // Don't append an underscore if this is the last digit, only put one
        // after the right-most digit of each digit grouping.
        if i != base.len() - 1 && (i as isize - base.len() as isize) % 3 == -1 {
            result.push('_');
        }
    }
    
    result
}
```
Let me know if I should rebuilt or if the current impl. is fine.

compiler-core/src/format.rs
```diff
@@ -754,14 +754,55 @@ impl<'comments> Formatter<'comments> {
...
-            let suffix = "0".to_doc();
-            doc.append(suffix)
-        } else {
-            doc
+    fn float<'a>(&self, value: &'a str) -> Document<'a> {
+        let mut parts = value.split('.');
+        let integer_part = match parts.next() {
+            None => "",
+            Some(str) => str,
+        };
```
lpil, line 762: There is a method for this: `parts.next().unwrap_or_default()`

lpil (approved):
Awesome! Thank you

PR #1899, opened 2022-12-17
reformat any number but do not insert underscores for numbers with less than 5 digits

inoas:
before this PR in 25.3 short numbers could:
- either have underscores or not
- have underscores in irregular places


lpil (commented):
Hello! I do not want to enforce a format for numbers with fewer than 5 digits.

inoas:
so 12_34 is fine? or just 1_234 and/or 1234?

lpil:
Yes, if the programmer has a reason to write that.

inoas:
What would such a reason be?
Wouldn‘t such a reason not also apply to larger numbers? Thus far I fail to replicate the distinction.

I can see the underscoring integers with less than  digits would add more noise than help so this PR just removes those instead of allowing say 1_234 or 12_34/123_4.

I can also see for 1234/12345 and 1_234/12_345 as both desireable formats, so thus PR then wouldn‘t be sufficient, yet.

Should I work on the last case or do you prefer for the formatter to not correct what appears like typos to me?

lpil (approved):
OK, you have convinced me, let's format it without underscores.

Issue #1875, opened 2022-12-02
'gleam new' shouldn't create git repository if already in one


rlipscombe:
If I run `gleam new foo` while I'm already in a git work tree, it shouldn't run "git init".

lpil:
Sounds good!

rlipscombe:
More generally, I wonder whether this behaviour should be controlled by an option (and default to off) along with the `.github/` creation.

lpil:
Sounds good also 👍

bookshelfdave:
The `git init` seems to happen after the root is created, are you referring to checking parent directories recursively for a `.git` dir?
https://github.com/gleam-lang/gleam/blob/67def5b4a72b104a32f2a45c13f2cbf6d1109881/compiler-cli/src/new.rs#L57-L62

lpil:
Yup that's what we would do

rlipscombe:
It's what "cargo new" does: https://github.com/rust-lang/cargo/blob/master/src/cargo/ops/cargo_new.rs#L752 and https://github.com/rust-lang/cargo/blob/1023fa5c1ed51602933f7b938d7fde2e7598767c/src/cargo/util/vcs.rs#L11

rlipscombe:
Gleam just seems to run "git init", so "git rev-parse --is-inside-work-tree" would seem like the simplest comparable thing to do?

lpil:
Sounds good to me.

bookshelfdave:
> More generally, I wonder whether this behaviour should be controlled by an option (and default to off) along with another option for the `.github/` creation.

any thoughts on what these options should be named?

PR #1901, opened 2022-12-18
add `gleam new` flags to control git init and .github content creation

bookshelfdave:
Addresses the cli feature request from #1875.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,11 @@
 - 4 digit integers are now always formatted without underscores.
 - Running `gleam new` will skip `git init` if the new project directory is
   already part of a git work tree.
-- The `gleam new` command gains the `--skip-git` flag to skip initialization
-  of both git and .github content.
-- The `gleam new` command gains the `--skip-github` flag to skip initialization
-  of .github content.
+- The `gleam new` command gains the `--skip-git` flag to skip creation of
+  `.git/*`, `.gitignore` and `.github/*` files.
+- The `gleam new` command gains the `--skip-github` flag to skip creation of
+  `.github/*` files.
+
 
 ## v0.25.3 - 2022-12-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - 4 digit integers are now always formatted without underscores.
 - Running `gleam new` will skip `git init` if the new project directory is
   already part of a git work tree.
+- The `gleam new` command gains the `--git-init` and `--github-init` flags
+  to initialize git and .github (respectively). Unless these flags are
+  specified, new projects default to skipping initialization of both git
+  and .github content.
 
 ## v0.25.3 - 2022-12-16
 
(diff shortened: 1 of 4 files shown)


compiler-cli/src/new.rs
```diff
@@ -72,14 +72,22 @@ impl Creator {
...
+        }
+        if self.options.git_init {
+            crate::fs::git_init(&self.root)?;
+        }
 
         match self.options.template {
             Template::Lib => {
-                self.gitignore()?;
-                self.github_ci()?;
+                if self.options.git_init {
```
bookshelfdave, line 85: I wasn't sure if this should be lumped in with `git init` functionality. What if the dev is using something other than git?

compiler-cli/src/new.rs
```diff
@@ -72,14 +72,22 @@ impl Creator {
...
+                    self.gitignore()?;
```
bookshelfdave, line 86: I wasn't sure if this should be lumped in with `git init` functionality. What if the dev is using something other than git?

inoas:
IMHO at least git should be opt-out not opt-in as it seems to be a sane default.

Also it feels like github files make no sense if git isn't there so these options depend on each other.

```shell
--skip=git # also sets no-github
--skip=github # just opt out of github files but not git
```

```shell
--no-git # also sets no-github
--no-github
``

```shell
--skeleton=github # default
--skeleton=gitlab
--skeleton=none
--ci=github
--ci=gitlab
--ci=none
``

**Review by lpil — COMMENTED** [review]
Wonderful! Let's do as @inoas says and default to rendering everything.

We can have a `--skip-git` which disables both, and `--skip-github` which disables the `.github` stuff.

File: CHANGELOG.md [hunk 3]
```diff
@@ -5,6 +5,10 @@
 - 4 digit integers are now always formatted without underscores.
 - Running `gleam new` will skip `git init` if the new project directory is
   already part of a git work tree.
+- The `gleam new` command gains the `--skip-git` flag to skip initialization
+  of both git and .github content.
```
inoas, line 9: The `gleam new` command gains the `--skip-git` flag to skip initialization of `.git/*`, `.gitignore` and `.github/*` files.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
...
+- The `gleam new` command gains the `--skip-github` flag to skip initialization
+  of .github content.
```
inoas, line 11: The `gleam new` command gains the `--skip-github` flag to skip creation of `.github/*`  files.

compiler-cli/src/new.rs
inoas: If this skips .gitignore copy-over if skip-git is set, I think that's sane _imho_.

compiler-cli/src/main.rs
```diff
@@ -235,6 +235,14 @@ pub struct NewOptions {
         default_value = "lib"
     )]
     pub template: new::Template,
+
+    /// Skip git initialization and creation of .github content
```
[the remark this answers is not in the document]
bookshelfdave: per https://github.com/gleam-lang/gleam/pull/1901#pullrequestreview-1223052086, `--skip-git` should also skip `.github/*`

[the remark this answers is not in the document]
inoas: sure but the comment just says github

bookshelfdave:
changes pushed

lpil (approved):
Brilliant, thank you very much!

Issue #1758, opened 2022-09-09
Scientific notation for floating point numbers


manveru:
Would be really nice to have :)

lpil:
Agreed! With a syntax like `10e2`?

manveru:
Yeah, exactly :)

bookshelfdave:
👋 I'd like to take this on.

lpil:
Thank you!

PR #1903, opened 2022-12-20
Add support for floats with scientific notation

bookshelfdave:
#1758


compiler-core/src/parse/lexer.rs
```diff
@@ -617,6 +617,18 @@ where
         if self.chr0 == Some('.') {
             value.push(self.next_char().expect("lex_normal_number float"));
             value.push_str(&self.radix_run(10));
+
+            // If scientific:
```
bookshelfdave, line 621: I wasn't sure if `lex_normal_number` should be renamed, or if the scientific notation stuff should be pulled out into a separate function.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__int_literals.snap
```diff
@@ -11,6 +10,8 @@ function go() {
   0b00001111;
   0o17;
   0xF;
+  1.6e5;
```
inoas, line 13: Maybe add examples with more digits?
Also add examples with underscores if that makes any sense and check before-after in tests (I think the current formatter/agreement is to not touch anything but every-day notation)

bookshelfdave: > ... check before-after in tests

is there an existing example that I could follow?

inoas: https://github.com/gleam-lang/gleam/pull/1899/files#diff-eace0c72b35f221894649a2a178e1799e60899a428b0db25e1aeca46562c5a0dR1205-R1238

test/language/src/main.gleam
```diff
@@ -108,6 +108,45 @@ fn float_tests() -> List(Test) {
...
+
+
+    // scientific notation
+    basic_addition(0.0e0, 0.0, 0.0),
+    basic_addition(0.0, 0.0e0, 0.0),
+    basic_addition(0.0e-0, 0.0, 0.0),
+    basic_addition(0.0, 0.0e-0, 0.0),
+    basic_addition(1.0e3, 1.0, 1001.0),
+    basic_addition(5.0, 1.0e3, 1005.0),
+    basic_addition(1.0e5, -3.0e5, -200000.0),
```
inoas, line 120: -200_000.0

compiler-core/src/erlang/tests/numbers.rs
```diff
@@ -32,3 +32,28 @@ pub fn main() {
...
+const i = 100.001e523
+const j = -100.001e-523
+
+const k = 100.001e1_230
+const l = -100.001e-1_230
+
+const m = 100.001e123_456_789
+const n = -100.001e-123_456_789
+
+pub fn main() {
```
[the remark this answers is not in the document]
lpil: It's to test the correct Erlang is generated

compiler-core/src/javascript/tests/numbers.rs
```diff
@@ -12,6 +12,14 @@ fn go() {
...
     0o17
     0xF
+    0.01e-1
+    0.01e-0
+    -10.01e-1
+    -10.01e-0
+    100.001e523
+    -100.001e-523
+    100.001e123_456_789
+    -100.001e-123_456_789
```
lpil, line 22: Could these go in a new test please? This one is already testing far too many things! Thanks

compiler-core/src/format/tests.rs
```diff
@@ -1241,6 +1241,43 @@ fn expr_float() {
...
+  1.0e-10
+  -1.0e10
+  -11.0e-10
+  1.0e100
+  1.0e-100
+  -1.0e100
+  -11.0e-100
+  1.0000e100
+  1.0000e100_100
+}
```
lpil, line 1280: What's the rewrite here? It looks the same to me?

compiler-core/src/parse/lexer.rs
lpil: I think `lex_decimal_number` might be a more appropriate name?

compiler-core/src/erlang/tests/numbers.rs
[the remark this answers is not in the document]
inoas: So the variable references in main() are required because else the constants don't exist in erl representation? If that's the case sorry for the noise.

lpil (commented):
Really nice work! I've left a few small comments inline

compiler-core/src/erlang/tests/numbers.rs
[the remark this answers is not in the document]
lpil: Yup that's right 👍

compiler-core/src/format/tests.rs
inoas: I think the idea is to show that there is no rewrite for the underscores and that it expected to stay the same?

lpil: This macro is for rewrites, the `assert_format` macro is for testing formats otherwise. I wasn't sure which it was here

bookshelfdave: My understanding is that it's to show that the formatter doesn't do anything here. If they're unneeded, I'm happy to remove!

lpil: This macro is for rewrites, the `assert_format` macro is for testing formats otherwise. I wasn't sure which it was here

compiler-core/src/javascript/tests/numbers.rs
bookshelfdave: yes, and I just noticed that I put them in the `int_literals` test by mistake.

lpil (approved):
Fantastic, thank you!

lpil:
Oops, there's an indentation issue in the tests @bookshelfdave

bookshelfdave:
I left out a .snap file, I'll add and push in a moment.

PR #1908, opened 2022-12-24
docs generator accessibility, search discoverability, layout stability

inoas:
This PR improves accessibility, discoverability and layout stability (removal of almost all cumulative layout shifts, preload directives, handling of initial dark mode toggle entirely via CSS).

1. added `page_meta_description` to `Template`.
   - used as meta-description html-tag in docs. 
   - on regular pages, renders `LIB_NAME vLIB_VERS · LIB_DESCRIPTION`, where `LIB_DESCRIPTION` = `gleam.toml` => `description` property.
   - on module doc pages, renders:
     - `LIB_NAME vLIB_VERS · MODULE_NAME`, **if** given module doesn't have a module docblock, or
     - `LIB_NAME vLIB_VERS · MODULE_NAME · MODULE_DESCRIPTION`, **if** given module has a module docblock. The module docblock passed through markdown, stripped off its html-tags and newlines, and trimmed by its its wrapping whitespaces.
2. use data attributes instead of class attribuets to hold state on the html instead of the body element:
   - the base documentElement (html-tag) is always available => no need to wait for the tag to become avaiable and nest javascript in it (reason for this change).
   - html-data-attributes are easier to toggle than classes => removed a bit of JS.
3. added preload directives to a lot of common resources.
   - autoloading of dark or light syntax-stylesheet per media query => we don't need to wait for JS execution.
4. Cumulative Layout Shifts (CLS) fixes
   - code blocks gain `display: block` and the same padding they would have after hljs kicks in. Result: The layout stops to jump vertically on first/slow load.
   - shield badge row gains a static hight to stop vertical jumping.
5. accessibility 
   - added aria labels to buttons to improve accessibility
   - converted pride button into a semantically true button
   - added lang-attribute to html-elem to improve lighthouse score
6. browser ui theme-color now correctly switches when the main page theme is switched.


inoas:
meta description start page:

Screenshot 2022-12-24 at 01 16 14

meta description module with docblock

Screenshot 2022-12-24 at 01 15 12

meta description module without docblock

Screenshot 2022-12-24 at 01 15 26

inoas:
### Lighthouse results

Screenshot 2022-12-23 at 23 52 08

Screenshot 2022-12-23 at 23 47 28

Screenshot 2022-12-23 at 23 47 26

Screenshot 2022-12-23 at 23 47 23

Screenshot 2022-12-23 at 23 52 09

Screenshot 2022-12-23 at 23 52 11

inoas:
## Lighthouse Results / Mobile

### Mobile / Live

Screenshot 2022-12-23 at 23 47 23

### Mobile / Main

Screenshot 2022-12-23 at 23 52 09

### Mobile / This PR

Screenshot 2022-12-23 at 23 52 11

compiler-core/src/docs.rs
```diff
@@ -198,21 +201,35 @@ pub fn generate_html(
             url: format!("{}.html", module.name),
         });
 
+        let rendered_documentation = render_markdown(&module.ast.documentation.iter().join("\n"));
+        let rendered_text_documentation =
+            manipulate::trim(rendered_documentation._strip_tags().as_str(), "\n ");
+        let page_meta_description = &manipulate::replace_all(
```
inoas, line 207: Is there a better/more elegant way than to use `voca_rs` crate?

michaeljones:
This PR shows the issue with the constructor name code sample having too large a padding value which #1911 attempts to address.

If you build the docs for the standard lib, go to `gleam/order` and look at the list of the constructors for `Order` you'll see it. When inspecting the padding on the list entries you see that the `.constructor-name > code`  rule which sets the padding to `var(--small-gap)` isn't being applied as the `pre code.hljs` rule from `atom one dark` or `atom one light` has a higher css specificity and is applied instead.

compiler-core/src/docs.rs
lpil: What does this do?

compiler-core/templates/docs-css/index.css
```diff
@@ -439,8 +440,8 @@ p code {
   font-size: calc(0.8 * var(--sidebar-toggle-size));
 }
 
-body.drawer-closed .label-open,
-body.drawer-open .label-closed {
+html[data-drawer="closed"] .label-opened,
+html[data-drawer="opened"] .label-closed {
```
lpil, line 444: Why has this changed?

compiler-core/templates/docs-css/index.css
```diff
@@ -1229,3 +1230,9 @@ body.theme-dark .pride .blue {
     padding-top: 0;
   }
 }
+
+/* stabilizes computed-height of remote loaded images */
+main > h1:first-child + p {
```
lpil, line 1235: This is a very precise selector, what it it?

compiler-core/templates/docs-js/index.js
```diff
@@ -25,7 +25,7 @@ window.Gleam = (function () {
     return Array.from(arguments).reduce(
       (acc, name) =>
         `${acc}
-        <svg class="icon icon-${name}"><use xlink:href="#icon-${name}"></use></svg>`,
+        <svg alt="icon-${name}" class="icon icon-${name}"><use xlink:href="#icon-${name}"></use></svg>`,
```
lpil, line 28: `alt` isn't a valid tag on SVGs, and I don't think this value will help people using screen readers

compiler-core/templates/documentation_layout.html
```diff
@@ -1,18 +1,41 @@
...
     <title>{{ page_title }}</title>
+    <meta name="description" content="{{project_name}} v{{project_version}} · {{page_meta_description}}"/>
+    <!-- Preload most used font variants -->
+    <link rel="font" as="font" href="{{ unnest }}/fonts/karla-v23-regular-latin.woff2" type="font/woff2"/>
+    <link rel="font" as="font" href="{{ unnest }}/fonts/karla-v23-bold-latin.woff2" type="font/woff2"/>
+    <link rel="font" as="font" href="{{ unnest }}/fonts/ubuntu-mono-v15-regular-latin.woff2" type="font/woff2"/>
+    <!-- Preload base script and version switcher -->
+    <link rel="script" as="script" href="{{ unnest }}/js/index.js?v={{ rendering_timestamp }}" type="application/javascript"/>
+    <link rel="script" as="script" href="{{ unnest }}/docs_config.js" type="application/javascript"/>
+    <!-- Preload source code highlighting -->
```
lpil, line 15: I don't think we should load these first, we prioritise content over syntax highlighting

compiler-core/templates/documentation_layout.html
```diff
@@ -1,18 +1,41 @@
...
+<html lang="en" data-theme="light" data-drawer="closed" data-prewrap="off">
   <head>
     <meta charset="utf-8"/>
     <meta name="viewport" content="width=device-width, initial-scale=1"/>
     <title>{{ page_title }}</title>
+    <meta name="description" content="{{project_name}} v{{project_version}} · {{page_meta_description}}"/>
+    <!-- Preload most used font variants -->
+    <link rel="font" as="font" href="{{ unnest }}/fonts/karla-v23-regular-latin.woff2" type="font/woff2"/>
+    <link rel="font" as="font" href="{{ unnest }}/fonts/karla-v23-bold-latin.woff2" type="font/woff2"/>
+    <link rel="font" as="font" href="{{ unnest }}/fonts/ubuntu-mono-v15-regular-latin.woff2" type="font/woff2"/>
```
lpil, line 11: Where were these before? I can't see them in the PR

lpil (commented):
This is a hard PR to review! There's a lot of unrelated changes and it's not clear what the motivations are or which changes are related to what feature.

Each feature or improvement or bug fix should get its own PR, it is much more understandable that way.

compiler-core/src/docs.rs
inoas: It creates the meta description from the module documentation doc block?

compiler-core/templates/docs-css/index.css
inoas: Stabilizes computed-height of remote loaded images of the shield badges.
I agree this should be changed, but I wouldn't know where they get inserted.

compiler-core/templates/docs-js/index.js
inoas: Right
https://stackoverflow.com/questions/4697100/accessibility-recommended-alt-text-convention-for-svg-and-mathml
swap for title?

compiler-core/templates/documentation_layout.html
inoas: Would be good if we can move in in all the things that stop cumulative layout shifts via HLJS application. Then the hljs application just adds colors and nothing else changes. I will remove this.

compiler-core/templates/documentation_layout.html
inoas: These are just woff2 preloads. The entire fonts are loaded in `index.css`.

inoas:
> Each feature or improvement or bug fix should get its own PR, it is much more understandable that way.

... okay!

compiler-core/templates/docs-css/index.css
inoas: the data attrib lives on the html elem and is easily toggable and available very early versus classes on body which are available after `<head>` not before.

PR #1911, opened 2022-12-27
Adjust constructor-name css

michaeljones:
This css is meant to be applied but it currently isn't due to classes in the atom-one theme overriding the padding on `pre code.hljs` blocks.

We add the extra class here to up the specificity of this rule and have it win against the atom-one classes.

We should either have this or consider removing the rule entirely as it currently isn't being applied.

### DIFF
--- a/compiler-core/templates/docs-css/index.css
+++ b/compiler-core/templates/docs-css/index.css
@@ -568,7 +568,9 @@ body.drawer-open .label-closed {
   margin: 0;
 }
 
-.constructor-name > code {
+/* .hljs included to have higher specificity than
+   matching classes in atom-one themes */
+.constructor-name > code.hljs {
   padding: var(--small-gap);
 }
 


inoas:
Does the problem you describe appear if you change this?

https://github.com/gleam-lang/gleam/pull/1908/files#diff-0a2b885976116f3f1750e64dbd5bad302d7261a174ec104e53b48940ee765e67L252-R253

michaeljones:
Adjusting the padding there would work but would change it for all code blocks and not just those inside `.constructor-name`? But I don't know if there is a better way to do it. I'm not attached to this approach just flagging an issue with the layout (as far as I can tell.)

inoas:
> Adjusting the padding there would work but would change it for all code blocks and not just those inside `.constructor-name`? But I don't know if there is a better way to do it. I'm not attached to this approach just flagging an issue with the layout (as far as I can tell.)

At the moment the set padding means that once hljs kicks in the layout jumps around.
The padding I have set in the other PR makes sure that does not happen (no cumulative layout shift by the code blocks). I had hoped that it works for constructors too, I will check.

michaeljones:
Ah, I hadn't seen your PR. I didn't take the time to look. Maybe we should close this until yours is in? I imagine you'll resolve the issue.

Certainly best to avoid layout shifts.

michaeljones:
I'll close it and we can reopen if needed.

inoas:
@michaeljones If you have the time please check out if it covers your problem. If not could you let me know which docs and specific module and/or function you used to test it out, so I can replicate it locally.

Issue #1905, opened 2022-12-22
Allow hiding internal modules from package documentation


aslilac:
If I have, say, `nakai/internal/type_that_is_not_intended_for_consumers`, which I want to be its own module to keep the code organized and idiomatic, I have to accept a couple of unfortunate things:

- This module is going to be accessible by consumers of my package
- This module is going to be shown in my public documentation

There's potentially a nice solution here that solves both together, but the documentation bit could also be solved separately

Discord conversation for reference: https://discord.com/channels/768594524158427167/768594524158427170/1055625580487377059

lpil:
I agree, this is something we want to improve.

I think we can split this into two features in order to make it easier to tackle:
1. Omitting internal modules from documentation
2. More fine-grained privacy control (so that code can be made public only the the same package, only to a subset of a project, etc).

2 is a much larger and less clear item of work that requires more research, so I'll consider this issue to be convering 1.

Thinking about this for a while my favourite so far is to have a `documentation.hidden_modules` or similar which could be used to filter out whatever modules from the documentation.

```toml
name = "nakai"
version = "1.0.0"

[documentation]
hidden_modules = [
  "nakai/internal/*",
]
```

lpil:
I would advise reading the documentation in the source code if you are working in the project.

aslilac:
the use of globs immediately brings up some questions for me:

- would `nakai/internal/cool/stuff` be excluded?
- would `nakai/internal` be excluded?

and if I instead had
```
[documentation]
hidden_modules = [
  "nakai/internal",
]
```

- would that only ignore `nakai/internal` and not anything underneath it?

lpil:
All behaviour is up for grabs!

PR #1922, opened 2022-12-29
add ability to hide modules from docs

massivefermion:
addresses #1905 

Just to make it clear, right now the behavior is that adding `package/internal` to the `hidden_modules` list, where `internal` is a directory, will hide everything inside `internal`. But if `internal` is not a directory and instead there is a `package/internal.gleam`, then only `package/internal.gleam` will be hidden. And I think if both are present, then both will be hidden. This seems to be the simplest behavior to implement.

### DIFF
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.25.3"
 authors = ["Louis Pilfold <louis@lpil.uk>"]
 edition = "2021"
 license-file = "LICENCE"
- 
+
 [dependencies]
 # Immutable data structures
 im = "15.1.0"
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -1,9 +1,9 @@
 [package]
-name = "gleam-core"
-version = "0.25.3"
 authors = ["Louis Pilfold <louis@lpil.uk>"]
 edition = "2021"
 license-file = "LICENCE"
+name = "gleam-core"
+version = "0.25.3"
 
 [dependencies]
 # Immutable data structures
@@ -26,16 +26,16 @@ lazy_static = "1.4.0"
 # Levenshtein string distance for typo suggestions
 strsim = "0.10.0"
 # Data (de)serialisation
-serde = { version = "1.0.130", features = ["derive"] }
+serde = {version = "1.0.130", features = ["derive"]}
 serde_json = "*"
 # Cap'n Proto binary format runtime
 capnp = "0.14.3"
 # Enum trait impl macros
-strum = { version = "0.24.0", features = ["derive"] }
+strum = {version = "0.24.0", features = ["derive"]}
 # Template rendering
 askama = "0.10.5"
 # Markdown parsing
-pulldown-cmark = { version = "0.8.0", default-features = false }
+pulldown-cmark = {version = "0.8.0", default-features = false}
 # Non-empty vectors
 vec1 = "1.8.0"
 # Hex package manager client
@@ -59,7 +59,7 @@ tar = "0.4.37"
 # gzip compression
 flate2 = "1.0.22"
 # Helper for wrapping text onto lines based upon width
-textwrap = { version = "=0.15.0", features = ["terminal_size"] }
+textwrap = {version = "=0.15.0", features = ["terminal_size"]}
 # base encoding
 base16 = "0.2.1"
 # toml config file parsing
@@ -68,6 +68,8 @@ toml = "0.5.8"
 spdx = "0.8.0"
 # Binary format de-serialization
 bincode = "1.3.3"
+# cross platform single glob and glob set matching
+globset = "0.4.9"
 
 [build-dependencies]
 # Data (de)serialisation
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
+name = "gleam-core"
+version = "0.25.3"
 authors = ["Louis Pilfold <louis@lpil.uk>"]
 edition = "2021"
 license-file = "LICENCE"
-name = "gleam-core"
-version = "0.25.3"
-
+ 
 [dependencies]
 # Immutable data structures
 im = "15.1.0"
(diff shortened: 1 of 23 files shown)


compiler-core/src/build.rs
```diff
@@ -98,9 +98,14 @@ pub struct Package {
 }
 
 impl Package {
-    pub fn attach_doc_and_module_comments(&mut self) {
+    pub fn attach_doc_and_module_comments(&mut self, hidden_modules: &Vec<String>) {
         for mut module in &mut self.modules {
-            module.attach_doc_and_module_comments();
+            if !hidden_modules
+                .iter()
+                .any(|pattern| module.name.starts_with(pattern))
```
lpil, line 105: Does this work on Windows?

Would `one_two` get ignored by a pattern of `one`? I think we likely want to make the globbing explicit using git-like `one/*` syntax.

lpil (commented):
Nice! Though I'm not entirely sure this is the right API. It would be good to get some input from the community.

A unit test for the filtering function would be fab. We could extract the "should this be included" logic into a function and test just that seeing as we don't have tests for the full documentation generation

massivefermion:
I made some changes, and now glob matching is used. I should also apologize if the changes I made are a mess, but I'm still at the "fighting with the compiler" stage of my rust journey.
I can add a test if the functionality is acceptable, although I'm still not sure how to test functionalities that require a `gleam.toml` file.
And if there is still any doubt about windows support, someone else should confirm it because I'm on linux.

compiler-core/Cargo.toml
```diff
@@ -68,6 +68,8 @@ toml = "0.5.8"
 spdx = "0.8.0"
 # Binary format de-serialization
 bincode = "1.3.3"
+# cross platform single glob and glob set matching
+globset = "0.4.9"
```
lpil, line 72: Could you revert the other changes to this file please

compiler-core/src/build.rs
```diff
@@ -99,8 +99,11 @@ pub struct Package {
 
 impl Package {
     pub fn attach_doc_and_module_comments(&mut self) {
+        let hidden_modules = self.specify_hidden_modules();
         for mut module in &mut self.modules {
-            module.attach_doc_and_module_comments();
+            if !hidden_modules.contains(&module.name) {
+                module.attach_doc_and_module_comments();
+            }
```
lpil, line 106: We still want to attach documentation for hidden modules as it's not only used by the HTML documentation generator, so best revert this

compiler-core/src/config.rs
```diff
@@ -138,6 +140,19 @@ impl PackageConfig {
             }
         })
     }
+
+    pub fn specify_hidden_modules(&self, modules: Vec<String>) -> Vec<String> {
```
lpil, line 144: Could we get some documentation please?

I think this likely also wants to take a reference rather than ownership as it doesn't do anything other than reference and then return the data. It could return a vector of &str too

compiler-core/src/config.rs
```diff
@@ -138,6 +140,19 @@ impl PackageConfig {
...
+        let mut glob_builder = GlobSetBuilder::new();
+        for pattern in self.documentation.hidden_modules.iter() {
+            let _ = glob_builder.add(Glob::new(pattern).expect("invalid pattern!"));
+        }
+        let glob_set = glob_builder.build().expect("problem!");
```
lpil, line 149: Could we get a detailed expect message here please?

compiler-core/src/config.rs
```diff
@@ -138,6 +140,19 @@ impl PackageConfig {
             }
         })
     }
+
+    pub fn specify_hidden_modules(&self, modules: Vec<String>) -> Vec<String> {
+        let mut glob_builder = GlobSetBuilder::new();
+        for pattern in self.documentation.hidden_modules.iter() {
+            let _ = glob_builder.add(Glob::new(pattern).expect("invalid pattern!"));
```
lpil, line 147: Can this ever happen? This is user input right? If so we'll need a user facing error message.

Does the glob type implement serde? We could use it directly in the config struct definition

compiler-core/src/docs.rs
```diff
@@ -25,7 +25,16 @@ pub fn generate_html(
...
-    let modules = analysed.iter().filter(|module| !module.is_test());
+    let hidden_modules = config.specify_hidden_modules(
+        analysed
+            .iter()
+            .map(|module| module.name.clone())
+            .collect_vec(),
+    );
+    let modules = analysed
+        .iter()
+        .filter(|module| !module.is_test() && !hidden_modules.contains(&module.name));
```
lpil, line 36: Something like this may be nicer

```rust
let modules = analysed
    .iter()
    .filter(|module| !module.is_test())
    .filter(|module| !config.is_hidden_module(module))

File: test-package-compiler/src/generated_tests.rs [hunk 8]
```diff
@@ -183,144 +183,144 @@ fn imported_external_fns() {
...
+fn variable_or_module() {
     let output =
-        crate::prepare("./cases/opaque_type_construct");
+        crate::prepare("./cases/variable_or_module");
     insta::assert_snapshot!(
-        "opaque_type_construct",
+        "variable_or_module",
         output,
-        "./cases/opaque_type_construct"
+        "./cases/variable_or_module"
```
lpil, line 324: Looks like I need to sort this or something to make the order stable! Oops my bad

lpil (commented):
Nice! I've left some comments

massivefermion:
Sorry about the `compiler-core/Cargo.toml`, I wasn't careful and my toml formatter was on. I turned it off and copied the original file, with the addition of `globset`, but it didn't seem to work. I also tried doing it in github's editor, but didn't seem to work again!
Anyway, turns out the Glob type does implement the serde trait, so now the globs are created when the `gleam.toml` file is being read. The implementation is really simplified now.

lpil (commented):
Fab! Could we have some tests for the filtering please?

massivefermion:
I need some help for that. Because `is_hidden_module` accepts a `Module` as an argument and not its name, I need to create modules for testing it. But all the fields are required to create a module. Right now I can provide some dummy values for all fields except for `ast`. I don't know how to create a dummy value for it!

lpil:
How about we change it to just accept the module name? That'll be a bit easier.

massivefermion:
It's done!

lpil (commented):
Looks fab! Thank you! Would you mind updating the changelog also? Thanks

lpil (approved):

lpil (approved):
Thanks!

lpil:
I've rebased this in, thank you!

PR #1930, opened 2022-12-29
Remove generalise_statement function

michallepicki:
It seems mostly unnecessary as functions are generalised inside `infer_statement` already, and its main responsibility seemed to be adding functions to module's interface.

I couldn't figure out how to use the `field_map` twice more nicely, sorry!

### DIFF
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -628,12 +628,6 @@ pub fn infer_module(
         statements.push(statement);
     }
 
-    // Generalise functions now that the entire module has been inferred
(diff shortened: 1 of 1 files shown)


lpil:
This is still needed as functions are not always generalised in the first pass, but either way this generalisation code is going to be entirely written as there's a soundness bug in the approach used here.

michallepicki:
In that case, I think some tests that exercise this codepath and fail on this branch should be added?

lpil:
Yup, there will be more added with the rewrite.

Issue #1728, opened 2022-08-15
Restricted Project Naming


coopbri:
## Description

When creating a new Gleam project with `gleam new`, there are situations where the `gleam` CLI will respond with errors due to disallowed project naming patterns:

<img src="https://user-images.githubusercontent.com/20056195/184559681-372e9ce0-7d7a-49c8-a650-470b8995cf8e.png" width="400px" />

Context: based on [this Twitter thread](https://twitter.com/0xcoop/status/1558609481352859648)

## Ideas

1.  Document this in the getting started guide/docs (may be a good temporary fix)
    - Mention reserved keywords situations (`gleam` prefix) as well as underscore convention
2.  Combine the error messages (though might be too much output to the user)
3.  Allow more project naming patterns; decouple Gleam project names from their parent folder name, allowing project directory names such as `gleam-demo`, `gleam`, and `gleam_demo`
    - I like this idea the most of the three proposed here; this is allowed in, for example, a new Rust/Cargo project. You can create projects with analogous name `rust-demo`, `rust`, and `rust_demo`:

      <img src="https://user-images.githubusercontent.com/20056195/184559734-97e7309e-8c17-425f-87af-494a706ee46a.png" width="500px" />

Let me know if you have other ideas. I'm happy to take this on after a consensus is reached!

lpil:
Thanks for the report, this must have been very annoying.

I would suggest that in these instances we could print something like this

"We were not able to create your project as `gleam` as etc etc....

Would you like is to create a project named `my_gleam` in a directory named `gleam`? (Y/n)"

> decouple Gleam project names from their parent folder name, allowing project directory names such as gleam-demo, gleam, and gleam_demo

This is already the case! Unfortunately we take the project name as the argument here and use the director name as the same.

inoas:
Maybe instead of saying what it should not be it should say what it should be:

Like:

> Project name must be **lowercase**, **alphanumeric**, may **not** start with a number and may not be named **gleam** or begin with **gleam_**.

coopbri:
> Thanks for the report, this must have been very annoying.
> 
> I would suggest that in these instances we could print something like this
> 
> "We were not able to create your project as `gleam` as etc etc....
> 
> Would you like is to create a project named `my_gleam` in a directory named `gleam`? (Y/n)"
> 
> > decouple Gleam project names from their parent folder name, allowing project directory names such as gleam-demo, gleam, and gleam_demo
> 
[...]

PR #1932, opened 2022-12-30
Update error message for invalid names when creating new gleam project

pascalwengerter:
Figured I could potentially fix parts of #1728 while I'm familiarizing myself with Rust/Gleam...a bit confused to not have any test failures after the changes?

### DIFF
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -1,6 +1,6 @@
 use gleam_core::{
     erlang,
-    error::{Error, FileIoAction, FileKind, InvalidProjectNameReason},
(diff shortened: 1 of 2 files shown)


lpil:
Hi @pascalwengerter ! There are no tests for that part of the project generator currently. If you're like to write some that'd be cool!

lpil (commented):
I would like to retain the detailed error message as was there previously as without it is unclear what the issue is. Suggesting an alternative and also printing some more information would be good.

lpil:
Closing due to inactivity, please feel free to reopen. Thank you!

PR #1935, opened 2022-12-31
feat: scoped and global lock files

### LINKED ISSUE #1933: GIve each mode + target their own build lock
So instead of always using `build/gleam-compile.lock` we check the `Mode` and do something like this:

| Mode | Target | Lock |
| --- | --- | --- |
| Dev | Erlang | build/dev/erlang/build.lock |
| Prod | Erlang | build/prod/erlang/build.lock |
| Lsp | Erlang | build/lsp/erlang/build.lock |
| Dev | JavaScript | build/dev/javascript/build.lock |
| Prod | JavaScript | build/prod/javascript/build.lock |
| Lsp | JavaScript | build/lsp/javascript/build.lock |

Checking deps versions and downloading deps is still global so we'll need to use `build/packages.lock` for that

HarryET:
Closes #1933

### DIFF
--- a/compiler-cli/src/build_lock.rs
+++ b/compiler-cli/src/build_lock.rs
@@ -10,10 +10,10 @@ pub(crate) struct BuildLock {
 
 // TODO: return errors rather than crashing.
 impl BuildLock {
-    pub fn new_scoped(mode: &Mode, target: &Target) -> Result<Self> {
-        let mut build = paths::build();
-        build = build.join(mode.to_string());
-        build = build.join(target.to_string());
+    pub fn new_scoped(mode: Mode, target: Target) -> Result<Self> {
+        let build = paths::build()
+            .join(mode.to_string())
+            .join(target.to_string());
         crate::fs::mkdir(&build)?;
         Ok(Self {
             path: build.join("build.lock"),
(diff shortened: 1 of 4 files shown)


compiler-cli/src/build_lock.rs
```diff
@@ -9,11 +10,21 @@ pub(crate) struct BuildLock {
 
 // TODO: return errors rather than crashing.
 impl BuildLock {
-    pub fn new() -> Result<Self> {
+    pub fn new_scoped(mode: &Mode, target: &Target) -> Result<Self> {
```
lpil, line 13: These implement `Copy` so can pass them in rather than as references

compiler-cli/src/build_lock.rs
```diff
@@ -9,11 +10,21 @@ pub(crate) struct BuildLock {
...
+        let mut build = paths::build();
+        build = build.join(mode.to_string());
+        build = build.join(target.to_string());
```
lpil, line 16: Favour method chains over reassignment

`x.join(y).join(z)`

compiler-cli/src/build_lock.rs
```diff
@@ -33,8 +44,50 @@ impl BuildLock {
...
+    let _guard1 = lock.lock(&crate::telemetry::NullTelemetry);
+    println!("Locked!")
+}
+
+#[test]
+fn locking_lsp_javascript() {
+    let lock = BuildLock::new_scoped(&Mode::Lsp, &Target::JavaScript).expect("make lock");
+    let _guard1 = lock.lock(&crate::telemetry::NullTelemetry);
+    println!("Locked!")
+}
```
lpil, line 93: EOF newline here please

compiler-cli/src/build_lock.rs
```diff
@@ -33,8 +44,50 @@ impl BuildLock {
 pub(crate) struct Guard(fslock::LockFile);
 
 #[test]
-fn locking() {
-    let lock = BuildLock::new().expect("make lock");
+fn locking_global() {
+    let lock = BuildLock::new_global().expect("make lock");
+    let _guard1 = lock.lock(&crate::telemetry::NullTelemetry);
+    println!("Locked!")
```
lpil, line 50: Lol this test is a bit rubbish, why did I write it like this? Weird that it just prints. I should test to see if the lock can be aquired a second time by a different locker.

lpil (commented):
Really nice!

compiler-cli/src/build_lock.rs
HarryET: done

compiler-cli/src/build_lock.rs
HarryET: fmt fixed that

compiler-cli/src/build_lock.rs
HarryET: done

compiler-cli/src/build_lock.rs
HarryET: Want me to do that or this a future issue?

lpil: Up to you :)

HarryET: left as a future issue to limit scope of this PR: https://github.com/gleam-lang/gleam/issues/1936

lpil (commented):
Steller work!

lpil (approved):

Issue #1934, opened 2022-12-31
Ints do not get formatted in patterns and constants


lpil:
```gleam
assert 123456789876543 = ...
```
```gleam
const x = 123456789876543
```

Would you like to take this one @inoas ?

inoas:
I can at least give it a try, yes!

inoas:
If I am not mistaken `const foo = 1.` does not get reformatted correctly neither?

lpil:
AH yes, floats also

inoas:
Also for patterns, I hope I didn't miss anything else. I am a but confused that there are different places for this.

lpil:
They are different things.

PR #1938, opened 2023-01-01
Fix int and float format in consts and patterns

inoas:
Fixes https://github.com/gleam-lang/gleam/issues/1934


compiler-core/src/format/tests.rs
```diff
@@ -1123,6 +1123,89 @@ fn expr_int() {
 "#
     );
 
+    assert_format_rewrite!(
```
lpil, line 1126: These assertions are very hard to read. Where're the rewrites here? Could we break it down into multiple assertions please. There should also only be one assertion per test.

compiler-core/src/format.rs
```diff
@@ -1325,9 +1327,9 @@ impl<'comments> Formatter<'comments> {
     fn pattern<'a>(&mut self, pattern: &'a UntypedPattern) -> Document<'a> {
         let comments = self.pop_comments(pattern.location().start);
         let doc = match pattern {
-            Pattern::Int { value, .. } => value.to_doc(),
+            Pattern::Int { value, .. } => self.int(value.as_str()),
 
-            Pattern::Float { value, .. } => value.to_doc(),
+            Pattern::Float { value, .. } => self.float(value.as_str()),
```
lpil, line 1332: Is `as_str` needed here?

compiler-core/src/format/tests.rs
inoas: Do we need to test all the number rewrites anyway, it seems we test the same thing again and again? Or would it be suffcient to a few of the rewrites instead of the whole pattern?

compiler-core/src/format.rs
inoas: No, it simply existed before for the number formatter and I wasn't sure if you wanted to keep it like before.

compiler-core/src/format/tests.rs
lpil: I don't know what you mean. Tests should cover the changed functionality as appropriate

inoas: Let me know if its good now or if you just want one line `let #(1_234, _) = #(1_234, Nil)` to `let #(1234, _) = #(1234, Nil)` instead of a subset of the int underscore formatter that I picked now.

inoas:
Is this good now?

lpil (approved):
Perfect! Thank you

Issue #1896, opened 2022-12-15
Prohibit compilation if there is a .gleam file and a .erl file with the same file/module name


inoas:
in `src/gleamish_extlib.gleam`, I had:

```gleam
import gleam/io

pub fn main() {
  io.println("Welcome to a Gleamish ExtLib!")
  io.println(
    "Notice: This package is meant to be used only as library and not a stand-alone application.",
  )
}
```
as well as a `gleamish_extlib.erl` with `-module(gleamish_extlib).` and `gleam test --target javascript` worked but `gleam test --target erlang` failed and I didn't know why until I transferred one by one files from the current to an empty new project.

lpil:
Sounds good!

bookshelfdave:
Is it good enough to check for .erl files with the same basename [here](https://github.com/gleam-lang/gleam/blob/c43c9dc989936b93ce99382a0c71e4b3f9b394d6/compiler-core/src/build/package_loader.rs#L144) and [here](https://github.com/gleam-lang/gleam/blob/c43c9dc989936b93ce99382a0c71e4b3f9b394d6/compiler-core/src/build/package_loader.rs#L155)?

Also, is there a good way to test this type of compile error?

lpil:
Closing as duplicate of https://github.com/gleam-lang/gleam/issues/1267

PR #1942, opened 2023-01-04
raise an error if a Gleam and Erlang file exists with the same basename

bookshelfdave:
#1896 

I wasn't sure of the best way to test this (although it seems to work locally), is there an example of something similar in the source tree that I could use?


lpil:
We will need a version of this that works for nested modules and for modules in other packages.

PR #1949, opened 2023-01-10
Do not remove new lines in expression blocks within case branches

### LINKED ISSUE #1919: Formatter removes new lines in expression blocks within case branches
Gleam 0.25.3

Saving this:

```gleam
fn foo() {
  "Foo"
  |> io.println

  "Bar"
  |> io.println

  let _quux = {
    "Foo"
    |> io.println

    "Bar"
    |> io.println
    Nil
  }

  let bool = True
  case bool {
    True -> {
      "Foo"
      |> io.println

      "Bar"
      |> io.println

      Nil
    }
    False -> Nil
  }
}
```

Results in this:

```gleam
fn foo() {
  "Foo"
  |> io.println

  "Bar"
  |> io.println

  let _quux = {
    "Foo"
    |> io.println

    "Bar"
    |> io.println
    Nil
  }

  let bool = True
  case bool {
    True -> {
      "Foo"
      |> io.println
      "Bar"
      |> io.println
      Nil
    }
    False -> Nil
  }
}
```

cobbinma:
Closes #1919 

Seems a bit too good to be true but didn't break any tests...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 - Fixed `int` and `float` formatting in `const`s and patterns.
 - Fixed a bug where piping into a function capture expression with a pipe as one
   of the arguments would produce invalid Erlang code.
+- Formatter no longer removes new lines in expression blocks within case branches
 
 ## v0.25.3 - 2022-12-16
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Huh, I guess whatever that code was for doesn't matter any more. Thanks!

Issue #1951, opened 2023-01-12
Ring fails to compile on CI for Linux musl


lpil:
https://github.com/gleam-lang/gleam/actions/runs/3901221151/jobs/6662804910

This is blocking the release!

PR #1952, opened 2023-01-12
regenerate cache-apt-pkgs-action cache

tranzystorek-io:
Should fix https://github.com/gleam-lang/gleam/issues/1951 for now


lpil (approved):
Wow! Thank you

PR #1959, opened 2023-01-14
Upgrade the version of docker/build-push-action to v3

### LINKED ISSUE #1955: Fix warnings on release containers CI workflow
https://github.com/gleam-lang/gleam/actions/runs/3903608863

Ian-GL:
Should fix #1955 . The warnings were not about code on our actual workflows but rather the ones on this action used from docker. The new version seems to not use any more the commands that are causing the warnings.


lpil (approved):
Thank you

PR #1962, opened 2023-01-14
Order functions by use prior to analysis

### LINKED ISSUE #1814: Typer incorrectly infers function as generic
```gleam
pub fn main() {
  call(fn() {
    "Hello"
  })
}

fn call(f: fn() -> a) {
  f()
}
```

Here `main` should be detected as `String` but instead it is detected as an unbound type variable.

Sounds like we've a bug in the generalisation code. May be quite annoying to find the issue with this one.

lpil:
Closes #1814


CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 # Changelog
 
 ## Unreleased
+
+- The unused private type and constructor detection has been improved.
+- Fixed a bug in the type inferrer were unannotated functions used that were
```
michallepicki, line 6:
```suggestion
- Fixed a bug in the type inferrer were unannotated functions that were
```

CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
...
+  used before they were defined in a module could in rare cased be inferred with
+  a more general type than is correct.
 - Fixed a bug where pattern matches on custom types with mixed labelled and 
-  unlabelled arguments could not be compiled with target=javascript
+  unlabelled arguments could not be compiled with targeting JavaScript.
```
michallepicki, line 10:
```suggestion
  unlabelled arguments could not be compiled when targeting JavaScript.
```

Issue #1963, opened 2023-01-15
using constant in constants generates invalid JS


CrowdHailer:
Source
```rs
// demo/foo.gleam
pub const x = 5

// demo/foo_test.gleam
import demo/foo

const a = foo.x
```

Generates the correct foo.mjs file but invalid foo_test.mjs file
```js
import * as $foo from "../demo/foo.mjs";

const a = x;
```

Issue on Gleam 0.25.0 and 0.26.0-rc1

lpil:
Thank you

MarkusPettersson98:
I can work on this, if that's cool with you guys :)

PR #1965, opened 2023-01-18
Fix import of constant var dropping module prefix in JS code generation.

MarkusPettersson98:
As detailed in #1963, importing constants from another module would not be qualified properly in the output `.mjs` file.

Therefore, when emitting code for a `Constant Var` expression in the Javascript backend, we check to see if that constant variable is defined in some module. If so, we prefix the constant variable with the qualified name of the module.

# Note for the reviewer
To test the code change, I created a small sample project as the bug reporter did in #1963. What I would like to do is to adhere to your principle of snapshot testing, but I am a bit unsure where I would place the gleam source files (module with constant + importing file) in this case :)

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -1218,7 +1218,13 @@ pub(crate) fn constant_expression<'a>(
             feature: "Bit string syntax".into(),
             location: *location,
         }),
-        Constant::Var { name, .. } => Ok(name.to_doc()),
+
+        Constant::Var { name, module, .. } => Ok({
+            match module {
+                None => name.to_doc(),
+                Some(module) => docvec!["$", module, ".", name],
+            }
+        }),
     }
 }
 
--- a/compiler-core/src/javascript/tests/modules.rs
+++ b/compiler-core/src/javascript/tests/modules.rs
@@ -70,6 +70,21 @@ const z = y
 }
 
 #[test]
+fn renamed_module() {
+    assert_js!(
+        (
+            CURRENT_PACKAGE,
+            vec!["x".to_string()],
+            r#"pub const v = 1"#
+        ),
+        r#"
+import x as y
+const z = y.v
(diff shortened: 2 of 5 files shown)


lpil (commented):
Hello! Thank you for this 💜 

Some snapshot tests to ensure this doesn't break in future would be fab.

Here's an example test: https://github.com/gleam-lang/gleam/blob/989bbf4ccfb12ed03fac76a0daef3ef1b8c911de/compiler-core/src/javascript/tests/custom_types.rs#L111-L124

You'll need to have [cargo insta](https://crates.io/crates/cargo-insta) installed to work with that test macro.

It would be fab to have tests covering these cases:

- An imported constant
- An imported constant where the module has a/nested/name
- An imported constant where the value was aliased `import one.{two as three}`

And any others you can think of :)

lpil:
Nice!

How about the module being renamed? `import x as y`

lpil:
Are you still working on this @MarkusPettersson98 ? Thanks!

MarkusPettersson98:
> Are you still working on this @MarkusPettersson98 ? Thanks!

Yes! Life got in the way a bit, but will resume work on this issue during the weekend 😊

MarkusPettersson98:
Hello again @lpil :wave:
How are you doing? :smiley: 

Since last time I added the `import x as y` test case you suggested, thanks!
I also rebased on top of the [current main](https://github.com/gleam-lang/gleam/commit/847e40f358065dd5a93c4ea0f59fca68848c2013) and [fixed the regression](https://github.com/gleam-lang/gleam/pull/1965/commits/61db1feaf56a3eab8a7c0216b23664b147032662) that caused the last [CI runs to fail](https://github.com/gleam-lang/gleam/actions/runs/3961247230/jobs/6832940267) :smile:

Take care!

lpil (approved):
Hello! Thank you so much, fab work 💜

Issue #1960, opened 2023-01-14
🎨 Formatter formats long `|` patterns weirdly for constructors.


hayleigh-dot-dev:
I have a fairly long pattern like this...
```gleam
pat.Typeof("Boolean", pat) | pat.Typeof("Number", pat) | pat.Typeof("String", pat) -> {
  ...
}
```
Because this exceeds the 80 char width, I'd expect this to wrap to something like...
```gleam
pat.Typeof("Boolean", pat) | pat.Typeof("Number", pat) | 
pat.Typeof("String", pat) -> {
  ...
}
```
...or...
```gleam
pat.Typeof("Boolean", pat) | 
pat.Typeof("Number", pat) | 
pat.Typeof("String", pat) -> {
  ...
}
```
Instead what ends up happening is it's formatted like this...
```gleam
pat.Typeof("Boolean", pat) | pat.Typeof("Number", pat) | pat.Typeof(
  "String",
  pat,
) -> {
  ...
}
```
I find this really hard/confusing to read.

lpil:
Yeah it basically ignores `|` at the moment, it is real ugly!

PR #1973, opened 2023-01-22
Format long | pattern with sensible breaks

cobbinma:
Fixes the formatter issue in #1960 🚀


lpil (commented):
Sweet! Could we have the `|` as the start of the line? I think that would be clearer.

lpil (approved):
Thank you!

Issue #1974, opened 2023-01-23
Command to build but not publish a hex tarball


keroami:
Write out the tar file, rather than push to hex.pm directly.
Allows for copying to a personal hex repo, which a rebar3 project can use.

(pull request comping up; I have an implementation)

PR #1975, opened 2023-01-23
Allow publishing a package to a local tarball

keroami:
The pull request mentioned #1974


lpil (commented):
Nice!

I think maybe this would be best as `gleam export hex-tarball`?

For the error handling let's return errors to be printed with the usual error handling code rather than pattern matching on it. If you use the function in the `fs` module the Gleam errors will be constructed automatically so you won't need to convert the Rust stdlib ones into them.

Issue #1292, opened 2021-10-09
Support `--warnings-as-errors` for build tool


lpil:

vic:
I'll also take this one.

lpil:
Thank you

pascalwengerter:
@lpil has this happened via https://github.com/gleam-lang/gleam/commit/c202ebcf6b177be4d80adcff3f6380ed651fb3d0 ? :)

lpil:
Possibly! We'll need to check it is working as that was some time  go and I don't remember if it was before or after this issue was created 😅

pascalwengerter:
> Possibly! We'll need to check it is working as that was some time go and I don't remember if it was before or after this issue was created 😅

Turns out it's been "disabled" via https://github.com/gleam-lang/gleam/blob/2af8497e5f24f300d4c5b01b1dbe2bb6fa83d2cc/compiler-cli/src/main.rs#L359, I have a running version locally but need to refine it a bit more. Is it intended that warnings get swallowed in consecutive builds?

lpil:
Oops 😁 

> Is it intended that warnings get swallowed in consecutive builds?

Currently there's a limitation with conditional compilation, there's an issue for it here:

https://github.com/gleam-lang/gleam/issues/1913

pascalwengerter:
@lpil I guess this was closed by mistake? I'll still try to pick up my (now closed) PR sometime soon :)

lpil:
Nop, it was implemented here: https://github.com/gleam-lang/gleam/commit/3d394fda30828c79531317badf003d3b986f107e

Sorry for closing yours, I thought you had lost interest!

pascalwengerter:
> Nop, it was implemented here: [3d394fd](https://github.com/gleam-lang/gleam/commit/3d394fda30828c79531317badf003d3b986f107e)
> 
> Sorry for closing yours, I thought you had lost interest!

Ah, nice to see it having been re-introduced (and the solution looks more straightforward than mine 😅 )! Let me check for another first-timer friendly issue, then

lpil:
The codebase changed a little since you made your attempt so it was easier for me :)

PR #1976, opened 2023-01-25
Reintroduce warnings_as_errors build flag

pascalwengerter:
Potential fix for #1292, looking for feedback for the approach since it doesn't feel optimal to having to pass it into various build cmds throughout the project (mostly with `false`, but some might actually profit from the `true`, e.g. the publish cmd)

### DIFF
--- a/compiler-cli/src/export.rs
+++ b/compiler-cli/src/export.rs
@@ -28,9 +28,10 @@ pub(crate) fn erlang_shipment() -> Result<()> {
 
     // Build project in production mode
     let package = crate::build::main(Options {
-        perform_codegen: true,
         mode,
+        perform_codegen: true,
         target: Some(target),
+        warnings_as_errors: true,
     })?;
 
     for entry in crate::fs::read_dir(&build)?
--- a/compiler-cli/src/export.rs
+++ b/compiler-cli/src/export.rs
@@ -1,5 +1,5 @@
 use gleam_core::{
-    build::{Mode, Options, Target},
+    build::{Mode, Options, Target, WarningLevel},
     paths, Result,
 };
 
@@ -31,7 +31,7 @@ pub(crate) fn erlang_shipment() -> Result<()> {
         mode,
         perform_codegen: true,
         target: Some(target),
-        warnings_as_errors: true,
+        warnings_as_errors: WarningLevel::Error,
     })?;
 
     for entry in crate::fs::read_dir(&build)?
--- a/compiler-cli/src/export.rs
+++ b/compiler-cli/src/export.rs
@@ -31,7 +31,7 @@ pub(crate) fn erlang_shipment() -> Result<()> {
         mode,
         perform_codegen: true,
         target: Some(target),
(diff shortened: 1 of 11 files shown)


lpil (commented):
This looks great to me!

Could you either add a test or a `// TODO: test` to the bit where it returns the error? I forget how easy it is to test the ProjectCompiler so I don't mind if a todo is easier for now :)

Could we replace that bool with something like this please? To prevent me misusing the bool later during a refactoring

```rust
enum WarningLevel {
  Warn
  Error
}
```

pascalwengerter:
> This looks great to me!
> 
> Could you either add a test or a `// TODO: test` to the bit where it returns the error? I forget how easy it is to test the ProjectCompiler so I don't mind if a todo is easier for now :)
> 
> Could we replace that bool with something like this please? To prevent me misusing the bool later during a refactoring
> 
> ```rust
> enum WarningLevel {
>   Warn
>   Error
> }
> ```

Could you take another look? I got it working but not sure if the overall approach is solid/the way you imagined it?

compiler-cli/src/export.rs
```diff
@@ -28,9 +28,10 @@ pub(crate) fn erlang_shipment() -> Result<()> {
 
     // Build project in production mode
     let package = crate::build::main(Options {
-        perform_codegen: true,
         mode,
+        perform_codegen: true,
         target: Some(target),
+        warnings_as_errors: WarningLevel::Error,
```
lpil, line 34: I don't think we should ever default to having this on as it means some people would not be able to deploy their code.

compiler-cli/src/main.rs
```diff
@@ -444,15 +451,17 @@ fn command_check() -> Result<(), Error> {
         perform_codegen: false,
         mode: Mode::Dev,
         target: None,
+        warnings_as_errors: WarningLevel::Error,
```
lpil, line 454: This should be `Warn` or be passed in.

compiler-cli/src/publish.rs
```diff
@@ -37,8 +37,9 @@ impl PublishCommand {
         // Build the project to check that it is valid
         let mut compiled = build::main(Options {
             mode: Mode::Prod,
-            target: Some(Target::Erlang),
             perform_codegen: true,
+            target: Some(Target::Erlang),
+            warnings_as_errors: WarningLevel::Error,
```
lpil, line 42: Likely this one too

lpil (commented):
Looking good!

I think from reading this that it'll also include warnings from dependencies, is that right? I suspect it should only include your code as you can't fix the warnings for deps

pascalwengerter:
@lpil rebased to resolve conflicts that have arisen, could you provide feedback? :)

lpil:
Thank you @pascalwengerter ! I couldn't immediately see- does this permit warnings from dependencies? I feel like we would likely not want to fail the build in the event there are some there

lpil:
I'm going to close this due to inactivity, please feel free to reopen!

PR #1979, opened 2023-01-26
Update many GitHub actions

Ian-GL:
I saw that there are still warnings in the actions workflows, so I took a closer look to see where they were happening and I've updated all the actions where I could find warnings. Hopefully this solves the ones remaining!


lpil (approved):
Thank you

Issue #1992, opened 2023-02-09
gleam 0.26.2 downloads outdated stdlib resulting to compiling error


Sterver47:
Hello,

I accidentally found a bug when I wanted to use the [newest release 0.26.2](https://github.com/gleam-lang/gleam/releases/tag/v0.26.2):
```ps
PS> .\gleam.exe test
  Compiling gleam_stdlib
error: Unlabelled argument after labelled argument
    ┌─ build\packages\gleam_stdlib\src\gleam\list.gleam:576:34
    │
576 │ pub fn prepend(to list: List(a), item: a) -> List(a) {
    │                                  ^^^^^^^

All unlabelled arguments must come before any labelled arguments.
```

The compiler downloads an outdated version of the _gleam_stdlib_, in which the `list.prepend` function's second argument isn't labeled.
I fixed it locally by downloading the [last release of the _gleam_stdlib_](https://github.com/gleam-lang/stdlib/releases/tag/v0.26.1) and replacing it in the _/build/packages_ directory, as the label was added in commit [fc955a79cb2d15052d8e4d99aaba4dbd17671b77](https://github.com/gleam-lang/stdlib/commit/fc955a79cb2d15052d8e4d99aaba4dbd17671b77) in [v0.26.1](https://github.com/gleam-lang/stdlib/releases/tag/v0.26.1).

Sterver47:
Oh, I see. The problem was in my package, not in the compiler itself. Good to know.

Thank you and sorry for the otiose issue and PR.

lpil:
No worries! Thanks for the report, I agree the issue was unclear

PR #1993, opened 2023-02-09
fixed outdated version of gleam_stdlib

Sterver47:
Hopefully, a fix for: https://github.com/gleam-lang/gleam/issues/1992


lpil:
Thanks for the contribution! Gleam uses Hex's version requirement scheme so `~> 0.26` is `>= 0.26.0 and < 1.0.0`, and `~> 0.26.1` is `>= 0.26.1 and < 0.27.0`, which is more constrained than we want for the default.

Issue #1990, opened 2023-02-08
Comment above case branch splits alternative clauses.


brettkolodny:
The gleam formatter will take the following code:

```gleam
case num {
  // 1 or 2
  1 | 2 -> ...
}
```
into
```gleam
case num {
  // 1 or 2
  1
  | 2 -> ...
}
```

lpil:
Thank you. We need to wrap the pattens with a `.group()` that excludes the comment.

PR #1995, opened 2023-02-10
Comment above case branch splits alternative clauses. #1990

Ashish020202:
closes   https://github.com/Ashish020202/gleam/pull/new/ashish/issue1190

commented case 
splited alternative clauses

### DIFF
--- a/compiler-core/src/erlang/tests/patterns.rs
+++ b/compiler-core/src/erlang/tests/patterns.rs
@@ -9,7 +9,9 @@ pub fn test() {
   let duplicate_name = 1
 
   case 1 {
-    1 | 2 -> {
+    // 1 or 2
+    1 
+    | 2 -> {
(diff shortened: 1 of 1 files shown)


lpil:
Hey @Ashish020202 ! Thanks for the contribution. Did you forgot to push your changes perhaps? 😁

Ashish020202:
Pls elaborate more about this issue I think I took some wrong assumption.

Ashish020202:
case num {
  // 1 or 2
  1
  | 2 -> ...
}
I have to change in this format isn't it .
Or there is any other case .

lpil:
The issue is that when a comment is added about an alternative pattern it now forces it to be wrapped onto multiple lines, this should not be the case.

The solution will be to use the `.group()` method on the printing document for the pattern (https://github.com/gleam-lang/gleam/issues/1990#issuecomment-1424704922)

PR #1997, opened 2023-02-12
Export hex tarball

keroami:
Implements #1974
Replaces #1975, incorporated feedback (moved to `export`, using `fs` for errors)
Some other improvements, in separate commits.


compiler-cli/src/export.rs
```diff
@@ -83,3 +83,12 @@ the entrypoint.sh script.
 
     Ok(())
 }
+
+pub fn hex_tarball() -> Result<()> {
+    let config = crate::config::root_config()?;
+    let data: Vec<u8> = crate::publish::build_hex_tarball(&config)?;
+
+    let path = paths::build_tarball(&config.name, &config.version.to_string());
+    println!("Creating package file '{}'", &path.display());
```
lpil, line 92: Can we have this last and in past tense? To match the message for `export erlang-shipment`. Without the quotes around the path too.

TIL path.display, nice!

lpil (commented):
Nice! It would be fab to have a test for this. Could be in the makefile tests in the `test` directory

keroami:
Added a test. checks that correct file is made (will fail when version is changed in gleam.toml). Does not check content.

wording updated.

lpil (commented):
Nice! As it's not a unit test the CI configuration will also need to be updated.

Could you also add a TODO comment to convert it into a unit test? Thank you.

keroami:
What does CI run, then?
`make test` will run the hex tarball test.

lpil:
It runs specific tests each in their own step for debugging. `make` is not used for testing currently, it is intended as a developer convenience.

keroami:
OK, I assumed it is `.github/ci.yaml` that you want.
For me, the CI is the convenience. `make test` or equivalent is leading, CI has to follow.

I do not know how you'd want to test the hex tarball in a unit test. It'd still be an integration test, but ran from a unit testing framework? If so, which command (or CI config?) should run it? then I I can add that to the comment so at least I'd understand it myself ;)

lpil:
We don't have a specifically unit testing framework so nothing to worry about there, just the regular Rust test runner. The test would pass in the in memory file system we use in tests and then assert the tarball is written to an appropriate location and can be extracted.

Sorry for the confusion but the top level makefile is not an authority, we don't use or document it anywhere outside of development and it doesn't do anything you couldn't trivially do with conventional Rust shell commands. I believe most people don't use it, instead using the regular Rust workflow.

keroami:
Like this?

A lot to learn about the Rust eco-system... It's all completely new to me :)

lpil:
Yup! I've made it more confusing here by adding a makefile, which is unusual for Rust. My bad

PR #2002, opened 2023-02-13
Fix compiler panics while inferring type of expression

### LINKED ISSUE #1978: compiler panics while inferring type of expression
Hi,

First of all, thanks for your work on this language !
While toying with it, I've stumbled across a compiler bug:

while inferring the type of this `case` expression,
 the compiler reaches an ["unreachable!()" statement](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_/expression.rs#L1808):

```gleam
import gleam/regex.{Match}

pub fn main() {
  let str = ""
  let _ = case 1 {
    _ if Match(str, []) -> True // this line panics
    _ -> False
  }
}
```

The exact output is:
```
bash $ gleam run
  Compiling bug
error: Fatal compiler bug!

with this information and the code that produces the crash.

        compiler-core/src/type_/expression.rs:1808
        internal error: entered unreachable code
 
```

I'm using a `gleam` binary directly compiled from sources, from the current lastest commit (2af8497e5f24f300d4c5b01b1dbe2bb6fa83d2cc).

Best regards,

evomassiny

maurobalbi:
This PR should fix #1978.

I started off adding the LocalVariable constructor-variant to the pattern match, which fixed the panic when inferring guard-constants. However during testing I found that pattern variables are introduced after the clause guards which might reference them, causing some runtime exceptions in the JS-backend.

For this I added the guard_constant_expression function which gets passed the generated assignments to be able to use the proper variable names in the guard clause. However this is a lot of code duplication and seems a bit hacky, so I'm more than happy to follow up / redo it after some pointers!


lpil (commented):
Brilliant, thank you! Could we also get some snapshot tests for this functionality? Including the issues you discovered in your testing.

https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/javascript/tests/case.rs

You will need the "cargo-insta` program installed to save snapshots.

maurobalbi:
I added a js snapshot (and found and fixed a bug) and added the changelog entry.

lpil (approved):
Wonderful! Thank you

lpil:
Looks like there's some small linting issues, would you mind fixing those? Thank you 🙏

maurobalbi:
Oh okay, it is fixed now!

lpil (approved):
Perfect, thank you!!

Issue #1940, opened 2023-01-02
formatter: floats with only zeros as fp part such as 1.00, 1.000 should be shorted to one fp zero


inoas:
Let me know that/if you want this, then once I have some time I will take a look at this.

inoas:
> juliaailuj: clang-format doesn't seem to trim trailing zeros, Prettier does
https://discord.com/channels/768594524158427167/768594524158427170/1059480877316448386

lpil:
Sounds good

inoas:
#2009 should close this, if it doesn't I will reopen this issue.

PR #2009, opened 2023-02-19
formatter: `trim_right` meaningless zeroes from floats' floating point part

inoas:
Fixes https://github.com/gleam-lang/gleam/issues/1940


inoas:
If the formatting is what you'd expected I will add more tests for floats and floats with scientific notation as well.

compiler-core/src/format.rs
```diff
@@ -773,18 +773,51 @@ impl<'comments> Formatter<'comments> {
...
+            }
+            fp_doc.push(ch);
+        }
+        // If there is no FP then append a 0 (thus that 1. becomes 1.0 etc.)
+        let fp_doc_len = fp_doc.len();
+        if fp_doc_len == 0 {
+            fp_doc.push('0');
+        }
+
+        let fp_doc = fp_doc.chars().rev().collect::<SmolStr>();
```
lpil, line 815: I think we can do this like this:

```rust
    let mut y = x.trim_end_matches('0').to_string();
    if y.ends_with('.') {
        y.push('0')
    };
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b674702762b650073948c92781f965cc

inoas:
This is hopefully good now!

PR #2012, opened 2023-02-20
formatter: case alt patterns break late and not if they are preceded by a comment

inoas:
alt pr to https://github.com/gleam-lang/gleam/pull/1995

as this did not solve the issue yet, I am also happy if 1995 just takes this test and implements the formatter change correctly (I don't want to steal PR opportunities).


inoas:
I have no clue why the 2nd comment gets indendet like that...

lpil:
It looks like you're inserting the comments before the newline above the clause, will need to have a line above the comments also

compiler-core/src/format/tests.rs
```diff
@@ -2985,6 +2985,28 @@ fn expr_case_alternative_patterns() {
...
+        r#"fn main() {
+  case 1 {
+    // Hello Jose!
+    1 | 2 -> Nil
+    // Hello Louis!
+    n -> Nil
+  }
+}
+"#
+    );
```
lpil, line 3008: Could we have this test also? 🙏 

```rust
    assert_format!(
        r#"fn main() {
  case 1 {
    // Hello Jose!
    1 | 2 -> Nil

    // Hello Louis!
    n -> Nil
  }
}
"#
    );
```

inoas: ok

inoas:
This should hopefully now be GTM @lpil

lpil (approved):
Nice! Thank you

Issue #2005, opened 2023-02-16
Duplicate unused variable warnings from Erlang


digitalcora:
Versions:

* Gleam 0.26.2
* OTP 25.2.1

Minimal example:

```gleam
// src/example.gleam

import gleam/io

pub fn main() {
  let unused = True
  io.println("Hello world!")
}
```

Compile output:

```
build/dev/erlang/example/_gleam_artefacts/example.erl:8:5: Warning: variable 'Unused' is unused
%    8|     Unused = true,
%     |     ^


warning: Unused variable
  ┌─ ./src/example.gleam:4:7
  │
4 │   let unused = True
  │       ^^^^^^ This variable is never used.

Hint: You can ignore it with an underscore: `_unused`.
```

Both Erlang and Gleam print separate warnings about the variable being unused. There is otherwise no consequence (I think), but it can make the output quite noisy when there are multiple unused variables.

lpil:
I agree, this is quite annoying! A short term fix would be to see if we can disable unused variable and value warnings only (I wish to keep the others).

Longer term once we have improved our dead code detection we can avoid unused code.

PR #2014, opened 2023-02-21
Erlang unused variable warning

lpil:
Closes https://github.com/gleam-lang/gleam/issues/2005

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -223,7 +223,7 @@ fn module_document<'a>(
     ));
 
     Ok(header
-        .append("-compile(no_auto_import).")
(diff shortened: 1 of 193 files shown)


compiler-core/src/erlang.rs
```diff
@@ -223,7 +223,7 @@ fn module_document<'a>(
     ));
 
     Ok(header
-        .append("-compile(no_auto_import).")
+        .append("-compile([no_auto_import, nowarn_unused_vars]).")
```
inoas, line 226: Might this result in us not detecting bugs?
Could this be hidden behind a flag? Would that make sense?

lpil: Gleam detects these already so the Erlang warning is a duplicate

PR #2017, opened 2023-02-23
Deprecate assert

lpil:


cognivore:
The PR sounded scarier than it is.

I'm convinced that `assert` in BEAM languages should be available to feature developers (not just testers) to implement features via offensive programming!

lpil:
Totally!!!

PR #2026, opened 2023-02-25
fix gleam format warnings about assert

inoas:
warnings about try are staying until the lang feature is removed


inoas:
This should hopefully now be GTM @lpil

lpil (approved):
Fab! Thank you

PR #2029, opened 2023-02-26
JS: replace `throwError` with `throw makeError` in codegen

inoas:
did you mean it like this?


lpil (commented):
Fab! There's also a TypeScript file for the prelude that'll need updating too, as well as the changelog

lpil (approved):
Fab!!

Issue #1983, opened 2023-01-30
Add unary negation operator


jiegillet:
When doing math, it's useful to be able to negate variables
```gleam
fn(a, b) { -a - 3 * b }
```
but `-a` won't parse. A workaround is `0 - a - 3 * b`, but a unary negation operator would be nicer.

lpil:
Thank you!

zphixon:
Hello! This feature seems pretty straightforward in the implementation but I'm beginning to wonder:

Should there be a separate negation operator for Int and Float? That would fit with the rest of the arithmetic operators but IMO there isn't a real conceptual difference between negating the two.

I'd also note that it should be supported in the implementation of #1756. There might also be type checking implications if unsigned integers are ever added.

I've played with this a bit already, I'd like to continue if it's welcome :)

inoas:
I don't think a float/integer denotation is necessary, is it?

`let b = -a` means it must be an integer or a float, but the type of a must be known upfront via inference anyway? `-a .+ b` is cleary so is `-a + b`?

Also if `-a.` or something like that existed then what would be a way to explicitly declare something as a positive integer or float?

lpil:
Great question!

Let's have `-x` ensure that `x` is an int. In future we can have a syntax for floats or overload the `-` operator, but for now we don't have any operator overloading so this should be done in a considered fashion rather than part of this work.

vistuleB:
Coming upon this thread later. I don't understand why would we consider overloading the unary `-` operator for floats when we can fit with existing syntax/philosophy by making `-.` a unary operator as well as a binary operator, just like `-` is a unary as well as a binary operator for integers? Why depart from the existing pattern?

PR #2033, opened 2023-03-01
Unary negation for integers

zphixon:
Howdy! This is part of #1983, only looking at integers for now. E.g.

```gleam
let a = 6
let b = -a
```

now parses, type checks, and codegens, where previously `-a` would be a syntax error.


CHANGELOG.md
```diff
@@ -49,6 +49,7 @@
   runtime if executed.
 - Fixed bug where raised errors on JS would have an extra stack frame recorded
   in them.
+- Added support for unary negation of integers.
```
lpil, line 52: Could you rebase on main and put this in a new `## Unreleased` section please?

compiler-core/src/ast/typed.rs
```diff
@@ -152,7 +152,12 @@ pub enum TypedExpr {
         args: Vec<TypedRecordUpdateArg>,
     },
 
-    Negate {
+    NegateBool {
+        location: SrcSpan,
+        value: Box<Self>,
+    },
+
+    NegateInteger {
```
lpil, line 160:
```suggestion
    NegateInt {
```

To be consistent within the codebase 🙏

compiler-core/src/ast/untyped.rs
```diff
@@ -123,7 +123,12 @@ pub enum UntypedExpr {
         arguments: Vec<UntypedRecordUpdateArg>,
     },
 
-    Negate {
+    NegateBool {
+        location: SrcSpan,
+        value: Box<Self>,
+    },
+
+    NegateInteger {
```
lpil, line 131:
```suggestion
    NegateInt {
```

compiler-core/src/type_/expression.rs
```diff
@@ -370,12 +374,31 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
         unify(bool(), value.type_()).map_err(|e| convert_unify_error(e, value.location()))?;
 
-        Ok(TypedExpr::Negate {
+        Ok(TypedExpr::NegateBool {
+            location,
+            value: Box::new(value),
+        })
+    }
+
+    fn infer_negate_integer(
```
lpil, line 392:
```suggestion
    fn infer_negate_int(
```

compiler-core/src/format.rs
```diff
@@ -692,7 +692,9 @@ impl<'comments> Formatter<'comments> {
 
             UntypedExpr::TupleIndex { tuple, index, .. } => self.tuple_index(tuple, *index),
 
-            UntypedExpr::Negate { value, .. } => self.negate(value),
+            UntypedExpr::NegateInteger { value, .. } => self.negate_with("-", value),
```
lpil, line 695: We had someone the other day confused by the behaviour of the code `--x`, let's make it clear there's no `--` operator in Gleam by adding spaces or blocks

compiler-core/src/parse.rs
```diff
@@ -577,7 +577,27 @@ where
...
+                    None => {
+                        return parse_error(
+                            ParseErrorType::ExpectedExpr,
+                            SrcSpan { start, end: start },
+                        )
+                    }
+                }
+            }
+
+            // Integer negation
```
lpil, line 596:
```suggestion
            // Int negation
```

lpil (commented):
Very nice! Could we get some tests for the Erlang and JavaScript code generation, as well as for the formatter. Thanks!

compiler-core/src/format.rs
inoas: `NegateInteger`

Would `NegateInt` be consistent with the other change requests?
`Integer` currently only exists 5 times in the whole gleam-lang/gleam repo, `Int` 1161 times.

zphixon: So

```gleam
let a = -b
```

should format to

```gleam
let a = - b
```

or just the case of repeated negation,

```gleam
let a = --b
```

to one of

```gleam
let a = - { -b }
let a = - -b
```

inoas: I guess?:

- `let a = -b` is fine and expected.
- `let a = - b` should format to `let a = -b`
- `let a = --b` should format to `let a = b`

zphixon: That's an interesting idea. Bool negation doesn't do that right now, maybe that could be a feature request?

I think blocks look the nicest, e.g. `--a` becomes `- { -a }`. The newest changes do that

lpil: The formatter doesn't change the AST at all so we wouldn't collapse negation operators, even if they are doing something pointless.  Using a block sounds great

compiler-core/src/format.rs
```diff
@@ -1529,13 +1531,24 @@ impl<'comments> Formatter<'comments> {
...
             _ => docvec!["!", self.wrap_expr(expr)],
         }
     }
 
+    fn negate_int<'a>(&mut self, expr: &'a UntypedExpr) -> Document<'a> {
+        match expr {
+            // Always nest repeated negation in a block to avoid confusion with
+            // the pre-decrement operator (which does not exist)
+            UntypedExpr::BinOp { .. } | UntypedExpr::NegateInt { .. } => {
+                docvec!["- { ", self.expr(expr), " }"]
```
lpil, line 1546: This unfortunately won't wrap and indent correctly when the line is too long. Let's extract the block wrapping code from `wrap_expr` into a helper function and use that here.

zphixon: I ended up cribbing from `operator_side`, `wrap_expr` includes some newlines that weren't always necessary.

lpil (approved):
Wonderful work! Thank you for being so thorough with the testing as well!

Issue #2035, opened 2023-03-01
Make semicolon tokens an error.


CreatorSiSo:
As just discussed on the discord semicolons are not really needed anymore.
However they are currently considered whitespace.
This means that one might try to split up expressions with them like this:
```gleam
pub fn with_semi() {
  5 == 2 + 3;
  --5;
}
pub fn without_semi() {
  5 == 2 + 3
  --5
}
```
The resulting Js is in both cases the following:
```js
export function with_without_semi() {
  return 5 === ((2 + 3) - -5);
}
```
Adding an error for each semicolon that is encountered would make it very clear why this happens.

CreatorSiSo:
Pull request: #2036

PR #2036, opened 2023-03-01
Add lexer error for semicolons

CreatorSiSo:
New error message: `Semicolons used to be white space, but are no longer allowed.`
Hint: `You can safely delete them.`

Issue: #2035


lpil (approved):
Fab work! Thank you

PR #2039, opened 2023-03-02
Set end.line to #lines instead of u32::MAX in text_edit_replace

VlkrS:
Always returning u32::MAX as the last line in text_edit_replace relies on the LSP client exhibiting a behavior that is not covered by the LSP specification.

See related discussion in [helix-editor/helix](https://github.com/helix-editor/helix/issues/6022) and [microsoft/vscode-languageserver-node](https://github.com/microsoft/vscode-languageserver-node/issues/1199#issuecomment-1449544173)

The proposed change works for me, but since I'm not really familiar with Rust, might not be a very idiomatic solution.

### DIFF
--- a/compiler-cli/src/lsp.rs
+++ b/compiler-cli/src/lsp.rs
@@ -4,6 +4,7 @@
 
 use std::{
     collections::{HashMap, HashSet},
+    convert::TryFrom,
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you!!!

PR #2042, opened 2023-03-04
--warnings-as--errors

inoas:
My stab at https://github.com/gleam-lang/gleam/issues/1292

If the approach is good enough I'd need pointers how to read the boolean flag from argv and push it through into the compile steps.


lpil:
Hello! You may want to see this PR for some existing work on this: https://github.com/gleam-lang/gleam/pull/1976

inoas:
I will not work on this PR (for now) and give #1976 some more time.

PR #2046, opened 2023-03-05
Check source hashes when loading from cache

### LINKED ISSUE #1914: Check file hashes as well as mtime for detecting staleness
More precise incremental compilation

laplab:
Resolves #1914.

Currently, gleam module loader checks if the modification time of the source file is greater than the one stored in the cache. If it is, the module gets recompiled. This PR changes this behaviour by adding one more check to ensure that both modification time AND source code changed before recompiling the module.

### DIFF
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -72,6 +72,8 @@ bincode = "1.3.3"
 globset = { version = "0.4.9", features = ["serde1"] }
 # Compact and cheap to clone immutable string type
 smol_str = { version = "0.1", features = ["serde"] }
+# Checksums
+sha2 = "0.9.8"
 
 [build-dependencies]
 # Data (de)serialisation
--- a/compiler-core/src/build/module_loader.rs
+++ b/compiler-core/src/build/module_loader.rs
@@ -22,7 +22,7 @@ use crate::{
     Error, Result,
 };
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(try_from = "String", into = "String")]
 pub(crate) struct SourceDigest(Output<Sha256>);
 
@@ -80,18 +80,19 @@ where
         let artefact = name.replace("/", "@");
         let source_mtime = self.io.modification_time(&path)?;
 
-        let read_source = |name| Ok(Input::New(self.read_source(path, name, source_mtime)?));
+        let module = self.read_source(path, name.clone(), source_mtime)?;
 
         let meta = match self.read_cache_metadata(&artefact)? {
             Some(meta) => meta,
-            None => return read_source(name),
+            None => return Ok(Input::New(module)),
         };
 
-        // If there's a timestamp and it's newer or the same than the source
-        // file modification time then we read the cached data.
-        if meta.mtime < source_mtime {
+        // If the timestamp of the source is newer than the cache entry and
+        // the hash of the source differs from the one in the cache entry,
+        // then we need to recompile.
+        if meta.mtime < source_mtime && meta.digest != SourceDigest::new(&module.code) {
             tracing::debug!(?name, "cache_stale");
-            return read_source(name);
+            return Ok(Input::New(module));
         }
 
         // The cache currently does not contain enough data to perform codegen,
@@ -99,7 +100,7 @@ where
         // that codegen has already been performed before using a cache.
         if self.codegen.is_required() && !meta.codegen_performed {
             tracing::debug!(?name, "codegen_required_cache_insufficient");
-            return read_source(name);
+            return Ok(Input::New(module));
         }
 
         Ok(Input::Cached(self.cached(name, meta)))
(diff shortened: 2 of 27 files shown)


compiler-core/src/build/module_loader.rs
```diff
@@ -49,26 +80,34 @@ where
...
-
         let meta = match self.read_cache_metadata(&artefact)? {
             Some(meta) => meta,
-            None => return read_source(name),
+            None => return self.read_source(path, name, source_mtime).map(Input::New),
         };
 
-        // If there's a timestamp and it's newer or the same than the source
-        // file modification time then we read the cached data.
+        let mut module = None;
```
laplab, line 88: This looks a bit awkward, but it ensures that we do not read source module multiple times in case hash check fails and we need the source for the codegen branch later.

laplab:
@lpil Could you please approve the CI run?

compiler-core/Cargo.toml
```diff
@@ -72,6 +72,8 @@ bincode = "1.3.3"
 globset = { version = "0.4.9", features = ["serde1"] }
 # Compact and cheap to clone immutable string type
 smol_str = { version = "0.1", features = ["serde"] }
+# Checksums
+sha2 = "0.9.8"
```
lpil, line 76: Let's use xxhash, which is much faster than sha256.

https://crates.io/crates/xxhash-rust

compiler-core/src/build/module_loader.rs
```diff
@@ -20,6 +22,35 @@ use crate::{
     Error, Result,
 };
 
+#[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
+#[serde(try_from = "String", into = "String")]
+pub(crate) struct SourceDigest(Output<Sha256>);
```
lpil, line 27: Can we use the word "fingerprint" rather than "digest"? We may use digest for other things :)

compiler-core/src/build/module_loader.rs
```diff
@@ -20,6 +22,35 @@ use crate::{
...
+        hasher.update(source.as_bytes());
+        SourceDigest(hasher.finalize())
+    }
+}
+
+impl TryFrom<String> for SourceDigest {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        let raw_digest = base16::decode(value.as_bytes()).map_err(|e| e.to_string())?;
```
lpil, line 41: This is stored in a binary file so no need to convert to a string like this, we can just store the raw bytes of the fingerprint. Avoid some extra work.

compiler-core/src/build/module_loader.rs
```diff
@@ -49,26 +80,34 @@ where
...
         // The cache currently does not contain enough data to perform codegen,
         // so if codegen is required in this compiler run then we must check
         // that codegen has already been performed before using a cache.
         if self.codegen.is_required() && !meta.codegen_performed {
             tracing::debug!(?name, "codegen_required_cache_insufficient");
-            return read_source(name);
+            return match module {
+                Some(m) => Ok(Input::New(m)),
+                None => self.read_source(path, name, source_mtime).map(Input::New),
+            };
```
lpil, line 110: This method has become a lot harder to read, going from a sequence to checks to a function with nesting and the src value shared between the branches.

If it can be flat again that would be my preference, if not let's commit to nested conditional logic and move this into the branch above, so that the logic is written in a consistent style and we can remove the `module` variable.

lpil (commented):
Very nice!! left a few comments inline

compiler-core/src/build/module_loader.rs
laplab: This is no longer needed, since xxhash outputs integers, which are serializable.

compiler-core/src/build/module_loader.rs
laplab: I was able to get rid of this variable by reordering if statements. Now we check for codegen case first and then we check for modification time. The code is flat now and I think that semantics is fine, but let me know if this is not the case.

compiler-core/src/error.rs
```diff
@@ -2258,16 +2258,16 @@ package to Hex.\n"
                         .to_string();
                 text.push_str(if *description_missing && *licence_missing {
                     r#"Add the licences and description fields to your gleam.toml file.
-                
+
```
laplab, line 2261: These changes are due to my IDE settings, I can revert them if needed.

lpil: nah this is an improvement 👍

laplab:
The CI seems to be failing because xxhash-rust does not use a license type from the allowlist. It uses [BSL](https://github.com/DoumanAsh/xxhash-rust/blob/master/LICENSE), which is still a permissive license without warranty, so it might be okay to just add it to the list.

lpil:
Oh yes! If you could add that that would be fab 🙏

laplab:
@lpil Tested locally - should be fine now 👍

inoas:
is BSL-1.0 this https://choosealicense.com/licenses/bsl-1.0/
or this: https://mariadb.com/bsl-faq-mariadb/

laplab:
@lpil Would you be okay with merging it now? It seems like a good time before any more merge conflicts arise.

lpil:
@laplab We don't use merge commits in this project as we want a linear history, so I can only merge if I squash the commits. Is that OK with you?

laplab:
@lpil Sure, as long as it still mentions me as the author, I am fine with it.

lpil (approved):
Thank you!!!

PR #2059, opened 2023-03-10
formatter removes block around pipline

despairblue:
The formatter removes the block around the pipeline, but it should not.

I wasn't sure if this test should go into the `use` test file or the general test file.
Also not sure how to name the test. I tried to make it similar to the single expression block test.


lpil (approved):
Thank you!

Should go in the general test file as it's unrelated to use. I'll move it :)

PR #2064, opened 2023-03-12
attempt to disable nightly builder for forked repos by default

inoas:
Do not build nightlies out of the box for people who fork gleam.

- Those nightlies are probably almost never used
- The innocent forking creates a lot of energy overhead in the mid long run if lots of people fork

### DIFF
--- a/.github/workflows/release-nightly.yaml
+++ b/.github/workflows/release-nightly.yaml
@@ -14,6 +14,7 @@ jobs:
     name: Check latest commit
     outputs:
       should_run: ${{ steps.should_run.outputs.should_run }}
+    if: github.repository_owner == "lpil"
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-nightly.yaml
```diff
@@ -14,6 +14,7 @@ jobs:
     name: Check latest commit
     outputs:
       should_run: ${{ steps.should_run.outputs.should_run }}
+    if: github.repository_owner == "lpil"
```
lpil, line 17: I think this should be `gleam-lang`?

lpil (approved):
Thank you

lpil:
I'm getting syntax errors :( I thought you said you tested it?

PR #2068, opened 2023-03-14
Support running modules with `gleam run -m`

### LINKED ISSUE #1558: Enable running on commands from dependency packages
It would be useful to be able to run code defined in packages that our Gleam projects depend upon.

For example, a database library may define a way to perform migrations of the database scheme in some controlled way, such as how Ecto does for Elixir. The Elixir programmer would run `mix ecto.migrate` to update their database.

Currently to do this in Gleam you would call the library code from the project main function, but this has the unfortunate drawback of making it so the application code in production has to depend on any package used, even if they are only used for local development. We wish to remove this restriction.

There are 2 parts to this system.
1. Command registration
2. Command execution

# Command registration

This is how we determine that some of the dependency code is a command that can be run by the programmer.

## No registration

Nothing at all! The user instead always provides us with which module to run each time and any module with a `main` function is runnable.

## Package main functions

Any main function in a module with the same name as the package is detected as a command. It would have the same name as the package.

```gleam
// In the wibble package
// In the src/wibble.gleam module
pub fn main() {
  // ...
}
```

This matches the existing API of `gleam run` very closely and has no boilerplate.

## Project level configuration

The programmer adds configuration to their top level `gleam.toml` to define each command, similar to Elixir Mix's aliases feature.

```toml
# In the top level gleam.toml
name = "thingy"
version = "1.0.0"

[commands]
migrate = "db_library/migrate"

[dependencies]
db_library = "~> 1.0"
```

This would allow the user to define a command for any module.

A disadvantage is that programmers would need to include the same command boilerplate for each project that uses

## Package level configuration

The author of a dependency defines commands in their `gleam.toml` and any project that depends on their package would automatically have that command registered.

```toml
# In the dependency's gleam.toml
name = "db_library/migrate"
version = "1.0.0"

[commands]
migrate = "db_library/migrate" # this module would be defined in a dependency package
```

This is more convenient for the programmers depending on packages with commands.

We would need to ensure that multiple packages do not define the same command. Non unique command naming would result in build failures.
This suggests that all commands should be the name of the dependency package and then arguments are used for the subcommands. This does to some degree make the ability to define and name multiple commands seem rather redundant.

## Implicit path based registration

As suggest in https://github.com/gleam-lang/gleam/pull/1557 we could have a convention where certain module paths are exposed as commands, similar to how Elixir's Mix tasks work.

e.g. `src/task/wibble.gleam` is a command called `wibble`.

# Command execution

This is the way that the programmer invokes the command code

## Flags for `gleam run`

Curerntly `gleam run` runs the main function in the module with the same name as the current package. This could be extended to permit the running of a specified module (`gleam run -m some/module`). This is similar to Python's `python -m module` functionality.

We could also base it around package names rather than module names. `gleam run -p some_package`.

One advantage of this is that no registration is required. Any module could be run and we don't need to detect them in advance, the user tells us which one to use and we only check it has a `main` function. We could still use it with a registration system if desired.

This would also be the simplest and also most flexible API to add to the build tool.

## A new command

For example `gleam invoke commandname`

This could get confusing with `gleam run`.

## New top level commands for each dep command

If there was a command named `wibble` then `gleam wibble` would run it.

This API is nice but we would need to think carefully about how help messages work.
We would still want to show a help message if `gleam something` where `something` isn't a command.
It would be non-trivial to include command help information in our help output as it's generated by the `clap` library and it entirely static at present.

Lunarmagpie:
This PR supports running modules with `gleam run -m module_name`. This closes #1558.

Im opening this as a draft since I have a few questions and comments on the implementation.

## What is supported
- Run a gleam module with `gleam run -m module_name`.
- Submodules can also be run with `gleam run -m module/sub_module`.
- Modules need a `pub fn main` function to be run, so its exactly like `gleam run` but for a different module.
- Error handling for `gleam run` when there is no `pub fn main` function was improved in this PR.


lpil:
This issue is just for specifying the module, the function cannot be specified.

compiler-cli/src/main.rs
```diff
@@ -168,6 +168,10 @@ enum Command {
         #[clap(long, ignore_case = true)]
         runtime: Option<Runtime>,
 
+        /// The module to run
+        #[clap(short, long, ignore_case = true)]
```
lpil, line 172: Let's not ignore case, `-m MYAPP/THINGY` should probably not be accepted

compiler-cli/src/run.rs
```diff
@@ -21,24 +23,71 @@ pub fn command(
     arguments: Vec<String>,
     target: Option<Target>,
     runtime: Option<Runtime>,
+    module: Option<String>,
     which: Which,
 ) -> Result<(), Error> {
+    // Validate the module to make sure it is a gleam module path
+    match &module {
+        Some(module_name) => {
+            // TODO: Check if this can be replaced with a function that
```
lpil, line 32: This regex is roughly what you want, though it also checks that there's `.gleam` at the end.

https://github.com/gleam-lang/gleam/blob/67c2d94a2afbfe14ba9075c7761cb6fd1a9c7eaf/compiler-cli/src/fs.rs#L301-L308

compiler-cli/src/run.rs
```diff
@@ -21,24 +23,71 @@ pub fn command(
...
+    {
+        Some(module_data) => match module_data.values.get("main") {
+            Some(ValueConstructor {
+                public: _,
+                variant:
+                    ValueConstructorVariant::ModuleFn {
+                        name: _,
+                        field_map: _,
+                        module: _,
+                        arity: _,
```
lpil, line 77: We'll need to also check that it doesn't take any arguments as otherwise it'll crash when we fail to supply them

compiler-cli/src/run.rs
```diff
@@ -73,6 +122,9 @@ fn run_erlang(package: &str, module: &str, arguments: Vec<String>) -> Result<i32
         args.push(entry.path().join("ebin").to_string_lossy().into());
     }
 
+    // gleam modules are seperated by `/`. Erlang modules are seperated by `@`.
+    let module = module.replace('/', "@");
```
lpil, line 126: We also need to update the JavaScript running functions. If the module is not in the same package then we'll need to have a different import path to support this.

compiler-cli/src/build.rs
```diff
@@ -9,10 +9,10 @@ use crate::{
     build_lock::BuildLock,
     cli,
     dependencies::UseManifest,
-    fs::{self, ConsoleWarningEmitter},
+    fs::{self, ConsoleWarningEmitter, ProjectIO},
 };
 
-pub fn main(options: Options) -> Result<Package> {
+pub fn main(options: Options) -> Result<(ProjectCompiler<ProjectIO>, Package)> {
```
lpil, line 15: Let's not expose the entire ProjectCompiler, it's a very large class and we don't need all that capability here.

Perhaps we can return something like this:

```rust
pub struct Built {
    pub root_package: Package,
    module_interfaces: im::HashMap<SmolStr, type_::Module> 
}
```

Then we can have a function for getting the main function for a given module

```rust
pub fn get_main_function(&self, module: &str) -> Result<StructWithTheInfo, Error> {
    todo!()
}
```

Generally we want to have as little in the IO providing `gleam_cli` as possible, with all the logic living in `gleam_core`, so this could live in `gleam_core`.

Unit tests for it would be fab also

lpil (commented):
Fab work! Thank you. This is looking really nice. 💜 

I've left notes inline, do ask any further questions and clarifications

compiler-cli/src/run.rs
Lunarmagpie: I added a check for arity with a new error message.

compiler-cli/src/main.rs
Lunarmagpie: I removed `ignore_case = true`.

compiler-cli/src/build.rs
Lunarmagpie: I did something similar to this. I thought it would be best to make `module_interfaces` public in that struct. I felt adding a `get_main_function` method to it would add some difficulty to reading the code for not much gain with encapsulation since I would need to access that has map outside of the compiler core crate anyway. I added a `get_function` method to `type_::Module` that is used to get the main function instead.

Since I haven't touched unit tests yet im not going to resolve this issue.

compiler-cli/src/run.rs
Lunarmagpie: Does a `is_valid_module_name` function belong in this file? Or should I create a new function/regex in `compiler-cli/run.rs`?

compiler-cli/src/run.rs
Lunarmagpie: This was fixed my modifying the `write_javascript_entrpoint` function to use the module name that the main function was instead of the module from config.

compiler-cli/src/run.rs
Lunarmagpie: Since [this command](https://github.com/gleam-lang/gleam/pull/2068#issuecomment-1467889908) i think is pretty likely to be a feature in the future im going to use regex validation that meant specifically for this one thing.

Lunarmagpie:
>
> Could we also have:
> 
> ```shell
> gleam run -m module:function
> gleam run -m module/sub_module:function
> ```
> 
> (I like the main function default but) that way one can run arbitrary functions which is very beam like and can help with a lot of things IMHO.

After getting the other stuff done, I feel this won't add much complexity to the PR. I can go implement this if its wanted.

Lunarmagpie:
I need help with writing integration tests for this feature.
I want to test commands similar to these on all platforms:
- `cargo run -- run -m module`
- `cargo run -- run -m module/submodule`
- `cargo run -- run -m module/doesnt_exist`

 I looked at things that are already tested but nothing existing fit what I wanted. The best i came up with what doing something similar to the tests in the `test` directory but this doesn't work because `gleam test` and `cargo test` don't test what I want. I would have to add a test script in python or similar which I won't do because im probably overthinking this :sweat_smile:

compiler-cli/src/build.rs
lpil: I would like not to expose `module_interfaces`. It's an internal data structure that is going to change a lot so I want to keep it an internal implementation detail.

When would it need to be accessed?

lpil:
> I looked at things that are already tested but nothing existing fit what I wanted. The best i came up with what doing something similar to the tests in the test directory but this doesn't work because gleam test and cargo test don't test what I want. I would have to add a test script in python or similar which I won't do because im probably overthinking this 😅

If you look at the the GitHub CI workflow you'll see some steps that call makefiles to test extra things. It's an ugly solution, but it's what we're doing for now.

In future it'll be integrated into `cargo test`.

compiler-cli/src/build.rs
```diff
@@ -13,7 +14,12 @@ use crate::{
     fs::{self, ConsoleWarningEmitter},
 };
 
-pub fn main(options: Options) -> Result<Package> {
+pub struct Built {
+    pub root_package: Package,
+    pub module_interfaces: im::HashMap<SmolStr, type_::Module>,
```
lpil, line 19: This is an internal data structure and not to be exposed 🙏

compiler-cli/src/build.rs
```diff
@@ -13,7 +14,12 @@ use crate::{
     fs::{self, ConsoleWarningEmitter},
 };
 
-pub fn main(options: Options) -> Result<Package> {
+pub struct Built {
```
lpil, line 17: Let's move this into core and have ProjectCompiler.compile return it

lpil (commented):
This is looking fantastic, thank you! Left a few notes

compiler-cli/src/build.rs
Lunarmagpie: I was going to access it in the built struct. Now that i realize built should be internal as well what I was worrying about doesn't matter.

lpil: ah right! Thank you 💜

compiler-cli/src/build.rs
Lunarmagpie: This is done. To make `Built` follow lifetime restrictions, I had to make the compile consume the project compiler. Otherwise uses built outside of the function that builds that project would be a use after free. The alternative I found to this was clone but cloning the module info I think would be an extremely expensive clone.

This didn't cause any issues with existing code luckily.

compiler-cli/src/build.rs
Lunarmagpie: Same as my other comment.

Lunarmagpie:
> If you look at the the GitHub CI workflow you'll see some steps that call makefiles to test extra things. It's an ugly solution, but it's what we're doing for now.
> 
> In future it'll be integrated into `cargo test`.

I did my best with this but I needed to check stdout included certain values for me to be confident the right module was actually being run. I think running the wrong module is a very possible bug. I had to write a python script for that but i hate it. Please tell me if theres a better way to do this, and also apologies for doing so many convoluted things in this PR lol.

test/running_modules/run_tests
```diff
@@ -0,0 +1,46 @@
...
+            print("Failed when running command", args)
+            print("with stdout:", output.stdout.decode("utf-8"))
+            return False
+
+        print("Passed")
+    return True
+
+
+if __name__ == "__main__":
+    main()
```
lpil, line 46: This is cool but I would prefer to not introduce Python. Could we use shell?

```sh
should_succeed() {
  echo Running: "$@"
  eval "$@"
  if [ $? -ne 0 ]
  then
    echo ERROR: Command should have succeeded
    exit 1
  fi
}
 should_fail() {
  echo Running: "$@"
  eval "$@"
  if [ $? -eq 0 ]
  then
    echo ERROR: Command should have failed
    exit 1
  fi
 }

# No module
should_succeed gleam run --target erlang
should_succeed gleam run --target javascript --runtime nodejs
should_succeed gleam run --target javascript --runtime deno

# Module from same package
should_succeed gleam run --module one --target erlang
should_succeed gleam run --module one --target javascript --runtime nodejs
should_succeed gleam run --module one --target javascript --runtime deno

# Nested module from same package
should_succeed gleam run --module one/two --target erlang
should_succeed gleam run --module one/two --target javascript --runtime nodejs
should_succeed gleam run --module one/two --target javascript --runtime deno

# Nested module from another package
should_succeed gleam run --module dep/mod --target erlang
should_succeed gleam run --module dep/mod --target javascript --runtime nodejs
should_succeed gleam run --module dep/mod --target javascript --runtime deno

# Unknown module
should_fail gleam run --module wibble

# No main function
should_fail gleam run --module wobble

# Main function with wrong arity
should_fail gleam run --module wabble
```

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,37 @@ pub struct Options {
...
+
+pub enum ModuleFunctionResult<'a> {
+    Found(ModuleFunction<'a>),
+    ModuleNotFound,
+    FunctionNotFound,
+}
+
+impl<IO> Built<IO>
+where
+    IO: CommandExecutor + FileSystemWriter + FileSystemReader + Clone,
```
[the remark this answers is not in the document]
Lunarmagpie: done

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,37 @@ pub struct Options {
     pub warnings_as_errors: bool,
 }
 
+#[derive(Debug)]
+pub struct Built<IO> {
+    pub root_package: Package,
+    compiler: ProjectCompiler<IO>,
+}
```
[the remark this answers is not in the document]
Lunarmagpie: done

compiler-core/src/type_.rs
```diff
@@ -409,6 +409,12 @@ impl ModuleValueConstructor {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ModuleFunction<'a> {
+    pub arity: usize,
+    pub package: &'a SmolStr,
+}
```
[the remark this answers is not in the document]
Lunarmagpie: done

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,37 @@ pub struct Options {
...
+
+impl<IO> Built<IO>
+where
+    IO: CommandExecutor + FileSystemWriter + FileSystemReader + Clone,
+{
+    pub fn get_module_function(
+        &self,
+        module: &SmolStr,
+        function: &SmolStr,
+    ) -> ModuleFunctionResult {
```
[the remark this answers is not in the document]
Lunarmagpie: done

test/running_modules/run_tests
Lunarmagpie: done. I didn't test a dependency since I don't have a dependency to test. Is there one i should use?

compiler-core/src/error.rs
```diff
@@ -415,6 +427,44 @@ This prefix is intended for official Gleam packages only.",
...
+                location: None,
+                hint: Some(format!("Add a function with the singature `pub fn main() {{}}` to `src/{module}.gleam`").into()),
+            },
+
+            Error::MainFunctionHasWrongArity { module, arity } => Diagnostic {
+                title: format!("Main function has wrong arity"),
+                text: format!("`{module}:main` should have an arity of 0 to be run but its arity is {arity}.").into(),
+                level: Level::Error,
+                location: None,
+                hint: Some(format!("Change the function signature of main to `pub fn main() {{}}`").into()),
```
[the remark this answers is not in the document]
Lunarmagpie: fixed

compiler-core/src/error.rs
```diff
@@ -415,6 +427,44 @@ This prefix is intended for official Gleam packages only.",
                 }
             }
 
+            Error::InvalidModuleName { module } => Diagnostic {
+                title: format!("Invalid module name"),
+                text: format!(
+                    "`{module}` is not a valid module name.
+Module names can only contain lowercase letters, underscore, and forward slash and must not end with a slash."
```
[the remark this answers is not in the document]
Lunarmagpie: fixed by wrapping the second line.

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,37 @@ pub struct Options {
...
+    Found(ModuleFunction<'a>),
+    ModuleNotFound,
+    FunctionNotFound,
+}
+
+impl<IO> Built<IO>
+where
+    IO: CommandExecutor + FileSystemWriter + FileSystemReader + Clone,
+{
+    pub fn get_module_function(
```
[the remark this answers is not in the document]
Lunarmagpie: fixed. This function and the module struct `get_main_function` only can get main.

compiler-core/src/error.rs
```diff
@@ -127,6 +127,18 @@ pub enum Error {
...
+    InvalidModuleName { module: String },
+
+    #[error("{module} is not module")]
+    ModuleDoesNotExist { module: String },
+
+    #[error("{module} does not have a main function")]
+    ModuleDoesNotHaveMainFunction { module: String },
+
+    #[error("{module} has the wrong arity so it can not be run.")]
+    MainFunctionHasWrongArity { module: String, arity: usize },
```
lpil, line 140:
```suggestion
    InvalidModuleName { module: SmolStr },

    #[error("{module} is not module")]
    ModuleDoesNotExist { module: SmolStr },

    #[error("{module} does not have a main function")]
    ModuleDoesNotHaveMainFunction { module: SmolStr },

    #[error("{module} has the wrong arity so it can not be run.")]
    MainFunctionHasWrongArity { module: SmolStr, arity: usize },
```

We can keep it as a SmolStr to remove the code and extra work converting to the mutable string type.

test/running_modules/Makefile
```diff
@@ -0,0 +1,12 @@
+.phony: build
+build: clean erlang javascript
+
+.phony: clean
+clean:
+	@rm -rf target-*
+
```
lpil, line 7:
```suggestion
```

test/running_modules/run_tests
```diff
@@ -0,0 +1,48 @@
...
+
+# Module from same package
+should_succeed cargo run -- run --module module --target erlang
+should_succeed cargo run -- run --module module --target javascript --runtime nodejs
+should_succeed cargo run -- run --module module --target javascript --runtime deno
+
+# Nested module from same package
+should_succeed cargo run -- run --module module/sub_module --target erlang
+should_succeed cargo run -- run --module module/sub_module --target javascript --runtime nodejs
+should_succeed cargo run -- run --module module/sub_module --target javascript --runtime deno
```
lpil, line 39: Could we get a test for a module from a different package? To ensure the JavaScript import logic is working correctly. Could add `gleeunit` as a dep, that has a main function.

lpil (commented):
Fab! Looking really great. I've left some notes but I think that's the last of what's required.

Could you update the changelog file and add a step to the GitHub actions test workflow to run the makefile on CI? Thank you

test/running_modules/run_tests
lpil: Anything would work. How about gleeunit? That has a main function.

Lunarmagpie:
> Could you update the changelog file and add a step to the GitHub actions test workflow to run the makefile on CI? Thank you

Done :smile:

test/running_modules/run_tests
Lunarmagpie: done

compiler-core/src/error.rs
Lunarmagpie: Done. `InvalidModuleName` is still a string though because we never converted to `SmolStr` for that one.

compiler-core/src/error.rs
```diff
@@ -415,6 +427,45 @@ This prefix is intended for official Gleam packages only.",
...
+            },
+
+            Error::ModuleDoesNotExist { module } => Diagnostic {
+                title: "Module does not exist".into(),
+                text: format!("Module `{module}` was not found"),
+                level: Level::Error,
+                location: None,
+                hint: Some(
+                    format!("Try creating the file `src/{}.gleam`.",
+                    Path::new(module).to_str().expect("pretty error print Path to_str"))
```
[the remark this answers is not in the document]
Lunarmagpie: done

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,28 @@ pub struct Options {
...
+pub struct Built {
+    pub root_package: Package,
+    module_interfaces: im::HashMap<SmolStr, type_::Module>,
+}
+
+impl Built {
+    pub fn get_main_function(&self, module: &SmolStr) -> Result<ModuleFunction, Error> {
+        match self.module_interfaces.get(module) {
+            Some(module_data) => match module_data.get_main_function() {
+                Some(module_function) => Ok(module_function),
```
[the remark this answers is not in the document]
Lunarmagpie: Done, moved the arity check.

test/running_modules/run_tests
Lunarmagpie: I used gleeunit.

lpil (approved):
Thank you so much!! I'm really excited to use this one 💜

lpil:
We enforce linear history in this repo so I'll rebase this PR in to remove the merge commit in this branch. Thanks!

Issue #2069, opened 2023-03-14
errors reporting: naming a constant like a a function results


inoas:
```gleam
import gleam/io

pub const main = "foo"

pub fn main() {
  io.println("Hello from temp!")
}
```

```
error: Duplicate constant definition
  ┌─ ./src/temp.gleam:3:11
  │
3 │ pub const main = "foo"
  │           ^^^^ Redefined here
4 │ 
5 │ pub fn main() {
  │ ^^^^^^^^^^^^^ First defined here

`main` has been defined multiple times.
Names in a Gleam module must be unique so one will need to be renamed.
```

But the constant is not duplicate, instead it collides with a function name.
... and the error appears on the const line instead of further down on the function line.

for functions it is:

```gleam
import gleam/io

pub fn main() {
  io.println("Hello from temp!")
}

pub fn main() {
  io.println("Hello from temp!")
}
```

```
error: Duplicate definition
  ┌─ ./src/temp.gleam:3:1
  │
3 │ pub fn main() {
  │ ^^^^^^^^^^^^^ First defined here
  ·
7 │ pub fn main() {
  │ ^^^^^^^^^^^^^ Redefined here

`main` has been defined multiple times.
Names in a Gleam module must be unique so one will need to be renamed.
```

So the error seems recycled?

lpil:
Let's use the duplicate definition error for both, it is more correct.

inoas:
> Let's use the duplicate definition error for both, it is more correct.

1. So there is no plan to allow both next to each other?
2. Remove `TypeError::DuplicateConstName` or just replace the warning string? Latter I will PR now as it is quick

lpil:
To remove DuplicateConstName, yup

PR #2071, opened 2023-03-14
move duplicate const/function into the same error

inoas:
Closing https://github.com/gleam-lang/gleam/issues/2069 (maybe)

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -110,9 +110,6 @@ pub fn infer_module(
     }
 
     // Register values so they can be used in functions earlier in the module.
-    for t in &statements.custom_types {
-        register_values_from_custom_type(t, &mut hydrators, &mut env, &mut value_names, &name)?;
-    }
     for c in &statements.constants {
         assert_unique_name(&mut value_names, &c.name, c.location)?;
     }
@@ -121,6 +118,9 @@ pub fn infer_module(
     }
     for ef in &statements.external_functions {
         register_external_function(ef, &mut value_names, &mut env)?;
+    }
+    for t in &statements.custom_types {
+        register_values_from_custom_type(t, &mut hydrators, &mut env, &mut value_names, &name)?;
     }
 
     // Infer the types of each statement in the module
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -109,18 +109,18 @@ pub fn infer_module(
         register_type_alias(t, &mut type_names, &mut env, &name)?;
     }
 
+    // Register values so they can be used in functions earlier in the module.
+    for t in &statements.custom_types {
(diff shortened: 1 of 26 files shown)


lpil (commented):
Let's just remove the DuplicateConstName entirely and always use DuplicateName

compiler-core/src/error.rs
```diff
@@ -859,7 +859,7 @@ Names in a Gleam module must be unique so one will need to be renamed."
                     };
                     let text = format!(
                         "`{name}` has been defined multiple times.
-Names in a Gleam module must be unique so one will need to be renamed."
+Function and constant names in a Gleam module must be unique so one will need to be renamed."
```
lpil, line 862: Please revert this.

inoas:
```gleam
pub const main = Nil

pub fn main() {
  Nil
}

pub fn main() {
  Nil
}
```

shows:

```
error: Duplicate definition
  ┌─ ./src/temp.gleam:5:1
  │
5 │ pub fn main() {
  │ ^^^^^^^^^^^^^ First defined here
  ·
9 │ pub fn main() {
  │ ^^^^^^^^^^^^^ Redefined here
```

while:

```gleam
pub const main = Nil

pub fn main() {
  Nil
}
```

shows:

```
error: Duplicate constant definition
┌─ ./src/temp.gleam:3:11
│
3 │ pub const main = Nil
│           ^^^^ First defined here
4 │
5 │ pub fn main() {
│ ^^^^^^^^^^^^^ Redefined here
```
... so there is still something wrong

inoas:
Any idea how to munch in functions and consts for duplicate checking in one go, so that it always picks the 1st and 2nd colissions to report an error on?

lpil:
I'm not sure what you mean. It's always in-order in the test snapshots?

lpil:
When printing the error we can put the one with the smaller start index first and the larger one second. That'll do it

inoas:
@lpil I did not expect to solve this on my own knowing barely no Rust, but I did :)

I hope it isn't too bad in terms of patterns/code.

compiler-core/src/analyse.rs
```diff
@@ -109,24 +109,47 @@ pub fn infer_module(
...
+            .push(ef.location)
+    }
+    for (name, locations) in name_locations {
+        let mut sorted_locations = locations.clone();
+        sorted_locations.sort_by_key(|l| l.start);
+        for location in sorted_locations {
+            assert_unique_name(&mut value_names, name, location)?;
+        }
+    }
+
```
lpil, line 139: Rather than having this extra looping code we can order the two spans in `error.rs` just before printing them. Something like this:

```rust
let (first, second) = if a.start < b.start {
    (a, b)
} else {
    (b, a)
};
```

lpil (approved):
Brilliant! Thank you! 💜

PR #2091, opened 2023-03-26
Exhaustiveness

### LINKED ISSUE #34: Full exhaustiveness checking of assignments and case expressions

lpil:
Closes #34


compiler-core/src/warning.rs
```diff
@@ -636,6 +636,41 @@ Run this command to add it to your dependencies:
...
+The missing patterns are:\n"
+                            .into();
+                    for missing in missing {
+                        text.push_str("\n    ");
+                        text.push_str(&missing);
+                    }
+                    text.push_str(
+                        "
+
+In a future version of Gleam this will become a compile error.
```
inoas, line 654: Is this meant to become an error in 1.0?

lpil: Probably before then.

PR #2109, opened 2023-04-06
Add proper module name for prelude

### LINKED ISSUE #2108: "Not exhaustive pattern match" when shadowing prelude types
In the example below I define my own `Result` type with its own `Ok` and `Error` constructors, plus an extra one. Then I try to pattern-match on a function (`list.at`) that returns the original `gleam.Result` type.

```gleam
import gleam
import gleam/list

type Result {
  Ok(Int)
  Halt(Int)
  Error(String)
}

fn test() -> Result {
  case list.at([1, 2, 3], 0) {
    gleam.Ok(val) -> Ok(val)
    gleam.Error(Nil) -> Halt(0)
  }
}
```

The compiler appears to be confused by this and thinks `list.at` now returns my custom `Result` type, and errors because I haven't matched `Halt`.

```
error: Not exhaustive pattern match
   ┌─ ./src/scratch.gleam:11:3
   │
11 │ ╭   case list.at([1, 2, 3], 0) {
12 │ │     gleam.Ok(val) -> Ok(val)
13 │ │     gleam.Error(Nil) -> Halt(0)
14 │ │   }
   │ ╰───^

This case expression does not match all possibilities.
Each constructor must have a pattern that matches it or
else it could crash.

These values are not matched:

  - Halt
```

Of course, if I add a `Halt(_)` pattern as this message suggests, it correctly errors because of a type mismatch. I can only get it to compile by adding a `_ -> panic` case.

graphman65:
Hi, this is my first contribution to the project!

## Summary
This main change in this PR is to start using a proper name for the prelude module instead of the empty string that was used up until now. 

## Why
The existing solution is causing some issue: if you define a type "A" in a module and try to build that module, this type will have an empty module name because it's in the current module being build. Sadly it means that as soon as you create a type that has the same name as one from the prelude, they will both end up with empty module name and many part of the compiler will then be confused because of that and might end up choosing one or the other resulting in some type errors.

## Solution
I added a new PRELUDE_MODULE_NAME public constant is used instead of the empty string and also updated any related code that was performing a random `module.is_empty()` (not explicit or clear) check with a proper `is_prelude_module` call. 

Should resolve #2108

### DIFF
--- a/compiler-core/src/type_/prelude.rs
+++ b/compiler-core/src/type_/prelude.rs
@@ -2,6 +2,7 @@ use crate::{ast::SrcSpan, build::Origin, uid::UniqueIdGenerator};
 
 use super::{Module, Type, TypeConstructor, TypeVar, ValueConstructor, ValueConstructorVariant};
 use std::{cell::RefCell, collections::HashMap, sync::Arc};
+use smol_str::SmolStr;
 
 const BIT_STRING: &str = "BitString";
 const BOOL: &str = "Bool";
@@ -13,13 +14,17 @@ const RESULT: &str = "Result";
 const STRING: &str = "String";
 const UTF_CODEPOINT: &str = "UtfCodepoint";
 
-// TODO: use "gleam" as the prelude module name, not ""
+pub const PRELUDE_MODULE_NAME: &str = "gleam";
+
+pub fn is_prelude_module(module: &SmolStr) -> bool {
+    module == PRELUDE_MODULE_NAME
+}
 
 pub fn int() -> Arc<Type> {
     Arc::new(Type::App {
         public: true,
         name: INT.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
         args: vec![],
     })
 }
@@ -29,7 +34,7 @@ pub fn float() -> Arc<Type> {
         args: vec![],
         public: true,
         name: FLOAT.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
     })
 }
 
@@ -38,7 +43,7 @@ pub fn bool() -> Arc<Type> {
         args: vec![],
         public: true,
         name: BOOL.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
     })
 }
 
@@ -47,7 +52,7 @@ pub fn string() -> Arc<Type> {
         args: vec![],
         public: true,
         name: STRING.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
     })
 }
 
@@ -56,7 +61,7 @@ pub fn nil() -> Arc<Type> {
         args: vec![],
         public: true,
         name: NIL.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
     })
 }
 
@@ -64,7 +69,7 @@ pub fn list(t: Arc<Type>) -> Arc<Type> {
     Arc::new(Type::App {
         public: true,
         name: LIST.into(),
-        module: "".into(),
+        module: PRELUDE_MODULE_NAME.into(),
         args: vec![t],
     })
 }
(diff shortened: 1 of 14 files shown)


compiler-core/src/type_/prelude.rs
```diff
@@ -13,13 +14,17 @@ const RESULT: &str = "Result";
 const STRING: &str = "String";
 const UTF_CODEPOINT: &str = "UtfCodepoint";
 
-// TODO: use "gleam" as the prelude module name, not ""
+pub const PRELUDE_MODULE_NAME: &str = "gleam";
+
+pub fn is_prelude_module(module: &SmolStr) -> bool {
```
lpil, line 19: Let's make this &str so it works with all strings, rather than only references to SmolStrs
```suggestion
pub fn is_prelude_module(module: &str) -> bool {
```

lpil (commented):
Wonderful! Thank you very much for the contribution. A nice and clear fix.

Would you mind adding a test to ensure that the faulty behaviour is fixed? Perhaps in here: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_/tests/imports.rs

graphman65:
> Wonderful! Thank you very much for the contribution. A nice and clear fix.
> 
> Would you mind adding a test to ensure that the faulty behaviour is fixed? Perhaps in here: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_/tests/imports.rs

Sure 😃

compiler-core/src/type_/prelude.rs
graphman65: TIL about deref coercion :D Thought I would need to as `as_str` on all the call sites but nope.

graphman65:
@lpil After adding tests for this I've noticed we have a similar issue on the pretty printer. Basically we never display the qualified name in the pretty print output for prelude types. I have two potential solutions for this in the pretty printer:
- If a type is a prelude type AND there is a custom type with the same name in the module then display the full qualified prelude name `gleam.Result`. Otherwise, display the non qualified one. The issue with that is that it might end up being weird because of the inconsistencies. For instance:

  ```gleam
  type Result(a) {
    Blah(a)
  }
  const a = Ok(Result(0)) //  `gleam.Result(Result(Int))`
  ``` 
  I think that's weird to see some prelude with the module (`gleam.Result`) but some without (`Int`).

- Always display the fully qualified name for the prelude or ANY types that are not part of the current module. That would be more verbose (especially when used in the LSP etc) but would probably be more future proof and explicit. I'm not sure how much a breaking change that would be though. 

Not sure if the output of the pretty print is supposed to be backward compatible. But it's clearly broken right now for instance, this code prints the same type name for both vars even though they are completely different. That could lead to massive confusion. 
```gleam
type Result(a) {
  Foo(a)
}

const a = Ok(1) // Result(Int, a)
const b = Foo(1) // Result(Int)
``` 

I've the code for the option 2 (the easiest one to implement) working but would like to get your thought on this.
Option one would require a bigger refactor. As the printer would now need a lot more context (especially for the diagnostic printing, you'd need to pass the whole module to most errors to check if the name should be qualified or not).

lpil:
Why is it that the qualifier would not get printed? The pretty printer does print qualifiers for any two types with the same name that are both printed.

graphman65:
> Why is it that the qualifier would not get printed? The pretty printer does print qualifiers for any two types with the same name that are both printed.

Figured out the printer is meant to be used to print a "single type" and not a full module.
In my example above:
```gleam
type Result(a) {
  Foo(a)
}

const a = Ok(1) // Result(Int, a)
const b = Foo(1) // Result(Int)
```

I was using the `infer_module` method from the test file. This infer_module is creating a new printer for each value it's printing so that's how I ended up with the same "non qualified" name printed for both.

It makes more sense to me now that I know it's only to print one (for lsp hover for instance) so it's not "broken".

The only thing I'm wondering for the example above. What a user would expect? In that case if they hover `b` in the lsp it will say `Result(Int, a)` (because when you only print that type, there is no clash) and if they over `a` it will say `Result(Int)` so again might be confusing to have both type printed the same when printed individually. Thus my suggestion to either make the printer more context aware or always print qualified name for prelude.

lpil:
Let's open up an issue or discussion and we can work out any improvements to the printing of types there, so this branch doesn't get blocked. Better to get the fix in rather than batch it with some other change 👍

lpil:
I've rebased this into main, thank you!

Issue #1864, opened 2022-11-24
Type annotations in `use` expressions


tynanbe:
Currently fail to compile.
```gleam
456 │     use acc: ValidateAcc, axis <- list.fold( // ..
    │            ^ I was not expecting this.
```

lpil:
This is not part of `use`'s design currently. We could add it straightforwardly, but what might be some use cases for this addition? Are they ever needed?

tynanbe:
Probably not ever needed, but since one can annotate the unsugared version, and e.g. `let` assignments, my thought is that it's unexpected for `use` annotations to be invalid syntax.

PR #2110, opened 2023-04-06
Add support for type annotations in use statements

graphman65:
Right now you cannot use type annotations in use statements.

This PR adds support for such syntax:
```gleam
fn main() {
    use blah: Int <- foo
}
```

Solves #1864

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed a bug where the language server would unset error diagnostics when
   displaying information on hover.
+- Added support for type annotations in use statements.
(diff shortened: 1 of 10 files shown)


lpil (commented):
Wonderful! Thank you!

Could we get some tests for the formatter, and an item in the changelog please? Thank you.

graphman65:
> Wonderful! Thank you!
> 
> Could we get some tests for the formatter, and an item in the changelog please? Thank you.

Done.

lpil (approved):
Thank you!! 💜

Issue #1925, opened 2022-12-29
Order of type aliases shouldn't matter


jiegillet:
```
$ gleam --version
gleam 0.25.0
```

The code
```gleam
type Bar(a) { Bar(a)  }
type Foo(a) = Map(a, Bar(a))
```
compiles, but the flipped version
```gleam
type Foo(a) = Map(a, Bar(a))
type Bar(a) { Bar(a)  }
```
doesn't:
```
19 │ type Foo(a) = Map(a, Bar(a))
   │                      ^^^^^^ Did you mean `Map`?
```

Other functions don't seem to mind the order of definition, so it feels to me like type definitions shouldn't either.

Interestingly enough, the co-recursive (although probably nonsensical)
```gleam
type Bar(a) { Bar(Foo(a))  }
type Foo(a) = Map(a, Bar(a))
```
does compile but
```gleam
type Foo(a) = Map(a, Bar(a))
type Bar(a) { Bar(Foo(a))  }
```
doesn't.

lpil:
Sounds good! Let's sort these prior to type inference.

asib:
Hiya, I'd be interested in tackling this if you're open to a newbie having a go? Some initial thoughts (feel free to ignore if you'd rather I don't work on this!):

The error seems to be due to the fact that the hydrator looks for the `Bar` type constructor in the environment,  but before it has been registered. High level, I can imagine a couple of approaches but I don't have a good idea of which is preferable/less complex.
- The `infer_module` function loops over statements, registering types for each. If it hits one that produces an error, it could try processing the following statement and then retry immediately after having processed that one additional statement. Could maybe be smarter about when the retry happens. This feels like not great performance, though.
- Rather than producing an error, register some sort of forward reference that can later be filled in with the actual type information, when the relevant statement is processed. I'm not sure whether this would work, as later bits of `type_from_ast` seem to depend on the actual type information fetched from the environment. However, perhaps these could be made optional and then similarly updated when the forward reference is updated?

lpil:
That's interesting, it doesn't work for me.

![image](https://user-images.githubusercontent.com/6134406/230352456-84135292-4de8-4e98-a0df-ac02978b1e61.png)

What code did you test with?

PR #2112, opened 2023-04-07
Type alias ordering

graphman65:
Right now type aliases are registered in their definition order. It means they must be defined in the correct order in the source file. 

This PR is changing that so that we compute type alias dependencies, detect type alias "cycles" and then register them in topological order.

Long term it would probably makes sense to do that for all types and not just alias. 

This PR solves #1925.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1,6 +1,7 @@
 #[cfg(test)]
 mod tests;
 
+use crate::dep_tree;
 use crate::{
     ast::{
         self, BitStringSegmentOption, CustomType, DefinitionLocation, ExternalFunction,
@@ -103,9 +104,8 @@ pub fn infer_module(
         register_types_from_custom_type(t, &mut type_names, &mut env, &name, &mut hydrators)?;
     }
     // TODO: Extract a Type alias class to perform this.
-    // TODO: Sort type aliases by dependency order so they don't have to be
-    // ordered in the file
-    for t in &statements.type_aliases {
+    let sorted_aliases = sorted_type_aliases(&statements.type_aliases)?;
+    for t in sorted_aliases {
         register_type_alias(t, &mut type_names, &mut env, &name)?;
     }
 
@@ -1329,4 +1329,72 @@ fn get_compatible_record_fields<A>(
     }
 
     compatible
+}
+
+/// Given a type, return a list of all the types it depends on
+fn get_type_dependencies(typ: &TypeAst) -> Vec<SmolStr> {
+    let mut deps = Vec::with_capacity(1);
+
+    match typ {
+        TypeAst::Var { .. } => (),
+        TypeAst::Hole { .. } => (),
+        TypeAst::Constructor {
+            name,
+            arguments,
+            module,
+            ..
+        } => {
+            deps.push(match module {
+                Some(module) => format!("{}.{}", name, module).into(),
+                None => name.clone(),
+            });
+
+            for arg in arguments {
+                deps.extend(get_type_dependencies(arg))
+            }
+        }
+        TypeAst::Fn {
+            arguments, return_, ..
+        } => {
+            for arg in arguments {
+                deps.extend(get_type_dependencies(arg))
+            }
+            deps.extend(get_type_dependencies(return_))
+        }
+        TypeAst::Tuple { elems, .. } => {
+            for elem in elems {
+                deps.extend(get_type_dependencies(elem))
+            }
+        }
+    }
+
+    deps
+}
+
+fn sorted_type_aliases(aliases: &Vec<TypeAlias<()>>) -> Result<Vec<&TypeAlias<()>>, Error> {
+    let mut deps: Vec<(SmolStr, Vec<SmolStr>)> = Vec::with_capacity(aliases.len());
+
+    for alias in aliases {
+        deps.push((alias.alias.clone(), get_type_dependencies(&alias.type_ast)))
+    }
+
+    let sorted_deps = dep_tree::toposort_deps(deps).map_err(|err| {
+        let dep_tree::Error::Cycle(cycle) = err;
+
+        let last = cycle.last().expect("Cycle should not be empty");
+        let alias = aliases
+            .iter()
+            .find(|alias| alias.alias == *last)
+            .expect("Could not find alias for cycle");
+
+        Error::RecursiveTypeAlias {
+            cycle,
+            location: alias.location,
+        }
+    })?;
+
+    Ok(aliases
+        .iter()
+        .sorted_by_key(|alias| sorted_deps.iter().position(|x| x == &alias.alias))
+        .collect())
 }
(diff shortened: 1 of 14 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
...
+        }
+    }
+
+    deps
+}
+
+/// Given a list of type aliases, return a list of the same type aliases sorted
+/// such that any type alias that depends on another type alias appears after
+/// the type alias it depends on.
+/// NOTE: Right now it's only sorting aliases but could actually be used to sort all the types later
```
lpil, line 1383: We can remove this note, there's no reason to sort other types.

compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
 
     compatible
 }
+
+/// Given a type, return a list of all the types it depends on
+fn get_type_dependencies(type_name: SmolStr, typ: &TypeAst) -> Vec<SmolStr> {
```
lpil, line 1348: Best take as a reference here as it is not always consumed
```suggestion
fn get_type_dependencies(type_name: &SmolStr, typ: &TypeAst) -> Vec<SmolStr> {
```

compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
...
+    let mut deps = Vec::with_capacity(1);
+
+    match typ {
+        TypeAst::Var { name, .. } => deps.push(name.clone()),
+        TypeAst::Hole { name, .. } => deps.push(name.clone()),
```
lpil, line 1353: These are not dependencies, we don't need to collect them.
```suggestion
        TypeAst::Var { name, .. } =>(),
        TypeAst::Hole { name, .. } => (),
```

compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
...
+        ))
+    }
+
+    let sorted_deps = dep_tree::toposort_deps(deps).map_err(|err| {
+        let dep_tree::Error::Cycle(cycle) = err;
+
+        let mut cycle = cycle.clone();
+        cycle.reverse();
+
+        let first = cycle.first().unwrap();
```
lpil, line 1400: We can avoid cloning and reversing the vector by taking a reference to the last element.

compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
...
+        let alias = aliases.iter().find(|alias| alias.alias == *first).unwrap();
```
lpil, line 1401:
```suggestion
        let alias = aliases.iter().find(|alias| alias.alias == *first).expect("Could not find alias from cycle");
```

compiler-core/src/analyse.rs
```diff
@@ -1343,3 +1343,72 @@ fn get_compatible_record_fields<A>(
...
+            cycle,
+            location: alias.location,
+        }
+    })?;
+
+    Ok(aliases
+        .iter()
+        .sorted_by_key(|alias| sorted_deps.iter().position(|x| x == &alias.alias))
+        .cloned()
+        .collect())
```
lpil, line 1413: If this function takes ownership of the type aliases vector rather than taking a reference we don't have to clone here.

compiler-core/src/type_/error.rs
```diff
@@ -254,6 +254,12 @@ pub enum Error {
     UnlabelledAfterlabelled {
         location: SrcSpan,
     },
+
+    /// A type cycle was detected
+    TypeCycle {
```
lpil, line 259:
```suggestion
    /// A type alias was defined directly or indirectly in terms of itself, which would
    /// cause it to expand to infinite size.
    /// e.g.
    ///     type ForkBomb = #(ForkBomb, ForkBomb)
    RecursiveTypeAlias {
```

compiler-core/src/type_/tests/type_alias.rs
```diff
@@ -0,0 +1,51 @@
...
+fn alias_cycle() {
+    assert_module_error!(
+        r#"
+type A = B
+type B = C
+type C = D
+type D = E
+type E = A
+"#);
+}
```
lpil, line 51: Can we have a test for an alias that directly recurses? Thanks

compiler-core/src/error.rs
```diff
@@ -2041,6 +2041,27 @@ value, or use `assert` rather than `let`."
                         extra_labels: vec![],
                     }),
                 },
+
+                TypeError::TypeCycle { location, cycle } =>  {
+                    let mut text = "The definition of that type forms a cycle:\n".into();
```
lpil, line 2046:
```suggestion
                    let mut text = "This type alias is defined in terms of itself.\n".into();
```

compiler-core/src/error.rs
```diff
@@ -2041,6 +2041,27 @@ value, or use `assert` rather than `let`."
...
+                    write_cycle(&mut text, cycle);
+                    text.push_str("Gleam doesn't support type cycles like these, please break the cycle to continue.");
```
lpil, line 2048:
```suggestion
                    text.push_str("If we tried to compile this recursive type it would expand
forever in a loop, and we'd never get the final type.");
```

I think it's nice to explain why it doesn't work when possible, so the programmer thinks we're not locking away functionality for no reason.

lpil (commented):
Fab work! Thank you.

I've left some small comments inline, but generally this looks ideal. Would you mind also updating the changelog?

lpil:
Hi @graphman65 , do you want to continue this or shall I? Thanks 🙏

graphman65:
> Hi @graphman65 , do you want to continue this or shall I? Thanks pray

Hey sorry @lpil, last few weeks were pretty busy :/ 
Updating the PR now ;)

compiler-core/src/error.rs
graphman65: Makes sense

graphman65:
PR updated and rebased @lpil

compiler-core/src/analyse.rs
```diff
@@ -1330,3 +1330,64 @@ fn get_compatible_record_fields<A>(
...
+/// Given a type, return a list of all the types it depends on
+fn get_type_dependencies(typ: &TypeAst) -> Vec<SmolStr> {
+    let mut deps = Vec::with_capacity(1);
+
+    match typ {
+        TypeAst::Var { .. } => (),
+        TypeAst::Hole { .. } => (),
+        TypeAst::Constructor {
+            name, arguments, ..
+        } => {
```
lpil, line 1343: We'll also need to take the module into account here. `Box`, `one.Box`, and `two.Box` are all different types.

lpil (commented):
Very nice! I've left a comment inline highlighting an issue. Once we have that fixed and a test written I think we are good to go 💜

compiler-core/src/analyse.rs
graphman65: Good point, just added this and a test for it

lpil (approved):
Thank you!

Issue #2114, opened 2023-04-08
Odd/incorrect formatter behaviour when a comment occurs before a `use`


TanklesXL:
I had the following, and expected it to remain unchanged:
```gleam
// get expected flag value, return any error that occurs
use caps <- result.then(flag.get_bool(from: input.flags, for: "caps"))
  ```
When the formatter ran, it modified it to be:
```gleam
use caps <- // get expected flag value, return any error that occurs
  result.then(flag.get_bool(from: input.flags, for: "caps"))
  ```

**Comment by TanklesXL** [comment]
this is a highlighter test
```gleam
let assert x = 1
```

lpil:
Thank you

PR #2115, opened 2023-04-09
Odd/incorrect formatter behaviour when a comment occurs before a `use`

cobbinma:
Fixes the formatter issue in #2114 🎨


lpil (approved):
Thank you!

PR #2117, opened 2023-04-10
Use dependency target when running dependency module

### LINKED ISSUE #2096: `gleam --module` should prioritize using the language target for a dependency over the project's target
For example, if i do `gleam -m rad` it should use the `javascript` target because that is set in rad's `gleam.toml`. Otherwise, a project like this would crash and that's not intended lol.

Lunarmagpie:
This pr makes it so `gleam run -m mod` will use the dependencies target in the `gleam.toml` when running the module.
Since gleam doesn't support filepath dependencies yet, I published a package on hex.pm that im testing with: https://hex.pm/packages/gleam_module_javascript_test.
This works by looking through all the dependency files to try and find one with the module name you are looking for. Then from there the dependencies folder will be checked for a `gleam.toml`. This implementation means it works with multiple modules under the same root namespace.

I also fixed a bug with the automated testing. It looks like it isn't being called as well so I can investigate that once the automated CI run starts.

closes #2096


compiler-cli/src/module.rs
```diff
@@ -0,0 +1,38 @@
+use lazy_static::lazy_static;
+
+/// Check if a module name is a valid gleam module name.
+pub fn is_gleam_module(module: &str) -> bool {
```
lpil, line 4: How come this function is exported now? I couldn't spot a new use of it. I'd prefer it to be private, if it's public I'd like it moved to core, though not entirely sure where it should go.

test/running_modules/run_tests.sh
```diff
@@ -55,3 +58,6 @@ should_fail run --module module/no_main_function
 
 # Main function with wrong arity
 should_fail run --module module/wrong_arity
+
+# try running gleam_module_javascript_test, which will crash if target is erlang
+should_succeed run --module gleam_module_javascript_test
```
lpil, line 63: Can we have cases for explicitly setting the Erlang and JavaScript values too please?

lpil (commented):
Fab! Thank you. I've left some comments

compiler-cli/src/config.rs
```diff
@@ -12,6 +13,35 @@ pub fn root_config() -> Result<PackageConfig, Error> {
...
 
+/// Get the config for a dependency module. Return the config for the current
+/// project if a dependency doesn't have a config file.
+pub fn module_config(mod_path: &str, project_paths: &ProjectPaths) -> Result<PackageConfig, Error> {
+    let package_path = fs::read_dir(project_paths.build_packages_directory()).map_or(None, |x| {
+        {
+            x.filter_map(|file| match file {
+                Ok(file) => Some(file),
+                Err(_) => None,
+            })
```
lpil, line 24:
```suggestion
            x.filter_map(Result::ok)
```

compiler-cli/src/config.rs
```diff
@@ -12,6 +13,35 @@ pub fn root_config() -> Result<PackageConfig, Error> {
...
+            .find(|file| {
+                file.path()
+                    .join("src")
+                    .join(mod_path.to_string() + ".gleam")
+                    .is_file()
+            })
+        }
+    });
+
+    println!("{:?}", package_path);
```
lpil, line 34: What's this print for? Leftover debugging?

compiler-cli/src/config.rs
```diff
@@ -12,6 +13,35 @@ pub fn root_config() -> Result<PackageConfig, Error> {
     read(paths.root_config())
 }
 
+/// Get the config for a dependency module. Return the config for the current
+/// project if a dependency doesn't have a config file.
+pub fn module_config(mod_path: &str, project_paths: &ProjectPaths) -> Result<PackageConfig, Error> {
```
lpil, line 18: We'll also want to ensure this is a Gleam project and not some other build tool somehow.

lpil, line 18: It wasn't clear to me what this function did until I looked at it for a while, so let's be more verbose with the naming.
```suggestion
pub fn find_package_config_for_module(module_name: &str, project_paths: &ProjectPaths) -> Result<PackageConfig, Error> {
```

compiler-cli/src/config.rs
```diff
@@ -12,6 +13,35 @@ pub fn root_config() -> Result<PackageConfig, Error> {
...
+    let package_path = fs::read_dir(project_paths.build_packages_directory()).map_or(None, |x| {
+        {
+            x.filter_map(|file| match file {
+                Ok(file) => Some(file),
+                Err(_) => None,
+            })
+            .find(|file| {
+                file.path()
+                    .join("src")
+                    .join(mod_path.to_string() + ".gleam")
```
lpil, line 28: This will fail on Windows! We need to handle their paths too.

test/running_modules/run_tests.sh
Lunarmagpie: Added the tests:
```
should_succeed run --module gleam_module_javascript_test --target=javascript
should_fail run --module gleam_module_javascript_test --target=erlang
```

compiler-cli/src/module.rs
Lunarmagpie: reverted this change

compiler-cli/src/config.rs
Lunarmagpie: Does this work? `.join(PathBuf::from(mod_path.to_string() + ".gleam"))` Been reading the PathBuf docs for a while an I can not tell if im doing this right.

compiler-cli/src/config.rs
Lunarmagpie: Can you elaborate on what has to be done?

lpil: I think best way will be to iterate over the packages in the manifest rather than scanning the filesystem, as then we have the information at hand and won't get confused by any other directories.

compiler-cli/src/config.rs
Lunarmagpie: I changed the method to do this.

lpil (commented):
Nice!

compiler-cli/src/config.rs
```diff
@@ -12,6 +14,52 @@ pub fn root_config() -> Result<PackageConfig, Error> {
...
+    project_paths: &ProjectPaths,
+) -> Result<PackageConfig, Error> {
+    let gleam_projects: Vec<&String> = manifest
+        .packages
+        .iter()
+        .filter(|package| package.build_tools.contains(&"gleam".to_string()))
+        .map(|package| &package.name)
+        .collect();
+
+    let package_path = fs::read_dir(project_paths.build_packages_directory()).map_or(None, |x| {
```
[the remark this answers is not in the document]
Lunarmagpie: Oh lol, I can't believe I didn't see that.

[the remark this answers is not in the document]
Lunarmagpie: Fixed by doing what you suggested

compiler-cli/src/config.rs
```diff
@@ -12,6 +14,52 @@ pub fn root_config() -> Result<PackageConfig, Error> {
...
+            x.filter_map(Result::ok)
+                .filter(
+                    |project_folder| match project_folder.file_name().into_string() {
+                        Ok(utf8_str) => gleam_projects.contains(&&utf8_str),
+                        Err(_) => false,
+                    },
+                )
+                .find(|file| {
+                    let mut path = file.path();
+                    path.push("src");
```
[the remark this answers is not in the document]
Lunarmagpie: I *think* I got this working but I can't test because the `test` directory isn't included in gleam package builds. Since this code is only triggered on a dependency I don't think that feature needs to be supported cause of that.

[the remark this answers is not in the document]
lpil: Oh you're completely right, there's no need to check for test inside packages. I'm a fool.

Would you like to remove the test checking bit? I can do that if you're prefer, don't wanna waste your time.

lpil (approved):
Wonderful! Thank you so much!

I've left one question below.

Would you mind also updating the changelog file?

compiler-cli/src/config.rs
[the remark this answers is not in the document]
Lunarmagpie: I'll do it :)

Lunarmagpie:
> Would you mind also updating the changelog file?

Changelog updated!

lpil (approved):
Thank you so much!!

PR #2120, opened 2023-04-12
2050 errors and warnings for double negation

bcpeinhardt:
Going ahead and creating this pull request so you can easily review the diff. The current state of this pull request implements the following:

1. It makes the unary minus (-x) operator compatible with floats as well as ints. I think this is a very practical change worth considering, as having an operator that works on a literal but not on a variable of the same value is pretty weird. 
2. It adds warnings for unnecessary back to back unary negations (--x or !!x)
3. It renames the `NegateInt` and `NegateBool` variants of the `UntypedExpr` enum to `UnaryMinus` and `UnaryBang` respectively, and renames the `infer_negate_int` method to `infer_unary_minus`
4. It adds tests for these changes

Regarding the change to have unary minus work on all floats => I think floats should definitely have some kind of unary minus operator, as it makes writing formulas in their standard order much easier. Personally I would support a unique operator, following the split between arithmetic operators for floats and ints, but a simple -x also seems like a good compromise.

Let me know how you feel about this proposed change, and if you disagree I can simplify the pull request to simply add the warnings for ints and booleans.


bcpeinhardt:
Alright so this has been updated to be just the requested warnings.
--7, --x (where x is an int), !!True, and !!x (where x is a boolean) will emit warnings.
I didn't add - { -x } because it gets into arbitrary levels of depth and if they took the time to type it it may have semantic value.

compiler-core/src/type_/expression.rs
```diff
@@ -166,9 +166,46 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                if let UntypedExpr::NegateBool {
+                    location: _,
+                    value: _,
+                } = &*value
+                {
+                    self.environment
+                        .warnings
+                        .emit(Warning::DoubleUnary { location });
+                }
+                self.infer_negate_bool(location, *value)
```
lpil, line 179: Could you put this logic inside the infer functions please, and have it check after type checking so we don't warn for invalid code.

test/hello_world/build/packages/packages.toml
```diff
@@ -0,0 +1 @@
+[packages]
```
lpil, line 1: Remove this file please 🙏

compiler-core/src/type_/error.rs
```diff
@@ -322,6 +322,10 @@ pub enum Warning {
         location: SrcSpan,
         name: SmolStr,
     },
+
+    DoubleUnary {
```
lpil, line 326: Can we call this something like PointlessDoubleNegation? Make it clear what the error is. Other unary operators could be valid

compiler-core/src/warning.rs
```diff
@@ -382,6 +382,18 @@ expression.",
                         extra_labels: Vec::new(),
                     }),
                 },
+                type_::Warning::DoubleUnary { location } => Diagnostic {
+                    title: "Unnecessary double unary expression".into(),
```
lpil, line 386: This is a bit abstract, which people less familiar with PL jargon will not understand. How about "Pointless double negation" or something like that?

It would also be good to explain to the user why this is a warning. Perhaps something like this:

```
Negating an Int twice returns the original Int, so this expression
does nothing! You can safely remove the two `-`s.
```

compiler-core/src/warning.rs
```diff
@@ -382,6 +382,18 @@ expression.",
...
+                    text: "".into(),
+                    hint: Some("If you're trying to type cast, look for a `from_<type>` function in the target type's module".into()),
+                    level: diagnostic::Level::Warning,
+                    location: Some(Location {
+                        src: src.clone(),
+                        path: path.to_path_buf(),
+                        label: diagnostic::Label { text: Some("Unnecessary double unary expression".into()), span: *location },
```
lpil, line 393: This looks like a duplicate of the title, we can skip it.

lpil (commented):
Nice!! Thank you. I've left some small comments, and if you could update the changelog file too that would be fab.

bcpeinhardt:
Yup! Should be able to get to this in a couple hours :)

lpil (approved):
Thank you!!

lpil:
Would you mind resolving the conflicts please? Also if you could change the word "boolean" to "bool" that would be fab

bcpeinhardt:
Sure :) I think the conflict is the changelog will look at it in a moment

bcpeinhardt:
Changed boolean to bool and the changelog is updated! I've never worked with a changelog before I didn't know if I was doing that right haha.

compiler-core/src/type_/error.rs
```diff
@@ -322,6 +322,14 @@ pub enum Warning {
         location: SrcSpan,
         name: SmolStr,
     },
+
+    UnnecessaryDoubleIntNegation {
+        location: SrcSpan,
+    },
+
+    UnnecessaryDoubleBooleanNegation {
```
inoas, line 330: Bool

compiler-core/src/type_/expression.rs
```diff
@@ -361,6 +361,12 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
 
         unify(bool(), value.type_()).map_err(|e| convert_unify_error(e, value.location()))?;
 
+        if let TypedExpr::NegateBool { .. } = value {
+            self.environment
+                .warnings
+                .emit(Warning::UnnecessaryDoubleBooleanNegation { location });
```
inoas, line 367: Bool

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__double_unary_boolean_literal.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: compiler-core/src/type_/tests/warnings.rs
+expression: "pub fn main() { let _ = !!True }"
+---
+
+warning: Unnecessary double negation (!!) on boolean.
```
inoas, line 6: maybe bool ?

also possibly rename the snap

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__double_unary_boolean_variable.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: compiler-core/src/type_/tests/warnings.rs
+expression: "\n        pub fn main() { \n            let x = True\n            let _ = !!x \n        }\n        "
+---
+
+warning: Unnecessary double negation (!!) on boolean.
```
inoas, line 6: maybe bool

then also the snap file name

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -525,3 +525,41 @@ pub const make_two = one.Two
...
+
+// https://github.com/gleam-lang/gleam/issues/2050
+#[test]
+fn double_unary_boolean_literal() {
+    assert_warning!("pub fn main() { let _ = !!True }");
+}
+
+// https://github.com/gleam-lang/gleam/issues/2050
+#[test]
+fn double_unary_boolean_variable() {
```
inoas, line 556: bool

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -525,3 +525,41 @@ pub const make_two = one.Two
...
+            let x = 7
+            let _ = --x 
+        }
+        "#
+    );
+}
+
+// https://github.com/gleam-lang/gleam/issues/2050
+#[test]
+fn double_unary_boolean_literal() {
```
inoas, line 550: bool

compiler-core/src/warning.rs
```diff
@@ -382,6 +382,36 @@ expression.",
...
+                        src: src.clone(),
+                        path: path.to_path_buf(),
+                        label: diagnostic::Label {
+                            text: None,
+                            span: *location,
+                        },
+                        extra_labels: Vec::new(),
+                    }),
+                },
+                type_::Warning::UnnecessaryDoubleBooleanNegation { location } => Diagnostic {
```
inoas, line 400: Bool

compiler-core/src/warning.rs
```diff
@@ -382,6 +382,36 @@ expression.",
...
+                    title: "Unnecessary double negation (!!) on boolean.".into(),
```
inoas, line 401: bool maybe

inoas (commented):
I think it should be Bool or bool over Boolean or boolean everywhere except the hints where it should say: `Unnecessary double negation (!!) on boolean.` But I am not sure how prevalent bool versus boolean is to non CS people.

lpil:
Agreed with @inoas , let's use bool everywhere as that's what it is called in Gleam.

bcpeinhardt:
Sounds good, driving most of today but can update this tonight

bcpeinhardt:
Everything has been renamed to bool :)

lpil (approved):
Wonderful work, thank you so much!

lpil:
I've rebased this into main, thank you!

lpil:
oops I forgot to close after rebasing

PR #2126, opened 2023-04-16
2103 Allow spread when constructor has no labeled fields

### LINKED ISSUE #2103: Spread match errors when constructor has no labeled fields
Submitting as requested in Discord!

Example:

```gleam
type Thing {
  NoLabel(Int, Int)
  HasLabel(Int, label: Int)
}

fn assert_has_label(thing: Thing) -> Nil {
  let assert HasLabel(..) = thing
  Nil
}

fn assert_no_label(thing: Thing) -> Nil {
  let assert NoLabel(..) = thing
  Nil
}
```

The compiler errors only on the `NoLabel` match:

```
error: Incorrect arity
   │
12 │   let assert NoLabel(..) = thing
   │              ^^^^^^^^^^^ Expected 2 arguments, got 0
```

DanielleMaywood:
Fixes #2103. This change allows the spread operator to be used on a record constructor.

### DIFF
--- a/compiler-core/src/type_/pattern.rs
+++ b/compiler-core/src/type_/pattern.rs
@@ -486,7 +486,7 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                                         },
                                         location: spread_location,
                                         label: None,
-                                        implicit: false,
+                                        implicit: true,
                                     });
                                 }
                             };
--- a/compiler-core/src/type_/pattern.rs
+++ b/compiler-core/src/type_/pattern.rs
@@ -465,8 +465,33 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                         field_map.reorder(&mut pattern_args, location)?
                     }
 
-                    // The fun has no field map and so we error if arguments have been labelled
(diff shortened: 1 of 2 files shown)


compiler-core/src/type_/pattern.rs
```diff
@@ -465,8 +465,33 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
...
+                                while pattern_args.len() < usize::from(*arity) {
+                                    pattern_args.push(CallArg {
+                                        value: Pattern::Discard {
+                                            name: "_".into(),
+                                            location: spread_location,
+                                            type_: (),
+                                        },
+                                        location: spread_location,
+                                        label: None,
+                                        implicit: false,
```
lpil, line 489: Not that it does anything here, but I think these are implicit arguments

DanielleMaywood: Yeah I can change that, I was just copying the segment above as that has `implicit: false`.

lpil (approved):
Perfect, thank you! 💜 

I'll update the changelog

PR #2137, opened 2023-04-28
Add language server compilation tests.

samwightt:
Hello! I'm interested in helping with work on the language server, so I figured I'd learn how it works by writing some more tests.

This PR does the following:
- Moves the test helpers for the language server into `tests/common/mod.rs` (suggested in the Rust book, lemme know if this isn't preferred).
- Renames some of the test methods so it's a little more clear what's going on.
- Refactors a few of the existing test cases.
- Adds doc comments to methods I touched.
- Adds test cases to cover more use cases.


compiler-core/src/language_server/engine.rs
```diff
@@ -80,6 +84,11 @@ where
         })
     }
 
+    /// Compiles the current project, returning a `Response` with any warnings and the
+    /// paths of any modules that were compiled.
+    ///
+    /// If there is no project, or the project is not a Gleam project, then
+    /// nothing is compiled.
```
lpil, line 91:
```suggestion
```

Not quite! Each engine always has a project. If the request is for a file outside a project then there is no engine.

compiler-core/src/language_server/tests/common/mod.rs
```diff
@@ -0,0 +1,253 @@
+use std::{
```
lpil, line 1: Could you revert this and return this code to its original location please? We don't have a `common/mod.rs` convention.

compiler-core/src/language_server/tests/compilation.rs
```diff
@@ -1,95 +1,157 @@
...
+    assert_eq!(
+        actual_actions,
+        [actions_during_new, actions_during_compile].concat()
+    );
+}
+
+#[test]
+fn compile_project_successfully() {
+    let io = LanguageServerTestIO::new();
+    let mut engine = setup_language_server(&io);
```
lpil, line 35: This method creates an engine instance rather than a language server instance, let's revert to the previous name.

compiler-core/src/language_server/tests/compilation.rs
```diff
@@ -1,95 +1,157 @@
...
 }
 
 #[test]
 fn compile_error_in_src() {
     let io = LanguageServerTestIO::new();
-    let mut engine = setup_engine(&io);
+    let mut engine = setup_language_server(&io);
 
-    io.src_module("app/error", "pub type Error {");
+    io.write_source_module("app/error", "pub type Error {");
```
lpil, line 136: Revert the name of this method please.

compiler-core/src/language_server/tests/compilation.rs
```diff
@@ -1,95 +1,157 @@
...
 
     let response = engine.compile_please();
+
```
lpil, line 139: Revert the deletion of the remaining assertions please.

lpil (commented):
Thank you! I've left some notes inline.

compiler-core/src/language_server/tests/compilation.rs
samwightt: Were you wanting each test method to have the action assertions, or just the error cases here?

compiler-core/src/language_server/tests/compilation.rs
```diff
@@ -70,26 +172,30 @@ fn compile_error_in_test() {
...
+        Action::DependencyDownloadingFinished,
+        Action::LockBuild,
+        Action::UnlockBuild,
+    ];
+    let actions_during_compile = vec![
+        Action::CompilationStarted,
+        Action::LockBuild,
+        Action::UnlockBuild,
+        Action::CompilationFinished,
+    ];
```
lpil, line 194: Could you revert these changes that split the vector in two please? 🙏  And in the other test also.

compiler-core/src/language_server/tests/mod.rs
```diff
@@ -22,8 +20,10 @@ use crate::{
     Result,
 };
 
+mod compilation;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
-enum Action {
+pub enum Action {
```
lpil, line 26: Could you revert the changes to this file that make the definitions public? Thanks

compiler-core/src/language_server/tests/compilation.rs
lpil: Up to you. :)

Issue #1964, opened 2023-01-18
Do not download optional dependencies by default


Nicd:
Start a new project with the following deps:

```toml
[dependencies]
gleam_stdlib = "~> 0.25"
phoenix = "~> 1.6.15"

[dev-dependencies]
gleeunit = "~> 0.7"
```

Use `gleam deps download` to get the deps. Observe `gleam deps list`:

```
castore 1.0.0
cowboy 2.9.0
cowboy_telemetry 0.4.0
cowlib 2.11.0
decimal 2.0.0
gleam_stdlib 0.26.0
gleeunit 0.9.0
jason 1.4.0
mime 2.0.3
phoenix 1.6.15
phoenix_html 3.2.0
phoenix_pubsub 2.1.1
phoenix_template 1.0.0
phoenix_view 2.0.2
plug 1.14.0
plug_cowboy 2.6.0
plug_crypto 1.2.3
ranch 1.8.0
telemetry 1.2.1
```

There is a whole bunch of stuff here. But looking at the Phoenix package, jason and plug_cowboy are optional but still included. Indirectly a lot of other stuff is pulled like cowboy, ranch, phoenix_html, decimal… I guess the default should be to _not_ download optional deps.

lpil:
Sounds good

Issue #1338, opened 2021-11-22
Support git deps with build tool


lpil:
Gleam only, no need for rebar3 currently

https://github.com/elixir-lang/elixir/blob/79cd891eb86ecb7654a7acdb63769cfdd950a5c0/lib/mix/lib/mix/scm/git.ex

- [x] Update gleam.toml to include git deps. Will require a fancy serde deserializer.
- [ ] Download top level git deps
- [ ] Parse gleam.toml for them
- [ ] Add info to manifest
- [ ] Integrate into the version resolver
- [ ] Assert there are no git deps when publishing

Anything else? Can't think of anything.

## Config

```toml
[deps]
short_hex_package = "~> 1.0"
long_hex_package = { version = "~> 1.0" }
git_package = { git = "git@github.com:gleam-lang/stdlib.git", ref = "284b9f0" }

# We could have this shortcut too
github_package = { github = "gleam-lang/stdlib", ref = "284b9f0" }
```

HarryET:
I'd like to have a look at this, but would do #1409 first as that is the basis for this issue.

PgBiel:
I have been experimenting with this feature locally, and would like to discuss some of my findings so we can better define what to do next.

First, I noticed that there are two moments within `gleam-cli/src/dependencies.rs` in which we may need to clone or update a Git repository.

1. When creating the manifest, we need access to the updated Git repository to be able to check for its transitive dependencies and add them to the manifest, as well as check if there are no version conflicts.
2. After the manifest is ready, we still need to ensure all Git dependencies were updated, since it's possible the manifest didn't have to be updated but no dependencies were fetched yet (e.g. we ran `gleam clean`). Therefore, in the "downloading packages" step, we also need to consider Git dependencies.

Therefore, in my initial approach, I simply cloned the Git dependency to `build/packages/<name>` on those two occasions whenever I detected the need for it, and it sort of works.

[...]

PR #2153, opened 2023-05-15
feat: Local dependencies

### LINKED ISSUE #1409: Allow specifying a local project as a build dependency
Build tool feature request: Allow specifying a local project as a build dependency.

For example, I've previously used the following with rebar3 to specify a local project `local_project` as a dependency:

```
{deps, [
    {local_project, {path, "local_project"}},
]}. 
```

It would be cool if the new Gleam build tool supported this kind of build pattern as well.

I hope I am not creating a duplicate issue :-)

NthTensor:
Resolves #1409. Includes the groundwork for #1338.

Dependencies can now be specified in multiple ways in `gleam.toml`.
```toml
dep1 = "~> 1.0.0"
dep2 = { version = "== 2.1.4" }
dep3 = { path = "./relative/path/to/project" }
```

Path dependencies are walked recursively, so that local dependencies of local dependencies can still be located.

Todo:
- [x] Fix package linking
- [x] Fix `Recipe` toml encoding
- [x] Fix `ManifestPackage::Local` toml encoding
- [x] Prevent path cycles and infinite loops
- [x] Add support for absolute paths
- [x] Ensure that local packages are added to `packages.toml` after linking
- [x] Verify that packages with local dependencies cannot be pushed to hex
- [x] Fix checks and format

Future work related to this PR:
+ #1338
+ #1964
+ Add optional version range specified to local dependencies as a fallback/for publishing (like in rust)
+ Change detection

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -210,6 +210,7 @@ async fn add_missing_packages<Telem: Telemetry>(
         match &package.source {
             ManifestPackageSource::Hex { .. } => Ok(()),
             ManifestPackageSource::Local { path } => {
+                fs.delete(&package_dest)?;
                 fs.mkdir(&package_dest)?;
                 fs.copy_dir(&path.join("src"), &package_dest)?;
                 fs.copy(&path.join("gleam.toml"), &package_dest.join("gleam.toml"))
@@ -322,7 +323,12 @@ impl LocalPackages {
         manifest
             .packages
             .iter()
-            .filter(|p| p.name != root && self.packages.get(&p.name) != Some(&p.version))
+            .filter(|p| {
+                p.name != root && {
+                    self.packages.get(&p.name) != Some(&p.version)
+                        || matches!(p.source, ManifestPackageSource::Local { .. })
+                }
+            })
             .collect()
     }
 
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -794,7 +794,7 @@ fn provide_wrong_package() {
     let mut provided = HashMap::new();
     let result = provide_local_package(
         "wrong_name".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
     );
@@ -815,7 +815,7 @@ fn provide_existing_package() {
 
     let result = provide_local_package(
         "hello_world".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
     );
@@ -826,7 +826,7 @@ fn provide_existing_package() {
 
     let result = provide_local_package(
         "hello_world".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
     );
@@ -842,7 +842,7 @@ fn provide_conflicting_package() {
 
     let result = provide_local_package(
         "hello_world".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
     );
@@ -853,9 +853,9 @@ fn provide_conflicting_package() {
 
     let result = provide_package(
         "hello_world".into(),
-        &Path::new("../test/other"),
+        &Path::new("./test/other"),
         ProvidedPackageSource::Local {
-            path: Path::new("../test/other").to_path_buf(),
+            path: Path::new("./test/other").to_path_buf(),
         },
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
@@ -872,7 +872,7 @@ fn provided_is_absolute() {
     let mut provided = HashMap::new();
     let result = provide_local_package(
         "hello_world".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "subpackage".into()],
     );
@@ -893,7 +893,7 @@ fn provided_recursive() {
     let mut provided = HashMap::new();
     let result = provide_local_package(
         "hello_world".into(),
-        &Path::new("../test/hello_world"),
+        &Path::new("./test/hello_world"),
         &mut provided,
         &mut vec!["root".into(), "hello_world".into(), "subpackage".into()],
     );
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -213,6 +213,7 @@ async fn add_missing_packages<Telem: Telemetry>(
                 fs.delete(&package_dest)?;
                 fs.mkdir(&package_dest)?;
                 fs.copy_dir(&path.join("src"), &package_dest)?;
+                fs.copy_dir(&path.join("priv"), &package_dest)?;
                 fs.copy(&path.join("gleam.toml"), &package_dest.join("gleam.toml"))
             }
             ManifestPackageSource::Git { .. } => Ok(()),
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -713,76 +713,74 @@ fn provide_package(
     provided: &mut HashMap<SmolStr, ProvidedPackage>,
     parents: &mut Vec<SmolStr>,
 ) -> Result<hexpm::version::Range> {
+    // Return early if a package cyle is detected
     if parents.contains(&package_name) {
-        // Dependency cycled detected
-        Err(Error::ProvidedDependencyCycle(
+        return Err(Error::ProvidedDependencyCycle(
             package_name.to_string(),
             parents.join(" -> "),
-        ))
-    } else {
-        // This package is not in a dependency cycle
-        match provided.get(&package_name) {
-            None => {
-                // No package with this name has been provided load the project config
-                let config = crate::config::read(package_path.join("gleam.toml"))?;
-                // Check that we are loading the correct project
-                if config.name != package_name {
-                    return Err(Error::DependencyResolutionFailed(format!(
-                        "{} was expected but {} was found",
-                        package_name, config.name
-                    )));
-                };
-                // Walk the requirements of the package
-                let mut requirements = HashMap::new();
-                parents.push(package_name);
-                for (name, requirement) in config.dependencies.into_iter() {
-                    let name = SmolStr::from(name);
-                    let version = match requirement {
-                        Requirement::Hex { version } => version,
-                        Requirement::Path { path } => {
-                            let resolved_path = if path.is_absolute() {
-                                path
-                            } else {
-                                package_path.join(path)
-                            };
-                            // Recursively walk local packages
-                            provide_local_package(name.clone(), &resolved_path, provided, parents)?
-                        }
-                        Requirement::Git { git } => {
-                            provide_git_package(name.clone(), &git, provided)?
-                        }
-                    };
-                    let _ = requirements.insert(name, version);
-                }
-                let _ = parents.pop();
-                // Add the package to the set
-                let version = hexpm::version::Range::new(format!("== {}", &config.version));
-                let _ = provided.insert(
-                    config.name,
-                    ProvidedPackage {
-                        version: config.version,
-                        source: package_source,
-                        requirements,
-                    },
-                );
-                // Return the version
-                Ok(version)
-            }
-            Some(package) if package.source == package_source => {
-                // This package has already been provided from this source, return the version
-                let version = hexpm::version::Range::new(format!("== {}", &package.version));
-                Ok(version)
-            }
-            Some(package) => {
-                // This package has already been provided from a different source which conflicts
-                Err(Error::ProvidedDependencyConflict(
-                    package_name.to_string(),
-                    package_source.to_toml(),
-                    package.source.to_toml(),
-                ))
-            }
-        }
+        ));
     }
+
+    // Check that we do not have a cached version of this package already
+    match provided.get(&package_name) {
+        Some(package) if package.source == package_source => {
+            // This package has already been provided from this source, return the version
+            let version = hexpm::version::Range::new(format!("== {}", &package.version));
+            return Ok(version);
+        }
+        Some(package) => {
+            // This package has already been provided from a different source which conflicts
+            return Err(Error::ProvidedDependencyConflict(
+                package_name.to_string(),
+                package_source.to_toml(),
+                package.source.to_toml(),
+            ));
+        }
+        None => (),
+    }
+
+    // Load the package
+    let config = crate::config::read(package_path.join("gleam.toml"))?;
+    // Check that we are loading the correct project
+    if config.name != package_name {
+        return Err(Error::DependencyResolutionFailed(format!(
+            "{} was expected but {} was found",
+            package_name, config.name
+        )));
+    };
+    // Walk the requirements of the package
+    let mut requirements = HashMap::new();
+    parents.push(package_name);
+    for (name, requirement) in config.dependencies.into_iter() {
+        let name = SmolStr::from(name);
+        let version = match requirement {
+            Requirement::Hex { version } => version,
+            Requirement::Path { path } => {
+                let resolved_path = if path.is_absolute() {
+                    path
+                } else {
+                    package_path.join(path)
+                };
+                // Recursively walk local packages
+                provide_local_package(name.clone(), &resolved_path, provided, parents)?
+            }
+            Requirement::Git { git } => provide_git_package(name.clone(), &git, provided)?,
+        };
+        let _ = requirements.insert(name, version);
+    }
+    let _ = parents.pop();
+    // Add the package to the provided packages dictionary
+    let version = hexpm::version::Range::new(format!("== {}", &config.version));
+    let _ = provided.insert(
+        config.name,
+        ProvidedPackage {
+            version: config.version,
+            source: package_source,
+            requirements,
+        },
+    );
+    // Return the version
+    Ok(version)
 }
 
 #[test]
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -218,10 +218,7 @@ async fn add_missing_packages<Telem: Telemetry>(
     let mut num_to_download = 0;
     let mut missing_hex_packages = missing_packages
         .into_iter()
-        .filter(|package| match package.source {
-            ManifestPackageSource::Hex { .. } => true,
-            _ => false,
-        })
+        .filter(|package| package.is_hex())
         .map(|package| {
             num_to_download += 1;
             package
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -543,7 +543,7 @@ fn resolve_versions<Telem: Telemetry>(
     let resolved = dependency::resolve_versions(
         PackageFetcher::boxed(runtime.clone()),
         provided_packages.clone(),
-        config.name.to_string(),
+        config.name.clone(),
         version_requirements.into_iter(),
         &locked,
     )?;
@@ -577,7 +577,7 @@ fn resolve_versions<Telem: Telemetry>(
 
 #[derive(Clone, Eq, PartialEq, Debug)]
 enum ProviderInfo {
-    Git { repo: String, commit: String },
+    Git { repo: SmolStr, commit: SmolStr },
     Local { path: PathBuf },
 }
 
@@ -626,8 +626,8 @@ fn provide_git_package(
 ) -> Result<hexpm::version::Range> {
     // TODO
     let _git = ProviderInfo::Git {
-        repo: "repo".to_string(),
-        commit: "commit".to_string(),
+        repo: SmolStr::new_inline("repo"),
+        commit: SmolStr::new_inline("commit")
     };
     Err(Error::DependencyResolutionFailed(
         "Git dependencies are not supported".to_string(),
@@ -724,8 +724,8 @@ async fn lookup_package(
                 build_tools: vec!["gleam".to_string()],
                 requirements,
                 source: ManifestPackageSource::Git {
-                    repo: repo.to_string(),
-                    commit: commit.to_string(),
+                    repo: repo.clone(),
+                    commit: commit.clone(),
                 },
             })
         }
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -208,9 +208,10 @@ async fn add_missing_packages<Telem: Telemetry>(
     // Link local paths
     let packages_dir = paths.build_packages_directory();
     for package in missing_packages.iter() {
+        let package_dest = packages_dir.join(project_name.to_string());
         match &package.source {
             ManifestPackageSource::Hex { .. } => Ok(()),
-            ManifestPackageSource::Local { path } => fs.symlink_dir(&path, &packages_dir),
+            ManifestPackageSource::Local { path } => fs.symlink_dir(&path, &package_dest),
             ManifestPackageSource::Git { .. } => Ok(()),
         }?
     }
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -785,7 +785,7 @@ fn provide_wrong_package() {
         "wrong_name".into(),
         &Path::new("../test/hello_world"),
         &mut provided,
-        &mut vec!["root".into(), "subpackage".into()]
+        &mut vec!["root".into(), "subpackage".into()],
     );
     let error = Err(Error::DependencyResolutionFailed(
         "wrong_name was expected but hello_world was found".to_string(),
@@ -801,7 +801,7 @@ fn provide_existing_package() {
         "hello_world".into(),
         &Path::new("../test/hello_world"),
         &mut provided,
-        &mut vec!["root".into(), "subpackage".into()]
+        &mut vec!["root".into(), "subpackage".into()],
     );
     assert_eq!(
         result,
@@ -812,7 +812,7 @@ fn provide_existing_package() {
         "hello_world".into(),
         &Path::new("../test/hello_world"),
         &mut provided,
-        &mut vec!["root".into(), "subpackage".into()]
+        &mut vec!["root".into(), "subpackage".into()],
     );
     assert_eq!(
         result,
@@ -828,7 +828,7 @@ fn provide_conflicting_package() {
         "hello_world".into(),
         &Path::new("../test/hello_world"),
         &mut provided,
-        &mut vec!["root".into(), "subpackage".into()]
+        &mut vec!["root".into(), "subpackage".into()],
     );
     assert_eq!(
         result,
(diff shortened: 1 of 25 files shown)


lpil:
Amazing!! Do you want me to review this now?

I think we'll need to do at least these ones before releasing, but it can be in later PRs.

- Prevent path cycles and infinite loops
- Add support for absolute paths
- Ensure that local packages are added to packages.toml after linking
- Verify that packages with local dependencies cannot be pushed to hex

compiler-cli/src/dependencies.rs
```diff
@@ -201,30 +203,52 @@ pub fn download<Telem: Telemetry>(
...
-        .missing_local_packages(manifest, &project_name)
+    let missing_packages = local.missing_local_packages(manifest, &project_name);
+
+    // Link local paths
+    let packages_dir = paths.build_packages_directory();
+    for package in missing_packages.iter() {
+        let package_dest = packages_dir.join(package.name.to_string());
+        match &package.source {
+            ManifestPackageSource::Hex { .. } => Ok(()),
+            ManifestPackageSource::Local { path } => fs.symlink_dir(&path, &package_dest),
```
lpil, line 222: Symlinks are not supported on Windows so we'll need to do something else here.

Long term this is likely using the build directory within the target directory itself, but for now we could copy the `gleam.toml`, `src`, and `priv`.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+
+    let provided_package_requirements = provided_packages
+        .into_iter()
+        .map(|(name, package)| {
+            (
+                name,
+                package.releases[0].requirements.keys().cloned().collect(),
+            )
+        })
+        .collect();
```
lpil, line 554: Looks like we can avoid taking ownership of the data and doing all the cloning here as it's only used by reference.

Why is it that we have the structure `package.releases[0]` when we constructed this packages and so earlier had their requirements to hand? If we don't put them into a vector we can statically ensure they are present rather than doing a runtime check.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+
+fn provide_local_package(
+    package_name: &str,
+    package_path: &Path,
+    info: &mut HashMap<String, ProviderInfo>,
+    provided: &mut HashMap<String, hexpm::Package>,
+) -> Result<hexpm::version::Range> {
+    let canonical_path = package_path
+        .canonicalize()
+        .expect("local package cannonical path");
```
lpil, line 587: This comes from user input, right? If so it could fail so we'd need to return an error rather than crashing.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+    let package_info = ProviderInfo::Local {
+        path: canonical_path.clone(),
     };
-    Ok(manifest)
+
+    // Determine if package has already been walked
+    match info.insert(package_name.to_string(), package_info.clone()) {
```
lpil, line 593: Let's use SmolStr here.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+        None => {
+            // No package with this name has been found yet
+            provide_package(package_name, &canonical_path, info, provided)
+        }
+        Some(existing_package_info) => {
+            // A package with this name has already been found
+            // True only if they are both local with the same canonical path, or both git with the same repo and commit
```
lpil, line 600: When can this happen?

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+            if existing_package_info == package_info {
+                // It is the same package, do not parse it again
+                let config = crate::config::read(package_path.join("gleam.toml"))?;
+                Ok(hexpm::version::Range::new(format!("== {}", config.version)))
```
lpil, line 604: Can we avoid performing the IO again here? It'll be the same each time.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+            } else {
+                // A different source was provided for this package
+                Err(Error::DependencyResolutionFailed(format!(
+                    "{} has multiple conflicting definition",
+                    package_name
+                )))
```
lpil, line 610: Let's make a proper error variant rather than using a string here.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +516,230 @@ fn resolve_versions<Telem: Telemetry>(
...
+    _provided: &mut HashMap<String, hexpm::Package>,
+) -> Result<hexpm::version::Range> {
+    // TODO
+    let _git = ProviderInfo::Git {
+        repo: "repo".to_string(),
+        commit: "commit".to_string(),
+    };
+    Err(Error::DependencyResolutionFailed(
+        "Git dependencies are not supported".to_string(),
+    ))
```
lpil, line 629: Let's make an error variant rather than using a string here.

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+use crate::config::PackageConfig;
+use crate::{Error, Result};
+use hexpm::version::Range;
+use serde::de::{self, Deserializer, MapAccess, Visitor};
+use serde::ser::{Serialize, SerializeMap, Serializer};
+use serde::Deserialize;
+
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(untagged, remote = "Self")]
+pub enum Recipe {
```
lpil, line 14: Can we call this a requirement?

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+    Hex { version: Range },
+    Path { path: PathBuf },
+    Git { git: String },
```
lpil, line 17:
```suggestion
    Git { git: SmolStr },
```

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+}
+
+impl Recipe {
+    pub fn version_range(&self) -> Result<Range, Error> {
+        match self {
+            Recipe::Hex { version: range } => Ok(range.clone()),
+            Recipe::Path { path } => {
+                let mut config_path = path.clone();
+                config_path.push("gleam.toml");
```
lpil, line 26:
```suggestion
                let config_path = path.join("gleam.toml");
```
Bit more concise and remove an allocation.

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+                let toml = std::fs::read_to_string(&config_path).map_err(|_| {
+                    Error::DependencyResolutionFailed("Local dependency could not be found".into())
+                })?;
+                let config: PackageConfig = toml::from_str(&toml).map_err(|_| {
+                    Error::DependencyResolutionFailed(
+                        "Local dependency config could not be parsed".into(),
+                    )
+                })?;
+                Ok(Range::new(format!("== {}", config.version.to_string())))
```
lpil, line 35:
```suggestion
                Ok(Range::new(format!("== {}", config.version)))
```

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+                let mut config_path = path.clone();
+                config_path.push("gleam.toml");
+                let toml = std::fs::read_to_string(&config_path).map_err(|_| {
+                    Error::DependencyResolutionFailed("Local dependency could not be found".into())
+                })?;
+                let config: PackageConfig = toml::from_str(&toml).map_err(|_| {
+                    Error::DependencyResolutionFailed(
+                        "Local dependency config could not be parsed".into(),
+                    )
+                })?;
```
lpil, line 34: Let's use the existing error code for loading the config file. If we want a custom error `gleam.toml` being absent (which is a good idea!) let's check if the file exists first.

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+                Ok(Range::new(format!("== {}", config.version.to_string())))
+            }
+            Recipe::Git { .. } => Err(Error::DependencyResolutionFailed(
+                "Git dependencies are currently unsuported".to_string(),
```
lpil, line 38: Let's use make a proper error for this

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+
+    pub fn path(path: &str) -> Recipe {
+        Recipe::Path { path: path.into() }
+    }
+
+    pub fn git(url: &str) -> Recipe {
+        Recipe::Git {
+            git: url.to_string(),
+        }
+    }
```
lpil, line 57: Let's move these into the test module as we don't want them use in application code.

compiler-core/src/recipe.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn to_string(&self) -> String {
+        match self {
+            Recipe::Hex { version: range } => {
+                format!(r#"{{ version = "{}" }}"#, range.to_string())
+            }
+            Recipe::Path { path } => format!(r#"{{ path = "{}" }}"#, path.display()),
+            Recipe::Git { git: url } => format!(r#"{{ git = "{}" }}"#, url),
+        }
+    }
+}
```
lpil, line 72: What's this used for? I don't think we'd want the ToString trait implemented for these, and this looks specifically like toml serialisation rather than a string representation.

compiler-core/src/manifest.rs
```diff
@@ -247,6 +262,10 @@ impl Default for ManifestPackage {
 pub enum ManifestPackageSource {
     #[serde(rename = "hex")]
     Hex { outer_checksum: Base16Checksum },
+    #[serde(rename = "git")]
+    Git { repo: String, commit: String },
+    #[serde(rename = "local")]
+    Local { path: PathBuf }, // should be the canonical path
```
lpil, line 268: What does this comment mean?

compiler-core/src/manifest.rs
```diff
@@ -166,10 +180,10 @@ packages = [
...
 
 [requirements]
-aaa = "> 0.0.0"
-gleam_stdlib = "~> 0.17"
-gleeunit = "~> 0.1"
-zzz = "> 0.0.0"
+aaa = { version = "> 0.0.0" }
+gleam_stdlib = { version = "~> 0.17" }
+gleeunit = { version = "~> 0.1" }
+zzz = { version = "> 0.0.0" }
```
lpil, line 186: Could we have tests for the new variants also? 🙏

compiler-core/src/hex.rs
```diff
@@ -174,6 +154,15 @@ impl Downloader {
         &self,
         package: &ManifestPackage,
     ) -> Result<bool, Error> {
+        let outer_checksum = if let ManifestPackageSource::Hex { outer_checksum } = &package.source
+        {
+            outer_checksum
+        } else {
+            return Err(Error::DependencyResolutionFailed(
+                "Attempt to download non-hex package from hex".to_string(),
+            ));
```
lpil, line 163: Let's panic here, it's a bug rather than user error.

compiler-cli/src/publish.rs
```diff
@@ -186,12 +187,25 @@ fn check_config_for_publishing(config: &PackageConfig) -> Result<()> {
...
+        .map(|(name, recipe)| match recipe {
+            Recipe::Hex { version } => Ok(ReleaseRequirement {
+                name,
+                requirement: version,
+            }),
+            _ => Err(Error::PublishNonHexDependencies {
+                package: name.to_string(),
+            }),
+        })
+        .collect();
```
lpil, line 208: Would be good to have a test for this.

compiler-core/src/manifest.rs
```diff
@@ -247,6 +262,10 @@ impl Default for ManifestPackage {
 pub enum ManifestPackageSource {
     #[serde(rename = "hex")]
     Hex { outer_checksum: Base16Checksum },
+    #[serde(rename = "git")]
+    Git { repo: String, commit: String },
```
lpil, line 266:
```suggestion
    Git { repo: SmolStr, commit: SmolStr },
```

compiler-core/src/dependency.rs
```diff
@@ -0,0 +1,639 @@
...
+pub type PackageVersions = HashMap<String, Version>;
+
+pub type ResolutionError = PubGrubError<String, Version>;
+
+type PubgrubRange = pubgrub::range::Range<Version>;
+
+pub fn resolve_versions<Requirements>(
+    package_fetcher: Box<dyn PackageFetcher>,
+    provided_packages: HashMap<String, hexpm::Package>,
+    root_name: String,
```
lpil, line 24:
```suggestion
    root_name: SmolStr,
```

compiler-core/src/dependency.rs
```diff
@@ -0,0 +1,639 @@
...
+    //             package,
+    //             Dependency {
+    //                 app: None,
+    //                 optional: false,
+    //                 repository: None,
+    //                 requirement,
+    //             },
+    //         )
+    //     })
+    //     .collect()
```
lpil, line 140:
```suggestion
```

NthTensor:
Thanks for the review! Shouldn't take too long to get these fixed! And yep, coverage is definitely lacking. I know how I want to unit test most of this stuff (except for the `providing_x` functions, those may be tricky) but it will be simpler for me to just implement it than try to plan it out here. That ok with you?

lpil:
Yeah totally! Thank you

compiler-cli/src/dependencies.rs
NthTensor: The most common example would just be cycles, where A has a path dependency to B and B has a path dependency to A. In this case we only want to load each once (to prevent infinite recursion) and let pubgrub explain that theres a cycle with it's fancy errors.

A (much less common) example is where A has path dependencies B and C, and both B and C depend on D, but B and C use different local versions of D. In this case we do need to return an error. This probably will never happen currently but (1) its correct to check for it and (2) possibly more likely with a mixture of git and relative local dependencies.

compiler-cli/src/dependencies.rs
NthTensor: Yeah I was looking at that when I did my last commit but I figured we could get to it later. I just need to add versions to the `info` enum.

compiler-core/src/recipe.rs
NthTensor: This comes from a very frustrating trait impl error somewhere in the manifest that I didn't want to spend time dissecting. I will see if I can move it.

compiler-core/src/manifest.rs
NthTensor: By the time this path reaches the manifest it should be in canonical format. At one point this was important, but I don't think it is any more. 

Nightly has a type for canonical paths, but I might just make a newtype to fill the gap.

compiler-cli/src/dependencies.rs
NthTensor: Is this suggestion intended for all platforms to maintain uniformity, or just for windows?

compiler-core/src/recipe.rs
NthTensor: I didn't answer your question. Yes this is used for toml serialization, and yes it 100% shouldn't be there.

NthTensor:
Currently `PackageConfig` uses `SmolStr` but `ManifestPackage` and some of the `hexpm` stuff is done with `String`. Its going to take a little longer for me to make sure I get the boundary between these right, unless you want me to just convert everything to `SmolStr`.

lpil:
Everything should be smolstr up until it reaches the hex library 👍 

A thought: If running `gleam deps download` is required to pull in any new changes from the local dep how does the programmer get the latest code when using the language server?

NthTensor:
Also I've responded to a few of your inline questions, and had a specific question about this suggestion https://github.com/gleam-lang/gleam/pull/2153#discussion_r1196596462.

compiler-cli/src/dependencies.rs
lpil: Each approach would have different user facing behaviour when files get edited, so I think we should do the same thing on all platforms.

lpil:
Your idea of some coordination between each project in the LSP could be the way to go, though needs some more thought as the dependency project may not be open in the editor at all.

Rather than getting hung-up on that let's have an issue for that aspect of the work also.

compiler-core/src/dependency.rs
```diff
@@ -0,0 +1,639 @@
...
+    .filter(|(name, _)| name.as_str() != root_name.as_str())
+    .collect();
+
+    Ok(packages)
+}
+
+fn root_dependencies<Requirements>(
+    base_requirements: Requirements,
+    locked: &HashMap<String, Version>,
+) -> Result<HashMap<String, Dependency>, ResolutionError>
```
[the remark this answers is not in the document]
NthTensor: This feeds into `Hexpm::Release` which has field `requirements: HashMap<String, Dependency>`. I think it makes sense to leave this as is. With this change I would still need to call `to_string` and probably add an extra iterator [here](https://github.com/gleam-lang/gleam/pull/2153/files#diff-977b3c4663dc111b507dc49768a4ab2bfc6841e9bf043610499307e2d11999efR40).

compiler-cli/src/dependencies.rs
NthTensor: This would require changing `package_name` to `&SmolStr` but `package_name` is called with dependency names from the config where the type is `Dependencies = HashMap<String, Recipe>`. I can change this definition but I get worrying errors from in codegen. Otherwise I could call `SmallStr::from<String>` but it seems kind of contrived.

NthTensor:
I've replaced most uses of `String` with `SmolStr`, but notably not [here](https://github.com/gleam-lang/gleam/pull/2153#discussion_r1199659809) or [here](https://github.com/gleam-lang/gleam/pull/2153#discussion_r1199657376) for compatibility with existing code. The changes involved in switching these to `SmolStr` in any useful way seem large enough to deserve their own PRs (which I would be happy to open after this is merged).

compiler-cli/src/dependencies.rs
NthTensor: This is because the `Package` provided by `hexpm` expects a vector of releases. We can't port the package struct over, but We could perhaps be working with releases directly here instead of packages.

NthTensor: This is a map of `Package` rather than `Release` because we assembled it to pass into `resolve_versions`. I just to bite the bullet and make a `LocalPackage` struct which all these other things can be properly derived from.

NthTensor:
Ok, I *think* that should resolve all the comments. Going to keep going through and adding unit tests for now. Ready for review sometime tomorrow or the day after.

NthTensor:
Ok, I *think* that should resolve all the comments. Going to keep going through and adding unit tests for now. Ready for review sometime tomorrow or the day after.

NthTensor:
I think I've also worked out how to use local dependencies directly instead of copying them into `/build`. But since that will need to touch project compilation maybe we can bundle that together with the lang-server stuff and change-detection/compatibility-checking in a separate PR or PRs.

compiler-core/src/manifest.rs
NthTensor: I ended up adding a test for this instead. Its more important that paths are absolute.

NthTensor:
After more tested I realized I was a little hasty with that review request. Now be debuged, complete and ready for review.

lpil:
> I think I've also worked out how to use local dependencies directly instead of copying them into /build. But since that will need to touch project compilation maybe we can bundle that together with the lang-server stuff and change-detection/compatibility-checking in a separate PR or PRs.

Assuming you mean build each package in their own directory, I think that's likely the way to go. It would have other nice benefits also such as shared caches between different projects using the same path deps, which would be fantastic for monorepos.

One this we'll need to consider with this is different compiler versions clobbering each others state. I think we'd need to put the compiler version into the path within the build directory to get isolation. This would mean that old state from unused compilers never gets removed, but the build directory isn't huge, so I think that's OK.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +607,307 @@ fn resolve_versions<Telem: Telemetry>(
...
+        "hello_world".into(),
+        &Path::new("../test/hello_world"),
+        &mut provided,
+        &mut vec!["root".into(), "hello_world".into(), "subpackage".into()],
+    );
+    assert_eq!(
+        result,
+        Err(Error::ProvidedDependencyCycle(
+            "hello_world".to_string(),
+            "root -> hello_world -> subpackage".to_string()
```
lpil, line 882: This should be a vector which then gets printed using the cycle printing code in the error module.

compiler-cli/src/dependencies.rs
```diff
@@ -581,3 +965,196 @@ impl hexpm::version::PackageFetcher for PackageFetcher {
...
+        build_tools: vec!["gleam".to_string()],
+        requirements: vec!["req_1".into(), "req_2".into()],
+        source: ManifestPackageSource::Git {
+            repo: "https://github.com/gleam-lang/gleam.git".into(),
+            commit: "bd9fe02f72250e6a136967917bcb1bdccaffa3c8".into(),
+        },
+    };
+
+    assert_eq!(
+        provided_package.to_manifest_package(&SmolStr::new_inline("package")),
```
lpil, line 1157: Hot damn, I didn't know about this inline method!

compiler-core/src/config.rs
```diff
@@ -103,6 +104,13 @@ impl PackageConfig {
         }
     }
 
+    pub fn dependency_versions_for(&self, mode: Mode) -> Result<HashMap<String, Range>> {
```
lpil, line 107: Where does this function get used? I couldn't find it.

I ask because I was very surprised that `requirement.version_range` reads from the filesystem (and presumably will have to call `git clone` for git deps), the IO seemed out of place. I wanted to learn the context in which it was used so that I could understand better.

compiler-core/src/error.rs
```diff
@@ -180,18 +180,33 @@ pub enum Error {
...
+    DependencyCanonicalizationFailed(String),
+
     #[error("Dependency tree resolution failed: {0}")]
     DependencyResolutionFailed(String),
 
     #[error("The package {0} is listed in dependencies and dev-dependencies")]
     DuplicateDependency(String),
 
+    #[error("The package {0} is provided multiple times, as {1} and {2}")]
+    ProvidedDependencyConflict(String, String, String),
```
lpil, line 196: Could we name these fields please? I can't tell what the second and third strings are.

compiler-cli/src/dependencies.rs
```diff
@@ -261,6 +286,8 @@ fn write_manifest_to_disc(paths: &ProjectPaths, manifest: &Manifest) -> Result<(
     fs::write(&path, &manifest.to_toml())
 }
 
+// This is the container for locally pinned packages.
+// For packages provided by paths and git deps, see the ProvidedPackage struct
```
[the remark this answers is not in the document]
NthTensor: I will try to make that comment more clear. The two are orthogonal, and "provided dependencies" still also show up as "local dependencies" after they are added to `/build`. The names could be confusing, so I wanted to add a disambiguation.

compiler-core/src/config.rs
NthTensor: Good catch, it doesn't! Remanent of early work when I thought I might need to pre-check all the versions before actually loading the packages.

compiler-cli/src/dependencies.rs
```diff
@@ -492,38 +607,307 @@ fn resolve_versions<Telem: Telemetry>(
...
+        }
+    }
+}
+
+#[test]
+fn provide_wrong_package() {
+    let mut provided = HashMap::new();
+    let result = provide_local_package(
+        "wrong_name".into(),
+        &Path::new("../test/hello_world"),
```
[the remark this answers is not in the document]
NthTensor: It was, and I have moved it to `compiler-cli/test`.

NthTensor:
Great! All those changes should be in.

>  Sorry about all the back and forth, this is quite a complex feature!

Complex and with quite a lot of surface area for you to review! The quick turnaround time is a big help.

compiler-cli/src/dependencies.rs
NthTensor: Just used the existing error code for package cycles.

NthTensor:
> Assuming you mean build each package in their own directory, I think that's likely the way to go long-term. It would have other nice benefits also such as shared caches between different projects using the same path deps, which would be fantastic for monorepos.

Have you thought about the issue with multiple manifests? Each projects does its own solve for packages, and we ignore the target's package configuration when using it as a local dependency. Thats going to give us errors when the dependency and parent pin a different version of the same package. 

Concrete example: B has dependency `A >= 1.0.0` pinned to `1.2.3`. C has dependencies `B >= 1.0.0; A >= 1.4.0` with `A` pinned to `1.6.8`. This is a valid solve, but if we try to build `C` using the build caches of `B` we are going to get errors about `A`.

Even if this can't be resolved, we still should use the live versions of `src` and `priv` instead of copying them.

lpil:
> Have you thought about the issue with multiple manifests when using the in-project cache? ...

I did not, that is that idea out of the window then 😅 Good catch.

> Or we could go the "umbrella project" route, and formalize groupings of projects which share the same manifest.

I think something like this may be best. Some kind of opt-in shared manifest seems like a good approach. It's largely a compile performance optimisation so I think can ignore it for now, or perhaps make an issue after this is merged.

compiler-core/src/requirement.rs
```diff
@@ -0,0 +1,156 @@
...
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+#[serde(untagged, remote = "Self")]
+pub enum Requirement {
+    Hex { version: Range },
+    Path { path: PathBuf },
+    Git { git: SmolStr },
+}
+
+impl Requirement {
+    pub fn version_range(&self) -> Result<Range, Error> {
```
lpil, line 22: I think this can be removed too?

lpil (commented):
Fab! I think the feature might be ready? What do you think?

One last thing could be to have an integration test on CI for a package that has a path dep, similar to this: https://github.com/gleam-lang/gleam/blob/01f672bde28675aa0e5bf3a04b117ed314aa56b9/.github/workflows/ci.yaml#L128

NthTensor:
> That is that idea out of the window then.

We can still use the source directly instead of cloning it on update, but yeah cache and manifest sharing umbrella projects seems like a separate feature. Would you like me to open  a tracking issue/issues for these features?

NthTensor:
I've not written a CI test before, so hopefully I did it right. Otherwise I think this is ready!

test/project_path_deps/project_a/gleam.toml
```diff
@@ -0,0 +1,9 @@
+name = "project_a"
+version = "0.1.0"
+
+[dependencies]
+project_b = { path = "../project_b" }
+project_c = { path = "../project_c" }
```
lpil, line 6: That's so damn cool

test/project_path_deps/README.md
```diff
@@ -0,0 +1,12 @@
...
+
+The dependency graph of these projects is as follows:
+
+```
+         _-> project_b -_
+        /                v
+project_a                project_d
+        \                ^
+         '-> project_c -'
+```
```
lpil, line 11: Love it

lpil (approved):
This is incredible work! Thank you so much. I really appreciate both the quality of your work and your patience with how long this took.

lpil:
@NthTensor There was a few bits of future work mentioned in the discussion on this PR. Would you like to open issues for them or shall I read though and make some?

Issue #2155, opened 2023-05-15
Error message improvement for running modules


dangdennis:
On first encountering `gleam run -m codegen` in the gleam package index, I was unsure what's considered a valid module. I had a file at `src/project_name/x.gleam`. Then I figured I could write `gleam run -m src/project_name/x`. 

This yielded error:
```
error: Module does not exist

Module `src/project_name/x` was not found
Hint: Try creating the file `src/src/project_name/x.gleam`.
```

I think a specific error message can be added that informs gleam users of some these incidental knowledge:
1. only files placed inside `/test` and `/src` are modules
2. you don't need the `src/` or `test/` path prefix

New error message to handle this case. Should handle both /test and /src prefixes accordingly.
```
error: Module does not exist

You are calling a module with `/src` ((replace with /test when appropriate)). 
Only files within /src and /test are valid modules. 
Hint: Try running the file `project_name/the_path/without/src_or_test`.
```

lpil:
Good idea! Thank you.

We could even check to see if it starts with `src` or `test` and recommend the correct value if that module exists.

```
$ gleam run -m src/one/two
error: Module does not exist

No module was found with the name `src/one/two`.

Did you mean `one/two`, from the package `wibble`?
```

dangdennis:
@lpil can you assign this to me? Would like to see it through!

dangdennis:
Gasp there’s a gleam 1.0 milestone

lpil:
Sure thing! Thank you

dangdennis:
@lpil Can you point me in the direction on how to best find if a module is available within the project's dependency graph?

Got the first part done where I handle the src/ and test/ prefixes with the hint using a trimmed version of the module. Looking into the your suggestion now too.

Assuming there's an easy to use API already, I'll likely do that work in the run command handler, and pass the various values along into the `Error::ModuleDoesNotExist`.

lpil:
Fab! Thanks.

Here's where the searching for the module is done:

https://github.com/gleam-lang/gleam/blob/9507ea6876133ee2fc78f2c96db0736b56c03212/compiler-cli/src/run.rs#L43-L48

The logic will want to be something like this:

```rust
if found(module) {
    ok()
} else if (module.starts_with("src/") && found(module.strip_prefix("src/"))) {
    error_about_src_prefix()
} else if (module.starts_with("test/") && found(module.strip_prefix("test/"))) {
    error_about_test_prefix()
} else {
    error()
}
```

PR #2160, opened 2023-05-20
Error message for src/ and test/ prefixes when running modules

dangdennis:
Resolves the original issue https://github.com/gleam-lang/gleam/issues/2155

## Changes

I added a new error enum that is specific to calling modules with the extra `src/` and `test/` prefixes. The issue was that I attempted to execute modules with `src/x/y/z` instead of just `x/y/z`. It's also not clear that `src/` and `test/` are special directories. The new error message teaches users this.

I did not complete the additional work of suggesting an in-project module or 3rd party module by checking the dependency graph.

The results

```
error: Module does not exist

You are calling a module with `src/`.
Only files within src/ and test/ are valid modules.
Hint: Try running `okay/yo`.
```

### DIFF
--- a/compiler-cli/src/run.rs
+++ b/compiler-cli/src/run.rs
@@ -7,7 +7,7 @@ use gleam_core::{
 };
 use lazy_static::lazy_static;
 use smol_str::SmolStr;
-use std::path::PathBuf;
+use std::{path::PathBuf, println};
 
 use crate::fs::ProjectIO;
 
@@ -27,6 +27,8 @@ pub fn command(
 ) -> Result<(), Error> {
     let paths = crate::project_paths_at_current_directory();
 
+    println!("{:?}", paths);
+
     // Validate the module path
     if let Some(mod_path) = &module {
         if !is_gleam_module(mod_path) {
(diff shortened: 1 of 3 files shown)


compiler-cli/src/run.rs
```diff
@@ -34,6 +34,20 @@ pub fn command(
                 module: mod_path.to_owned(),
             });
         }
+
+        if mod_path.starts_with("src/") {
+            return Err(Error::InvalidModulePrefix {
+                module: mod_path.strip_prefix("src/").unwrap().to_string(),
```
dangdennis, line 40: unwrapping seemed safe here because we do a starts_with check. but what's a neater way of handling this? 

i considered handling the option directly or use unwrap_or_default, but both didnt feel right

example 1:
```
 match mod_path.strip_prefix("src/") {
                Some(path) => {
                    return Err(Error::InvalidModulePrefix {
                        module: path.to_string(),
                        prefix: "src/".to_string(),
                    });
                }
                _ => (),
            }
```

lpil: We'll still need to check if the module exists as `src/src/wibble.gleam` is a valid module.

So we'll need to look up the module and if it does not exist and it starts with src or test, then check if it exists once this prefix has been removed.

dangdennis: src/src/x is valid? Okay I’ll def check. 

1. Do I simply check that a file exists by using the std fs or is there something else you have in mind? I had thought you’d built a dependency graph prior so I can just check that, but direct file check was my initial thought.
2. Are you able to run a module within a third party package? Double checking here.

Lunarmagpie: > unwrapping seemed safe here because we do a starts_with check. but what's a neater way of handling this?

You can use if let for this!.

```rust
if let Some(path) = mod_path.strip_prefix("src/") {
	// logic
}
```

lpil: > Do I simply check that a file exists by using the std fs or is there something else you have in mind? I had thought you’d built a dependency graph prior so I can just check that, but direct file check was my initial thought.

You can use this function here, to match how we do it for the correct module name.

https://github.com/gleam-lang/gleam/blob/deb9f26fd66b33bcc5e2d7e96f0707d1519419b9/compiler-cli/src/run.rs#L72

Any checks for `src/` and `test/` should go after this existing check for the module name as-is.

> Are you able to run a module within a third party package? Double checking here.

Yup! This will all be handled by that method.

dangdennis: Thanks yall. I believe it's working correctly now. I wrote my manual testing strategy in the description. 
I feel inclined to write at least a couple tests here, but is this advised? And how should I go about it?
I updated the shell file `test/running_modules/run_tests.sh` as an initial test, but I'm not asserting anything yet.

lpil (approved):
Thank you!!

compiler-cli/src/run.rs
lpil: We don't have a good way to test this sort of build tool stuff yet I'm afraid :(

lpil:
Sorry, I had to squash as for some reason it wouldn't let me rebase.

dangdennis:
Ty! I was going to ask you to squash too. Much more of a squash person myself. 😇

dangdennis:
Ci failed on linting. Small change. Oops. And here I was wondering whether is _ok or if let Ok(_) is better.

PR #2180, opened 2023-05-28
Add warning when using `list.length` to check for empty list instead of `list.is_empty`

### LINKED ISSUE #2067: warn about performance issues that could arise from using list.length to find out if a list is empty
When using `list.length` to find out if a list is empty the compiler could emit a warning telling the developer to use `list.empty` instead and linking to some part of the documentation that goes into detail about `list` being a linked list and `list.length` iterating over the list to find it's length.

Code that should produce a warning:
```gleam
pub fn non_empty_list(list: List(a)) -> Option(List(a)) {
  // O(n), so potentially super slow.
  case list.length(list) == 0 {
    True -> Some(list)
    False -> None
  }
}

pub fn non_empty_list(list: List(a)) -> Option(List(a)) {
  // O(n), so potentially super slow.
  case list.length(list) {
    0 -> Some(list)
    _ -> None
  }
}
```

What to replace it with:
```gleam
pub fn non_empty_list(list: List(a)) -> Option(List(a)) {
  case list.is_empty(list) {
    True -> Some(list)
    False -> None
  }
}

pub fn non_empty_list(list: List(a)) -> Option(List(a)) {
  case list == [] {
    True -> Some(list)
    False -> None
  }
}
```

For reference see the discord conversation around here: https://discord.com/channels/768594524158427167/768594524158427170/1084786897680879696

maxdeviant:
This PR adds a warning when using `list.length` to check for the empty list instead of using `list.is_empty`.

Currently we're able to detect and issue the warning for the following expressions:

```gleam
list.length(a_list) == 0
list.length(a_list) != 0

0 == list.length(a_list)
0 != list.length(a_list)

list.length(a_list) == -0
list.length(a_list) != -0

-0 == list.length(a_list)
-0 != list.length(a_list)
```

### Motivation

`list.length` runs in linear time (`O(n)`), while `list.is_empty` runs in constant time (`O(1)`). This makes using `list.length` potentially far less efficient for checking if a list if empty for large lists.

Resolves #2067.


compiler-core/src/type_/error.rs
```diff
@@ -333,6 +333,11 @@ pub enum Warning {
     UnnecessaryDoubleBoolNegation {
         location: SrcSpan,
     },
+
+    PerfListLength {
```
maxdeviant, line 337: This is just a placeholder name for the warning while I think of something better.

Suggestions welcome!

compiler-core/src/type_/expression.rs
```diff
@@ -691,6 +691,27 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
                     .map_err(|e| convert_unify_error(e, right.location()))?;
 
+                match (&left, &right) {
+                    (TypedExpr::Call { fun, .. }, TypedExpr::Int { value, .. })
+                    | (TypedExpr::Int { value, .. }, TypedExpr::Call { fun, .. }) => {
+                        match fun.as_ref() {
+                            TypedExpr::ModuleSelect {
+                                module_name, label, ..
+                            } => {
+                                if module_name == "gleam/list" && label == "length" && value == "0"
```
maxdeviant, line 701: A general question I have is whether this is the best way to detect the usage of `list.length`.

It does feel a little weird since the compiler doesn't know about the standard library (AFAIK).

maxdeviant:
While working on this it got me thinking if the compiler core itself is the best place for these kinds of warnings (especially since this one, in particular, could be something that is auto-fixable).

Has there been any discussion about what the equivalent of a `cargo clippy` / `cargo clippy --fix` might be in Gleam?

lpil:
> While working on this it got me thinking if the compiler core itself is the best place for these kinds of warnings (especially since this one, in particular, could be something that is auto-fixable).

I think generally it is not, but seeing as this one is so common I think it may warrant inclusion in the compiler itself.

> Has there been any discussion about what the equivalent of a cargo clippy / cargo clippy --fix might be in Gleam?

I've been working on [a parser](https://github.com/lpil/glance/) for the purpose of making a specialised linter for the [Exercism Gleam track](https://github.com/exercism/gleam/). It'll be specific to Exercism as a teaching tool, but it could be the inspiration for a more general linter project.

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -563,3 +563,113 @@ fn double_unary_bool_variable() {
         "#
     );
 }
+
+#[test]
+fn prefer_list_is_empty_over_list_length() {
```
lpil, line 568: Let's use snapshots for these tests rather than comparing to the warning object. That way we can see what the rendered error looks like and test that part too.

compiler-core/src/type_/expression.rs
```diff
@@ -691,6 +691,30 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
 
+                match (&left, &right) {
+                    (TypedExpr::Call { fun, .. }, TypedExpr::Int { value, .. })
+                    | (TypedExpr::Int { value, .. }, TypedExpr::Call { fun, .. }) => {
+                        match fun.as_ref() {
+                            TypedExpr::ModuleSelect {
+                                module_name, label, ..
+                            } => {
+                                let is_list_length_call =
+                                    module_name == "gleam/list" && label == "length";
```
lpil, line 702: Do you think we should also check that this comes from the `gleam_stdlib` package?

compiler-core/src/type_/expression.rs
```diff
@@ -691,6 +691,30 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 unify(left.type_(), right.type_())
                     .map_err(|e| convert_unify_error(e, right.location()))?;
 
+                match (&left, &right) {
```
lpil, line 694: Let's extract this into a new function, to make it easier to read.

There's also a very large amount of nesting in this function- let's rewrite it in a flat style using early returns.

compiler-core/src/type_/expression.rs
inoas: I think:
- We should also test that it comes form gleam_stdlib. Let's say someone has a todo list and that is streamed or async or something, then calling .length makes sense?
- Once we also test for gleam stdlib we can convert this warning into an error? What would be the drawback?

inoas:
Would moving from `string.length(str) == 0` to `str == ""` have a similar impact?

inoas:
> Would moving from `string.length(str) == 0` to `str == ""` have a similar impact?

https://github.com/gleam-lang/stdlib/blob/v0.29.0/src/gleam_stdlib.mjs#L132

... maybe a simlar implementation should be added for list.length and string.length for all targets?
... or we teach good manners and add this warning/error here for both list and string?

compiler-core/src/type_/tests/warnings.rs
maxdeviant: Good call, that makes this much nicer.

I no longer need to build the compiler and run it against a test Gleam project every time I want to visually inspect the warnings 😄

compiler-core/src/type_/expression.rs
```diff
@@ -741,6 +743,76 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+        };
+
+        // TODO: Figure out how to simulate loading the `gleam_stdlib` package in the tests.
+        const STDLIB_MODULE_NAME: &str = if cfg!(test) {
+            // In the tests we currently declare all modules ourselves in a package
+            // named "thepackage".
+            //
+            // Therefore, we need to use that as the "stdlib" module name when running
+            // the tests if we want the warning to work as expected.s
+            "thepackage"
```
maxdeviant, line 791: This part here is kind of gross, since it means that our tests won't fully line up with reality.

One way we might be able to potentially get around this is reworking the test fixtures to be able to define modules that come from different packages. Not sure how difficult that will be.

@lpil Do you have any ideas on the best way to proceed here?

compiler-core/src/type_/expression.rs
maxdeviant: > Do you think we should also check that this comes from the `gleam_stdlib` package?

I took a pass at this in the latest revision.

> Once we also test for gleam stdlib we can convert this warning into an error? What would be the drawback?

One drawback would be that making this an error would be a breaking change.

I think it would be nice to have this as an auto-fixable thing at some point, at which point it probably doesn't matter too much if it's a warning or an error.

A warning seems like a good place to start.

compiler-core/src/type_/expression.rs
maxdeviant: Done!

compiler-core/src/type_/expression.rs
maxdeviant: > One drawback would be that making this an error would be a breaking change.

I guess this is still technically a breaking change for anyone who:
1. Has usages of `list.length([]) == 0` in their codebase
2. Is using `gleam build --warnings-as-errors` 

I'm new around here, so I don't know what our stance on breaking changes like this is 😄

compiler-core/src/type_/expression.rs
lpil: Let's permit the package name to be specified in the test.

compiler-core/src/type_/expression.rs
```diff
@@ -691,6 +691,8 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 unify(left.type_(), right.type_())
                     .map_err(|e| convert_unify_error(e, right.location()))?;
 
+                self.check_for_inefficient_empty_list_check(name, &left, &right, location);
```
lpil, line 694: I think we can remove this one in favour of the one below?

lpil (commented):
Nice! I like how clear the function is now.

compiler-core/src/type_/tests.rs
```diff
@@ -199,7 +212,15 @@ pub fn stringify_tuple_strs(module: Vec<(&str, &str)>) -> Vec<(SmolStr, String)>
...
 }
 
-pub fn infer_module(src: &str, dep: Vec<(&str, &str)>) -> Vec<(SmolStr, String)> {
+/// A module loaded as a dependency in the tests.
+///
+/// In order the tuple elements indicate:
+/// 1. The package name.
+/// 2. The module name.
+/// 3. The module source.
+type DependencyModule<'a> = (&'a str, &'a str, &'a str);
```
maxdeviant, line 221: With a third tuple element being added I felt the inline type definition was starting to get a bit unclear.

I moved it out into a type alias to document what each element of the tuple represents.

I think a struct would probably be clearer, but it's a bit more complicated to refer to from within the macro (but I can go that way, if we think it's worth it).

compiler-core/src/type_/tests.rs
```diff
@@ -234,15 +255,15 @@ pub fn compile_module(
     // place.
     let _ = modules.insert(PRELUDE_MODULE_NAME.into(), build_prelude(&ids));
 
-    for (name, module_src) in dep {
+    for (package, name, module_src) in dep {
```
maxdeviant, line 258: The dependencies now provide their own package name so that we can use it to simulate loading modules from different packages.

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -437,39 +437,39 @@ fn unused_imported_module_with_alias_warnings_test() {
 #[test]
 fn unused_imported_module_no_warning_on_used_function_test() {
     assert_no_warnings!(
-        ("gleam/foo", "pub fn bar() { 1 }"),
+        ("thepackage", "gleam/foo", "pub fn bar() { 1 }"),
```
maxdeviant, line 440: I had to update these tests to pass in the package name due to changes with the macro.

The package name has not changed from what it was before.

compiler-core/src/type_/expression.rs
maxdeviant: We're now setting the package name to `gleam_stdlib` in the tests, to reflect reality.

compiler-core/src/type_/expression.rs
lpil: Oh yes! Thank you for explaining

compiler-core/src/type_/expression.rs
lpil: I don't think this will ever be an error, a warning is OK.

lpil (approved):
Bravo! Thank you very much.

PR #2181, opened 2023-05-28
Promote comments to doc comments, where applicable

maxdeviant:
While I was working in the compiler I noticed a number of cases where functions had regular comments instead of doc comments above them.

This PR promotes those comments to doc comments so that they show up in the editor/related tooling.


lpil (approved):
Thank you

PR #2186, opened 2023-05-29
Add blockquote styling

giacomocavalieri:
This PR adds styling to block quote elements in Gleam documentation. It just mimics the style of the `.member-name` class adding a vertical accent color border and an accented background.

Here's how it looks with the light theme on:
Screenshot 2023-05-29 alle 08 49 19

And here's with the dark theme:
Screenshot 2023-05-29 alle 08 43 27


lpil (approved):
Wonderful, thank you!

PR #2187, opened 2023-05-29
Fix path dependency paths

### LINKED ISSUE #2182: manifest.toml uses absolute paths for path deps
Follows on from https://github.com/gleam-lang/gleam/pull/2153

These need to be relative.

```toml
packages = [
  { name = "project_b", version = "0.1.0", build_tools = ["gleam"], requirements = ["project_d"], source = "local", path = "/Users/louis/src/gleam/gleam/test/project_path_deps/project_b" },
]
```

NthTensor:
Resolves #2182.

This adds a dependency to the [pathdiff](https://docs.rs/pathdiff) package, a utility for turning absolute paths into relative ones. Normally I don't like to add tiny packages like this in PRs but in this case its a somewhat complex implementation extracted from the rust compiler backed so I feel it's warranted.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -511,7 +511,7 @@ fn get_manifest<Telem: Telemetry>(
     };
 
     if should_resolve {
-        let manifest = resolve_versions(runtime, mode, config, None, telemetry)?;
+        let manifest = resolve_versions(runtime, mode, paths, config, None, telemetry)?;
         return Ok((true, manifest));
     }
 
@@ -524,7 +524,7 @@ fn get_manifest<Telem: Telemetry>(
         Ok((false, manifest))
     } else {
         tracing::debug!("manifest_outdated");
-        let manifest = resolve_versions(runtime, mode, config, Some(&manifest), telemetry)?;
+        let manifest = resolve_versions(runtime, mode, paths, config, Some(&manifest), telemetry)?;
         Ok((true, manifest))
     }
 }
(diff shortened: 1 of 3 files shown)


NthTensor:
Gah! I am not sure why windows is failing.

lpil (commented):
`%s/relitive/relative/g` 😁

lpil:
Ah, very cool! Is that something to be done in this PR?

NthTensor:
Yeah, I can't test it but I am pretty sure this is currently properly broken on windows. I will mark for review when I've got it working!

NthTensor:
Ok, should be fixed on windows. Packages given by absolute paths will still show up as absolute in the manifest, but now relative paths will stay relative.

compiler-cli/src/dependencies.rs
```diff
@@ -612,9 +613,32 @@ impl ProvidedPackageSource {
...
+                ProvidedPackageSource::Git {
+                    repo: own_repo,
+                    commit: own_commit,
+                },
+                ProvidedPackageSource::Git {
+                    repo: other_repo,
+                    commit: other_commit,
+                },
+            ) => own_repo == other_repo && own_commit == other_commit,
+            _ => false,
```
lpil, line 633: Can we list al the cases rather than having a catch all here? We've had multiple bugs from variants being added but matches like these silently returning the wrong value rather than having to be modified for the new variant.

lpil (approved):
Thanks! One small note, but actually I'll do it :)

PR #2188, opened 2023-05-30
Add ability to open docs in the browser

maxdeviant:
This PR adds the ability to have `gleam docs build` automatically open the docs in the browser after they are built.

`gleam docs build` now accepts an optional `--open` flag to indicate that the docs should be opened after generation:

```
λ gleam docs build --help
gleam-docs-build
Render HTML docs locally

USAGE:
    gleam docs build [OPTIONS]

OPTIONS:
    -h, --help    Print help information
        --open    Opens the docs in a browser after rendering
```

Passing the `--open` flag will cause the `index.html` file for the docs to be opened in the system default program for `.html` files (typically the browser):

Screenshot 2023-05-29 at 9 49 09 PM

### DIFF
--- a/compiler-cli/Cargo.toml
+++ b/compiler-cli/Cargo.toml
@@ -78,6 +78,8 @@ lsp-types = "0.92"
 fslock = "0.2.1"
 # Compact and cheap to clone immutable string type
 smol_str = "0.1"
+# Open generated docs in browser
+opener = "0.6"
 
 [dev-dependencies]
 # Test assertion errors with diffs
--- a/compiler-cli/src/docs.rs
+++ b/compiler-cli/src/docs.rs
@@ -5,7 +5,7 @@ use crate::{cli, hex::ApiKeyCommand, http::HttpClient};
 use gleam_core::{
     build::{Codegen, Mode, Options, Package},
     config::{DocsPage, PackageConfig},
-    error::Error,
+    error::{Error, FileIoAction, FileKind},
     hex,
     io::HttpClient as _,
     Result,
@@ -88,7 +88,7 @@ pub fn build(options: BuildOptions) -> Result<()> {
     );
 
     if options.open {
-        open_docs(&index_html);
+        open_docs(&index_html)?;
     }
 
     // We're done!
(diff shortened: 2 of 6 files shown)


compiler-cli/Cargo.toml
```diff
@@ -78,6 +78,8 @@ lsp-types = "0.92"
 fslock = "0.2.1"
 # Compact and cheap to clone immutable string type
 smol_str = "0.1"
+# Open generated docs in browser
+opener = "0.6"
```
maxdeviant, line 82: We're pulling in the [`opener`](https://crates.io/crates/opener) crate to handle opening the browser for us.

This is also what `cargo doc --open` uses when there is no browser set in the workspace config: https://github.com/rust-lang/cargo/blob/c9b89c67723b93151f023ac3a932a471fe9fa6a0/src/cargo/ops/cargo_doc.rs#L66-L71

compiler-cli/src/docs.rs
```diff
@@ -72,16 +78,32 @@ pub fn build() -> Result<()> {
...
     Ok(())
 }
 
+/// Opens the indicated path in the default program configured by the system.
+///
+/// For the docs this will generally be a browser (unless some other program is
+/// configured as the default for `.html` files).
+fn open_docs(path: &Path) {
+    if let Err(error) = opener::open(path) {
+        eprintln!("Failed to open docs:\n{}", error);
```
maxdeviant, line 103: If we failed to open the docs, log the error to stderr, but don't fail the operation overall (since the docs themselves were still built successfully).

lpil: Let's use the same error printing mechanism as the others, for consistency. The file IO errors are probably a good one to look at.

lpil (commented):
Fab idea! Would you mind updating the changelog too?

lpil (commented):
Sorry, I have been unclear. I mean making a new error variant and returning that so it will be printed and the program exits with a non-zero error code.

maxdeviant:
> Sorry, I have been unclear. I mean making a new error variant and returning that so it will be printed and the program exits with a non-zero error code.

Alright, a new variant has been added.

Here's what it looks like when opening the docs fails:

Screenshot 2023-05-30 at 4 18 58 PM

lpil (approved):
Wonderful! Thank you very much 💜

maxdeviant:
> Sorry, would you mind rebasing on main to resolve the merge conflicts? Thank you

No problem! Conflicts have been resolved.

lpil (approved):
bravo!

Issue #2192, opened 2023-05-30
`gleam format` to upgrade conditional compilation syntax


lpil:
Follows on from https://github.com/gleam-lang/gleam/issues/2191

See the linked issue for details on the new syntax.

Issue #2191, opened 2023-05-30
Convert conditional compilation to use attribute syntax


lpil:
Old syntax:

```rust
if erlang {
  pub fn main() { todo }
}

if javascript {
  pub fn main() { todo }
}
```

New syntax

```rust
@if(erlang)
pub fn main() { todo }

@if(javascript)
pub fn main() { todo }
```

We will also want `gleam fix` to update this syntax. https://github.com/gleam-lang/gleam/issues/2192

Prior discussion: https://github.com/gleam-lang/gleam/discussions/2133. Note this does not include the change to the external function syntax.

lpil:
Yes, the attribute should be usable for any definition, the same as the `if` syntax can today be used for any definition.

MarkusPettersson98:
Sorry for doing it a bit backwards this time, but I threw together [a PR](https://github.com/gleam-lang/gleam/pull/2198) which aims to implement this feature :blush: 

However, I expect it to become a bit stalled while waiting for #2192. Anyone keen on working on that? Otherwise, I could give it a shot later this week :calendar:

PR #2198, opened 2023-05-30
Use attribute syntax for conditional compilation

MarkusPettersson98:
This PR aims to implement #2191.

### Compiling a toy program
So far I've gotten the following piece of Gleam compiling: 

```gleam
@if(erlang)
pub fn erl_main() { }

@if(javascript)
pub fn js_main() { }
```

So if I target `erlang`:
```bash
$ gleam build --target erlang
  Compiling playground

warning: Unimplemented function
  ┌─ /home/markus/Projects/gleam/playground/src/playground.gleam:2:1
  │
2 │ pub fn erl_main() {
  │ ^^^^^^^^^^^^^ This code is incomplete
```

the compiler will correctly strip the `javascript` version of `main` out.
If we do the same, but but targetting `javascript`: 

```bash
$ gleam build --target javascript
  Compiling playground

warning: Unimplemented function
  ┌─ /home/markus/Projects/gleam/playground/src/playground.gleam:5:1
  │
5 │ pub fn js_main() {
  │ ^^^^^^^^^^^^^ This code is incomplete
```
the compiler will correctly strip the `erlang` version of `main` out.

Since the guts of `parse_target_group` remain intact, the patch will allow use of the attribute syntax to conditionally compile any definition: imports, external functions, external types etc ..

### What remains
To be able to properly test this PR, #2192 is probably needed first. Until then, CI will fail miserably.
To test my toy program, I had to disregard the entire standard library, which is less than ideal.

Also, the code could do with some cleanup. I am not too keen on the amount of nesting which is starting to pile up in the `parse_target_group` function.

Always fun hacking on this codebase :rocket:


lpil:
Hey @MarkusPettersson98 !! Thanks for this, looks grand. 💜 

The compiler needs to support both syntaxes at once so people can migrate, and also so we can implement `gleam fix`. Could you put-back the code for the previous conditional compilation syntax?

MarkusPettersson98:
I added back the traditional `if`-based syntax. It should be easy to simply rip it out when the time comes to deprecate it :wink: 

Appreciate the always so positive feedback, makes contributing such a blast! :purple_heart:

maurobalbi:
Should
```gleam
@if(erlang)
fn wobble() { ... }

fn main() { ... }
```
be equal to
```gleam
if erlang {
  fn wobble() { ... }

  fn main() { ... }
)
```

or
```gleam
if erlang {
  fn wobble() { ... }
}

fn main() { ... }
```
?

I think the latter feels more intuitive.

lpil:
It is the latter, aye! Each definition is to have its own attribute on it

inoas:
This is great!

inoas:
is it planned to have other constraints in `@if`
If this is imaginable:

`@if(target=erlang)` adds very little boilerplate and means `@if` is more generic

compiler-core/src/parse.rs
```diff
@@ -192,8 +192,25 @@ where
...
+            // Attribute-syntax
+            Some((_, Token::Attribute, _)) => {
+                let _ = self.next_tok();
+                match &self.tok0 {
+                    Some((_, Token::If, _)) => {
+                        let _ = self.next_tok();
+                        let _ = self.expect_one(&Token::LeftParen)?;
+                        let target = self.expect_target()?;
+                        let _ = self.expect_one(&Token::RightParen)?;
+                        let statements = self.expect_module_statements()?;
```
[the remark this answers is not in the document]
MarkusPettersson98: Oh, that's not on you imho, I could have read the requirements more carefully :blush:

MarkusPettersson98:
Wow, nice to see so much engagement around this :smile:
I will fix the scoping of the CC attribute asap, which I suspect will be during the upcoming weekend :blush:

lpil:
A good point @inoas. Let's use this syntax instead:

```rust
@target(erlang)
pub fn main() {}
```

This has the added benefit of not using a keyword here.

MarkusPettersson98:
I worked a bit on this today, but it is still not completely finished.

## What has been done since last update
* Use `@target` attribute for conditional compilation
* Only parse the following statement after a `@target` attribute, rather than all following ones

I still need to add tests, and I am not completely sure if I should've implemented `expect_module_statement` by introducing the new `ParseErrorType`: `ExpectedStatement`. Here I would appreciate some feedback :blush:

MarkusPettersson98:
Since it was decided to not use reserved keywords, I am looking for feedback on how we should treat erroneous attributes in a gleam program.

For instance, consider the following program:

```gleam
import gleam/io

@targetx(erlang)
pub fn hello_erlang() {
  io.println("Hello from erlang!")
}

pub fn main() {
  hello_erlang()
}
```

`@target` is supposed to be used for conditional compilation, not `@targetx`. For this situation, I implemented the current error message + hint:

```bash
gleam run
  Compiling playground
error: Syntax error
  ┌─ /home/markus/Projects/gleam/playground/src/playground.gleam:3:2
  │
3 │ @targetx(erlang)
  │  ^^^^^^^ I was not expecting this.

Expected one of: 
target
Hint: Did you mean to write @target?
```

Currently this is an error in the parser. But should that be the case, if we allow `Name`s instead of keywords, such as `if`, after the `@` sign? Or should this be resolved at a later stage, for example in the type checker?

I noticed that their is this concept of `Diagnostic`, which is used more frequently with type errors and can be used to provide prettier (and more helpful) error messages. I.e, if I try to run the following erroneous program:

```gleam
import gleam/io

pub fn main() {
  io.printlnn("Hello World")
}
```
This helpful error message is printed:
```bash
gleam run
  Compiling playground
error: Unknown module field
  ┌─ /home/markus/Projects/gleam/playground/src/playground.gleam:4:5
  │
4 │   io.printlnn("Hello World")
  │     ^^^^^^^^^ Did you mean `println`?

The module `gleam/io` does not have a `printlnn` field.
```

This would be nice to have for my new error message as well, since we can easily point to what is wrong and what is expected from the user :blush:

compiler-core/src/parse/token.rs
```diff
@@ -11,6 +11,7 @@ pub enum Token {
     Float { value: SmolStr },
     String { value: SmolStr },
     CommentDoc { content: String },
+    Attribute,
```
lpil, line 14: Could you call this `At` and put it in the punctuation section please?

lpil (commented):
Having it in the parser for now is good. We could optionally move it later as we introduce more attributes.

compiler-core/src/parse/token.rs
MarkusPettersson98: Done!

lpil (commented):
Wonderful! Could we get some tests that use the new syntax please? 🙏 

Say one of each of a JavaScript and Erlang code generation test. Nothing complex, just so we can see definitions being included or not depending on what target is selected.

lpil (approved):
Thank you!

PR #2203, opened 2023-05-31
Filter out optional dependencies before version resolution

### LINKED ISSUE #1964: Do not download optional dependencies by default

NthTensor:
Resolves #1964.

Ignores all optional dependencies. Not only are optional dependencies not downloaded, they are not considered during version resolution.

We probably want a way to re-enable specific optional dependencies before we merge this. We could perhaps not filter out optional dependencies if at-least one non-optional dependency exists to the same package, but that seems really insanely complicated to implement. Looking for feedback on other ways we could accomplish this.

If we want to follow the Hex convention, then we should filter out all optional dependencies which are not dependencies of the root.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -691,11 +691,28 @@ fn resolve_versions<Telem: Telemetry>(
     )?;
 
     // Convert the hex packages and local packages into manfiest packages
-    let manifest_packages = runtime.block_on(future::try_join_all(
+    let lookup_result = runtime.block_on(future::try_join_all(
         resolved
             .into_iter()
             .map(|(name, version)| lookup_package(name, version, &provided_packages)),
     ))?;
+
+    // Form a set of activated packages which have at-least one non-optional dependency
+    let activated_packages =
+        lookup_result
+            .iter()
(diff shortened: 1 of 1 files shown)


NthTensor:
> I think this works as desired, right?

I don't believe so. Right now I am ignoring a bunch of constraints. But if we do end up having a non-optional dependency and an optional dependency on the same package, we need to re-activate the optional constraint to get a valid version.

> Is this something we can test?

Absolutely.

NthTensor:
Closing. This solution breaks version resolution.

PR #2205, opened 2023-05-31
Add supported compiler version to config

### LINKED ISSUE #1813: In `gleam.toml` one should be able to specify a minimum gleam version
- Either in absolute terms or as a semver string (like mix.exs deps?)
- Could this make dep resolvement harder?

NthTensor:
Resolves #1813.

This PR adds a new optional entry to the config called `gleam-version` which uses the normal dependency range syntax and describes the supported compiler versions. You can add something like
```
name = "example"
version = "0.1.0"
gleam-version = "~> 0.29"
```
to support compiler versions between the current release and `1.0`.

Whenever a config file is loaded the compiler checks to see if it is compatible with the compiler version and produces a helpful error if it is not. The main config file and all gleam-based dependencies are validated during build.

This option doesn't effect dependency resolution. The value of `gleam-version` should probably not change between minor releases.

### DIFF
--- a/compiler-cli/src/config.rs
+++ b/compiler-cli/src/config.rs
@@ -55,12 +55,14 @@ pub fn find_package_config_for_module(
 
 pub fn read(config_path: PathBuf) -> Result<PackageConfig, Error> {
     let toml = crate::fs::read(&config_path)?;
-    toml::from_str(&toml).map_err(|e| Error::FileIo {
+    let config: PackageConfig = toml::from_str(&toml).map_err(|e| Error::FileIo {
         action: FileIoAction::Parse,
         kind: FileKind::File,
         path: config_path,
         err: Some(e.to_string()),
-    })
+    })?;
+    config.check_gleam_compatability()?;
+    Ok(config)
 }
 
 pub fn ensure_config_exists(paths: &ProjectPaths) -> Result<(), Error> {
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -129,7 +129,6 @@ impl PackageConfig {
             path: path.as_ref().to_path_buf(),
             err: Some(e.to_string()),
         })?;
-        config.check_gleam_compatability()?;
         Ok(config)
     }
 
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -2,6 +2,7 @@ use crate::error::{FileIoAction, FileKind};
 use crate::io::FileSystemReader;
 use crate::manifest::Manifest;
 use crate::requirement::Requirement;
+use crate::version::COMPILER_VERSION;
 use crate::{Error, Result};
 use globset::{Glob, GlobSetBuilder};
 use hexpm::version::Version;
@@ -181,7 +182,7 @@ impl PackageConfig {
     // with the current compiler version
     pub fn check_gleam_compatability(&self) -> Result<(), Error> {
         if let Some(required_version) = &self.gleam_version {
-            let compiler_version = hexpm::version::Version::parse(env!("CARGO_PKG_VERSION"))
+            let compiler_version = hexpm::version::Version::parse(COMPILER_VERSION)
                 .expect("Parse compiler semantic version");
             let range = hexpm::version::Range::new(required_version.to_string())
                 .to_pubgrub()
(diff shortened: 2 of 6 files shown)


compiler-core/src/config.rs
```diff
@@ -173,6 +177,29 @@ impl PackageConfig {
         .expect("internal module globs")
         .is_match(module)
     }
+
+    // Checks to see if the gleam version specified in the config is compatable
+    // with the current compiler version
+    pub fn check_gleam_compatability(&self) -> Result<(), Error> {
+        if let Some(required_version) = &self.gleam_version {
+            let compiler_version = hexpm::version::Version::parse(env!("CARGO_PKG_VERSION"))
```
lpil, line 185: Let's use a constant for the Gleam version. Currently it's in `compiler-cli/src/main.rs`, let's move it to `compiler-core/src/lib.rs`.

compiler-core/src/error.rs
```diff
@@ -223,6 +223,13 @@ pub enum Error {
         package: String,
         build_tools: Vec<String>,
     },
+
+    #[error("The package {package} requires gleam version {required_version} and you are using {gleam_version}")]
```
lpil, line 227:
```suggestion
    #[error("The package {package} requires Gleam v{required_version} and you are using v{gleam_version}")]
```

compiler-core/src/error.rs
```diff
@@ -2455,6 +2462,28 @@ issue in our tracker: https://github.com/gleam-lang/gleam/issues",
...
             }
+
+            Error::IncompatibleCompilerVersion {
+                package,
+                required_version,
+                gleam_version
+            } => {
+                let text = format!(
+                    "The package {} requires gleam version {} \
+but you are using version {}.",
```
lpil, line 2473:
```suggestion
                    "The package {} requires Gleam v{} \
but you are using v{}.",
```

compiler-core/src/error.rs
```diff
@@ -2455,6 +2462,28 @@ issue in our tracker: https://github.com/gleam-lang/gleam/issues",
...
+                    package,
+                    required_version,
+                    gleam_version
+                );
+                Diagnostic {
+                    title: "Incompatible gleam version".into(),
```
lpil, line 2479:
```suggestion
                    title: "Incompatible Gleam version".into(),
```

lpil (commented):
Fab, thank you! I've left some comments inline.

compiler-core/src/error.rs
NthTensor: Required version is going to be a range, so this interpolates as `Gleam v~>1.0`.

NthTensor: Required version is going to be a range, so this interpolates as `Gleam v~>1.0`.

lpil: Oh! Thank you

compiler-cli/src/config.rs
```diff
@@ -55,12 +55,14 @@ pub fn find_package_config_for_module(
...
     let toml = crate::fs::read(&config_path)?;
-    toml::from_str(&toml).map_err(|e| Error::FileIo {
+    let config: PackageConfig = toml::from_str(&toml).map_err(|e| Error::FileIo {
         action: FileIoAction::Parse,
         kind: FileKind::File,
         path: config_path,
         err: Some(e.to_string()),
-    })
+    })?;
+    config.check_gleam_compatability()?;
```
lpil, line 64: I'd like to be able to test that if you run (or perhaps instantiate) the package compiler with a version too low it errors, to ensure this behaviour isn't lost in future refactorings. Could we move it there and have a test? 🙏 

Anything that isn't going to be obviously wrong with normal manual testing it helps to have tests for. We accidentally lost the printing of warnings once 😁

NthTensor: Currently there are two checks: one in the cli and one in the package compiler. The cli check covers stuff for the root package, and gives a nice error when for things like `deps download` that don't build anything. Do you want me to remove this check?

lpil: Ah! That makes more sense, thank you. Keeping and having a test for the core one sounds good.

lpil (approved):
Thank you!

PR #2206, opened 2023-05-31
feat: add unstable flag for Deno config

brettkolodny:
Adds the `unstable` flag to the Deno config so that Gleam projects can opt into new features.


lpil (approved):
Thanks!

PR #2207, opened 2023-05-31
Unlock new dependencies

### LINKED ISSUE #1754: Adding a new dep to gleam.toml that is already in the manifest does not update the version
How to reproduce:

```
> gleam new server
> gleam add gleam_http
> gleam add gleam_elli
> gleam build
Manually add gleam_otp = "0.4.0" line to gleam.toml
```
Real: Got error `gleam_otp is specified with the requirement 0.4.0, but it is locked to 0.5.0, which is incompatible`  
Expected: The gleam_otp version will change to 0.4 in the manifest.

Now to avoid this problem, you need to delete the manifest file, delete the folder.build, and then build the project again.

NthTensor:
Resolves #1754.

Unlocks newly added packages. Previously you would get a version conflict error if you added a top-level dependency to package which was already pinned as a sub-dependency of another package. Now the old version will be unpinned, and dependency resolution will be used to find a new version.

Sorry about the PR spam by the way! Several of these package management tweaks are tiny, and I thought it would be good to knock them out all at once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Adding a new dependency unlocks the target package.
 - Dependencies can now be loaded from paths. Path dependencies currently use
   the same semantics as hex dependencies, and must be updated using the command
   `gleam deps update` to load changes.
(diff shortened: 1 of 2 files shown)


lpil (approved):
Don't apologise for making PRs fixing things 😁 

Would you mind updating the changelog? And either adding a test or adding a TODO comment saying to add a test for this behaviour. I forget if we have code in place to test this :P

NthTensor:
Should be good to go! Not sure why the tests keep failing but if I don't think its related to this PR.

lpil (approved):
Thank you!

PR #2211, opened 2023-06-01
Feature ignore optional dependencies by defult

### LINKED ISSUE #1964: Do not download optional dependencies by default

NthTensor:
Resolves #1964.

My previous attempt at this (#2203) just ignored optional dependency constraints entirely, but this could result in hard-to-diagnose errors arising from version mismatches.

This PR always uses version constraints (optional or not) and adds a new `active` flag to manifest packages, to which the default is `true`. Packages which are only given by optional requirements are listed in the manifest `active = false` and are not downloaded. Once #2207 is merged, you will be able to active an inactive package by adding it to your dependencies (right now the inactive package will still be locked and this will produce a version conflict error).

Unfortunately, creating the set of active packages has to happen at a really specific stage between when the final package releases are downloaded from hex and when the packages are written to the manifest. This makes it pretty hard to write unit tests for.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -693,6 +693,10 @@ fn resolve_versions<Telem: Telemetry>(
         .map(|(name, package)| (name.to_string(), package.to_hex_package(name)))
         .collect();
 
+    // Add the root dependencies to the set of active packages
+    let activated_packages: HashSet<String> = config.all_dependencies()?.into_keys().collect();
+
+    // Do version solve
     let resolved = dependency::resolve_versions(
         PackageFetcher::boxed(runtime.clone()),
         provided_hex_packages,
@@ -712,8 +716,8 @@ fn resolve_versions<Telem: Telemetry>(
     let activated_packages =
         lookup_result
             .iter()
-            .fold(HashSet::new(), |mut acc, (activated, _pkg)| {
-                acc.extend(activated);
+            .fold(activated_packages, |mut acc, (activated, _pkg)| {
+                acc.extend(activated.clone());
                 acc
             });
 
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -333,10 +333,10 @@ impl LocalPackages {
             .packages
             .iter()
             .filter(|p| {
-                p.name != root && {
-                    self.packages.get(&p.name) != Some(&p.version)
-                        || matches!(p.source, ManifestPackageSource::Local { .. })
-                }
+                let is_local = matches!(p.source, ManifestPackageSource::Local { .. });
+                let is_outdated = self.packages.get(&p.name) != Some(&p.version);
+                let should_update = is_local || is_outdated;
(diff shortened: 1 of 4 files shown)


lpil (commented):
I'm not very familiar with this code so I may be about to ask some silly questions 😁 

What would happen if we just dropped all the optional dependencies from the collection of packages prior to downloading and writing the manifest? They would be unlocked, but if a new package was added that required the optional package the unlocked constraints from the existing packages would still be respected.
I think that would permit us to remove the concept of active (which I found a bit hard to understand at first). Mix and rebar3 don't seem to record that state in their manifests, so it seems we can do without it.

I would like to have some tests for this as it's it quite a complicated task and I don't feel confident being able to know if it is working or not in future without them.
I think we'd need to move `lookup_package`'s HTTP side effects into an IO trait object to achieve this, but that would be a good thing in my eyes as I would like to move this code into `-core` in future. `-cli` is intended to be a thin wrapper over core that injects IO, so having the logic in there is the long term goal.

NthTensor:
> I may be about to ask some silly questions

Not at all!

> What would happen if we just dropped all the optional dependencies from the collection of packages prior to downloading and writing the manifest? 

I tried this first, but it interferes with the code to detect stale packages. The current system assumes at several points that if a package is pinned then all of it's dependencies are pinned.

> I would like to have some tests for this as it's it quite a complicated task and I don't feel confident being able to know if it is working or not in future without them. I think we'd need to move lookup_package's HTTP side effects into an IO trait object... would like to move this code into -core in future.

Absolutely agree. Things are getting a little too ad-hoc with all the little incremental changes I have been proposing. I can think of several places we could decouple different stages of the dependency flow, or move logic package management back into the core app. 

Would you be open a separate PR which does some light refactoring/documenting/testing and lays the groundwork for more complex features like this one? After that, I can see about updating this one to get the drop-from-manifest stuff working.

lpil:
Got it, thank you.

> I tried this first, but it interferes with the code to detect stale packages. The current system assumes at several points that if a package is pinned then all of it's dependencies are pinned.

I think this would be the system to modify in that case, to remove that assumption. The end result of the unused dependencies not being locked or present would be best and more in keeping with the other build tools.

lpil:
Gunna close this as stale, but please feel free to reopen. Thank you 💜

Issue #2210, opened 2023-06-01
Emit a warning when a .gleam file has an invalid name


greenfork:
I tried having a file `20230601211333_create_accounts_table.gleam` somewhere in `src/`, it was not compiled to Erlang and LSP didn't see it. After I removed the number prefix from the file name, it became visible. This probably happens because Erlang modules can't start with a number.

I think the compiler should at least emit a warning about such cases. Would be best to include such files in compilation, maybe prefix them with `@` automatically when converting to Erlang modules.

lpil:
I like the warning idea, let's do that.

PR #2213, opened 2023-06-06
Add a warning when invalid files are in the source.

etopiei:
This attempts to address #2210. It seems to work okay after doing some testing, but I am pretty new to Rust (and the Gleam compiler) so I'm not sure if this approach is ideal or not. Happy to hear any and all feedback to make this better 🙂

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -41,7 +41,7 @@ impl ProjectIO {
 }
 
 impl FileSystemReader for ProjectIO {
-    fn gleam_ext_files(&self, dir: &Path) -> Vec<PathBuf> {
+    fn gleam_extension_files(&self, dir: &Path) -> Vec<PathBuf> {
         if !dir.is_dir() {
             return vec![];
         }
@@ -52,12 +52,12 @@ impl FileSystemReader for ProjectIO {
             .filter_map(Result::ok)
             .filter(|e| e.file_type().is_file())
             .map(|d| d.into_path())
-            .filter(move |d| is_gleam_ext(d, dir.clone()))
+            .filter(move |d| is_gleam_extension(d, dir.clone()))
             .collect()
     }
 
     fn gleam_source_files(&self, dir: &Path) -> Vec<PathBuf> {
-        self.gleam_ext_files(dir)
+        self.gleam_extension_files(dir)
             .into_iter()
             .filter(move |d| is_gleam_path(d, dir.clone()))
             .collect()
@@ -330,18 +330,18 @@ pub fn is_gleam_path(path: &Path, dir: impl AsRef<Path>) -> bool {
     )
 }
 
-fn is_gleam_ext(path: &Path, dir: impl AsRef<Path>) -> bool {
+fn is_gleam_extension(path: &Path, dir: impl AsRef<Path>) -> bool {
     use regex::Regex;
     lazy_static! {
         static ref RE: Regex =
-            Regex::new(&format!("^.*\\.gleam$",)).expect("is_gleam_ext() RE regex");
+            Regex::new(&format!("^.*\\.gleam$",)).expect("is_gleam_extension() RE regex");
     }
 
     RE.is_match(
         path.strip_prefix(dir)
-            .expect("is_gleam_ext(): strip_prefix")
+            .expect("is_gleam_extension(): strip_prefix")
             .to_str()
-            .expect("is_gleam_ext(): to_str"),
+            .expect("is_gleam_extension(): to_str"),
     )
 }
 
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -41,7 +41,7 @@ impl ProjectIO {
 }
 
 impl FileSystemReader for ProjectIO {
-    fn gleam_source_files(&self, dir: &Path) -> Vec<PathBuf> {
+    fn gleam_ext_files(&self, dir: &Path) -> Vec<PathBuf> {
         if !dir.is_dir() {
             return vec![];
         }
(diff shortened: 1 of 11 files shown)


compiler-cli/src/fs.rs
```diff
@@ -41,7 +41,7 @@ impl ProjectIO {
 }
 
 impl FileSystemReader for ProjectIO {
-    fn gleam_source_files(&self, dir: &Path) -> Vec<PathBuf> {
+    fn gleam_ext_files(&self, dir: &Path) -> Vec<PathBuf> {
```
lpil, line 44: We avoid abbreviations, so let's use `extension` rather than `ext` in the method names.

compiler-cli/src/fs.rs
```diff
@@ -323,6 +330,21 @@ fn is_gleam_path(path: &Path, dir: impl AsRef<Path>) -> bool {
     )
 }
 
+fn is_gleam_ext(path: &Path, dir: impl AsRef<Path>) -> bool {
```
lpil, line 333: No need to check the path prefix here so we can remove the `dir` argument, and can check the extension directly rather than using a regex. Something like this:

```rust
path.extension() == Some("gleam")
```

compiler-core/src/build/package_loader.rs
```diff
@@ -162,9 +166,18 @@ where
         };
 
         // Src
-        for path in self.io.gleam_source_files(&src) {
-            let input = loader.load(path)?;
-            inputs.insert(input)?;
+        for path in self.io.gleam_ext_files(&src) {
+            if self.io.gleam_source_files(&src).contains(&path) {
```
lpil, line 170: This iterates over the Gleam files on disc count^2 times. We can instead do something like this:

```rust
for path in self.io.gleam_files(&src) {
    if !valid_module_path(path) {
        emit_warning();
        continue;
    }
    // ...
```

compiler-core/src/type_/tests.rs
```diff
@@ -118,8 +118,9 @@ fn get_warnings(src: &str, deps: Vec<DependencyModule<'_>>) -> Vec<Warning> {
     warnings
         .take()
         .into_iter()
-        .map(|warning| match warning {
-            crate::Warning::Type { warning, .. } => warning,
+        .filter_map(|warning| match warning {
+            crate::Warning::Type { warning, .. } => Some(warning),
+            _ => None,
```
lpil, line 123: Could you match on all the variants rather than using a catch all please? That way the compiler will prompt us to update this code when new ones are added. Outdated logic from catch-all patterns has been a common source of bugs in the past.

compiler-core/src/warning.rs
```diff
@@ -123,11 +123,23 @@ pub enum Warning {
...
+    InvalidSource {
+        name: SmolStr,
+    },
 }
 
 impl Warning {
     pub fn to_diagnostic(&self) -> Diagnostic {
         match self {
+            Warning::InvalidSource { name } => Diagnostic {
+                title: "Invalid gleam source filename.".into(),
```
lpil, line 135:
```suggestion
                title: "Invalid module name".into(),
```

compiler-core/src/warning.rs
```diff
@@ -123,11 +123,23 @@ pub enum Warning {
...
+                text: "".into(),
```
lpil, line 136: Let's explain what a valid module name is here. The friendlier and more helpful we can make messages from the compiler the better.

lpil (commented):
Fab! Thank you. I've left some small notes inline

etopiei:
Hi @lpil, thanks for the review 🙏 I've applied the changes you suggested now and rebased to fix conflicts - lmk if there is anything more to fix up 🙂

compiler-core/src/build/package_loader/tests.rs
```diff
@@ -1,13 +1,18 @@
...
     build::module_loader::SourceFingerprint,
     io::{memory::InMemoryFileSystem, FileSystemWriter},
     parse::extra::ModuleExtra,
-    warning::NullWarningEmitterIO,
+    warning::{NullWarningEmitterIO, TypeWarningEmitter},
+    Warning,
+};
+use std::{
+    sync::{atomic::AtomicUsize, Arc},
+    time::Duration,
```
lpil, line 14: I'm confused, why are these extra imports here?

compiler-core/src/io.rs
```diff
@@ -164,6 +164,7 @@ pub trait FileSystemReader {
     fn read_bytes(&self, path: &Path) -> Result<Vec<u8>, Error>;
     fn reader(&self, path: &Path) -> Result<WrappedReader, Error>;
     fn is_file(&self, path: &Path) -> bool;
+    fn is_valid_module_path(&self, path: &Path, dir: &Path) -> bool;
```
lpil, line 167: This doesn't perform IO so it shouldn't be on the IO trait that gets injected in, rather it should be a regular function in a module.

compiler-core/src/type_/tests.rs
```diff
@@ -118,9 +118,10 @@ fn get_warnings(src: &str, deps: Vec<DependencyModule<'_>>) -> Vec<Warning> {
     warnings
         .take()
         .into_iter()
-        .map(|warning| match warning {
-            crate::Warning::Type { warning, .. } => warning,
+        .filter_map(|warning| match warning {
+            crate::Warning::Type { warning, .. } => Some(warning),
             crate::Warning::Parse { .. } => panic!("Unexpected parse warning"),
+            crate::Warning::InvalidSource { .. } => None,
```
lpil, line 124: Panic for this one too please

compiler-core/src/warning.rs
```diff
@@ -128,12 +128,14 @@ pub enum Warning {
...
         warning: crate::type_::Warning,
     },
-
     Parse {
         path: PathBuf,
         src: SmolStr,
         warning: crate::parse::Warning,
     },
+    InvalidSource {
+        name: SmolStr,
```
lpil, line 137:
```suggestion
        path: PathBuf,
```

compiler-core/src/warning.rs
```diff
@@ -186,6 +188,15 @@ impl Warning {
                 }
             },
 
+            Warning::InvalidSource { name } => Diagnostic {
+                title: "Invalid module name. Module names must begin with a lowercase letter and contain only lowercase alphanumeric characters or underscores.".into(),
```
lpil, line 192: This should be text rather than the title. The title is short and more detail goes in the text field, wrapped at around 65 characters to make it read nicely.

lpil (commented):
Fab, thank you! I've left some notes inline

lpil (approved):
Thank you

PR #2230, opened 2023-06-13
New external type syntax

lpil:


inoas:
https://github.com/gleam-lang/website/blob/main/book-src/tour/external-types.md needs to be updated. Does it go away entirely?

PR #2235, opened 2023-06-20
Add gap at end of doc page

giacomocavalieri:
Right now the doc pages have no bottom margin, this can make the page feel a bit cramped and the Lucy message, when toggled, can cover the README text:
Screenshot 2023-06-20 alle 11 03 11
Screenshot 2023-06-20 alle 11 03 14

In this PR I added a gap at the end of the page to make it breath a little:
Screenshot 2023-06-20 alle 11 27 57
Screenshot 2023-06-20 alle 11 28 01

### DIFF
--- a/compiler-core/templates/docs-css/index.css
+++ b/compiler-core/templates/docs-css/index.css
@@ -255,6 +255,7 @@ p code {
 
 .content {
   margin-left: var(--sidebar-width);
+  margin-bottom: var(--large-gap);
   padding: calc(var(--header-height) + var(--gap)) var(--gap) 0 var(--gap);
   width: calc(100% - var(--sidebar-width));
   max-width: var(--content-width);
@@ -469,6 +470,10 @@ body.drawer-open .label-closed {
 
 .module-members {
   margin-top: var(--large-gap);
+}
+
+.module-members:last-of-type .member:last-of-type {
+  margin-bottom: 0;
 }
 
 .module-member-kind {


compiler-core/templates/docs-css/index.css
```diff
@@ -471,6 +472,10 @@ body.drawer-open .label-closed {
   margin-top: var(--large-gap);
 }
 
+.module-members:last-of-type .member:last-of-type {
```
lpil, line 475: This seems like quite an unwieldy selector. Is there a single class or element we can reference? Perhaps adding padding to the bottom of the page?

lpil (commented):
Oh fab! Thank you

compiler-core/templates/docs-css/index.css
giacomocavalieri: Yeah it's bad... at first I simply added the bottom margin to `.content` but that would result in a quite big bottom margin in the code section: `.member` already has some margin so this rule is removing the already existing margin for just the last bit of code of the page.

What would you do in this case? There's no single class to reference since I just want to remove the margin of the last member of the last module of a page

lpil: `.content *:last-child` perhaps? It expresses the intention better too I think as it's relating to the margin added to `.content`.

giacomocavalieri: The problem is that `.content > *:last-child` would do nothing on the margin of the code page: it selects the last `.module-member` in the code page but it's not the responsible for the bottom margin (the element adding the margin is the `.member` _inside_ the `.module-member`). Otherwise the any page with code would still have a big chin:
Screenshot 2023-06-23 alle 12 05 47

I mean, it's not the worst thing we could also just accept it removing it is not worth the extra complexity in the `index.css`'s rules

lpil: Let's go with it as it is then :)

lpil (approved):

Issue #2028, opened 2023-02-26
Permit a message to be specified with `panic`


inoas:
Would that be possible?

## 0.26 gleam/stdlib

```gleam
[1, 2, 3]
|> list.all(fn(x) {
  case x {
    1 -> True
    2 -> False
    // Crash if no short-circuit
    _ -> {
      assert True = False
    }
  }
})
```

## 0.27 gleam/stdlib

```gleam
[1, 2, 3]
|> list.any(fn(x) {
  case x {
    1 -> False
    2 -> True
    // Crash if no short-circuit
    _ -> panic
  }
})
```

## imaginary panic

```gleam
[1, 2, 3]
|> list.any(fn(x) {
  case x {
    1 -> False
    2 -> True
    _ -> panic("Crash if no short-circuit")
  }
})
```

```gleam
[1, 2, 3]
|> list.any(fn(x) {
  case x {
    1 -> False
    2 -> True
    _ -> panic(#("Crash if no short-circuit", x))
  }
})
```

* * *

Edit this is what we see : 
```erlang
gleam/list_test.any_test
#{function => <<"any_test">>,gleam_error => panic,line => 474,
  message => <<"panic expression evaluated">>,
```

```javascript
gleam/list_test.any_test:474: Error: panic expression evaluated
```

The panic data could replace the `message` `"panic expression evaluated"`? Maybe the panic data/message should/could only be literals, function calls should not be allowed at that point IMHO?

Ref: https://github.com/gleam-lang/gleam/issues/1961

inoas:
Ref: https://github.com/gleam-lang/gleam/issues/1961

lpil:
Totally! We want a custom error message for sure. I've held back on this for now because I would like `assert`, `assert let`, `todo` and `panic` to all share the same syntax for their message, and we've not found something that works for all of them yet.

If you've got any ideas for them let me know!

Lunarmagpie:
One way to do this that would work is copying python's assert error message syntax. I don't love this but finding what doesn't work is a good step to finding what does work lol.
```gleam
let assert Ok(value) = f(), "error message"
todo, "error message"
panic, "error message"
```

lpil:
Could also use `as`

```rust
let assert Ok(value) = f() as "error message"
todo as "error message"
panic as "error message"
```
I'm not sold with any of these syntaxes just yet, good to explore more

greenfork:
If the message is optional, it could be

```rust
[...]

PR #2248, opened 2023-06-29
Panic message

lpil:
Closes https://github.com/gleam-lang/gleam/issues/2028


inoas:
I like that we _do not_ have something that looks like a built-in function call for `panic` and `todo`. Thank you for this!

lpil:
Totally! Should avoid some confusion

PR #2263, opened 2023-07-08
feat: allow `gleam new` in existing directories

### LINKED ISSUE #1879: Support creating a new Gleam project in the current directory with `gleam new .`

msutkowski:
Fixes #1879. Per the feedback from @lpil in #1881, this also allows running in an existing folder as long as the generated files do not already exist. It'll produce an output like the screenshot below.

image

This will allow you to:
- `gleam new --name something .` (in an existing folder)
- `gleam new --name existing existing` (will add in generated files in addition to whatever exists when not empty)

I opted for the `enum` to prevent typos. I saw that this same treatment could be applied to the compiler... perhaps it worth adopting across both workspaces?

Anddd, we could also consider a `--force` flag here to overwrite the existing files, but will wait on feedback before considering.

Thanks and let me know if you have any feedback! I'd be happy to address it 🎉

### DIFF
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -4,8 +4,8 @@ use gleam_core::{
     parse, Result,
 };
 use serde::{Deserialize, Serialize};
-use std::fs::File;
 use std::path::{Path, PathBuf};
+use std::{collections::HashMap, fs::File};
 use std::{env, io::Write};
 use strum::{Display, EnumString, EnumVariantNames};
 
@@ -36,6 +36,16 @@ pub struct Creator {
     gleam_version: &'static str,
     options: NewOptions,
     project_name: String,
+    files_to_create: HashMap<FileToCreate, String>,
+}
+
+#[derive(PartialEq, Eq, Debug, Hash)]
+enum FileToCreate {
+    Readme,
+    Gitignore,
+    ProjectGleam,
+    TestGleam,
+    ConfigToml,
 }
 
 impl Creator {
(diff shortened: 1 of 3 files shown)


compiler-cli/src/new.rs
```diff
@@ -48,8 +58,22 @@ impl Creator {
         .trim()
         .to_string();
 
+        let files_to_create = HashMap::from([
```
lpil, line 61: If we want to use types to model the files to create let's use this pattern here:

https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_/prelude.rs#L180

We could then implement a `pub fn location(&self) -> PathBuf` which would be the replacement for the hashmap.

The use of EnumIter means that we are guaranteed to cover each of the variants in the enum and we don't have to have the runtime checks to ensure the entries are in the hashmap.

compiler-core/src/error.rs
```diff
@@ -149,6 +149,12 @@ pub enum Error {
     #[error("project root already exists")]
     ProjectRootAlreadyExist { path: String },
 
+    #[error("File(s) already exist in {path}: {}", file_names.join(", "))]
+    OutputFilesAlreadyExist {
+        path: String,
+        file_names: Vec<String>,
```
lpil, line 155:
```suggestion
        path: PathBuf,
        file_names: Vec<PathBuf>,
```

lpil (commented):
Really nice! Thank you!

I like your use of an enum. I've made a suggestion inline about how we could take it further.

compiler-cli/src/new.rs
msutkowski: I gave it a shot. Let me know if this is the right track :)

compiler-cli/src/new.rs
```diff
@@ -197,7 +222,8 @@ jobs:
 
     fn gleam_toml(&self) -> Result<()> {
         write(
-            self.root.join("gleam.toml"),
+            self.root
+                .join(FileToCreate::ConfigToml.location(&self.project_name)),
```
lpil, line 226: If we're to go down the enum path these methods need to all be removed and replaced by iterating over the enum. The code would look something like this:

```rust
for file in FileToCreate::iter() {
    let path = directory_name.join(file.location(project));
    let contents = file.contents(project)
    write(path, file)?;
}
```

msutkowski: Ah - I intentionally left those as-is, but can go ahead and move the content up and iterate here as well 🎉

lpil:
Closing this due to inactivity. Please feel free to re-open. Thanks!

dxixp:
Hi @msutkowski! I'm reaching out to see if you have plans to work on this issue. If not I would gladly take it over.

msutkowski:
> Hi @msutkowski! I'm reaching out to see if you have plans to work on this PR and https://github.com/gleam-lang/gleam/issues/1879. If not I would gladly take it over.

Feel free! I won't have time for a bit, and I'd appreciate this getting in 🎉

Issue #1421, opened 2021-12-29
Field and recursive access in guard tests


J3RN:
_Probably just enqueuing more work for myself, but..._

For case guard expressions, the parser appears to [limit you to using variables and integer indexed values of tuple variables](https://github.com/gleam-lang/gleam/blob/7e05ca47bf1d09f1b5b21ae24d2cad7f93127153/compiler-core/src/parse.rs#L1022-L1048), so your options are pretty much limited to:
```gleam
case thing {
  a if a.1 == 5 -> True
  b if b < 4 -> False
  // etc
}
```

This is well and good, but it's an arbitrary constraint. [Erlang supports record syntax in guards](https://github.com/gleam-lang/gleam/blob/7e05ca47bf1d09f1b5b21ae24d2cad7f93127153/compiler-core/src/parse.rs#L1022-L1048), so there's no reason why we can't support named field access:
```gleam
case cat {
  c if c.name != "Nubi" -> True
  // ...
}
```

Though, this could be done via pattern-match instead:
```gleam
case cat {
  Cat(name: name) if name != "Nubi" -> True
  //...
}
```
But the same could be said for integer-based indexing of tuples, so :shrug: 

Also, recursive access could be supported, if desired:
```gleam
case thing {
  a if a.1.foobar != 5 -> True
  b if a.1.2.3.4 == 6 -> True
  // ...
}
```

lpil:
We definitely want this! Thank you

J3RN:
As a potential extension of this in the general notion of "things I expected to work in guard tests but didn't": arithmetic operators. For instance (contrived, again, of course):
```gleam
case x {
  y if y * 5 > 20 -> True
  // ...
}
```

Yields this error message:
```
error: Syntax error
   ┌─ src/testapp.gleam:13:12
   │
13 │     y if y * 5 > 20 -> True
   │            ^ I was not expecting this.

Expected one of: "->"
```

lpil:
Generally anything that works in Erlang guards and Gleam has a literal expression syntax I'd be happy to add to Gleam's guard

bcpeinhardt:
I'd like to try my hand at this. I think to start I'll just add support for named field syntax and go from there.

PR #2272, opened 2023-07-14
Allow field access in clause guard units

massivefermion:
I did all that I could figure out, but there are errors that I have no idea how to resolve!

### DIFF
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
@@ -1,0 +1,24 @@
+---
+source: compiler-core/src/erlang/tests/case.rs
+assertion_line: 23
+expression: "\npub fn myfun(mt) {\n  case mt {\n    1 | _ ->\n      1\n      |> Ok\n  }\n  1\n  |> Ok\n}\n"
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars]).
+
+-export([main/0]).
+-export_type([person/0]).
+
+-type person() :: {person, binary(), binary(), integer()}.
+
+-spec main() -> binary().
+main() ->
+    Given_name = <<"jill"/utf8>>,
+    Person = {person, <<"credo"/utf8>>, <<"jack"/utf8>>, 32},
+    case Given_name of
+        Name when Name =:= erlang:element(2, Person) ->
+            <<"it's jack"/utf8>>;
+
+        Name@1 ->
+            <<"it's "/utf8, Name@1/binary>>
+    end.
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
@@ -17,7 +17,7 @@ main() ->
     Raiden = {person, <<"raiden"/utf8>>, <<"jack"/utf8>>, 31},
     case Given_name of
         Name when Name =:= erlang:element(3, Raiden) ->
-            <<"it's jack"/utf8>>;
+            <<"It's jack"/utf8>>;
 
         _ ->
             <<"It's not jack"/utf8>>
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
@@ -1,7 +1,7 @@
 ---
-source: compiler-core/src/erlang/tests/case.rs
-assertion_line: 23
-expression: "\npub fn myfun(mt) {\n  case mt {\n    1 | _ ->\n      1\n      |> Ok\n  }\n  1\n  |> Ok\n}\n"
+source: compiler-core/src/erlang/tests/guards.rs
+assertion_line: 446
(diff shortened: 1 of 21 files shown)


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
```diff
@@ -0,0 +1,24 @@
...
+-export_type([person/0]).
+
+-type person() :: {person, binary(), binary(), integer()}.
+
+-spec main() -> binary().
+main() ->
+    Given_name = <<"jill"/utf8>>,
+    Person = {person, <<"credo"/utf8>>, <<"jack"/utf8>>, 32},
+    case Given_name of
+        Name when Name =:= erlang:element(2, Person) ->
```
lpil, line 19: The first element in the tuple is the the tag, so this would need to get the 3rd element rather than the 2nd (Erlang tuples are 1 indexed)

compiler-core/src/parse.rs
```diff
@@ -1289,6 +1289,23 @@ where
                             }
                         }
 
+                        Some((_, Token::Name { name: label }, int_e)) => {
```
lpil, line 1292: This doesn't support `thing.one.two`, which I think is a requirement. We'll need to parse more than a name

lpil (commented):
Looking really good! I've left a couple comments inline

compiler-core/src/type_/expression.rs
```diff
@@ -1046,6 +1046,44 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                    typ if typ.is_unbound() => Err(Error::NotATupleUnbound {
+                        location: container.location(),
+                    }),
+
+                    _ => Err(Error::NotATuple {
+                        location,
+                        given: container.type_(),
+                    }),
+                }
```
lpil, line 1084: I think this section shouldn't be talking about tuples?

compiler-core/src/type_/expression.rs
```diff
@@ -1046,6 +1046,57 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                                    Ok(ClauseGuard::FieldAccess {
+                                        location,
+                                        label,
+                                        index: Some(index + 1),
+                                        type_: typ,
+                                        container: Box::new(typed_container),
+                                    })
+                                }
+
+                                _ => panic!("Expected RecordAccess"),
```
lpil, line 1077: I think this will crash the compiler when it's a module being accessed.

compiler-core/src/type_/expression.rs
```diff
@@ -1046,6 +1046,57 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            ClauseGuard::FieldAccess {
+                location,
+                label,
+                container,
+                ..
+            } => {
+                let typed_container = self.infer_clause_guard(*container.clone())?;
+
+                match typed_container.type_().as_ref() {
+                    Type::App { .. } => match typed_container.clone() {
```
lpil, line 1058: It doesn't look like we do anything with the knowledge that this is a `Type::App` here, so this flow control can be removed.

lpil (commented):
Getting there! Thank you

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__guards__field_access.snap
massivefermion: I think I forgot to update the snapshot, it seems to be fixed now!

compiler-core/src/type_/expression.rs
```diff
@@ -1046,6 +1046,57 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                    typ if typ.is_unbound() => Err(Error::NotATupleUnbound {
+                        location: typed_container.location(),
+                    }),
+
+                    _ => Err(Error::NotATuple {
+                        location,
+                        given: typed_container.type_(),
+                    }),
+                }
```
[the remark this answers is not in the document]
massivefermion: removing the `match` with the `Type::App` clause mostly solved it. But there is still the `match` to see whether the container is itself a field access or not which should have a catch all clause to satisfy exhaustiveness but otherwise is not needed.

compiler-core/src/type_/expression.rs
massivefermion: I tested it, it doesn't crash because the out of scope error for the module happens before this.

compiler-core/src/type_/expression.rs
```diff
@@ -1046,6 +1046,39 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                            Ok(ClauseGuard::FieldAccess {
+                                location,
+                                label,
+                                index: Some(index + 1),
+                                type_: typ,
+                                container: Box::new(var),
+                            })
+                        }
+
+                        _ => panic!("Expected RecordAccess"),
```
lpil, line 1074: We shouldn't make assertions about the variant returned here, rather any shared code should be agnostic to which AST is being used. We can largely adapt `infer_known_record_access` to be agnostic and have a small wrapper for each kind of AST.

lpil (commented):
Could we get tests for nested access too? `one.two.three` 🙏 

And for invalid cases such as `"hello".field`

lpil:
Hey @massivefermion ! What are your plans here? Thanks

massivefermion:
Truly sorry about this. Every day I try to work on this, but I'm still in the middle of something that doesn't let me focus(And I really need to focus for this).
Maybe I can revert my changes to the place that at least one level field access works and just merge that? Or if that's not enough and you can't keep it open anymore, we can just close it until I can come back to it or someone else just does it!
Again, really sorry, I really wanted to do this.

lpil:
No worries at all!! If you're happy for me to continue on your behalf I can finish this one. You've done the lion's share of the work already :)

massivefermion:
No problem, thanks for the work!

lpil (approved):

lpil:
Closes https://github.com/gleam-lang/gleam/issues/1421

Issue #2259, opened 2023-07-03
Poor formatting of external function


lpil:
Current formatting
```gleam
@external(erlang, "ffi", "improper_list_append")
fn improper_list_append(a: item_a, b: item_b, c: improper_tail) -> List(
  anything,
)
```
It should instead be this:
```gleam
@external(erlang, "ffi", "improper_list_append")
fn improper_list_append(
  a: item_a,
  b: item_b,
  c: improper_tail,
) -> List(anything)
```

PR #2273, opened 2023-07-15
Poor formatting of external function

cobbinma:
#2259 

👋 

I think the line in the annotation meant the formatter was ignoring the head group.
Fixed this by adding a group for the signature before appending it to the attributes.


lpil (approved):
Brilliant! Thank you!!

PR #2277, opened 2023-07-17
use UTF8 paths from camino

### LINKED ISSUE #2079: Convert from Path to Utf8Path etc
https://crates.io/crates/camino

bcpeinhardt:
closes #2079 

This pull request is a straight replacement of std::Path and std::PathBuf with their camino equivalents Utf8Path and Utf8PathBuf in the compiler-core, compiler-cli, compiler-wasm, and test-package-compiler crates. It's a lot of files, but there's absolutely nothing fancy.
When we get a Path or Pathbuf from a third party crate, we convert to Utf8 equivalent and expect with a nice error message.

### DIFF
--- a/compiler-cli/src/build_lock.rs
+++ b/compiler-cli/src/build_lock.rs
@@ -36,7 +36,7 @@ impl BuildLock {
         crate::fs::mkdir(&self.directory).expect("Could not create lock directory");
 
         let lock_path = self.directory.join("gleam.lock");
-        let mut file = fslock::LockFile::open(&lock_path.to_string()).expect("LockFile creation");
+        let mut file = fslock::LockFile::open(lock_path.as_str()).expect("LockFile creation");
 
         if !file.try_lock_with_pid().expect("Trying build locking") {
             telemetry.waiting_for_build_directory_lock();
--- a/compiler-cli/src/build_lock.rs
+++ b/compiler-cli/src/build_lock.rs
@@ -1,14 +1,14 @@
+use camino::Utf8PathBuf;
 use gleam_core::{
     build::{Mode, Target, Telemetry},
     paths::ProjectPaths,
     Result,
 };
-use std::path::PathBuf;
 use strum::IntoEnumIterator;
 
 #[derive(Debug)]
 pub(crate) struct BuildLock {
-    directory: PathBuf,
+    directory: Utf8PathBuf,
 }
 
 impl BuildLock {
@@ -36,7 +36,7 @@ impl BuildLock {
         crate::fs::mkdir(&self.directory).expect("Could not create lock directory");
 
         let lock_path = self.directory.join("gleam.lock");
-        let mut file = fslock::LockFile::open(&lock_path).expect("LockFile creation");
+        let mut file = fslock::LockFile::open(&lock_path.to_string()).expect("LockFile creation");
 
         if !file.try_lock_with_pid().expect("Trying build locking") {
             telemetry.waiting_for_build_directory_lock();
--- a/compiler-cli/src/compile_package.rs
+++ b/compiler-cli/src/compile_package.rs
@@ -11,7 +11,7 @@ use gleam_core::{
     type_::ModuleInterface,
     uid::UniqueIdGenerator,
     warning::WarningEmitter,
-    Result
+    Result,
 };
 use smol_str::SmolStr;
 use std::sync::Arc;
--- a/compiler-cli/src/compile_package.rs
+++ b/compiler-cli/src/compile_package.rs
@@ -3,7 +3,7 @@ use crate::{
     fs::{self, ConsoleWarningEmitter, ProjectIO},
     CompilePackage,
 };
-use camino::{Utf8Path, Utf8PathBuf};
+use camino::Utf8Path;
 use gleam_core::{
     build::{Mode, PackageCompiler, StaleTracker, Target, TargetCodegenConfiguration},
     metadata,
@@ -11,7 +11,7 @@ use gleam_core::{
     type_::ModuleInterface,
     uid::UniqueIdGenerator,
     warning::WarningEmitter,
-    Result,
+    Result
 };
 use smol_str::SmolStr;
 use std::sync::Arc;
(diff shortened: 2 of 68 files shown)


bcpeinhardt:
Hmmmm. The error the ci is getting is a warning for an unused import, but the import actually is used. Removing it is an error. Hmmmmmmmm it's a thinker.

compiler-cli/src/build_lock.rs
```diff
@@ -36,7 +36,7 @@ impl BuildLock {
         crate::fs::mkdir(&self.directory).expect("Could not create lock directory");
 
         let lock_path = self.directory.join("gleam.lock");
-        let mut file = fslock::LockFile::open(&lock_path).expect("LockFile creation");
+        let mut file = fslock::LockFile::open(&lock_path.to_string()).expect("LockFile creation");
```
lpil, line 39: We should be able to use this as a reference to an OsStr rather than allocate a string and then discard it immediately. Perhaps `lock_path.as_ref()`?

compiler-cli/src/compile_package.rs
```diff
@@ -56,12 +57,13 @@ pub fn command(options: CompilePackage) -> Result<()> {
...
 fn load_libraries(
     ids: &UniqueIdGenerator,
-    lib: &Path,
+    lib: &Utf8Path,
 ) -> Result<im::HashMap<SmolStr, ModuleInterface>> {
     tracing::info!("Reading precompiled module metadata files");
     let mut manifests = im::HashMap::new();
     for lib in fs::read_dir(lib)?.filter_map(Result::ok) {
-        let path = lib.path().join(paths::ARTEFACT_DIRECTORY_NAME);
+        let path = Utf8PathBuf::from_path_buf(lib.path().join(paths::ARTEFACT_DIRECTORY_NAME))
```
lpil, line 65: fs::read_dir should return utf8 paths, there should be no `PathBuf` usage within our codebase if this library has been adopted.

compiler-cli/src/main.rs
```diff
@@ -520,7 +520,10 @@ fn initialise_logger() {
 }
 
 fn project_paths_at_current_directory() -> ProjectPaths {
-    let current_dir = std::env::current_dir().expect("Could not get current directory");
+    let current_dir = Utf8PathBuf::from_path_buf(
+        std::env::current_dir().expect("Could not get current directory"),
+    )
+    .expect("Non Utf-8 Path");
```
lpil, line 526: This code exists in a few places, let's move it to a helper function in `fs`.

compiler-cli/src/new/tests.rs
```diff
@@ -25,7 +27,8 @@ fn new() {
     assert!(path.join("test/my_project_test.gleam").exists());
     assert!(path.join(".github/workflows/test.yml").exists());
 
-    let toml = crate::fs::read(path.join("gleam.toml")).unwrap();
+    let toml =
+        crate::fs::read(Utf8PathBuf::from_path_buf(path.join("gleam.toml")).unwrap()).unwrap();
```
lpil, line 31: Let's move this converting to utf8 path to the place where the path is created rather than here where it is used. All converting to our desired data structures should happen at the boundary of our system as much as possible.

lpil (commented):
Brilliant, thank you! I've left a few small comments.

Let's also add rules to clippy.toml to ban use of Path, PathBuf, and the side effecting methods from the new path classes.

compiler-cli/src/main.rs
bcpeinhardt: We use this in compiler-core as well. Are you okay if I put the helper in compiler-core::io?

lpil: Oh yikes, it shouldn't be in there- it performs IO. We'll have to move it to the IO trait, or at least put a huge TODO: FIXME on it for now

bcpeinhardt: I think this might involve a more serious think haha. I think a big TODO fix me (with an accompanying issue) would be a good idea. Compiler core has a concept of DirEntry, ReadDir, OutputFile, FileIOError, ProjectPaths, etc. all based on paths with real implementations instead of just traits. There could be a clever way to extract all this stuff to be based on a simple enough trait for compiler-cli and others to implement if you really want compiler-core's conception of a file and a path to be abstract.

lpil: > all based on paths with real implementations instead of just traits

What do you mean by this?

bcpeinhardt: OH I meant converting from path to utf8 path in general (which is in the language server and paths. Language server needs to convert from lsp crates `Url` type, paths needs to convert from `dirs_next::cache_dir` call).

lpil: Oh right! Anything pure is fine into do in core 👍 

Sounds good

bcpeinhardt: Done :)

bcpeinhardt:
I've updated the clippy.toml in compiler-core and added one for compiler-cli (just to handle `Path::new` and `PathBuf::new`).
What else do we need to get this merged?

compiler-cli/src/fs.rs
```diff
@@ -17,15 +17,23 @@ use std::{
...
 
 #[cfg(test)]
 mod tests;
 
+/// This is a helper function for converting from `PathBuf` to `Utf8PathBuf`.
+/// We simply panic on non-utf8 paths.
+///
+pub fn utf8_or_panic(input: std::path::PathBuf) -> Utf8PathBuf {
+    Utf8PathBuf::from_path_buf(input).expect("Non Utf8 Path")
+}
```
lpil, line 35: Sorry, I've been unclear. It was IO that should be moved into `compiler-cli`. This is a pure operation so there's no need to move it out of core, and I don't want to have a helper function that does nothing but call `expect`. We want to discourage `expect` so we don't have helpers that do it.

bcpeinhardt:
How's that look?

lpil (approved):
Heroic effort! Thank you so much

lpil:
Would you mind rebasing on main? Alternatively I could do a squash commit if you're good with that.

bcpeinhardt:
Squash it up, I honestly don't think the history is important for this 😂

PR #2290, opened 2023-07-31
Added check to  contructor to auto convert ArrayBuffer to Uint8Array

### LINKED ISSUE #2200: Help with encoding and decoding BitStrings in Node/Deno
@lpil you already have the context but I'll put it here anyway. I am writing a shim for basic file operations with erlang/node/deno. I am trying to provide some functions for working with BitStrings, but getting a strange error. Here is the test
```
pub fn bits_test() {
  let filepath = "./test/hello_bits.txt"
  let assert Ok(_) = <<"Hello, World":utf8>> |> write_bits(to: filepath)
  let assert Ok(_) = <<"Goodbye, Mars":utf8>> |> append_bits(to: filepath)
  let assert Ok(hello_goodbye) = read_bits(from: filepath)
  hello_goodbye |> should.equal(<<"Hello, WorldGoodbye, Mars":utf8>>)
  let assert Ok(_) = delete(filepath)
  let assert Error(_) = read_bits(from: filepath)
}
```
and here is the error
```
simplifile_test.bits_test: Error: 
        <<72, 101, 108, 108, 111, 44, 32, 87, 111, 114, 108, 100, 71, 111, 111, 100, 98, 121, 101, 44, 32, 77, 97, 114, 115>>
        should equal 
        <<72, 101, 108, 108, 111, 44, 32, 87, 111, 114, 108, 100, 71, 111, 111, 100, 98, 121, 101, 44, 32, 77, 97, 114, 115>>
```
On the surface these bit strings seem to match, but I'm not super familiar with how bitstrings work. Here is the ffi code:
```
export function readBits(filepath) {
    try {
        const contents = fs.readFileSync(path.normalize(filepath))
        return new Ok(new BitString(contents))
    } catch(e) {
        return new GError(undefined)
    }
}

export function writeBits(contents, filepath) {
    try {
        fs.writeFileSync(path.normalize(filepath), contents.buffer)
        return new Ok(undefined)
    } catch (e) {
        return new GError(undefined)
    }
}

export function appendBits(contents, filepath) {
    try {
        fs.appendFileSync(path.normalize(filepath), contents.buffer)
        return new Ok(undefined)
    } catch (e) {
        return new GError(undefined)
    }
}
```

BitString here is imported from "./gleam.mjs".

bcpeinhardt:
Closes #2200

When constructing a `BItString` from the gleam.mjs in js, if you pass an `ArrayBuffer` to the constructor, it will now default to being wrapped in a `Uint8Array`.


compiler-core/templates/prelude.js
```diff
@@ -116,6 +116,9 @@ export class BitString {
   }
 
   constructor(buffer) {
+    if (buffer instanceof ArrayBuffer) {
+      buffer = new Uint8Array(buffer);
+    }
```
lpil, line 121: Nice!

Looking at the code I don't understand why it is that it doesn't work with an array buffer. Did you work out why? Perhaps it should always be an array buffer.

How about we throw an exception if it's not the desired type? Someone could make the same mistake with one of the other buffer classes.

bcpeinhardt: Yeah I may have been a bit hasty to make a pull request, there may be a better way here or this may not even really require a fix. 

My understanding is that a javascript `ArrayBuffer` is an "array of bytes" ([docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)), i.e. it can never be a true `BitString`. 

In the context of the issue, I was confused because I was trying to write a `BitString`'s contents to a file, and then read the content's back out, and they were failing an equality check. I think this was because my `BitString` literal used the `:utf8` option, so I think the constructed `BitString` must have been wrapped in a `Uint8Array` view, which is why the same data failed an equality check.

My solution was to cast the contents read out with the same `Uint8Array` data view. When working on this, since the buffer will never not be bytes, it made sense to just set a default for the 90% use case. If you need a different view you can always `const uint16Array = new Uint16Array(uint8Array.buffer)`. 

But now I'm thinking it may be more consistent to throw an exception on anything that isn't an `ArrayBuffer` as you suggest. I think I'll try it an see if some errors pop up regarding the construction of bitstrings with utf8 encodings, and try and get a deeper understanding of the issue.

lpil: That sounds fab! Thank you.

lpil: Are you still working on this @bcpeinhardt ? :)

bcpeinhardt: Yes :)
Sorry it's been kind of a crazy month for me, I can't believe it's been 20 days since I was looking at this 😂
I'm going to try to finish up the issues I've been working on this weekend 😌

bcpeinhardt:
The error I was getting was very simply that I didn't realize I needed to wrap the ArrayBuffer in a Uint8Array before constructing the BitString I was returning.
Everywhere you construct a BitString, you do so with a Uint8Array, so I have simply add a check to throw an exception if the argument passed to the constructor is not a Uint8Array. I think that's a fine requirement to hold.

compiler-core/templates/prelude.js
```diff
@@ -116,6 +116,9 @@ export class BitString {
   }
 
   constructor(buffer) {
+    if (!(buffer instanceof Uint8Array)) {
```
lpil, line 119: Shouldn't this be `ArrayBuffer`? As per https://github.com/gleam-lang/gleam/pull/2290/files#r1283405089?

As that's the underlying storage and Uint8Array is a wrapper.

bcpeinhardt: That was the original plan, but everywhere you construct one (for instance in the prelude tests), you pass in a `Uint8Array`. I have to imagine folks writing ffi code right now are doing the same.
I think either is a reasonable choice, but this seemed like it would break fewer uses. Going to ask in the discord and see if anyone other than us is currently manually constructing `BitString`s in js

lpil: I suspect very few people are constructing bit strings directly, and now is the time to make a breaking change.

Having said that, couldn't we do something like `thing.buffer` for any of the typed array classes?

bcpeinhardt: Alright, I think I've come full circle on this, and it's served as a lesson to try a few different things before considering making a pull request haha. 
Lets use just `ArrayBuffer` and throw an exception if it's anything else. 
`ArrayBuffer` is the actual chunk of memory, the simplest piece, so I think it's what should underpin the internals (as you stated before). 
Encouraging an API that takes any Typed Array like a `Uint8Array` or a `DataView` and implicitly pulling out the buffer is probably a bad idea, because the view might be only of segment of the underlying buffer.
I'm going to update the `BitString` class to be based around an `ArrayBuffer` and push for your review.

lpil: Ah right! That makes much more sense, thank you.

>  because the view might be only of segment of the underlying buffer.

This is the key bit here- we need to be able to support having bit strings being a slice in to the underlying buffer, to avoid having lots of copying when iterating over a bit string or slicing new bit strings from them. I foolishly didn't understand the relationship between these classes and slices.

With that constraint in mind what's the best one to use? If we used `ArrayBuffer` we would need to implement the logic ourselves for only using a portion of it, which could be slower than an optimised implementation that comes from one of the other classes.

bcpeinhardt: Right now I have an ArrayBuffer backed impl. For the various operations it uses Uint8Array internally to work with the buffer, but the actual stored buffer is an ArrayBuffer. 
I edited unequalBuffer so the isEqual function works for any ArrayBuffer backed item, and now the only issue is that inlining a BitString into another is failing so I need to fix that.

bcpeinhardt: I honest to god think we could merge this as is. It's not like we ran into some actual performance constraint, I just didn't understand I needed to use a Uint8Array, but now the constructor makes that much more explicit. I think limiting to Uint8Array makes perfect sense, as it's the closest thing JavaScript has to a BitString, meets your requirements of cheaply creating multiple BitStrings from the same memory, and seems to be how the class was designed in the first place. I say we take the small win, be grateful for having the chance to learn more about JavaScript together, and focus energy on more important tasks lol.

lpil (approved):
Thank you so much! Especially for talking this all through with me.

PR #2292, opened 2023-08-01
properly inline integer constants in bitstring patterns

### LINKED ISSUE #2166: Constants are failing when used with bit_strings
Giving this code:
```gleam
pub fn get_next(index_file: IndexFile) -> Result(Index, file.Reason) {
  case fs.read(index_file.handler, index_size_bytes) {
    read.Ok(<<
      offset:size(offset_size_bits),
      size:size(event_size_bits),
      file_id:size(file_id_size_bits),
    >>) -> Ok(Index(offset, size, file_id))
    read.Eof -> Error(file.Espipe)
    _ -> Error(file.Einval)
  }
}
```
and defined the following constants:
```gleam
const offset_size_bits = 48
const event_size_bits = 24
const file_id_size_bits = 128
```
Gleam compiler isn't giving any issue but the Erlang compiler is failing and generating this code:
```erlang
get_next(Index_file) ->
    case file:read(erlang:element(2, Index_file), 25) of
        {ok,
            <<Offset:Offset_size_bits,
                Size:Event_size_bits,
                File_id:File_id_size_bits>>} ->
            {ok, {index, Offset, Size, File_id}};

        eof ->
            {error, espipe};

        _ ->
            {error, einval}
    end.
```
As you can see, `Offset_size_bits`, `Event_size_bits`, `File_id_size_bits` are the constants but they are not defined in the scope of the Erlang function.

macroby:
closes #2166 

To allow inlining of integer constants in bit string patterns, specifically in the bit string size option, I added an optional constructor field to Pattern::VarUsage. When the Pattern is typed, then the constructor field will hold Some(constructor), otherwise it is None. The constructor is used during codegen to get the const integer value to inline.

I still need to write tests for the new behavior as well as show that variables used in an option in a bitstring pattern still work as intended.

Gleam code that I wrote to replicate the issue:
```
pub fn main() {
  reveal_bug()
}

pub const sixteen = 16

fn reveal_bug() {
  let x = 16
  case arbitrary() {
    <<5:size(sixteen)>> -> <<5:size(sixteen)>>
    <<6:size(x)>> -> <<5:size(x)>>
  }
}

fn arbitrary() -> BitString {
  <<5:size(sixteen)>>
}
```

and corrected erlang output showing that "Sixteen" has been replaced by "16":
```
-spec arbitrary() -> bitstring().
arbitrary() ->
    <<5:(lists:max([(16), 0]))>>.

-spec reveal_bug() -> bitstring().
reveal_bug() ->
    X = 16,
    case arbitrary() of
        <<5:16>> ->
            <<5:(lists:max([(16), 0]))>>;

        <<6:X>> ->
            <<5:(lists:max([(X), 0]))>>
    end.

-spec main() -> bitstring().
main() ->
    reveal_bug().
```

### DIFF
--- a/compiler-core/src/erlang/pattern.rs
+++ b/compiler-core/src/erlang/pattern.rs
@@ -31,7 +31,18 @@ fn print<'a>(
 
         Pattern::Discard { .. } => "_".to_doc(),
 
-        Pattern::VarUsage { name, .. } => env.local_var_name(name),
+        Pattern::VarUsage { name, constructor, .. } => {
+            let v = &constructor.as_ref().unwrap().variant;
+            match v {
+                ValueConstructorVariant::ModuleConstant{literal, ..} => {
+                    match literal {
+                        Constant::Int { value, .. } => int(&value),
+                        _ => env.local_var_name(name),
+                    }
+                },
+                _ => env.local_var_name(name)
+            }
+        }
 
         Pattern::Var { name, .. } if define_variables => {
             vars.push(name);
--- a/compiler-core/src/erlang/pattern.rs
+++ b/compiler-core/src/erlang/pattern.rs
@@ -34,7 +34,12 @@ fn print<'a>(
         Pattern::VarUsage {
             name, constructor, ..
         } => {
-            let v = &constructor.as_ref().unwrap().variant;
+            let v = &constructor
+                .as_ref()
+                .expect(
+                    "Expected some constructor since this is a variable but instead none was found",
+                )
+                .variant;
             match v {
                 ValueConstructorVariant::ModuleConstant { literal, .. } => {
                     const_inline(literal, env)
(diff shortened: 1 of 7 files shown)


compiler-core/src/erlang/pattern.rs
```diff
@@ -31,7 +31,18 @@ fn print<'a>(
 
         Pattern::Discard { .. } => "_".to_doc(),
 
-        Pattern::VarUsage { name, .. } => env.local_var_name(name),
+        Pattern::VarUsage { name, constructor, .. } => {
+            let v = &constructor.as_ref().unwrap().variant;
+            match v {
+                ValueConstructorVariant::ModuleConstant{literal, ..} => {
```
lpil, line 37: Rather than handling just ints we'll need to inline any constant of any type, so we'll need to call the `const_inline` function here.

lpil (commented):
I had a little peek even though the PR is still a draft. Looking great so far! Thanks

lpil (approved):
Very nice! Thank you!

Would you mind updating the changelog and also running `cargo fmt`?

lpil:
Looks good! The linter needs a fix but other than that I think we're good to go.

lpil (approved):
Thank you!!

PR #2293, opened 2023-08-05
2286 manifest paths should be relative

bcpeinhardt:
I think this is working how I want it to, just need to do some cleanup and test writing. Creating the draft so you know it's being worked on.


compiler-cli/src/fs.rs
```diff
@@ -27,11 +27,48 @@ use crate::{dependencies::UseManifest, lsp::LspLocker};
...
+    }
+}
+
+impl MakeRelative for MakeRelativeToCurrentDir {
+    /// Given an absolute path, this method will return a relative path
+    /// from the current directory to the provided path.
+    /// A relative path will be returned as is.
+    fn make_relative(&self, path: &Utf8Path) -> Utf8PathBuf {
+        match path.is_absolute() {
+            true => pathdiff::diff_utf8_paths(path, &self.curr_dir)
```
lpil, line 64: This looks to be a pure function, we can use it directly in core rather than injecting it as if it were IO.

compiler-core/src/manifest.rs
```diff
@@ -333,3 +212,171 @@ where
...
+aaa = { version = "> 0.0.0" }
+awsome_local1 = { path = "../path/to/package" }
+awsome_local2 = { git = "https://github.com/gleam-lang/gleam.git" }
+gleam_stdlib = { version = "~> 0.17" }
+gleeunit = { version = "~> 0.1" }
+zzz = { version = "> 0.0.0" }
+"#
+        );
+
+        // The deserialized path no longer matches because it contains a relative path for the
```
lpil, line 360: What does this mean?

compiler-core/src/manifest.rs
```diff
@@ -333,3 +212,171 @@ where
...
+        // path dependency awesome_local1
+
+        // let deserialised: Manifest = toml::from_str(&buffer).unwrap();
+        // manifest.packages.sort_by(|a, b| a.name.cmp(&b.name));
+        // assert_eq!(deserialised, manifest);
```
lpil, line 365: Some commented out code here to remove

compiler-cli/src/fs.rs
```diff
@@ -27,11 +27,48 @@ use crate::{dependencies::UseManifest, lsp::LspLocker};
 #[cfg(test)]
 mod tests;
 
+/// Return the current directory as a UTF-8 Path
+/// # Panics
+/// This function will panic if it fails to determine the current directory,
+/// or if your current directory is not a UTF8 Path.
```
lpil, line 33: This function doesn't panic?

lpil (commented):

compiler-cli/src/fs.rs
bcpeinhardt: It did when I wrote the comment and I forgot to update it xD

bcpeinhardt: Changing this to return a `FileIoError`. What would the `FileIoAction` be for failing to locate the current directory? Open?

compiler-core/src/manifest.rs
bcpeinhardt: For this part of the test, we were testing that when you run `manifest.to_toml`, and the deserialize that back to a `Manifest` struct, you get the same result. That is no longer the expected behavior, because `manifest.to_toml` will now possibly convert an absolute path to a relative path.
I don't love it, but I think it's okay. It would be different if this was the serialize/deserialize impl or something. I left it commented out so you would see it, but if you're okay with it I'll remove the commented code, if not we'll find another way to do it.

compiler-core/src/io.rs
```diff
@@ -9,6 +9,21 @@ use tar::{Archive, Entry};
 
 use camino::{Utf8Path, Utf8PathBuf};
 
+/// Takes in a source path and a target path and determines a relative path
+/// from source -> target.
+/// If given a relative target path, no calculation occurs.
+/// # Panics
+/// The provided source path should be absolute, otherwise will panic.
+pub fn make_relative(source_path: &Utf8Path, target_path: &Utf8Path) -> Utf8PathBuf {
+    assert!(source_path.is_absolute());
```
lpil, line 18: Why do we panic in this case?

compiler-core/src/manifest.rs
```diff
@@ -20,7 +21,7 @@ impl Manifest {
     // manually so that we can control the formatting.
     // We want to keep entries on a single line each so that they are more
     // resistant to merge conflicts and are easier to fix when it does happen.
-    pub fn to_toml(&self) -> String {
+    pub fn to_toml(&self, current_directory: &Utf8Path) -> String {
```
lpil, line 24:
```suggestion
    pub fn to_toml(&self, root_directory: &Utf8Path) -> String {
```

compiler-cli/src/dependencies.rs
```diff
@@ -277,7 +277,7 @@ fn read_manifest_from_disc(paths: &ProjectPaths) -> Result<Manifest> {
 
 fn write_manifest_to_disc(paths: &ProjectPaths, manifest: &Manifest) -> Result<()> {
     let path = paths.manifest();
-    fs::write(&path, &manifest.to_toml())
+    fs::write(&path, &manifest.to_toml(&get_current_directory()?))
```
lpil, line 280: We can't use the current directory as there's no guarantee that the root of the project will be in the current directory.

The `ProjectPaths` is what to use here as it is positioned at the root of the project.

compiler-core/src/io.rs
bcpeinhardt: The source path passed in has to be absolute, otherwise the function doesn't work :) So I'd rather panic than continue with an incorrect path.

compiler-core/src/error.rs
```diff
@@ -796,6 +800,18 @@ Second: {}",
                 }
             }
 
+            Error::NonUtf8Path { path: _ } => {
+                let text =
+                    "Encountered a non utf8 path. Only valid UTF-8 paths are supported".to_owned();
```
lpil, line 805:
```suggestion
                    "Encountered a non UTF-8 path, but UTF-8 paths are supported".to_owned();
```

lpil (approved):
Thank you so much!!

Issue #2243, opened 2023-06-27
Assign names to sub patterns in string pattern matching


giacomocavalieri:
When pattern matching on strings using `<>` it seems it's not possible to assign names to sub patterns:

```gleam
pub fn main() {
  case "foo" {
    "1" as digit <> rest -> todo
  }
}
```

Results in the error:

```text
error: Syntax error
  ┌─ ./src/main.gleam:5:9
  │
5 │     "1" as digit <> rest -> todo
  │                  ^ I was not expecting this.

Expected one of: "->"
```

lpil:
Let's add this!

lpil:
Fab!

You'll want to add a new optional name field to this AST variant: https://github.com/gleam-lang/gleam/blob/23d6800bf1792317cc428e22597912137071c786/compiler-core/src/ast.rs#L1138

Then you'll need to change the parser to accept this: https://github.com/gleam-lang/gleam/blob/23d6800bf1792317cc428e22597912137071c786/compiler-core/src/parse.rs#L992

And lastly a variable needs to be inserted during analysis which has that string prefix as a constant https://github.com/gleam-lang/gleam/blob/23d6800bf1792317cc428e22597912137071c786/compiler-core/src/type_/pattern.rs#L253

giacomocavalieri:
Thank you so much for the useful info! As soon as I have some free time I'll start working on it 💜

OndrejValenta:
I was just wondering what is that code trying to achive? 
i don't mean technically but from programmer's perspective, when would I use it? how?

lpil:
That code example would be very commonly found in a parser where you want to have flow control based on the contents of a string and also save that content so it can be used to produce a token stream, abstract syntax tree, or similar.

giacomocavalieri:
Yes, that was exactly my use case! I had something like this:
```gleam
case string {
  "1" <> rest -> parse_number("1", rest)
  "2" <> rest -> parse_number("2", rest)
  // the same for digits 3 through 9...
}
```

With this It could become
```gleam
case string {
  "1" as digit <> rest
  | "2" as digit <> rest
  // | the same with digits 3 through 9
  -> parse_number(digit, rest) 
}
```

timjs:
Should it be possible to shorten  your example to:

```gleam
case string {
  ("1" | "2" | "3" ) as digit <> rest // And all other digits till 9
  -> parse_number(digit, rest) 
}
```

giacomocavalieri:
[...]

PR #2295, opened 2023-08-06
String pattern alias

### LINKED ISSUE #2294: ✨ Allow assigning names to string prefix patterns.
Gleam allows you to assign names to subpatterns for usage in the expression body using `as`. This is useful for many reasons but one reason is to deduplicate work in combination with alternative clause patterns.

We could write something like:
```gleam
case list {
  [1 as x, ..rest] | [2 as x, ..rest] -> 
    // do something with `x` here

  ... ->
    ...
}
```

Unfortunately, the equivalent for string patterns is a syntax error:
```gleam
case string {
  "a" as x <> rest | "b" as x <> rest ->
    // do something with `x` here

  ... ->
    ...
}
```

Instead, you'd have to write something like:
```gleam
case string {
  "a" <> rest | "b" <> rest -> {
    let assert Ok(#(x, rest)) = string.pop_grapheme(string)
    // do something with `x`
  }

  ... ->
    ...
}
```

Which both duplicates work we've already done and makes the code less immediately clear what it does.

giacomocavalieri:
This should close #2294 and #2243

### DIFF
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -274,6 +274,11 @@ pub enum ValueConstructorVariant {
         literal: Constant<Arc<Type>, SmolStr>,
     },
 
+    /// A constant defined locally, for example when pattern matching on string literals
+    LocalConstant {
+        literal: Constant<Arc<Type>, SmolStr>,
+    },
+
     /// A function belonging to the module
     ModuleFn {
         name: SmolStr,
@@ -332,6 +337,12 @@ impl ValueConstructorVariant {
                 documentation: documentation.clone(),
             },
 
+            Self::LocalConstant { literal } => ModuleValueConstructor::Constant {
+                literal: literal.clone(),
+                location: literal.location(),
+                documentation: None,
+            },
+
             Self::LocalVariable { location, .. } => ModuleValueConstructor::Fn {
                 name: function_name.clone(),
                 module: module_name.clone(),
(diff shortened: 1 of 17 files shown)


lpil:
Hi @giacomocavalieri ! Would you like me to pick this work?

giacomocavalieri:
@lpil sorry if it took long, I think the PR is ready for review now 👍

compiler-core/src/type_.rs
```diff
@@ -676,6 +693,7 @@ impl ValueConstructor {
     pub(crate) fn get_documentation(&self) -> Option<&str> {
         match &self.variant {
             ValueConstructorVariant::LocalVariable { .. } => Some("A locally defined variable."),
+            ValueConstructorVariant::LocalConstant { .. } => Some("A locally defined constant."),
```
lpil, line 696:
```suggestion
            ValueConstructorVariant::LocalConstant { .. } => Some("A locally defined variable."),
```

This is an implementation detail of the compiler, it's still a regular variable from the user's point of view.

compiler-core/src/type_/pattern.rs
```diff
@@ -75,6 +75,47 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
...
+                        let initial_typ = initial.type_.clone();
+                        unify(initial_typ, literal.type_())
+                    }
+
+                    // This variable was not defined in the Initial multi-pattern
+                    _ => Err(UnifyError::ExtraVarInAlternativePattern { name: name.into() }),
+                }
+            }
+        }
+    }
```
lpil, line 117: We only ever insert strings so the vast majority of the code head is dead code and cannot be tested in any way. We can instead have a `insert_constant_string` function that does much less, and we can know that it works.

lpil (commented):
Fab! Two tiny little things inline.

lpil (approved):

Issue #2241, opened 2023-06-25
LSP: Display function signature when hovering over function name at definition site


palexanderm:
Consider the following as an example:
```gleam
fn first_if_true(test, first, second) {
  case test(first) {
    True -> first
    False -> second
  } 
}
```
Hovering over `first_if_true` doesn't display the function's signature.
The function signature is only displayed when hovering over the function's name at the call site. For example:
```gleam
fn main() {
  first_if_true(fn(a) -> { a > 0 }, 1, -1)
}
```

lpil:
Good idea, thank you!

preveen-stack:
new to gleam; came across a video on youtube interviewing @lpil on gleam. And LSP means?

septemhill:
> new to gleam; came across a video on youtube interviewing @lpil on gleam. And LSP means?

@preveen-stack LSP -> Language Server Protocol

inoas:
This is a great idea!

palexanderm:
I'm not sure how easy this would be to implement, but I came across this cool feature of OCaml's where the function signature is inlined as virtual text next to the function head.
![image](https://github.com/gleam-lang/gleam/assets/61638460/74074ac3-3db8-4a31-ac80-ef9a1b6bfe58)

lpil:
We already have that feature :)

palexanderm:
I should've done a better job at highlighting the feature I was referring to in the screenshot. I'm referring to the the virtual text that's on line 43 where the inferred function signature is displayed to the left.

lpil:
Yes, we have that. You can see it in my screenshot to the right of the list of completion suggestions

palexanderm:
Is there any way for that information to be available inline as virtual text like in the screenshot I posted?

lpil:
No, we don't want to add features that would encourage the programmer to omit annotation for module function arguments and return types.

PR #2302, opened 2023-08-19
Display signature when hovering over function definitions or module constants

nino:
Hi!

I had a go at #2241.

Hovering over constants now displays their type:
CleanShot 2023-08-19 at 19 28 41@2x

And hovering over function definitions displays their signature:
CleanShot 2023-08-19 at 19 31 10@2x

Happy to make any changes you need to this PR :)

Thanks!


nino:
On closer inspection I noticed hovering over function arguments still just shows the whole function signature. (See picture.)

CleanShot 2023-08-19 at 21 32 20@2x

I'll try to handle hovering on function arguments too.

compiler-core/src/language_server/engine.rs
```diff
@@ -476,10 +478,29 @@ fn hover_for_pattern(pattern: &TypedPattern, line_numbers: LineNumbers) -> Hover
...
+            // Show the type of the hovered argument to the user
+            for arg in &fun.arguments {
+                if arg.location.contains(byte_index) {
+                    let contents = Printer::new().pretty_print(&arg.type_, 0);
+                    return Some(Hover {
+                        contents: HoverContents::Scalar(MarkedString::String(contents)),
+                        range: Some(src_span_to_lsp_range(arg.location, &line_numbers)),
+                    });
+                }
+            }
```
lpil, line 501: This finding logic should live with the other finding logic in `node_at_position`. This function is only about showing the information rather than hunting through the AST.

lpil (commented):
Fab! Thank you. I've left a note inline, and we'll need tests for the new functionality.

compiler-core/src/language_server/engine.rs
nino: I think that would involve extending the `Located` enum, right? I can have a go at that in the next few days (and also try to write some tests 🤞)

compiler-core/src/language_server/engine.rs
```diff
@@ -241,13 +245,33 @@ where
...
+                        if let Some(arg) = fun
+                            .arguments
+                            .iter()
+                            .find(|arg| arg.location.contains(byte_index))
+                        {
+                            Some(hover_for_function_argument(arg, lines))
+                        } else {
+                            Some(hover_for_function_head(fun, lines))
+                        }
+                    }
```
lpil, line 270: This finding logic should go in the `node_at_position` functions with the other finding logic that deals with byte indexes and AST. This function here doesn't perform any finding logic, it just maps from a found node to a hover response function.

Sorry, I was unclear previously!

nino: That sounds like the best way to do this, but I wanted to check if my understanding/approach sounds reasonable before I try to do stuff (also this is my first open-source contribution of non-trivial size, so this is a bit scary).

Does this sound right:

`node_at_position` should be capable of finding the function head and each function arg. The main type `node_at_position` returns is `Located`, and I don't think there's a way to represent a function argument in a `Located`. So I think I'd add an entry to `Located` called `Arg(&'a TypedArg)`. Then in [`TypedDefinition::find_node`](https://github.com/nino/gleam/blob/1149636aa3112abb2f7701506396e588abee024a/compiler-core/src/ast.rs#L548-L560), I'd try to return a `Located::Arg` if possible, and a `Located::ModuleStatement` if not?

(And also I'll look at how to write tests for the new behavior.)

nino:
I think this might be ready for another look. I've tidied things up a bit and added some tests.

The main issues now are:

1. The `hover` function still does a bit of checking whether we're inside the function head or body:

```rust
Located::ModuleStatement(Definition::Function(fun)) => {
    if fun.location.contains(byte_index) {
        Some(hover_for_function_head(fun, lines))
    } else {
        // Don't show hover info for function body
        None
    }
}
```

Which could be fixed, but I'm not sure if that would break any assumptions we're making about `find_node` finding a `Function` when looking at a position inside the function body.

2. There are two places where I'm defining a temporary variable `let empty_smolstr = SmolStr::from("");`, because I couldn't figure out how to do `unwrap_or_default` with `SmolStr`s. Maybe there's a better solution for that.

compiler-core/src/language_server/engine.rs
lpil: Yup! That's perfect

lpil: Why is there still some location logic in this function? What is the case in which it is needed?

compiler-core/src/language_server/engine.rs
```diff
@@ -472,6 +490,52 @@ fn hover_for_pattern(pattern: &TypedPattern, line_numbers: LineNumbers) -> Hover
     }
 }
 
+fn hover_for_function_head(
+    fun: &Function<Arc<Type>, TypedExpr>,
+    line_numbers: LineNumbers,
+) -> Hover {
+    let empty_smolstr = SmolStr::from("");
+    let documentation = fun.documentation.as_ref().unwrap_or(&empty_smolstr);
```
lpil, line 498: Does `.unwrap_or_default()` not work here? Or `.unwrap_or("".into())`

lpil (commented):
This is looking really nice! I've left a couple tiny questions inline.

Would you mind also updating the CHANGELOG file? Thank you

compiler-core/src/language_server/engine.rs
nino: `.unwrap_or_default()` was the first thing I'd tried since you were using it for all the results of `.get_documentation()`. That didn't work because `.documentation` itself is a `SmolStr`, and `&SmolStr` doesn't have `Default` implemented.

Trying some more stuff, the only other way I can see is

```rust
    let documentation = fun
        .documentation
        .as_ref()
        .map(|s| s.as_str())
        .unwrap_or_default();
```

which feels "cleaner" in some way, but also less readable?

compiler-core/src/language_server/engine.rs
nino: The remaining logic is to prevent the function signature coming up when you hover leading/trailing whitespace or comments in the function body. It's essentially undoing the [`.full_location()` logic in `ast.rs`](https://github.com/nino/gleam/blob/051c6b04191681ff98f4db9786c3fd4bf6d4c685/compiler-core/src/ast.rs#L561-L567).

lpil (approved):
Fantastic!! Thank you so much

lpil:
Would you like to rebase the branch to remove the merge commits and conflicts, or should I squash merge?

nino:
Probably easiest if you just squash merge? But I can also do a rebase if that's the preferred workflow

lpil:
Works for me! Thanks

PR #2307, opened 2023-08-29
Use `import type` for importing type declarations

vonagam:
Right now when I open a generated typescript declaration there are errors about imports of types, like those:
```ts
import * as _ from "../gleam.d.ts";
```
```
A declaration file cannot be imported without 'import type'. 
Did you mean to import an implementation file '../gleam.js' instead?
```

This one changes printing of import statements - if an import path ends with `.d.ts` add the type modifier.

There is a downside through - `import type` was [introduced at the start of 2020](https://devblogs.microsoft.com/typescript/announcing-typescript-3-8-beta/#type-only-imports-exports) with 3.8 release. 3.5 years have passed but I don't know what minimum version of typescript is supposed to be supported by gleam.

### DIFF
--- a/compiler-core/src/javascript/import.rs
+++ b/compiler-core/src/javascript/import.rs
@@ -4,7 +4,7 @@ use itertools::Itertools;
 
 use crate::{
     docvec,
-    javascript::INDENT,
+    javascript::{JavaScriptCodegenTarget, INDENT},
     pretty::{break_, concat, line, Document, Documentable},
 };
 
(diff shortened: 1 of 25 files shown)


vonagam:
Fixed formatting (otherwise all tests were green).

compiler-core/src/javascript/import.rs
```diff
@@ -92,9 +92,16 @@ impl<'a> Import<'a> {
 
     pub fn into_doc(self) -> Document<'a> {
         let path = Document::String(self.path.clone());
+        let import_modifier = if self.path.ends_with(".d.ts") {
```
lpil, line 95: Could we pass this in as an argument or other form of configuration rather than checking the path please.

lpil:
This is a fab change! Thank you. I've left 1 note inline

compiler-core/src/javascript/import.rs
vonagam: Initially I thought about adding an argument but I don't know Rust or this code base so went with a change of least surface level. I'll look into converting it into an argument.

vonagam: Added boolean `is_type` property to `Import`. Can rename if there are better suggestions.

vonagam:
Anything else holding this back?

compiler-core/src/javascript/import.rs
```diff
@@ -28,14 +28,15 @@ impl<'a> Imports<'a> {
 
     pub fn register_module(
         &mut self,
-        js_path: String,
+        path: String,
+        is_type: bool,
```
lpil, line 32: I think `is_type` is a little confusing in this context. How about this?

```rust
target: JavaScriptCodegenTarget,
```
```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq)]
pub enum JavaScriptCodegenTarget {
    JavaScript,
    TypeScriptDeclarations,
}
```

vonagam: I've changed the thing from being property of an import to an argument for `to_doc` since there should be no type imports in js and no non-type imports in ts, so it is not a property of an import instance. Also less code changes.

As for `enum` - from my perspective it makes sense to introduce the enum if you expect or already know other places where it can be useful too. And instead of `TypeScriptDeclarations` I prefer shorter `TypeScript`.

Do you still want enum after current change? If yes, can you add it (into this PR or after merging)? I feel like you will implement it more properly and consistently with code base than me.

lpil: An enum with `TypeScriptDeclarations` please, as it's specifically that format and not regular TypeScript.

vonagam: I've updated with the enum. Is that correct usage/placement?

lpil (approved):

PR #2308, opened 2023-08-29
Use `.d.mts` extension for ts type declarations

vonagam:
Gleam generates `.mjs` files but instead of `.d.mts` it generates `.d.ts`. This mismatch prevents typescript from finding corresponding type declarations. Ability to find those declarations depend on t sconfig and resolver - some things did work for me with current `.d.ts` extension, others were producing errors, with `.d.mts` all bases are covered.

This PR is based on previous one, will rebase once it is merged or rejected.


lpil:
Thank you.

I'm not familiar with .mts files. Is there a place I can read more? Or how would I replicate the issue?

vonagam:
Ok, I've updated changelog for PR about `import type`. When it is merged I'll rebase and update changelog here.

vonagam:
I've rebased this PR and dollar one. Let me know if there is anything else to be done here.

lpil (approved):
Thank you so much!!

PR #2309, opened 2023-08-29
Use opaque ts types for types without constructors

vonagam:
I was surprised to find that all type info is lost in typescript when it comes to maps from std lib. Currently the map type is defined like this in `.d.ts`:
```typescript
export type Map$<MA, LZ> = any;
```
To retain type info opaque types can be used. With this PR all types without constructors become opaque types instead of simple `any`. The map type becomes:
```typescript
export type Map$<MA, LZ> = { [_.OpaqueType]: Map$<LZ, MA> };
```
Where `_.OpaqueType` is a virtual non-existent unique symbol. So now type info about values and keys is preserved and can being used by ts to check arguments and type returns. 

(Based on two previous currently open PRs, will rebase after they are merged/rejected.)


lpil:
Unfortunately this results in an incorrect type specification.

For example, if I define a type like this:

```gleam
pub type MyType
```

it would generate this TypeScript:

```
export type MyType$ = { [_.OpaqueType]: MyType$ };
```

It says that a `MyType` is an object with this symbol key- but it may not be an object and it never has that symbol key.

It also says the type has infinite size at runtime as it's recursive with no case to terminate the recursion.

vonagam:
> It says that a MyType is an object

Not really correct. `type X = {}; const x: X = true` - is a valid code. `{}` does not mean that the thing is an object as even other primitive types have methods like `toString` in js.

> with this symbol key - and it never has that symbol key.

Yeah, that's the point. The opaque type. There is no symbol. The only way to create an opaque type is through code base that defines it. 

> It also says the type has infinite size at runtime as it's recursive with no case to terminate the recursion.

Do you get such error from some diagnostic report? Because such thing should be supported by ts without much problem, js is not immutable and this is the same as having a reference on an object of itself - `const a = {}; a.self = a;`.

vonagam:
The only problem I see is that in some cases you do not want a type to be opaque.

I'm glancing at `gleam/javascript` and I assume that the api can work not only with map/promise created with `new` in that module but also random map/promises created in js land. For such cases `any` works (though again with all type params lost) but an opaque type will not.

So there needs to be some way to annotate what type to generate in ts. I would say - generate opaque type by default unless a custom definition is provided.

vonagam:
Ok, makes sense, so where the discussion should be opened about this? I thought PR is as good place to discuss as any.

> {} is a special type, is it not the same as an object with a single key, so we cannot treat it the same way.

It was an example. Again, `type X = {toString: Function}; let x: X = true`, is a valid code. It ts it is not an object, it is about properties. Structural typing and stuff.

> External types are to be constructed in the target language, so using types to forbid their construction means that they could not be constructed in either language.

Part about "either" is not right. From gleam side everything works. And from js side that uses gleam to construct opaque instance everything works. The only thing opaqueness forbids is to come up with a value outside of gleam.

lpil:
Issues are for problems or work that is ready to work on, so discord or discussions would be a good place to start. When ready to work on, or a problem is identified with the current design, then it's time for an issue.

> type X = {toString: Function}; let x: X = true, is a valid code. It ts it is not an object, it is about properties. Structural typing and stuff.

That's not comparable I'm afraid. That definition is structurally compatible with all JS types, the one you have defined in the PR is not. We could use `{toString: Function}`, but I think that would be less clear than using `any`.

> The only thing opaqueness forbids is to come up with a value outside of gleam.

External types are for types that are constructed outside of Gleam.

vonagam:
Opened the discussion at https://github.com/gleam-lang/gleam/discussions/2312.

lpil:
> Depends.

It does not depend, they are always constructed outside of Gleam.

> Map type from stdlib is not supposed to be created by any other means than by calling new from stdlib, right?

No, it can be constructed anywhere. This is probably one of the most commonly constructed types outside of Gleam, and even the stdlib constructs it outside of Gleam and then calls that code from Gleam.

vonagam:
Depends on what you mean by outside of Gleam. What I meant is that the user of stdlib Map is not supposed to create a map by importing `PMap` from `persistent-hash-map.mjs`. He was supposed to call a function `new` written in gleam. So from perspective of a user `Map` from stdlib is an opaque type.

It is in contrast to new javascript gleam library which is supposed to work with types which can be created and worked without gleam api - native js Map and Promise.

Anyway this is irrelevant now as I understand that external types are not meant to be opaque, they just to happen to be in stdlib for js. I'm more interested if there are any issues with proposed usage of `@external` for `type`.

PR #2310, opened 2023-08-30
Display type when hovering over function arguments

### LINKED ISSUE #2240: LSP: Display type when hovering over function parameter
Consider the following as an example:
```gleam
fn first_if_true(test, first, second) {
  case test(first) {
    True -> first
    False -> second
  } 
}
```
When hovering over `test`, `first`, or `second` inside the function head, the type information is not displayed. 
You would need to hover over instances of those parameters inside the function body to get the type information.

Chriscbr:
Initial pass at adding LSP hover support for function arguments.

Closes #2240

### DIFF
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -16,7 +16,8 @@ pub use self::project_compiler::{Built, Options, ProjectCompiler};
 pub use self::telemetry::{NullTelemetry, Telemetry};
 
 use crate::ast::{
-    CustomType, DefinitionLocation, TypedDefinition, TypedExpr, TypedPattern, TypedStatement,
+    CustomType, DefinitionLocation, TypedArg, TypedDefinition, TypedExpr, TypedPattern,
+    TypedStatement,
 };
 use crate::{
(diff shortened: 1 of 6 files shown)


compiler-core/src/build.rs
```diff
@@ -266,6 +267,7 @@ pub enum Located<'a> {
     Statement(&'a TypedStatement),
     Expression(&'a TypedExpr),
     ModuleStatement(&'a TypedDefinition),
+    FunctionArgument(&'a TypedArg),
```
Chriscbr, line 270: I wasn't too sure if adding a new enum case here would be appropriate or not.

nino:
I think this functionality is included in the [PR I made a while ago](https://github.com/gleam-lang/gleam/pull/2302) (assuming that will be merged at some point). (Sorry for not tagging the function argument ticket in the PR!)

lpil:
Very sorry @Chriscbr , but I have just merged @nino's patch. Thank you for your contribution though, the code looked good!

Chriscbr:
My bad! Thanks for the heads up 👍

PR #2316, opened 2023-09-01
Prefix module names with dollar in typescript to avoid name collisions

vonagam:
```gleam
// foo.gleam
pub fn stuff() {
  0
}
```
```gleam
// bar.gleam
import foo

pub fn foo() {
  foo.stuff()
}
```
Produces declaration for bar:
```ts
// bar.d.ts
import type * as foo from "./foo.d.mts";

export function foo(): number;
```
Which shows this error on `foo` in imports:
```
Import declaration conflicts with local declaration of 'foo'.
```
Js file does not have such conflict as star imports are prefixed with dollar sign:
```js
// bar.mjs
import * as $foo from "./foo.mjs";

export function foo() {
  return $foo.stuff();
}
```
The same method was used in typescript before, there is still a comment that module names are supposed to be prefixed with "$".

So this PR returns that prefixing with dollar for typescript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@
   or the type of the hovered argument.
 - Use `import type` for importing types from typescript declarations.
 - Use `.d.mts` extension for typescript declarations to match `.mjs`.
+- Prefix module names with dollar sign in typescript to avoid name collisions.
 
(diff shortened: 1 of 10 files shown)


lpil (approved):
Thank you!

PR #2321, opened 2023-09-02
LSP: add code action to remove unused import

### LINKED ISSUE #2318: LSP: code action to remove unused imports
This is a feature request for a new language server code action to remove unused imports. When import are unused, next to the warning, the server should provides an action to delete the lines importing unused module.

TristanCacqueray:
This PR adds a new code action to fix unused import warnings.

Fixes #2318

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -219,6 +219,64 @@ where
         })
     }
 
+    pub fn action(
+        &mut self,
+        params: lsp::CodeActionParams,
+    ) -> Response<Option<Vec<lsp_types::CodeAction>>> {
+        eprintln!("Got a new action request: {:?}", params);
+        self.respond(|this| {
+            let mut actions = vec![];
+
+            // question: how to get the current module warnings?
+            // These could be used to generate a single code action to fix all the unused warnings at once?
+            eprintln!("Why is this empty? {:?}", this.compiler.warnings);
+
+            // here is a demo action:
+            match params
+                .context
+                .diagnostics
+                .iter()
+                .find(|diag| diag.message.starts_with("Unused imported"))
+            {
+                Some(diag) => {
+                    let edit = lsp_types::TextEdit {
+                        range: lsp_types::Range::new(
+                            lsp_types::Position {
+                                line: diag.range.start.line,
+                                character: 0,
+                            },
+                            lsp_types::Position {
+                                line: diag.range.start.line + 1,
+                                character: 0,
+                            },
+                        ),
+                        new_text: "".to_string(),
+                    };
+                    let mut changes = std::collections::HashMap::new();
+                    let _ = changes.insert(params.text_document.uri, vec![edit]);
+                    let action = lsp_types::CodeAction {
+                        title: "Remove unused imports".to_string(),
+                        kind: None,
+                        diagnostics: None,
+                        edit: Some(lsp_types::WorkspaceEdit {
+                            changes: Some(changes),
+                            document_changes: None,
+                            change_annotations: None,
+                        }),
+                        command: None,
+                        is_preferred: Some(true),
+                        disabled: None,
+                        data: None,
+                    };
+                    actions.push(action)
+                }
+                None => (),
+            }
+
+            Ok(Some(actions))
+        })
+    }
+
     fn respond<T>(&mut self, handler: impl FnOnce(&mut Self) -> Result<T>) -> Response<T> {
         let result = handler(self);
         let warnings = self.take_warnings();
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -228,7 +228,12 @@ where
 
             // Check if unused removal can be performed
             if let Some(ranges) = this.get_unused_ranges(&params.text_document.uri) {
-                actions.push(make_unused_code_action(params.text_document.uri, &ranges))
+                if ranges
+                    .iter()
+                    .any(|unused_range| range_includes(&params.range, unused_range))
+                {
+                    actions.push(make_unused_code_action(params.text_document.uri, &ranges))
+                }
             }
 
             Ok(if actions.is_empty() {
@@ -609,4 +614,10 @@ fn make_unused_code_action(uri: Url, ranges: &[lsp_types::Range]) -> lsp_types::
         disabled: None,
         data: None,
     }
+}
+
+// Check if the inner range is included in the outer range.
+fn range_includes(outer: &lsp_types::Range, inner: &lsp_types::Range) -> bool {
+    (outer.start >= inner.start && outer.start <= inner.end)
+        || (outer.end >= inner.start && outer.end <= inner.end)
 }
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -227,10 +227,8 @@ where
             let mut actions = vec![];
 
             // Check if unused removal can be performed
-            if is_action_attached_to_unused_diagnostic(&params) {
-                if let Some(ranges) = this.get_unused_ranges(&params.text_document.uri) {
-                    actions.push(make_unused_code_action(params.text_document.uri, &ranges))
-                }
+            if let Some(ranges) = this.get_unused_ranges(&params.text_document.uri) {
+                actions.push(make_unused_code_action(params.text_document.uri, &ranges))
             }
 
             Ok(if actions.is_empty() {
@@ -581,14 +579,6 @@ fn hover_for_expression(expression: &TypedExpr, line_numbers: LineNumbers) -> Ho
         contents: HoverContents::Scalar(MarkedString::String(contents)),
         range: Some(src_span_to_lsp_range(expression.location(), &line_numbers)),
     }
-}
-
-fn is_action_attached_to_unused_diagnostic(params: &lsp::CodeActionParams) -> bool {
-    params
-        .context
-        .diagnostics
-        .iter()
-        .any(|diag| diag.message.starts_with("Unused"))
 }
 
 // Convert a list of unused range into a "Remove unused" code action.
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -223,58 +223,39 @@ where
         &mut self,
         params: lsp::CodeActionParams,
     ) -> Response<Option<Vec<lsp_types::CodeAction>>> {
-        eprintln!("Got a new action request: {:?}", params);
         self.respond(|this| {
             let mut actions = vec![];
 
-            // question: how to get the current module warnings?
-            // These could be used to generate a single code action to fix all the unused warnings at once?
-            eprintln!("Why is this empty? {:?}", this.compiler.warnings);
-
-            // here is a demo action:
-            match params
-                .context
-                .diagnostics
-                .iter()
-                .find(|diag| diag.message.starts_with("Unused imported"))
-            {
-                Some(diag) => {
-                    let edit = lsp_types::TextEdit {
-                        range: lsp_types::Range::new(
-                            lsp_types::Position {
-                                line: diag.range.start.line,
-                                character: 0,
-                            },
-                            lsp_types::Position {
-                                line: diag.range.start.line + 1,
-                                character: 0,
-                            },
-                        ),
-                        new_text: "".to_string(),
-                    };
-                    let mut changes = std::collections::HashMap::new();
-                    let _ = changes.insert(params.text_document.uri, vec![edit]);
-                    let action = lsp_types::CodeAction {
-                        title: "Remove unused imports".to_string(),
-                        kind: None,
-                        diagnostics: None,
-                        edit: Some(lsp_types::WorkspaceEdit {
-                            changes: Some(changes),
-                            document_changes: None,
-                            change_annotations: None,
-                        }),
-                        command: None,
-                        is_preferred: Some(true),
-                        disabled: None,
-                        data: None,
-                    };
-                    actions.push(action)
+            // Check if unused removal can be performed
+            if is_action_attached_to_unused_diagnostic(&params) {
+                if let Some(ranges) = this.get_unused_ranges(&params.text_document.uri) {
+                    actions.push(make_unused_code_action(params.text_document.uri, &ranges))
                 }
-                None => (),
             }
 
-            Ok(Some(actions))
+            Ok(if actions.is_empty() {
+                None
+            } else {
+                Some(actions)
+            })
         })
+    }
+
+    fn get_unused_ranges(&mut self, uri: &Url) -> Option<Vec<lsp_types::Range>> {
+        let module = self.module_for_uri(uri)?;
+        let unused = &module.ast.type_info.unused;
+        if unused.is_empty() {
+            None
+        } else {
+            // Convert to lsp range
+            let line_numbers = LineNumbers::new(&module.code);
+            Some(
+                unused
+                    .iter()
+                    .map(|location| src_span_to_lsp_range(*location, &line_numbers))
+                    .collect(),
+            )
+        }
     }
 
     fn respond<T>(&mut self, handler: impl FnOnce(&mut Self) -> Result<T>) -> Response<T> {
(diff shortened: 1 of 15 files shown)


TristanCacqueray:
Well, I couldn't figure how to get the current warnings to collect all the unused imports and fix them all at once. I left a question in the code about the `compiler.warnings` being empty. If you could help me with that, I should be able to complete this feature.

The PR currently implements a proof of concept demo that works with the current diagnostic attached to the action request:

[Screencast from 2023-09-02 12-10-49.webm](https://github.com/gleam-lang/gleam/assets/154392/4aa02ec1-31af-4f0b-ba60-f0c9602d4539)

compiler-core/src/language_server/engine.rs
```diff
@@ -219,6 +219,64 @@ where
...
+        &mut self,
+        params: lsp::CodeActionParams,
+    ) -> Response<Option<Vec<lsp_types::CodeAction>>> {
+        eprintln!("Got a new action request: {:?}", params);
+        self.respond(|this| {
+            let mut actions = vec![];
+
+            // question: how to get the current module warnings?
+            // These could be used to generate a single code action to fix all the unused warnings at once?
+            eprintln!("Why is this empty? {:?}", this.compiler.warnings);
```
lpil, line 232: Warnings are events that are dispatched to the user, so this vector is an IO buffer that has recently been flushed.

The analyser will need to be modified to store the locations of all unused imports, and then that can be used here.

lpil (commented):
Awesome! Left a note about how to identify the unused ones

compiler-core/src/language_server/engine.rs
TristanCacqueray: Great, thank you for the explanation. So I did that in d0e9718, and the action now seems to work for multiple unused imported module. However it's a bit annoying because neither the ast nor the warning contains the "import" token, so I had to do some string lookups to make it works. If that looks good to you, then I'd like to add some tests and perhaps also handle unused imported values.

TristanCacqueray:
Here is a new demo:

[Screencast from 2023-09-04 11-36-03.webm](https://github.com/gleam-lang/gleam/assets/154392/ed8fd2be-6886-4eb8-9229-d1be5ed6c0a5)

The last import is not removed because unused imported values are not implemented yet. Though the named qualifier should have a warning, but it seems like unqualified imports does not yield an unused warning for the as qualifier. That is going to be tricky to handle from the lsp server side, and perhaps we should add this missing warning to the compiler first.

TristanCacqueray:
I am marking this as ready for review since the feature is now complete (minus the issue with `as name` not being marked as unused when there are qualified imports). However I'm not super happy with the implementation, it was tricky to satisfy the borrow checker while collecting the warnings for qualified imports.

Let me know if this strategy looks good for you, I'd be happy to improve the code if needed :)

compiler-core/src/language_server/engine.rs
TristanCacqueray: Ok, like this https://github.com/gleam-lang/gleam/pull/2325 ?

lpil: Perfect!

compiler-core/src/language_server/engine.rs
```diff
@@ -219,13 +224,125 @@ where
...
+
+        // Group warnings per module
+        let mut warnings_per_module: HashMap<Utf8PathBuf, Vec<&crate::type_::Warning>> =
+            HashMap::new();
+        for warning in warnings {
+            if let Warning::Type { path, warning, .. } = warning {
+                if let Some(vec) = warnings_per_module.get_mut(path) {
+                    vec.push(warning);
+                } else {
+                    let _ = warnings_per_module.insert(path.clone(), vec![warning]);
```
lpil, line 273: Can filter for the unused imports here, simplify the second loop a bit.

compiler-core/src/language_server/engine.rs
```diff
@@ -219,13 +224,125 @@ where
...
+                                store_unused_warning_range(&mut ranges_per_location, &path, range);
+                            }
+                        }
+
+                        crate::type_::Warning::UnusedImportedValue { location, .. } => {
+                            if let Some(Located::ModuleStatement(
+                                crate::language_server::engine::Definition::Import(import),
+                            )) = module.find_node(location.start)
+                            {
+                                // Store the qualified import details now, we'll process it after.
```
lpil, line 299: What does this mean?

compiler-core/src/language_server/engine.rs
```diff
@@ -219,13 +224,125 @@ where
...
+                                }
+                            };
+                        }
+                        _ => (),
+                    }
+                }
+
+                for (_, info) in qual_per_module.drain() {
+                    if let Some(ranges) =
+                        handle_unused_imported_module_qualified(&line_numbers, &code, &info)
```
lpil, line 322: What does handle mean?

lpil (commented):
Nice! Thank you.

I'm not sure about the approach of inspecting the warnings to learn about the code. Up until here they've not been an API that has been depended on, they've been all about presenting information to the user. I was imagining the compiler returning a vector of locations for unused items.

What do you think? I'm unsure what is best.

TristanCacqueray:
Oh right, I should have looked at the compiler environment module, it seems better to do this work from there :) I guess I was aiming for a no-cost solution outside of the language server, but I don't think this will impact much the compiler. So to confirm your suggestion, we would pass a new argument to the `handle_unused` function all the way from the `LspProjectCompiler` using such a type:

```rust
type UnusedLocations = Arc<HashMap<Utf8PathBuf, Vec<SrcSpan>>>
```

Could this be added to the existing WarningEmitter struct to avoid updating the whole call stack?

In `handle_unused`, we would add an extra pass for the following cases:

- Unused module, the import location can be removed.
- Unqualified import when all the names are unused, the full import location can also be removed.
- Remaining unused unqualified names needs to be removed indivdually. For that, it seems like we would need to extend the name's location with the surrounding comma depending on the position of the name. For example, in `import m.{unused, a}`, the unused location should be `unused, `. Can this comma lookup be done in place, or should we update the parser to include this extra bit of information?

Finally, we might want to keep track of other unused declarations so that the language server can offer an action to remove them too, but I think that should be done in a follow-up.

Let me know if that sounds good to you, thanks!

compiler-core/src/language_server/engine.rs
```diff
@@ -542,3 +659,142 @@ fn hover_for_expression(expression: &TypedExpr, line_numbers: LineNumbers) -> Ho
...
+        }
+        Some(vec) => vec.push(range),
+    }
+}
+
+fn import_token_location(before: &str) -> Option<usize> {
+    let pos = before.rfind("import")?;
+    // Consume the previous line return if it exists.
+    Some(if pos > 0 { pos - 1 } else { pos })
+}
```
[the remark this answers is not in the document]
TristanCacqueray: Nope, that's good.

compiler-core/src/language_server/engine.rs
```diff
@@ -542,3 +659,142 @@ fn hover_for_expression(expression: &TypedExpr, line_numbers: LineNumbers) -> Ho
...
+
+#[derive(Debug)]
+struct QualifiedUnusedInfo {
+    unqualified_count: usize,
+    has_name: bool,
+    location: SrcSpan,
+    unused: Vec<SrcSpan>,
+}
+fn handle_unused_imported_module_qualified(
```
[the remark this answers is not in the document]
TristanCacqueray: As explained below this is to include the surrounding comma so that unqualified name can be safely removed. I moved this logic to the parser in 8d8af27, though I'm not sure that's better.

compiler-core/src/language_server/engine.rs
```diff
@@ -219,13 +224,125 @@ where
...
+
+    fn module_code(&self, path: &Utf8PathBuf) -> Option<(LineNumbers, &'_ Module, SmolStr)> {
+        let uri = crate::language_server::server::path_to_uri(path.clone());
+        let module = self.module_for_uri(&uri)?;
+        let line_numbers = LineNumbers::new(&module.code);
+        Some((line_numbers, module, module.code.clone()))
+    }
+
+    // This function remember unused warning locations for the related code action.
+    fn store_unused_warnings(&mut self, modules: &[Utf8PathBuf], warnings: &[Warning]) {
```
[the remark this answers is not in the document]
TristanCacqueray: Yes, that's now done in the compiler environment.

lpil:
That type seems more complex than I was thinking, I would just return a vector of the locations of unused things and what kind of thing they are with the compiled module information. No need for shared mutable state as it will be constructed each time the module is compiled and can be returned with the other info.

TristanCacqueray:
You mean adding such vector of unusued things to the `ModuleInterface` ?

lpil:
Yup!

TristanCacqueray:
Alright, the branch now uses the ModuleInterface to store the unused locations for the lsp server. I removed support for removing unqualified import because I'm not sure how to do it properly. I guess we should wait for the future call graph traversal mentioned in https://github.com/gleam-lang/gleam/pull/2327#discussion_r1322959716 so that we can assess when an import can be removed.

compiler-core/src/type_/environment.rs
```diff
@@ -491,18 +491,22 @@ impl<'a> Environment<'a> {
         }
     }
 
-    /// Converts entities with a usage count of 0 to warnings
-    pub fn convert_unused_to_warnings(&mut self) {
+    /// Converts entities with a usage count of 0 to warnings.
+    /// Returns the list of location for the removed unused lsp action.
```
lpil, line 495: Looks like it's specifically unused imported modules, can we say that here please

compiler-core/src/type_.rs
```diff
@@ -459,6 +459,7 @@ pub struct ModuleInterface {
     pub types_constructors: HashMap<SmolStr, Vec<SmolStr>>,
     pub values: HashMap<SmolStr, ValueConstructor>,
     pub accessors: HashMap<SmolStr, AccessorsMap>,
+    pub unused: Vec<SrcSpan>,
```
lpil, line 462: I think these are specifically unused imports?
```suggestion
    pub unused_imports: Vec<SrcSpan>,
```

compiler-core/src/metadata/module_decoder.rs
```diff
@@ -75,6 +75,8 @@ impl ModuleDecoder {
             ),
             values: read_hashmap!(reader.get_values()?, self, value_constructor),
             accessors: read_hashmap!(reader.get_accessors()?, self, accessors_map),
+            // Unused locations are not serialized
+            unused: Vec::new(),
```
lpil, line 79: How come?

compiler-core/src/language_server/tests/action.rs
```diff
@@ -0,0 +1,225 @@
...
+                            - offset;
+                        let range = (start as usize)..(end as usize);
+                        offset += end - start;
+                        result.replace_range(range, &edit.new_text);
+                    }
+                }
+            }
+            None => (),
+        },
+        _ => (),
```
lpil, line 95: Can we panic for these other cases please? So we know why it doesn't work for others in future.

compiler-core/src/language_server/tests/action.rs
```diff
@@ -0,0 +1,225 @@
...
+}
+
+fn apply_code_action(src: &str, action: &lsp_types::CodeAction) -> String {
+    let mut result = src.to_string();
+    let line_numbers = LineNumbers::new(src);
+    let mut offset = 0;
+    match &action.edit {
+        Some(WorkspaceEdit { changes, .. }) => match changes {
+            Some(changes) => {
+                for change in changes.values() {
```
lpil, line 79: Could the code in this branch be extracted into another function please? To split out deciding what needs to be done from how it is done.

compiler-core/src/metadata/module_decoder.rs
TristanCacqueray: I wasn't sure if they needed to be serialized. I've updated the schema in the last commit.

compiler-core/src/language_server/engine.rs
```diff
@@ -543,3 +582,41 @@ fn hover_for_expression(expression: &TypedExpr, line_numbers: LineNumbers) -> Ho
         range: Some(src_span_to_lsp_range(expression.location(), &line_numbers)),
     }
 }
+
+fn is_action_attached_to_unused_diagnostic(params: &lsp::CodeActionParams) -> bool {
+    params
+        .context
+        .diagnostics
+        .iter()
+        .any(|diag| diag.message.starts_with("Unused"))
```
lpil, line 591: Check this from the `unused_imports` vector please. Diagnostic messages are not a reliable interface, and it's possible this buffer has already been flushed at this point.

TristanCacqueray: Right, so for this part I'm not entirely sure how LSP works. If we remove this check, and use the `unused_imports` vector, then the server will always suggest to remove unused imports, even when the click on unrelated warnings. Another solution would be to check if the code action request params range falls into an import from the ast.

One thing I couldn't find yet, is how to suggest code action without waiting for the user request, I guess that would be the right thing to do instead of a waiting for a code action request.

TristanCacqueray:
I've rebased this PR on top of #2327 and confirmed that the removed unused alias test works as expected.

compiler-core/src/language_server/engine.rs
lpil: It must not be based off of warnings at all. They are not an interface for any kind of reflection.

From reading the code it adds the actions for each item in the `unused_imports` vector, so if there's none then none will be added.

lpil (commented):
Sorry for the slow reply, I've been unwell with covid 🤒

TristanCacqueray:
> Sorry for the slow reply, I've been unwell with covid 🤒

No worries, get well soon! I'll update this PR next week or the week after, after I get back from vacation.
Cheers!

compiler-core/src/language_server/engine.rs
TristanCacqueray: Ok, but if there is an unused import, then the action will be proposed, even when the user click on an unrelated warning. May I lookup the action request location to see if it match an import location?

TristanCacqueray:
To recap, this PR implement unused imports and unused import aliases removal.

Handling qualified import is no longer implemented, and perhaps this should be done in a follow up as it requires ground work to be done on the environment analysis (e.g. to be able to check if all the qualified names for a given import are unused so that the whole import statement can be safely removed) and ast annotations (e.g. to get the surrounding comma separator location for removing a single qualified name, and the full `{...}` location to remove all qualified name. See the commented `test_remove_unused_qualified_` for example scenarios.

compiler-core/src/language_server/engine.rs
lpil: Sorry, I don't understand. Why would the `unused_imports` vector be filled with locations of unrelated warnings?

TristanCacqueray: Oh no, the `unused_imports` only contains the locations of unused imports. What I am trying to say is that presently, the LanguageServerEngine.action method is called when the user requests a code action and this [request](https://docs.rs/lsp-types/latest/lsp_types/struct.CodeActionParams.html) can come from anywhere in the document.

Therefor, only checking the `unused_imports` vector is not enough, we need to make sure the user actually request a code action that is coming from an import location, and not, for example, from a non exhaustive pattern match error.

Currently the server always offers to remove unused import (if there are any), but we could improve that by checking that the request range is overlapping with import node from the ast, that way, the action would be only available when the user clicks on an unused import warning.

Hope that is more clear.

lpil: Yes, that's right. The user presses the hotkey for suggestions for code actions, the editor asks the server, the server returns a code action if the position is in the `unused_imports`, so the user only gets a code action suggested if they are positioned in an unused value.

TristanCacqueray: Oh right, we can use the location of the unused_imports directly instead of the looking up the ast. I'll give this a try shortly.

Sorry for the confusion, I got a bit lost here :)

TristanCacqueray: Alright, this is now implemented.

lpil (approved):
Brilliant, thank you so much ! Sorry it took me so long to get this though 💜

TristanCacqueray:
You are welcome, I am happy to help you build such an amazing language. Also, since this work, I'm now adopting capnp in my projects :) Thank you for taking the time to review this PR!

PR #2325, opened 2023-09-05
Adjust the location field on the import to include the keyword

TristanCacqueray:
This change updates the parser to include the "import" keyword in the ast location.


lpil (approved):
Perfect, thank you!!

Issue #2326, opened 2023-09-06
Import with as name and unqualified are not marked as unused


TristanCacqueray:
Given this module:

```gleam
import gleam/map.{drop} as map
```

The compiler only warns about `drop` being unused:

```ShellSession
warning: Unused imported value
  ┌─ test_project.gleam:1:19
  │
1 │ import gleam/map.{drop} as map
  │                   ^^^^ This imported value is never used.

Hint: You can safely remove it.
```

Shouldn't the warning include the full statement? e.g.:

```ShellSession
warning: Unused imported module
  ┌─ test_project.gleam:1:1
  │
1 │ import gleam/map.{drop} as map
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This imported module is never used.

Hint: You can safely remove it.
```

lpil:
Thank you

TristanCacqueray:
What about (from unused_imported_module_with_alias_and_unqualified_name_no_warnings_test):

```gleam
import gleam/foo.{bar} as foo

pub fn baz() { bar() }
```

Shouldn't this produce a warning? e.g.:

```
warning: Unused imported module name
  ┌─ test:1:1
  │
1 │ import gleam/foo.{bar} as foo
  │                        ^^^^^^ This imported module name is never used.

Hint: You can safely remove it.
```

lpil:
Makes sense!

massivefermion:
Oh...I just realized there already is a PR for this!

PR #2327, opened 2023-09-07
Handle unused imported module with alias and unqualified name

TristanCacqueray:
Fixes: #2326


compiler-core/src/type_/tests/warnings.rs
```diff
@@ -434,6 +434,38 @@ fn unused_imported_module_with_alias_warnings_test() {
...
+// https://github.com/gleam-lang/gleam/issues/2326
+#[test]
+fn unused_imported_module_with_alias_and_unqualified_name_warnings_test() {
+    let warnings = get_warnings(
+        "import gleam/foo.{bar} as foo",
+        vec![("thepackage", "gleam/foo", "pub fn bar() { 1 }")],
+    );
+    assert!(!warnings.is_empty());
+    assert_eq!(
+        Warning::UnusedImportedValue {
```
TristanCacqueray, line 446: Perhaps this warning should be omitted in that situation?

compiler-core/src/type_/environment.rs
```diff
@@ -482,11 +482,17 @@ impl<'a> Environment<'a> {
...
-        self.handle_unused(unused);
+        self.handle_unused(unused.clone());
 
         for (name, location) in self.unused_modules.clone().into_iter() {
-            self.warnings
-                .emit(Warning::UnusedImportedModule { name, location });
+            // Check if the module is really unused
+            if !unused
+                .iter()
+                .any(|(_, (_, loc, used))| *used && location.contains(loc.start))
```
lpil, line 491: What is this doing? To me it seems something is off if we have to do this additional checking, and I don't understand exactly what this is from the code and the comment.

compiler-core/src/type_/tests/warnings.rs
lpil: Oh maybe! If we can cleanly do that.

compiler-core/src/type_/environment.rs
TristanCacqueray: This is checking if one of the unqualified name is used. I agree it seems something is off. I guess the issue is that `as name` usage needs to be handled independently, e.g. by using a new `UnusedImportedModuleName` warning. To emit the right location, could we add this new struct:

```rust
pub struct ImportName {
  pub name: SmolStr,
  pub location: SrcSpan,
}
```

... for the `as_name` field of the Import:

```diff
  pub struct Import<PackageName> {
    pub documentation: Option<SmolStr>,
    pub location: SrcSpan,
    pub module: SmolStr,
-   pub as_name: Option<SmolStr>,
+   pub as_name: Option<ImportName>,
    pub unqualified: Vec<UnqualifiedImport>,
    pub package: PackageName,
}
```

compiler-core/src/type_/tests/warnings.rs
TristanCacqueray: What I had trouble with in `handle_unused`, is that the imported names are no longer associated with their import, so it's difficult to check if a whole import statement is unused. I think we would need to add the original module name in the EntityKind as proposed here: https://github.com/gleam-lang/gleam/blob/8d8af276334f78c92a7011897702809a18aabf91/compiler-core/src/type_/environment.rs#L94-L98

What do you think about that?

lpil: I don't think we should make this any more complex as it will be replace with a proper call graph traversal in future.

TristanCacqueray: Alright, so until then, is it ok to emit the extra warning as shown in the test? If so then I think this PR is ready, and I would rebase #2321 on top of this.

lpil: Yup sounds good 👍

compiler-core/src/type_/environment.rs
```diff
@@ -497,11 +501,15 @@ impl<'a> Environment<'a> {
...
             .pop()
             .expect("Expected a bottom level of entity usages.");
-        self.handle_unused(unused);
+        self.handle_unused(unused.clone());
 
         for (name, location) in self.unused_modules.clone().into_iter() {
-            self.warnings
-                .emit(Warning::UnusedImportedModule { name, location });
+            let warning = if self.imported_module_names.contains(&name) {
+                Warning::UnusedImportedModuleName { name, location }
```
lpil, line 508: Took me quite a while to understand this warning and the `imported_module_names`. They're not module names, they're module aliases. A module always has a name, but only sometimes has an alias. Let's rename these.

compiler-core/src/warning.rs
```diff
@@ -437,6 +437,22 @@ expression.",
                     }),
                 },
 
+                type_::Warning::UnusedImportedModuleName { location, .. } => Diagnostic {
+                    title: "Unused imported module name".into(),
```
lpil, line 441: Alias, not a name

compiler-core/src/type_/environment.rs
TristanCacqueray: Oops, sorry about that, module aliases looks better, thanks!

compiler-core/src/type_/environment.rs
```diff
@@ -497,11 +501,15 @@ impl<'a> Environment<'a> {
             .entity_usages
             .pop()
             .expect("Expected a bottom level of entity usages.");
-        self.handle_unused(unused);
+        self.handle_unused(unused.clone());
```
lpil, line 504: I don't think we need to clone here? I can't see it being used afterwards

lpil (commented):
Sorry I'm a bit confused now. Is this sharing code with the other PR?

compiler-core/src/type_/environment.rs
TristanCacqueray: You are right, thanks!

TristanCacqueray:
> Sorry I'm a bit confused now. Is this sharing code with the other PR?

No, this is just adding a new warning and the locations of module aliases. Once this is merged, the code action would make use of it.

lpil (approved):
Thank you! I'll just hold off on merging this as I'm doing a release.

lpil (commented):
Could you update the changelog too please? 🙏

TristanCacqueray:
Sure, I've rebased the PR and updated the CHANGELOG.

lpil (approved):
Thank you!!

TristanCacqueray:
That's great, you are welcome, thank you for the many reviews!

PR #2329, opened 2023-09-07
Fix links in error messages

polynomialherder:
The [getting-started](https://gleam.run/getting-started/) page linked in the existing error messages don't actually reference Erlang or rebar3 at all -- I think the documentation being referenced is the material in [installing](https://gleam.run/getting-started/installing/).


lpil (approved):
Thank you

PR #2330, opened 2023-09-08
Remove boxing from single arity opaque types

lpil:
A small performance optimisation for niche uses.


lpil:
Closing as the implementation became more complex than I wanted it to be for the value of the feature.

PR #2334, opened 2023-09-12
2278 path dependencies module resolution bug

### LINKED ISSUE #2278: Path dependencies module resolution bug
## Description of behaviour

## Steps to reproduce

```
foo/
  gleam.toml
  src/
    foo.gleam
bar/
  gleam.toml
  src/
    bar.gleam
```

```toml
# foo/gleam.toml 

[dependencies]
bar = { path = "../bar" }
```

```gleam
// foo/src/foo.gleam

import bar

pub fn main() {
  bar.say_hello()
}
```

```toml
# bar/gleam.toml

[dependencies]
gleam_stdlib = "~> 0.29"
```

```gleam
// bar/src/bar.gleam

import gleam/io

pub fn say_hello() {
  io.println("Hello from bar!")
}
```

```sh
~/foo $ gleam build
~/foo $ cd ../bar
~/bar $ gleam add simplifile
```

```gleam
// bar/src/bar.gleam

import simplifile
import gleam/io

pub fn say_hello() {

  io.println("Hello from bar!")
}
```

```sh
~/foo $ gleam build
```

```
  Compiling gleam_stdlib
  Compiling bar
error: Unknown module
  │
1 │ import simplifile
  │        ^^^^^^^^^^

```

macroby:
Fixes #2278 
When building a project, we must check that a manifest file already exists. If the manifest exists, then we must check that it is up to date. It is at this stage that the issue occurs. When verifying the manifest, we currently only check for changes in the root config file. This does not account for changes in the config files of local packages that occur after the root manifest is generated. To resolve this, I've added a check for changes in local package config files.

### DIFF
--- a/compiler-core/src/docs/snapshots/gleam_core__docs__tests__tables.snap
+++ b/compiler-core/src/docs/snapshots/gleam_core__docs__tests__tables.snap
@@ -1,0 +1,357 @@
+---
+source: compiler-core/src/docs/tests.rs
+assertion_line: 48
+expression: "compile(config, modules)"
+---
+//// app.html
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1"/>
+    <title>app ·  · v0.1.0</title>
+    <meta name="description" content=""/>
+    <meta name="theme-color" content="#ffaff3" media="(prefers-color-scheme: light)"/>
+    <meta name="theme-color" content="#33384d" media="(prefers-color-scheme: dark)"/>
+    <link rel="stylesheet" href="./css/index.css?v=0.32.0-dev" type="text/css"/>
+    <!-- The docs_config.js file is provided by HexDocs and shared
+         between multiple versions of the same package. -->
+    <script src="./docs_config.js"></script>
+    <link id="syntax-theme" rel="stylesheet" href="./css/atom-one-light.min.css?v=0.32.0-dev"/>
+  </head>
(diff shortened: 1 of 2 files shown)


compiler-cli/src/dependencies.rs
```diff
@@ -512,6 +514,18 @@ fn get_manifest<Telem: Telemetry>(
...
+fn get_all_dependencies(config: &PackageConfig) -> Result<Dependencies> {
+    let mut dependencies = config.all_dependencies()?;
+    for (_name, requirement) in dependencies.clone().into_iter() {
+        if let Requirement::Path { path } = requirement {
+            let config = crate::config::read(path.join("gleam.toml"))?;
+            dependencies.extend(config.all_dependencies()?);
+        };
+    }
+    Ok(dependencies)
+}
```
lpil, line 527: I think this will fail for nested path deps, as the path will be relative to the dep and not the root of the project.

compiler-cli/src/dependencies.rs
```diff
@@ -500,9 +500,11 @@ fn get_manifest<Telem: Telemetry>(
 
     let manifest = read_manifest_from_disc(paths)?;
 
+    let dependencies = get_all_dependencies(config)?;
```
lpil, line 503: This detection is being done at the very edge of the system, when the CLI starts. It works for the CLI build tool as it always starts from scratch, but the problem will still exist for the language server which is a long-running process. We'll need to determine an approach that works in that case also.

lpil (commented):
Thank you for this! I've added some questions inline.

PR #2336, opened 2023-09-15
Add support for module access in case clause guards

massivefermion:
There is still a problem with the js side and that is why it's a draft.

Consider the following module(`hero.gleam`):
```gleam
pub type Hero {
  Hero(name: String)
}

pub const ironman = Hero("Tony Stark")
```

The following code:
```gleam
import hero

pub fn main() {
  let name = "Tony Stark"

  case name {
    n if n == hero.ironman.name -> True
    _ -> False
  }
}
```

Would produce:

```javascript
import * as $hero from "./hero.mjs";

export function main() {
  let name = "Tony Stark";
  if (name === new Hero("Tony Stark").name) {
    let n = name;
    return true;
  } else {
    return false;
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - A warning is now emitted if a module from a transitive dependency is imported.
 - Record access can now be used in case clause guards.
+- Module access can now be used in case clause guards.
 - Fixed a bug where `manifest.toml` could contain absolute paths for path
   dependencies.
 - The `as` keyword can now be used to assign the literal prefix to a variable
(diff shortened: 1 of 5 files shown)


massivefermion:
I know what the problem is now. Apparently the method I've used here is completely wrong. Instead of taking the value from the module at compile time and putting it in the AST, I need to add a `ModuleSelect` variant to `ClauseGuard` and put that in the AST. The fact that this method works for erlang is completely accidental.
The problem is I don't know how to work with such a variant.

PR #2337, opened 2023-09-15
add support for module access in case clause guards

massivefermion:


compiler-core/src/type_/expression.rs
```diff
@@ -1062,24 +1062,72 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
-                    label,
-                )?;
-                Ok(ClauseGuard::FieldAccess {
-                    container,
-                    label,
-                    index: Some(index),
-                    location,
-                    type_,
-                })
+            } => match self.infer_clause_guard(*container.clone()) {
```
lpil, line 1065: Finding this function quite hard to read with its mix of different concerns and very large amount of nesting. Could you extract some functions from this? Perhaps one for records and one for modules.

lpil (commented):
Brilliant! Thank you!

Left a small note inline. Would you also mind adding some tests to https://github.com/gleam-lang/gleam/blob/main/test/language/test/language_test.gleam? Thanks

lpil (approved):
Thank you!!

Issue #2236, opened 2023-06-21
✨ Allow imports to be aliased with `as _`.


hayleigh-dot-dev:
This would allow you to do:

```gleam
import wibble.{wobble} as _
```

to effectively hide the `wibble` module, only exposing the `wobble` function. This is particularly useful if there are multiple modules with the same name but you only want to import specific things from each:

```gleam
import mist/internal/handler.{HandlerResponse} as _
import glisten/handler.{HandlerMessage} as _
```

thefloofe22:
Would love to work on this if it's still up for grabs

lpil:
Brilliant, thank you @thefloofe22 !

TristanCacqueray:
Shouldn't the module be implicitly hidden when it is imported with unqualified names? That's how Haskell's import work.

lpil:
No, it should not.

PR #2339, opened 2023-09-16
allow discarding imported modules

massivefermion:
addresses #2236 

I should mention that I allowed any name staring with `_` to be used. I think it could be useful. Imagine the situation below:

```gleam
import mist/internal/handler.{HandlerResponse} as _internal_handler
import glisten/handler.{HandlerMessage} as _glisten_handler
```

This way the developer can just go to the imports and remove the `_` and now they can just use the module. I like the convenience.

EDIT: also addresses #2326


compiler-core/src/ast.rs
```diff
@@ -378,6 +378,8 @@ pub struct Import<PackageName> {
     pub location: SrcSpan,
     pub module: SmolStr,
     pub as_name: Option<SmolStr>,
+    pub as_span: Option<SrcSpan>,
```
lpil, line 381: Let's merge `discarded`, `as_name` and `as_span` into `alias: Option<(AssignName, SrcSpan)>` as it's not valid to have one but not the other.

compiler-core/src/warning.rs
```diff
@@ -437,6 +437,22 @@ expression.",
                     }),
                 },
 
+                type_::Warning::UnusedImportedModuleAlias { location, .. } => Diagnostic {
+                    title: "Unused alias for imported module".into(),
+                    text: "".into(),
+                    hint: Some("You can safely remove it.".into()),
```
lpil, line 443: This might not be true in the case where there is another import using the default name for the module.

compiler-core/src/analyse.rs
```diff
@@ -342,33 +345,40 @@ pub fn register_import(
         }
     }
 
-    if unqualified.is_empty() {
-        // When the module has no unqualified imports, we track its usage
-        // so we can warn if not used by the end of the type checking
+    if !discarded {
```
lpil, line 348: This is going to change due to the removal of the`discarded` field, but generally let's use guard clauses rather than if with indentation.

Indentation is hard to read because you have to read the conditional and then scan down to the bottom of the function to see if there's an `else`, while with a guard you don't have to hold this unresolved question while reading the rest.

lpil (commented):
Thank you!!

compiler-core/src/warning.rs
massivefermion: I'm not sure what you mean by this.
Is this the situation you're talking about?
```gleam
import gleam/io
import two.{four} as _one
import one.{three} as two

pub fn main() {
  io.debug(three)
  io.debug(four)
}
```
I ran this and I got:
```
warning: Unused alias for imported module
  ┌─ <path>/src/dojo.gleam:3:20
  │
3 │ import one.{three} as two
  │                    ^^^^^^ This alias for the imported module is never used.

```
Which is correct!

lpil:
```gleam
import a/one
import a/one as two
```

Here it'll say you can safely remove `as two`, but that is not the case as it will cause a compile error.

massivefermion: It's fixed!

compiler-core/src/ast.rs
massivefermion: I changed it there and now I'm going all over the code to fix the errors. But I can't figure out how to fix this one(I'm fighting with the borrow checker and just can't win). This is in `src/javascript/typescript.ts` in the `collect_imports` function:
```rust
Definition::Import(Import {
                    module,
                    package,
                    as_name,
                    ..
                }) => {
                    if let Some(alias) = as_name {
                        let _ = self.aliased_module_names.insert(module, alias);
                    }
                    self.register_import(&mut imports, package, module);
                }
``
I've changed it to:
```rust
Definition::Import(Import {
                    module,
                    package,
                    alias,
                    ..
                }) => {
                    if let Some((alias, _)) = alias {
                        let alias: SmolStr = (*alias).clone().into();
                        let _ = self.aliased_module_names.insert(module, alias.as_str());
                    }
                    self.register_import(&mut imports, package, module);
                }
```
(I've implemented `From<AssignName>` for `SmolStr`)
But now I get this error at `alias.as_str()`:
```
'alias' does not live long enough
borrowed value does not live long enough
```
I tried a lot of things but I just can't figure out how to solve this.

lpil: Could you remove the `From<AssignName>` from `SmolStr` in favour of a pattern matching please. It's not appropriate as `AssignName` contains 2 different names.

The error is that you're cloning the string, so it gets deallocated when the block finishes, leaving a dangling pointer in the hashmap. If you remove the clone it'll work.

Something like this:

```rust
if let Some((AssignName(name), _)) = alias.as_ref() {
    let _ = self.aliased_module_names.insert(module, alias.as_str());
}
```

test-package-compiler/src/generated_tests.rs
```diff
@@ -13,6 +13,18 @@ fn alias_unqualified_import() {
     );
 }
 
+#[rustfmt::skip]
```
lpil, line 16: Could you remove these build tool tests please? We always use the simplest possible tests for each feature so these should be `type_` tests.

compiler-core/src/warning.rs
lpil: Sorry, how has it been fixed? It still says you can remove it.

compiler-core/src/type_/environment.rs
```diff
@@ -503,6 +505,17 @@ impl<'a> Environment<'a> {
             self.warnings
                 .emit(Warning::UnusedImportedModule { name, location });
         }
+
+        for (alias, location) in self.unused_module_aliases.clone().into_iter() {
+            if let Some((_, module)) = self.imported_modules.get(&alias) {
+                if let Some(name) = module.name.split('/').last() {
+                    if self.imported_modules.get(name).is_none() {
```
lpil, line 512: This is going to not raise an unused import warning if there's another module with a similar name. Another import doesn't stop the other one from being unused, so we want to emit the warning here.

compiler-core/src/ast.rs
```diff
@@ -377,15 +377,16 @@ pub struct Import<PackageName> {
...
     pub package: PackageName,
+    pub alias: Option<(AssignName, SrcSpan)>,
 }
 impl<T> Import<T> {
     pub(crate) fn used_name(&self) -> SmolStr {
-        self.as_name
+        self.alias
             .as_ref()
             .cloned()
+            .map(|(name, _)| name.name().into())
```
lpil, line 389: When the module is discarded this will result in it being treated as if it was aliased. We'd want the variable name only, and the method will have to return none when it is discarded.

Probably will need a test in the language server bit to ensure this is correct.

compiler-core/src/analyse.rs
```diff
@@ -342,33 +343,47 @@ pub fn register_import(
...
 
-    if unqualified.is_empty() {
-        // When the module has no unqualified imports, we track its usage
-        // so we can warn if not used by the end of the type checking
-        let _ = environment
-            .unused_modules
-            .insert(module_name.clone(), *location);
-    }
+    match alias {
+        Some((AssignName::Variable(alias), span)) if !unqualified.is_empty() => {
```
lpil, line 347: We've checking whether the module is aliased multiple times here, but then using a `module_name` value which exists regardless of whether it is discarded. I think this module name should be optional and then we don't have the same check multiple times and potential for invalid data when one exists but the other does not.

compiler-core/src/warning.rs
massivefermion: I think there is a misunderstanding. I just tested this code
```gleam
import a/one
import b/one as two 

pub fn main() {
  todo
}
```
The only warning I get is for modules themselves not being used which is not incorrect. It does not mention the alias at all:
```
warning: Unused imported module
  ┌─ src/dojo.gleam:1:1
  │
1 │ import a/one
  │ ^^^^^^^^^^^^ This imported module is never used.

warning: Unused imported module
  ┌─ src/dojo.gleam:2:1
  │
2 │ import b/one as two
  │ ^^^^^^^^^^^^^^^^^^^ This imported module is never used.

   Compiled in 0.36s
```

massivefermion: I should add that this won't emit a warning, as It shouldn't:
```gleam
import gleam/io
import a/one.{three}
import b/one.{four} as two

pub fn main() {
  io.debug(three)
  io.debug(four)
}
```

PR #2340, opened 2023-09-16
LSP: variables defined using use should show the variable type in the tooltip

### LINKED ISSUE #2335: LSP: Variables defined using "use" should show the variable type in the tooltip
Currently it just shows the return type of the whole `use` expression. When hovering one of the variable names it should show the variable's type instead.

For example below, the tooltip should show `Connection`.

image

Chriscbr:
Fixes #2335


compiler-core/src/ast/typed.rs
```diff
@@ -153,7 +153,9 @@ impl TypedExpr {
     // This could be optimised in places to exit early if the first of a series
     // of expressions is after the byte index.
     pub fn find_node(&self, byte_index: u32) -> Option<Located<'_>> {
-        if !self.location().contains(byte_index) {
+        // arguments of a function may not be in the same location as the function
+        // e.g. in "use" expressions, so we handle it separately
+        if !self.location().contains(byte_index) && !matches!(self, Self::Fn { .. }) {
```
Chriscbr, line 158: I wasn't sure if this was a bug in the compiler or just a side effect of how use expressions are modeled, so I decided to try this naive solution to start. Basically the issue I had was in my unit test, `with_increment` internally receives two `CallArg`'s, the second of which is an inferred / generated function. That function's `location` only reflected the span of its body, and not the span of its arguments.

I'll paste the full debug printout of the Call expr from the unit test I added, if it provides any context. Byte 91 is the location of `a`, and bytes 116-122 are the location of the `"done"` string.

```
Call {
    location: SrcSpan {
        start: 87,
        end: 122,
    },
    typ: App {
        public: true,
        module: "gleam",
        name: "String",
        args: [],
    },
    fun: Var {
        location: SrcSpan {
            start: 96,
            end: 110,
        },
        constructor: ValueConstructor {
            public: false,
            deprecation: NotDeprecated,
            variant: ModuleFn {
                name: "with_increment",
                field_map: None,
                module: "app",
                arity: 2,
                location: SrcSpan {
                    start: 1,
                    end: 50,
                },
                documentation: None,
            },
            type_: Fn {
                args: [
                    App {
                        public: true,
                        module: "gleam",
                        name: "Int",
                        args: [],
                    },
                    Fn {
                        args: [
                            App {
                                public: true,
                                module: "gleam",
                                name: "Int",
                                args: [],
                            },
                        ],
                        retrn: App {
                            public: true,
                            module: "gleam",
                            name: "String",
                            args: [],
                        },
                    },
                ],
                retrn: App {
                    public: true,
                    module: "gleam",
                    name: "String",
                    args: [],
                },
            },
        },
        name: "with_increment",
    },
    args: [
        CallArg {
            label: None,
            location: SrcSpan {
                start: 111,
                end: 112,
            },
            value: Int {
                location: SrcSpan {
                    start: 111,
                    end: 112,
                },
                typ: App {
                    public: true,
                    module: "gleam",
                    name: "Int",
                    args: [],
                },
                value: "3",
            },
            implicit: false,
        },
        CallArg {
            label: None,
            location: SrcSpan {
                start: 116,
                end: 122,
            },
            value: Fn {
                location: SrcSpan {
                    start: 116,
                    end: 122,
                },
                typ: Fn {
                    args: [
                        Var {
                            type_: RefCell {
                                value: Link {
                                    type_: App {
                                        public: true,
                                        module: "gleam",
                                        name: "Int",
                                        args: [],
                                    },
                                },
                            },
                        },
                    ],
                    retrn: App {
                        public: true,
                        module: "gleam",
                        name: "String",
                        args: [],
                    },
                },
                is_capture: false,
                args: [
                    Arg {
                        names: Named {
                            name: "a",
                        },
                        location: SrcSpan {
                            start: 91,
                            end: 92,
                        },
                        annotation: None,
                        type_: Var {
                            type_: RefCell {
                                value: Link {
                                    type_: App {
                                        public: true,
                                        module: "gleam",
                                        name: "Int",
                                        args: [],
                                    },
                                },
                            },
                        },
                    },
                ],
                body: [
                    Expression(
                        String {
                            location: SrcSpan {
                                start: 116,
                                end: 122,
                            },
                            typ: App {
                                public: true,
                                module: "gleam",
                                name: "String",
                                args: [],
                            },
                            value: "done",
                        },
                    ),
                ],
                return_annotation: None,
            },
            implicit: true,
        },
    ],
}
```

lpil: Rather than having a special case here for one variant lets move this logic into each of the cases below.

We'll also need to have tests to ensure that all `a`, `b`, and `c` are being found for this expression, as the logic has changed

```gleam
use a <- b
c
```

lpil (commented):
Thank you! One small thing inline, and if you could update the changelog too that would be excellent.

Chriscbr:
Sorry for the delay - I got another chance to work on this and I think it's shaping up to better now. Let me know whenever you're able to give another pass 🙏

lpil (approved):
Thank you  💜

PR #2341, opened 2023-09-17
added support for bit syntax for module consts for js target

massivefermion:


lpil (commented):
Awesome, thank you!

Could you add some integration tests please? Seeing as this is all tricky runtime stuff. https://github.com/gleam-lang/gleam/blob/main/test/language/test/language_test.gleam

test/language/test/language_test.gleam
```diff
@@ -1060,20 +1060,97 @@ fn bit_string_tests() -> List(Test) {
...
+          0x1,
+          2,
+          2:size(16),
+          0x4:size(32),
+          "Gleam":utf8,
+          4.2:float,
+          <<<<1, 2, 3>>:bit_string, "Gleam":utf8, 1024>>:bit_string,
+        >> == importable.data,
+      )
+    }),
```
lpil, line 1108: Could you remove the duplicated tests out of the target-specific tests please 🙏 

Only tests that cannot be performed on the other targets should be in here.

massivefermion: I removed them, but now I'm confused. This PR is for the js target only, if there shouldn't be any js target specific tests for this, what was I supposed to do in the first place when you asked me to add more tests there?

massivefermion: I removed them but now I'm confused. This PR is for the js target only, so if there isn't supposed to be any js specific test in that file, what was I supposed to do there when you asked me to add more tests?

lpil (approved):
Thank you

PR #2362, opened 2023-10-04
allow discarding imported modules

massivefermion:
addresses #2236


compiler-core/src/ast.rs
```diff
@@ -386,9 +386,10 @@ impl<T, E> Function<T, E> {
 pub type UntypedImport = Import<()>;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ImportName {
-    pub location: SrcSpan,
-    pub name: SmolStr,
+pub enum ImportName {
+    Alias(SrcSpan, SmolStr),
+    Original(SrcSpan, SmolStr),
+    Discarded(SrcSpan, SmolStr),
```
lpil, line 392: The `Original` variant here makes it possible for the "original" name to not match the actual name of the module, which should be impossible. Please use `AssignName` instead.

compiler-core/src/format.rs
```diff
@@ -255,10 +255,11 @@ impl<'comments> Formatter<'comments> {
                     ".{".to_doc().append(unqualified).append("}")
                 };
                 let doc = docvec!["import ", module.as_str(), second];
-                if let Some(as_name) = as_name {
-                    doc.append(" as ").append(&as_name.name)
-                } else {
-                    doc
+                match as_name {
+                    ImportName::Alias(_, name) | ImportName::Discarded(_, name) => {
```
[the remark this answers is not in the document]
massivefermion: That wouldn't happen because the name inside the `Discareded` variant starts with an underscore. I added tests anyway.

compiler-core/src/javascript.rs
```diff
@@ -345,18 +345,17 @@ impl<'a> Generator<'a> {
...
-                    .last()
-                    .expect("JavaScript generator could not identify imported module name.")
-            });
+        let module_name = match as_name {
+            ImportName::Alias(_, name) | ImportName::Original(_, name) => name,
+            ImportName::Discarded(_, _) => module
+                .split('/')
+                .last()
+                .expect("JavaScript generator could not identify imported module name."),
+        };
```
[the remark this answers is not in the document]
massivefermion: good catch. I fixed it and added tests.

compiler-core/src/fix/bit_string_rename.rs
```diff
@@ -48,7 +48,11 @@ impl Fixer {
 
             Definition::Import(i) => {
                 if i.module == "gleam" {
-                    self.prelude_module_import_alias = Some(i.used_name());
+                    self.prelude_module_import_alias = match i.used_name() {
+                        ImportName::Alias(_, name)
+                        | ImportName::Original(_, name)
+                        | ImportName::Discarded(_, name) => Some(name),
```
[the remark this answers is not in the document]
massivefermion: fixed!

compiler-core/src/ast.rs
lpil: There must be no duplication of state in any of the data structures that could result in invalid states being representable. It must use `AssignName`, and the other import struct fields can be checked to determine other facts about the import.

massivefermion: I think this one is resolved too!

compiler-core/src/ast.rs
```diff
@@ -1159,29 +1149,35 @@ impl Default for Inferred<()> {
...
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum AssignVariableType {
+    Alias,
+    Original,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum AssignName {
-    Variable(SmolStr),
+    Variable(SmolStr, SrcSpan, AssignVariableType),
```
lpil, line 1160: Please do not change this type. This new value doesn't make sense in any of the other places this is used and it is again a duplication of information that is already in the import struct. There must be no duplication of any data that could result in invalid states being representable.

massivefermion: I was a bit confused, hopefully not anymore!

compiler-core/src/ast.rs
```diff
@@ -406,19 +400,15 @@ pub struct Import<PackageName> {
...
 
 impl<T> Import<T> {
-    pub(crate) fn used_name(&self) -> SmolStr {
-        self.as_name
-            .as_ref()
-            .map(|as_name| as_name.name.clone())
-            .or_else(|| self.module.split('/').last().map(|s| s.into()))
-            .expect("Import could not identify variable name.")
+    pub(crate) fn used_name(&self) -> Option<(AssignName, SrcSpan)> {
+        self.as_name.clone()
```
lpil, line 411: This function should still return the used name, as the name says. If other code wants to access the AssignName they can do `import.as_name`

lpil (commented):
One last thing I think! Thank you! 💜

compiler-core/src/analyse/imports.rs
```diff
@@ -242,28 +224,72 @@ impl<'a> Importer<'a> {
...
+                self.check_not_a_duplicate_import(&used_name, import.location)?;
+
+                if import.unqualified_types.is_empty() && import.unqualified_values.is_empty() {
+                    // When the module has no unqualified imports, we track its usage
+                    // so we can warn if not used by the end of the type checking
+                    let _ = self
+                        .environment
+                        .unused_modules
+                        .insert(used_name.clone(), import.location);
+                }
```
lpil, line 243: I think this check is exactly the same below. Could we move these out to a shared block please?

```rust
if let Some(name) = import.used_name() {
    self.check_not_a_duplicate_import(&used_name, import.location)?;

    self.environment.imported_modules.insert(name, ...)

    if !import.has_any_unqualified() {
        ...
    }
}
```

compiler-core/src/fix/bit_string_rename.rs
```diff
@@ -50,7 +50,11 @@ impl Fixer {
 
             Definition::Import(i) => {
                 if i.module == "gleam" {
-                    self.prelude_module_import_alias = Some(i.used_name());
+                    self.prelude_module_import_alias = match i.clone().as_name {
```
lpil, line 53: We don't need to clone the entire import here. Generally cloning should be avoided, and when it is done it should only be with the smallest item required, so it is almost always on the right hand side of the `=>`. Here we can take this by reference and then clone the name if needed.

```suggestion
                    self.prelude_module_import_alias = match i.as_name.as_ref() {
```

compiler-core/src/javascript.rs
```diff
@@ -345,18 +345,20 @@ impl<'a> Generator<'a> {
...
-            .unwrap_or_else(|| {
-                module
-                    .split('/')
-                    .last()
-                    .expect("JavaScript generator could not identify imported module name.")
-            });
+        let name = module
+            .split('/')
+            .last()
+            .expect("JavaScript generator could not identify imported module name.");
```
lpil, line 354: This now performs the work even if it is not needed- could make it into an anonymous perhaps?

massivefermion:
I think all the issues are resolved!

compiler-core/src/type_.rs
```diff
@@ -533,6 +533,15 @@ impl ModuleInterface {
             .cloned()
             .collect_vec()
     }
+
+    pub fn used_name(&self) -> SmolStr {
+        SmolStr::from(
+            self.name
+                .split('/')
+                .last()
+                .expect("Could not identify imported module name."),
```
lpil, line 542: This code now panics even when the data is valid. Please have it return an optional.

compiler-core/src/language_server/engine.rs
```diff
@@ -360,7 +359,12 @@ where
                 if !type_.public {
                     continue;
                 }
-                completions.push(type_completion(Some(&alias), name, type_));
+                let module = match import.clone().as_name {
```
lpil, line 362: This is cloning the entire import, this should take the import by reference instead.
```suggestion
                let module = match import.as_name.as_ref() {
```

compiler-core/src/language_server/engine.rs
```diff
@@ -360,7 +359,12 @@ where
...
+                    Some((AssignName::Variable(name), _)) => Some(name),
+                    Some((AssignName::Discard(_), _)) => None,
```
lpil, line 364: This is registering types as being in scope unqualified when they are not actually in scope. Unqualified types are handled in the block below.

compiler-core/src/language_server/engine.rs
```diff
@@ -360,7 +359,12 @@ where
...
+                    None => Some(module.name.clone()),
```
lpil, line 365: This is registering the name of the module in the scope, but it should instead be the used name.

```rust
let Some(alias) = import.used_name() else {
    continue;
}

for (name, type_) in &module.types {
    if !type_.public {
        continue;
    }
    completions.push(type_completion(Some(&alias), name, type_));
}
```

compiler-core/src/language_server/engine.rs
```diff
@@ -399,7 +401,13 @@ where
...
                     continue;
                 }
-                completions.push(value_completion(Some(&alias), name, value));
+
+                let module = match import.as_name.clone() {
+                    Some((AssignName::Variable(name), _)) => Some(name),
+                    Some((AssignName::Discard(_), _)) => None,
+                    None => Some(module.used_name()),
+                };
+                completions.push(value_completion(module.as_deref(), name, value));
```
lpil, line 410: Same as above, this should use `used_name` and should not insert values when the module is discarded.

compiler-core/src/javascript.rs
```diff
@@ -345,18 +345,22 @@ impl<'a> Generator<'a> {
...
+            module
+                .split('/')
+                .last()
+                .expect("JavaScript generator could not identify imported module name.")
+        };
+
+        let (discarded, module_name) = match as_name {
+            None => (false, get_name(module)),
+            Some((AssignName::Discard(_), _)) => (true, get_name(module)),
+            Some((AssignName::Variable(name), _)) => (false, name.as_str()),
```
lpil, line 361: This isn't right, it's using the full name of the module but that may not be a valid JavaScript variable.

It'll also need to handle that two modules could have the same last segment as their name, but now one of them could have been discarded.

compiler-core/src/fix/bit_string_rename.rs
```diff
@@ -50,7 +50,11 @@ impl Fixer {
 
             Definition::Import(i) => {
                 if i.module == "gleam" {
-                    self.prelude_module_import_alias = Some(i.used_name());
+                    self.prelude_module_import_alias = match i.as_name.as_ref() {
+                        Some((AssignName::Variable(name), _)) => Some(name.clone()),
+                        Some((AssignName::Discard(_), _)) => None,
+                        None => Some(i.module.clone()),
+                    }
```
lpil, line 57: This is using the full module name, which isn't a valid variable name.

Please use `used_name` here.

compiler-core/src/analyse/imports.rs
```diff
@@ -242,28 +224,53 @@ impl<'a> Importer<'a> {
...
-                .unused_modules
-                .insert(as_name.name.clone(), as_name.location);
-            // We also register it's name to differentiate between unused module
-            // and unused module name. See 'convert_unused_to_warnings'.
+    fn register_module(
+        &mut self,
+        import: &Import<()>,
+        import_info: &'a ModuleInterface,
+    ) -> Result<(), Error> {
+        let (used_name, location, aliased) = match import.as_name.as_ref() {
```
lpil, line 232: Please use `used_name` here instead of pattern matching on `as_name` as detailed here:

https://github.com/gleam-lang/gleam/pull/2362#discussion_r1367033440

compiler-core/src/type_.rs
massivefermion: Sorry, I don't understand this. In what situation would it panic? The tests are passing, so I don't see how this could be true.

compiler-core/src/javascript.rs
massivefermion: If I understand the first objection correctly, it's actually not using the full name(do you mean `a/b/c`?). Because I'm using the `get_name` lambda to get the last part. Is that the problem or am I misunderstanding it?

Discarded modules are actually handled on line 379. If the module is discarded, an empty vector is passed to `imports.register_module` which then proceeds to only register unqualified imports.

massivefermion:
@lpil In several places you insist on using the `used_name` method of the `Import` struct. The problem is, the way that method was written, it wouldn't distinguish between an import being discarded or not, or whether the name is an alias or the original name. I would have to change the signature to reflect those information but I can't come up with something that does not look weird. Should I define an enum and return the proper variant along with the name to indicate those? something like `(Option<SmolStr>, ImportStatus)` where the possible return values would be `(Some(alias), ImportStatus:Aliased)`, `(None, ImportStatus::Original)` or `(None, ImportStatus::Discarded)`.

lpil:
>  it wouldn't distinguish between an import being discarded or not, or whether the name is an alias or the original name.

That's right, in all the places I've asked for this you don't need to know that information as it is not used in those conditional block, and those blocks have bugs as a result of misusing the import, which is not possible if `used_name` is used instead.

Later code which uses the alias information would match on the alias information directly.

massivefermion:
> > it wouldn't distinguish between an import being discarded or not, or whether the name is an alias or the original name.
> 
> That's right, in all the places I've asked for this you don't need to know that information as it is not used in those conditional block, and those blocks have bugs as a result of misusing the import, which is not possible if `used_name` is used instead.
> 
> Later code which uses the alias information would match on the alias information directly.

You made that comment for changes in `compiler-core/src/analyse/imports.rs` where I'm matching over `import.as_name`. As you can see, that's the place where that distinction is important so I can't use `used_name` there.

lpil:
There is nothing in that block that uses information from `as_name` directly in a correct implementation, and the code becomes clearer. It also calls `used_name` within that block, which means there's a hidden `.expect` there, and the Gleam compiler codebase doesn't have any functions that are unexpectedly partial, or any `.expect`s that can be trivially avoided by using the type system to enforce correct behaviour.

The block `if aliased {` is about aliases, so this alone should match on the alias information.

compiler-core/src/javascript.rs
lpil: Sorry, I misread this line and didn't see the `get_name`. Thank you

massivefermion:
yeah, you're right. I don't know why it didn't come into my mind. Sorry if I made it harder than it needed to be.

lpil (approved):
Thank you! And sorry this took so long to get though 💜

PR #2370, opened 2023-10-17
Add deprecated annotation support for types

omprakaash:
Add deprecated annotation support for types.

Added updated snapshots here as well. The docs snapshot has been updated as the deprecated annotation is included in the generated documentation for types now. (the updated snapshot seems to lose the assertion_line metadata that is usually added by the insta crate.)

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -1085,13 +1085,13 @@ impl<'comments> Formatter<'comments> {
     }
 
     pub fn custom_type<'a, A>(&mut self, ct: &'a CustomType<A>) -> Document<'a> {
-
-        let _ = self.pop_empty_lines(location.start);
+        let _ = self.pop_empty_lines(ct.location.end);
 
         // @deprecated attribute
-        let doc = self.deprecation_attr(deprecation);
+        let doc = self.deprecation_attr(&ct.deprecation);
 
-        let doc =  doc.append(pub_(ct.public))
+        let doc = doc
+            .append(pub_(ct.public))
             .to_doc()
             .append(if ct.opaque { "opaque type " } else { "type " })
             .append(if ct.parameters.is_empty() {
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -19,8 +19,8 @@ use smol_str::SmolStr;
 use std::sync::Arc;
 use vec1::Vec1;
 
-use camino::Utf8Path;
 use crate::type_::Deprecation;
+use camino::Utf8Path;
 
 const INDENT: isize = 2;
 
@@ -1079,6 +1079,7 @@ impl<'comments> Formatter<'comments> {
     pub fn custom_type<'a, A>(&mut self, ct: &'a CustomType<A>) -> Document<'a> {
 
         let _ = self.pop_empty_lines(ct.location.end);
+    
         let doc = docvec![];
 
         // @deprecated attribute
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -1079,7 +1079,6 @@ impl<'comments> Formatter<'comments> {
     pub fn custom_type<'a, A>(&mut self, ct: &'a CustomType<A>) -> Document<'a> {
 
         let _ = self.pop_empty_lines(ct.location.end);
-    
         let doc = docvec![];
 
         // @deprecated attribute
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -213,8 +213,9 @@ impl<'comments> Formatter<'comments> {
                 parameters: args,
                 type_ast: resolved_type,
                 public,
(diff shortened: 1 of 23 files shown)


lpil (commented):
Thank you! Though this seems to not deprecate the types themselves, only their constructors. Would you like to do that part as well?

omprakaash:
Oh I see. Yeah I would like to do that as well.  So in a program like this : 

```java 
@deprecated("This is deprecated ")
pub type Cat {
  Cat(name: String, cuteness: Int)
}

pub fn cat_name(arg1: Cat) {
  arg1.name
}

pub fn main() {
  let c = Cat(name: "numi", cuteness: 2001)
  io.println(cat_name(c))
}
``` 
the user would ideally get a deprecation warning in the function definition of cat_name and the first line of the main function/(wherever constructors are called) ?

compiler-core/src/format.rs
```diff
@@ -1083,6 +1093,8 @@ impl<'comments> Formatter<'comments> {
         commented(doc_comments.append(doc).group(), comments)
     }
 
+    // TODO : This takes too many arguments
+    #[allow(clippy::too_many_arguments)]
```
lpil, line 1097: Could you fix this to take the CustomType struct rather than each of its fields? Thank you.

compiler-core/src/type_/hydrator.rs
```diff
@@ -121,12 +121,23 @@ impl Hydrator {
...
                     ..
                 } = environment
                     .get_type_constructor(module, name, *location)
                     .map_err(|e| convert_get_type_constructor_error(e, location))?
                     .clone();
 
+                match deprecation {
+                    Deprecation::NotDeprecated => {}
+                    Deprecation::Deprecated { message } => {
+                        environment.warnings.emit(Warning::DeprecatedValue {
```
lpil, line 134: This results in a warning saying that the _value_ is deprecated, but it's a type. Could you make a `DeprecatedType` variant or make this `DeprecatedItem` and have it hold a `Layer` to indicate whether it's a value or a type.

compiler-core/src/format.rs
```diff
@@ -1091,9 +1103,23 @@ impl<'comments> Formatter<'comments> {
...
+
+        // @deprecated attribute
+        let doc = match deprecation {
+            Deprecation::NotDeprecated => doc,
+            Deprecation::Deprecated { message } => doc
+                .append("@deprecated(\"")
+                .append(message)
+                .append("\")")
+                .append(line()),
+        };
```
lpil, line 1119: Could you add a test for the formatting of this please? Thanks

lpil (commented):
This is really nice code! Super clean. Thank you.

I've left a few small notes inline, and would you be able to also add it for type aliases too?

compiler-core/src/format.rs
omprakaash: Yep will add a test for this.

compiler-core/src/type_/hydrator.rs
omprakaash: Oh Yeah. Will try making a DeprecatedItem with a Layer.

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -914,6 +914,65 @@ pub fn b() {
...
+        r#"
+@deprecated("Don't use this!")
+pub type Cat {
+    Cat(name: String)
+}
+
+pub type Dog = Cat
+
+pub fn b() {
+  let c = Dog("Bob")
```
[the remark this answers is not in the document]
omprakaash: Ah yeah, my bad. Will redo this test.

compiler-core/src/format.rs
omprakaash: Got you. Will change the signature to take in a struct instead.

compiler-core/src/format/tests.rs
```diff
@@ -5113,3 +5113,46 @@ fn empty_line_after_crash() {
...
+        r#"pub type Animal {
+  Cat(name: String)
+}
+
+@deprecated("Deprecated type")
+pub type Tiger =
+  Animal
+"#
+    );
+}
```
lpil, line 5158:
```suggestion
fn deprecated_custom_type() {
    assert_format!(
        r#"@deprecated("Deprecated type")
pub type One {
  One
}
"#
    );
}

#[test]
fn deprecated_type_alias() {
    assert_format_rewrite!(
        r#"@deprecated("Deprecated type")
pub type Tiger =
  Nil
"#
    );
}
```

lpil (approved):
Thank you!

lpil:
Thank you! Would you mind rebasing on main to resolve the conflicts? Thank you.

lpil (approved):
Thank you!

I think the metadata encoding is missing but that requires the capnp compiler so I'll do that.

Issue #2371, opened 2023-10-17
Different error message for list literals with wrong types


giacomocavalieri:
When defining a list literal with heterogeneous values the error message is not super useful:

```gleam
pub fn wrong_list() {
  [1, "2", 3]
}
```

We get this error message:

```
error: Type mismatch
  ┌─ /Users/giacomocavalieri/Desktop/prova/src/prova.gleam:8:3
  │
8 │   [1, "2", 3]
  │   ^^^^^^^^^^^

Expected type:

    Int

Found type:

    String
```

Having the entire list highlighted is a bit misleading, it looks like the compiler is expecting an `Int` instead of the whole list.

It would be more useful to get something like this:

```
error: Type mismatch
  ┌─ /Users/giacomocavalieri/Desktop/prova/src/prova.gleam:8:3
  │
8 │   [1, "2", 3]
  │       ^^^

Expected type:

    Int

Found type:

    String
```

In my opinion it would be even better to get something along the lines of Elm's error message explaining how lists can only have elements of one type:

```
error: Type mismatch
  ┌─ /Users/giacomocavalieri/Desktop/prova/src/prova.gleam:8:3
  │
8 │   [1, "2", 3]
  │       ^^^

All elements of a list should have the same type.
The 2nd element has an unexpected type:

Expected type:

    Int

Found type:

    String
```

lpil:
This is a bug! It should point to the specific element that is wrong.

I'm not convinced that the third error is worth the work to implement, but I would accept a PR for it.

lpil:
Heh, it turns out it's pretty easy to implement the error!

giacomocavalieri:
Nice! I can give that a try, I might have already figured out how to fix the span error :)

lpil:
Sorry, I did it!

PR #2372, opened 2023-10-17
fix mismatched tail error

giacomocavalieri:
There was a bug similar to the one in #2371 so I gave it a shot and tried to fix it!
I'm not really sold on the note in the error message, any idea on a better way to explain the problem?


lpil:
Perhaps something like this?

> All elements in a list must have the same type, but the elements of this list don't match the type of the elements being prepended to it.

giacomocavalieri:
Thanks! It's way better 😁

lpil (approved):
Thank you!

Issue #2374, opened 2023-10-19
Inconsistency in error messages for unknown field acces


giacomocavalieri:
There's a little discrepancy between two record field accesses depending on the place where it happens:

```gleam
import gleam/io

pub type Person {
  Person(name: String)
}

pub fn main() {
  let person = Person("Rob")
  io.println(person.id)
//           ^^^^^^^^^ Did you mean `name`?
  case True {
    _ if person.id -> todo
//             ^^^ Did you mean `name`?
  }
}
```

When accessing a wrong field in a case guard the error message highlights just the field access, if the same happens anywhere else the whole record is highlighted. I feel like the correct behaviour is the one shown in the guard access given the suggestion.
What do you think?

In any case, if that's the desired behaviour I may already have a fix ready 😁

giacomocavalieri:
Now that I think about it, a possible alternative would be to highlight just the field and not even the dot, like this:

```gleam
pub fn main() {
  let person = Person("Rob")
  io.println(person.id)
//                  ^^ Did you mean `name`?
  case True {
    _ if person.id -> todo
//              ^^ Did you mean `name`?
  }
}
```

@lpil what do you think would be the best behaviour here?

lpil:
Highlighting the field with or without the dot sounds good to me!

PR #2375, opened 2023-10-19
Fix span for error reporting in UnknownRecordField

giacomocavalieri:
This PR addresses #2374
(I've also made [a PR](https://github.com/gleam-lang/website/pull/282) to update the homepage's look accordingly!)
 
I decided to add another field in the field access data structure instead of changing the current `location` because it was used by a lot more of error messages that would then break. The new field holds a span for just the thing being accessed so error reporting can be more precise when giving hints instead of highlighting the whole record and field

### DIFF
--- a/compiler-core/src/ast/untyped.rs
+++ b/compiler-core/src/ast/untyped.rs
@@ -68,7 +68,14 @@ pub enum UntypedExpr {
     },
 
     FieldAccess {
+        // This is the location of the whole record and field
+        //   user.name
+        //   ^^^^^^^^^
         location: SrcSpan,
+        // This is the location of just the field access
+        //   user.name
+        //       ^^^^^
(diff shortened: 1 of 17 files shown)


compiler-core/src/ast/untyped.rs
```diff
@@ -68,7 +68,14 @@ pub enum UntypedExpr {
...
     FieldAccess {
-        location: SrcSpan,
+        // This is the location of the whole record and field
+        //   user.name
+        //   ^^^^^^^^^
+        location: SrcSpan,
+        // This is the location of just the field access
+        //   user.name
+        //       ^^^^^
+        access_location: SrcSpan,
```
lpil, line 78: Look at these beautiful comments! 💜

Issue #2380, opened 2023-10-25
Powershell for Erlang shipment entrypoint


bunopnu:
Currently, `gleam export erlang-shipment` generates a file named "entrypoint.sh" file, allowing us to run and create an Erlang shell for build.

To improve usability on Windows, we can provide a PowerShell version of the original script, for Windows version of Gleam.

PR #2381, opened 2023-10-25
Port `entrypoint.sh` for PowerShell

bunopnu:
addresses #2380

### DIFF
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -37,6 +37,22 @@ macro_rules! assert_infer {
 
 #[macro_export]
 macro_rules! assert_infer_with_module {
+    (
+        ($name1:expr, $module_src1:literal),
+        ($name2:expr, $module_src2:literal),
+        $src:expr, $module:expr $(,)?
+    ) => {
+        let constructors = $crate::type_::tests::infer_module(
+            $src,
+            vec![
+                ("thepackage", $name1, $module_src1),
+                ("thepackage", $name2, $module_src2),
+            ],
+        );
+        let expected = $crate::type_::tests::stringify_tuple_strs($module);
+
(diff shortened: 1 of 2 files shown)


lpil (commented):
Wonderful! Thank you!

There's a test for this here

https://github.com/gleam-lang/gleam/blob/91df6e02cb59c0510b7b0a799a9405a5de5d8bc3/.github/workflows/ci.yaml#L121-L127

Would you mind making this run on Windows too? And updating the CHANGELOG.md file.

bunopnu:
It appears that the [Windows Test](https://github.com/bunopnu/gleam/actions/runs/6657111450/job/18091133643) is encountering a failure, due to issues with the compilation of the `bcrypt` module. The easiest solution might be to replace `bcrypt` with a more straightforward alternative, and it seems that this is currently marked as a TODO.

I'd like to hear your thoughts, @lpil

lpil:
Bcrypt is deliberately there to ensure that nifs work, but I guess they do not on Windows! or at least not that one.

Having a minimal project that doesn't use a nif and testing that on windows would be good for now, until we work out what is wrong with the full one.

bunopnu:
I've made a copy of `project_erlang` called `project_erlang_windows`. The main differences are that I've updated the dependencies and left out bcrypt.

I think it is ok for now and can be removed after replacing `bcrypt` with something that supports all three major platforms (maybe something that uses rustler?)

lpil:
> I think it is ok for now and can be removed after replacing bcrypt with something that supports all three major platforms (maybe something that uses rustler?)

I don't want to use ruster, I want to use regular NIFs as this is a more useful and realistic test.

I didn't realise that library doesn't support Windows! That is very disappointing.

lpil (approved):
Thank you!!

lpil:
I have rebased these onto the main branch, thank you very much!

PR #2382, opened 2023-10-27
Replace `smol_str` with `ecow`

### LINKED ISSUE #2298: Switch from smol_str to ecow
https://github.com/typst/ecow

https://www.reddit.com/r/rust/comments/117ksvr/ecow_compact_cloneonwrite_vector_and_string/j9eh35d/

MarkusPettersson98:
Closes #2298

Have not done any benchmarks, so I don't know how big of a performance/memory gain this will actually result in.
Ran `make test`, and it seems to work on my machine :desktop_computer: 
Mostly a search & replace job, but it was fun anyway :) :purple_heart:


compiler-cli/src/dependencies.rs
```diff
@@ -298,7 +298,7 @@ impl LocalPackages {
             .collect();
         self.packages
             .iter()
-            .filter(|(n, v)| !manifest_packages.contains(&(&SmolStr::from(n), v)))
+            .filter(|(n, v)| !manifest_packages.contains(&(&EcoString::inline(n), v)))
```
lpil, line 301: The inline method will panic if the string is too large to be lined, so we can't use it with anything but a string literal.

I think we should always use the Into trait maybe?

lpil (commented):
Amazing! Thank you. I've just had a little review and left a note inline.

compiler-cli/src/dependencies.rs
MarkusPettersson98: Ah, good catch! 

Got rid of all uses of `EcoString::inline`, so please have a look at my last commit. If it looks good I will go ahead and squash it :blush:

lpil (approved):
Brilliant! Thank you!

I'm just going to hold off on merging this as I'm doing a release now :)

MarkusPettersson98:
Sounds reasonable! Keep up the good work on Gleam :purple_heart:

lpil (approved):
Thank you!!

Issue #2319, opened 2023-09-02
proposal: Support inlay hint feature


septemhill:
Hi gleam team,

I would like to suggest support `inlay hint` feature in vscode-gleam. 
This feature be supported in most of popular programming languages (e.g. Rust, Go, etc.) 
It would save us much time to hover over the variables/functions to check the types.

Looking forward to hearing from your reply !

inoas:
I like this IF it is something I can disable per project via vscode settings.

septemhill:
> Seeing as Gleam has labels does it still make sense to show the argument names in function calls?

Agree. For this case, I think we could ignore. But another question is, if the function without named arguments how do we display, may be just show the type ?

lpil:
Fab! Ok so the next steps are to determine how to send these hints via the language server protocol, and how to turn them on and off with the language server protocol.

inoas:
> > Are there any other places we might want them? The language server can put them anywhere.
> 
> This proposal is all I can think of now.
> 
> > Seeing as Gleam has labels does it still make sense to show the argument names in function calls?
> 
> Agree. For this case, I think we could ignore. But another question is, if the function without named arguments how do we display, may be just show the type ?
> 
> I understand we don't want to make whole UI with too much noise. So here I provide Go inlay-hint features in LSP settings. There're many hints in Go and we could decide which hint would be enabled in a language rather than a project. But the config could be created under `.vscode/setting` in different projects.
> 
> ![Screenshot 2023-09-04 at 9 13 21 AM](https://user-images.githubusercontent.com/1216949/265299767-ae3f5324-7ebf-4cb8-a90a-19afeadc2dbf.png)
> 
> Thanks.

would we want the comments around it, i think the rust lsp skips those. just grey and some opacity, and maybe (not sure) you cannot select them ever with the cursor

lpil:
We don't have an inline comment syntax so we couldn't wrap them like Go does.

jamesbirtles:
I'm currently working on this https://github.com/jamesbirtles/gleam/tree/feat/inlay-hints

lpil:
Very exciting!

Let's start with the MVP feature-set and we can add more over time. That way it is easier to review and merge.

lpil:
https://github.com/gleam-lang/gleam/pull/2393

GearsDatapacks:
Also another thing inlay hints could be used for is inferred types for arguments and return-types, since those might not be obvious in certain cases

[...]

PR #2393, opened 2023-11-07
Add inlay hints for function definitions and module constants

jamesbirtles:
- Add inlay hints for function definitions and module constants
- Add config options to enable/disable specific inlay hints
  - Before now the language server had no config so this this PR adds the requesting of the config, as well as listening for config updates.
  - These will need to be exposed by the vscode plugin which I can put a PR in for too
- Adds code actions to annotate types for function definitions and module constants
- This PR also expands the 'location' on module constants to include the `(pub) const` tokens to make it easier to target a module constant with a code action as, without this change, the cursor would need to specifically be on the name to trigger it.
- The code for the code action and inlay hints are similar enough that it made sense to share some code between them. This is done with the `TypeAnnotations` struct which once created can be turned into either a code action or inlay hint.

I have made the inlay hints default to being disabled as I think the consensus is that they may add too much noise and also, particularly for function definitions, may undesirably discourage annotating top level types.

This PR does not yet address all the use cases for inlay hints as presented in #2319 
Specifically called out in that issue are function calls and variable assignments. These are a little more involved to support than what this PR adds support for, due to needing to walk the ast to find these nodes within function bodies. A subsequent PR may add support for these.

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -22,7 +22,10 @@ use std::sync::Arc;
 use strum::IntoEnumIterator;
 
 use super::{
-    code_action::CodeActionBuilder, src_span_to_lsp_range, type_annotations::TypeAnnotations,
+    code_action::CodeActionBuilder,
+    configuration::{InlayHintsConfig, VersionedConfig},
+    src_span_to_lsp_range,
+    type_annotations::TypeAnnotations,
     DownloadDependencies, MakeLocker,
 };
 
@@ -57,6 +60,8 @@ pub struct LanguageServerEngine<IO, Reporter> {
     // Used to publish progress notifications to the client without waiting for
     // the usual request-response loop.
     progress_reporter: Reporter,
+
+    pub(crate) user_config: VersionedConfig,
 }
 
 impl<'a, IO, Reporter> LanguageServerEngine<IO, Reporter>
@@ -76,6 +81,7 @@ where
         progress_reporter: Reporter,
         io: FileSystemProxy<IO>,
         paths: ProjectPaths,
+        user_config: VersionedConfig,
     ) -> Result<Self> {
         let locker = io.inner().make_locker(&paths, config.target)?;
 
@@ -97,6 +103,7 @@ where
             progress_reporter,
             compiler,
             paths,
+            user_config,
         })
     }
 
@@ -417,7 +424,12 @@ where
             let line_numbers = LineNumbers::new(&module.code);
 
             let mut hints = vec![];
-            add_hints_for_definitions(&module.ast.definitions, &line_numbers, &mut hints);
+            add_hints_for_definitions(
+                &this.user_config.inlay_hints,
+                &module.ast.definitions,
+                &line_numbers,
+                &mut hints,
+            );
 
             Ok(Some(hints))
         })
@@ -661,17 +673,18 @@ fn code_action_annotate_types(
 }
 
 fn add_hints_for_definitions(
+    config: &InlayHintsConfig,
     definitions: &[TypedDefinition],
     line_numbers: &LineNumbers,
     hints: &mut Vec<InlayHint>,
 ) {
     for def in definitions {
         match def {
-            Definition::Function(function) => hints.extend(
+            Definition::Function(function) if config.function_definitions => hints.extend(
                 TypeAnnotations::from_function_definition(function, line_numbers)
                     .into_inlay_hints(),
             ),
-            Definition::ModuleConstant(constant) => hints.extend(
+            Definition::ModuleConstant(constant) if config.module_constants => hints.extend(
                 TypeAnnotations::from_module_constant(constant, line_numbers).into_inlay_hints(),
             ),
             _ => {}
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -15,7 +15,7 @@ use crate::{
 };
 use camino::Utf8PathBuf;
 use ecow::EcoString;
-use lsp::CodeAction;
+use lsp::{CodeAction, InlayHint, Position};
 use lsp_types::{self as lsp, Hover, HoverContents, MarkedString, Url};
 use std::sync::Arc;
 use strum::IntoEnumIterator;
@@ -229,6 +229,7 @@ where
             };
 
             code_action_unused_imports(module, &params, &mut actions);
+            code_action_annotate_types(module, &params, &mut actions);
 
             Ok(if actions.is_empty() {
                 None
@@ -282,24 +283,12 @@ where
         })
     }
 
-    fn module_node_at_position(
-        &self,
-        params: &lsp::TextDocumentPositionParams,
-        module: &'a Module,
-    ) -> Option<(LineNumbers, Located<'a>)> {
-        let line_numbers = LineNumbers::new(&module.code);
-        let byte_index = line_numbers.byte_index(params.position.line, params.position.character);
-        let node = module.find_node(byte_index);
-        let node = node?;
-        Some((line_numbers, node))
-    }
-
     fn node_at_position(
         &self,
         params: &lsp::TextDocumentPositionParams,
     ) -> Option<(LineNumbers, Located<'_>)> {
         let module = self.module_for_uri(&params.text_document.uri)?;
-        self.module_node_at_position(params, module)
+        module_node_at_position(&params.position, module)
     }
 
     fn module_for_uri(&self, uri: &Url) -> Option<&Module> {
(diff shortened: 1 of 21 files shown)


jamesbirtles:
ci failure is unrelated, should be fixed by https://github.com/gleam-lang/gleam/pull/2394

compiler-core/src/language_server/engine.rs
```diff
@@ -417,6 +415,44 @@ where
...
+            let start =
+                line_numbers.byte_index(params.range.start.line, params.range.start.character);
+            let end = line_numbers.byte_index(params.range.end.line, params.range.end.character);
+            add_hints_for_definitions(
+                &this.user_config.inlay_hints,
+                module.ast.definitions.iter().filter(|stmt| {
+                    let loc = stmt.location();
+                    (loc.start > start && loc.start < end) // Is the start of the node in the range
+                        || (loc.end > start && loc.end < end) // or is the end of the node in the range
+                        || (start > loc.start && end < loc.end) // or is range within the node
```
lpil, line 436: Could we extract this to a `fn overlaps(self, other: SrcSpan) -> bool` or such please?

compiler-core/src/language_server/engine.rs
```diff
@@ -417,6 +415,44 @@ where
...
+        self.respond(|this| {
+            let Some(module) = this.module_for_uri(&params.text_document.uri) else {
+                return Ok(None);
+            };
+            let line_numbers = LineNumbers::new(&module.code);
+
+            let mut hints = vec![];
+            let start =
+                line_numbers.byte_index(params.range.start.line, params.range.start.character);
+            let end = line_numbers.byte_index(params.range.end.line, params.range.end.character);
```
lpil, line 429: How about something like this:

```suggestion
            let selected = line_numbers.src_span(params.range);
```

compiler-core/src/language_server/engine.rs
```diff
@@ -417,6 +415,44 @@ where
 
         completions
     }
+
+    pub fn inlay_hint(&mut self, params: lsp::InlayHintParams) -> Response<Option<Vec<InlayHint>>> {
```
lpil, line 419: What's the difference between `Ok(None)` and `Ok(Some(vec![]))`?

It seems they are semantically the same, so if that's the case let's remove the option as the engine class is intended to be a simpler API rather than a copy of the protocol.

compiler-core/src/language_server/type_annotations.rs
```diff
@@ -0,0 +1,92 @@
...
+                tooltip: None,
+                padding_left: text.starts_with(' ').then_some(true),
+                padding_right: None,
+                data: None,
+                label: InlayHintLabel::String(text.trim_start().to_owned()),
+                text_edits: Some(vec![TextEdit::new(Range::new(position, position), text)]),
+            })
+    }
+
+    pub fn into_code_action(self, uri: Url) -> Option<CodeActionBuilder> {
```
lpil, line 72: This isn't always used so let's take it by reference, rather than force the clone.
```suggestion
    pub fn into_code_action(self, uri: &Url) -> Option<CodeActionBuilder> {
```

compiler-core/src/language_server/type_annotations.rs
```diff
@@ -0,0 +1,92 @@
...
+            Self::default()
+        }
+    }
+
+    pub fn from_function_definition(function: &TypedFunction, line_numbers: &LineNumbers) -> Self {
+        let mut annotations = vec![];
+        if function.return_annotation.is_none() {
+            let linecol = line_numbers.line_and_column_number(function.location.end);
+            let position = Position::new(linecol.line - 1, linecol.column - 1);
+            let type_text = Printer::new().pretty_print(&function.return_type, 0);
```
lpil, line 40: I think this will generate invalid code when the type is imported in a qualified fashion, when there are type parameters and they have already been specified for some arguments, or when there are multiple types in the type with the same name.

Could we have tests for these please?

Seeing as we don't care about formatting etc I think it may be better to have a dedicated type printer for this use case rather than reusing the UI pretty printer. It'll be simpler than teaching the pretty printer about scopes, having to track type variable names, and having to conditionally disable the other features of the pretty printer.

compiler-core/src/language_server/tests/action.rs
```diff
@@ -78,18 +82,22 @@ fn apply_code_edit(
...
-                line_numbers.byte_index(edit.range.end.line, edit.range.end.character) - offset;
+            let start = line_numbers.byte_index(edit.range.start.line, edit.range.start.character)
+                as i32
+                + offset;
+            let end = line_numbers.byte_index(edit.range.end.line, edit.range.end.character) as i32
+                + offset;
             let range = (start as usize)..(end as usize);
-            offset += end - start;
+            offset -= end - start;
+            offset += edit.new_text.len() as i32;
```
lpil, line 100: What are these changes for? I couldn't figure it out.

compiler-core/src/language_server/server.rs
```diff
@@ -181,12 +205,57 @@ where
...
+                                version: self.config.version + 1,
+                                config,
+                            };
+                        }
+                    }
+                    Err(err) => eprintln!("unable to parse configuration: {err}"),
+                }
+            }
+        }
+    }
```
lpil, line 235: I find this function hard to read, could you rewrite it without the nesting please?

compiler-core/src/language_server/server.rs
```diff
@@ -181,12 +205,57 @@ where
...
+                match serde_json::from_value::<Vec<Configuration>>(result) {
+                    Ok(result) => {
+                        if let Some(config) = result.into_iter().next() {
+                            self.config = VersionedConfig {
+                                version: self.config.version + 1,
+                                config,
+                            };
+                        }
+                    }
+                    Err(err) => eprintln!("unable to parse configuration: {err}"),
```
lpil, line 231: This silently discards the problem from the user's point of view, we should surface it to them somehow. Crashing would be the easiest way, but an actual error message would be better. I'm happy with either.

compiler-core/src/language_server/server.rs
```diff
@@ -181,12 +205,57 @@ where
...
     }
 
+    fn handle_response(&mut self, handler: ResponseHandler, result: Json) {
+        match handler {
+            ResponseHandler::UpdateConfiguration => {
+                match serde_json::from_value::<Vec<Configuration>>(result) {
+                    Ok(result) => {
+                        if let Some(config) = result.into_iter().next() {
+                            self.config = VersionedConfig {
+                                version: self.config.version + 1,
```
lpil, line 226: What's the version thing here?

lpil (commented):
Wonderful work! Thank you very much. I've left some comments inline with various bits of feedback and questions.

If you could liberally add comments describing both what code does and the reasoning behind it that would be much appreciated, and the changelog file will need to be updated before merging, but that can come last as it tends to result in merge conflicts.

Thank you!

compiler-core/src/language_server/server.rs
jamesbirtles: Its used to quickly check if the config cached in the engine is out of date. Its checked in the `respond_with_engine` function to lazily update the config if the versions don't match. An alternative might be to eagerly loop through the engines and update the cached configs, or I suppose share the config with an Arc RwLock or similar. I don't have a particular preference here

compiler-core/src/language_server/tests/action.rs
jamesbirtles: I have added some comments to the code to help explain what its attempting to do.

The reason these changes were needed is that the previous code only worked on text edits that removed text. The offset was used to update to locations in subsequent edits in the same batch. As the annotate types action adds text, a solely negative offset no longer worked.

compiler-core/src/language_server/type_annotations.rs
jamesbirtles: Could you give me some code examples of the cases you listed?

compiler-core/src/language_server/server.rs
lpil: Oh cool! Could we use the RwLock approach? That sounds nice to me.

compiler-core/src/language_server/tests/action.rs
lpil: Fab, thank you

compiler-core/src/language_server/type_annotations.rs
lpil:
```gleam
pub identity(value: some_type) {
  value
}
```
```gleam
```gleam
pub equal(v1: some_type, v2) {
  v1 == v2
}
```
```gleam
import mod1 as a
import mod2 as b

pub fn main() {
  let x: a.Value
  let y: b.Value
  #(x, y)
}
```
```gleam
import mod1.{Value as V1}
import mod2.{Value as V2}

pub fn main() {
  let x: V1
  let y: V2
  #(x, y)
}
```

lpil:
Hello! What's the status of this work? Is it ready for a review?

lpil:
Great, thank you.

I'm afraid it'll need to work for all types, or more accurately, the work to identify which types can be printed without modification is pretty much the same work as printing them correctly, so if that was done you'd be almost finished.

jamesbirtles:
unfortunately I don't know when i'll get time to pick this back up, closing this PR for now. Happy for someone else to pick up where I left of if they so desire.

omprakaash:
Hey can I continue working on this ?

PR #2394, opened 2023-11-07
removed unused import as flagged by clippy

jamesbirtles:
import was only being used in tests so moved the tests to a test module and imported it there


lpil:
Sorry, I have already done this!

jamesbirtles:
> Sorry, I have already done this!

I don't think your fix was quite right as the import _is_ used in tests so they fail to compile if its removed

jamesbirtles:
Looks like you got it, I'll update my other PR now :)

PR #2406, opened 2023-11-14
also accept js and erl for --target

massivefermion:


lpil (approved):
Thanks! Could you update the changelog please 🙏

lpil (approved):
Thank you

PR #2407, opened 2023-11-16
when an imported type is used, the alias is used

massivefermion:
I was writing some gleam and I came across this bug:
```gleam
import a/foo as bar

pub fn do_sth(a: bar.A) {
  a
}
```
This would emit a warning that the alias `bar` is not used, which is incorrect.
This is the fix.

P.S.
I had to use `assert_warnings_with_imports` for the test because `assert_no_warnings!` was not failing without the fix!


compiler-core/src/type_/tests/warnings.rs
```diff
@@ -1138,6 +1138,21 @@ fn unused_alias_warning_test() {
     );
 }
 
+#[test]
+fn used_type_with_import_alias_no_warning_test() {
+    assert_warnings_with_imports!(
+        ("gleam/foo", "pub type A");
+        r#"
+            import gleam/foo as bar
+            fn fun(a: bar.A) { a }
```
lpil, line 1147: Could you make this function public to remove the other warning please? It's not very clear to me what this is testing currently.

massivefermion: >  I had to use assert_warnings_with_imports for the test because assert_no_warnings! was not failing without the fix!

This is the issue, I had to do this so the test would fail without the fix!

lpil: oh! How come it wasn't failing? 🤔

massivefermion: The weird thing is, even if the module name in the first argument and the one imported in the second argument don't match, it still does not complain. It only started complaining when there was syntax errors I think, but it seems like there are a few issues that `assert_no_warnings!` simply ignores!

lpil: We will need to use `assert_no_warnings` as this test is unclear.

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -1140,16 +1140,12 @@ fn unused_alias_warning_test() {
 
 #[test]
 fn used_type_with_import_alias_no_warning_test() {
-    assert_warnings_with_imports!(
-        ("gleam/foo", "pub type A");
+    assert_no_warnings!(
+        ("gleam", "foo", "pub type A"),
```
lpil, line 1144: Here the module is called `foo` but the test is importing `gleam/foo`, so I think the is a compile error rather than a test that replicates the problem. If the fix is commented out does this test fail?

massivefermion: My bad, I thought the "gleam" in `("gleam", "foo", "pub type A")` is part of the import path.
I fixed the this test and one other place where I made the same mistake. Now the test fails without the fix.

lpil (approved):
Thank you! Would you mind updating the changelog file also?

lpil (approved):
Thank you!

PR #2411, opened 2023-11-19
Allow as to take an expression when used together with panic

omprakaash:
Allow as to take an expression when used together with panic. Expression given should evaluate to a string.

### DIFF
--- a/compiler-core/src/erlang/tests/panic.rs
+++ b/compiler-core/src/erlang/tests/panic.rs
@@ -22,6 +22,20 @@ pub fn main() {
     );
 }
 
+#[test]
+fn panic_as_function() {
+    assert_erl!(
+        r#"
+pub fn retstring() {
+  "wibble"
+}
+pub fn main() {
+  panic as retstring() <> "wobble"
+}
+"#
+    );
+}
+
 // https://github.com/gleam-lang/gleam/issues/2176
 #[test]
 fn piped() {
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -3158,6 +3158,18 @@ fn expr_panic_as() {
 }
 
 #[test]
+fn expr_panic_as_value(){
+    assert_format!(
+        r#"fn main() {
+  let x = "panicking" <> "with a value"
+  panic as x
+}
+"#
+    );
+}
+
+
+#[test]
 fn expr_todo() {
     assert_format!(
         "fn main() {
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -416,9 +416,10 @@ where
                 let mut label = None;
                 let _ = self.next_tok();
                 if self.maybe_one(&Token::As).is_some() {
-                    let (_, l, e) = self.expect_string()?;
-                    label = Some(l);
-                    end = e;
+                    label = self.parse_expression()?.map(Box::new);
+                    if let Some(msg_expr_val) = &label {
+                        end = msg_expr_val.location().end;
+                    }
                 }
                 UntypedExpr::Panic {
                     location: SrcSpan { start, end },
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -76,11 +76,13 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             // body, instead only giving an external implementation for this
             // target. This placeholder implementation will never be used so we
             // treat it as a `panic` expression during analysis.
-            UntypedExpr::Placeholder { location } => Ok(self.infer_panic(location, None)),
+            UntypedExpr::Placeholder { location } => {
+                self.infer_panic(location, None)
+            }
 
             UntypedExpr::Panic {
                 location, message, ..
-            } => Ok(self.infer_panic(location, message)),
+            } => self.infer_panic(location, message),
 
             UntypedExpr::Var { location, name, .. } => self.infer_var(name, location),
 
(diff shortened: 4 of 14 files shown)


compiler-core/src/format/tests.rs
```diff
@@ -3157,6 +3157,17 @@ fn expr_panic_as() {
...
+#[test]
+fn expr_panic_as_value() {
+    assert_format!(
+        r#"fn main() {
+  let x = "panicking" <> "with a value"
+  panic as x
+}
+"#
+    );
+}
```
lpil, line 3169: Could we have Erlang and JS tests for `todo as`, and tests for the formatter too please? Thank you

compiler-core/src/type_/expression.rs
```diff
@@ -202,12 +202,30 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    type_: typ,
+                    message: Some(Box::new(message)),
+                })
+            }
+            None => Ok(TypedExpr::Panic {
+                location,
+                type_: typ,
+                message: None,
+            }),
+            Some(Err(e)) => Err(e),
```
lpil, line 228: I'm finding this a bit hard to read- could it be re-written without using `.map`? I think this would be clearer if it's used with just pattern matching as it means that `?` can be used directly and `Some(Err(e))` won't need to be handled.

```rust
let message = match message {
   ....
};
Ok(TypedExpr::Panic { ..... })
```

lpil (commented):
Fab! Thank you. I've left a few comments inline, and once it is ready the changelog file will need updating too.

compiler-core/src/parse.rs
```diff
@@ -416,9 +416,10 @@ where
                 let mut label = None;
                 let _ = self.next_tok();
                 if self.maybe_one(&Token::As).is_some() {
-                    let (_, l, e) = self.expect_string()?;
-                    label = Some(l);
-                    end = e;
+                    label = self.parse_expression()?.map(Box::new);
+                    if let Some(msg_expr_val) = &label {
+                        end = msg_expr_val.location().end;
+                    }
```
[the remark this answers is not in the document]
omprakaash: Oh yeah. Missed that. Will change in the next commit

compiler-core/src/format/tests.rs
omprakaash: Sure. Will add more tests.

compiler-core/src/type_/expression.rs
omprakaash: Gotcha. Will try making this part easy to read.

lpil:
Thank you

compiler-core/src/parse.rs
```diff
@@ -401,9 +401,9 @@ where
                 let _ = self.next_tok();
                 let mut message = None;
                 if self.maybe_one(&Token::As).is_some() {
-                    let (_, l, e) = self.expect_string()?;
-                    message = Some(l);
-                    end = e;
+                    let msg_expr = self.expect_expression()?;
```
lpil, line 404: Parsing an expression unit will do the trick here for the parsing bug

compiler-core/src/erlang/tests/todo.rs
```diff
@@ -33,6 +33,21 @@ pub fn main() {
...
+  "wibble"
+}
+pub fn main() {
+  todo as retstring() <> "wobble"
+}
+"#
+    );
+}
+
+#[ignore = "This would fail now. See https://github.com/gleam-lang/gleam/issues/2440"]
```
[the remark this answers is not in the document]
omprakaash: Oh no worries. Will fix this.

compiler-core/src/parse.rs
omprakaash: Wouldn't parsing an expression unit in this context disable string concatenation ? 
eg: ```panic as "Hello" <> " There"  ``` or  ``` panic as var1 <> var2 ```

lpil: It would be written like this:

```gleam
panic as { "Hello " <> name }

Comment (omprakaash) [reply compiler-core/src/parse.rs]: Oh yeah ! Thank you.

File: compiler-core/src/erlang/tests/panic.rs [hunk 6]
```diff
@@ -22,15 +22,29 @@ pub fn main() {
     );
 }
 
-// https://github.com/gleam-lang/gleam/issues/2176
 #[test]
-fn piped() {
```
lpil, line 27: Please put back this function and comment that have been deleted.

compiler-core/src/parse.rs
```diff
@@ -2509,6 +2509,14 @@ where
         }
     }
 
+    fn expect_expression_unit(&mut self) -> Result<UntypedExpr, ParseError> {
+        if let Some(e) = self.parse_expression_unit()? {
+            Ok(e)
+        } else {
+            self.next_tok_unexpected(vec!["An expression unit".into()])
```
lpil, line 2516:
```suggestion
            self.next_tok_unexpected(vec!["An expression".into()])
```

"expression unit" is a parser implementation detail, it doesn't mean anything to a Gleam user.

compiler-core/src/erlang/tests/panic.rs
omprakaash: Got it. Sorry, thought that the comment was referring to a issue that was closed.

compiler-core/src/parse.rs
omprakaash: Gotcha, was wondering about this. Will change it to just expression.

compiler-core/src/erlang/tests/panic.rs
lpil: Yes, it's a test to prevent a regression on that bug.

lpil (approved):
Thank you!

PR #2425, opened 2023-11-27
Fix a bug with case branch formatting

giacomocavalieri:
There was a little bug I overlooked in the formatter.
It would format case branches like this:
```gleam
case x {
  1 -> some_module.some_long_name_function([
      some_module.some_long_name_function(),
    ])
  _ -> todo
}
```
Now it works correctly and does this:
```gleam
case x {
  1 ->
    some_module.some_long_name_function([
      some_module.some_long_name_function(),
    ])
  _ -> todo
}
```


lpil (approved):
💜

PR #2429, opened 2023-11-28
Fix a bug where the formatter would move comments out of blocks

### LINKED ISSUE #2428: Formatter moves comments out of blocks
The formatter doesn't preserve this code
```gleam
pub fn main() {
  {
    // foo
    2
  }
}
```
and turns it into
```gleam
pub fn main() {
  // foo
  { 2 }
}
```

giacomocavalieri:
Closes #2428

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Formatter
 
 - The formatter now tries to keep a function body and its arguments on a single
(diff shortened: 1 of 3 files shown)


giacomocavalieri:
Oops I just noticed that there already was a fix for this on the main branch, sorry!

PR #2432, opened 2023-11-28
Fix a bug where the formatter would move comments out of blocks

giacomocavalieri:
Before the formatter would yeet comments out of blocks:
```gleam
{
  1
  // I'm a comment
}
```
would turn into:
```gleam
{ 1 }
// I'm a comment
```

Now it's a bit more respectful of comments inside of blocks (even case clause blocks!)
```gleam
case foo {
  1 -> {
    1
    // I'm a comment
  }
}
```
no longer turns into:
```gleam
case foo {
  1 -> {
    1
  }
}
// I'm a comment
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,8 @@
 ### Formatter
 
 - The formatter now tries to keep a function body and its arguments on a single
-  line by first trying to split only its last argument on multiple lines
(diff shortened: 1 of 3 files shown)


lpil (approved):
Wonderful, thank you!

Issue #2434, opened 2023-11-29
Ignore `build` dir by default in `gleam format` (`--check`)


tynanbe:
Currently, ignoring `build` relies on `.gitignore`.
The `build` dir should by default be excluded from the `gleam format` and `gleam format --check` commands.

lpil:
Thank you

PR #2445, opened 2023-12-03
#2434 Removed build from `gleam format`

ascandone:
Fixed https://github.com/gleam-lang/gleam/issues/2434 (ignore the `./build` folder when running `gleam format`)

**Note**: this PR also makes the `./build` folder be ignored by the `gleam fix` and `gleam publish` commands


lpil (commented):
Thank you! This will need to be any build directory that is in a directory with a `gleam.toml` as in other directories it could be a regular directory with that name rather than the build directory of a Gleam project.

compiler-cli/src/fs.rs
```diff
@@ -342,10 +342,34 @@ fn is_gleam_path(path: &Utf8Path, dir: impl AsRef<Utf8Path>) -> bool {
...
+    if !e.path().is_dir() || !e.path().ends_with("build") {
+        return Some(false);
+    }
+
+    let has_gleam_toml_as_sibiling = e
+        .path()
+        .parent()?
+        .read_dir()
+        .ok()?
+        .any(|e| e.unwrap().file_name().to_str() == Some("gleam.toml"));
```
lpil, line 355: This will iterate over the whole directory again. We can avoid that work by doing something like this pseudocode:

```rust
e.path().parent().join("gleam.toml").is_file()
```

compiler-cli/src/fs.rs
```diff
@@ -342,10 +342,34 @@ fn is_gleam_path(path: &Utf8Path, dir: impl AsRef<Utf8Path>) -> bool {
     )
 }
 
+fn is_gleam_build_dir(e: &ignore::DirEntry) -> Option<bool> {
```
lpil, line 345: Let's just return a bool here and assume that there's no gleam.toml if we can't read it.

lpil (commented):
Nice!! I've left a couple comments inline

compiler-cli/src/fs.rs
ascandone: Thanks! I missed that :)

lpil (approved):
Thank you! I'll update the changelog

PR #2447, opened 2023-12-04
Go to docs for functions

ascandone:
Implements https://github.com/gleam-lang/gleam/issues/2454

This PR provides a "View in hexdocs" link when hovering over public symbols from external packages
Currently only working for functions and constants (no records)

As an example, in the following file:
```gleam
import gleam/dict

pub fn f() {
  dict.delete
}
```

Hovering on `dict.delete` looks like:

Screenshot 2023-12-04 at 14 30 42

Implementation works on both qualified and unqualified imports

**current limitations:** the "view on hexdocs" link does not work on records.


lpil:
Hi @ascandone ! Have you linked the right issue? That one is for adding links in the generated HTML documentation.

ascandone:
> Hi @ascandone ! Have you linked the right issue? That one is for adding links in the generated HTML documentation.

That's true! I'll remove the issue mention from the PR description
Can the PR still be useful? It provides same functionality as elixir's "View in Hexdocs" link on hover

lpil:
It could be useful but we'd need to make sure to only show a link if the code exists on hexdocs. Do you have any thoughts on how the LSP would know if it is on there or not?

ascandone:
> It could be useful but we'd need to make sure to only show a link if the code exists on hexdocs. Do you have any thoughts on how the LSP would know if it is on there or not?

The way Elixir behaves is that this link is always shown, regardless of whether the package was published.
I don't think is doable to check if the module you are writing was published easily, however if we want to make sure there aren't any 404 links I could try to only show the link for external dependencies

lpil:
We definitely do not want to copy what that Elixir language server does as clicking a link and getting an error is very poor UX. We will need to only show links when the documentation is on Hex. Let's discuss how to do this in an issue and then it can be implemented once we have determined a suitable design.

ascandone:
> We definitely do not want to copy what that Elixir language server does as clicking a link and getting an error is very poor UX. We will need to only show links when the documentation is on Hex. Let's discuss how to do this in an issue and then it can be implemented once we have determined a suitable design.

That makes sense! I created an issue here: https://github.com/gleam-lang/gleam/issues/2454

compiler-core/src/language_server/engine.rs
```diff
@@ -613,3 +640,57 @@ fn code_action_unused_imports(
...
+        Definition::Import(p)
+            if p.module == module_name && hex_deps.contains(&p.package.to_string()) =>
+        {
+            Some(&p.package)
+        }
+        _ => None,
+    })?;
+
+    let link = format!("https://hexdocs.pm/{package_name}/{module_name}.html#{name}");
+    Some(format!("\nView on [hexdocs]({link})"))
```
lpil, line 695:
```suggestion
    Some(format!("\nView on [HexDocs]({link})"))
```

compiler-core/src/language_server/tests/mod.rs
```diff
@@ -82,6 +82,21 @@ impl LanguageServerTestIO {
...
+    pub fn hex_dep_module(&self, dep: &str, name: &str, code: &str) -> Utf8PathBuf {
+        let path = self
+            .paths
+            .root()
+            .join("build")
+            .join("packages")
+            .join(dep)
+            .join("src")
+            .join(name)
+            .with_extension("gleam");
```
lpil, line 94: Please use the method on the path class rather than duplicating this logic here.

compiler-core/src/language_server/tests/mod.rs
```diff
@@ -254,6 +269,44 @@ impl ProgressReporter for LanguageServerTestIO {
...
+        },
+    );
+
+    let toml_path = engine
+        .paths
+        .root()
+        .join("build")
+        .join("packages")
+        .join(name)
+        .join("gleam.toml");
```
lpil, line 300: Please use the method on the path class rather than duplicating this logic here.

compiler-core/src/language_server/tests/mod.rs
```diff
@@ -82,6 +82,21 @@ impl LanguageServerTestIO {
         path
     }
 
+    pub fn hex_dep_module(&self, dep: &str, name: &str, code: &str) -> Utf8PathBuf {
```
lpil, line 85: Could you rename `dep_module` to `path_dep_module` to make it clear the difference please.

compiler-core/src/language_server/tests/hover.rs
```diff
@@ -65,6 +76,269 @@ fn(Int) -> Int
...
+}
+
+fn main() {
+  my_fn
+}
+";
+
+    assert_eq!(
+        positioned_hover(code, Position::new(6, 3)),
+        Some(Hover {
```
lpil, line 93: It would be good to use snapshot tests with `cargo-insta` here rather than typing out the returned value, to match the majority of tests in the codebase and to reduce the amount of work maintaining the tests.

compiler-core/src/language_server/tests/hover.rs
```diff
@@ -65,6 +76,269 @@ fn(Int) -> Int
     );
 }
 
+#[test]
+fn hover_local_function() {
```
lpil, line 80: Could we get tests for different types of imports please?

For example, nested modules, aliased modules, discarded modules that would have the same name, and anything else you can think of.

I think there may be a bug in finding the right import and I would like full test coverage to verify that.

compiler-core/src/language_server/engine.rs
```diff
@@ -613,3 +640,57 @@ fn code_action_unused_imports(
         .preferred(true)
         .push_to(actions);
 }
+
+fn get_expr_qualified_name(expression: &TypedExpr) -> Option<(&EcoString, &EcoString)> {
```
lpil, line 644: What is `module_name` here? Is it the name or the alias of the imported module? I'm not totally sure but I think this function mixes them up.

ascandone:
Sorry about the stale PR! I've been unable to have access to the computer for few days and I'll be able to implement the suggestion later this week

compiler-core/src/language_server/engine.rs
```diff
@@ -55,6 +55,8 @@ pub struct LanguageServerEngine<IO, Reporter> {
     // Used to publish progress notifications to the client without waiting for
     // the usual request-response loop.
     progress_reporter: Reporter,
+
+    hex_deps: std::collections::HashSet<String>,
```
lpil, line 59: Could you document this with a comment and use a `EcoString` here please? We don't use `String` other than with library that force it.

compiler-core/src/language_server/engine.rs
```diff
@@ -95,6 +97,7 @@ where
             progress_reporter,
             compiler,
             paths,
+            hex_deps: std::collections::HashSet::new(),
```
lpil, line 100: The hex deps should be set immediately here as what they are is known even before compilation.

lpil (commented):
Wonderful! Thank you very much 💜

compiler-core/src/language_server/tests/hover.rs
ascandone: You were right! A function marked with the `@external` modifier was not behaving properly

https://github.com/gleam-lang/gleam/blob/4c436129cf3e86bd4d3aede644111582a5a2ee2a/compiler-core/src/language_server/tests/hover.rs#L223

Other than that, I tried to test as many edge cases as possible

compiler-core/src/language_server/engine.rs
ascandone: it should be the original name... there are tests to confirm that

ascandone:
I think I fixed everything!
I've converted the tests I wrote to snaèshot test, would you like me to convert other existing tests in the same module to snapshot tests too?

The only thing I'm not sure about is whether to set the `hex_deps` property once actually works in case you add another dependency, however I tested it locally and it looks like it does. Could there be any edge cases?

lpil (approved):
Thank you!

lpil:
Would you mind rebasing on main to remove the merge commit? Thank you.

If you could also update the change log that would be fab!

ascandone:
> Would you mind rebasing on main to remove the merge commit? Thank you.
> 
> If you could also update the change log that would be fab!

Done :+1:
Let me know if the description is clear enough

lpil (approved):
Thank you

PR #2449, opened 2023-12-04
Add leading wildcard to doc searches

fschwalbe:
Previously, searches only matched the beginning of a word. For example searching for "shift" showed no results, while it should clearly be able to find `int.bitwise_shift_(left/right)`.

I'm not sure if the chosen solution is the best way or if the boost values are optimal, so please do give feedback and suggestions for improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@
   patterns.
 - Fixed a bug where the formatter would misplace comments at the start of a
   block.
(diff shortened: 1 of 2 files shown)


lpil (commented):
Brilliant, thank you. Could you update the changelog file too? Thank you.

lpil (approved):
Thank you

Issue #2448, opened 2023-12-04
Absolute path for local dependency in gleam.toml always resolves dependencies


erikareads:
## Problem Statement

Having an absolute path for a local dependency leads to Gleam resolving dependencies on every run. This hits the Hex API and can trigger rate limits.

## Expected Behaviour

An absolute path for a local dependency should behave the same as a relative path for a local dependency.

## Reproduction

```
gleam new local_package
gleam new use_local_package
cd use_local_package
```

Add to `gleam.toml`

```toml
local_package = { path = "/absolute/path/to/local_package" }
```

```
gleam run
```

## Additional information

Here is the `GLEAM_LOG=trace gleam run` output truncated to the relevant part:

```
DEBUG download_deps: locking_build_directory path="/home/erika/projects/gleam/use_local_package/build/packages"
TRACE download_deps: reading_file path="/home/erika/projects/gleam/use_local_package/gleam.toml"
DEBUG download_deps: reading_manifest_toml
TRACE download_deps: reading_file path="/home/erika/projects/gleam/use_local_package/manifest.toml"
DEBUG download_deps: manifest_outdated
INFO download_deps: unlocking_stale_package name="local_package"
```

By comparison with a relative path:

```
DEBUG download_deps: locking_build_directory path="/home/erika/projects/gleam/use_local_package/build/packages"
TRACE download_deps: reading_file path="/home/erika/projects/gleam/use_local_package/gleam.toml"
DEBUG download_deps: reading_manifest_toml
TRACE download_deps: reading_file path="/home/erika/projects/gleam/use_local_package/manifest.toml"
DEBUG download_deps: manifest_up_to_date
```

The change in output localizes the issue to within:
https://github.com/gleam-lang/gleam/blob/main/compiler-cli/src/dependencies.rs#L511

Credit to @keroami for reporting the issue in the Gleam discord.

erikareads:
I'm looking into the issue, but I wanted to throw up an issue while I work.

erikareads:
This seems to be a likely culprit [source](https://github.com/gleam-lang/gleam/blob/main/compiler-cli/src/dependencies.rs#L511):

```rust
if manifest.requirements == config.all_dependencies()?
```

https://github.com/gleam-lang/gleam/issues/2286

It's comparing the manifest deps against the config deps to check,
The problem is that the manifest deps are all rewritten to be relative,
whereas the rewriting (probably) hasn't occured in the config.
So they're different, triggering a resolve.

[...]

PR #2451, opened 2023-12-05
fix: local dependencies with absolute paths no longer always resolve

erikareads:
This is a first crack at addressing https://github.com/gleam-lang/gleam/issues/2448.
I think we may want to fix the problem somewhere else, 
perhaps by canonicalizing all paths before comparison in both the manifest and config.

This commit also lacks tests, though I have manually tested that it resolves the issue.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -508,7 +508,11 @@ fn get_manifest<Telem: Telemetry>(
 
     // If the config has unchanged since the manifest was written then it is up
     // to date so we can return it unmodified.
-    if manifest.requirements == config.all_dependencies()? {
+    if is_same_requirements(
+        manifest.clone().requirements,
+        config.clone().all_dependencies()?,
+        paths.root(),
+    )? {
         tracing::debug!("manifest_up_to_date");
         Ok((false, manifest))
     } else {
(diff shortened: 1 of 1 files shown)


compiler-core/src/requirement.rs
```diff
@@ -2,22 +2,48 @@ use std::fmt;
...
     Hex { version: Range },
     Path { path: Utf8PathBuf },
     Git { git: EcoString },
 }
 
+impl PartialEq for Requirement {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Path { path: own_path }, Self::Path { path: other_path }) => {
+                is_same_file(own_path, other_path).unwrap_or(false)
```
lpil, line 25: We don't want to do take symlinks etc into consideration so we should not use this library. Instead we can use the same logic that is used to make the absolute path in the manifest.

https://github.com/gleam-lang/gleam/blob/4f6ba0a88b5c3589636637721460dfd47c6b44e1/compiler-cli/src/dependencies.rs#L701-L705

lpil: Thanks! `is_same_requirements` sounds fab to me, and easy to test that function. Thank you.

compiler-cli/src/dependencies.rs
```diff
@@ -508,7 +508,11 @@ fn get_manifest<Telem: Telemetry>(
 
     // If the config has unchanged since the manifest was written then it is up
     // to date so we can return it unmodified.
-    if manifest.requirements == config.all_dependencies()? {
+    if is_same_requirements(
+        manifest.clone().requirements,
+        config.clone().all_dependencies()?,
```
lpil, line 513: I've suggested that this method only borrows rather than taking ownership so these clones are no longer required, but for reference if items are to be cloned it should always be the smallest clone possible. Here the whole parent container is cloned and then the rest of it is immediately deallocated, which is wasted work.

compiler-cli/src/dependencies.rs
```diff
@@ -518,6 +522,49 @@ fn get_manifest<Telem: Telemetry>(
...
+                        path: fs::canonicalise(&root_path.join(path))?,
+                    },
+                    Requirement::Git { .. } => requirement,
+                },
+            ))
+        })
+        .collect::<Result<Vec<_>, gleam_core::Error>>()?
+        .into_iter()
+        .collect();
+    Ok(canonicalised_requirements1 == canonicalised_requirements2)
```
lpil, line 565: This function could have the duplication removed and run faster if instead of making two new hashmaps it checks that they are the same size, and then iterates over one and pulls the same key from the other, checking if they match as appropriate.

lpil (approved):
Thank you 💜

PR #2453, opened 2023-12-05
Add support for unicode escape sequences on both targets

### LINKED ISSUE #1509: Support \u escape sequence in strings
I haven't researched this much, but it looks like both Erlang and JavaScript support the `"\x"` escape sequence, e.g. `"\x1b[36mCYAN STRING\x1b[0m"`.

Currently, Gleam only allows `"\e"`, but Node.js doesn't seem to recognize it.

We should investigate further and document findings.

abs0luty:
Resolves #1509.


lpil (commented):
This looks great! Very clear and understandable code.

I think we could simplify the codegen a lot by turning `\u{ABCD}` into `\x{ABCD}` in Erlang.

https://www.erlang.org/doc/reference_manual/data_types.html#escape-sequences

abs0luty:
Oh god, I missed that. Thanks ❤

lpil:
I only learnt about it just now!

abs0luty:
@lpil fixed it.

compiler-core/src/erlang.rs
```diff
@@ -464,8 +464,32 @@ fn escape_atom_string(value: String) -> String {
...
+
 fn string(value: &str) -> Document<'_> {
-    value.to_doc().surround("<<\"", "\"/utf8>>")
+    Document::String(
+        unicode_escape_sequence_pattern()
+            // `\\u`-s should not be affected, so that "\\u..." is not converted to
+            // "\\x...". That's why capturing groups is used to exclude cases that
+            // shouldn't be replaced.
+            .replace_all(value, |caps: &Captures<'_>| {
+                if caps.get(1).is_some() {
```
lpil, line 481: I think this would not handle `\\\u{...}` correctly. We'd need to check how many slashes there are, rewriting if there's an odd number, not if there's an even number.

abs0luty:
@lpil fixed handling situations with multiple slashes

Issue #2299, opened 2023-08-16
Add --no-print-progress flag to gleam run and gleam build


PastMoments:
This is especially useful for piping program's stdout to other things.

Brief discussion about it here:
https://discord.com/channels/768594524158427167/1139046296439758858

lpil:
Good suggestion, thank you. I think I would like to use a more verbose flag as `--silent` is is more likely to be something the Gleam program would want to use. Perhaps `--no-print-progress`?

inoas:
Would this be only about the stdout infos or also about warnings?

lpil:
Only the progress information, warnings and errors would still be printed.

ohamill:
I'd like to work this ticket, if possible. Should I proceed with naming the flag `--no-print-progress`?

lpil:
Yes please 🙏

erikareads:
I would love this!

codereport:
Not sure if the `--silent` would suppress warnings as well but that would be very helpful.

lpil:
It would not, no.

inoas:
--warnings-as-errors would also be helpful for CI purposes (modifying the posix exit code to be 0)
I am personally agains --no-warnings

PR #2455, opened 2023-12-05
Add `no-print-progress` flag to `run` and `build`

ohamill:
Implements #2299

* `--no-print-progress` flag has been added to the `build` and `run` commands to suppress progress output
* Add `print_progress` parameter to `downloading_package`, `packages_downloaded`, and `compiling_package` methods in `Telemetry` trait
* Update implementations of `Telemetry` trait to include new parameter

### DIFF
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -14,9 +14,9 @@ use crate::{
     fs::{self, get_current_directory, ConsoleWarningEmitter},
 };
 
-pub fn download_dependencies() -> Result<Manifest> {
+pub fn download_dependencies(print_progress: bool) -> Result<Manifest> {
     let paths = crate::project_paths_at_current_directory();
-    crate::dependencies::download(&paths, cli::Reporter::new(), None, UseManifest::Yes)
+    crate::dependencies::download(&paths, cli::Reporter::new(), None, UseManifest::Yes, print_progress)
 }
 
 pub fn main(options: Options, manifest: Manifest) -> Result<Built> {
@@ -32,6 +32,7 @@ pub fn main(options: Options, manifest: Manifest) -> Result<Built> {
         options.target.unwrap_or(root_config.target),
     )?;
     let current_dir = get_current_directory().expect("Failed to get current directory");
+    let print_progress = options.print_progress;
 
     tracing::info!("Compiling packages");
     let compiled = {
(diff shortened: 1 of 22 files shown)


fschwalbe:
Wouldn't you typically let the tracing subscriber filter events/spans based on the level or target (or other metadata fields) instead of adding another layer of our own ad-hoc filtering on top?

lpil:
Hello! We already have a system for this. Please pass in the Null reporter instead of the Cli reporter when this option is set. Thank you

compiler-cli/src/build.rs
```diff
@@ -48,9 +54,12 @@ pub fn main(options: Options, manifest: Manifest) -> Result<Built> {
...
-    match perform_codegen {
-        Codegen::All | Codegen::DepsOnly => cli::print_compiled(start.elapsed()),
-        Codegen::None => cli::print_checked(start.elapsed()),
-    };
+    if !no_print_progress {
+        match perform_codegen {
+            Codegen::All | Codegen::DepsOnly => cli::print_compiled(start.elapsed()),
+            Codegen::None => cli::print_checked(start.elapsed()),
+        };
+    }
```
lpil, line 62: Move these onto the telemetry object please, don't want to have any conditional logic for this.

compiler-core/src/build/project_compiler.rs
```diff
@@ -49,6 +49,7 @@ pub struct Options {
     pub target: Option<Target>,
     pub codegen: Codegen,
     pub warnings_as_errors: bool,
+    pub no_print_progress: bool,
```
lpil, line 52: This is a command line interface, but the time this is run we already have a telemetry object, so this can be removed.

compiler-cli/src/shell.rs
```diff
@@ -14,8 +14,9 @@ pub fn command() -> Result<(), Error> {
             codegen: Codegen::All,
             mode: Mode::Dev,
             target: Some(Target::Erlang),
+            no_print_progress: false,
         },
-        crate::build::download_dependencies()?,
+        crate::build::download_dependencies(false)?,
```
lpil, line 19: Pass a telemetry object into this function please 🙏

giacomocavalieri:
Hello @ohamill! Are you still working on this PR? If not I could help merge this, you did a great job here so it will probably need just some really minor adjustments 😊

ohamill:
> Hello @ohamill! Are you still working on this PR? If not I could help merge this, you did a great job here so it will probably need just some really minor adjustments 😊

Hello! Yes, sorry for the delay, I unfortunately haven't had any time lately to keep working on this. If you're able to help merge it, that would be a great help, thank you!

giacomocavalieri:
> Hello! Yes, sorry for the delay, I unfortunately haven't had any time lately to keep working on this. If you're able to help merge it, that would be a great help, thank you!

No, thank you! I'll try my best

@lpil I have a question about the `Telemetry` trait, could we change the code to use a `Telemetry` enum with different variants `NullTelemetry | LogTelemetry | CliReporterTelemetry` instead of the current design based on a trait?
It would make passing around the telemetry object easier and remove the need for all the `<Telem: Telemetry>` type annotation that would pop up where it's added as an argument.
I'm not sure using a trait is worth the extra complexity, especially since all implementations are empty structs but maybe I'm not seeing the problems with the enum-based approach

lpil:
That cannot be done, no. `gleam_core` is pure, it does not have any ability to perform IO. There's lots of reasons why this is a good thing, but one of the hardest barrier to changing this is that if `gleam_core` could perform arbitrary IO then we couldn't run it in a browser any more.

lpil:
Hello gang! Is anyone currently working on this?

giacomocavalieri:
I was but got stuck on the refactoring, it wasn't as easy as I expected 🥲 feel free to pick this up if you want to!

lpil:
OK! I'll close this as no one is working on it. Please feel free to reopen, thank you.

PR #2456, opened 2023-12-06
Make Unicode escape sequences 1 to 6 digits long

fschwalbe:
The maximum length of 6 hex digits fits a 21 bit Unicode codepoint, and being clearly designated by the surrounding braces there is no reason not to allow 1, 3 or 5 digits too.

This behaviour seems to be in line with most other languages that have a similar feature.


lpil (approved):
Thank you!!

PR #2459, opened 2023-12-08
Allow `gleam new` in existing directories

### LINKED ISSUE #1879: Support creating a new Gleam project in the current directory with `gleam new .`

abs0luty:
Resolves #1879.


lpil (approved):
Wow, very nice! Very clean. Thank you

lpil:
I'm going to hold off on merging this until I've finished this release. Thank you

abs0luty:
@lpil Release is finished, so it's time?

lpil (commented):
Hello! I've just given this a test and I've found a few problems. The git repo isn't being initialised, the github direction is missing, and any existing files get overwritten.

Tests to prevent regressions for these would be useful 🙏

abs0luty:
Oh it seems like it only checks for overwritten files when initialising in current directory. Thanks for pointing this out!

abs0luty:
@lpil with git I think the issue is that it calls `git rev-parse` in the current directory:

```rs
pub fn is_inside_git_work_tree(path: &Utf8Path) -> Result<bool, Error> {
    tracing::trace!(path=?path, "checking_for_git_repo");

    let args: Vec<&str> = vec!["rev-parse", "--is-inside-work-tree", "--quiet"];

    // Ignore all output, rely on the exit code instead.
    // git will display a fatal error on stderr if rev-parse isn't run inside of a git work tree,
    // so send stderr to /dev/null
    let result = std::process::Command::new("git")
        .args(args)
        .stdin(std::process::Stdio::null())
        .stdout(std::process::Stdio::null())
        .stderr(std::process::Stdio::null())
        .current_dir(path)
        .status();

    match result {
        Ok(status) => Ok(status.success()),
        Err(error) => match error.kind() {
            io::ErrorKind::NotFound => Ok(false),

            other => Err(Error::ShellCommand {
                program: "git".into(),
                err: Some(other),
            }),
        },
    }
}
```

And if there is no git repo outside, it doesn't create the git folder. Should I somehow fix this?

abs0luty:
I think checking if `.git` directory exists is fine.

compiler-cli/src/new.rs
```diff
@@ -179,64 +172,79 @@ jobs:
...
+          # elixir-version: "{ELIXIR_VERSION}"
       - run: gleam deps download
       - run: gleam test
       - run: gleam format --check src test
 "#,
-                ERLANG_OTP_VERSION, self.gleam_version, REBAR3_VERSION, ELIXIR_VERSION,
-            ),
-        )
+            )),
+            _ => None,
```
lpil, line 184: Remove this `_` please. We want to match all variants explicitly so that when a new variant is added then the Rust compiler prompts the programmer to update this.

compiler-core/src/error.rs
```diff
@@ -556,6 +560,24 @@ to `src/{module}.gleam`"
                 location: None,
             },
 
+            Error::OutputFilesAlreadyExist { file_names } => Diagnostic {
+                title: "File(s) already exist in target directory".into(),
```
lpil, line 564: When there's only 1 file this should say "File", and when there's another number it should say "Files"

compiler-core/src/error.rs
```diff
@@ -556,6 +560,24 @@ to `src/{module}.gleam`"
...
+                text: format!(
+                    "{}
+
+If you want to overwrite these files, delete them
+and run the command again.
+",
+                    file_names
+                        .iter()
+                        .map(|name| format!("- {}", name.as_str()))
```
lpil, line 573:
```suggestion
                        .map(|name| format!("  - {}", name.as_str()))
```

lpil (commented):
Almost there! Thank you

abs0luty:
@lpil fixed everything needed :3

lpil (commented):
I think something has gone wrong with this PR, there's a bunch of unrelated changes here!

abs0luty:
@lpil seems like I rebased main

PR #2472, opened 2023-12-15
String alias bug fix

### LINKED ISSUE #2471: String alias is always bound to first alternative value
When pattern matching on strings one could assign a name to a constant prefix, for example:
```gleam
case foo {
  "bar" as prefix <> _ | "baz" as prefix <> _ -> prefix
  _ -> "no prefix"
}
```
However, `prefix` is always bound to the first alternative, that is `"bar"` in this example.
For example this is the generated javascript code:
```js
export function go(foo) {
  if (foo.startsWith("bar")) {
    return "bar";
  } else if (x.startsWith("baz")) {
    return "baz";
  } else {
    return "no prefix";
  }
}
```

I've already started working on a fix!

giacomocavalieri:
This PR closes #2471

Now aliases for string prefixes are treated similarly to aliases for list items: if there's an alias for a string prefix the compiler assigns that to a local variable with the given name.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Fixed a bug where string prefix aliases defined in alternative case branches
+  would all be bound to the same constant.
+
 ## v0.33.0-rc2 - 2023-12-07
 
 ### Language changes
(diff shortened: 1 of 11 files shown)


lpil (approved):
Amazing! Thank you! Would you mind updating the changelog and adding some comments explaining what the new logic does and why.

giacomocavalieri:
Thanks for the feedback! Somehow I keep forgetting about the Changelog 😆
I've also added some comments explaining the changes in the code generation

lpil (approved):
Wow, those are fantastic comments!! I'm gunna copy this style.

Issue #835, opened 2020-11-07
Formatter: no line-breaking for long boolean expressions


edkelly303:
In Gleam 0.12.0
```gleam
fn something() -> Bool {
  let some_long_variable_name = 1
  let some_other_long_variable_name = 2
  let some_third_long_variable_name = True
  let some_fourth_long_variable_name = False
  some_long_variable_name > some_other_long_variable_name || some_third_long_variable_name && some_fourth_long_variable_name
}
```
`gleam format` doesn't seem to attempt to break the the boolean expression over multiple lines, no matter how long it gets.

Not sure what the best place to break would be. Perhaps start by breaking on any || or &&s?
```gleam
some_long_variable_name > some_other_long_variable_name 
  || some_third_long_variable_name 
  && some_fourth_long_variable_name
```

lpil:
Something like that would be good, though I don't mind the exact formatting so much.

I think we could expand it more generally to any binary operations too.

massivefermion:
There is also something like [this](https://github.com/exercism/gleam/blob/main/exercises/practice/bowling/.meta/example.gleam#L41). I also think it looks better if the break is on operators!

inoas:
Another example:

```gleam
pub fn to_string(color color: Color) -> String {
  int.to_string(color.red) <> ", " <> int.to_string(color.green) <> ", " <> int.to_string(
    color.blue,
  ) <> ", " <> float.to_string(color.alpha)
}
```

giacomocavalieri:
I think this can be closed now!

lpil:
Thank you

PR #2473, opened 2023-12-16
Change binary operation formatting

giacomocavalieri:
This PR tries to address issue #835, I tried to keep it general and handle chain of binary operations in general instead of just boolean ones.

I'm not 100% sure about the look of some of the formatted code snippets I had to change, any feedback is more than welcome!


lpil (approved):
Thank you so much!

PR #2483, opened 2023-12-21
Generalize module constants

### LINKED ISSUE #2204: Constant's type is determined only once which causes strange errors
This code:

```gleam
import gleam/io
import gleam/result

pub fn with_smth(f: fn() -> a) -> a {
  f()
}

pub const with_smth2 = with_smth

pub fn main() {
  use <- with_smth2()
  Ok(Nil)
}

pub fn another() {
  use <- with_smth2()
  Nil
}
```

produces this error:

```
error: Type mismatch
   ┌─ /tmp/const_forward/src/const_forward.gleam:17:3
   │
17 │   Nil
   │   ^^^

Expected type:

    fn() -> Result(Nil, a)

Found type:

    fn() -> Nil
```

The problem is that when this constant is public and used in multiple files, these files fail in random order depending on slight code changes, and the error doesn't give enough context.

omprakaash:
This should fix #2204.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -157,24 +157,29 @@ pub fn infer_module<A>(
         let statement = insert_type_alias(t, &mut env)?;
         typed_statements.push(statement);
     }
-    for c in statements.constants {
-        let statement = infer_module_constant(c, &mut env, &name)?;
-        typed_statements.push(statement);
-    }
 
-    // Sort the functions into dependency order for inference. Functions that do
-    // not depend on other functions are inferred first, then ones that depend
+    // Sort functions and constants into dependency order for inference. Definitions that do
+    // not depend on other definitions are inferred first, then ones that depend
     // on those, etc.
-    let function_groups = into_dependency_order(statements.functions)?;
+    let definition_groups = into_dependency_order(statements.functions, statements.constants)?;
     let mut working_group = vec![];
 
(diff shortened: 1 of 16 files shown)


omprakaash:
Modified call-graph to work with constants too. Just wanted to see if this was a good approach before modifying the tests for that as well.

compiler-core/src/analyse.rs
```diff
@@ -157,24 +157,29 @@ pub fn infer_module<A>(
...
+            match definition {
+                Definition::Function(f) => {
+                    let statement = infer_function(f, &mut env, &mut hydrators, &name)?;
+                    working_group.push(statement);
+                }
+                Definition::ModuleConstant(c) => {
+                    let statement = infer_module_constant(c, &mut env, &name)?;
+                    working_group.push(statement);
+                }
+                _ => unreachable!("Only funtions and constants should be in a definition group"),
```
lpil, line 181: Rather than crash at runtime let's have some type that is either a function or a module constant, making the invalid state impossible.

lpil (commented):
Looks great!! Let me know when it's ready for a full review 💜

compiler-core/src/analyse.rs
omprakaash: Got it. Will refactor this.

lpil (approved):
Thank you!

PR #2485, opened 2023-12-22
Formatter fix

giacomocavalieri:
This PR fixes some quirks with the formatting of binary operations:
- binary operations at the beginning of a pipe are now kept together
  ```diff
  // this
  + 1 * 2
  + |> inc
  // not this
  - 1
  - * 2
  - |> inc
  ```
- binary operations in constructors/function calls only get broken if needed (and are indented, if actually broken)
  ```diff
  // this
  +Dog(
  +  name: "James " <> last_name,
  +  age: 12,
  +)
  // not this
  -Dog(
  -  name: "James "
  -  <> last_name,
  -  age: 12,
  -)
  ```
- binary operations in case branches only get broken if needed
  ```diff
  // this
  +case foo {
  +  _ -> 1 + 2
  +}
  // not this
  -case foo {
  -  _ ->
  -    1
  -    + 2
  -}
  ```

Thanks to everyone who pointed out all the little quirks I missed in the previous release, turns out binary operations are quite challenging to print correctly

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -689,6 +689,17 @@ pub enum BinOp {
     Concatenate,
 }
 
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum OperatorKind {
+    BooleanLogic,
+    Equality,
+    IntComparison,
+    FLoatComparison,
+    IntMath,
+    FloatMath,
+    StringConcatenation,
+}
+
 impl BinOp {
     pub fn precedence(&self) -> u8 {
         // Ensure that this matches the other precedence function for guards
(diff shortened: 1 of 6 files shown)


compiler-core/src/ast.rs
```diff
@@ -747,6 +747,40 @@ impl BinOp {
             Self::Concatenate => "<>",
         }
     }
+
+    pub fn can_be_grouped_with(self, other: BinOp) -> bool {
+        let is_boolean_binop = |binop| matches!(binop, BinOp::And | BinOp::Or);
+        let is_int_comparison = |binop| {
+            matches!(
```
lpil, line 754: Could you convert these to match expressions that list every possible value? With this macro and wildcards the compiler doesn't tell you if a new variant is added, which can result in bugs as logic doesn't get updated.

Might be good to extract them to methods too.

compiler-core/src/ast.rs
```diff
@@ -747,6 +747,40 @@ impl BinOp {
...
+                binop,
+                BinOp::AddFloat | BinOp::SubFloat | BinOp::DivFloat | BinOp::MultFloat
+            )
+        };
+        self == other
+            || (is_boolean_binop(self) && is_boolean_binop(other))
+            || (is_float_comparison(self) && is_float_comparison(other))
+            || (is_int_comparison(self) && is_int_comparison(other))
+            || (is_float_math(self) && is_float_math(other))
+            || (is_int_math(self) && is_int_math(other))
```
lpil, line 782: Maybe this could be written as `self .operator_kind() == other.operator_kind()`, where this method returns a new enum saying if it is int, float, etc.

lpil (commented):
Gorgeous work! Thank you! Wouldn't you mind rebasing to remove the merge commit please 🙏

compiler-core/src/ast.rs
giacomocavalieri: Great idea! I'll do that

giacomocavalieri:
I should have addressed all comments 😁
Adding the `OperatorKind` enum was a huge readability improvement, tysm for the suggestion!

lpil (approved):
Thank you!!!

compiler-core/src/format/tests/binary_operators.rs
```diff
@@ -0,0 +1,172 @@
...
+    );
+
+    assert_format!(
+        r#"pub fn main() {
+  Ok(Lesson(
+    name: names.name,
+    text: text,
+    code: code,
+    path: chapter_path
+      <> "/"
```
jonmellman, line 111: By any chance, would you mind explaining how the formatter knows to indent here but knows _not_ to indent on line 9 of this file?

I'm trying to follow along with this PR, but I'm new to this formatter code. Thanks!

lpil: If you check the call_arg function in the formatter it add a layer of indentation for binary operators

giacomocavalieri: @jonmellman if you want to chat about pretty printing you can reach out in the Gleam discord server! I'd be more than happy to share what I've learned about the topic 💜

PR #2488, opened 2023-12-23
fix(cli): change `deps list` to look in current directory

### LINKED ISSUE #2439: gleam deps list path resolution with issue with local depedencies
when a dependency comes from a local path rather than hex, gleam dep list fails to find it as I believe it assumes a path from root:

https://github.com/gleam-lang/gleam/blob/main/compiler-cli/src/dependencies.rs#L37

leading to the error like so

```bash
> gleam deps list
  Resolving versions
error: File IO failure

An error occurred while trying to canonicalise this file:

    /../mist

```

I'll try see if I can throw up a PR for this soon

fffnite:
Previously this was looking at the file system root, which meant it could not find the `manifest.toml` or resolve relative paths to local libraries.

fixes #2439

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Fixed a bug where `gleam add` would not update `manifest.toml` correctly.
 - `Utf8Codepoint` has been renamed to `UtfCodepoint` in `prelude.d.mts`.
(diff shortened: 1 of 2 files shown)


lpil (approved):
Oh dear! How did that happen.

Would you mind updating the changelog? Thank you.

lpil (approved):

PR #2489, opened 2023-12-23
feat(config): add support for gitea/forgejo/codeberg to repository config

### LINKED ISSUE #2470: toml configuration support for forgejo/gitea
In gleam.toml

Currently to add a repository for codeberg.org for example one has to use 

```repository = { type = "custom", url = "https://codeberg.org/{user}/{project}/" }```

would be nicer to be able to use something like:

```repository = { type = "codeberg", user = "username", repo = "project" }```

to get it on par with github.

### PR DESCRIPTION
**Pull request by fffnite**
Note that this link assumes that it is linking to a git tag rather than a branch. There doesn't seem to be a generic path for both on these platforms that I can see, this may cause them to link to a non-existent path? Open to suggestions.

resolves #2470

### CODE REVIEW

File: CHANGELOG.md [hunk 1]
```diff
@@ -11,6 +11,12 @@
 
 - Fixed some quirk with the formatting of binary operators.
 
+### Config
```
lpil, line 14: Could you change this to `Build tool` and put it above the bugs section? Thank you.

compiler-core/src/config.rs
```diff
@@ -718,10 +718,35 @@ pub struct DenoConfig {
...
+    Forgejo {
+        user: String,
+        repo: String,
+        host: String,
+    },
+    Gitea {
+        user: String,
+        repo: String,
+        host: String,
+    },
```
lpil, line 746: These two are the same, so one can be a serde alias of the other.

compiler-core/src/config.rs
```diff
@@ -733,6 +758,12 @@ impl Repository {
             Repository::BitBucket { repo, user } => {
                 Some(format!("https://bitbucket.com/{user}/{repo}"))
             }
+            Repository::CodeBerg { repo, user } => {
+                Some(format!("https://codeberg.org/{user}/{repo}"))
+            }
+            Repository::Forgejo { repo, user, host } | Repository::Gitea { repo, user, host } => {
+                Some(format!("https://{host}/{user}/{repo}"))
```
lpil, line 765: Let's only add the `https://` if it or `http://` has not already been specified.

compiler-core/src/docs/source_links.rs
```diff
@@ -49,6 +49,22 @@ impl SourceLinker {
...
+                format!(
+                    "https://codeberg.org/{}/{}/src/tag/{}/{}#L",
+                    user, repo, project_config.version, path_in_repo
+                ),
+                "-".into(),
+            )),
+            Repository::Forgejo { user, repo, host } | Repository::Gitea { user, repo, host } => {
+                Some((
+                    format!(
+                        "https://{host}/{user}/{repo}/src/tag/{}/{}#L",
```
lpil, line 62: Here too, let's be careful not to add the scheme twice

lpil (commented):
Wonderful! Thank you. I've added some minor notes inline

fffnite:
I've change it over to use the custom serde uri deserializer in that module, though it appears to have the behaviour of adding a https:// and sanitizing the user provided scheme if it exists. Not sure this is really the preferred behaviour.

Should I maybe write some other helper function/deserializer?

lpil:
> should I still add default to https case or something in a new decoder?

Yes please! That would be fab

lpil (approved):
Thank you!

Issue #2490, opened 2023-12-24
Formatter: removes @external line without gleam def


keroami:
when I add `@external(erlang, "module" "fun")` but not yet the gleam declaration `pub fn mod_fun() -> Type`, the formatter removes the `@external` line.

Then I have to type it again.

lpil:
Thank you. This should be a parse error.

abs0luty:
I can work on this one

lpil:
Massivefermion is working it currently!

abs0luty:
Ok :3

michallepicki:
I think this can be closed now since https://github.com/gleam-lang/gleam/pull/2497 got merged?

keroami:
checked w/ 0.34.rc1
format does not do anything, as it is (now) a syntax error.
Closing.

PR #2497, opened 2023-12-28
not providing a definition after attributes is a parse error

massivefermion:
addresses #2490

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -240,7 +240,7 @@ where
 
     fn parse_definition(&mut self) -> Result<Option<TargetedDefinition>, ParseError> {
         let mut attributes = Attributes::default();
-        self.parse_attributes(&mut attributes)?;
+        let location = self.parse_attributes(&mut attributes)?;
 
         let def = match (self.tok0.take(), self.tok1.as_ref()) {
             // Imports
@@ -293,10 +293,18 @@ where
             }
         }?;
 
-        Ok(def.map(|definition| TargetedDefinition {
-            definition,
-            target: attributes.target,
-        }))
+        match def {
+            Some(definition) => Ok(Some(TargetedDefinition {
+                definition,
+                target: attributes.target,
+            })),
+
+            None if location.is_some() => {
+                parse_error(ParseErrorType::ExpectedDefinition, location.expect(""))
+            }
+
+            None => Ok(None),
+        }
     }
 
     //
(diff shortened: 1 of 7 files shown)


compiler-core/src/parse.rs
```diff
@@ -292,10 +292,18 @@ where
...
-            target: attributes.target,
-        }))
+        match def {
+            Some(definition) => Ok(Some(TargetedDefinition {
+                definition,
+                target: attributes.target,
+            })),
+
+            None if location.is_some() => {
+                parse_error(ParseErrorType::ExpectedDefinition, location.expect(""))
```
lpil, line 302: Please match on `(def, location)`  rather than calling expect in this clause.

compiler-core/src/parse.rs
```diff
@@ -2795,37 +2803,53 @@ where
...
+                start_of_attributes = Some(start);
+            }
+
+            end_of_attributes = Some(self.parse_attribute(start, attributes)?);
         }
-        Ok(())
+
+        Ok(start_of_attributes.map(|start| SrcSpan {
+            start,
+            end: end_of_attributes.expect(""),
```
lpil, line 2823: Do not use `.expect` if it can be avoided, and always give it a message.

The change above to use SrcSpan means this isn't needed 👍

compiler-core/src/parse.rs
```diff
@@ -2795,37 +2803,53 @@ where
         }
     }
 
-    fn parse_attributes(&mut self, attributes: &mut Attributes) -> Result<(), ParseError> {
+    fn parse_attributes(
+        &mut self,
+        attributes: &mut Attributes,
+    ) -> Result<Option<SrcSpan>, ParseError> {
+        let mut start_of_attributes = None;
+        let mut end_of_attributes = None;
```
lpil, line 2811: These either both exist or both don't exist, so use `let mut location: Option<SrcSpan>`

compiler-core/src/parse/tests.rs
```diff
@@ -631,6 +631,16 @@ pub fn main() -> Nil {
...
 
+#[test]
+fn attributes_with_no_definition() {
+    assert_module_error!(
+        r#"
+@deprecated("1")
+@target(erlang)
+"#
+    );
+}
```
lpil, line 642: Could we have test for the original issue reported? That is having an external on a non-function definition.

compiler-core/src/parse/snapshots/gleam_core__parse__tests__attributes_with_no_definition.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: compiler-core/src/parse/tests.rs
+expression: "\n@deprecated(\"1\")\n@target(erlang)\n"
+---
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:1
+  │  
+2 │ ╭ @deprecated("1")
+3 │ │ @target(erlang)
+  │ ╰───────────────^ I was expecting a definition after this
```
lpil, line 10: I think this error message will be confusing when there is a definition but it's one that is not supported, like external on a type.

lpil (commented):
Thank you!

lpil (approved):
Thank you

lpil:
Could you rebase on main please?

lpil (approved):
Thank you!

lpil:
Could you rebase on main to remove the merge commit please?

lpil:
Thankyou!

PR #2498, opened 2023-12-28
Permit to define target specific functions without needlessly failing compilation for a different target

### LINKED ISSUE #2324: Permit deps with other-target only functions to compile
Not sure how to title this because in a way it’s two issues. title could be fixed

## Problem(s)

1. If I import a javascript only dependency, using `@target(javascript)` and build for erlang, Gleam gives an error regarding that dependency not implementing a function in erlang.

2. If do the same as before but not import that javascript only dependency, I still get the same error even though I have not used the function.

## Steps to recreate the issue

1. `gleam new abc`
2. `cd abc`
3. `gleam add gleam_javascript`
4. `gleam build` 

> `error: Function without an implementation`

## Possible solutions

- In `gleam.toml` add an option to specify the target for a dependency. 
ex:
`gleam_javascript = { version = "0.6.0", target = "javascript" }`

- Build dependencies based on the target specified by the gleam.toml of the dependency. 
Assuming I add `gleam_javascript` and in its gleam.toml it says `target=javascript`, if I compile to erlang that dependency should not build

- Do not build a dependency if it’s not used. This could potentially speed up compile time by not building a dependency for no reason.

giacomocavalieri:
This closes #2324

To get a better feeling of the problem this PR addresses you can have a look at this code snippet:
```gleam
import gleam/io

@external(javascript, "foo", "bar")
pub fn js_only() -> Int

pub fn main() {
  io.println("I'm Erlang code!")
}
```
Trying to compile it on the Erlang target would fail with an error complaining that `js_only` doesn't have an implementation for the Erlang target. But this is needlessly limiting! Since `js_only` is never actually referenced by functions that are going to be compiled down to Erlang this code is still safe to compile and run!

After this PR compilation will only fail if someone tries to use a target-specific function in a code path that could be used in a different target:
```gleam
import gleam/io

@external(javascript, "foo", "bar")
pub fn js_only() -> Int

pub fn main() {
  io.println("I'm Erlang code!")
  js_only()
//^^^^^^^^^ error: this function doesn't have an implementation for the Erlang target.
}
```

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -182,7 +182,7 @@ impl CommandExecutor for ProjectIO {
             .args(args)
             .stdin(stdio.get_process_stdio())
             .stdout(stdio.get_process_stdio())
-            .envs(env.iter().map(|(a, b): &(&str, String)| (a, b)))
+            .envs(env.iter().map(|(a, b)| (a, b)))
             .current_dir(cwd.unwrap_or_else(|| Utf8Path::new("./")))
             .status();
 
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -182,7 +182,7 @@ impl CommandExecutor for ProjectIO {
             .args(args)
             .stdin(stdio.get_process_stdio())
             .stdout(stdio.get_process_stdio())
-            .envs(env.iter().map(|(a, b)| (a, b)))
+            .envs(env.iter().map(|(a, b): &(&str, String)| (a, b)))
             .current_dir(cwd.unwrap_or_else(|| Utf8Path::new("./")))
             .status();
 
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -484,6 +484,7 @@ fn register_value_from_function(
         body: _,
         return_type: _,
         supported_targets,
+        target: _,
     } = f;
     assert_unique_name(names, name, *location)?;
     assert_valid_javascript_external(name, external_javascript.as_ref(), *location)?;
@@ -582,6 +583,7 @@ fn infer_function(
         external_erlang,
         external_javascript,
         return_type: (),
+        target,
         supported_targets: _,
     } = f;
     let preregistered_fn = environment
@@ -623,6 +625,14 @@ fn infer_function(
 
     let external_targets = external_supported_targets(&external_erlang, &external_javascript);
 
+    // If the function has a target annotation, the only required target for its
+    // body is the annotated target. Otherwise the required targets are all the
+    // targets that do not have an external implementation.
+    let required_targets = match target {
+        Some(target) => SupportedTargets::from_target(target),
+        None => SupportedTargets::all().difference(external_targets),
+    };
+
     // Infer the type using the preregistered args + return types as a starting point
     let (type_, args, body, mut supported_targets) = environment.in_new_scope(|environment| {
         let args_types = arguments
@@ -630,7 +640,7 @@ fn infer_function(
             .zip(&args_types)
             .map(|(a, t)| a.set_type(t.clone()))
             .collect();
-        let mut expr_typer = ExprTyper::new(environment, external_targets);
+        let mut expr_typer = ExprTyper::new(environment, required_targets);
         expr_typer.hydrator = hydrators
             .remove(&name)
             .expect("Could not find hydrator for fn");
@@ -647,7 +657,7 @@ fn infer_function(
         // supported targets the ones that have an external implementation.
         supported_targets = external_targets;
     } else {
-        supported_targets.merge(&external_targets);
+        supported_targets = supported_targets.merge(external_targets);
     }
 
     // Assert that the inferred type matches the type of any recursive call
@@ -686,6 +696,7 @@ fn infer_function(
         body,
         external_erlang,
         external_javascript,
+        target,
         supported_targets,
     }))
 }
@@ -947,9 +958,15 @@ fn infer_module_constant(
         annotation,
         public,
         value,
+        target,
         ..
     } = c;
-    let mut expr_typer = ExprTyper::new(environment, SupportedTargets::none());
+
+    let required_targets = match target {
+        Some(target) => SupportedTargets::from_target(target),
+        None => SupportedTargets::all(),
+    };
+    let mut expr_typer = ExprTyper::new(environment, required_targets);
     let typed_expr = expr_typer.infer_const(&annotation, *value)?;
     let type_ = typed_expr.type_();
     let variant = ValueConstructor {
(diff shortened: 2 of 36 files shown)


compiler-cli/src/fs.rs
```diff
@@ -182,7 +182,7 @@ impl CommandExecutor for ProjectIO {
             .args(args)
             .stdin(stdio.get_process_stdio())
             .stdout(stdio.get_process_stdio())
-            .envs(env.iter().map(|(a, b)| (a, b)))
+            .envs(env.iter().map(|(a, b): &(&str, String)| (a, b)))
```
lpil, line 185: What's this for?

compiler-core/src/analyse.rs
```diff
@@ -697,11 +732,19 @@ fn target_function_implementation<'a>(
     }
 }
 
-fn ensure_body_given(body: &Vec1<UntypedStatement>, location: SrcSpan) -> Result<(), Error> {
-    if body.first().is_placeholder() {
-        Err(Error::NoImplementation { location })
-    } else {
-        Ok(())
+fn is_empty_body(body: &Vec1<UntypedStatement>) -> bool {
```
lpil, line 735: This is called a placeholder, not empty.

lpil (commented):
Nice! DM'd you some Qs on discord. I'll do a full review after 🙏

compiler-cli/src/fs.rs
giacomocavalieri: I noticed CI was failing because the linter wanted me to remove this map, it's a false positive because it thinks it's the same as applying the identity function so I had to make the type explicit

compiler-core/src/analyse.rs
abs0luty: Isn't placeholder _used when parsing is incomplete or when a function body is missing_?

compiler-cli/src/fs.rs
abs0luty: You can alternatively use `#[allow(clippy::map_identity)]`

giacomocavalieri: Ah I see, thank you! I'll revert this then

compiler-cli/src/compile_package.rs
```diff
@@ -47,6 +47,7 @@ pub fn command(options: CompilePackage) -> Result<()> {
         &target,
         ids,
         ProjectIO::new(),
+        todo!("I'm not sure what should go here"),
```
giacomocavalieri, line 50: @lpil I'm not sure what to do here, in the compiler-core package I had access to a `is_root` variable but here there's no such thing

compiler-core/src/type_/expression.rs
```diff
@@ -10,28 +10,131 @@ use crate::{
...
+    ) -> Self {
         let mut hydrator = Hydrator::new();
+
         hydrator.permit_holes(true);
         Self {
             hydrator,
             environment,
+            // This will be narrowed down as the expression type is inferred
+            supported_targets: SupportedTargets::all(),
+            external_supported_targets,
```
lpil, line 137: Why are these tracks independently? Seems they always get merged together before use so it can be simplified to a single value. This would also permit the `merge` method to be removed.

compiler-core/src/type_/expression.rs
```diff
@@ -10,28 +10,131 @@ use crate::{
...
+            Target::Erlang => SupportedTargets {
+                erlang: true,
+                javascript: self.javascript,
+            },
+            Target::JavaScript => SupportedTargets {
+                erlang: self.erlang,
+                javascript: true,
+            },
+        }
+    }
```
lpil, line 88: Favour mutation for these sorts of operations
```suggestion
    pub fn add(&mut self, target: Target) {
        match target {
            Target::Erlang => self.erlang = true,
            Target::JavaScript => self.javascript = true,
        }
    }
```

compiler-core/src/type_/expression.rs
```diff
@@ -10,28 +10,131 @@ use crate::{
...
+
+    pub fn intersect(&self, targets: SupportedTargets) -> SupportedTargets {
+        SupportedTargets {
+            erlang: self.erlang && targets.erlang,
+            javascript: self.javascript && targets.javascript,
+        }
+    }
+
+    pub fn merge(&self, targets: SupportedTargets) -> SupportedTargets {
```
lpil, line 70: This can be removed once external support was not tracked in a second data structure.

compiler-core/src/type_/expression.rs
```diff
@@ -10,28 +10,131 @@ use crate::{
...
+            Target::JavaScript => SupportedTargets::javascript(),
+        }
+    }
+
+    pub fn intersect(&self, targets: SupportedTargets) -> SupportedTargets {
+        SupportedTargets {
+            erlang: self.erlang && targets.erlang,
+            javascript: self.javascript && targets.javascript,
+        }
+    }
```
lpil, line 68: Perhaps this could be called `narrow` as that's the word used elsewhere in this PR.

Let's also use mutation here.

compiler-core/src/type_/expression.rs
giacomocavalieri: I think we have to keep track of those separately in the `ExprTyper`: when narrowing I have to perform this check 
```rust
self.supported_targets.narrow(new_targets);
if self.environment.target_support == TargetSupport::Enforced
   && !self.supported_targets
      .merge(self.external_supported_targets)
      .supports(self.environment.target) {
  // error this function has no supported targets and we must fail!
}
```
Supported targets get narrowed down starting from `SupportedTargets::all()`, the external targets, however, never change: imagine you had a function with an external js implementation that uses an Erlang-only function in its body:
```gleam
@external(javascript, "foo", "bar")
pub fn foo() {
  erlang_only()
}
```
If you had a single `supported_targets` field it would be narrowed down to just `SupportedTargets::erlang()` and you'd loose track of the fact that this function can actually support javascript as well. Now If you're compiling to JavaScript the check would fail:
```rust
self.supported_targets.narrow(new_targets);
if self.environment.target_support == TargetSupport::Enforced
   && !self.supported_targets.supports(self.environment.target) {
  // error this function has no supported targets and we must fail!
```
Because the body's supported target was narrowed down to just `Erlang` and you're not considering that you can still support JavaScript since you have an external implementation for that target.

Though the check could be written in a different way to make things a bit clearer and not rely on the merge function:

```rust
self.supported_targets.narrow(new_targets);
if self.environment.target_support == TargetSupport::Enforced
   && !(
     self.supported_targets.supports(self.environment.target)
     || self.external_targets.supports(self.environment.target)
   ) {
  // error this function has no supported targets and we must fail!
```

We would still need `merge` to merge the external targets and the body supported targets once the typer did its thing and record all the targets supported by the function, both those of its body and those with and external implementation.

lpil: Ah I get it now, and you can't do that I originally said because that wouldn't give you the error at the point the non-supporting value is referenced. Thank you

PR #2501, opened 2023-12-28
updated types on `prelude.d.mts` to match `prelude.mjs`

### LINKED ISSUE #2500: Types on `prelude.d.mts` are missing/incomplete
*Note: These are all errors that come up with the LSP, nothing related to runtime*
## Some examples

This is due to declaration only exporting `OkStatic`
```js
import { Ok } from "./gleam.mjs"; // Module '"./gleam.mjs"' has no exported member 'Ok'
```

This is due to `Result` being an interface instead of a class.
```ts
import { Result } from "./gleam.mjs"; // 'Result' is a type and cannot be imported in JavaScript files. Use 'import("./gleam.mjs").Result' in a JSDoc type annotation.
```

Additionally, some functions regarding `byteArray` are missing.

Enderchief:
Closes #2500 

- interfaces swapped out for classes
- added static methods and removed static instances of classes
- filled in missing information for all classes
- new functions regarding `byteArray` added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@
 - The `repository` section now supports additional VCS types in the form of
   codeberg, forgejo and gitea allowing a `user`, `repo` and additionally a
   `host` url.
+- TypeScript declaration for the prelude exports previously missing functions and classes. Additionally, swaps interfaces for classes and adds missing attributes to classes.
 
 ### Bug fixes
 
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you. Would you mind updating the changelog?

Enderchief:
> Thank you. Would you mind updating the changelog?

Done.

lpil (approved):
Thank you!

PR #2502, opened 2023-12-29
Warn for unused function body

### LINKED ISSUE #2499: Warn for unused function body
A function with an external implementation can still define a default implementation to be used on other targets:
```gleam
@external(javascript, "foo", "bar")
pub fn foo() -> Int {
  // Implementation that is going to be used on the
  // Erlang target.
  11
}
```
When a function defines an external implementation for all the possible targets its body is never going to be used anywhere:
```gleam
@external(javascript, "foo", "bar")
@external(erlang, "foo", "bar")
pub fn foo() -> Int {
  // This is dead code since there already is an
  // external implementation for all targets.
  11
}
```
This could result in an error/warning so that unused code doesn't go undetected. Not sure if it should be an error or warning though, I'm leaning towards warning similarly to unused module functions/constants though

abs0luty:
Resolves #2499


lpil (commented):
Thank you. Could you move this into the inference code please as this is unrelated to parsing. Tests and a changelog entry are needed too

abs0luty:
@lpil the reason why it cannot be moved out of the parsing is that when function body is empty, parser places `todo` expression as a placefolder and inference can't differentiate between empty body and `todo`.

abs0luty:
Wow @lpil very quick response. Sure

abs0luty:
@lpil Added to CHANGELOG, added tests and moved from parser to inference.

compiler-core/src/warning.rs
```diff
@@ -196,6 +196,23 @@ a future version of Gleam."
                 )),
             },
             Self::Type { path, warning, src } => match warning {
+                type_::Warning::UnusedFunctionBody { location } => Diagnostic {
+                    title: "Unused function body.".into(),
```
giacomocavalieri, line 200: To be consistent with #2506 this shouldn't have a full stop at the end of the title

lpil (approved):
Thank you

PR #2505, opened 2023-12-30
JS codegen: export prelude.d.mts in gleam.d.mts

### LINKED ISSUE #2487: JavaScript target: `gleam.d.mts` should export from `prelude.d.mts`?
For example, file content of `util.gleam`:

```gleam
pub fn hello(result: Result(Int, String)) {
  case result {
    Ok(i) -> i
    Error(_) -> 0
  }
}
```

The generated `util.d.mts`.

```typescript
import type * as _ from "./gleam.d.mts";

export function hello(result: _.Result<number, string>): number;
```

There's an error under `_.Result<number, string>`: `Type 'Result' is not generic.deno-ts(2315)`.

It will be OK if changing the content of `gleam.d.mts` from:

```typescript
export * from "../prelude.mjs";
```

to

```typescript
export * from "../prelude.d.mts";
```

I'm not sure if the above change is correct. If it is, the related source code seems to be at [compiler-core/src/codegen.rs#L218](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/codegen.rs#L218):

```rust
 fn write_prelude(&self, writer: &impl FileSystemWriter) -> Result<()> {
    let rexport = format!("export * from \"{}\";\n", self.prelude_location);

    writer.write(&self.output_directory.join("gleam.mjs"), &rexport)?;

    if self.typescript == TypeScriptDeclarations::Emit {
        writer.write(&self.output_directory.join("gleam.d.mts"), &rexport)?; // <==========
    }
    Ok(())
}
```

xhh:
This should fix #2487.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@
 - Fixed a bug with the `isEqual` function in `prelude.js` where RegExps were
   being incorrectly structurally compared and being falsely reported as being
   equal.
+- JavaScript: export from `prelude.d.mts` in `gleam.d.mts` to fix the error:
+  "Type 'Result' is not generic".
 
 
 ## v0.33.0 - 2023-12-18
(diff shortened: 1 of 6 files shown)


lpil (commented):
Neat! Are there tests to update? If not we should add one to prevent a regression

lpil (approved):
Thank you!

PR #2506, opened 2023-12-30
Fix some inconsistencies with error messages

### LINKED ISSUE #2503: Inconsistency of periods in compiler diagnostics
There exists a noticable inconsistency of periods in diagnostic error messages:

# Titles

Period is present here:
```
warning: Unnecessary double negation (!!) on bool.
  ┌─ /src/warning/wrn.gleam:1:25
  │
1 │ pub fn main() { let _ = !!True }
  │                         ^^^^^^

Hint: You can safely remove this.
```

But not, for instance here:
```
warning: Deprecated type used
  ┌─ /src/warning/wrn.gleam:7:22
  │
7 │ pub fn cat_name(cat: Cat) {
  │                      ^^^ This type has been deprecated

It was deprecated with this message: Don't use this!
```

# Hints

Period is present here in this hint:
```
warning: Unnecessary double negation (!!) on bool.
  ┌─ /src/warning/wrn.gleam:1:25
  │
1 │ pub fn main() { let _ = !!True }
  │                         ^^^^^^

Hint: You can safely remove this.
```

But not here:

```
warning: Unnecessary double negation (--) on integer.
  ┌─ /src/warning/wrn.gleam:1:25
  │
1 │ pub fn main() { let _ = --7 }
  │                         ^^^

Hint: You can safely remove this
```

# Label titles

Period is present here in the title of the label:
```
error: Invalid bit array segment
  ┌─ /src/one/two.gleam:1:19
  │
1 │ let x = <<<<1:1>>:bytes>> x
  │                   ^^^^^ This option is only allowed in BitArray patterns.

Hint: This option has no effect in BitArray values.
See: https://gleam.run/book/tour/bit-strings.html
```

But not here:

```
warning: Deprecated value used
   ┌─ /src/warning/wrn.gleam:11:5
   │
11 │     Cat -> "Meow"
   │     ^^^ This value has been deprecated

It was deprecated with this message: The type Animal has been deprecated.
```

# The overall description

The period isn't present here:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:1:5
  │
1 │ foo = 4
  │     ^ There must be a 'let' to bind variable to value

Hint: Use let for binding
See: https://gleam.run/book/tour/let-bindings
```

But is present here:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:1:21
  │
1 │ let x = [1, 2, 3, ..]
  │                     ^ I was expecting a value here

If a list expression has a spread then a tail must also be given.
```

This must be a low priority for now, but just sometimes this sight makes my eyes bleed out.

giacomocavalieri:
This PR closes #2503.

I opted to:
- Never put a full stop `.` after an error/warning title or label pointing to code
- Put a full stop after an error/warning text and hint

Since I was there I've also fixed a couple of typos I found in the messages and rewrote some `format!` calls to do `format("{foo}")` instead of `format("{}", foo)` to be consistent with what other messages do.

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -290,8 +290,7 @@ impl Error {
                 version,
                 source,
             } => format!(
-                "An error occured while trying to retrieve dependencies of {}@{}: {}",
-                package, version, source
+                "An error occured while trying to retrieve dependencies of {package}@{version}: {source}",
             ),
 
             ResolutionError::DependencyOnTheEmptySet {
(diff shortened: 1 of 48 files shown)


abs0luty:
Seems like because of new version of Rust CI is broken. You should fix clippy lints :3

lpil:
Could you rebase on main please? That'll fix the lint issues

abs0luty:
Seems like there are some cases in which diagnostics is baked into some `compiler-core` modules like `parse`: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/parse.rs#L2493.

lpil (approved):
Thank you

Issue #2509, opened 2023-12-30
Error message for using a function in `case` guard clause is unhelpful


erikareads:
Guard clauses (`if` inside a match arm) do not allow functions, the current error message doesn't adequately communicate this to the user:

```rust
import gleam/io
import gleam/string

pub fn main() {
  let my_string = "hello"
  case my_string {
    _ if string.length(my_string) > 2 -> io.debug("doesn't work!'")
  }
}
```

```
error: Syntax error
  ┌─ /src/main.gleam:7:23
  │
7 │     _ if string.length(my_string) > 2 -> io.debug("doesn't work!'")
  │                       ^ I was not expecting this.

Expected one of: 
"->"
Hint: Did you mean to wrap a multi line clause in curly braces?
```

Can we improve this to call out the specific error: that we're trying to use a function in a guard clause?

abs0luty:
I can work on this one!

abs0luty:
Added this error message:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:4:21
  │
4 │   _ if string.length(my_string) > 2 -> io.debug("doesn't work!'")
  │                     ^ Call expression used in case `if` guard clause

Equality and comparison of numbers is all that can be used in an `if` guard clause.
See: https://gleam.run/book/tour/case-expressions.html#checking-equality-and-ordering-in-patterns.
```

kwshi:
What is the rationale for this being disallowed, by the way? This feels natural to have to me.

lpil:
It's not supported by the virtual machine. Largely the only operations it supports are constant time, which helps with how it optimises the performance of pattern matching.

wmstack:
I also think the compiler saying
```
expected one of:
"->"
```
is not reasonable when there is only a single expected item. It should be

```
expected "->"
```

PR #2511, opened 2023-12-31
Add error message for using call expression in case guard clause

abs0luty:
Resolves https://github.com/gleam-lang/gleam/issues/2509.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Language changes
+
+- Added a better error message for situation when user tries to use call expression in case `if` guard.
+
 ### Language Server Changes
 
 - Added a `View on HexDocs` link on function hover
@@ -268,7 +272,7 @@
 - The content has been made wider in rendered HTML documentation.
 - Dependencies that can be built with both `mix` and `rebar3` are now built
   with `mix` if it exists on the system, and with `rebar3` if it doesn't.
-  
+
 ### Bug fixes
 
 - "Compiling $package" is now only printed when a package has new changes to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Language changes
 
-- Added a better error message for situation when user tries to use call expression in case `if` guard.
+- The error message presented when a function is called in a guard has been improved.
 - Warn about function body not being used, because it already has external
   implementations for all targets.
 
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1170,6 +1170,7 @@ where
             let mut estack = vec![];
             let mut last_op_start = 0;
             let mut last_op_end = 0;
+
             loop {
                 if let Some(unit) = self.parse_case_clause_guard_unit()? {
                     estack.push(unit)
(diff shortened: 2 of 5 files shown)


compiler-core/src/parse.rs
```diff
@@ -1213,6 +1214,15 @@ where
                             &mut estack,
                             &do_reduce_clause_guard,
                         );
+                    } else if t == Token::LeftParen {
```
giacomocavalieri, line 1217: So this would give the same error even if it wasn't actually a function call but just an open parentheses?
```gleam
match foo {
  _ if hello( -> todo
//          ^ CallInCaseGuard
}
```
That doesn't seem ideal

abs0luty: Well I think practically this is what mostly corresponds to the user, what can it be other than a call? If no, what do you suggest?

lpil: It should have the previous behaviour in that case

compiler-core/src/parse/snapshots/gleam_core__parse__tests__call_in_case_if_guard.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: compiler-core/src/parse/tests.rs
+expression: "\nlet my_string = \"hello\"\ncase my_string {\n  _ if string.length(my_string) > 2 -> io.debug(\"doesn't work!'\")\n}"
+---
+error: Syntax error
+  ┌─ /src/parse/error.gleam:4:21
+  │
+4 │   _ if string.length(my_string) > 2 -> io.debug("doesn't work!'")
+  │                     ^ Call expression used in case `if` guard clause
```
lpil, line 9: Let's have this point to the whole `string.length(my_string)`

compiler-core/src/parse/error.rs
```diff
@@ -262,6 +267,7 @@ pub enum ParseErrorType {
     ConcatPatternVariableLeftHandSide,
     ListSpreadWithoutTail, // let x = [1, ..]
     ExpectedFunctionBody,  // let x = fn()
+    CallInCaseGuard,       // call expression used in case guard
```
lpil, line 270: Show the syntax in the comment please, to match the other comments

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Language changes
 
+- Added a better error message for situation when user tries to use call expression in case `if` guard.
```
lpil, line 7:
```suggestion
- The error message presented when a function is called in a guard has been improved.
```

compiler-core/src/parse/error.rs
```diff
@@ -210,6 +210,11 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
             ),
             ParseErrorType::UnknownTarget => ("I don't know what this attribute is", vec![]),
             ParseErrorType::ExpectedFunctionBody => ("This function does not have a body", vec![]),
+            ParseErrorType::CallInCaseGuard => (
+                "Call expression used in case `if` guard clause",
+                vec!["Equality and comparison of numbers is all that can be used in an `if` guard clause.
+See: https://gleam.run/book/tour/case-expressions.html#checking-equality-and-ordering-in-patterns.".into()]
```
lpil, line 216: This isn't quite right

```suggestion
                "Unsupported expression",
                vec!["Functions cannot be called in clause guards.".into()]
```

compiler-core/src/parse/tests.rs
```diff
@@ -670,3 +670,15 @@ fn reserved_macro() {
...
+fn call_in_case_if_guard() {
+    // https://github.com/gleam-lang/gleam/issues/2509
+    assert_error!(
+        r#"
+let my_string = "hello"
+case my_string {
+  _ if string.length(my_string) > 2 -> io.debug("doesn't work!'")
+}"#
+    );
+}
```
lpil, line 684: Could you add a test case for `_ if string.length(` to ensure it shows the correct error? Thank you.

PR #2513, opened 2023-12-31
🎨 Formatter improvements!

### LINKED ISSUE #2496: `gleam format` breaks up function call strangely when close to line length
Hello!  Tagging @giacomocavalieri as discussed in the Discord.

The line is formatted like this:

```gleam
let assert Ok(new_stream) =
  stream.new(identifier, settings.initial_window_size, handler, conn,
  )
```

It seems like this should either put the closed paren on the same line, or just break out the arguments into separate lines.  The latter is what happens when I make one of the variable names longer.

Thank you!

### LINKED ISSUE #2507: 🎨 Format tuples with the same rules as function calls.
## Background

Currently a function call with a callback as the final argument will be formatted neatly with all args on one line save for the function body, thanks to @giacomocavalieri's great work:

```gleam
wibble("value", fn(attr) {
  attr
  |> ...
})
```

## Request

I'd like this formatting applied consistently to tuple construction also:

```gleam
#("value", fn(attr) {
  attr
  |> ...
})
```

☝️ currently the above formats like this 👇 

```gleam
#(
  "value",
  fn(attr) {
    attr
    |> ...
  },
)
```

## Context

In [lustre](https://github.com/lustre-labs/lustre) component definitions must supply a dictionary of attribute names to listen to and a decoder to run when those attributes change:

```gleam
fn on_attribute_change() -> Dict(String, Decoder(Msg)) {
  dict.from_list([
    #(
      "value",
      fn(attr) {
        attr
        |> dynamic.int
        |> result.map(Value)
        |> result.map(AttributeChanged)
      },
    ),
  ])
}
```

The recent formatter changes have made lots of common lustre code much more legible but this particular aspect produces quite a lot of vertical noise. For non-trivial components that may want to listen to more than one or two attributes this can end up occupying a lot of space in a module.

### LINKED ISSUE #2512: 🎨 When a list is split into multiple lines its binary operator items get split regardless if they fit or not
As @tynanbe pointed out in #2478 if a binary operator is used in a list that gets split into multiple lines the operator gets split as well, regardless if it fits or not; currently we get this formatting:
```gleam
[
  "Success!"
  |> ansi(apply: [1, 31]),
  "",
  "Wrote `"
  <> bin
  <> "`, `"
  <> pwsh_bin
  <> "`",
]
```
while ideally it should be:
```gleam
[
  "Success!"
  |> ansi(apply: [1, 31]),
  "",
  "Wrote `" <> bin <> "`, `" <> pwsh_bin <> "`",
]
```

giacomocavalieri:
- closes #2512
- closes #2496
- closes #2507

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -877,7 +877,7 @@ impl<'comments> Formatter<'comments> {
         )
     }
 
-    fn tuple<'a>(&mut self, elements: &'a Vec<UntypedExpr>) -> Document<'a> {
+    fn tuple<'a>(&mut self, elements: &'a [UntypedExpr]) -> Document<'a> {
         self.append_inlinable_wrapped_args("#".to_doc(), elements, |e| e, |self_, e| self_.expr(e))
     }
 
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -842,11 +842,7 @@ impl<'comments> Formatter<'comments> {
         }
     }
 
-    fn call<'a>(
-        &mut self,
-        fun: &'a UntypedExpr,
-        args: &'a Vec<CallArg<UntypedExpr>>,
-    ) -> Document<'a> {
+    fn call<'a>(&mut self, fun: &'a UntypedExpr, args: &'a [CallArg<UntypedExpr>]) -> Document<'a> {
         let expr = match fun {
             UntypedExpr::Placeholder { .. } => panic!("Placeholders should not be formatted"),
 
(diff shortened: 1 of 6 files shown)


compiler-core/src/format.rs
```diff
@@ -845,7 +842,11 @@ impl<'comments> Formatter<'comments> {
         }
     }
 
-    fn call<'a>(&mut self, fun: &'a UntypedExpr, args: &'a [CallArg<UntypedExpr>]) -> Document<'a> {
+    fn call<'a>(
+        &mut self,
+        fun: &'a UntypedExpr,
+        args: &'a Vec<CallArg<UntypedExpr>>,
```
lpil, line 848: Why is this `&'a Vec<...>` instead of `&'a [...]` now? It looks like all the pointers to vecs could be slices here.

lpil (commented):
Thank you! One tiny note below

compiler-core/src/format.rs
giacomocavalieri: Oops my bad, fixed it!

lpil (approved):
Thank you

PR #2516, opened 2024-01-01
Get rid of some boilerplate in parser's code

abs0luty:
I don't know whether Ipil will want to merge this so I will wait for his response. I don't want to make an issue for this, because there is a lot more high priority stuff to work on. This didn't take this much time, so I don't really care that much if this will get blocked. But in my opinion, it will be great if this gets merged, as it will make the parser's code a lot easier to go through.


lpil (commented):
Could you revert the changes other than the addition of the `advance` method please? I find the new names less clear. For example `expect` already has a meaning in Rust and it doesn't match what this new one does.

abs0luty:
Sure!

lpil (approved):
Thank you

PR #2519, opened 2024-01-03
change `build` to `build/` in generated .gitignore

### LINKED ISSUE #2517: 🔧 Change `build` in gitignore to `/build`
[lustre_ssg](https://github.com/lustre-labs/lustre_ssg) recommends writing your build script in a file called `build.gleam`. It's not uncommon to want to have some utils/functionality just for build-related things and these naturally end up in a `src/build/` or `test/build/` directory.

The current gitignore means anything in those folders will be ignored and this has tripped me a up a few times before I've realised they're not being tracked!

I think it would be better if the generated `.gitignore` were more specific and only ignored `./build` instead.

steven-mathew:
Fixes #2517

### DIFF
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -121,7 +121,7 @@ pub fn main() {{
             self.root.join(".gitignore"),
             "*.beam
 *.ez
-build
+/build
 erl_crash.dump
 ",
         )


compiler-cli/src/new.rs
```diff
@@ -121,7 +121,7 @@ pub fn main() {{
             self.root.join(".gitignore"),
             "*.beam
 *.ez
-build
+build/
```
lpil, line 124: It needs to match only at the root, so the `/` should be at the start

steven-mathew: Sorry, fixed!

lpil (approved):
Thank you!

PR #2520, opened 2024-01-03
Fix bug with Javascript codegen of function literals

### LINKED ISSUE #2518: 💥 Bug in the javascript codegen
Compiling this gleam code to javascript:

```gleam
pub type Foo {
  Foo(year: Int)
}

pub fn prova(foo: Foo) -> Foo {
  Foo(1)
}

pub fn main() {
  {fn (it) { prova(it).year }}(Foo(10))
}
```

Produces this javascript code:

```js

export class Foo extends $CustomType {
  constructor(year) {
    super();
    this.year = year;
  }
}

export function prova(foo) {
  return new Foo(1);
}

export function main() {
  return (it) => { return prova(it).year; }(new Foo(10));
}
```

Which is wrong, because the lambda `(it) => { return prova(it).year }` isn't wrapped in parenthesis, which results in syntax error.

giacomocavalieri:
This PR closes #2518

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -310,12 +310,14 @@ impl<'module> Generator<'module> {
 
     /// Wrap an expression in an immediately involked function expression if
     /// required due to being a JS statement, or in parens if required due to
-    /// being an operator
+    /// being an operator or a function literal.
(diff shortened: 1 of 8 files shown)


lpil (approved):
Thank you!

PR #2524, opened 2024-01-04
improve error message for unknown types

### LINKED ISSUE #2515: Improve error message when using an imported constructor instead of a type with the same name
In this piece of code we erroneously try to use an imported constructor instead of the homonymous type:
```gleam
import gleam/dynamic.{type Dynamic, DecodeError} as d

pub fn decode(instance_state: Dynamic) -> Result(Int, DecodeError) {
  d.int(instance_state)
}
```
The error message is the following:
```text
error: Unknown type
  ┌─ /tmp/problem/src/debug.gleam:3:55
  │
3 │ pub fn decode(instance_state: Dynamic) -> Result(Int, DecodeError) {
  │                                                       ^^^^^^^^^^^ Did you mean `BitArray`?

The type `DecodeError` is not defined or imported in this module.
```
This is technically correct but not super useful to fix the problem. The error message could be a bit more useful if the compiler can figure out that the `gleam/dynamic` module also exposes a type with the same name and the programmer has already imported the constructor.
```text
error: Unknown type
  ┌─ /tmp/problem/src/debug.gleam:3:66
  │
3 │ pub fn decode(instance_state: Dynamic) -> Result(List(Int), List(DecodeError)) {
  │                                                                  ^^^^^^^^^^^

The type `DecodeError` is not defined or imported in this module.
Hint: Maybe you wanted to import `type DecodeError` from the `gleam/dynamic` module?
```
This is just an idea, I'm not really sure about what the best wording would be here

giacomocavalieri:
This PR closes #2515

### DIFF
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -18,7 +18,6 @@ use super::FieldAccessUsage;
 pub struct UnknownType {
     pub location: SrcSpan,
     pub name: EcoString,
-    pub hint: UnknownTypeHint,
 }
 
 #[derive(Debug, Eq, PartialEq, Clone)]
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -15,6 +15,13 @@ use pretty_assertions::assert_eq;
 use super::FieldAccessUsage;
 
 #[derive(Debug, Eq, PartialEq, Clone)]
+pub struct UnknownType {
+    pub location: SrcSpan,
+    pub name: EcoString,
+    pub hint: UnknownTypeHint,
+}
+
+#[derive(Debug, Eq, PartialEq, Clone)]
 pub enum Error {
     SrcImportingTest {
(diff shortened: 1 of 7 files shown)


compiler-core/src/type_/error.rs
```diff
@@ -14,6 +14,13 @@ use pretty_assertions::assert_eq;
 
 use super::FieldAccessUsage;
 
+#[derive(Debug, Eq, PartialEq, Clone)]
+pub struct UnknownType {
+    pub location: SrcSpan,
+    pub name: EcoString,
+    pub hint: UnknownTypeHint,
+}
```
lpil, line 22: Is this used anywhere?

lpil (commented):
Thank you! One question below

compiler-core/src/type_/error.rs
giacomocavalieri: You're right, it's never used!

lpil (approved):
Thank you!!

PR #2525, opened 2024-01-05
Feat/inlay hints

omprakaash:
Adds basic support for inlay-hints and annotation insertion for types.


lpil:
Hi @omprakaash are you working on this?

omprakaash:
Hi, Yes. Sorry, progress has been slow for a while due to college. I believe I am close to putting this up for review by this week.

compiler-core/src/type_/lsp_pretty.rs
omprakaash, line 1: Created a separate lsp_printer. I feel that this pretty similar to the pretty_printer so used an instance of one as a field here. Not sure if this is a good design decision.

lpil:
Hello! Are you currently working on this?

omprakaash:
Hey yes. Wanted to get this reviewed before potentially going further. Felt like this was getting too large for a single PR.

compiler-core/src/type_/lsp_pretty.rs
lpil: I think this is a good idea 👍

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,113 @@
...
+use ecow::EcoString;
+use im::HashMap;
+use itertools::Itertools;
+
+use crate::{
+    //ast::TypeAst,
+    pretty::{break_, concat, nil, Document, Documentable},
+};
+
+use super::{pretty::Printer, Type, TypeVar};
```
lpil, line 12: I think we can just append to a string as inline hints don't need to be wrapped onto multiple lines. Would make it simpler and faster

lpil (commented):
Had a quick look through and it looks great so far! I can do a thorough review when you feel it's more ready 🙏

compiler-core/src/type_/lsp_pretty.rs
omprakaash: Ah gotcha ! Will change this.

compiler-core/src/language_server/configuration.rs
```diff
@@ -0,0 +1,45 @@
...
+pub struct InlayHintsConfig {
+    #[serde(default = "InlayHintsConfig::default_function_definitions")]
+    pub function_definitions: bool,
+
+    #[serde(default = "InlayHintsConfig::default_module_constants")]
+    pub module_constants: bool,
+
+    #[serde(default = "InlayHintsConfig::default_variable_assignments")]
+    pub variable_assignments: bool,
+}
```
lpil, line 23: Could you document in detail what these do please 🙏

compiler-core/src/language_server/configuration.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl InlayHintsConfig {
+    fn default_function_definitions() -> bool {
+        true
+    }
+    fn default_module_constants() -> bool {
+        true
+    }
+    fn default_variable_assignments() -> bool {
+        true
```
lpil, line 43: I suspect we want this one off by default

compiler-core/src/language_server/engine.rs
```diff
@@ -246,7 +259,17 @@ where
                 return Ok(None);
             };
 
+            let type_qualifiers = get_type_qualifiers(module);
+            let module_qualifiers = get_module_qualifiers(module);
```
lpil, line 263: These are only used in the code action function so let's move these function calls in there.

compiler-core/src/language_server/engine.rs
```diff
@@ -444,6 +453,82 @@ where
...
+
+// Given a module, returns a hashMap mapping between module names and qualified names
+// eg: import mod1 as m => key: mod1, value: m
+fn get_module_qualifiers(module: &Module) -> HashMap<EcoString, EcoString> {
+    let mut module_qualifiers = HashMap::new();
+    for def in &module.ast.definitions {
+        if let Definition::Import(import) = def {
+            if let Some((AssignName::Variable(v), _)) = &import.as_name {
+                let _ = module_qualifiers.insert(import.module.clone(), v.clone());
+            }
```
lpil, line 517: Not import.used_name() here?

compiler-core/src/language_server/server.rs
```diff
@@ -444,6 +531,49 @@ where
...
+                    section: Some("gleam".into()),
+                }],
+            },
+            Some(ResponseHandler::UpdateConfiguration),
+        )
+    }
+
+    /// Response handler for the `workspace/configuration` request
+    fn configuration_update_received(&mut self, result: Json) {
+        if result.is_array() && !result.get(0).is_some_and(|config| config.is_null()) {
```
lpil, line 563: How come we're not using serde here?

compiler-core/src/type_.rs
```diff
@@ -3,6 +3,7 @@ pub(crate) mod error;
 pub(crate) mod expression;
 pub(crate) mod fields;
 pub(crate) mod hydrator;
+pub mod lsp_pretty;
```
lpil, line 6: Move this to the LSP module please

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,93 @@
+use std::sync::Arc;
+
+use ecow::EcoString;
+use im::HashMap;
+use itertools::Itertools;
+
+
+use crate::ast::TypeAst;
+
+use super::{pretty::Printer, Type, TypeVar};
```
lpil, line 10: We can't use the type pretty printer as it adds linebreaks and indentation, we need to have a language server type printer that appends to a string

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,93 @@
...
+            type_parameters
+        }
+    }
+
+    /// Render a Type as a well formatted string.
+    pub fn pretty_print(&mut self, typ: &Type) -> String {
+        self.print(typ)
+    }
+
+    pub fn print(&mut self, typ: &Type) -> String {
```
lpil, line 39: This allocates a very large number of strings and concats them together. Rather than do this we should create one mutable string and append to it

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,93 @@
...
+        match typ {
+            Type::Named {
+                name, args, module, ..
+            } => {
+                let key = module.clone() + "." + name.clone();
```
lpil, line 44: Let's use `(&str, &str)` as the key to avoid the 3 allocations here.

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,93 @@
...
+                let res = self.type_qualifiers.get(&key).unwrap_or(&key).to_string();
+                if args.is_empty() {
+                    if module == "gleam" {
+                        // Ignoring module names for in-built gleam types like Int
+                        name.to_string()
```
lpil, line 49: This could produce incorrect results when the prelude types have been shadowed locally.

compiler-core/src/type_/lsp_pretty.rs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    pub fn print(&mut self, typ: &Type) -> String {
+        match typ {
+            Type::Named {
+                name, args, module, ..
+            } => {
+                let key = module.clone() + "." + name.clone();
+                let res = self.type_qualifiers.get(&key).unwrap_or(&key).to_string();
+                if args.is_empty() {
```
lpil, line 46: We're gunna want to have the correct qualifier even when there are args, no?

compiler-core/src/language_server/type_annotations.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    pub fn into_inlay_hints(self) -> impl Iterator<Item = InlayHint> {
+        self.annotations
+            .into_iter()
+            .map(|(position, text)| InlayHint {
+                position,
+                kind: Some(InlayHintKind::TYPE),
+                tooltip: None,
+                padding_left: text.starts_with(' ').then_some(true),
```
lpil, line 176: What does this mean?

compiler-core/src/language_server/type_annotations.rs
```diff
@@ -0,0 +1,204 @@
...
+
+            Self {
+                annotations: vec![(position, text)],
+            }
+        } else {
+            Self::default()
+        }
+    }
+
+    pub fn extract_type_parameters_from_fn(function: &TypedFunction) -> HashMap<u64, TypeAst> {
```
lpil, line 37: Why do we do this?

lpil (commented):
Thank you. I've left a bunch of notes inline.

This is a very large PR and it is proving hard to merge. Would you like it if we split it into a series of smaller and easier PRs? For example, the annotation printer first, then a code action for annotate assignment, then module constant, etc etc.

omprakaash:
Hey, yes I think it is better if we split this up into separate PRs. What order do you think works best? I think the order you mentioned is good(LSP Printer, variable assignment, module constant etc. ).

compiler-core/src/language_server/configuration.rs
omprakaash: Yep, will do

compiler-core/src/language_server/configuration.rs
omprakaash: Yeah, was just testing this. Will switch it back to false when complete

compiler-core/src/language_server/engine.rs
omprakaash: Got it.

compiler-core/src/type_/lsp_pretty.rs
omprakaash: Gotcha. Will change this.

lpil:
Sounds good to me!

compiler-core/src/language_server/type_annotations.rs
omprakaash: Oh we do this to make a note of all user given type-parameter names anywhere in the function. We call this function before we process a function definition to generate inlay-hints. 
eg:
 `
fn equals(a: value, b) {
    a == b
} // In this case b's type would also be annotated as (value)
`

compiler-core/src/type_/lsp_pretty.rs
omprakaash: Yep !

PR #2526, opened 2024-01-06
Provides argument types when piping to an anonymous function

### LINKED ISSUE #2504: Provide argument type when piping in to an anonymous function
This works fine:

```gleam
  let today =
    birl.now()
    |> birl.get_day()
  let year =
    today.year
    |> int.to_string()
```

But this does not:

```gleam
  let year =
    birl.now()
    |> birl.get_day()
    |> fn (it) { it.year }
    |> int.to_string()
```

You'll get a squiggle under the usage of `it` in the body of the function, because the compiler is failing to deduce the type.

Using that latest of versions of gleam_stdlib and birl off of hex
Using Gleam 0.33

steven-mathew:
Resolves #2504.


compiler-core/src/type_/pipe.rs
```diff
@@ -99,6 +103,25 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
...
+                    is_capture,
+                    location,
+                } if arguments.len() == expected_args.len() => self.infer_apply_to_fn_pipe(
+                    arguments,
+                    &expected_args,
+                    body,
+                    is_capture,
+                    &return_annotation,
+                    location,
+                )?,
```
lpil, line 123: I think this would be best done in the context of function application, not in the sugar rewriting code here. It would keep this code clearer, and it would mean that this works outside of pipes too.

```gleam
  let z = #(1,2)
  fn(x) { x.0 }(z)
```

lpil (commented):
Thank you! Left a note inline

compiler-core/src/type_/pipe.rs
steven-mathew: Thanks, looking into this!

lpil:
Are you still working on this @steven-mathew ?

steven-mathew:
> Are you still working on this @steven-mathew ?

Hi, unfortunately I’ve had a busy school semester and this went over my head. It’s probably best someone else takes on this issue.

lpil:
ok! I'll close this for now and folks who want to pick it up can find the PR linked from the issue 👍 

Thanks again!

PR #2527, opened 2024-01-06
Fix spelling of transitive in warning message

michaeljones:
It is spelled correctly in the enum variant name but not the text of the warning message.


lpil (approved):
Thank you

PR #2528, opened 2024-01-06
Improve formatting of multiline strings in string concatenation

### LINKED ISSUE #2478: Prevent formatter wrapping lines in multi-line string concatenation.
Because Gleam doesn't have string interpolation, the recommended way to inject values into some template is to just use string concatenation.

This means it's plausible to have code that looks something like:

```rs
"pub fn " <> really_long_function_name <> "(" <> lots_of_args <> ") -> " <> a_very_long_return_type <> " {
  ...
}"
```

It's important for legibility that the formatter doesn't start wrapping lines here, because we want to see the template holistically, the same as we would if we had string interpolation. 

In previous versions of Gleam the formatter worked as expected, preserving the formatting above. On 0.33.0, however, the above is now formatted as:

```rs
  "pub fn "
  <> really_long_function_name
  <> "("
  <> lots_of_args
  <> ") -> "
  <> a_very_long_return_type
  <> " {
    ...
  }"
```

This is a real blow for applications that might rely on this sort of injection into templates. I think an interesting rule to explore applying is if any of the strings being concatenated span multiple lines, then the formatter should not attempt to re-arrange things.

giacomocavalieri:
This PR closes #2478

### DIFF
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -5391,6 +5391,25 @@ fn multiline_string_are_not_broken_with_string_concatenation_if_they_fit() {
 }
 
 #[test]
+fn nesting_goes_back_to_normal_after_multiline_string() {
+    assert_format!(
+        r#"pub fn main() {
+  let x = {
+    "
+1
+2
+" <> long_name_function_call(
+      1_111_111_111_111_111,
+      222_222_222_222,
+      3_333_333_333_333_333,
+    )
+  }
+}
+"#
+    );
+}
+
+#[test]
 fn multiline_string_get_broken_on_newlines_as_function_arguments() {
     assert_format!(
         r#"pub fn main() {
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -5377,3 +5377,32 @@ fn function_call_close_to_line_limit() {
 "#
     );
 }
+
+#[test]
+fn multiline_string_are_not_broken_with_string_concatenation_if_they_fit() {
+    assert_format!(
+        r#"pub fn main() {
+  "pub fn foo(" <> arg <> ") ->" <> type_ <> "{
+    body
+}"
+}
+"#
+    );
+}
+
+#[test]
+fn multiline_string_get_broken_on_newlines_as_function_arguments() {
+    assert_format!(
+        r#"pub fn main() {
+  foo(
+    bar,
+    "bar
+  asd
+       baz",
+    foo,
+    bar,
+  )
+}
+"#
+    );
+}
--- a/compiler-core/src/pretty.rs
+++ b/compiler-core/src/pretty.rs
@@ -584,7 +584,7 @@ impl<'a> Document<'a> {
         Self::Group(Box::new(self))
     }
 
-    pub fn set_nest(self, indent: isize) -> Self {
+    pub fn set_nesting(self, indent: isize) -> Self {
         Self::Nest(indent, NestMode::Set, NestCondition::Always, Box::new(self))
     }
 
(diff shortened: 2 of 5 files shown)


compiler-core/src/pretty.rs
```diff
@@ -218,6 +218,22 @@ pub enum NestCondition {
     IfBroken,
 }
 
+/// Used to change the way nesting of documents work.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum NestMode {
```
lpil, line 223: Why is this needed? I couldn't figure out why one couldn't set it back to zero but have the normal indent rules still apply.

lpil (commented):
Thank you! Left a question below.

compiler-core/src/format/tests.rs
```diff
@@ -5377,3 +5377,32 @@ fn function_call_close_to_line_limit() {
...
+    bar,
+    "bar
+  asd
+       baz",
+    foo,
+    bar,
+  )
+}
+"#
+    );
```
lpil, line 5407: Could we have a test for an expression to the right of a multi-line string? To ensure that indentation works after it being set.

```gleam
pub fn main() {
  {
    let x = {
      "
1
2
"
      <> long_name_function_call(
        1111111111111111,
        222222222222,
        3333333333333333,
      )
  }
}
```
Or whatever the correct format is. The important bit is that the level of indentation afterwards is correct.

compiler-core/src/pretty.rs
giacomocavalieri: I'm not sure I get the question right, but this change is needed because before adding this enum everything implicitly worked in the `Increase` mode. Indentation could only be increased, with `nest(doc, 2)` we're saying increase the indentation level by 2 but we have no way to know how much the document is actually going to be indented because it might depend _where_ the document is used.

```rust
// This is pseudo syntax
let doc = group(["foo", break(" ", ""), "bar"])

doc.nest(2).to_string(1)
// foo
//   bar
// ^^---- `.nest(2)` increased the nesting by 2

doc.nest(2).nest(0).to_string(1)
// foo
//   bar
// ^^---- `.nest(0)` increases the nesting by 0
//        so it is a no-op, it doesn't set the
//        indentation back to zero

doc.nest(2).nest(-2).to_string(1)
// foo
// bar
// ^---- this is a hack one can use to reduce nesting
//       but it is unreliable as one would have to know
//       beforehand how much a document is going to be
//       nested
```

Multiline strings require that the indentation of each of their lines is always set to 0 because adding indentation would change the content of the multiline string. The set mode is a way to reliably set the indentation of a piece of document to an exact value:

```rust
doc.set_nesting(0).nest(2).to_string(1)
// foo
// bar
// ^---- no matter how many following nests, the
//       indentation of `doc` will always be 0
//       this is what we need for multiline strings:
//       an outer nest must not change the string's
//       content
```

compiler-core/src/format/tests.rs
giacomocavalieri: Added!

compiler-core/src/pretty.rs
lpil: Sorry, I'm still finding this hard to understand.

Why is it that Elixir doesn't need this algebra extension to do this?

giacomocavalieri: Yes it needs it ([here](https://github.com/elixir-lang/elixir/blob/74bfab8ee271e53d24cb0012b5db1e2a931e0470/lib/elixir/lib/inspect/algebra.ex#L220)), it uses the atom `:reset` passed to the nest function and [it always puts nesting at 0](https://github.com/elixir-lang/elixir/blob/74bfab8ee271e53d24cb0012b5db1e2a931e0470/lib/elixir/lib/inspect/algebra.ex#L1044C12-L1044C12), it would be the same as doing `doc.set_nesting(0)` here.

lpil (approved):
Thanks again!!

PR #2537, opened 2024-01-10
Exporting package metadata as json

### LINKED ISSUE #1475: Generate package info JSON as part of HTML docs
https://github.com/gleam-lang/gleam/blob/b47eb344d9025c82f4f68071d91ac576a35fe6ed/test/package.jsonc

This could be used by external tools to enable searching through documentation, for types, functions, and packages in general.

giacomocavalieri:
This PR closes #1475

### ~~Why is this a draft PR?~~
- [x] I still have to add a bunch of tests to make sure everything works as expected (maybe snapshot tests with insta? Sounds like the best way to do it)
- [x] I still have to understand when this file should be generated and where to put it (should this only be enabled with a flag? Or be always generated together with the html documentation? Should I be able to choose the destination path with a CLI argument?)
- [x] The target tracking should change to take into account pure Gleam functions supported by _all_ modules (even future ones) and those that support all _current_ modules through external implementations
- [x] There's a bug where package documentation is not included in the interface
- [x] I still have to add some documentation


giacomocavalieri:
I think the PR is now ready for review, there's a [point in the tests I added](https://github.com/gleam-lang/gleam/pull/2537/files#diff-cba801b5bd04aeee35b67c4cbb7114e5259da681de23c9e589dedce95d30b84aR32-R80) that I think could be improved: it's a bit of code that is copy pasted and exactly the same for the `erlang` end `javascript` tests as well.
I could make a single function to avoid the code duplication, but I'm not really sure where to put such function

compiler-cli/src/main.rs
```diff
@@ -231,15 +231,22 @@ enum Command {
...
     HexTarball,
-    /// The JavaScript prelude module.
+    /// The JavaScript prelude module
     JavascriptPrelude,
+    /// The TypeScript prelude module
     TypescriptPrelude,
+    /// The package json interface
+    PackageInterface {
+        #[clap(long = "out", required = true)]
+        /// A file to write the json interface to, relative to the package root
```
lpil, line 247:
```suggestion
        /// The path to write the JSON file to
```
Might not be relative

compiler-cli/src/main.rs
```diff
@@ -231,15 +231,22 @@ enum Command {
...
     ErlangShipment,
-    /// The package bundled into a tarball, suitable for publishing to Hex.
+    /// The package bundled into a tarball, suitable for publishing to Hex
     HexTarball,
-    /// The JavaScript prelude module.
+    /// The JavaScript prelude module
     JavascriptPrelude,
+    /// The TypeScript prelude module
     TypescriptPrelude,
+    /// The package json interface
```
lpil, line 244:
```suggestion
    /// Information on the modules, functions, and types in the project in JSON format
```

compiler-core/src/manifest.rs
```diff
@@ -193,7 +193,7 @@ pub enum ManifestPackageSource {
     Local { path: Utf8PathBuf }, // should be the canonical path
 }
 
-fn ordered_map<S, K, V>(value: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::Error>
+pub fn ordered_map<S, K, V>(value: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::Error>
```
lpil, line 196: If this is public it should not be in the manifest module as it's unrelated to the manifest.

compiler-cli/src/main.rs
```diff
@@ -231,15 +231,22 @@ enum Command {
...
+    PackageInterface {
+        #[clap(long = "out", required = true)]
+        /// A file to write the json interface to, relative to the package root
+        output: String,
```
lpil, line 248:
```suggestion
        output: Utf8PathBuf,
```

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+pub enum TypeInterface {
+    Tuple {
+        elements: Vec<TypeInterface>,
+    },
+    Fn {
+        parameters: Vec<TypeInterface>,
+        #[serde(rename = "return")]
+        return_: Box<TypeInterface>,
+    },
+    /// A generic type variable.
```
lpil, line 70:
```suggestion
    /// A type variable.
```

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+            }
+
+            // Private definitions are not included.
+            Definition::Function(_) => {}
+            Definition::CustomType(_) => {}
+            Definition::ModuleConstant(_) => {}
+
+            // Type aliases are ignored since those are also erased from the
+            // generated documentation.
+            Definition::TypeAlias(_) => {}
```
lpil, line 212: Not true, they can have the pub keyword!

compiler-core/src/package_interface/tests.rs
```diff
@@ -0,0 +1,201 @@
...
+        &modules,
+        &TypeWarningEmitter::null(),
+        &direct_dependencies,
+        TargetSupport::Enforced,
+    )
+    .expect("should successfully infer");
+
+    // TODO: all the bits above are basically copy pasted from the javascript
+    // and erlang test helpers. A refactor might be due here, but I'm not sure
+    // where to put such function...
```
lpil, line 80: Yeah we really need something that does all this. So much duplication.

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+                    .iter()
+                    .map(|elem| TypeInterface::from_type(elem.as_ref()))
+                    .collect(),
+            },
+
+            Type::Var { type_ } => match type_.as_ref().try_borrow() {
+                // Package serialisation happens after inference has taken place
+                // At this point no one should be helding a mutable reference of
+                // the type: the borrow should never fail.
+                Err(_) => unreachable!(),
```
lpil, line 243: Use `expect` here, that's how we make assertions about things that must hold true and otherwise be a fatal error.

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+    /// The number of the type's type variables
+    parameters: usize,
+    /// A list of the names of its constructors
+    constructors: Vec<EcoString>,
+}
+
+/// Holds serialisable data about a top-level public value defined in a module
+/// (either a constant or a function)
+#[derive(Serialize, Debug)]
+pub struct ValueInterface {
```
lpil, line 51: Could you include the supported targets too please 🙏 

Need to decide what's a good encoding for "all". Could be a null value means everything, or it could be that there's a "gleam" target which means all (`["gleam", "erlang"]`). Or something else.

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+                body: _,
+                return_annotation: _,
+                external_erlang: _,
+                external_javascript: _,
+            }) => {
+                let _ = values.insert(
+                    name.clone(),
+                    ValueInterface {
+                        supported_targets: supported_targets.to_vec(),
+                        deprecation: deprecation.clone(),
```
lpil, line 192: This should be null if it's not deprecated, not the string "NotDeprecated"

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,280 @@
...
+                    //       definitions.
+                    //       However, `Unbound` and `Generic` ids are generated
+                    //       using the same generator so we have no problem treating
+                    //       unbound variables as generic ones since ids will never
+                    //       overlap.
+                    //       Once #2533 is closed this branch can be turned back to
+                    //       be unreachable!().
+                    TypeVar::Unbound { id } => TypeInterface::Variable { id: *id },
+                    TypeVar::Link { type_ } => TypeInterface::from_type(type_),
+                    TypeVar::Generic { id } => TypeInterface::Variable { id: *id },
```
lpil, line 259: This should use new ints for each value, not the internal ints from the compiler. So if a function's type has 3 variables they will always be 0, 1, and 2.

lpil (commented):
Amazing! Thank you. I've left some notes

compiler-core/src/package_interface.rs
giacomocavalieri: There's a `supported_targets` field in this struct; it is an array of targets so "all" would just become a list with all the targets like this: `["erlang", "javascript"]`

lpil: Oh yes! Sorry I missed that.

It needs to distinguish between there being an external implementation for both and for there being a Gleam implementation as otherwise if a wasm backend arrives then we can no longer tell which are supported, and all packages will need to be republished. I think we'd need to extend the target tracking to have this level of information.

giacomocavalieri: Aha! You're right it'd be quite a problem if we don't track that... I'll have to put this on hold and update the target tracking code first, It shouldn't be too hard

compiler-core/src/package_interface.rs
giacomocavalieri: Ok, I'll include type aliases as well!

compiler-core/src/manifest.rs
giacomocavalieri: What would be a good place to put this function?

lpil: It's annoyingly generic isn't it? Perhaps `io` at a push

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,596 @@
...
+
+use crate::build::{Module, Package};
+
+/// The public interface of a package that gets serialised as a json object.
+#[derive(Serialize, Debug)]
+pub struct PackageInterface {
+    name: EcoString,
+    package_version: Version,
+    /// The minimum Gleam version required by the package.
+    gleam_version: Option<EcoString>,
```
lpil, line 25:
```suggestion
    /// The Gleam version constraint that the package specifies in its `gleam.toml`
    gleam_version_constraint: Option<EcoString>,
```

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// A tuple type like `#(Int, Float)`.
+    Tuple {
+        /// The types composing the tuple.
+        elements: Vec<TypeInterface>,
+    },
+    /// A function type like `fn(Int, String) -> String`.
+    Fn {
+        parameters: Vec<ParameterInterface>,
+        #[serde(rename = "return")]
+        return_: Box<TypeInterface>,
```
lpil, line 223: Labels are not part of the type system so this information cannot be part of the type AST.

The only places one can have a label is on a module function or on a record constructor so I think we'll need to have fields on those that hold this information.

lpil (commented):
Looking really good! Some small comments

giacomocavalieri:
I've addressed the review comments and did some little changes:
- I have [removed the duplicate step in the CI](https://github.com/gleam-lang/gleam/pull/2537/commits/db837c095239e5e14a31d543c2b263c96156257f) as you suggested
- I have changed the `gleam_version` to `gleam_version_constraint` and updated its documentation as you suggested. I've also [changed the `Version` type to `String`](https://github.com/gleam-lang/gleam/pull/2537/commits/d47780114b8561ef60ff8d564e2a5cf165a53403) since the its `Serialize` implementation already turns it into a string and this makes it easier to see it without having to go look for the `Serialize` implementation
- I have defined a `ImplementationsInterface` that replaces `Implementations`, you can read more about my reasoning in the [comments I've added](https://github.com/gleam-lang/gleam/pull/2537/commits/c8e8382e6d18610507aed8ef20b5b4c67a9dc6ca)
- Like you suggested I have [removed labels from the definition of a function type](https://github.com/gleam-lang/gleam/pull/2537/commits/85d96f96dadd8ad61d712f572ecc551735b45405). This also led to updating the module interface to have two listing fields: `constants` and `functions`. And module `functions` now have labelled arguments

lpil:
Q: Why are the version and the Gleam constraint strings rather then their normal types? Wouldn't it be better to use the precise types to ensure they are valid? So far the compiler avoids using strings instead of precise types whenever possible.

giacomocavalieri:
> Q: Why are the version and the Gleam constraint strings rather then their normal types? Wouldn't it be better to use the precise types to ensure they are valid? So far the compiler avoids using strings instead of dedicated types whenever possible.

As for `gleam_version_constraint` I went with a `String` because the compiler itself is using a string and [I don't have access to anything more specific in the Package's `PackageConfig`.](https://github.com/gleam-lang/gleam/pull/2537/files#diff-713f951c0b7f4c757dd5cccc55c5cd1ad26d2ff609cd2cc4cc961e92e9a628a3R331)

As for the `version` field I used a string since that's what it gets serialised into and it makes it easier to see. I agree that it would be better to use a more specific structure though:
```rust
#[derive(Serialize)]
struct VersionInterface {
  major: u32,
  minor: u32,
  patch: u32,
  pre: Vec<PreIdentifier>,
  build: Option<String>,
}

#[derive(Serialize)]
enum PreIdentifier {
  Numeric(u32),
  AlphaNumeric(String),
}
```

CHANGELOG.md
```diff
@@ -7,19 +7,24 @@
...
-- The search bar in generated docs now has a darker background color.
-
 ### Formatter
 
 - Now the formatter will nest pipelines and binary operators that are used as
   function arguments, list items or as tuple items.
 
 ### Build tool
 
+- The search bar in generated docs now has a darker background color.
```
lpil, line 17:
```suggestion
```

compiler-core/src/package_interface/snapshots/gleam_core__package_interface__tests__generic_function.snap
```diff
@@ -0,0 +1,71 @@
+---
+source: compiler-core/src/package_interface/tests.rs
+expression: "\npub type Bar(a) { Bar }\n@deprecated(\"deprecation message\")\npub fn main() { Bar }\n"
+---
+{
+  "name": "my_package",
+  "package_version": {
```
lpil, line 7: This should be `"version": "11.10.9-1foo"`. I think the default encoder for a version uses that format so you could try that.

compiler-core/src/package_interface.rs
```diff
@@ -0,0 +1,680 @@
...
+    type_::{expression::Implementations, Deprecation, Type, TypeVar},
+};
+
+use crate::build::{Module, Package};
+
+/// The public interface of a package that gets serialised as a json object.
+#[derive(Serialize, Debug)]
+pub struct PackageInterface {
+    name: EcoString,
+    package_version: VersionInterface,
```
lpil, line 23:
```suggestion
    version: VersionInterface,
```

lpil (commented):
Could you have the JSON keys use `-` rather than `_` for word delimiters please? As on the web `_` means it's the one word, while `-` means it's two words. Thank you.

giacomocavalieri:
> Could you have the JSON keys use - rather than _ for word delimiters please? As on the web _ means it's the one word, while - means it's two words. Thank you.

Thanks for the review, I wasn't aware of that! Now it should be fixed :)

lpil (approved):
Thank you so much!! 🎉

PR #2539, opened 2024-01-13
A small lexer code refactoring

Animemchik:
When i was reading [parser/lexer.rs](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/parse/lexer.rs), i saw a lot of lines of code, that can be written smaller.
What do u think about it?


lpil (approved):
Nice! Thank you!

PR #2541, opened 2024-01-14
Fix target tracking

giacomocavalieri:
This PR fixes a problem with the current target tracking code. Now the compiler understands the difference between a "pure Gleam" function and a function that happens to support all current targets through FFI:
```gleam
@external(javascript, ...)
@external(erlang, ...)
pub fn one() -> Int

pub fn two() -> Int { 1 + 1 }
```
Under the current implementation, both these two functions would be inferred to support Erlang and Javascript. While that is true for the current version of Gleam, if Gleam were to add more targets (say WASM) `two` would support those automatically; saying it only supports Erlang and Javascript would be wrong.

With this fix a function like `two` would be a "pure Gleam" function and automatically support _all_ targets, even future ones.


compiler-core/schema.capnp
```diff
@@ -92,8 +92,14 @@ struct ValueConstructor {
...
 struct SupportedTargets {
-  erlang @0 :Bool;
-  javascript @1 :Bool;
+  union {
+    gleam @0 :Void;
+
+    externals :group {
+      erlang @1 :Bool;
+      javascript @2 :Bool;
+    }
```
lpil, line 101: Let's not throw away the information about externals when there is a Gleam impl

compiler-core/src/analyse.rs
```diff
@@ -703,7 +703,7 @@ fn external_supported_targets(
     external_javascript: &Option<(EcoString, EcoString)>,
 ) -> SupportedTargets {
     match (external_erlang, external_javascript) {
-        (Some(_), Some(_)) => SupportedTargets::all(),
+        (Some(_), Some(_)) => SupportedTargets::all_externals(),
```
lpil, line 706: I think this has become less clear than this:

```rust
Implementations {
    gleam: has_body,
    erlang: external_erlang.is_some(),
    javascript: external_javascript.is_some(),
}

File: compiler-core/src/metadata/module_decoder.rs [hunk 3]
```diff
@@ -445,14 +445,19 @@ impl ModuleDecoder {
     }
 
     fn supported_targets(&self, reader: supported_targets::Reader<'_>) -> SupportedTargets {
-        let mut supported_targets = SupportedTargets::none();
-        if reader.get_erlang() {
-            supported_targets = supported_targets.add(Target::Erlang);
-        }
-        if reader.get_javascript() {
-            supported_targets = supported_targets.add(Target::JavaScript);
+        match reader.which().expect("supported targets reader") {
```
lpil, line 448: This should return an error, not crash

compiler-core/schema.capnp
```diff
@@ -92,8 +92,14 @@ struct ValueConstructor {
 }
 
 struct SupportedTargets {
-  erlang @0 :Bool;
-  javascript @1 :Bool;
+  union {
```
lpil, line 95: Perhaps this should be called `Implementations` now, since Gleam isn't a target

lpil (commented):
Looks good! Though I think we need to change the design here based off our conversation in discord. I would like us to not throw away the information about externals when there's a Gleam implementation.

giacomocavalieri:
Yeah thank for pointing it out! I'll update it now

compiler-core/src/analyse.rs
```diff
@@ -644,17 +647,9 @@ fn infer_function(
...
     })?;
 
-    if is_placeholder {
-        // If the function has an empty body we are only going to consider as
-        // supported targets the ones that have an external implementation.
-        supported_targets = external_targets;
-    } else {
-        supported_targets = supported_targets.merge(external_targets);
-    }
-
```
giacomocavalieri, line 657: Just a bit of explanation on these changes to make it easier to follow: this piece of target tracking logic was moved from here inside the expression typer. This way everything is done by the typer and the target tracking logic is not split in two different places

compiler-core/src/error.rs
```diff
@@ -2292,11 +2292,9 @@ implementation but the function name `{function}` is not valid."
                 TypeError::UnsupportedTarget {
                     location,
                     target: current_target,
-                    kind,
                 } => {
                     let text = wrap_format!(
-                        "This {} doesn't have an implementation for the {} target.",
-                        kind,
+                        "This value doesn't have an implementation for the {} target.",
```
giacomocavalieri, line 2297: I opted to remove the `kind` (`function`/`constant`) because it felt a bit awkward reading the error message

compiler-core/src/metadata/tests.rs
```diff
@@ -444,7 +444,7 @@ fn module_fn_value_regression() {
                         start: 52,
                         end: 1100,
                     },
-                    supported_targets: SupportedTargets::all(),
```
giacomocavalieri, line 447: This change and the following one was just to have some more variety in the tests

compiler-core/src/type_/expression.rs
```diff
@@ -18,123 +18,184 @@ use im::hashmap;
...
+            gleam: false,
+            erlang: true,
+            javascript: true,
         }
     }
 
+    /// Given the implementations of a function update those with taking into
+    /// account the `implementations` of another function (or constant) used
+    /// inside its body.
+    pub fn update_from_use(&mut self, implementations: &Implementations) {
```
giacomocavalieri, line 126: I removed all other combinators like `merge`/`intersect`/... in favour of just this function that is actually all that is needed by the typer. I think it makes things easier to follow and harder to mess up inadvertently 😁

compiler-core/src/analyse.rs
lpil: Thank you, super helpful!

compiler-core/src/metadata/tests.rs
lpil: I'd prefer new tests for these, to make it clear that it's deliberate with each test only testing one thing.

compiler-core/src/type_/tests/expression.rs
```diff
@@ -0,0 +1,164 @@
...
+                }
+            ),
+            (
+                "pure_gleam",
+                Implementations {
+                    gleam: true,
+                    erlang: true,
+                    javascript: true
+                }
+            )
```
lpil, line 115: This should be only Gleam. `erlang_external_and_pure_body` sets JS to false, `javascript_external_and_pure_body` sets Erlang to false.

compiler-core/src/type_/tests/expression.rs
```diff
@@ -0,0 +1,164 @@
...
+        r#"
+@external(erlang, "foo", "bar")
+@external(javascript, "foo", "bar")
+pub fn all_externals_1() -> Int
+
+pub fn all_externals_2() { all_externals_1() * 2 }
+"#,
+        vec![
+            ("all_externals_1", Implementations::all_externals()),
+            ("all_externals_2", Implementations::all_externals())
```
lpil, line 72: Could we write all the implementation structs in full rather than these helper methods? Sorry I kept having to check what they did again so the tests were a bit hard to read.

compiler-core/src/type_/tests/expression.rs
```diff
@@ -0,0 +1,164 @@
...
+        vec![
+            ("all_externals", Implementations::all_externals()),
+            ("erlang_only", Implementations::erlang()),
+            (
+                "javascript_external_and_erlang_body",
+                Implementations::all_externals()
+            ),
+        ],
+    );
+}
```
lpil, line 164: Can we have a test for one that ends up with all false?

compiler-core/src/type_/expression.rs
```diff
@@ -18,123 +18,184 @@ use im::hashmap;
...
+        // }
+        //
+        // pub fn main() { erlang_only_with_pure_gleam_default() }
+        // ```
+        // Both functions will end up having the following `Implementations`:
+        // `Implementations { gleam: true, erlang: true, javascript: false}`.
+        // They have a pure gleam implementation and an erlang specific external
+        // implementation.
+        self.erlang = self.erlang || *erlang;
+        self.javascript = self.javascript || *javascript;
```
lpil, line 156: I think these should be &&, not ||. If code only supports JavaScript and then calls a function which does support Erlang then it does not start supporting Erlang. The previous code would still mean the function doesn't work on Erlang.

compiler-core/src/type_/expression.rs
```diff
@@ -18,123 +18,184 @@ use im::hashmap;
...
+            javascript: true,
         }
     }
 
+    /// Given the implementations of a function update those with taking into
+    /// account the `implementations` of another function (or constant) used
+    /// inside its body.
+    pub fn update_from_use(&mut self, implementations: &Implementations) {
+        // With this pattern matching we won't forget to deal with new targets
+        // when those are added :)
```
lpil, line 128: Love it!

compiler-core/src/type_/expression.rs
lpil: Yes I agree!

lpil (commented):
This is beautiful work! I love all the comments you've added. Putting me to shame ✨

compiler-core/src/type_/tests/expression.rs
giacomocavalieri: Yes for sure!

compiler-core/src/type_/expression.rs
giacomocavalieri: Ah I think we're interpreting these booleans in a different way.
To me the field `erlang/javascript` set to `true` means that the function relies somewhere on an external Erlang/JavaScript implementation.

I interpret this `Implementations { gleam: true, erlang: true, javascript: false }` as:
- The function has a pure gleam implementation that can be run on any target
- On the Erlang target this function is going to use external code
- On the JavaScript target this function is not going to use external code

And I interpret the following `Implementation { gleam: false, erlang: true, javascript: true }` as:
- The function doesn't have a gleam implementation that can be run on any target
- On the Erlang target it's going to use external code
- On the JavaScript target it's going to use external code
- So it supports all _current_ targets but not future ones

Does that make sense to you?

compiler-core/src/type_/tests/expression.rs
giacomocavalieri: Ah this is a real nice case of boolean blindness, it all comes down to the fact that we're thinking about those booleans in different ways, we can discuss this under [this comment](https://github.com/gleam-lang/gleam/pull/2541#discussion_r1457677570)

compiler-core/src/type_/tests/expression.rs
giacomocavalieri: Depending on how what we un up agreeing [here](https://github.com/gleam-lang/gleam/pull/2541#discussion_r1457677570) it might or might not be possible to have Implementations end up with all fields set to false because the compiler would emit an error earlier.
- If we go along with the current implementation than it's not possible because the only way to get it would be by having a function with no body and no annotations (and that would result in a compilation error)
- Otherwise it might be possible to have that

lpil: It could be a test that results in an error, that's cool, we just need to make sure it covers all the paths.

compiler-core/src/type_/expression.rs
lpil: Ah I see! That seems good to me! Thank you

Maybe we can find names that make this super clear. How about:

```rust
Implementations {
    pure_gleam: bool,
    uses_erlang_externals: bool,
    users_javascript_externals: bool,  
}
```

lpil (commented):
Wonderful, thank you. I've suggested renaming 1 thing

compiler-core/src/type_/expression.rs
giacomocavalieri: Yeah I think that's perfect, thank you!

lpil (approved):
Beautiful work as per usual!!

PR #2542, opened 2024-01-15
Make sure function heads don't go over the line limit

giacomocavalieri:
This fixes a strange behaviour with the formatter where function signatures would sometimes go over the line limit of 80 chars


lpil (approved):
Thank you

PR #2554, opened 2024-01-18
Fix (really) minor date typo in changelog

tupini07:
Unless you traveled to the future, made the commit, and then came back then the release of `v0.34.0-rc1` was set one month ahead 😅


lpil (approved):
Thank you! Our time-travelling Gleam developers must have written that entry

PR #2555, opened 2024-01-18
Fix formatting of pipelines and binary operators

giacomocavalieri:
This PR changes the formatter so that pipelines and binary operators that are used as function arguments, list items or as tuple items get nested. This way it should be easier to tell where an item starts and the other ends.

### DIFF
--- a/compiler-core/src/format/tests/use_.rs
+++ b/compiler-core/src/format/tests/use_.rs
@@ -201,7 +201,7 @@ fn arity_1_var_call() {
         r#"pub fn main() {
   use x, y <- await(
     file.read()
-      |> promise.map(something),
+    |> promise.map(something),
   )
 }
 "#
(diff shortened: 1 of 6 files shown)


compiler-core/src/format/tests/use_.rs
```diff
@@ -214,7 +214,7 @@ fn arity_1_access_call() {
         r#"pub fn main() {
   use x, y <- promise.await(
     file.read()
-    |> promise.map(something),
+      |> promise.map(something),
```
lpil, line 217: Oh dear, I've been shortsighted. I'm not a fan of how it looks with use.

Perhaps your previous suggestion of only doing it when there's multiple arguments is better?

giacomocavalieri: Ok! I've made it so that pipes/binary operators won't get nested if they're the only argument of a function call or the only item of a list/tuple

lpil:
Sorry, would you mind rebasing to resolve the conflict?

Issue #2486, opened 2023-12-23
✨ Introduce an `@internal` attribute for individual declarations.


hayleigh-dot-dev:
## Background

Currently we have the `internal_modules` entry in our `gleam.toml` as a way of excluding certain modules from the generated documentation. Any `pub` code is always importable, but by excluding it from the docs it is de facto internal and not part of the public API.

This solution works well when you have large chunks of internal functionality, but quickly gets clunky when you have small bits of internal functionality next to public code.

It is good practice to define custom types as `pub opaque` unless there is good reason to do otherwise: this means consumers can't pattern match on your custom types and the implementation details remain hidden. Because of this though, it is not possible to have a `pub opaque` type part of the public API and a separate internal module that can pattern match on that type.

## Propsal

I'm proposing an `@internal` attribute that can be attached to declarations that hides that declaration from the generated documentation in an otherwise-public module:

```gleam
pub opaque type Wibble {
  Wibble
  Wobble
}

@internal 
pub fn encode(wibbble: Wibble) -> Json {
  case wibble {
    Wibble -> ...
    Wobble -> ...
  }
} 
```

Here this `encode` function needs access to the structure of the `Wibble` type in order to encode it to JSON properly, but the function itself is not part of the public API. This is handy if we have other parts of a package that might need to know how to encode a Wibble.

---

## Prior art

Elixir has this functionality by setting `@doc false`, see [here](https://hexdocs.pm/elixir/1.12/writing-documentation.html#hiding-internal-modules-and-functions)

lpil:
Seems reasonable to me. We will see if anyone has any thoughts or adjustments or objections.

It should also be excluded from autocompletions from other packages.

massivefermion:
Makes sense to me. Gonna keep my eye on this issue and start working on it when it's final.

erikareads:
I could use this too, I like the current `@internal` proposal.

cauebs:
Won't this be a little bit confusing? First, because it seems somewhat symmetric to `@external` while having nothing to do with it. Second, because then you can have some marked as `pub` that isn't exactly public? Users would then be able to use items that are not documented. Are breaking changes to these items allowed without affecting semver?

Looking at Rust for alternative ideas, there are two things that I understand would satisfy the current proposal's goals: there's the `#[doc(hidden)]` attribute, and there's the `pub(crate)` visibility modifier, both reasonably self-explanatory.

Couldn't we go one of these ways?

[...]

PR #2558, opened 2024-01-21
Add internal annotation

giacomocavalieri:
This PR adds the `@internal` annotation, closing #2486 

### ~~Why is this a WIP PR?~~

- [x] I'm waiting for #2537 to be merged and exclude internals from the package interface as well


compiler-core/src/language_server/engine.rs
```diff
@@ -368,6 +368,9 @@ where
 
         // Module types
         for (name, type_) in &module.ast.type_info.types {
+            if type_.internal {
+                continue;
+            }
```
DanielSherlock, line 373: Are these lines of code excluding `@internal` types from the same file from completions? I don't think we want that.

compiler-core/src/language_server/tests/completion.rs
```diff
@@ -726,3 +726,71 @@ pub fn wibble(
...
+@internal pub type Alias = Int
+@internal pub type Foo { Bar }
+@internal pub const foo = 1
+"#;
+
+    assert_eq!(expression_completions("", code), vec![]);
+}
+
+#[test]
+fn public_internal_definitions_from_the_same_module_are_in_the_completions() {
```
DanielSherlock, line 747: This test seems to only check value-completion, not type-completions. In particular the types `Foo` and `Alias`.

I think these would need a separate test (like `for_custom_type_definition()` at line 568 or the five following tests) which would then detect the I-think-incorrect behaviour in my above comment.

DanielSherlock:
_(I don't rust, so feel free to ignore me completely.)_

I joined the stream late so was curious to look at parts of the code again after, and noticed some things that seem suspicious about completions for types. Hopefully the comments I added are getting at the right bits of code for the issue.

compiler-core/src/language_server/engine.rs
giacomocavalieri: I think you're right! I will look into it
Thanks for pointing it out!!

giacomocavalieri: I checked and you're perfectly right! Thank you so much for the help 💜

giacomocavalieri:
Thank you @DanielSherlock for pointing this out! I've fixed the bug and added some tests to make sure that internal types do appear in the completions for the same module they're defined in

lpil:
Could you rebase on main please? 🙏

compiler-core/src/analyse.rs
```diff
@@ -537,7 +547,14 @@ fn register_value_from_function(
...
     };
-    environment.insert_variable(name.clone(), variant, typ, *public, deprecation.clone());
+    environment.insert_variable(
+        name.clone(),
+        variant,
+        typ,
+        *public,
+        *internal,
+        deprecation.clone(),
+    );
```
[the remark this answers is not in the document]
giacomocavalieri: Ok! Should it be a compile error or should it be a bit more permissive and the formatter could just remove the internal annotation on private functions?

compiler-core/src/language_server/engine.rs
```diff
@@ -382,7 +385,7 @@ where
 
             // Qualified types
             for (name, type_) in &module.types {
-                if !type_.public {
+                if !type_.public || type_.internal {
```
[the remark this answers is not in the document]
giacomocavalieri: Wouldn't removing it show _all_ internal types in the autocompletions - even those defined by external dependencies?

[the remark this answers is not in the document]
lpil: Aye, so it needs to check the package too

compiler-core/src/analyse.rs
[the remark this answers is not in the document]
lpil: Oh yes good point. Perhaps a compile error would be good?

lpil:
Is this WIP @giacomocavalieri ?

giacomocavalieri:
Yes I still have to finish addressing your review comments, I'll be working on this shortly!

compiler-core/src/language_server/engine.rs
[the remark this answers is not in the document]
giacomocavalieri: I'm not sure how I can check if a module come from the root package, how can I do that?

compiler-core/src/analyse.rs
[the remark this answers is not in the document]
giacomocavalieri: Ok! I went with an error

giacomocavalieri:
Turning `public` + `internal` into a single `publicity` attribute turned out to be quite the chunky refactoring but I think this is finally ready for review @lpil 🎉

compiler-core/src/docs.rs
```diff
@@ -164,6 +167,7 @@ pub fn generate_html(
             .ast
             .definitions
             .iter()
+            .filter(|statement| !statement.is_internal())
```
lpil, line 170: nit: is_public() would be a wee bit clearer here

compiler-core/src/docs.rs
```diff
@@ -559,7 +563,7 @@ fn type_<'a>(source_links: &SourceLinker, statement: &'a TypedDefinition) -> Opt
     let mut formatter = format::Formatter::new();
 
     match statement {
-        Definition::CustomType(ct) if ct.public && !ct.opaque => Some(Type {
+        Definition::CustomType(ct) if !ct.publicity.is_private() && !ct.opaque => Some(Type {
```
lpil, line 566: Same here: `.is_public()`

compiler-core/src/javascript/typescript.rs
```diff
@@ -294,41 +294,43 @@ impl<'a> TypeScriptGenerator<'a> {
         match statement {
             Definition::TypeAlias(TypeAlias {
                 alias,
-                public,
+                publicity,
                 type_,
                 ..
-            }) if *public => vec![self.type_alias(alias, type_)],
+            }) if !publicity.is_internal() => vec![self.type_alias(alias, type_)],
```
lpil, line 300: Why is this rendered for public and private but not internal?

compiler-core/src/language_server/engine.rs
```diff
@@ -377,10 +388,13 @@ where
...
-                    continue;
-                };
-                completions.push(type_completion(None, unqualified.used_name(), type_));
+                match module.get_public_type(&unqualified.name) {
+                    Some(type_) if type_.publicity.is_internal() => continue,
+                    Some(type_) => {
+                        completions.push(type_completion(None, unqualified.used_name(), type_))
+                    }
+                    None => continue,
+                }
```
lpil, line 397: In this case they've been imported explicitly so I think we should show them.

compiler-core/src/language_server/engine.rs
```diff
@@ -354,6 +355,8 @@ where
             completions.push(type_completion(None, name, type_));
         }
 
+        let root_package = self.compiler.project_compiler.config.name.clone();
```
lpil, line 358: I don't think we need to clone here, using a `&str` will be fine. A helper function `self.root_package_name()` could be good too

compiler-core/src/language_server/engine.rs
```diff
@@ -392,9 +406,14 @@ where
 
         // Module functions
         for (name, value) in &module.ast.type_info.values {
+            // Here we do not check for the internal attribute: we always want
+            // to show autocompletions for values defined in the same module,
+            // even if those are internal.
             completions.push(value_completion(None, name, value));
         }
 
+        let root_package = self.compiler.project_compiler.config.name.clone();
```
lpil, line 415: Same as above

compiler-core/src/language_server/engine.rs
```diff
@@ -418,10 +445,13 @@ where
 
             // Unqualified values
             for unqualified in &import.unqualified_values {
-                let Some(value) = module.get_public_value(&unqualified.name) else {
-                    continue;
-                };
-                completions.push(value_completion(None, unqualified.used_name(), value));
+                match module.get_public_value(&unqualified.name) {
+                    Some(value) if value.publicity.is_internal() => continue,
```
lpil, line 449: As above, they've been explicitly brought into scope so let's autocomplete them

compiler-core/src/language_server/tests/completion.rs
```diff
@@ -19,13 +23,33 @@ fn positioned_expression_completions(
...
     dep: &str,
     position: Position,
+) -> Vec<CompletionItem> {
+    positioned_expression_completions_with_dep(src, dep, false, position)
+}
+
+/// `external` is used to tell whether the specified dependency has to be
+/// treated as coming from an external imported package or as coming from a
+/// module from the same root package.
+/// In any case `src` will be able to import it using `import dep`.
```
lpil, line 33: External means it's written in FFI, dependency means it comes from another package. I think with the old name it should have always been in a dep package.

compiler-core/src/language_server/tests/completion.rs
```diff
@@ -19,13 +23,33 @@ fn positioned_expression_completions(
...
+    if external {
+        add_path_dep(&mut engine, "dep");
+        let _ = io.path_dep_module("dep", "dep", dep);
+    } else {
+        _ = io.src_module("dep", dep);
+    }
     _ = io.src_module("app", src);
+
     let response = engine.compile_please();
+    println!("{:#?}", response);
```
lpil, line 52: Printing here 😁

lpil (commented):
Nice!! And what an effort 💪  I've left a few small notes inline.

compiler-core/src/language_server/tests/completion.rs
giacomocavalieri: 🫢

compiler-core/src/docs.rs
giacomocavalieri: Aha the problem is that we want to do it with both public and internals, so anything that is not private, how would you call such a method?
```rust
fn wibble(&self) -> bool {
  case self {
    Publicity::Internal | Publicity::Public => true,
    Publicity::Private => false,
  }
}
```

compiler-core/src/javascript/typescript.rs
giacomocavalieri: Oh that's a typo! It was meant to be `!publicity.is_private()`

lpil: oh no! 😁

compiler-core/src/docs.rs
lpil: I realised this for the later ones 😓 

I wonder if `!x.is_internal()` should be `x.is_importable()` or something?

giacomocavalieri: Ah I really like that name, I'll go for it and we can always change it later 😁

compiler-core/src/language_server/tests/completion.rs
giacomocavalieri: The way it was implemented it was just considered as another module under the same root package. I can change it to make sure the dep is always under another package if you want

lpil (approved):
Thank you!!!

PR #2564, opened 2024-01-23
Do not show if an argument is discarded in the docs

### LINKED ISSUE #2561: Do not render arguments as discarded in HTML documentation
The following code produces a warning from the LSP and the compiler:
```gleam
@external(javascript, "./client-component.ffi.mjs", "register")
pub fn register(app: App(Nil, model, msg), name: String) -> Result(Nil, Error) {
  Error(NotABrowser)
}
```

The warning is that `app` and `name` are unused in the body of the function and should be ignored. This is a problem because if suppress this warning by writing `_app` and `_name` _now the generated documentation implies these parameters are unused_:

image

For now I have resigned to just ignoring the warning, but it would be great if it could go away entirely: the argument is used just Gleam can't know about it.

giacomocavalieri:
This PR closes #2561

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@
 - The warning for an unused module alias now shows how to not assign a name to
   the module.
 
+### Build tool changes
+
+- The search bar in generated docs now has a darker background color.
+- The generated docs no longer shows wether an argument is discarded or
+  not in a function signature.
+
 ### Formatter
 
 - Now the formatter will nest pipelines and binary operators that are used as
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 ### Build tool changes
 
 - The search bar in generated docs now has a darker background color.
-- The generated docs no longer shows wether an argument is discarded or
+- The generated docs no longer shows whether an argument is discarded or
   not in a function signature.
 
 ### Formatter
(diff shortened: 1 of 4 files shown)


compiler-core/src/format.rs
```diff
@@ -1267,19 +1267,31 @@ impl<'comments> Formatter<'comments> {
...
 
+    fn docs_fn_arg_name<'a>(&mut self, arg: &'a TypedArg) -> Document<'a> {
+        match &arg.names {
+            ArgNames::Named { name } => name.to_doc(),
+            ArgNames::NamedLabelled { label, name } => docvec![label, " ", name],
+            // We remove the underscore from discarded function arguments since we don't want to
+            // expose this kind of detail: https://github.com/gleam-lang/gleam/issues/2561
+            ArgNames::Discard { name } => name.chars().dropping(1).as_str().to_doc(),
+            ArgNames::LabelledDiscard { label, name } => {
+                docvec![label, " ", name.chars().dropping(1).as_str().to_doc()]
```
lpil, line 1290: Can we use strip_prefix? I think that'll be clearer and remove the risk of accidentally dropping the first char should we ever change whether we store the `_` or not

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ### Build tool changes
 
 - The search bar in generated docs now has a darker background color.
+- The generated docs no longer shows wether an argument is discarded or
```
lpil, line 13:
```suggestion
- The generated docs no longer shows whether an argument is discarded or
```

lpil (commented):
Wonderful! Thank you

CHANGELOG.md
giacomocavalieri: Oops, fixed!

lpil (approved):
Thank you

lpil (approved):
Thank you!

PR #2585, opened 2024-02-07
Do not compile root project if running module from a dependency

### LINKED ISSUE #2568: 🐛 Cannot run a dependency's module if my project can't compile to its target.
## The problem

Lustre ships with it a tiny web server to make onboarding for folks new to lustre/frontend really easy. The web server is written in _erlang_ but a user's project will be targeting _javascript_. If the user's project includes code that is only compilable on the javascript target, attempting to run

```
gleam run -m lustre/try
```

won't work because Gleam attempts to compile _everything_ with the erlang target and the user's project is not compatible.

## What I'd expect

I'd expect Gleam to compile _lustre_ (and any of its dependencies) with the erlang target and leave the users project untouched. As it stands, the majority of lustre projects will never be able to use this feature now because they will have javascript-specific code in them.

giacomocavalieri:
This pr closes #2568


compiler-core/src/build/package_compiler.rs
```diff
@@ -120,32 +135,39 @@ where
...
-            &self.config,
-            self.target.target(),
-            self.mode,
-            &self.ids,
-            loaded.to_compile,
-            existing_modules,
-            warnings,
-            self.target_support,
-        )?;
+        }?;
```
lpil, line 170: Finding this a bit hard to follow as there's lots of indentation. How about instead we only call the package loader if we want to load modules, and use empty vectors otherwise.

compiler-cli/src/build.rs
```diff
@@ -20,6 +20,18 @@ pub fn download_dependencies() -> Result<Manifest> {
 }
 
 pub fn main(options: Options, manifest: Manifest) -> Result<Built> {
+    compile(options, manifest, ModulesCompilation::CompileAll)
+}
+
+pub fn deps_only(options: Options, manifest: Manifest) -> Result<Built> {
+    compile(options, manifest, ModulesCompilation::IgnoreModules)
```
lpil, line 27: This name was confusing to me- it says to ignore modules but it compiles all modules except for the root package. Something like this would make more sense to me at this level:

```rust
pub enum CompilationScope {
    AllModules
    OnlyDependencyModules
}

File: compiler-core/src/build/package_compiler.rs [hunk 3]
```diff
@@ -29,13 +29,26 @@ use camino::{Utf8Path, Utf8PathBuf};
 
 use super::{ErlangAppCodegenConfiguration, TargetCodegenConfiguration, Telemetry};
 
+#[derive(Copy, Clone, Debug)]
+/// A flag that can be used to tell the package compiler to drop all the modules
+/// before performing code generation.
```
lpil, line 34: Can we change this from talking about dropping of modules and codegen to explain what the intent is? I find these docs a bit confusing atm, especially since we really want to not load them at all rather than drop them.

compiler-core/src/build/package_compiler.rs
```diff
@@ -29,13 +29,26 @@ use camino::{Utf8Path, Utf8PathBuf};
...
+/// For example we can use this to compile and run the dependencies of a project
+/// while ignoring the root package.
+pub enum ModulesCompilation {
+    /// Do not ignore the package's modules.
+    CompileAll,
+    /// Drop all the modules of a project before performing code generation.
+    IgnoreModules,
+}
+
```
lpil, line 43: Could remove the repetition of "compile" and "modules" in the names here?
```suggestion
pub enum CompileModules {
    /// Do not ignore the package's modules.
    All,
    /// Drop all the modules of a project before performing code generation.
    None,
}

```

test/run_dependency_with_broken_root/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Further documentation can be found at <https://hexdocs.pm/run_dependency_with_broken_root>.
+
+## Development
+
+```sh
+gleam run   # Run the project
+gleam test  # Run the tests
+gleam shell # Run an Erlang shell
+```
```
lpil, line 25: Could you document what this package is for here please 🙏 

Also could you move this from `test/` to `test/fixtures/` or something as this this package is not a test that can be run- it'll always fail.

test/run_dependency_with_broken_root/test/run_dependency_with_broken_root_test.gleam
```diff
@@ -0,0 +1,12 @@
...
+
+pub fn main() {
+  gleeunit.main()
+}
+
+// gleeunit test functions end in `_test`
+pub fn hello_world_test() {
+  1
+  |> should.equal(1)
+}
```
lpil, line 12: Remove this please, it's unused.

test/run_dependency_with_broken_root/.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+      - uses: actions/checkout@v3
+      - uses: erlef/setup-beam@v1
+        with:
+          otp-version: "26.0.2"
+          gleam-version: "0.34.1"
+          rebar3-version: "3"
+          # elixir-version: "1.15.4"
+      - run: gleam deps download
+      - run: gleam test
+      - run: gleam format --check src test
```
lpil, line 23: Could you delete this file please 🙏

test/run_dependency_with_broken_root/gleam.toml
```diff
@@ -0,0 +1,20 @@
...
+# licences = ["Apache-2.0"]
+# repository = { type = "github", user = "username", repo = "project" }
+# links = [{ title = "Website", href = "https://gleam.run" }]
+#
+# For a full reference of all the available options, you can have a look at
+# https://gleam.run/writing-gleam/gleam-toml/. 
+
+[dependencies]
+gleam_stdlib = "~> 0.34 or ~> 1.0"
+birdie = "~> 1.0"
```
lpil, line 17: This is a lot more code to compile than we need to, which makes it slower to run and introduces more chances of breakages. Let's run gleeunit in the tests instead as it is smaller.

lpil (commented):
Wonderful! Thank you. 💖  I've left some notes inline

compiler-cli/src/build.rs
giacomocavalieri: Yeah that is way clearer, thanks!

compiler-core/src/build/package_compiler.rs
giacomocavalieri: Ok! I've refactored it into a different function, now it should read better :)

giacomocavalieri:
Thanks for the comments! The code is looking much better now 💕
Let me know if it still needs something

lpil:
oh! the integration test is failing atm @giacomocavalieri !

compiler-core/src/build/project_compiler.rs
```diff
@@ -484,7 +500,12 @@ where
         };
         let config_path = package_root.join("gleam.toml");
         let config = PackageConfig::read(config_path, &self.io)?;
-        self.compile_gleam_package(&config, false, package_root)
+
+        // Since we're compiling a dependency package we always want to compile
+        // all its modules and check that it's compiling.
+        // That's why we're not using the project compiler's `compilation_scope`
+        // but we force the `AllModules` compilation scope.
+        self.compile_gleam_package(&config, false, package_root, CompilationScope::AllModules)
```
lpil, line 508: One slight thing here: The second argument indicates whether the package is the root, so it knows enough to know whether it should access `self.compilation_scope` or not.

Further due to there now being two different arguments saying whether the package is the root it's possible for them to disagree and result in a logic bug. If `is_root` is false it should be possible to get to the path where it doesn't compile modules.

In `compile_gleam_package` I'd do something like this:

```rust
        let compile_modules = match compilation_scope {
            CompilationScope::OnlyDependencyModules if is_root => CompileModules::None,
            _ => CompileModules::All,
        };
```

compiler-core/src/build/package_compiler.rs
lpil: Can we not do the thing I asked for with the loader? I would much prefer to not load them at at all instead of loading and then dropping them. I think it would read much clearer too.

giacomocavalieri:
Closing this one, thanks @lpil for picking up the work! 💕

Issue #1972, opened 2023-01-22
Remove all use of `matches!`


lpil:
This match doesn't force the programmer to make a decision when a new variant is added, which resulted in the bug #1971 . Replace it with match expressions to avoid this in future.

lpil:
Yup, the goal is to not have a default pattern, we want to make a decision when a new variant is added. Some uses like `is_variantname` methods is probably OK.

PR #2590, opened 2024-02-09
Permit compilation for unsupported targets when running dependency modules

lpil:
Unfortunately I'm unable to make this PR merge into @giacomocavalieri's branch as it's on a fork. We can just pretend it is I suppose!

Closes https://github.com/gleam-lang/gleam/issues/2568
Closes https://github.com/gleam-lang/gleam/pull/2585

## TODO

- [x] Error if an all-external function doesn't support the current target
- [x] Make an issue about the build cache thing


compiler-core/src/analyse.rs
```diff
@@ -93,6 +93,16 @@ pub enum TargetSupport {
     NotEnforced,
 }
 
+impl TargetSupport {
+    /// Returns `true` if the target support is [`Enforced`].
+    ///
+    /// [`Enforced`]: TargetSupport::Enforced
+    #[must_use]
+    pub fn is_enforced(&self) -> bool {
+        matches!(self, Self::Enforced)
```
giacomocavalieri, line 102: I remember an issue saying we should drop all uses of `matches!` https://github.com/gleam-lang/gleam/issues/1972

giacomocavalieri (approved):
Looks great! Just a small comment

compiler-core/src/analyse.rs
lpil: Oh yes, thank you. Auto-generated code!

PR #2592, opened 2024-02-09
Feat/inline var refactor as code action

katsmil:
Implementation for providing code action hints for inlining variables. Not all types of variables are to be inlined; for instance, a function will not be inlined with this implementation. This is a first step toward providing pipeline refactor suggestions.


compiler-core/src/ast.rs
```diff
@@ -1710,9 +1710,7 @@ pub type UntypedAssignment = Assignment<(), UntypedExpr>;
 
 impl TypedAssignment {
     pub fn find_node(&self, byte_index: u32) -> Option<Located<'_>> {
-        self.pattern
-            .find_node(byte_index)
-            .or_else(|| self.value.find_node(byte_index))
+        self.value.find_node(byte_index)
```
katsmil, line 1713: Can we change it to this?

PR #2601, opened 2024-02-12
Sort imports

### LINKED ISSUE #727: Formatter: Sort imports into alphabetical order
```rust
import b
import c
import a
```

Is formatted as 

```rust
import a
import b
import c
```

giacomocavalieri:
This PR closes #727

The formatter:
- Preserves comments
- Uses comments/empty lines to tell groups apart and sort those separately
- Sorts all imports of each group alphabetically

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -16,7 +16,6 @@ use crate::type_::expression::Implementations;
 use crate::type_::{
     self, Deprecation, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
 };
-use std::cmp::Ordering;
 use std::sync::Arc;
 
 use ecow::EcoString;
@@ -416,10 +415,6 @@ impl<T> Import<T> {
             Some((AssignName::Discard(_), _)) => None,
             None => self.module.split('/').last().map(EcoString::from),
         }
-    }
-
-    pub(crate) fn compare(&self, other: &Self) -> Ordering {
-        self.module.cmp(&other.module)
     }
 
     pub(crate) fn alias_location(&self) -> Option<SrcSpan> {
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -16,6 +16,7 @@ use crate::type_::expression::Implementations;
 use crate::type_::{
     self, Deprecation, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
 };
+use std::cmp::Ordering;
 use std::sync::Arc;
 
 use ecow::EcoString;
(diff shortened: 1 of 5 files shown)


compiler-core/src/ast.rs
```diff
@@ -420,6 +421,20 @@ impl<T> Import<T> {
     pub(crate) fn alias_location(&self) -> Option<SrcSpan> {
         self.as_name.as_ref().map(|(_, location)| *location)
     }
+
+    pub(crate) fn cmp(&self, other: &Self) -> Ordering {
```
lpil, line 425: This name looks like the Ord trait, which is a bit misleading.

What's the special case of `gleam/` for? I couldn't see a test that demonstrates what it does

giacomocavalieri: It's so that it can put `gleam/` imports on top of each group, the test is this one: <https://github.com/gleam-lang/gleam/pull/2601/files#diff-25116aa6d184c5ec1b2f472955472a17d316f0d04d79fa79f0a1de318fd38faaR41-R49>

compiler-core/src/ast.rs
```diff
@@ -417,6 +418,10 @@ impl<T> Import<T> {
         }
     }
 
+    pub(crate) fn compare(&self, other: &Self) -> Ordering {
```
lpil, line 421: Could you remove this please? I think it's clearer what's happening to compare the names in the formatter.

compiler-core/src/format.rs
```diff
@@ -150,21 +157,24 @@ impl<'comments> Formatter<'comments> {
...
-            if documents.is_empty() {
-                // We don't insert empty lines before the first definition
-            } else if previous_was_import && is_import {
-                documents.push(lines(1));
+        // Here we take consecutive groups of imports so that they can be sorted
+        // alphabetically.
+        for (is_import_group, definitions) in &module
+            .definitions
+            .iter()
+            .group_by(|definition| definition.definition.is_import())
```
lpil, line 166: Use `partition` or `partition_in_place` rather than `group_by` in situations like these please

lpil (commented):
Could you include some tests for when there are other non-import statements in with the imports please 🙏

compiler-core/src/format.rs
giacomocavalieri: Wouldn't that hold a different output though? Using partition would yield two collections: one of imports and one with non imports, that means that the formatter would move all imports to the top of the document:
```gleam
import wibble
pub fn wobble()
import wobble
// Turns into:
import wibble
import wobble
pub fn wobble()
```
I did that because the current formatter's behaviour is to keep imports where those are defined and not move them at the top of the file. What should I do?

lpil: ohhhh it produces multiple groups! Sorry this is my mistake, I thought it made `HashMap<bool, Definition>`. OK cool this is great

lpil (approved):
Let's go! 😁

Issue #2586, opened 2024-02-08
`gleam remove pkg` does not work when gleam.toml has dependencies on 1 line


keroami:
```sh
$ cat gleam.toml
name = "hubba"
dependencies = { gleam_erlang = "~> 0.24"}
$ gleam remove gleam_erlang
   Removed gleam_erlang
$ cat gleam.toml
name = "hubba"
dependencies = { gleam_erlang = "~> 0.24"}
```
yet when the dependencies are listed under a header:
```sh
$ cat gleam.toml
name = "hubba"

[dependencies]
gleam_erlang = "~> 0.24"
$ gleam remove gleam_erlang
   Removed gleam_erlang
$ cat gleam.toml
name = "hubba"

[dependencies]
```

Analogous for dev-dependencies.

lpil:
Thank you

PR #2602, opened 2024-02-13
Allow dependency removal from both inline and regular tables

axrez:
Use `as_table_like_mut` in favor `as_table_mut` since the later doesn't allow for modifying inline tables and the first handles both.

Fixes https://github.com/gleam-lang/gleam/issues/2586


lpil (approved):
Thank you so much!!

Issue #2607, opened 2024-02-16
Formatter shouldn't indent an inner item after a comment


tynanbe:
Current formatter (v1.0.0-rc2) produces this

```gleam
  let #(prompt, default, to_bool) = case default {
    True -> #(
      // break it up
        prompt <> " (Y/n)",
      "y",
      fn(x) { !string.starts_with(x, "n") },
    )
    // ..
  }
```

The line after the comment shouldn't be indented.

lpil:
Thank you

PR #2611, opened 2024-02-18
Fix formatter indenting inner item after comment

MarkusPettersson98:
This PR closes https://github.com/gleam-lang/gleam/issues/2607

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -920,9 +920,14 @@ impl<'comments> Formatter<'comments> {
                 // If there's more than one item in the tuple and there's a
                 // pipeline or long binary chain, we want to indent those to
                 // make it easier to tell where one item ends and the other
-                // starts.
+                // starts, except if they are preceded by a comment.
+                let indent = if self_.any_comments(e.location().start) {
+                    0
+                } else {
+                    INDENT
+                };
                 if elements.len() > 1 && (e.is_binop() || e.is_pipeline()) {
-                    self_.expr(e).group().nest(INDENT)
+                    self_.expr(e).group().nest(indent)
                 } else {
                     self_.expr(e).group()
                 }
--- a/compiler-core/src/format/tests/tuple.rs
+++ b/compiler-core/src/format/tests/tuple.rs
@@ -51,3 +51,42 @@ fn tuple_with_last_splittable_arg() {
 "#
     );
 }
+
+// https://github.com/gleam-lang/gleam/issues/2607
+#[test]
+fn inner_items_are_not_indented_after_comment() {
+    assert_format!(
+        r#"fn main() {
+  #(
+    // Don't indent simple expressions following a comment
+    1,
+    2,
+  )
+}
+"#
+    );
+
+    assert_format!(
+        r#"fn main() {
+  #(
+    // Don't indent binary operations following a comment
+    "Should we break up?" <> " (Y/n)",
+    "y",
+  )
+}
+"#
+    );
+
+    assert_format!(
+        r#"fn main() {
+  #(
+    // Keep nesting even if there is a comment here
+    wobble
+    |> fun
+    |> fun2,
+    wabble,
+  )
+}
+"#
+    );
+}


compiler-core/src/format.rs
```diff
@@ -920,11 +920,15 @@ impl<'comments> Formatter<'comments> {
                 // If there's more than one item in the tuple and there's a
                 // pipeline or long binary chain, we want to indent those to
                 // make it easier to tell where one item ends and the other
-                // starts.
-                if elements.len() > 1 && (e.is_binop() || e.is_pipeline()) {
+                // starts, except if they are preceded by a comment.
+                let mut comments = self_.pop_comments(e.location().start).peekable();
+                if comments.peek().is_none()
+                    && elements.len() > 1
+                    && (e.is_binop() || e.is_pipeline())
```
giacomocavalieri, line 927: I'm not sure this is the correct behaviour, we still want things to get nested even if there's a comment. How would this bit of code be indented with this change?
```gleam
wibble(
  // comment
  wobble
    |> fun
    |> fun2,
  wabble,
)
```
If I understand this correctly, with this change, the pipe wouldn't be nested

MarkusPettersson98: Gotcha. I added something equivalent to your code sample as a test case to ensure that we still nest properly.

MarkusPettersson98:
If this behavior is correct, should I add the same behavior for function calls as well?

compiler-core/src/format/tests/tuple.rs
```diff
@@ -51,3 +51,42 @@ fn tuple_with_last_splittable_arg() {
...
+"#
+    );
+
+    assert_format!(
+        r#"fn main() {
+  #(
+    // Keep nesting even if there is a comment here
+    wobble
+    |> fun
+    |> fun2,
```
giacomocavalieri, line 86: This bit here should be nested:
```suggestion
      |> fun
      |> fun2,
```

giacomocavalieri:
> If this behavior is correct, should I add the same behavior for function calls as well?

Thank you so much!! Yeah it would be great if you added the same behaviour to function calls, tuples and lists for consistency.

If it can be of any help I've recently worked on a similar issue and I can help out. The problem arises from the fact that when we have a commented expression (so a list/fn/tuple item) the formatter adds a `force_break` to make sure the comment stays on its own line and is not merged on the previous line. However, this doesn't play nicely with this bit of code that is responsible for indenting pipes and stuff in function calls (and lists/tuples):
```rust
// inside the `tuple` formatting function
if elements.len() > 1 && (e.is_binop() || e.is_pipeline()) {
  self_.expr(e).group().nest(INDENT);
}
```
To see why you can take as an example this tuple and the corresponding document:
```gleam
#(
  // wibble
  wobble
)
```
The document of the tuple's item being:
```rust
// The actual document is a bit more complex but this is all we care about
docvec!("//wibble".force_break(), line(), "wobble").nest(2)
```
So when the formatter breaks it, the `line()` gets nested by two additional spaces resulting in the original bug:
```gleam
// wibble
  wobble // <- the newline is rendered and the nesting is added here
```

The problem is that we only want the commented item to be nested, not the whole thing _including the comment_. The easiest way to do that is to remove the comments manually and add those later _after_ adding the nesting:
```rust
// inside the `tuple` formatting function
if elements.len() > 1 && (e.is_binop() || e.is_pipeline()) {
  // We pop the comments ourselves without relying on the default
  // `expression` way of commenting stuff. That's because we want to
  // avoid the comment block (and its newline) being indented by the
  // formatter. This way the comments gets added _after_ we increase
  // the nesting level.
  let comments = self_.pop_comments(e.start_byte_index());
  let doc = self_.expr(e).group().nest(INDENT);
  commented(doc, comments)
}
```
So now if you have a look at the document generated by the example tuple I showed you earlier you can see how the nesting only gets applied to `"wobble"` and not the whole commented block:
```rust
// The actual document is a bit more complex but this is all we care about
docvec!("//wibble".force_break(), line(), "wobble".nest(2))
```

giacomocavalieri:
Hello @MarkusPettersson98! Are you still planning to work on this? Otherwise I could pick it up for you :)

PR #2620, opened 2024-02-25
Fix a bug where the package interface would not have a module's documentation

giacomocavalieri:
Thank you @DanielleMaywood!!


lpil (approved):
Thank you!

PR #2627, opened 2024-02-29
Properly nest a record update arguments

### LINKED ISSUE #2624: Continuation of property is indented wrong when partly updating record
```gleam
pub type Wibble {
  Wibble(foo: String, bar: String)
}

pub fn wobble() {
  let orig_wibble = Wibble(foo: "1", bar: "2")

  Wibble(
    ..orig_wibble,
    foo: "this is a very long string that will force the formatter"
    <> " to split this into multiple lines",
  )

  Wibble(
    bar: "this won't fail",
    foo: "this is a very long string that will force the formatter"
      <> " to split this into multiple lines",
  )
}
```

The first instance of `<>` is indented to the same level as `foo`, but the second `<>` is indented further. Probably they should both be idented.

Gleam 0.34.1 and 1.0.0-rc2.

giacomocavalieri:
This PR closes #2624

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+### Formatter
+
+- Fixed a bug where a record update's arguments would not be indented correctly.
+
 ### Build tool
 
 - Fixed a bug where the exported package interface would not have a module's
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you

PR #2628, opened 2024-02-29
Do not move last comment out of bounded expressions

### LINKED ISSUE #2608: Formatter shouldn't move a comment from the end of a bounded expression to the beginning of the following expression
```gleam
fn foo() {
  [1, 2,
    // abc
  ]

  case True {
    _ -> Nil
    // def
  }

  #(1, 2,
    // ghi
  )

  case True {
    _ -> #(
      1,
      2,
      // jkl
    )
  }
}
```

Currently formats to:

```gleam
fn foo() {
  [1, 2]

  // abc
  case True {
    _ -> Nil
  }

  // def
  #(1, 2)

  // ghi
  case True {
    _ -> #(1, 2)
  }
  // jkl
}
```

Not sure how it should be improved, but current behaviour is surely undesirable.

giacomocavalieri:
This PR closes #2608

### DIFF
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -5549,3 +5549,40 @@ fn pipeline_inside_tuple_is_not_nested_if_only_item() {
 "#
     );
 }
+
+// github.com/gleam-lang/gleam/issues/2608
+#[test]
+fn comments_are_not_moved_out_of_bounded_expressions() {
+    assert_format!(
+        r#"fn main() {
+  [
+    1, 2,
+    // list
+  ]
+
+  [
+    wibble,
+    wobble,
+    // list
+  ]
+
+  case True {
+    _ -> Nil
+    // case
+  }
+
+  #(
+    1,
+    2,
+    // tuple
+  )
+
+  call(
+    1,
+    2,
+    // function call
+  )
+}
+"#
+    );
+}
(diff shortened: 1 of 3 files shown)


compiler-core/src/format/tests.rs
```diff
@@ -5549,3 +5549,40 @@ fn pipeline_inside_tuple_is_not_nested_if_only_item() {
 "#
     );
 }
+
+// github.com/gleam-lang/gleam/issues/2608
+#[test]
+fn comments_are_not_moved_out_of_bounded_expressions() {
```
lpil, line 5555: Could you make a test per thing being tested please 🙏

lpil (commented):
Fab! One small comment below 🙏

compiler-core/src/format/tests.rs
giacomocavalieri: Sure!

compiler-core/src/format/tests.rs
```diff
@@ -5549,3 +5549,67 @@ fn pipeline_inside_tuple_is_not_nested_if_only_item() {
...
+  ]
+
+  [
+    wibble,
+    wobble,
+    // list
+  ]
+}
+"#
+    );
```
lpil, line 5570: There's still two things being tested in this test!

giacomocavalieri: Oops fixed it!

lpil (approved):
Thank you!

PR #2632, opened 2024-03-02
Adds some documentation improvements on opaqueness

### LINKED ISSUE #2111: Docs reveal opaque type constructors for constants
Not sure if this is really an issue, but it seems like an inconsistency in how opaque types are treated in docs. Example:

```gleam
pub opaque type Secret {
  Number(Int)
}

pub const the_answer = Number(42)
```

This renders in the generated docs as:

```gleam
pub const the_answer: Secret = Number(42)
```

`Number` itself is otherwise not exposed in the docs and cannot be used from outside the module, so this could be confusing.

It could alternatively render like:

```gleam
pub const the_answer: Secret
```

But maybe this would also be confusing? As there is no explicit indication that the reason you're not seeing the value is because it has an opaque type.

floriank:
In an effort to get myself hacking on gleam, I attempted working on #2111. After some back and forth and asking in Discord, I ran with https://github.com/gleam-lang/gleam/pull/2446 and just hid the `const` implementation for the time being.

Since I also dove a bit into CSS, this offers an explicit marker should a type be `opaque`. I _think_ this should improve readability.

Potentially closes #2111.

### DIFF
--- a/compiler-core/templates/docs-css/index.css
+++ b/compiler-core/templates/docs-css/index.css
@@ -527,6 +527,19 @@ body.drawer-open .label-closed {
   margin: 0 0 0 var(--small-gap);
 }
 
+.visibility-tag {
+  background-color: var(--bg-shade-2); 
+  color: var(--text); 
+  padding: 2px 6px; 
+  border-radius: 4px; 
+  border-style: solid;
+  border-width: 1px;
+  border-color: var(--fg-shade-2);
+  font-size: 0.9em; 
+  margin-left: 8px;
+  float: right;
+}
+
 /* Custom type constructors */
 
 .constructor-list {
--- a/compiler-core/templates/docs-css/index.css
+++ b/compiler-core/templates/docs-css/index.css
@@ -540,17 +540,6 @@ body.drawer-open .label-closed {
   float: right;
 }
 
-.visibility-tag:hover::after {
-  content: attr(title);
-  display: block;
-  position: absolute;
-  background-color: var(--bg-shade-3);
-  color: var(--fg-shade-3);
-  padding: 6px; 
-  border-radius: 2px;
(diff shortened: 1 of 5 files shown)


floriank:
Quick screenshots

## Light mode

image

## Dark  mode

image

compiler-core/templates/docs-css/index.css
```diff
@@ -527,6 +527,30 @@ body.drawer-open .label-closed {
...
+.visibility-tag:hover::after {
+  content: attr(title);
+  display: block;
+  position: absolute;
+  background-color: var(--bg-shade-3);
+  color: var(--fg-shade-3);
+  padding: 6px; 
+  border-radius: 2px;
+  margin-top: 6px; 
+}
```
lpil, line 552: What's this bit for? Seem unused, and I think it replicates what the title attribute does by default?

lpil (commented):
Thank you! I've left a question inline.

Would you mind also updating the changelog?

compiler-core/templates/docs-css/index.css
floriank: Ah, that's a leftover from myself playing around with css (haven't done frontend stuff in quite some time).

Will remove shortly ❤️

floriank:
Thanks for the review - I made the changes as requested! @lpil

lpil (approved):
Thank you!!

lpil:
Looks like the tests are failing due to some whitespace having been added.

floriank:
I am not too familiar with the templating here, but I've removed an extra line that was added by ...probably me fat fingering my way through vim :wink:

floriank:
@lpil Tests run locally for me now

lpil:
Thank you!!

Issue #2617, opened 2024-02-23
(Generated docs) Pages generated from markdown have incorrect indentation


MystPi:
It looks like doc pages generated from markdown (such as a package's readme or lustre's new quickstart) have the first indentation level as only one space instead of two in code examples. Other indentation levels appear to be fine.

From Glen's readme:
![image](https://github.com/gleam-lang/gleam/assets/86574651/074f0821-ca5d-4e86-aff6-351ad3a0df69)

From Lustre's quickstart:
![image](https://github.com/gleam-lang/gleam/assets/86574651/4773e56a-9808-470c-8ed0-8c2bd9da3167)

It doesn't look like this happens in pages generated from source files.

lpil:
Thank you

PR #2635, opened 2024-03-03
Fix a bug where a record constructor's argument would not have a doc string

### LINKED ISSUE #2631: Bug in how the doc site is generated
In this snippet of code
```gleam
pub type Wibble {
  Wobble(
    /// Documentation!!
    wabble: Int,
  )
}

pub fn main() {
  io.println("Hello from prova!")
}
```
The generated documentation looks like this. Notice how the doc comment `/// Documentation!` is also erroneously rendered as the main's documentation:
Screenshot 2024-03-02 alle 17 00 18

giacomocavalieri:
This issue closes #2631

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,8 @@
 - Rust's Reqwest's `webpki-roots` are now used for TLS verification.
 - Update Deno config to allow passing `--location` runtime flag.
 - Fixed a bug with dependency resolution of exact versions of RC releases.
+- Fixed a bug where the documentation of a labelled record constructor could be
+  assigned to the wrong definition in the doc site.
 
 ### Language Server
 
@@ -95,7 +97,6 @@
   documentation.
 - Fixed a bug where the `export package interface` command would always
   recompile the project ignoring the cache.
-
 
 ## v1.0.0-rc2 - 2024-02-14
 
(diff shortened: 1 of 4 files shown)


lpil (commented):
The documentation isn't shown at all now. Should we add it to the template?

giacomocavalieri:
I think this is related to that issue I ran into with testing the generated docs I was talking to you the other day on discord. Locally it works fine but it doesn't end up in the test snapshots

lpil:
Sorry, I've forgotten the detail in the rush of the release. What's the problem again?

giacomocavalieri:
So when fixing when fixing [this issue](https://github.com/gleam-lang/gleam/issues/2617) I ran into some strange test behaviour I couldn't fix on my own.

To test that it works I've also updated the test function to include page docs, so far so good.
The problem is that when I run it the generated html has an empty [main element](https://github.com/gleam-lang/gleam/pull/2637/files#diff-affe2e9bf3502ba81b40568370d752e7a613f3599e35f96959b1d35c7ed5749cR226-R230)
I couldn't figure out why is that, what's even more strange is that I tried the fixed version of the compiler locally with make install and using that it works perfectly! My guess is that the test function is missing some important bit that the cli does but I couldn't figure out what.

My guess is that here there's a similar problem, things work perfectly if I install the patched version locally and run `gleam docs build` but somehow those pieces are missing from the test string that is getting snapped

lpil:
Coolio. Is this something you're working on or should I take a look?

giacomocavalieri:
It would be a great help if you could take a look at that 💖
Looking at the test code I couldn't figure it out on my own

giacomocavalieri:
@lpil did you manage to have a look at this?

lpil (approved):
Thank you!

Issue #1970, opened 2023-01-21
Mark modules as stale if they import a module that has since been deleted


lpil:
If the programmer deletes a source file and it is still imported elsewhere the build should fail. This would include external source files, metadata files, and all other files created in the build directory from the source directories.

lpil:
Reopening as I've introduced a bug when importing modules in other packages

PR #2636, opened 2024-03-03
Make sure the export package-interface command uses the project's cache

### LINKED ISSUE #2614: `gleam export package-interface` recompiles project and dependencies every time
When running `gleam export package-interface` it recompiles the project and dependencies from scratch every time, maybe it should be using the project's cache to avoid recompiling everything.

giacomocavalieri:
This PR closes #2614

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Fixed a bug where the exported package interface would not have a module's
   documentation.
+- Fixed a bug where the `export package interface` command would always
+  recompile the project ignoring the cache.
 
 ## v1.0.0-rc2 - 2024-02-14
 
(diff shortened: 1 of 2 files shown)


lpil (commented):
This will make it so deleted modules will show up in the interface. Is that OK?

giacomocavalieri:
Aha! That didn't occur to me, what would you say is best here? It shouldn't be a problem for published packages because the compiler rebuilds everything before publishing, but it could lead to some confusing errors when used locally 🤔

lpil:
Maybe we just accept the bug for now?

giacomocavalieri:
I'd say yeah I don't see that being a big source of bugs and we can always go back to deleting the cache (or finding an actual fix for when modules are deleted) if it turns out to be a bigger problem

I'd open an issue or add a TODO comment in the code to not forget about this, though. I can totally see myself forgetting about this quirk in a week 😂

lpil:
I think this issue will cover it: https://github.com/gleam-lang/gleam/issues/1970

lpil (approved):
Thanks!

PR #2649, opened 2024-03-07
Fix bug in target implementation enforcement

lpil:
This code previously would not result in a compile error on Erlang:

```gleam
import gleam/io

@external(javascript, "test.mjs", "now")
fn now() -> Nil

pub fn main() {
  io.debug(now())
}
```

This was because when checking if the current function had an external implementation we looked at the implementation struct, but this struct was also being used to track whether any of the functions previously called had external implementations, so it was no longer a reliable indicator of externals for the current function.

To fix this I've split it into two structs. One is mutated over time (tracking all referenced functions), one remains the same (the current function's externals).

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -44,6 +44,27 @@ pub struct Implementations {
     pub uses_javascript_externals: bool,
 }
 
+/// Trracking whether the function being currently type checked has externals
+/// implementations or not.
+/// This is used to determine whether an error should be raised in the case when
+/// a value is used that does not have an implementation for the current target.
+#[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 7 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -44,6 +44,27 @@ pub struct Implementations {
     pub uses_javascript_externals: bool,
 }
 
+/// Trracking whether the function being currently type checked has externals
```
giacomocavalieri, line 47: Tiny typo "Tracking"

giacomocavalieri (approved):
This looks great! Having two structures makes it quite easier to follow too

Just noticed a tini tiny typo

lpil:
Thank you

PR #2656, opened 2024-03-08
LSP message receive optimisation

lpil:
With this PR the language server batches changes before compiling rather than compiling once on each update from the client. This reduces the amount of work performed and should help responsiveness on weaker computers where they could struggle to keep up with the demand.


giacomocavalieri (approved):
I'm not really well versed in the LSP parts of the compiler but this looks good to me!

PR #2666, opened 2024-03-11
Add support for FORCE_COLOR

### LINKED ISSUE #2652: Allow forcing colored output
Hey there! I'm new to Gleam and really enjoying the live tour and how the LSP is built right in.

Since we don't have access to a live reload yet, I've been using `watch -n 0.01 -c -x  gleam build` to re-build my project all the time, but I was hoping to enable forcing color, even if Gleam isn't outputting to a terminal (which was added in https://github.com/gleam-lang/gleam/pull/492).

I hoped maybe `FORCE_COLOR=1` would help, but it seems like Gleam doesn't have support for that either.

Thanks for Gleam!

dvjn:
Fixes #2652 

Added support for forcing colored output even if stderr is not connected to terminal using `FORCE_COLOR` environment variable.
Helpful in situations when running a command like `watch gleam build`.

### DIFF
--- a/compiler-cli/src/cli.rs
+++ b/compiler-cli/src/cli.rs
@@ -185,7 +185,7 @@ pub fn stdout_buffer_writer() -> BufferWriter {
 }
 
 fn color_choice() -> ColorChoice {
-    if atty::is(atty::Stream::Stderr) {
+    if atty::is(atty::Stream::Stderr) || std::env::var("FORCE_COLOR").is_ok() {
         termcolor::ColorChoice::Auto
     } else {
         termcolor::ColorChoice::Never
--- a/compiler-cli/src/cli.rs
+++ b/compiler-cli/src/cli.rs
@@ -184,8 +184,16 @@ pub fn stdout_buffer_writer() -> BufferWriter {
     termcolor::BufferWriter::stdout(color_choice())
 }
 
+fn should_force_color() -> bool {
+    if let Ok(force_color_value) = std::env::var("FORCE_COLOR") {
+        !force_color_value.is_empty()
+    } else {
+        false
+    }
+}
+
 fn color_choice() -> ColorChoice {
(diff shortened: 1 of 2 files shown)


dvjn:
Should I add this to `CHANGELOG.md` too?

compiler-cli/src/cli.rs
```diff
@@ -185,7 +185,7 @@ pub fn stdout_buffer_writer() -> BufferWriter {
 }
 
 fn color_choice() -> ColorChoice {
-    if atty::is(atty::Stream::Stderr) {
+    if atty::is(atty::Stream::Stderr) || std::env::var("FORCE_COLOR").is_ok() {
```
Enderchief, line 188: `std::env::var("FORCE_COLOR").is_ok(()` returns `true` when `FORCE_COLOR` is an empty string. It should be `false`.

> When this variable is present and **not an empty string** (regardless of its value), it should force the addition of ANSI color.
	https://force-color.org/

Enderchief (changes requested):

Enderchief:
> Should I add this to `CHANGELOG.md` too?

Probably add it under "Build tool"

compiler-cli/src/cli.rs
lpil: Yes please!

lpil (commented):
Fab, thank you. One note inline and yes please to the changelog entry

compiler-cli/src/cli.rs
dvjn: Done

Enderchief:
@lpil should `should_force_color` and `force_color_value` be the British English spelling or remain American?
Most places in the Gleam source use British Spelling yet for some places like code relating to `NO_COLOR`, the American spelling used. An argument can be made for both.

lpil (approved):
Thank you!!

@Enderchief good shout. I'll adjust the variable.

Issue #2667, opened 2024-03-11
Broken link in string escape error message


bramvbilsen:
Invalid escape codes in string produce the following error:
```
Syntax error

I don't understand this escape code.

Hint: Add another backslash before it.
See: https://gleam.run/book/tour/strings.html#escape-sequences

```
The following link is dead: https://gleam.run/book/tour/strings.html#escape-sequences
I think it should link: https://tour.gleam.run/basics/strings instead

lpil:
Thank you

PR #2668, opened 2024-03-11
Fixed broken link in strings escape code error message

bramvbilsen:
The error message for escape codes in strings included a broken link (https://github.com/gleam-lang/gleam/issues/2667). This replaces that with the correct url.


lpil (approved):
Thank you!!!

PR #2669, opened 2024-03-11
Add nix flake support

ehllie:
Providing a flake for gleam would make it easier for people to contribute to the project, as it provides developers a dev shell with all the necessary dependencies to build it. Additionally, it would let people use the most recent version of gleam, regardless of what their package manager is shipping, by using gleam's flake as an input.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 erl_crash.dump
 *.lock
 node_modules/
+result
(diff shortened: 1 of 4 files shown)


lpil:
Hello! Thank you, but we are unable to support Nix as there's no one on the core team who is a Nix expert, and Nix problems generates a huge amount of work for us, it being the number one cause of support request. Please open this PR with the Nix folks as they are able to support Nix.

PR #2672, opened 2024-03-11
Add support for generating type references (javascript/typescript)

### LINKED ISSUE #2410: Improve generated types' interop with Deno
Currently if you try and import one of the *.mjs files generated by gleam into Deno directly Deno will ignore the corresponding *.d.mts file.  This leads to a lot of type "holes" that get filled with `any`.

The exact issue is [described in the deno docs](https://docs.deno.com/runtime/manual/advanced/typescript/types):

> This can cause a couple problems though. For example, let's say I want to consume a TypeScript file that has already been transpiled to JavaScript along with a type definition file. So I have mod.js and mod.d.ts. If I try to import mod.js into Deno, it will only do what I ask it to do, and import mod.js, but that means my code **won't be as well type checked as if TypeScript was considering the mod.d.ts file in place of the mod.js file**.

The "fix" (though I'm not sure it's fair to call it that, it's not exactly a bug) - is fairly straightforward and outlined [here](https://docs.deno.com/runtime/manual/advanced/typescript/types#using-the-triple-slash-reference-directive).

Basically each *.mjs file should contain a triple slash directive pointing to its corresponding *.d.mts file

```javascript
/// <reference types="./coolLib.d.ts" />
```

Deno should then use this to resolve the types correctly.

I think this is important for compatibility with Deno given Gleam is supporting Deno as a runtime. This will make the developer experience much smoother. When I first noticed this I initially assumed that perhaps Gleam's type generation was just lacking maturity and was a WIP because the language is relatively new, but Gleam's type generation actually seems fine - the concern is people new to Gleam might jump to a similar assumption.

In the meantime for anyone else facing this issue this is the script I'm using as a stopgap (expects gleam's build/ to be in the current directory):

```typescript
import { WalkEntry, expandGlob } from "https://deno.land/std@0.207.0/fs/mod.ts";
import { basename } from 'https://deno.land/std@0.207.0/path/mod.ts'

const files = await expandGlob("./build/**/*.mjs");
const tasks = []

for await (const file of files) {
  tasks.push(fixTypes(file));
}

await Promise.all(tasks);
console.log("Done!")

async function fixTypes(file: WalkEntry) {
  let content = await Deno.readTextFile(file.path);

  const refTypeStr = `/// <reference types="./${basename(file.path, '.mjs')}.d.mts" />`
  if (content.includes(refTypeStr)) {
    console.log(`Skipping ${file.path} already fixed.`)
    return;
  }

  content = `/// <reference types="./${basename(file.path, '.mjs')}.d.mts" /> \n\n${content}`
  await Deno.writeTextFile(file.path, content);

  console.log(`Fixed types for ${file.path}`)
}
```

ajkachnic:
Improves JavaScript codegen by adding [type references](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) to JS files when TypeScript types are generated.

Example:
```javascript
/// <reference types="./hello.d.mts" />

export class Woo extends $CustomType {}

export function wobble() {
  return new Woo();
}
```

References are only generated when TypeScript definitions are enabled

I did my best to follow existing conventions around documents and formatting. Closes #2410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 - Generate [type references](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-)
   when compiling to JavaScript with TypeScript definitions enabled.
 
+
 ### Formatter
 
 - The formatting of case expressions with multiple subjects has been improved.
@@ -47,6 +48,7 @@
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
 
+
 ## v1.0.0 - 2024-03-04
 
 ### Language changes
@@ -73,6 +75,7 @@
 - The format used by the formatter has been improved in some niche cases.
 - Improved the formatting of long case guards.
 - The formatter can now format groups of imports alphabetically.
+
 
 ## v1.0.0-rc1 - 2024-02-10
 
@@ -119,6 +122,7 @@
 - Fixed a bug where external only functions would "successfully" compile for a
   target they do not support, leading to a runtime error.
 
+
 ## v0.34.1 - 2023-01-17
 
 ### Build tool changes
@@ -131,6 +135,7 @@
   format incorrectly.
 - The `@deprecated` attribute can now be used to annotate module constants.
   This will cause a warning to be emitted when the constant is used.
+
 
 ## v0.34.0 - 2023-01-16
 
@@ -149,6 +154,7 @@
 - Fixed a bug where function heads would go over the line limit in the
   formatter.
 
+
 ## v0.34.0-rc2 - 2023-01-11
 
 ### Bug fixes
@@ -158,6 +164,7 @@
 - Fixed a bug where the compiler would in some cases fail to error when an
   application uses functions that do not support the current compilation
   target.
+
 
 ## v0.34.0-rc1 - 2024-01-07
 
@@ -188,7 +195,7 @@
 
 - The `gleam new` command now accepts any existing path, as long as there are
   no conflicts with already existing files. Examples: `gleam new .`, `gleam new
-..`, `gleam new ~/projects/test`.
+  ..`, `gleam new ~/projects/test`.
 - The format for the README created by `gleam new` has been altered.
 - The `gleam.toml` created by `gleam new` now has a link to the full reference
   for its available options.
@@ -222,6 +229,7 @@
   "Type 'Result' is not generic".
 - Not providing a definition after some attributes is now a parse error.
 
+
 ## v0.33.0 - 2023-12-18
 
 ## v0.33.0-rc4 - 2023-12-17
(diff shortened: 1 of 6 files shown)


compiler-core/src/javascript.rs
```diff
@@ -147,14 +164,18 @@ impl<'a> Generator<'a> {
         // Put it all together
 
         if imports.is_empty() && statements.is_empty() {
-            Ok(docvec!("export {}", line()))
+            Ok(docvec!(type_reference, "export {}", line()))
```
Enderchief, line 167: for consistency 
```suggestion
            Ok(docvec![type_reference, "export {}", line()])
```

compiler-core/src/javascript.rs
```diff
@@ -41,13 +42,16 @@ pub struct Generator<'a> {
     module_scope: im::HashMap<EcoString, usize>,
     current_module_name_segments_count: usize,
     target_support: TargetSupport,
+
```
Enderchief, line 45:
```suggestion
```
remove the extra newline

compiler-core/src/javascript.rs
```diff
@@ -41,13 +42,16 @@ pub struct Generator<'a> {
...
+    typescript: TypeScriptDeclarations,
```
Enderchief, line 46:
```suggestion
    typescript: TypeScriptDeclarations,
```
remove the extra newline

Enderchief (commented):
Tested it locally, tests passed and functionality works as expected.

Review is just for stylistic changes for consistency.

compiler-core/src/javascript.rs
```diff
@@ -58,10 +62,23 @@ impl<'a> Generator<'a> {
...
     }
 
+    fn type_reference(&self) -> Document<'a> {
+        if self.typescript == TypeScriptDeclarations::None {
+            return Document::Str("");
+        }
+
+        let name = Document::EcoString(self.module.name.clone());
+
+        docvec!["/// <reference types=\"./", name, ".d.mts\" />", line()]
```
lpil, line 76: What is this relative to? If it's the current file and the name of the module is for example `one/two` this would then point to `one/one/two.d.mts`, which I don't think is right.

lpil (commented):
Great! Thank you. I've left a question inline and would you mind also updating the changelog? 🙏

compiler-core/src/javascript.rs
ajkachnic: Should be fixed now

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
   implementation for the current targeting being compiled for.
 - Fixed a bug where some functions would not result in a compile error when
   compiled for a target that they do not support.
+- Generate [type references](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-) when compiling to JavaScript with TypeScript definitions enabled.
```
lpil, line 12: Could you reset all the formatting changes in this file please

lpil (commented):
Nice, thank you! One tiny comment inline

lpil (approved):
Thank you!!

PR #2678, opened 2024-03-12
Fix hovering over expressions in the middle of pipes

### LINKED ISSUE #2654: LSP: Add hover docs for functions used with the pipe operator
Example: 
This works 
Screenshot 2024-03-07 at 18 45 49

This doesn't work
Screenshot 2024-03-07 at 18 46 11

Acepie:
Closes #2654 

The issue ultimately comes down to the fact that pipelines before the final one are rewritten as assignments to a temp pipe variable. As such the hover behavior treats the thing being hovered over as the pipeline return value instead of the pipeline itself. This doesn't feel like the idiomatic way to solve this (I don't have much rust experience and this feels like something that maybe should be treated differently at the AST level?) but at the very least it provides a nice test case

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
   implementation for the current targeting being compiled for.
 - Fixed a bug where some functions would not result in a compile error when
   compiled for a target that they do not support.
+- Fixed a bug where hovering over an expression in the middle of a pipe would
+  give the wrong node.
 
 ### Formatter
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Fab, thank you. Would you mind updating the changelog also? Thank you

Acepie:
Thanks for taking a look :D changelog is updated. hope this looks good

lpil (approved):
Thank you!

PR #2686, opened 2024-03-12
Improve error message for wrong spread usage

### LINKED ISSUE #2683: Show helpful error message if trying to append to a string using the spread operator
Some newcomers try doing this `[..wibble, item]` to append an item to a list expecting it to work like prepending does. The error message is not particularly useful as it complains about an unexpected token. We could have a better error message that could explain why this is not possible in Gleam

```text
  [..wibble, wobble, wabble]
             ^^^^^^^^^^^^^^
It looks like you're trying to append items to a list using the `..` operator.
Some more in depth explanation
```

How would you implement this Louis? I would like to give this a try but I'm not sure where to start, I was thinking I could check if there's a comma and a `Parser::series_of` elements after the list's tail...

giacomocavalieri:
This PR closes #2683

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -156,6 +156,16 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                     "See: https://tour.gleam.run/basics/lists/".into(),
                 ],
             ),
+            ParseErrorType::ListSpreadFollowedByElements => (
+                "I wasn't expecting elements after this spread",
+                vec![
+                    "A spread can only be used to prepend elements to lists like this: `[first, ..rest]`."
+                        .into(),
+                    "Hint: If you need to append elements to a list you can use `list.append`."
+                        .into(),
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you!!

compiler-core/src/parse/error.rs
```diff
@@ -156,6 +156,16 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                     "See: https://gleam.run/book/tour/lists.html".into(),
                 ],
             ),
+            ParseErrorType::ListSpreadFollowedByElements => (
+                "I wasn't expecting elements after this spread",
+                vec![
+                    "A spread can only be used to prepend elements to lists like this: `[first, ..rest]`."
```
lpil, line 162:
```suggestion
                    "A spread can only be used to prepend elements to lists like this: `[first, ..rest]`.\n"
```

lpil:
Would you mind resolving the conflict? 🙏

PR #2687, opened 2024-03-12
Fixed a bug with JavaScript code generation of pattern matching guards

### LINKED ISSUE #2665: Javascript incorrect pattern matching with if clause containing or operator
Running the following code on javascript and erlang runtime yields different results
```gleam
import gleam/list
import gleam/io
import gleam/bool
import gleam/result

type Struct {
  Struct(optional: Bool, repeated: Bool, value: String)
}

fn insert_struct(structs, struct) {
  let Struct(optional: optional, repeated: repeated, ..) = struct
  let conflict = case list.last(structs) {
    Ok(Struct(repeated: True, ..) as s) -> {
      io.debug(s)
      "Anyting after repeated"
    }
    Ok(Struct(optional: True, ..) as s) if !optional || !repeated -> {
      io.debug(s)
      "Required after optional"
    }
    _ -> {
      io.debug("No conflict found")
      ""
    }
  }

  use <- bool.guard(conflict != "", Error(conflict))

  let structs = list.append(structs, [struct])

  Ok(structs)
}

pub fn main() {
  let structs =
    []
    |> insert_struct(Struct(optional: False, repeated: False, value: "1"))
    |> result.then(insert_struct(_, Struct(
      optional: False,
      repeated: False,
      value: "2",
    )))

  io.debug(structs)
  Nil
}
```
Running this code on erlang produces the following output:
```console
"No conflict found"
"No conflict found"
Ok([Struct(False, False, "1"), Struct(False, False, "2")])
```
While on javascript it produces this:
```console
Nil
Error("Required after optional")
```

giacomocavalieri:
This PR closes #2665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-## Compiler
+### Compiler
 
 - Prepending to lists in JavaScript (`[x, ..xs]` syntax) has been optimised.
 - Function stubs are no longer generated for functions that do not have an
@@ -11,6 +11,7 @@
   compiled for a target that they do not support.
 - Fixed a bug where sometimes a warning would not be emitted when a result is
   discarded.
+- Fixed a bug with JavaScript code generation of pattern matching guards.
 
 ### Formatter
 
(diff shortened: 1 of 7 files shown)


lpil (approved):
Beautiful. Thank you @giacomocavalieri and @ehllie !!

Issue #2655, opened 2024-03-08
LSP: go to definition for items defined in dependencies


ameknite:
I would like to be able to access the code definitions of the standard library directly in vscode.
I think it would be great  to learn how to write idiomatic gleam reading from the implementations of the std library.

PR #2698, opened 2024-03-13
add lsp tests for some basic goto_def cases

Acepie:
This pr just adds unit tests to the existing language server implementation. Was originally thinking about looking into https://github.com/gleam-lang/gleam/issues/2655 but realized I probably need to read more/wanted to start with something smaller so just leaving the test cases here that I saw in the todos (hopefully helps the next person with implementing)

### DIFF
--- a/compiler-core/src/language_server/tests/definition.rs
+++ b/compiler-core/src/language_server/tests/definition.rs
@@ -1,0 +1,367 @@
+use lsp_types::{
+    GotoDefinitionParams, Location, Position, Range, TextDocumentIdentifier,
+    TextDocumentPositionParams, Url,
+};
+
+use super::*;
+
+fn positioned_with_io(
+    src: &str,
+    position: Position,
+    io: &LanguageServerTestIO,
+) -> Option<Location> {
+    let mut engine = setup_engine(io);
+
+    let _ = io.src_module("app", src);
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/tests/definition.rs
```diff
@@ -0,0 +1,367 @@
...
+            TextDocumentIdentifier::new(url),
+            position,
+        ),
+        work_done_progress_params: Default::default(),
+        partial_result_params: Default::default(),
+    };
+    let response = engine.goto_definition(params);
+
+    response.result.unwrap()
+}
```
lpil, line 41: This looks familiar- is it copied from another test module? If so could we make that one public and use it across both?

lpil (commented):
Amazing! Thank you! I've left one note inline

compiler-core/src/language_server/tests/definition.rs
Acepie: Good call, yeah this is effectively copied from hover (I thought I'd have to change it more at first). Lemme separate the engine create into a helper func

Acepie: LMK if this is roughly what you had in mind

lpil (approved):
Thank you so much!

Issue #2653, opened 2024-03-08
LSP: show labelled arguments


ameknite:
Instead of only showing the types
![Screenshot 2024-03-07 at 19 18 47](https://github.com/gleam-lang/vscode-gleam/assets/104745335/09e5f178-f5e7-4f65-a27f-9f533e0a5b67)

show also labelled arguments like in the docs
![Screenshot 2024-03-07 at 19 19 56](https://github.com/gleam-lang/vscode-gleam/assets/104745335/75c1e074-83fb-4919-b8a3-8102cfd94896)

without that you don't know if a function can be used with labels or no, you will need to go to the docs to learn that.

ascandone:
@lpil how would you suggest approaching this problem?

This is the current logic:
```rust
let function_type = Type::Fn {
   args: fun.arguments.iter().map(|arg| arg.type_.clone()).collect(),
   retrn: fun.return_type.clone(),
};
let formatted_type = Printer::new().pretty_print(&function_type, 0);
// and then show formatted_type in the hover docs
```

One could either re-implement an ad-hoc version of the pretty printer that works for functions (referencing directly the last so that it can use the information about the label args), or extend the current type pprinter and include the notion of "labelled args" in the `Type` type (no pun intended), but that might be a little weird as labelled args are not part of the type system.
Do you have something better in mind?

wllfaria:
Hello! I've been really interested in improving the language server for gleam. I was exploring the code in order to implement this, but I have a question that might be due to my lack of context.

As far as I have explored, the hover information for function declarations is done by collecting the `type_` of the arguments, but the label information is contained within arg.names when the arg is a `NamedLabelled`.

After collected, they are formatted to be pretty-printed [(see here)](https://github.com/gleam-lang/gleam/blob/a6ab232e78224bd000b08004dc98ae286b859f32/compiler-core/src/language_server/engine.rs#L517), which takes a `Type`. in which the variants `Var` and `Named` have no field for a Optional label, at least this was the direction I was going. But by adding a new field to some Type variants, many places would require updates, and that didn't seem like a good solution to me.

So I wanted to ask which direction did you expect this to go, or if this is something that you personally haven't checked deeply, I'm open to discuss how this could be implemented.

I'll be looking into other areas of the LSP to improve in the meantime, thanks for your time

lpil:
[...]

PR #2705, opened 2024-03-13
edited hover on function definition

ascandone:


manuraj17:
@ascandone I believe you are tackling this issue https://github.com/gleam-lang/gleam/issues/2653, correct? I happen to be working on that too. Just wanted to check to avoid any confusion.

ascandone:
> @ascandone I believe you are tackling this issue #2653, correct? I happen to be working on that too. Just wanted to check to avoid any confusion.

Ciao! Thank you for writing me
Yeah I started working on it 'cause it looked like a low hanging fruit but then I noticed it required more work that the time I currently have. Feel free to continue on your fork and to take anything from the pr, if useful

PR #2711, opened 2024-03-14
Fix javascript return issue with multistatement blocks

### LINKED ISSUE #2693: Some curly brace blocks return Nil for no apparent reason.
Code:
```gleam
import gleam/io

pub fn main() {
  // prints Nil
  io.debug(foo(10))
  
  // prints Nil
  io.debug({10 {11 12}})
  
  // prints 10
  io.debug({    
    10    
    10
  })
}

fn foo(a) -> Int {
  {
    a
    a 
  }
}
```
Output:
```
Nil
Nil
10
```
Running on the https://tour.gleam.run website.

Acepie:
Analysis in https://github.com/gleam-lang/gleam/issues/2693#issuecomment-1996771438 was dead on. Multistatement blocks did not get wrapped in returns. Added some test cases and a call to wrap_return. Closes #2693

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   the spread syntax (like `[..rest, last]`).
 - Generate [type references](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html#-reference-types-)
   when compiling to JavaScript with TypeScript definitions enabled.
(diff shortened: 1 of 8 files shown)


lpil (approved):
Wow that was fast! Brilliant work, thank you. And thank you for the thorough tests.

PR #2714, opened 2024-03-14
fix issues with blocks returning functions in js

### LINKED ISSUE #2658: Compiler: At least one form of `use` misses a semicolon when compiled to JS
This Gleam
```gleam
fn b() {
  {
    use a <- fn(cb) { cb(1) }
    a
  }
  {
    use b <- fn(cb) { cb(2) }
    b
  }
  3
}
```
Produces this JavaScript with Gleam v1.0.0
```javascript
function b() {
  ((cb) => { return cb(1); })((a) => { return a; })
  ((cb) => { return cb(2); })((b) => { return b; })
  return 3;
}
```
Causing this error
```javascript
  ((cb) => { return cb(2); })((b) => { return b; })
  ^

TypeError: (intermediate value)(...) is not a function
```
Due to a missing semicolon.

Acepie:
Closes #2658 

Issue was ultimately because while functions require semicolons, blocks that happen to return functions do not. If a block returns a function then the function was immediately called on the next value if the value was parenthesised

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -1497,14 +1497,14 @@ fn requires_semicolon(statement: &TypedStatement) -> bool {
             | TypedExpr::NegateBool { .. }
             | TypedExpr::RecordUpdate { .. }
             | TypedExpr::RecordAccess { .. }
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you so much!!

Issue #2697, opened 2024-03-13
Compiler Does Not Suggest Functions from Imported Packages for Unknown Variables/Functions


JS-JeevaSaravanan:
### Description of the Problem:
While working with Gleam version 1.0.0, I attempted to use a function to print a list of numbers to the console but mistakenly omitted the package prefix from the function call. Instead of suggesting the correct function from the imported package, the compiler suggested an unrelated term.

### Code to Reproduce:
```gleam
import gleam/io

pub fn main() {
  let numbers = [1, 2, 3, 4, 5]
  debug(numbers) // Intended to use io.debug but mistakenly omitted 'io.'
}
```

### Expected Behavior:
I expected the compiler to suggest using `io.debug` or another relevant function from the `gleam/io` package since it was imported and contained functions for console output.

### Actual Behavior:
The compiler identified `debug` as an unknown variable and suggested unrelated terms, missing the opportunity to suggest the correct function from an already imported package.

**Error Message**:
```
error: Unknown variable
  ┌─ /path/to/project/src/my_project.gleam:5:3
  │
5 │   debug(numbers)
  │   ^^^^^ Did you mean `True`?
```

### Suggested Improvement:
Enhance the compiler's error suggestions to consider functions available in the imported modules/packages, especially when an unknown function or variable matches a name in those imports. This feature could significantly improve developer experience by providing more contextually relevant suggestions and reducing potential confusion, particularly for newcomers or in cases of simple typos.

### Additional Context:
This issue arose when I tried to print a list of numbers without correctly specifying the `io` module prefix for the `debug` function. The current compiler suggestions did not help in guiding me towards the correct usage of imported functions.

**Gleam Version**: 1.0.0
**Operating System**: macOS Ventura (Version 13.2)

![image](https://github.com/gleam-lang/gleam/assets/59015064/416ccaa9-f464-44dc-8230-fc01270b6d57)

lpil:
Good idea! Thank you

lpil:
Wonderful, thank you.

You'd need to find where the error is emitted and at that point look to see which (if any) of the imported modules have values with that name. If any are found then you can add the names of the modules to the error and then when printing the error include them in the error message in some way.

JS-JeevaSaravanan:
Thank you @lpil and @shellbear for resolving this issue ! Good work guys !

realraphrous:
[...]

PR #2715, opened 2024-03-15
feat(compiler): suggest better suggestions for imports on unknown var…

shellbear:
This improves the overall compiler unknown variables suggestions by listing all imported modules values in the scope so it can now recommends imports.

For example with the following code (described in the following [GitHub issue](https://github.com/gleam-lang/gleam/issues/2697)):
```gleam
import gleam/io

pub fn main() {
  let numbers = [1, 2, 3, 4, 5]
  debug(numbers) // Intended to use io.debug but mistakenly omitted 'io.'
}
```

The current version of the compiler provides the following error:
```gleam
error: Unknown variable
  ┌─ /path/to/project/src/my_project.gleam:5:3
  │
5 │   debug(numbers)
  │   ^^^^^ Did you mean `True`?
```

With my changes the suggestions are a bit smarter:
```gleam
error: Unknown variable
  ┌─ /Users/shellbear/code/gleamio/src/gleamio.gleam:5:3
  │
5 │   debug(numbers) // Intended to use io.debug but mistakenly omitted 'io.'
  │   ^^^^^ Did you mean `io.debug`?

The name `debug` is not in scope here.
```

It's my first contribution to the project so I would be really happy to receive any feedback or improvement. I didn't have a look at the full codebase, it was more a first attempt to implement this improvement and learn more about Gleam compiler. 

Maybe some unit tests for these changes would be great. 

Thanks ✌️


compiler-core/src/analyse.rs
```diff
@@ -1362,7 +1362,7 @@ fn get_type_dependencies(typ: &TypeAst) -> Vec<EcoString> {
             ..
         }) => {
             deps.push(match module {
-                Some(module) => format!("{}.{}", name, module).into(),
+                Some(module) => eco_format!("{}.{}", name, module),
```
lpil, line 1365: Oh wow I had no idea this was a thing!

lpil (commented):
Nice! Thank you! The code looks nice and clean.

We'll need some tests though! To ensure the functionality works and that there are no regressions in future.

I have a feeling that while this works well for `io.debug` it'll not work so well for modules with longer names. The `module.` is going to have a big impact on the edit distance so they won't ever be selected as a suggestion. We can write some tests to verify this, but if it turns out to be the case we'll need to use a different approach, like the one I suggested here <https://github.com/gleam-lang/gleam/issues/2697#issuecomment-1997241885>.

Thanks again! Very nice work

compiler-core/src/type_/environment.rs
```diff
@@ -606,13 +606,36 @@ impl<'a> Environment<'a> {
...
+    // List all possible imports values
+    pub fn imported_modules_values(&self) -> Vec<EcoString> {
+        self.imported_modules
+            .iter()
+            .map(|(module_name, (_, module))| {
+                module
+                    .values
+                    .keys()
+                    .map(|value| format!("{}.{}", module_name, value).into())
+                    .collect::<Vec<EcoString>>()
```
juntuu, line 618:
```suggestion
                    .map(move |value| format!("{}.{}", module_name, value).into())
```

Regarding the ci-lint error about `flat_map`, you can additionally avoid the intermediate `.collect()`s by moving the `module_name` reference into the closure. Its lifetime should last fine for the duration of this whole function (or otherwise the borrow checked would yell).

lpil: Could use that cool eco_format macro too

shellbear: Thanks for the feedback, appreciated. I will update this 🙂

compiler-core/src/analyse.rs
shellbear: I discovered it also by looking at `ecow` crate. Definitely useful and avoid to convert it back to `EcoString`. It can probably be used everywhere in the codebase 🙂

shellbear:
> Nice! Thank you! The code looks nice and clean.
> 
> We'll need some tests though! To ensure the functionality works and that there are no regressions in future.
> 
> I have a feeling that while this works well for `io.debug` it'll not work so well for modules with longer names. The `module.` is going to have a big impact on the edit distance so they won't ever be selected as a suggestion. We can write some tests to verify this, but if it turns out to be the case we'll need to use a different approach, like the one I suggested here [#2697 (comment)](https://github.com/gleam-lang/gleam/issues/2697#issuecomment-1997241885).
> 
> Thanks again! Very nice work

Yes definitely, it works well for this small example but this implementation might quickly shows his limits with more complex examples. I will continue to look on how to improve this and find a better approach that handle more complex suggestions.  If anyone else have some ideas or code solutions, I would be glad to take this as part of this PR.

Thanks 🙂

compiler-core/src/type_/environment.rs
juntuu: I think you missed the main part of my suggestion: removing the `.collect::<Vec<EcoString>>()`. Adding the `move` just enables that :)

shellbear: Updated it, Looks good to you? Thanks :)

lpil:

Issue #2716, opened 2024-03-15
Add link to hex package in generated docs (similar to recent ExDoc versions)


bwireman:
ExDoc now links directly to the hex package page, which makes learning about, installing, and using packages easier, especially when the projects README doesn't link to it or list install instructions

PR #2717, opened 2024-03-15
Add links to hex package in generated docs

bwireman:
Wanted to take a swing at this since I noticed the gap. Not sure if the team wants this or agrees on just adding it to links, but wanted to give it a go. Please lemme know what y'all think

Opened companion issue:
https://github.com/gleam-lang/gleam/issues/2716

### DIFF
--- a/compiler-core/src/docs.rs
+++ b/compiler-core/src/docs.rs
@@ -65,7 +65,12 @@ pub fn generate_html(
         path,
     });
 
-    let links: Vec<_> = doc_links.chain(repo_link).collect();
+    let package_link = Link {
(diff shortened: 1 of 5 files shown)


lpil (commented):
Great idea! Thank you!

Can we add it to the end and do it in the Rust code rather than in the template? Like this: <https://github.com/gleam-lang/gleam/blob/5cb2163a6f03ed4e3f1685023ed2cde041fb7450/compiler-core/src/docs.rs#L63-L68>

Thanks

lpil (approved):
Thank you!

PR #2724, opened 2024-03-15
Improve error messaging around types vs values

### LINKED ISSUE #2712: Improve error message when using type as term
Again, I was playing with Gleam at https://tour.gleam.run/.

I tried to print a number

```
error: Type mismatch
  ┌─ /src/main.gleam:4:14
  │
4 │   io.println(1)
  │              ^

Expected type:

    String

Found type:

    Int
```

no good. It says "Found type: Int", so I tried to print "Int".

```
error: Unknown variable
  ┌─ /src/main.gleam:5:14
  │
5 │   io.println(Int)
  │              ^^^ Did you mean `Nil`?

The name `Int` is not in scope here.

```

Now it says "The name `Int` is not in scope here" even though the previous error message just mentioned "Int".

This is confusing for me because

- I have used languages with dependent types, where `Int` should be viewed as a term there.
- `Nil` is a valid term, but `Int` is not. Why?
- `Integer` is a valid term in Elixir (as atom).

P.S. I have also noticed #2576 while playing with Gleam.

Acepie:
Closes #2712 

Updates some of the error messaging to make more clear when the compiler is expecting a value instead of a type

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1561,21 +1561,29 @@ constructing a new record with its values."
                     location,
                     variables,
                     name,
-                } => Diagnostic {
-                    title: "Unknown variable".into(),
-                    text: wrap_format!("The name `{name}` is not in scope here."),
-                    hint: None,
-                    level: Level::Error,
-                    location: Some(Location {
-                        label: Label {
-                            text: did_you_mean(name, variables),
-                            span: *location,
-                        },
-                        path: path.clone(),
-                        src: src.clone(),
-                        extra_labels: vec![],
-                    }),
-                },
+                    type_with_name_in_scope,
+                } => {
+                    let text = if *type_with_name_in_scope {
+                        wrap_format!("`{name}` is a type, it cannot be used as a value.")
+                    } else {
+                        wrap_format!("The name `{name}` is not in scope here.")
+                    };
+                    Diagnostic {
+                        title: "Unknown variable".into(),
+                        text,
+                        hint: None,
+                        level: Level::Error,
+                        location: Some(Location {
+                            label: Label {
+                                text: did_you_mean(name, variables),
+                                span: *location,
+                            },
+                            path: path.clone(),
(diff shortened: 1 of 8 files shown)


compiler-core/src/error.rs
```diff
@@ -1561,21 +1561,31 @@ constructing a new record with its values."
...
+                            label: Label {
+                                text: label_text,
+                                span: *location,
+                            },
+                            path: path.clone(),
+                            src: src.clone(),
+                            extra_labels: vec![],
+                        }),
+                    }
+                }
```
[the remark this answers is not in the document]
Acepie: Sounds good 👍🏽 I reverted the original text changes and implemented what you suggested (assuming I am reading this right). Lmk if this looks better

compiler-core/src/type_/error.rs
```diff
@@ -521,6 +522,7 @@ pub enum UnknownValueConstructorError {
     Variable {
         name: EcoString,
         variables: Vec<EcoString>,
+        types: Vec<EcoString>,
```
lpil, line 525: This one still has all the types here, let's use `type_with_name_in_scope` too to avoid the extra work and data.

Acepie: whoops thank you 👍🏽

lpil (commented):
Hello! The wasteful extra work is still being done at the other places this error is being constructed. If you delete the `local_type_names` method you'll see all those places.

Could you rebase to remove the conflicts please? We don't permit merge commits in this repo so they cannot be used to resolve conflicts.

Acepie:
ahhh whoops you are right I wasn't paying enough attention on that one. Lemme do a pass on this and push up a fix.

Yup can do (I'm so used to just squashing at work)

Acepie:
Alright should be all fixed up now 👍🏽

lpil (approved):
Brilliant! Thank you!

PR #2727, opened 2024-03-15
Fix erlang code generation for empty list pattern with spread

### LINKED ISSUE #2657: Compiler: `[..]` breaks Erlang
This Gleam
```gleam
fn a() {
  case [] {
    [..] -> 1
    _ -> 2
  }
}
```
Produces this Erlang in Gleam v1.0.0
```erlang
-spec a() -> integer().
a() ->
    case [] of
        [ | _] ->
            1;

        _ ->
            2
    end.
```

Which results in
```erlang
syntax error before: '|'
%   20|         [ | _] ->
%     |           ^
```

Note that the formatter currently converts `[..]` into `_`, sidestepping the issue.

No such problem was observed with the JS target.

Acepie:
Closes #2657

Issue was that we always assumed head has entries if a tail exists when generating. This is true for any validly constructed list but not true for a list matching case expression (i.e. you can't create a non-empty list without a comma but you can create a case pattern matching the rest of a list without accessing any heads technically).

### DIFF
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__case__spread_empty_list.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__case__spread_empty_list.snap
@@ -1,0 +1,15 @@
+---
+source: compiler-core/src/erlang/tests/case.rs
+expression: "\npub fn main() {\n  case [] {\n    [..] -> 1\n  }\n}\n"
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).
+
+-export([main/0]).
+
+-spec main() -> integer().
+main() ->
+    case [] of
+        [_] ->
+            1
+    end.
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__case__spread_empty_list.snap
```diff
@@ -0,0 +1,15 @@
...
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).
+
+-export([main/0]).
+
+-spec main() -> integer().
+main() ->
+    case [] of
+        [_] ->
```
juntuu, line 13: Sorry about the late comment.

This matches a list with exactly one element, not a list with any tail.

I was wondering if it would be sufficient to just return the tail pattern:

```rust
    let elems = if let Some(final_tail) = tail {
        if elems.is_empty() {
            // must not surround with [], but
            // does this need .nest() or .group()?
            return final_tail;
        }
        elems.append(break_(" |", " | ")).append(final_tail)
```

PR #2731, opened 2024-03-16
LSP go to definition supports dependency modules

### LINKED ISSUE #2655: LSP: go to definition for items defined in dependencies

Acepie:
Closes #2655

Change takes the already existing output from compiling the dependencies and stores source data in lsp state. Then when looking for the source info for go to definition calls it just checks against both the original sources from the local module as well as the dependency sources.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
   Neovim.
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
+- Go to definition now works for types in external Gleam modules.
 
 
 ## v1.0.0 - 2024-03-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,7 @@
   Neovim.
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
-- Go to definition now works for types in external Gleam modules.
+- Go to definition now works for values in dependency Gleam modules.
 
 
 ## v1.0.0 - 2024-03-04
--- a/compiler-core/src/language_server/compiler.rs
+++ b/compiler-core/src/language_server/compiler.rs
@@ -31,6 +31,9 @@ pub struct LspProjectCompiler<IO> {
     pub modules: HashMap<EcoString, Module>,
     pub sources: HashMap<EcoString, ModuleSourceInformation>,
 
+    // Information on compiled dependencies.
+    pub dependency_sources: HashMap<EcoString, ModuleSourceInformation>,
+
     /// The storage for the warning emitter.
     pub warnings: Arc<VectorWarningEmitterIO>,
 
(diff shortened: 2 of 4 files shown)


compiler-core/src/language_server/engine.rs
```diff
@@ -165,13 +161,13 @@ where
...
-                        // required LineNumbers and source file path information is
-                        // not stored in the module metadata.
-                        None => return Ok(None),
+                    let module = match (
+                        this.compiler.get_source(name),
+                        this.compiler.get_dependency_source(name),
+                    ) {
+                        (Some(module), _) => module,
+                        (_, Some(module)) => module,
+                        _ => return Ok(None),
```
lpil, line 170: Can you write this as `get_source(name).or_else(|_| get_dep_source(name))` please

compiler-core/src/language_server/compiler.rs
```diff
@@ -31,6 +31,9 @@ pub struct LspProjectCompiler<IO> {
     pub modules: HashMap<EcoString, Module>,
     pub sources: HashMap<EcoString, ModuleSourceInformation>,
 
+    // Information on compiled dependencies.
+    pub dependency_sources: HashMap<EcoString, ModuleSourceInformation>,
```
lpil, line 35: Why is it that there's a second hashmap for dependencies rather than all being in one?

CHANGELOG.md
```diff
@@ -53,6 +53,7 @@
   Neovim.
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
+- Go to definition now works for types in external Gleam modules.
```
lpil, line 56: This says types, but the test is for a value. Could we have a test for types too in that case.

---


External modules are Erlang or JavaScript modules
```suggestion
- Go to definition now works for types in dependency Gleam modules.
```

lpil (commented):
Nice! Thank you! I've left some notes in line.

compiler-core/src/language_server/compiler.rs
Acepie: Good point. I kinda just assumed we were using it in more places/would need to differentiate elsewhere but I guess not. Lemme just combine into 1

CHANGELOG.md
Acepie: This was a mistype on my part. It looks like this will only work for values (as far as I can tell goto type does not work for the actual type even in the same file rn)

compiler-core/src/language_server/compiler.rs
lpil: That might be the case! I'm not sure.

CHANGELOG.md
Acepie: But lemme add a test for record constructors/constants/etc at least to show things besides functions

lpil: Fab! Thank you

lpil (approved):
Fantastic!! Thank you very much

PR #2736, opened 2024-03-16
Fix parenthesised expr return

### LINKED ISSUE #2735: JavaScript codegen: incorrectly parenthesised return statement
Some expressions get parenthesised in the code generation. When such expression is the only one in a block, and also in tail position (e.i. returned), the return statement is parenthesised incorrectly.

For example this gleam code

```gleam
fn b() {
  {
    1 + 2
  }
}
```

produces the following JavaScript, which is syntactically incorrect

```js
function b() {
  (return 1 + 2;)
}
```

juntuu:
Attempt to fix #2735


lpil (approved):
Thank you!

PR #2737, opened 2024-03-16
Refactor: use pretty::join consistently

juntuu:
I noticed quite a few places doing `concat(Itertools::intersperse(docs, separator))`, which is what the [pretty::join()](https://github.com/gleam-lang/gleam/blob/0b1062543eb721893f6c9d8ad6ddb402d883c4e5/compiler-core/src/pretty.rs#L136) function does. 

Here I used the function for all those cases I could find. Remaining uses of `Itertools::intersperse` collect the result some other way than with `pretty::concat()`.


lpil (approved):
Thank you

PR #2743, opened 2024-03-17
Fix generated files in `.gitattributes`

sobolevn:
I think that `src/bindata_capnp.rs` is outdated and is replaced with `compiler-core/generated/schema_capnp.rs`.

I've reused our `generated` attribute from CPython: https://github.com/python/cpython/blob/b8d808ddd77f84de9f93adcc2aede2879eb5241e/.gitattributes#L66C1-L66C55

It is useful in a case that generated file are folded by default during reviews, you can still look through them, but noise is reduced.

I've also found one more generated file that is test-related.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,7 @@
+[attr]generated linguist-generated=true diff=generated
+
 # Generate Cap'n Proto code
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you

PR #2745, opened 2024-03-17
Warn for pattern matching on tuple literals

giacomocavalieri:


compiler-core/src/warning.rs
```diff
@@ -598,6 +598,25 @@ the same values.\n"
                         }),
                     }
                 }
+
+                type_::Warning::CaseMatchOnLiteralTuple { location } => Diagnostic {
+                    title: "Case expression with literal tuple as subject".into(),
```
lpil, line 603: How about `Redundant tuple`?

lpil (commented):
Nice, thank you! Looks like the build is faiing

compiler-core/src/warning.rs
giacomocavalieri: Way better! Thanks for the suggestion 💖

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
   returned by blocks.
 - Fix a bug where Erlang code generation would not properly handle list case patterns
   with no head and a spread tail.
+- The compiler will now raise a warning ig you're pattern matching on tuple
```
lpil, line 26:
```suggestion
- The compiler will now raise a warning if you're pattern matching on tuple
```

lpil (approved):
Thank you!

Issue #1980, opened 2023-01-28
🐛 Nested tuple access parses as a float.


hayleigh-dot-dev:
I'm in a situation where I have a list of nested tuples. When trying to do a bit of data wrangling, I've written:
```
let #(fields, refs) = list.unzip(list.map(fields, fn(field) { #(#(field.0, field.1.0), field.1.1) }))
```
This ends up throwing the following syntax error:
```
Syntax error

I was not expecting this.

Expected one of: 
A positive integer or a field name.
```
It's incorrectly parsing `field.1.0` as trying to access `1.0` on `field`. This should either work, or report an elm-style error scolding me for having bad code or something.

lpil:
Improving the error sounds good!

lpil:
I think we could possibly have the parse check if the next token is a float and then check to see if that can be used as nested indexing. Will need to also see how to handle further nesting in this case too.

hayleigh-dot-dev:
I would say this behaviour is incorrect, or at the very least _incredibly_ surprising, because the equivalent syntax with named fields works fine and from both a user's perspective and a semantic one, there is little difference between a record and a tuple. 

> I think we could possibly have the parse check if the next token is a float...

This is not sufficient as `foo.1.baz` is also a syntax error.

inoas:
What would be an example syntax where a float would be expected instead of tuple dot syntax?

Should this be possible and execute as is?
```gleam
pub fn foo() {
  let a = #(#(True, False), #(True, False))
  let b = 0.2
  let c = a.0
  .1 == b
}

pub fn foo2() {
  let a = #(#(True, False), #(True, False))
  let b = 0.2
  let c = a.1
  .2 == b
}
```

... because the formatter transforms it to:

```gleam
pub fn foo() {
  let a = #(#(True, False), #(True, False))
  let b = 0.2
  let c = {a.0}.1 == b
}

pub fn foo2() {
  let a = #(#(True, False), #(True, False))
  let b = 0.2
  let c = {a.1}.2 == b
}
```

Would not allowing floats without an integer part (aka `0.1` is a float where `.1` is an syntax error) fix this?

sobolevn:
I am working on this issue, it is nice to get started with a new language with its internals :)

My plan is to add fake block expressions to make `tuple.0.1` equavalent to `{tuple.0}.1` (it also works for structs as well.

Here's how the parsed structure looks like:

```rust
#[test]
fn nested_tuples() {
[...]

PR #2750, opened 2024-03-18
Fix nested tuple dot access lexing, refs #1980

sobolevn:
Design:
- Lexer has now a special case, after `::Name` check that `::Dot` and `::Int` are not following the toke flow
- If they are, parse them as sequence of `::Dot` and `::Int` tokens, not `::Float` token
- Other this automatically start to work correctly 🎉 

Testing:
- I've added a couple of parsing tests for the direct comparision with `{}` and without blocks
- I've also added several runtime tests for this feature, since it is more fundamental :)

Please, share your feedback, since this is my first Gleam contribution and I might miss obvious things (I know Gleam for 1 week now 😆)

Closes https://github.com/gleam-lang/gleam/issues/1980

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,12 @@
   with no head and a spread tail.
 - The compiler will now raise a warning if you're pattern matching on tuple
   literals and suggest you use multiple subjects instead.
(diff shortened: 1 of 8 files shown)


sobolevn:
It was very interesting to work on this project, thanks a lot :)

PR #2754, opened 2024-03-18
Annotate JavaScript runtime in `gleam run --help`

gubsey:
Currently, the javascript runtime selection is not documented when running `gleam run --help`. Change the help message to say
```
Usage: gleam run [OPTIONS] [ARGUMENTS]...

Arguments:
  [ARGUMENTS]...  

Options:
  -t, --target <TARGET>    The platform to target
      --runtime <RUNTIME>  Optionally specify the JavaScript runtime: nodejs, deno, bun
  -m, --module <MODULE>    The module to run
  -h, --help               Print help
```

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -188,6 +188,7 @@ enum Command {
         #[arg(short, long, ignore_case = true)]
         target: Option<Target>,
 
+        /// Optionally specify the JavaScript runtime: nodejs, deno, bun
         #[arg(long, ignore_case = true)]
         runtime: Option<Runtime>,
 
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -118,14 +118,14 @@ enum Command {
         warnings_as_errors: bool,
 
         /// The platform to target
-        #[arg(short, long, ignore_case = true)]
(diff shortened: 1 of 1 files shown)


compiler-cli/src/main.rs
```diff
@@ -188,6 +188,7 @@ enum Command {
         #[arg(short, long, ignore_case = true)]
         target: Option<Target>,
 
+        /// Optionally specify the JavaScript runtime: nodejs, deno, bun
```
lpil, line 191: This list of variants needs to be generated from the type itself rather than being a comment.

lpil (commented):
Thank you

compiler-cli/src/main.rs
gubsey: I've been looking and I don't think that's possible due to it being an `Option<Runtime>` as opposed to just `Runtime`. I'll keep looking though.

gubsey: I solved it by using `#[arg(help = runtime_doc()]` where `runtime_doc` is a new function I made in `compiler-cli/src/main.rs`. This could be changed to a method that's attached to the `Runtime` struct (maybe `Runtime::help_doc()`?)

lpil (approved):
Clever! Thanks

PR #2755, opened 2024-03-18
feat(lsp): complete variables

saiintbrisson:
Adds a naive implementation of variable name completion for the current scope. Please judge! This was put together in about 1 hour of getting to know the codebase, I have no idea if it is correct.

image
image

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -410,7 +410,7 @@ pub type TypedFunction = Function<Arc<Type>, TypedExpr>;
 pub type UntypedFunction = Function<(), UntypedExpr>;
 
 impl<T, E> Function<T, E> {
-    fn full_location(&self) -> SrcSpan {
+    pub fn full_location(&self) -> SrcSpan {
         SrcSpan::new(self.location.start, self.end_position)
     }
 }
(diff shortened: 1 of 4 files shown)


saiintbrisson:
Oh wait, I'll add arguments and consts as well. Forgot about those.

saiintbrisson:
Next step is probably some sort of sorting of completions based on the expected type. Better leave it to another PR

saiintbrisson:
I should probably evaluate whether the cursor if before or after the variable assignment shouldn't I...

saiintbrisson:
r-a has deeper access to information regarding the local variables, like `Environment.scope` has, but I gather the LSP can't access an `Environment`?

lpil:
Aye, the environment is an internal detail of the analysis code, by this point it does not exist. I think we would implement this by building and then inspecting a call graph.

saiintbrisson:
@lpil I think I got it working in almost all cases (or at least those I can think of).

But here's the thing, I went with a recursive approach descending all expressions surrounding the cursor, shouldn't be a problem in most cases, but can be inefficient with complex expression chains and long statements. We can build the call graph and than keep it in memory as long as the source is valid, but I don't really know where to start and what it should look like.

Another thing bugging me is that `engine.rs` will eventually become a mess. Should we relocate completions to something like `engine/completions.rs`?

Lemme know if the current way of doing things is acceptable for now, or if you prefer we hold the idea until we have a clearer idea of how to implement it.

lpil:
Hello! The way to implement this is to build a call graph during compilation and persist that, and then perform a lookup on that graph to get what's currently in scope.

Hold off on splitting any modules for now 👍

lpil:

Issue #1439, opened 2022-01-03
Binary checking doesn't work on macOS arm64


lpil:
As the runner is macOS amd64. We will likely need to wait for GitHub to support arm64

Disabling the check for now.

PR #2758, opened 2024-03-18
Add aarch64-apple-darwin CI target

sobolevn:
GitHub now supports Apple M1 runners: https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

So, I am changing `macos-latest` to be `macos-12`, since it is now just an alias for this version: https://github.com/gleam-lang/gleam/actions/runs/8331253881/job/22797719598#step:1:4

And I am adding this Tier2 Rust target. It should mostly work (until it does not as always).

But, keep in mind that this might slow the CI, for example, our jobs in CPython sometimes take hours to even start. And sometimes it is really fast: https://github.com/python/cpython/pull/116861

Refs https://github.com/gleam-lang/gleam/issues/1439

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,7 @@ jobs:
             use-cross: false
             run-integration-tests: true
           - os: macos-14  # aarch64
+            toolchain: stable
             target: aarch64-apple-darwin
             use-cross: false
             run-integration-tests: true
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,13 +85,13 @@ jobs:
           otp-version: "25.1"
           elixir-version: "1.14.1"
           rebar3-version: "3"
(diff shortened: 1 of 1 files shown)


sobolevn:
Done, it took just 6 minutes, amazing :)

I would like to work on your CI definitions in the next PR, there are a lot of things to improve, like:
- Security
- Resource usage
- Best practices 

If you don't mind :)

lpil (approved):
Thank you!

There's these workflows too which could do with updating
- https://github.com/gleam-lang/gleam/blob/5277f303cba8b9bfe2372c0be71c634584edb76c/.github/workflows/release-nightly.yaml#L53-L77
- https://github.com/gleam-lang/gleam/blob/5277f303cba8b9bfe2372c0be71c634584edb76c/.github/workflows/release.yaml#L15-L39

PR #2760, opened 2024-03-18
Fixed a bug with string pattern matching codegen on the JS target

### LINKED ISSUE #2759: Codegen bug with JS string pattern matching
```gleam
fn main() {
  case Ok(["a", "b c", "d"]) {
    Ok(["a", "b " <> _, "d"]) -> 1
    _ -> 2
  }
}
```
Here the result is `1` on the erlang target (as one would expect) and `2` on the js one.

I tried making it smaller and this example:
```gleam
fn main() {
  case Ok(["b c", "d"]) {
    Ok(["b " <> _, "d"]) -> 1
    _ -> 1
  }
}
```
Crashes on the js target with a `TypeError: Cannot read properties of undefined (reading 'head') `

giacomocavalieri:
This PR closes #2759


lpil (approved):
Thank you!

PR #2761, opened 2024-03-18
Fix formatting line breaks with `\r\n`

### LINKED ISSUE #2753: Formatter should not remove line breaks when converting `\r\n` into `\n`
https://github.com/gleam-lang/gleam/assets/25585136/f7bd4bc0-f059-4a57-b708-0d362d304926

juntuu:
Attempt to fix #2753

I noticed the same happened with trailing space before the empty line. I think this was partially caused by https://github.com/gleam-lang/gleam/blob/2a459ff2c02a59ce32b989a40ce07d88e1d14bcd/compiler-core/src/format.rs#L704, which didn't find the further than expected empty line. But I might be wrong, I don't quite have the full picture where all the spaces and line breaks are handled.

I also added some tests for cases I bumped into with another attempt.

Adding tests for different EOL cases, I noticed that comments would retain the `\r` from `\r\n`. I fixed this by moving the reported position to the `\r` character.

### DIFF
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -4465,6 +4465,127 @@ fn negation_block() {
 }
 
 #[test]
+fn empty_lines_work_with_trailing_space() {
+    let src = "pub fn main() {
+  let inc = fn(a) { a + 1 } 
+
+
+  pair.map_first(#(1, 2), inc)  
+  |> should.equal(#(2, 2)) 
+
+  // Comment
+
+  1  
+
+
+  // Comment 
+
+
+  2
+}
+";
+    let expected = "pub fn main() {
+  let inc = fn(a) { a + 1 }
+
+  pair.map_first(#(1, 2), inc)
+  |> should.equal(#(2, 2))
+
+  // Comment
+
+  1
+
+  // Comment 
+
+  2
+}
+";
+    assert_format!(expected); // Sanity check
+
+    assert_format_rewrite!(src, expected);
+}
+
+#[test]
+fn empty_lines_work_with_eol_normalisation() {
+    let src = "pub fn main() {
+  let inc = fn(a) { a + 1 }
+
+
+  pair.map_first(#(1, 2), inc)
+  |> should.equal(#(2, 2))
+
+  // Comment
+
+  1
+
+
+  // Comment
+
+
+  2
+}
+";
+    let expected = "pub fn main() {
+  let inc = fn(a) { a + 1 }
+
+  pair.map_first(#(1, 2), inc)
+  |> should.equal(#(2, 2))
+
+  // Comment
+
+  1
+
+  // Comment
+
+  2
+}
+";
+    assert_format!(expected); // Sanity check
+
+    assert_format_rewrite!(&src.replace('\n', "\r\n"), expected);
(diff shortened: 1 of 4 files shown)


compiler-core/src/format/tests.rs
```diff
@@ -18,20 +18,36 @@ mod use_;
...
+            $crate::format::pretty(&mut writer, &src, camino::Utf8Path::new("<stdin>")).unwrap();
+            assert_eq!(writer, $output);
+        }
+        // \r
+        {
+            let mut writer = String::new();
+            let src = $src.replace('\n', "\r").into();
+            $crate::format::pretty(&mut writer, &src, camino::Utf8Path::new("<stdin>")).unwrap();
+            assert_eq!(writer, $output);
+        }
```
lpil, line 50: Let's write a couple new tests instead of running the whole test suite 3 times.

compiler-core/src/format/tests.rs
```diff
@@ -4464,6 +4503,86 @@ fn negation_block() {
...
+
+  1
+
+  // bar
+
+  2
+}
+"
+    );
+}
```
lpil, line 4584: Do we not already have these tests?

lpil (commented):
Looks good! I've left a note inline RE the tests.

compiler-core/src/format/tests.rs
juntuu: Fair point. How about a third macro that tests with all the eol variations, and use that for selected test cases? Mainly those which check the empty line formatting.

Would that be good?

And if, would it be better as separate macros e.i. `assert_format_rewrite_with_eol_conventions`, or with extra argument (and same for `assert_format`):
```rs
#[macro_export]
macro_rules! assert_format_rewrite {
    ($src:expr, $output:expr  $(,)?) => { /* main impl */ };

    (check_eol_conventions, $src:expr, $output:expr  $(,)?) => {
        // Different EOLs in input should all format to \n in the output
        $crate::assert_format_rewrite!($src, $output);
        $crate::assert_format_rewrite!(&$src.replace('\n', "\r\n"), $output);
        $crate::assert_format_rewrite!(&$src.replace('\n', "\r"), $output);
    };
}
```

juntuu:
Added changelog entry.

For the tests I went with the extra macro argument for additional EOL checks for now, let me know how that looks to you.

What's the practice around individual commits, should each be passing tests?

Currently this branch first adds breaking test cases and only then the fix. I can combine the test addition & fix commits if desired.

I'd also like to rebase this a bit once things look good to remove some of the back and forth changes. But I'll keep the history intact for sake of review.

compiler-core/src/format/tests.rs
lpil: I'd prefer to keep it simple and write three tests in this case.

juntuu:
Changed tests as requested.

Also rebased and cleaned the history, since there was a conflict to resolve anyway.

compiler-core/src/format/tests.rs
juntuu: I somewhat disagree about this being a one-off. I might even consider this being a general property of formatting: eol normalisation should not change the end result.

The trailing white space case is not quite as general, since comments (at least currently) retain the space.

lpil (approved):
Thank you!!

Issue #2757, opened 2024-03-18
LSP: add new import via autocompletion (after #2756)


MarcThu:
It would be a better DX, if new modules could be imported while writing other code.

This feature could be implemented like in Go:
When I am in a function body and type `fmt` I get autocompletion for `fmt` even if this is not imported yet. When I use the autocompletion (press enter), the selected function from `fmt` gets autocompleted as well as imported.

(should be implemented after #2756)

lpil:
Good idea! Thank you

PR #2763, opened 2024-03-19
Adds import completion when writing imports

### LINKED ISSUE #2756: LSP: autocomplete when importing modules (after import keyword)
The LSP should assist the developer when adding imports to the current file. There should be autocompletion for the `import` keyword as well as the possible packages after the keyword.

The LSP should suggest both installed packages and custom packages from the current project.

Acepie:
Closes #2756

Adds module import completion. Includes the modules from dependencies and root package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -82,6 +82,7 @@
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
 - Go to definition now works for values in dependency Gleam modules.
+- Completions are now provided for module imports
 
 
 ## v1.0.0 - 2024-03-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@
 - Go to definition now works for values in dependency Gleam modules.
 - Completions are now provided for module imports.
 
+
 ## v1.0.0 - 2024-03-04
 
 ### Language changes
@@ -110,6 +111,7 @@
 - The format used by the formatter has been improved in some niche cases.
 - Improved the formatting of long case guards.
 - The formatter can now format groups of imports alphabetically.
+
 
 ## v1.0.0-rc1 - 2024-02-10
 
@@ -158,6 +160,7 @@
 - Fixed a bug where external only functions would "successfully" compile for a
   target they do not support, leading to a runtime error.
 
+
 ## v0.34.1 - 2023-01-17
 
 ### Build tool changes
@@ -170,6 +173,7 @@
   format incorrectly.
 - The `@deprecated` attribute can now be used to annotate module constants.
   This will cause a warning to be emitted when the constant is used.
+
 
 ## v0.34.0 - 2023-01-16
 
@@ -188,6 +192,7 @@
 - Fixed a bug where function heads would go over the line limit in the
   formatter.
 
+
 ## v0.34.0-rc2 - 2023-01-11
 
 ### Bug fixes
@@ -197,6 +202,7 @@
 - Fixed a bug where the compiler would in some cases fail to error when an
   application uses functions that do not support the current compilation
   target.
+
 
 ## v0.34.0-rc1 - 2024-01-07
 
@@ -227,7 +233,7 @@
 
 - The `gleam new` command now accepts any existing path, as long as there are
   no conflicts with already existing files. Examples: `gleam new .`, `gleam new
-..`, `gleam new ~/projects/test`.
+  ..`, `gleam new ~/projects/test`.
 - The format for the README created by `gleam new` has been altered.
 - The `gleam.toml` created by `gleam new` now has a link to the full reference
   for its available options.
@@ -261,6 +267,7 @@
   "Type 'Result' is not generic".
 - Not providing a definition after some attributes is now a parse error.
 
+
 ## v0.33.0 - 2023-12-18
 
 ## v0.33.0-rc4 - 2023-12-17
@@ -270,6 +277,7 @@
 - The deprecated `BitString` type has been removed.
 - The deprecated `inspect` functions and `BitString` type has been removed from
   the JavaScript prelude.
+
 
 ## v0.33.0-rc3 - 2023-12-17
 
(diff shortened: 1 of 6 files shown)


compiler-core/src/build.rs
```diff
@@ -234,14 +234,17 @@ impl Module {
...
+    pub fn get_module_comments(&self) -> Vec<EcoString> {
+        self.extra
             .module_comments
             .iter()
             .map(|span| Comment::from((span, self.code.as_str())).content.into())
-            .collect();
+            .collect()
+    }
+
+    pub fn attach_doc_and_module_comments(&mut self) {
```
Acepie, line 245: not sure if it would make sense to just call this on each module when the lsp compiles or not

CHANGELOG.md
```diff
@@ -68,6 +68,7 @@
 - Fixed a bug where hovering over an expression in the middle of a pipe would
   give the wrong node.
 - Go to definition now works for values in dependency Gleam modules.
+- Completions are now provided for module imports
```
lpil, line 71:
```suggestion
- Completions are now provided for module imports.
```

compiler-core/src/language_server/engine.rs
```diff
@@ -427,6 +429,35 @@ where
...
+
+    fn completion_imports<'b>(&'b self, module: &'b Module) -> Vec<lsp::CompletionItem> {
+        let mut completions = vec![];
+
+        // Module functions
+        for (module_name, dep_module) in &self.compiler.modules {
+            if module_name == &module.name {
+                continue;
+            };
+            let doc_string: String = dep_module.get_module_comments().join("\n");
```
lpil, line 441: These will commonly be thousands of words, it's not suitable for displaying in the LSP. Let's not include this.

compiler-core/src/language_server/engine.rs
```diff
@@ -211,10 +211,12 @@ where
                     Some(this.completion_types(module))
                 }
 
-                Located::ModuleStatement(Definition::Import(_) | Definition::ModuleConstant(_)) => {
-                    None
+                Located::ModuleStatement(Definition::Import(_)) => {
```
lpil, line 214: Unfortunately this isn't going to work most the time as you're not going to be inside an import statement when trying to complete an import, as the AST isn't updated until there has been a successful compilation. It might be if we had fault tolerant parsing though. I'm not sure how's best to approach this- do you have ideas?

compiler-core/src/language_server/compiler.rs
```diff
@@ -106,12 +105,17 @@ where
...
+
         // Store the compiled dependency module information
-        for module in &compiled_dependencies {
+        for module in compiled_dependencies {
+            compiled_modules.push(module.input_path.clone());
             let path = module.input_path.as_os_str().to_string_lossy().to_string();
             let line_numbers = LineNumbers::new(&module.code);
             let source = ModuleSourceInformation { path, line_numbers };
             _ = self.sources.insert(module.name.clone(), source);
+            _ = self.modules.insert(module.name.clone(), module);
```
lpil, line 118: Could you explain the changes in this file please. Thank you.

lpil:
Could we have some tests to check that `src` can't import `test` too please.

Here's the old implementation which was removed: <https://github.com/gleam-lang/gleam/commit/00a41a57>

compiler-core/src/language_server/compiler.rs
Acepie: The goal of the change was to add compiled dependency modules into the list of modules so that the lsp would have all potential modules that could be useful when doing operations. Also since we needed to loop through `compiled_dependencies` twice anyway (once to construct the list of paths and once to add the module info to our maps) I combined the iterations into 1 loop

Acepie: oh yeah good point I completely missed that somehow. Lemme switch

compiler-core/src/language_server/engine.rs
Acepie: Yeah that's a fair point. My original plan was to at least get the plumbing working to have import completions in this issue and figure out how to generally apply it outside of import statements as part of #2757 

I guess for most cases outside of the import statement itself we'd probably actually be completing for function/consts in other modules and just adding the import as an additional text insert so it'd be a bit different.

Fault tolerant parsing would definitely make this more usable as is but I would need to read more about how that actually gets implemented in practice

Acepie:
Added some more tests and an extra bit to exclude already imported modules. Not sure if you want to get this in now or if you'd prefer to solve the general issue for https://github.com/gleam-lang/gleam/issues/2766 first

lpil (commented):
I've given this a test and on vscode it gives incorrect results such as turning

```gleam
import gleam/
```
into 
```gleam
import gleam/gleam/string
```

which is more annoying than no completion.

We're going to need to provide completions with precise positioning to make sure that the existing code is updated correctly. I think the approach here may be to not use the AST, and to instead write a second simple parse that only looks for imports and can figure out if the person is writing an import and what text needs to be edited and how.

Acepie:
Hmmm ok that's annoying. I'll need to spend more time on this/test the full lsp running locally through vs code. It "feels" like there should be an easy way to swap the existing import with the new one if they accept the suggestion without having to do a separate pass. I'll try to come back to this over the week end

lpil:
The existing system doesn't have the information on the exact current state of the code so it has to be based off of the source file, or everything has to be changed to be fault tolerant.

Acepie:
Ok I think I get what you mean. Yeah I guess we'd have to run the new/changed source and attempt an incomplete parse. For now it could start with just figuring out if we are stopping the parse at an import but eventually we would use this as a base for general fault tolerant parsing. Does that sound right?

lpil:
I would just scan across the string once until we hit the desired line and then see if it starts with `import`.

Acepie:
Ahh fair enough. Probably getting ahead of myself. Lemme try that and get this going

Acepie:
https://github.com/gleam-lang/gleam/assets/5996838/309576dd-3b4c-41e9-a581-017f6336af57

Ok tried out what you suggested and tested it out a bit on local. Seems to work nicely and take into account the current src well. The only kinda "weird" edge case is that since we filter already imported modules if you are mid editting one of your existing imports it won't recommend that one (which seems like reasonable behavior to me but might be confusing)

lpil (approved):
Wonderful!! Really exciting to have this working now! Thank you

PR #2767, opened 2024-03-19
Improve ci.yml and add dependabot.yml

sobolevn:
Here's the full list of changes:
- `workflow_dispatch:` allows you to manually run actions from the admin panel: https://github.com/gleam-lang/gleam/actions/workflows/ci.yaml It is really useful sometimes, in other cases it just does nothing
- `permissions:` sets stricter security rules for this workflow. It should not be able to do anything. Upload / download of artifacts will continue to work: https://github.com/actions/upload-artifact/issues/197 `permissions:` can be set on individual jobs as well, if needed later
- `concurrency:` cancels previous jobs, when new commits are pushed to the CI. It helps to save extra resources and not to continue meaningless work
- I've changed `workflow.os` to `runner.os`, because the first is not really about the runner os, it is just the label we use
- `node-version:` is updated, because `16` is reaching its EOL, see https://nodejs.org/en/about/previous-releases `18` is still good 👍  
- I've also updated several actions' versions. No other changes here should be required
- I've also added `dependabot.yml` which is a nice tool to update deps, embeded into GitHub itself. It should not be very noisy and can help doing what I'm doing now :)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
@@ -12,6 +13,13 @@ env:
   CARGO_PROFILE_DEV_DEBUG: 0
   CARGO_PROFILE_TEST_DEBUG: 0
   CROSS_CONTAINER_UID: 0
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
 jobs:
   test:
@@ -61,7 +69,7 @@ jobs:
             run-integration-tests: true
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Install musl-tools incl. musl-gcc
         uses: awalsh128/cache-apt-pkgs-action@v1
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -13,6 +14,13 @@ env:
   CARGO_PROFILE_TEST_DEBUG: 0
   CROSS_CONTAINER_UID: 0
 
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
```
lpil, line 21: Is the head_ref the commit sha? If so, won't this be different every time? I'd have thought that maybe the PR number would be what we want

lpil (commented):
Great! Thank you! I've left 1 question

.github/workflows/ci.yaml
lpil: Oh fab!!

lpil (approved):
Thank again!!

sobolevn:
Hm, this is interesting. Looks like you use "Rebase and merge" strategy, which I don't use.
And since there are multiple commits at the same time to `main`, they cancel each other and this notification shows up.

Give me a moment please, I will search what the options are.
Revert this part always is an option.

lpil:
I think it may have been as I merged several PRs in quick succession. Maybe we can keep the config for a little while and see how often this happens.

PR #2781, opened 2024-03-19
[WIP] Update LSP hover action to return labelled definition

manuraj17:
Relates to https://github.com/gleam-lang/gleam/issues/2653 

Updating the LSP hover action to return the full function head, as per the documentation. 

The planned format is as below
```
<fn-accessor-type> fn <function-name>(<label> <argument-name>: <type>, ..) -> <return-type> 
```
Eg:
For a function like below
```
fn greet(firstname name:String) { io.println("Hello " <> name) }
```
The generated string would be
```
fn greet(firstname name: String) -> Nil 
```

Current implementation uses pretty print and existing Fn Type enum. Since labels are not part of the type system, we are going ahead with simple string interpolation for now. 

P.S : This is the first time working with the codebase and such a big Rust repo, hence please feel free to suggest changes and improvements. 

P.P.S: I am trying to test it out in local hence kept the `[WIP]` tag.

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -520,14 +520,44 @@ fn hover_for_function_head(
 ) -> Hover {
     let empty_str = EcoString::from("");
     let documentation = fun.documentation.as_ref().unwrap_or(&empty_str);
-    let function_type = Type::Fn {
-        args: fun.arguments.iter().map(|arg| arg.type_.clone()).collect(),
-        retrn: fun.return_type.clone(),
-    };
-    let formatted_type = Printer::new().pretty_print(&function_type, 0);
+
+    let mut function_string = format!("fn {}(", fun.name);
+    if fun.public {
+        function_string = format!("pub {}", function_string);
+    }
+
+    let default_label = ecow::EcoString::from("");
+
+    let args: Vec<_> = fun
+        .arguments
+        .iter()
+        .map(|arg| {
+            let label = match arg.names.get_label() {
+                Some(val) => val,
+                None => &default_label,
+            };
+
+            format!(
+                "{} {}:{}",
+                label,
+                arg.names.get_variable_name().unwrap(),
+                Printer::new().pretty_print(&arg.type_.clone(), 0)
+            )
+            .trim()
+            .to_string()
+        })
+        .collect();
+
+    function_string = format!(
+        "{}{}) -> {}",
+        function_string,
+        args.join(", "),
+        Printer::new().pretty_print(&fun.return_type.clone(), 0),
+    );
+
     let contents = format!(
         "```gleam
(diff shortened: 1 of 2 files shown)


compiler-core/src/language_server/engine.rs
```diff
@@ -520,14 +520,44 @@ fn hover_for_function_head(
 ) -> Hover {
     let empty_str = EcoString::from("");
     let documentation = fun.documentation.as_ref().unwrap_or(&empty_str);
-    let function_type = Type::Fn {
-        args: fun.arguments.iter().map(|arg| arg.type_.clone()).collect(),
-        retrn: fun.return_type.clone(),
-    };
-    let formatted_type = Printer::new().pretty_print(&function_type, 0);
+
+    let mut function_string = format!("fn {}(", fun.name);
```
Acepie, line 524: FWIW https://github.com/gleam-lang/gleam/pull/2705/files it looks like there is an existing draft pr that utilizes the function used by the doc generation to do roughly the same thing https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/format.rs#L1403 not sure if theres a preference to reuse vs write new here but worth considering

manuraj17: @Acepie Oh nice, I think it's targeting same set of changes. I wasn't aware of this `format`. Looks like that way is better. I will discuss on that PR 👍🏽 Thanks.

manuraj17:
Considering the sampe like below
```gleam
import gleam/io
import gleam/string

fn greet(n name: String) -> String {
  string.append(name, "hello")
}

pub fn main() {
  io.println(greet("Tom"))
}
```


Some of my findings (being new to the language)
- Hovering on `main()` our Lsp identifies the keyword as a function and it works fine
- Hovering on `greet` on this line ` io.println(greet("Tom"))` , it get idenfitied as `expression` [here](https://github.com/gleam-lang/gleam/blob/5ade71380454d298d542012b7be82bec461e4257/compiler-core/src/language_server/engine.rs#L279)

Will need to identify how to handle that case there. Plus, I am also trying to understand how to quickly identify if the `expression` in this method [here](https://github.com/gleam-lang/gleam/blob/5ade71380454d298d542012b7be82bec461e4257/compiler-core/src/language_server/engine.rs#L563)


Tried defining something like below but seems to not to work
```rust
    /// Returns `true` if the typed expr is [`Fn`].
    ///
    /// [`Fn`]: TypedExpr::Fn
    #[must_use]
    pub fn is_function(&self) -> bool {
        matches!(self, Self::Fn { .. })
    }
```

cc @lpil

Acepie:
The case you are probably looking for there is a `Self::Call` not a `Self::Fn` fn is a function definition whereas call is when the function is actually being used

lpil:

Issue #2733, opened 2024-03-16
Pattern aliases for strings produce bad Erlang


rchog:
I was delighted to learn that the pattern aliases described [here](https://tour.gleam.run/flow-control/pattern-aliases/) are also supported for strings, however, I ran into this problem when trying to run my code with Erlang as the target. I pared it way 
down to this minimal example - it's nonsense, but it seems clear where the problem is: 
```ruby
import gleam/io
import gleam/string

fn reproduce_bug(data) {
  case data {
    "" -> ""
    "(" as paren <> rest -> paren <> reproduce_bug(rest)
    s -> s <> reproduce_bug(string.drop_left(s, 1))
  }
}

pub fn main() {
  io.println(reproduce_bug("abcde(fgh)ijkl"))
}
```

..which produces the following when run with `gleam run`: 
```
[ bug ]$ gleam run
Downloading packages
 Downloaded 2 packages in 0.00s
  Compiling gleam_stdlib
  Compiling gleeunit
  Compiling bug
/home/r/projects/gleam/bug/build/dev/erlang/bug/_gleam_artefacts/bug.erl:12:20: syntax error before: '='
%   12|         <<"("/utf8 = Paren, Rest/binary>> ->
%     |                    ^

/home/r/projects/gleam/bug/build/dev/erlang/bug/_gleam_artefacts/bug.erl:6:2: spec for undefined function reproduce_bug/1
%    6| -spec reproduce_bug(binary()) -> binary().
%     |  ^

/home/r/projects/gleam/bug/build/dev/erlang/bug/_gleam_artefacts/bug.erl:21:22: function reproduce_bug/1 undefined
%   21|     gleam@io:println(reproduce_bug(<<"abcde(fgh)ijkl"/utf8>>)).
%     |                      ^

error: Shell command failure

There was a problem when running the shell command `escript`.
```

It runs as expected with `gleam run -t js`. My Gleam version is 1.0.0, installed from [here](https://aur.archlinux.org/packages/gleam). I'll add the generated Erlang for good measure:
```erlang
-module(bug).
-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).

-export([main/0]).

-spec reproduce_bug(binary()) -> binary().
reproduce_bug(Data) ->
    case Data of
        <<""/utf8>> ->
            <<""/utf8>>;

        <<"("/utf8 = Paren, Rest/binary>> ->
[...]

PR #2782, opened 2024-03-19
fix pattern aliases for strings when targeting erlang

wgancar:
Hello, this is my attempt to fix the issue described here

[#2733](https://github.com/gleam-lang/gleam/issues/2733)

Currently making aliases for string patterns doesn't work when complied to erlang. The problem is that erlang doesn't allow aliases for constant binary patterns.

My workaround is to match the number of bytes first and then write additional guard to check if matched bytes matches the provided pattern

so that this 
```gleam
case x {
  "Hello, " as greeting <> _  -> greeting
}

```
becomes this
```erlang
<<Greeting:7/binary, Name/binary>> when Greeting =:= <<"Hello, ">> ->
            Greeting;
```

instead of invalid code that we have today
```erlang
<<"Hello, "/utf8 = Greeting, Name/binary>> ->
            Greeting;
```

If this is acceptable approach I still have one issue to fix, If there are some user defined guards on the clause level I have to replace `when` with `andalso` to make sure that both guards are supported

Thanks


lpil:
The approach looks great to me! We do something similar to this on JavaScript. Let me know when you would like a code review, thank you.

wgancar:
Ok, I think im ready for review. Ive made sure that user defined guards are working properly and I've added test to verify that

compiler-core/src/pretty.rs
```diff
@@ -656,4 +656,18 @@ impl<'a> Document<'a> {
...
+            EcoString(s) => s.contains(pattern),
+            String(s) => s.contains(pattern),
+            Str(s) => s.contains(pattern),
+            // assuming `broken` and `unbroken` are equivalent
+            Break { broken, .. } => broken.contains(pattern),
+            ForceBroken(d) | Nest(_, _, _, d) | Group(d) | NextBreakFits(d, _) => d.contains_string(pattern),
+            Vec(docs) => docs.iter().any(|d| d.contains_string(pattern)),
+            _ => false
+        }
+    }
```
lpil, line 672: Remove this please, it's not appropriate to introspect the printing alegbra, and this isn't a correct implementation as it could be across multiple variants.

lpil (commented):
Thank you but we cannot introspect the algebra to learn things about the code. The code is source of information here.

Could you also add comments like the ones you removed please 🙏

compiler-core/src/pretty.rs
wgancar: You are absolutely right, sorry

wgancar:
> Thank you but we cannot introspect the algebra to learn things about the code. The code is source of information here.
> 
> Could you also add comments like the ones you removed please 🙏

Thank you for the review I appreciate it a lot 💜 Im still not sure where this guard check should happen. Right now I made simple function in patterns module but Im not sure if its feels right. Any suggestions?

lpil (approved):
Beautiful work! Thank you very much! 💜

PR #2790, opened 2024-03-21
Fix a bug where tuple/call/list items would be needlessly indented

### LINKED ISSUE #2607: Formatter shouldn't indent an inner item after a comment

giacomocavalieri:
This PR closes #2607 and closes #2611


lpil (approved):
Thank you. Love the comments!

PR #2791, opened 2024-03-21
Improve error message multi-variant custom type.

### LINKED ISSUE #2778: Improve error message on UnknownRecordField for custom types with multiple variants
Hello,

I propose enhancing the error message for `UnknownRecordField` scenarios to better communicate issues related to field access in custom types with multiple variants. The motivation for this change is to provide clearer guidance to developers, especially in situations where a field's presence or position varies across different variants.

### Current Situation:
When attempting to access a field in a custom type that has multiple variants, developers may encounter errors if the field is not consistently positioned or present in all variants. This is currently captured in our tests with comments like:

["We cannot access fields on custom types with multiple variants where they are in different positions e.g. 2nd and 3rd"](https://github.com/gleam-lang/gleam/blob/752fe2a30cb30feb4cba461e9233f53a9caa7fd8/compiler-core/src/type_/tests/errors.rs#L618-L644)
However, the existing error message does not explicitly clarify this aspect, potentially leading to confusion.

### Proposed Change:
To improve developer understanding and troubleshooting, I suggest updating the error message for UnknownRecordField instances to the following:

"The field you are trying to access might not be consistently present or positioned across the custom type's variants, preventing reliable access. Ensure the field exists in the same position and has the same type in all variants to enable direct accessor syntax."

This message directly addresses the potential inconsistency in field presence and position across variants and advises on how to rectify the situation.

### Justification:
This change will make error messages more informative and actionable, reducing the time developers spend diagnosing and fixing such issues. By explicitly stating the requirement for field consistency in the error message, we align with best practices for error handling and user experience, aiding in faster resolution and clearer understanding of the underlying type system constraints.

I believe this enhancement will significantly improve the developer experience and would appreciate your feedback and support in implementing this change.

### Changes:
I've drafted some changes. Let me know if you think a pull request is valid.
https://github.com/tubedude/gleam/tree/multi-variable-custom-type3

Thank you,

tubedude:
Closes #2778


compiler-core/src/error.rs
```diff
@@ -1224,6 +1226,13 @@ Hint: Add some type annotations and try again."
                         text.push_str(field);
                     }
 
+                    match variants {
+                        RecordVariants::HasVariants => {
+                            text.push_str("\n\nNote: The field you are trying to access might not be consistently present or positioned across the custom type's variants, preventing reliable access. Ensure the field exists in the same position and has the same type in all variants to enable direct accessor syntax.");
```
lpil, line 1231: Wrap this long string please! There's a `wrap` function which is useful for this. You can see it used below.

lpil (commented):
Nice! Thank you! I've left a note inline.

compiler-core/src/error.rs
tubedude: Done. I updated the branch from `gleam-lang:main` wrapped the msg and also separated the `\n\n` from the message itself.
Thanks for the tip.

lpil (approved):
Thank you!

Issue #1599, opened 2022-04-16
Switch from capnproto to serde


HarryET:
Switch from capnproto to serde with https://github.com/bincode-org/bincode, this is because capnproto is difficult to work with while serde is not. Capnproto is also very time-consuming to work with and is error prone.

lpil:
Fab, thank you!

lpil:
@HarryET are you still looking to do this?

lpil:
Cool I'll untag you for now :thum

PR #1912, opened 2022-12-27
feat: Switch from Cap'n'proto to Bincode

HarryET:
Closes #1599

PR #2800, opened 2024-03-21
WIP: replace `capnpronto` with `serde` + `bincode`

sobolevn:
Why is this PR marked as WIP?
Well, there's a problem I want to ask about.

Right now we can correctly serialize and deserialize all internal objects, except the fact that `TypeVar`'s ids are not regenerated. This is needed to uniquely identify type vars and to make sure that there are no conflicts.

Old module had this code: https://github.com/gleam-lang/gleam/blob/efc519c65fcf23dd1c0d9e890efe0a5f9260c021/compiler-core/src/metadata/module_decoder.rs#L140-L155

The first proposed solution by @HarryET https://github.com/gleam-lang/gleam/pull/1912 had this logic: https://github.com/HarryET/gleam/blob/b46f227ac18f9bbe2e36de2a7c9b73505ba74282/compiler-core/src/metadata.rs#L22-L42

But, the code above is incomplete (it does not have `types_value_constructors`, `values` and `accessors` processing). It is also not very clean and will require constant modification.

The problem is that `serde` does not allow simple stateful deserialization. It has `DeserializeSeed` API which is not an easy API to use and to connect with `bincode`.

The second problem is that `bincode@1.0` will soon be replaced with `bincode@2.0` which will not have `deserialize_from_seed` API. So, I don't think that it is worth doing that as well.

I've also tried https://github.com/Marwes/serde_state to work with the global state, but it is unmaintained.

So, what should we do? :)
Tests are failing because of that.

Most of the credit for this change goes to @HarryET 
Thank you!

Closes https://github.com/gleam-lang/gleam/issues/1599


HarryET:
Awesome to see this change back! Happy to have a look if you want but feel bad you've had to sift through that PR 🤣

sobolevn:
I totally need an advice :)

lpil:

PR #2802, opened 2024-03-21
Fix `test_wasm` to actually run tests

sobolevn:
Docs about `#[wasm_bindgen_test]`: https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/usage.html
Docs about `wasm-pack test`: https://rustwasm.github.io/wasm-pack/book/commands/test.html

I've also simplified the `wasm-pack` installation, we now don't have to use an intermediate file. We can just pipe the file into shell.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -251,13 +251,10 @@ jobs:
 
       - name: Install wasm-pack
         run: |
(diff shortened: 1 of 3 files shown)


sobolevn:
It now executes 6 tests: https://github.com/gleam-lang/gleam/actions/runs/8381435945/job/22952950160?pr=2802#step:6:191 🎉

lpil (approved):
Thank you! Great work!

PR #2806, opened 2024-03-22
Fix error message when trying to use single quotes to define a string

sobolevn:
Suggestions to improve the error message are welcome!


lpil (approved):
Thank you!

PR #2807, opened 2024-03-22
Add vscode recommended extensions

sobolevn:
This PR adds `.vscode/extensions.json` which contains a list of recommended extensions to install to get an easier time working on this project. As a new developer, it would be nice to have.

It includes: rust and gleam tools + a debugging backed (which is platform dependent and is different on Windows). But, I guess no one really uses Windows to develop Rust + Erlang tools, so it is fine? ¯\_(ツ)_/¯

Here's how it looks like inside the IDE: it is just a small section which suggests extensions, nothing more: 
Снимок экрана 2024-03-22 в 13 04 56

Docs: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

Since there's no maintaince cost involved (or close to it), I guess that this can be considered a net-positive for the project: improving DX is a good thing.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,8 +3,6 @@
   "recommendations": [
     // Rust:
     "rust-lang.rust-analyzer",
-    // Debugging support:
-    "vadimcn.vscode-lldb",
 
     // Gleam:
     "Gleam.gleam",
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,12 @@
+{
+  // @see https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions
+  "recommendations": [
+    // Rust:
+    "rust-lang.rust-analyzer",
+    // Debugging support:
+    "vadimcn.vscode-lldb",
+
+    // Gleam:
+    "Gleam.gleam",
+  ]
+}


lpil:
Hello! Let's drop the debug one as I can't speak with enough authority to recommend that.

Given code recommends the other two by default what is the goal here? I'd prefer to leave it up to them so we don't have to ever update this ourselves

sobolevn:
> Given code recommends the other two by default what is the goal here?

Does it? Here' how recommendations look inside the same project without `extensions.json`:
Снимок экрана 2024-03-22 в 18 34 13

Anyway, this is so minor that I don't want to waste your time on that :)
Feel free to close! 👍

inoas:
Would a similar thing make sense for skeletons generated by gleam new and/or in stdlib?

lpil:
It recommends for me when I open a Rust or a Gleam file

sobolevn:
> It recommends for me when I open a Rust or a Gleam file

Not for me :(
Снимок экрана 2024-03-22 в 19 24 27
Снимок экрана 2024-03-22 в 19 24 19

lpil:
I think you must have turned off the recommendations. It's there by default but there's a button with the popup to disable them.

I'm gunna close this, thank you!

PR #2810, opened 2024-03-22
Setup dependabot for Rust

### LINKED ISSUE #2803: Consider setting up Dependabot for Cargo
You already have a `dependabot.yml` file: https://github.com/gleam-lang/gleam/blob/efc519c65fcf23dd1c0d9e890efe0a5f9260c021/.github/dependabot.yml#L4

And there are many outdate dependencies in `Cargo.toml`s, so maybe it is a good idea to bump them automatically?

It would be something like:

```
  - package-ecosystem: "cargo"
    directory: "/"
    schedule:
      interval: "weekly"
```

+ several similar entries for submodules.
Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem (grep for `cargo`)

This way it is easier to spot changes that affect the project.
On the other hand, it increases the noise from the project.

If you want to try that, I can send a PR.

sobolevn:
Adds `dependabot` support for Rust.
Docs: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
See `cargo` section.

Possible risks:
- Noise, solution: increase the `interval` value, add `open-pull-requests-limit`
- Dependabot can have bugs specific to Rust and Cargo. Solution: patience :)

Closes #2803

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,6 @@ compiler-core/generated/schema_capnp.rs generated
 
 # Tests:
 test-package-compiler/src/generated_tests.rs generated
+
+# Lock files:
+Cargo.lock generated
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,27 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
+
+# Rust:
+- package-ecosystem: "cargo"
+  directory: "/"
+  schedule:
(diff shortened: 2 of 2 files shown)


sobolevn:
Here it goes: https://github.com/gleam-lang/gleam/pull/2815

PR #2829, opened 2024-03-25
Fix matching of string prefix with `\u` escape on JS target

### LINKED ISSUE #2828: String prefix matching doesn't handle unicode escapes on JS codegen
The program below prints `Result: fghi` instead of `Result: abcde fghi` when run on the JS target:

```js
pub fn main() {
  let z = "\u{0032} abcde fghi"
  let assert "\u{0032} " <> p = z

  io.println("Result: " <> p)
}
```

Peeking at the generated code, we can see that `p` is being assigned a substring longer than it should be:

```js
export function main() {
  let z = "\u{0032} abcde fghi";
  if (!z.startsWith("\u{0032} ")) {
    throw makeError(
      "assignment_no_match",
      "aog",
      33,
      "main",
      "Assignment pattern did not match",
      { value: z }
    )
  }
  let p = z.slice(8);  // <---- here
  return $io.println("Result: " + p);
}
```

This is because the code at `javascript/pattern.rs` to calculate the length of the prefix string, taking into account escape sequences, is assuming (as seen below) that every escape sequence in a Gleam string has (and expands into) a single character, but Unicode escape sequences always contain more than one character after the backslash (`\u{0032}` vs `\n`), so the `{`, `0`, `0`, `3`, `2` and `}` characters are not being filtered out.

https://github.com/gleam-lang/gleam/blob/5ade71380454d298d542012b7be82bec461e4257/compiler-core/src/javascript/pattern.rs#L864-L871

One possible solution is to just forgo of that logic in `pattern.rs` and generate `z.slice("\u{0032} ".length)` instead.
Alternatively, we can use a nested loop to skip upcoming `{`, (digit), `}` characters.

PgBiel:
Fixes #2828

When matching a string prefix, to get the rest of the string, we use `string.slice(UTF-16 length of prefix)` at the generated JS. https://github.com/gleam-lang/gleam/pull/2038 had fixed the case where the string has simple escape sequences generating only a single ASCII character each, such as `\f`, `\n` or `\t` - those should result in a skip of 1 character, not 2. However, it did not take into account `\u{codepoint}`, which not only has multiple characters in its escape sequence, but also may produce a UTF-16 length larger than 1, thus potentially requiring a skip of 2 characters at once in the generated `.slice` call, depending on the codepoint.

In my code, we properly convert `\u{codepoint}` into the character represented by that codepoint and push it to the filtered string whose UTF-16 length will be taken later. Therefore, we will always get the correct length of the prefix string, after `\u{...}` is applied.

I believe that should be the last exception we will need to make when considering the string's effective length (unless there are other character escape sequences I'm not aware of, which doesn't seem to be case, taking a quick look at the lexer code).


lpil (approved):
Bravo! This is fantastic work. Thank you very much.

lpil:
Could you update the changelog please 🙏

PgBiel:
> Bravo! This is fantastic work. Thank you very much.

Thank you!! And you're welcome! :smiley:

> Could you update the changelog please 🙏

Done!

lpil (approved):
Thank you!

lpil:
Could you resolve the merge conflicts please? 🙏

PgBiel:
Done! :rocket:

lpil (approved):
Thank you!

PR #2830, opened 2024-03-25
Rename `$prepend` to `listPrepend` on the JS target

### LINKED ISSUE #2823: Module named `prepend` shadows prelude's `prepend` on JS codegen
On latest main, if you import a name from a module named `prepend`, the JS codegen will import the module under a variable named `$prepend`. Similarly, if you write something like `[1, ..tail]`, JS codegen will convert that to `$prepend(1, tail)` and import `$prepend` from the prelude. Doing both in the same file leads to a name conflict, flagged below when running.

error: $prepend was already defined

Sample project attached below.

[showcase.tar.gz](https://github.com/gleam-lang/gleam/files/14733062/showcase.tar.gz)

One possible solution is to import `prepend` from the prelude aliased to `listPrepend` instead, or use some other naming. I am willing to make a PR for this change (to whichever name is agreed upon).

PgBiel:
Fixes #2823

Renames the `$prepend` alias for prelude's `prepend` to `listPrepend` instead. That way, it won't conflict with a Gleam module named `prepend`, and Gleam modules can't have uppercase characters, so the new name won't conflict with any Gleam modules (which are also imported with a dollar sign in front), variables (which also can't have uppercase characters) or types (which are in PascalCase, not camelCase).

Feel free to suggest an alternative solution, e.g. some name other than `listPrepend`. (Maybe `prependWith` could be better? Not sure.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@
 - Allow compilation of packages that require `"rebar"` using the rebar3 compiler.
 - A warning is now emitted when defining an opaque external type.
 - Improve error message when using incorrect quotes (`'`) to define a string
+- Fixed a bug where an imported module named `prepend` would conflict with the
+  `prepend` function imported from the prelude in the JavaScript target.
 
 ### Formatter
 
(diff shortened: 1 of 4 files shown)


PgBiel:
There you go! :+1:

lpil (approved):
Thank you

inoas:
great find!

PR #2842, opened 2024-03-26
fix issue with windows local manifest paths being absolute

### LINKED ISSUE #2837: manifest.toml uses absolute paths for relative path devs on Windows
An example *manifest.toml* with a relative path dep `shared = { path = "../shared" }`:

```toml
packages = [
# ...
{ name = "shared", version = "1.0.0", build_tools = [
    "gleam",
  ], requirements = [
    "gleam_stdlib",
  ], source = "local", path = "//?/D:/git/my_app/shared" }, # <=====
# ...
]

[requirements]
shared = { path = "../shared" }
mist = { version = "~> 0.17" }
wisp = { version = "~> 0.13" }
# ...
```

It doesn't hurt when developing locally, but won't work when e.g. deploying with Docker.

_Originally posted by @xhh in https://github.com/gleam-lang/gleam/issues/2182#issuecomment-2017918389_

Acepie:
Closes #2837 

The issue was that on Windows canonicalise adds a special root prefix https://stackoverflow.com/questions/41233684/why-does-my-canonicalized-path-get-prefixed-with which caused the diff to not properly work when creating the relative path. Since keeping the target canonical might be important for resolutions in some cases this pr updates the source to be canonicalized when doing the relative path.

I tested this against https://github.com/Acepie/p5js_gleam locally on my windows computer and it seems to work as expected now.


lpil (commented):
Could you add a test for this please? Thank you 🙏

Acepie:
So the difficulty for adding a test for this case is that canonicalize actually checks the file exists in the filesystem, which is valid/always true in real-world usage but makes it pretty hard to make a nonbrittle test that hits this case

Acepie:
Or at the very least, i don't know how we'd write a test that passes consistently on both local and CI that actually tests the intended behavior (i.e without writing some weird mock that just string appends the root without an FS check)

Acepie:
Ok based on how difficult it was to test the canonicalising for source and from what I read in https://github.com/rust-lang/rust/issues/42869 it seems like we might be better off just stripping from target. Lmk if that is fine with you. I could imagine theoretical ways this doesn't work (I'm thinking some path where the prefix is actually required to be resolvable) but considering the current state this is probably the better approach

lpil (approved):
Oh very nice! Thank you for digging into this so thoroughly. That issue is v useful for context

PR #2855, opened 2024-03-27
Raise a warning for unused binary operations

### LINKED ISSUE #2784: Unused string concatenation does not generate warning
It is very easy to make the mistake of leaving out a `<>` in a long statement of many string concatenations, i.e.:
`"foo" <> a <> "bar" b <> "baz`
The issue is the because this is valid code and will be parsed as two separate expressions, and only the last expression will be returned from the block, you'll end up with the previous concatenations before the missing `<>` being dead/unused code. You'll get no warning either to indicate that you've done something unintentional either.
Ideally this unused code should generate a warning to make it easier to see that you've made a mistake. This would require checking for expression that only contain `<>` and, are not the last statement in the block, and are not assigned to anything.
This issue also applies to basically any pure function or operator as well, such as `1 + 2 + 3 4 + 5 + 6` which will discard the `1 + 2 + 3` as it is parsed as a separate unused expression. The only way to solve this generally is to know whether all the function calls and operators within an expression are pure or not, and give a warning on any unused expressions that only contain pure functions and operators.
Given that there doesn't look to be a way currently to mark functions as pure (though a list of pure/unpure functions in the stdlib could be made for the compiler to use), I think the best solution currently is to just stick to operators, since as far as I'm aware there are no unpure operators, and just put the warning on any unused expressions that only contain operators (so no function calls).

giacomocavalieri:
Closes #2784

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -397,11 +397,7 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 .emit(Warning::ImplicitlyDiscardedResult {
                     location: discarded.location(),
                 });
-        } else if discarded.is_binop()
-            || discarded.is_record_constructor()
-            || discarded.is_record_access()
-            || discarded.is_record_update()
-        {
+        } else if discarded.is_pure_value_constructor() {
             self.environment.warnings.emit(Warning::UnusedValue {
                 location: discarded.location(),
             })
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -391,13 +391,16 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             self.environment.warnings.emit(Warning::UnusedLiteral {
                 location: discarded.location(),
             });
-        }
-        if discarded.type_().is_result() {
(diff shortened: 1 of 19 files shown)


lpil (commented):
Nice, thank you. Let's do this for any value that is constructed and immediately discarded rather than just binary operators.

compiler-core/src/type_/expression.rs
```diff
@@ -391,13 +391,20 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+        } else if discarded.type_().is_result() {
             self.environment
                 .warnings
                 .emit(Warning::ImplicitlyDiscardedResult {
                     location: discarded.location(),
                 });
+        } else if discarded.is_binop()
+            || discarded.is_record_constructor()
+            || discarded.is_record_access()
+            || discarded.is_record_update()
```
lpil, line 403: What do you think of something like `.is_pure_value_constructor()` or something like that? The if new variants are added the compiler will prompt us to decide if it's pure, while without pattern matching it won't.

giacomocavalieri: I think that's a great idea! I'll add it

lpil (approved):
Beautiful! Thank you!!

Issue #2844, opened 2024-03-26
`gleam publish` should be able to use a long-lived hex api key to auth publishing to hex.


Skenvy:
At the moment the `gleam publish` command requires `HEXPM_USER` and `HEXPM_PASS` to authorise publishing. It would be nice to not require the user password to need to be used for this, instead making available a preferred option of a hex api key ~ `HEXPM_API_KEY` ~ so that users can manage their own long lived API keys if they prefer.

Skenvy:
_I would like to start working on this._

lpil:
Fab! Thank you!

TomBinford:
Should this be closed now?

lpil:
Thank you

PR #2856, opened 2024-03-27
Let `ApiKeyCommand` use and prefer a long lived api key `HEXPM_API_KEY`

Skenvy:
Addresses #2844 

Add a `HEXPM_API_KEY` which `ApiKeyCommand` will prefer, but will check if it's supplied or not and fall back to requesting the username and password, unless either the api key or username AND password are supplied via the env vars, in which case, it will prefer the api key. But if an API key isn't supplied to the prompt and a username and password are supplied, it will use those, and only prompt for them if no API key is given to the prompt for one if no credentials are supplied from env vars. Then wrap the creating and removing of the temporary key produced using the username and password with the check from whether or not a long lived api key was provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -101,6 +101,8 @@
   instead of relative.
 - The `gleam publish` command now asks for confirmation if the package repository
   URL doesn't return a successful status code.
+- `gleam publish` can now optionally use a Hex API key to authorise publishing
+  and retiring packages, set via the environment variable `HEXPM_API_KEY`.
 
 ### Language Server
 
--- a/compiler-cli/src/hex.rs
+++ b/compiler-cli/src/hex.rs
@@ -9,6 +9,8 @@ const USER_PROMPT: &str = "https://hex.pm username";
 const USER_KEY: &str = "HEXPM_USER";
 const PASS_PROMPT: &str = "https://hex.pm password";
 const PASS_KEY: &str = "HEXPM_PASS";
+const API_KEY_PROMPT: &str = "https://hex.pm API key";
+const API_KEY_KEY: &str = "HEXPM_API_KEY";
 
 /// A helper trait that handles the provisioning and destruction of a Hex API key.
 pub trait ApiKeyCommand {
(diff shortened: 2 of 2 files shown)


lpil (commented):
Hello! Thank you for this.

I'm finding it a bit hard to follow this code as there's a lot of boolean logic in one function, and I struggle a bit with boolean blindless.

Could we do something like this instead:

```rust
let key = var(API_KEY_KEY).unwrap_or_default().trim();
if key.is_empty() {
  self.with_new_api_key(...)
} else {
  self.with_api_key(...)
}
```

Then the two paths have been split into two functions, making it easier to follow each one.

Skenvy:
The only problem then is that without just moving the same logic inside `with_new_api_key`, we wouldn't be able to check if the username and password are supplied or not before checking if it's a good idea to prompt for the api key before prompting for them or not, i.e. the api key will never be prompted for, if you're happy with that.

lpil:
Yup let's use environment variables for the key always.

Skenvy:
I'm not particularly rusty, is there a preference between;
```rust
if api_key.is_empty() {
    let result = self.with_new_api_key(runtime.handle(), &hex_config);
} else {
    let result = self.with_api_key(runtime.handle(), &hex_config, &api_key);
}

result
```
and
```rust
if api_key.is_empty() {
    self.with_new_api_key(runtime.handle(), &hex_config)
} else {
    self.with_api_key(runtime.handle(), &hex_config, &api_key)
}
```

Skenvy:
Ok it compiles and works as expected now (publishes / retires with the key and user+pass), and only prompts for user/pass with no api key given.

lpil:
> I'm not particularly rusty, is there a preference between;
> 
> ```rust
> if api_key.is_empty() {
>     let result = self.with_new_api_key(runtime.handle(), &hex_config);
> } else {
>     let result = self.with_api_key(runtime.handle(), &hex_config, &api_key);
> }
> 
> result
> ```
> 
> and
> 
> ```rust
> if api_key.is_empty() {
>     self.with_new_api_key(runtime.handle(), &hex_config)
> } else {
>     self.with_api_key(runtime.handle(), &hex_config, &api_key)
> }
> ```

The first is a compile error as the variable `result` isn't in scope any more after the block. The other looks good 👍

lpil (approved):
Great! Thank you.

Would you mind updating the changelog? Thank you

PR #2857, opened 2024-03-27
Improve error message for wrong patterns

### LINKED ISSUE #2841: Improve error message for wrong patterns using qualified names
Consider the following snippet of code:
```gleam
pub type Chest(a) {
  Chest(pwd: String, treasure: a)
}

pub fn open_chest(chest, pwd) {
  case pwd {
    chest.pwd -> Ok(chest.treasure)
    _ -> Error(Nil)
  }
}
```
That results in an error like this one:
```gleam
    chest.pwd -> Ok(chest.treasure)
//        ^^^
// Syntax error
// I'm expecting a type name here.
// Hint: Type names start with a uppercase letter,
// and can contain a-z, A-Z, or 0-9.
```
Which is totally right: when using the `.`-syntax in a pattern it means we're trying to match on a type coming from the `chest` module and `pwd` is not a type name as it should be.

However, it might not be super clear why this is wrong to someone who's just getting started with Gleam. Especially if they are wrongly assuming that pattern matching checks for equality: a slightly modified version of this snippet was sent to me by @GioeleBucci (thank you for pointing this out!!) who was trying to check that `pwd` equals the `pwd` field held in the `chest` variable.

A slightly better version of this error could highlight the whole pattern and point out that that's not a valid pattern:
```gleam
    chest.pwd -> Ok(chest.treasure)
//  ^^^^^^^^^
// Syntax error
// I'm expecting a pattern here.
// Hint: A pattern can be a type Name, a literal value or a variable
// to bind a value to.
```
Not sure if this is the best phrasing here but it could be a start

giacomocavalieri:
This PR closes #2841

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,8 @@
   `prepend` function imported from the prelude in the JavaScript target.
 - Fixed string prefix matching producing wrong results on the JavaScript target
   when the prefix had a Unicode codepoint escape sequence (`\u{...}`).
+- Improved error message for wrong patterns using constructors from other
+  modules.
 
 ### Formatter
 
@@ -90,7 +92,6 @@
 - Go to definition now works for values in dependency Gleam modules.
 - Completions are now provided for module imports.
 
-
 ## v1.0.0 - 2024-03-04
 
 ### Language changes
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -41,6 +41,8 @@
   `prepend` function imported from the prelude in the JavaScript target.
 - Fixed string prefix matching producing wrong results on the JavaScript target
   when the prefix had a Unicode codepoint escape sequence (`\u{...}`).
+- Improved error message for wrong patterns using constructors from other
+  modules.
```
lpil, line 45: Revert the other changes in this file please 🙏

lpil (approved):
Thank you

PR #2860, opened 2024-03-28
Fix issue with lsp not compiling on editor load

Acepie:
It looks like with the changes to https://github.com/gleam-lang/gleam/pull/2731 we only compile modules that have changed based on lsp messages for didOpen/didChange/didSave. However, on first lsp load/editor open (at least as tested with vs code) the very first call to compile_please would be caused by the initialization of the lsp itself (it was hard to tell exactly where but I assume it was the very first loop of the message `receive` not getting events in the first x milliseconds before the initial didOpen event). As such, the first call to compile_please would be https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/language_server/server.rs#L347 which would get called with the root directory itself. This caused issues because `find_gleam_project_parent` would not recognized that the root directory was itself the "parent project". This change fixes this issue by checking the directory itself first if the passed in path is a directory

### DIFF
--- a/compiler-core/src/language_server/router.rs
+++ b/compiler-core/src/language_server/router.rs
@@ -59,7 +59,6 @@ where
         // If the path is the root of a known project then return it. Otherwise
         // find the nearest parent project.
         let path = if self.engines.contains_key(&path) {
-            tracing::info!(?path, "found engine for path");
             path
         } else {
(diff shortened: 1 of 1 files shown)


lpil (approved):
Very nice, thank you.

PR #2861, opened 2024-03-28
fixed goto for cached dependencies

Acepie:
Realized while debugging https://github.com/gleam-lang/gleam/pull/2860/files that the solution I wrote for https://github.com/gleam-lang/gleam/pull/2731 didn't actually work for non-clean builds. This adds a basic way to handle cached modules by reading the package src directly and constructing the missing source information. This seems to work well when I tested locally though there would maybe be a long term benefit to sticking this line number info in the cache itself

### DIFF
--- a/compiler-core/src/language_server/compiler.rs
+++ b/compiler-core/src/language_server/compiler.rs
@@ -123,15 +123,11 @@ where
             let build_path = self
                 .project_compiler
                 .get_build_dir_for_module(&module.package, &module.name);
-            // Read the source
-            let code = self.project_compiler.io.read(&build_path);
-            if let Ok(code) = code {
-                // Create the source information
-                let path = build_path.as_os_str().to_string_lossy().to_string();
-                let line_numbers = LineNumbers::new(&code);
-                let source = ModuleSourceInformation { path, line_numbers };
-                _ = self.sources.insert(name.clone(), source);
-            }
+            // Create the source information
+            let path = build_path.as_os_str().to_string_lossy().to_string();
+            let line_numbers = module.line_numbers.clone();
+            let source = ModuleSourceInformation { path, line_numbers };
+            _ = self.sources.insert(name.clone(), source);
         }
 
         // Warnings from dependencies are not fixable by the programmer so
--- a/compiler-core/src/language_server/compiler.rs
+++ b/compiler-core/src/language_server/compiler.rs
@@ -114,6 +114,26 @@ where
             _ = self.sources.insert(module.name.clone(), source);
         }
 
+        // Since cached modules are not recompiled we need to manually add them
+        for (name, module) in self.project_compiler.get_importable_modules() {
+            if self.sources.contains_key(name) {
+                continue;
+            }
+            // Get the build path for the module
+            let build_path = self
+                .project_compiler
+                .get_build_dir_for_module(&module.package, &module.name);
+            // Read the source
+            let code = self.project_compiler.io.read(&build_path);
+            if let Ok(code) = code {
+                // Create the source information
+                let path = build_path.as_os_str().to_string_lossy().to_string();
+                let line_numbers = LineNumbers::new(&code);
+                let source = ModuleSourceInformation { path, line_numbers };
+                _ = self.sources.insert(name.clone(), source);
+            }
+        }
+
         // Warnings from dependencies are not fixable by the programmer so
         // we don't bother them with diagnostics for them.
         let _ = self.take_warnings();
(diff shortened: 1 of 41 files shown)


Acepie:
I managed to write a unit test that I can prove locally works intentionally by checking the code path it goes down but don't know of a way to show that in the test itself. Lmk if you can think of a better way to do this

lpil (commented):
Let's add it to the cache instead, rather than pulling information half from the cache and half not from the cache.

Acepie:
Sounds reasonable to me 👍🏽 do you think there's benefit to write the entire module src string or should I just make the line numbers themselves serializable and cache that?

lpil:
Hmm. Let's just pop in the line numbers for now as we already have the source code on disc

Acepie:
Ok so I know what bits of code need to be changed but am having trouble with getting the capn proto generator to generate in a way that is useful. it seems like something to do with versions of rust/capnp because it looks like it successfully generates the generated rs file but the change is massive and then there are compile issues when trying to use it. When you get time could you try doing a schema generate on main and see how much things change/if things change/work for you. And if it does "just work" could you let me know what versions you are using

Acepie:
So I technically got something that works well despite not actually figuring out the encoding/decoding with capnp. Since CacheMetadata is serialized without capnp and we need the line numbers in both structs to make passing the data along work, https://github.com/gleam-lang/gleam/pull/2861/files#diff-65daa92566552b98dec29695441a52791be43182c8183c2bcfbddc819e81bf37R138 this change to pass forward the line numbers from metadata to loaded cache object actually means the lsp sees the right line numbers. I'd still like to get the capnp serialization working correctly but maybe its not worth the effort if the long term goal is to replace that https://github.com/gleam-lang/gleam/issues/1599

Acepie:
Ok i investigated this a bit more and it looks like https://github.com/gleam-lang/gleam/commit/78c91b4e6bae6b36db0b7865466ffe6282aa79a5#diff-d92835fc3ab41cf44829a3f4ffaff8b03ffba535fa58a68dd4575377291e9341R71 this package bump might have had unintended consequences/seems to cause the capnp generated code to not build/not be compatible. I've reverted the change for now

Acepie:
Should I include a changelog entry for this? I didn't include one at first since its really a fix for https://github.com/gleam-lang/gleam/blob/main/CHANGELOG.md?plain=1#L103 which is still unreleased though the change does touch a lot of file and maybe the cache change in and of itself is worth mentioning?

compiler-core/src/language_server/compiler.rs
```diff
@@ -114,6 +114,22 @@ where
...
+        for (name, module) in self.project_compiler.get_importable_modules() {
+            if self.sources.contains_key(name) || name == "gleam" {
+                continue;
+            }
+            // Get the build path for the module
+            let build_path = self
+                .project_compiler
+                .get_build_dir_for_module(&module.package, &module.name);
+            // Create the source information
+            let path = build_path.as_os_str().to_string_lossy().to_string();
```
lpil, line 127: This is a utf8 path so should be able to do just `.to_string()`, no?

compiler-core/src/language_server/compiler.rs
```diff
@@ -114,6 +114,22 @@ where
             _ = self.sources.insert(module.name.clone(), source);
         }
 
+        // Since cached modules are not recompiled we need to manually add them
+        for (name, module) in self.project_compiler.get_importable_modules() {
```
lpil, line 118: This is loading information about cached modules every time the compiler runs, but they will have never changed. Instead we could do this once when the compiler is created. What do you think?

lpil (commented):
Thank you! Left a couple notes inline.

+ Yup I agree, the existing changelog entry is fine 👍

compiler-core/src/language_server/compiler.rs
Acepie: yup looks like this was a copy paste on my end. will fix

compiler-core/src/language_server/compiler.rs
Acepie: So the issue with that is that importable_modules won't actually have the right list until the first compile when the package loader is first run. For what its worth, https://github.com/gleam-lang/gleam/pull/2861/files/dcabe44c72ffcad5bf528a97dd0dc08adb6952dc#diff-9b6f5e817763c0c820d9ece57897c33972be558d7348018dc916fac80e5d9968R119 this check does prevent the bulk of the work since after the first compile the sources will already be present.

Acepie: I added a trace to the cache loading and confirmed it only goes through the loop on the very first compile which seems reasonable to me? I think the alternative would be to run the package_loader directly which could be doable though might be a bit more of a lift than necessary here?

Acepie:
The compiler does have logic to handle that by deleting the build dir/cache if the gleam version is different but since prs use the same 1.0.0 version as whatever the latest release is it doesn't trigger properly

lpil (approved):
Thank you!!

Issue #2862, opened 2024-03-28
Bit arrays in guards don't handle pattern variables in JS target


PgBiel:
Consider the following Gleam program:

```gleam
pub fn main() {
  case 5 {
    x if <<x>> == <<x>> -> True
    _ -> False
  }
}
```

This currently compiles to:

```js
import { isEqual, toBitArray } from "./gleam.mjs";

export function main() {
  let $ = 5;
  if (isEqual(toBitArray([x]), toBitArray([x]))) {
    let x = $;
    return true;
  } else {
    return false;
  }
}
```

which, when run, leads to

```
  if (isEqual(toBitArray([x]), toBitArray([x]))) {
                          ^

ReferenceError: x is not defined
    at main (file:///dev/shm/showcase/build/dev/javascript/showcase/showcase.mjs:5:27)
    at file:///dev/shm/showcase/build/dev/javascript/showcase/gleam.main.mjs:2:1
    at ModuleJob.run (node:internal/modules/esm/module_job:195:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:336:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:106:12)
```

The problem is in `guard_constant_expression` (`javascript/expression.rs`), introduced in #2002 (it appears that BitArrays were not supported in the JS target back then, so it's likely the function just became outdated after they were added). We can see the following line:

https://github.com/gleam-lang/gleam/blob/2dd5c731871f9bbf4077c9efa2a428e7b11cce7a/compiler-core/src/javascript/expression.rs#L1230

That is, it delegates any types of constant expressions it didn't explicitly handle to `constant_expression`. This means that `constant_expression` will handle BitArrays since they aren't handled by the former, and thus pattern variables won't be replaced by subject + path, leading to the error above.

The fix involves making the match in `guard_constant_expression` handle BitArrays separately. While it's possible that the `bit_array` function (which currently calls `constant_expression` within itself) could have to be duplicated, a bit of generics could possibly avoid that (requires investigation).

PR #2863, opened 2024-03-28
Fix bit arrays in guards not handling pattern variables in JS

PgBiel:
Fixes https://github.com/gleam-lang/gleam/issues/2862

Explicitly handles bitarrays in guards, avoiding problems with pattern-bound variables being used (within the compiled condition) before they are defined (inside the compiled `if`'s body).

I used generics here to avoid having to duplicate the `bit_array` function for each type of `constant_expression`-like function. Feel free to suggest alternative implementations if desirable.


lpil (approved):
Thank you

Issue #2826, opened 2024-03-24
provide a clearer message when using a reserved word incorrectly


grottohub:
as of right now, when using a keyword somewhere that's unexpected, such as:

```gleam
pub type Message {
  Message(type: String)
}
```

the compiler will error out with an `UnexpectedToken` [error](https://github.com/gleam-lang/gleam/blob/5ade71380454d298d542012b7be82bec461e4257/compiler-core/src/parse/error.rs#L184): `Expected one of: ")"`, which can be a bit confusing, especially for new programmers who aren't aware of the usual "can't use reserved words" rule.

happy to work on this myself

Acepie:
FWIW it looks like this is a general problem whenever the compiler expects a "series of things" (types in a record, args in a function definition, etc.). The compiler checks for N things until it fails to parse another thing and then immediately checks for a closing paren, bracket, etc. Would it make sense in these cases to have the expected series item in the list of expected things? so like: `Expected one of: ")", "a function argument declaration"` or something along those lines?

lpil:
Sounds good to me!

PR #2871, opened 2024-03-29
Improve messaging when failing to parse a series

Acepie:
See #2826 for more details. In cases where the parser was expecting a series of things followed by a closing, we were only suggesting the closing. This pr adds the expected series as part of the error message.


lpil (approved):
Beautiful!! Thank you

Issue #2867, opened 2024-03-29
`const` are not in scope inside binary scalars


dartheian:
While attempting the following:
```gleam
const bit_size = 3

fn my_function() -> BitArray {
    <<1:size(bit_size)>>
}
```
the compiler complains about the `const` not being in scope:
```bash
error: Unknown variable
The name `bit_size` is not in scope here.
```

lpil:
Thank you

larzconwell:
I believe this can be closed since it was fixed by https://github.com/gleam-lang/gleam/pull/2874, I verified this works as expected on `main`.

lpil:
Thank you

PR #2874, opened 2024-03-30
Fix issue with using `const` value in BitArray size option

kritik94:
Issue https://github.com/gleam-lang/gleam/issues/2867.
As I research, right now value from BitArray size ignored - and this affect how dependency sorted in `gleam_core::call_graph::into_dependency_order`.


lpil (commented):
Thank you. Could you update the changelog please?

kritik94:
Absolutely!
I've made the update in a separate commit.
Please feel free to suggest any modifications to the wording if needed.

Issue #2865, opened 2024-03-28
Check that the source code repo URL 200s when publishing and ask the user for confirmation if it does not


lpil:
To avoid typos in configuration

```toml
name = "wibble"
version = "1.0.0"
description = "wobble"
licences = ["Apache-2.0"]
repository = { type = "github", user = "wubble", repo = "oh-no-this-is-the-wrong-name" }

[dependencies]
gleam_stdlib = "~> 0.34 or ~> 1.0"

[dev-dependencies]
gleeunit = "~> 1.0"
```

BradLewis:
I can have a go at this.

I was having a quick look and the most obvious place to do this check is in the `setup` for the publisher command, along with the version and name checks. However, there is no tokio runtime at that stage, so I see 3 easy options

1. Add the tokio runtime to the `setup` method
2. Use something like `reqwest::blocking`
3. Run the check in the `with_api_key` method instead where the runtime exists.

Either way it should be pretty straightforward. LMK what you think or if there's an option I missed!

lpil:
Let's do it using the blocking API in setup 👍

PR #2881, opened 2024-04-02
Add check and prompt for when source code repo URL doesn't return 200s

BradLewis:
Resolves https://github.com/gleam-lang/gleam/issues/2865


lpil (approved):
Thank you! Perfect!

PR #2882, opened 2024-04-02
fix cached goto definition for local packages and refactor testing suite to be more consistent across lsp

Acepie:
Another fix for the goto definitions (hopefully the last one for real this time). Realized when I did some further testing with a package with local dependencies that the source path was not being computed correctly. Opted to store the source path in the cache as well. While writing up new test cases to handle this I ended up cleaning up/combining the testing patterns for the lsp

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -124,6 +124,7 @@ pub fn infer_module<A>(
     direct_dependencies: &HashMap<EcoString, A>,
     target_support: TargetSupport,
     line_numbers: LineNumbers,
+    src_path: EcoString,
 ) -> Result<TypedModule, Error> {
     let name = module.name.clone();
     let documentation = std::mem::take(&mut module.documentation);
@@ -278,6 +279,7 @@ pub fn infer_module<A>(
             unused_imports,
             contains_todo,
             line_numbers,
+            src_path,
         },
     })
 }
--- a/compiler-core/generated/schema_capnp.rs
+++ b/compiler-core/generated/schema_capnp.rs
@@ -489,6 +489,14 @@ pub mod module {
     pub fn has_line_numbers(&self) -> bool {
       !self.reader.get_pointer_field(7).is_null()
     }
+    #[inline]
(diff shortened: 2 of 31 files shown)


Acepie:
The issue was the path resolution when going to a definition that was in a local package. Since for local package we don't have sources in the packages dir instead we want to link to the actual source wherever the package was originally.

The easiest way to repro is to open a project with a local dep and try going to the definition from the local dep (i realized this while testing something in the p5js examples)

Acepie:
It would instead try to go to a file that doesn't actually exist which depending on the ide has different effects (when I tested it would create a new empty file in the packages dir where it expected the source to be)

Acepie:
https://github.com/gleam-lang/gleam/pull/2882/files#diff-a5468dcd1cd8deb85fe968b5881424b5125ebf61c339dce8b665494612e4190fR483 this test is the one that if written appropriately would fail on the current main due to the issue. its a bit hard to see with the test rewriting I did (lemme know if you'd prefer I undo that btw. I realized it was a bit frustrating realizing how much I essentially had to copy paste because of edge cases that didn't quite fit into the existing abstractions)

lpil:
Ah thank you! That makes sense. Rather than duplicate the source let's read path deps from their original location.

Acepie:
This change is just storing the source path in cache, not the source itself so as implemented it does just link to the original source. There was an alternative way to implement this without putting anything in the cache if we put another hashmap in the compiler itself but that seemed like it would be a bit more invasive/round about of a change

lpil:
oooh sorry! I misread the PR. 😵  Let me review again. Thank you

compiler-core/src/analyse.rs
```diff
@@ -124,6 +124,7 @@ pub fn infer_module<A>(
     direct_dependencies: &HashMap<EcoString, A>,
     target_support: TargetSupport,
     line_numbers: LineNumbers,
+    src_path: EcoString,
```
lpil, line 127: This is a path so let's use a Utf8Path

lpil (approved):
Thank you!

Let's always use paths and not strings to represent paths. I'll merge and then sort that myself :)

Issue #2888, opened 2024-04-03
Disallow creating new projects in non-utf8 paths


zahash:
gleam won't let me create a project with non utf8 name but i can create a project in a directory where one of the path segments is non-utf8.

eg:

```sh
mkdir $'\xC0\xC0'
cd $'\xC0\xC0'

gleam new samplegleam
cd simplegleam
```

however this is NOT the intended behavior beause when i run any command, like `gleam build` or `gleam run` etc... they all fail.

```sh
error: Fatal compiler bug!

This is a bug in the Gleam compiler, sorry!

Please report this crash to https://github.com/gleam-lang/gleam/issues/new
and include this error message with your report.

Panic: compiler-cli/src/main.rs:558
	Failed to get current directory: NonUtf8Path { path: "/home/zahash/source/\xC0\xC0/samplegleam" }
Gleam version: 1.0.0
Operating system: linux

If you can also share your code and say what file you were editing or any
steps to reproduce the crash that would be a great help.

You may also want to try again with the `GLEAM_LOG=trace` environment
variable set.
```

the problem is because when i do `gleam new samplegleam`, the `NewOptions.project_root: String` is set to "simplegleam". and the `Creator::new` function makes a `Utf8PathBuf` with just the "simplegleam" string and doesn't check its full path `/home/zahash/source/��/samplegleam`

the root cause of this is the improper path checking when a new instance of `Utf8PathBuf` is created from a `String`. it assumes that since the `Utf8PathBuf` is created from a `String` it must be valid utf8. But it fails to consider that the `String` might be just a relative path. The absolute path might contain non-utf8 segments.

```rust
impl Creator {
    fn new(options: NewOptions, gleam_version: &'static str) -> Result<Self, Error> {
        // ...
        let root = Utf8PathBuf::from(&options.project_root);
        // ...
    }
}
```

the reason why `gleam run` and `gleam clean` catch this error is because they use `find_project_paths` function which uses `get_current_directory` which uses `std::env::current_dir()` which gives the absolute path of the current directory. So, when it is converted to `Utf8PathBuf`, it fails. 

```rust
// compiler-cli/src/run.rs
pub fn command(
    arguments: Vec<String>,
    target: Option<Target>,
    runtime: Option<Runtime>,
    module: Option<String>,
    which: Which,
) -> Result<(), Error> {
    let paths = crate::find_project_paths()?;
    // ...
}

// compiler-cli/src/main.rs
[...]

PR #2889, opened 2024-04-03
disallow creating new projects in non-utf8 paths

zahash:
fixes https://github.com/gleam-lang/gleam/issues/2888


lpil (approved):
Thank you!!

PR #2896, opened 2024-04-05
Feat(lsp): Adding Pipeline Suggestions Feature to Gleam LSP

katsmil:
This extension to the Gleam LSP implementation aims to provide users with pipeline suggestions. To ensure the effectiveness of these suggestions, especially when functions are chained through intermediate variables, an additional code action implementation for inlining local variables has been included.

This PR is associated with [discussion](https://github.com/gleam-lang/gleam/discussions/2573) that I initiated:

Changes included in this PR:
* pruning AST tree.
* early exit AST tree.
* possibility to lazily resolve costly code actions.
* usage of coverage marks to potentially improve the test suite.
* Translation of AST nodes to string representation, such that it can be used as a text edit for the client to receive.
* LSP server can return code actions over a selection of code lines.

It's worth noting that I initially embarked on this project as part of my thesis work. As this is my first contribution to the project, I welcome any feedback. I'm committed to further refining this implementation.
Thank you! 🙂


compiler-core/src/ast/typed.rs
```diff
@@ -495,6 +556,112 @@ impl TypedExpr {
             _ => false,
         }
     }
+
+    pub fn to_string(&self) -> Option<EcoString> {
```
lpil, line 560: Let's move this out into its own printer class and implement it by appending to a buffer to avoid all the extra allocations.

lpil (commented):
Wow, thank you for this! Looks great.

I'm finding this quite hard to understand and review due to being a large amount of code and there being no documentation and very few comments.

Perhaps we break this down into multiple PRs to enable it to get in easier? For example, the AST printing could be a PR in itself.

Could you also add comments and documentation to the code throughout? To make it clear what the purpose and intent is for each type, function, and logical decision. Thank you.

> It's worth noting that I initially embarked on this project as part of my thesis work.

Oh really cool! What's your thesis work?

Issue #2891, opened 2024-04-03
LSP: do not autocomplete internal modules for other packages


lpil:

Acepie:
So I assume for this we would need to add information to the ModuleInterface's so that we can tell if a module is internal or not right? Because rn I think the only point where you can check if a module is internal is when looking at PackageConfig which the LSP can't access atm. I'm mainly asking for clarification because I was thinking about doing this and realized if so it would conflict with the other PR I have open atm which might be annoying to resolve if running the generator

lpil:
Sounds good! Let's do that.

PR #2900, opened 2024-04-06
Add warning for internal types in public API

### LINKED ISSUE #2234: ✨ Warn about internal types used in the public API.
The `internal_modules` property in our `gleam.toml` config lets us define modules that won't be documented when generating gleam's package docs. Because gleam has no formal notion of "internal" modules, these modules can still be imported and used freely.

This is all well and good until a function in a package refers to an internal type:

```gleam
pub fn with_handler(func: fn(Message, Subject(HandlerMessage)) ->
    Result(Nil, Nil)) -> WebsocketHandler
```

[Here](https://hexdocs.pm/mist/mist/websocket.html#with_handler) `WebsocketHandler` is an internal type and so has no documentation. The DX is very poor in this case, I had to dig into the source to work out where it was coming from!

It would be great if this emitted an error the same way that using a private type in the public API does, but I'd settle for a warning too ^.^

giacomocavalieri:
This PR closes #2234


compiler-core/src/type_.rs
```diff
@@ -55,6 +55,7 @@ pub enum Type {
     ///
     Named {
         publicity: Publicity,
+        from_internal_module: bool,
```
lpil, line 58: I don't think this information should be part of the type itself, rather we can look up whether the module that defines the type is internal.
Generally I'd like to avoid adding anything new to the type data structures.

We'll need to store whether a module is internal for this ticket too https://github.com/gleam-lang/gleam/issues/2891

lpil (commented):
Thank you! I've left a note inline about where we store the internal-or-not info

compiler-core/src/erlang/tests/records.rs
```diff
@@ -49,6 +49,7 @@ fn module_types() {
             "name",
             Arc::new(type_::Type::Named {
                 publicity: Publicity::Public,
+                from_internal_module: false,
```
lpil, line 52: Isn't this duplicating the publicity field? It has a `Publicity::Internal` variant

giacomocavalieri: I thought about it but then the formatter would then put the `@internal` annotation above everything inside an internal module

lpil: The formatter doesn't run analysis so it wouldn't know the module is internal

giacomocavalieri: Ah I didn't think about it! Thanks I've removed the redundant field

compiler-core/src/analyse.rs
```diff
@@ -405,9 +405,16 @@ fn register_types_from_custom_type<'a>(
 
     hydrator.clear_ridgid_type_names();
 
+    // If the type comes from an internal module we change it's publicity to
+    // `Internal`.
+    let publicity = if config.is_internal_module(module) {
```
lpil, line 410: Doesn't this now make private ones internal?

giacomocavalieri: Oh you're right! I have to restrict it to internal only if it's public. I'll add a test

compiler-core/src/analyse.rs
```diff
@@ -126,6 +127,7 @@ pub fn infer_module<A>(
     target_support: TargetSupport,
     line_numbers: LineNumbers,
     src_path: Utf8PathBuf,
+    config: &PackageConfig,
```
lpil, line 130: This is now duplicating the `package` parameter above. Let's remove that and take it from this config

lpil (commented):
One little thing below!

lpil (approved):
Thank you!

PR #2903, opened 2024-04-06
Check if package being published is a real project

lpil:


compiler-cli/src/publish.rs
```diff
@@ -131,6 +88,98 @@ The repository URL {} returned status {}
     }
 }
 
+fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
+    if package.modules.len() > 1 {
```
giacomocavalieri, line 92: I don't remember, does this count test modules as well? Because if those are counted as well then every default package would come with two modules by default: the `<name>` and `<name>_test` one

giacomocavalieri (approved):
I left just a couple of comments

compiler-cli/src/publish.rs
```diff
@@ -131,6 +88,98 @@ The repository URL {} returned status {}
...
+        println!("more than 1");
```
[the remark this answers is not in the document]
lpil: ooops hahaha. Very sloppy of me

compiler-cli/src/publish.rs
lpil: It doesn't in this case because it has been compiled in prod mode for publishing

PR #2905, opened 2024-04-07
Only autocomplete module or direct dep imports

### LINKED ISSUE #2902: Do not suggest completions for indirect deps

Acepie:
Closes #2902

Checks the package config to get the list of direct dep packages and filters out any packages that are not part of a direct dep, the root package, or the prelude. For tests it also includes the list of direct dev deps.

Also swaps the order of the filters so that the ones that are likely to remove the most packages (roughly) go first

### DIFF
--- a/compiler-core/src/language_server/tests/completion.rs
+++ b/compiler-core/src/language_server/tests/completion.rs
@@ -1216,6 +1216,208 @@ pub fn main() {
 }
 
 #[test]
+fn completions_for_an_import_not_from_dev_dependency() {
+    let code = "import gleam
+
+pub fn main() {
+  0
+}";
+    let dep = "";
+
+    let mut io = LanguageServerTestIO::new();
+    let _ = io.hex_dep_module("indirect_package", "indirect_module", "");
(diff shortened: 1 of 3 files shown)


compiler-core/src/language_server/tests/completion.rs
```diff
@@ -1144,6 +1144,77 @@ pub fn main() {
...
+}";
+    let dep = "";
+
+    let mut io = LanguageServerTestIO::new();
+    let _ = io.hex_dep_module("indirect_package", "indirect_module", "");
+    let test_project = TestProject::for_source(code).add_hex_module("example_module", dep);
+    let mut engine = test_project.build_engine(&mut io);
+
+    // Manually add an indirect dependency to the project
+    let compiler = &mut engine.compiler.project_compiler;
```
Acepie, line 1162: This feels kinda bad but I couldn't think of a better way to simulate a dep of dep. I did confirm that without the actual change in `engine.rs` this test fails though. Testing out the change in IDE looks good

lpil (approved):
Beautiful, thank you!!

Issue #2904, opened 2024-04-06
Non-Empty array (represented as `[Nil]`) that should be empty. JS target.


keroami:
Gleam git main (176334cd); gleam 1.0 runs fine.
JavaScript target - any runtime, since the generated JS contains
```
                                                                          return toList([
                                                                            ,
                                                                          ])
```
where JS has `[,]` the same as `[undefined]`.

 The following test should pass, but fails with `[Nil]` unequal to `[]`:
```
import gleam/dynamic as d
import gleam/json
import gleam/list.{map}
import gleam/result.{try}
import gleam/string.{split}
import gleeunit/should

pub fn bug_test() {
  "{ \"special\": \"\", \"f\": 42 }"
  |> json.decode(decoder)
  |> should.equal(Ok([]))
}

type Special {
  Special(String)
}

fn decoder(dyn) {
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use _ <- try(d.optional(d.field("f", d.int))(dyn))
  use special <- try(d.field("special", d.string)(dyn))
  case split(special, ", ") {
    [""] -> []
    props ->
      props
      |> map(Special)
  }
  |> Ok
}
```

PR #2908, opened 2024-04-07
Fix empty lists in tight positions having length of one on JS

PgBiel:
Fixes https://github.com/gleam-lang/gleam/issues/2904

Now we only add a trailing comma to arrays which generate lists if the lists actually have elements - `[ , ]` actually has a single `undefined` element, so we shouldn't compile an empty list to `toList([,])`. The same is done for newlines after the first `[` (if there are no elements, always render as `[]`, not `[\n]`).


compiler-core/src/javascript/tests/lists.rs
```diff
@@ -99,3 +99,33 @@ fn go(xs) {
...
+  use _ <- func(1, 2)
+  use _ <- func(1, 2)
+  use _ <- func(1, 2)
+  use _ <- func(1, 2)
+  use _ <- func(1, 2)
+  use _ <- func(1, 2)
+  case [] {
+    [] -> []
+    _ -> []
+  }
```
giacomocavalieri, line 127: Do you think this could be rewritten just using a variable with a very long name that would trigger the same error?
```gleam
fn go() {
  let aaaaaaaaaa...a = []
//              ^^^ This would have to be long enough to trigger the
//                  code generation bug    
} 
```
So the snapshot could be easier to review

compiler-core/src/javascript/expression.rs
```diff
@@ -1357,12 +1357,16 @@ pub fn string(value: &str) -> Document<'_> {
...
+    } else {
+        (break_("", ""), break_(",", ""))
+    };
     Ok(docvec![
         "[",
-        docvec![break_("", ""), elements].nest(INDENT),
-        break_(",", ""),
+        docvec![initial_break, elements].nest(INDENT),
+        trailing_comma,
         "]"
```
giacomocavalieri, line 1370: I think an easier way to go about this would be something like this:
```rs
if elements.is_empty() {
  Ok("[]".to_doc())
} else {
  Ok(docvec![
    "[",
    docvec![break_("", ""), elements].nest(INDENT),
    break_(",", ""),
    "]",
  ])
}
```
So it's easier to see that the list is never split when there's no elements

lpil: I like this idea here, but I would like to move away from this for the formatter so that you can have comments in an empty list

compiler-core/src/javascript/tests/lists.rs
PgBiel: Sure! Should be done now.

compiler-core/src/javascript/expression.rs
PgBiel: Good idea! It has been done.

giacomocavalieri:
Awesome! This looks great 💖

lpil (approved):
Amazing, thank you

PR #2914, opened 2024-04-07
initial implementation of goto types and separating annotations

Acepie:
An initial implementation for making annotations distinct from the thing being annotated. Currently in most places where the user hovers over an annotation it either doesn't work or hovers the larger object being annotated (function head, constant, custom type). This change is the first part of separating that out and allowing for operations on the located annotation. Currently that means hovering will display the type itself, completions will always be type completions (this seems to be exactly the same as current in all cases), and goto definition now will go to where the type is declared. There are a couple known gaps here but overall its still an improvement so want to start with this and get feedback

Known things to improve:

- Type aliases seem to be stored in the ast directly as the type being aliased meaning that hover/goto will use the underlying type
- ~Hovering over a type won't show documentation~
- ~Because of how locations/labels for CustomType constructor args are implemented hovering over the arg label will be treated as hovering over the annotation~
  -  ~This is probably fixable but honestly I kinda like it so wanna get thoughts~

### DIFF
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -300,9 +300,30 @@ pub enum Located<'a> {
 }
 
 impl<'a> Located<'a> {
+    fn type_location(
+        &self,
+        importable_modules: &'a im::HashMap<EcoString, type_::ModuleInterface>,
+        type_: std::sync::Arc<type_::Type>,
+    ) -> Option<DefinitionLocation<'_>> {
+        match type_.as_ref() {
+            type_::Type::Named { name, module, .. } => importable_modules
+                .get(module)
+                .and_then(|i| i.types.get(name))
+                .map(|t| DefinitionLocation {
+                    module: Some(&t.module),
+                    span: t.origin,
+                }),
+            type_::Type::Var { type_, .. } => type_
+                .borrow()
+                .nested_type()
+                .and_then(|v| self.type_location(importable_modules, v)),
+            _ => None,
+        }
+    }
+
     pub fn definition_location(
         &self,
-        importable_modules: &im::HashMap<EcoString, type_::ModuleInterface>,
+        importable_modules: &'a im::HashMap<EcoString, type_::ModuleInterface>,
     ) -> Option<DefinitionLocation<'_>> {
         match self {
             Self::Pattern(pattern) => pattern.definition_location(),
@@ -314,18 +335,7 @@ impl<'a> Located<'a> {
                 span: statement.location(),
             }),
             Self::Arg(_) => None,
-            Self::Annotation(_, type_) => match type_.as_ref() {
-                // For type annotations, we need to look up the type location
-                // From the module interface.
-                type_::Type::Named { name, module, .. } => importable_modules
-                    .get(module)
-                    .and_then(|i| i.types.get(name))
-                    .map(|t| DefinitionLocation {
-                        module: Some(&module),
-                        span: t.origin,
-                    }),
-                _ => None,
-            },
+            Self::Annotation(_, type_) => self.type_location(importable_modules, type_.clone()),
         }
     }
 }
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -315,7 +315,7 @@ impl<'a> Located<'a> {
                 }),
             type_::Type::Var { type_, .. } => type_
                 .borrow()
-                .nested_type()
+                .inner_type()
                 .and_then(|v| self.type_location(importable_modules, v)),
             _ => None,
         }
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -296,13 +296,25 @@ pub enum Located<'a> {
     ModuleStatement(&'a TypedDefinition),
     FunctionBody(&'a TypedFunction),
     Arg(&'a TypedArg),
-    Annotation(&'a TypeAst, std::sync::Arc<type_::Type>),
+    Annotation(SrcSpan, std::sync::Arc<type_::Type>),
 }
 
 impl<'a> Located<'a> {
+    // Looks up the type constructor for the given type and then create the location.
+    fn type_location(
+        &self,
+        importable_modules: &'a im::HashMap<EcoString, type_::ModuleInterface>,
+        type_: std::sync::Arc<type_::Type>,
+    ) -> Option<DefinitionLocation<'_>> {
+        type_constructor_from_modules(importable_modules, type_).map(|t| DefinitionLocation {
+            module: Some(&t.module),
+            span: t.origin,
+        })
+    }
+
     pub fn definition_location(
         &self,
-        type_constructor: Option<&'a type_::TypeConstructor>,
+        importable_modules: &'a im::HashMap<EcoString, type_::ModuleInterface>,
     ) -> Option<DefinitionLocation<'_>> {
         match self {
             Self::Pattern(pattern) => pattern.definition_location(),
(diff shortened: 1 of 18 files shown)


compiler-core/src/language_server/compiler.rs
```diff
@@ -124,6 +130,12 @@ where
             }
             // Create the source information
             let path = module.src_path.to_string();
+            // strip canonicalised windows prefix
+            #[cfg(target_family = "windows")]
```
Acepie, line 134: ran into this while testing on my other computer. Only causes issues with local deps (I'm guessing due to calling canonicalise similar to https://github.com/gleam-lang/gleam/pull/2842). Could pull this out into a separate thing if preferred

compiler-core/src/type_.rs
```diff
@@ -134,6 +134,28 @@ impl Type {
...
+
+    pub fn constructor_types(&self) -> Option<Vec<Arc<Self>>> {
+        match self {
+            Self::Named { args, .. } => Some(args.clone()),
+            Self::Var { type_, .. } => type_.borrow().constructor_types(),
+            _ => None,
+        }
+    }
+
+    pub fn nested_type(&self) -> Option<Arc<Self>> {
```
lpil, line 152: What's a nested type?

Could you document these three functions please 🙏

compiler-core/src/language_server/tests/definition.rs
```diff
@@ -632,3 +632,129 @@ fn main() {
...
+                    character: 0
+                },
+                end: Position {
+                    line: 1,
+                    character: 15
+                }
+            }
+        })
+    )
+}
```
lpil, line 760: Worth having a test for path deps?

lpil:
> Because of how locations/labels for CustomType constructor args are implemented hovering over the arg label will be treated as hovering over the annotation 

Does that mean if I have this code (`][ is cursor`)

```gleam
pub type Thing {
  Thing(value: ][ String)
}
```
then it shows documentation for `type String`?

Acepie:
Yeah because of how the parser sets the location the entire arg is String. We could add some additional information to the AST if we wanna split it but I think the way it seemed to me rn "value" is treated as a label in the AST

Acepie:
Huh actually, I wanted to test this out after I typed that and I guess somewhere along the line that statement became no longer true and I'm not sure why. Wondering if maybe I had a bug in an earlier version or if there is something I am not remembering rn

Acepie:
Ok yeah the actual behavior is that the label has no hover (which matches the previous behavior). I think i had a bug at some middle phase that meant i was checking the wrong location and fixed it when I switched to do the TypeAst method of finding the location

lpil (commented):
Smashing it as always! Thank you!

Left some v small questions

compiler-core/src/type_.rs
Acepie: I'm realizing inner_type is probably a better name for this. I just needed a way to follow the Var to the underlying type so we had proper linking. Lemme rename and also add comments

compiler-core/src/language_server/tests/definition.rs
Acepie: lemme add one since it'll be easy though i think? it'll use all the same code paths

compiler-core/src/language_server/compiler.rs
Acepie: Also, not sure if you want me to pull this out if this pr won't get into the 1.1 release (I'm fine with whatever but lmk what you think. I think the fix here really should be in 1.1 but the rest makes sense to leave out)

Acepie:
Ok figured out how to get the documentation on hover for custom types. Ultimately ended up adding documentation to TypeConstructors and then passing it through and looking up the type on hover

Screenshot 2024-04-09 at 8 24 06 AM

Acepie:
I pulled the windows path fix into https://github.com/gleam-lang/gleam/pull/2941

compiler-core/src/build.rs
```diff
@@ -296,10 +296,14 @@ pub enum Located<'a> {
     ModuleStatement(&'a TypedDefinition),
     FunctionBody(&'a TypedFunction),
     Arg(&'a TypedArg),
+    Annotation(&'a TypeAst, std::sync::Arc<type_::Type>),
```
lpil, line 299: What do you think of just having the location here? It seems that's all we use and I can't think of anything else that we could take from this data structure that would be useful.

compiler-core/src/language_server/engine.rs
```diff
@@ -154,7 +154,12 @@ where
                 None => return Ok(None),
             };
 
-            let location = match node.definition_location() {
+            let location = match node.definition_location(node.annotation_type().and_then(|t| {
+                type_constructor_from_modules(
+                    this.compiler.project_compiler.get_importable_modules(),
+                    t,
+                )
+            })) {
```
lpil, line 162: Could we instead pass in the importable modules? That would make this API simpler, it would remove some of the extra work done in the cases when we don't use the type constructor information to find the location, and it would avoid having to pattern match on the node twice (once in definition_location, once in annotation_type).

compiler-core/src/metadata/tests.rs
```diff
@@ -324,6 +329,7 @@ fn module_with_type_links() {
                     module: "a".into(),
                     parameters: vec![],
                     deprecation: Deprecation::NotDeprecated,
+                    documentation: Some("documentation".into()),
```
lpil, line 332: Could we have a dedicated test for the documentation and also have some tests for the origin too please. Thanks

compiler-core/src/type_.rs
```diff
@@ -134,6 +134,33 @@ impl Type {
...
+    }
+
+    /// Gets the inner type following any var links. Returns a pointer to self
+    /// if the type is not a var.
+    pub fn inner_type(&self) -> Option<Arc<Self>> {
+        match self {
+            Self::Var { type_, .. } => type_.borrow().inner_type(),
+            _ => Some(Arc::new(self.clone())),
+        }
+    }
```
lpil, line 162: I think this is called collapse_type_links elsewhere in the compiler and I think it doesn't clone the type if it is not a link.

compiler-core/src/language_server/engine.rs
```diff
@@ -225,6 +230,8 @@ where
                 }
 
                 Located::Arg(_) => None,
+
+                Located::Annotation(_, _) => Some(this.completion_types(module)),
```
lpil, line 234: Let's have some tests for this new functionality or remove it.

lpil (commented):
Fab! Really like this, will be a huge improvement to the language server.

There's a few tests I would like to have added here (autocompletions for the various places that annotations can happen) and I'd like to go back to how the definition location method worked before. See the note inline.

Thank you so much!!

compiler-core/src/build.rs
Acepie: Yeah I think replacing the TypeAst with just the location is reasonable. The type itself we need for figuring out the type constructor

compiler-core/src/language_server/engine.rs
Acepie: Yup sounds good :D for some reason I thought this was covered by existing tests

compiler-core/src/type_.rs
Acepie: I spent more time than i want to admit trying to figure out how to do the recursion on the inner `Arc<RefCell<TypeVar>>` and it looks like switching to `collapse_links` actually solves that much nicer than the current implementation. Thanks for pointing this out

compiler-core/src/build.rs
lpil: Sorry yes that's what I meant, in place of the AST

lpil (approved):
Beautiful! Thank you so much!!!

PR #2916, opened 2024-04-07
Fix bug with import sorting

### LINKED ISSUE #2915: Formatter: imports are not sorted correctly
@hayleigh-dot-dev tried the formatter on main and there's something wrong with the way imports are sorted:
```diff
 // IMPORTS
-
 import filepath
+
 import gleam/dynamic.{type Dynamic}
 import gleam/io
 import gleam/list
``` 
I don't want the space to be removed before that comment and it also looks like it's adding a spurious space after the first import. I must have messed something up with the way comments are handled. I'll be looking into this!

giacomocavalieri:
This PR closes #2915 
I took this as a chance to rewrite the import sorting bit and make it a bit tidier, I was having a really hard time making sense of what I wrote earlier. I think now it's quite simpler and better commented altogether

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,7 @@
   needlessly indented if preceded by a comment.
 - Line endings other than `\n` are now handled by the formatter, preserving
   blank lines and converting them to `\n`.
+- The formatter can now format groups of imports alphabetically.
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you formatter maestro!!

PR #2920, opened 2024-04-08
Make sure monospaced links have the same color as common links

### LINKED ISSUE #2880: Code doc links are hard to discover
In the generated docs monospaced words that are also links do not get the same pink colour that links usually have:
Screenshot 2024-04-01 alle 22 11 38
They are indistinguishable from normal monospaced words and a pink underline only appears when hovering:
Screenshot 2024-04-01 alle 22 11 42

I think those should be coloured pink like all links and still get an underline on hover, this way one could easily tell monospaced links apart from monospaced text

giacomocavalieri:
This PR closes #2880.

### DIFF
--- a/compiler-core/templates/docs-css/index.css
+++ b/compiler-core/templates/docs-css/index.css
@@ -262,6 +262,10 @@ p code {
   color: var(--inline-code);
 }
 
+p a code {
+    color: var(--links);
+}
+
 /* Page layout */
 
 .page {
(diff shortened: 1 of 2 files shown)


compiler-core/templates/docs-css/index.css
```diff
@@ -262,6 +262,10 @@ p code {
   color: var(--inline-code);
 }
 
+p a code {
+    color: var(--links);
```
lpil, line 266: 2 space indentation please 🙏

giacomocavalieri: Oops, fixed!

Issue #1117, opened 2021-05-21
Formatter: Support elm-format style line-break opt-in


MattCheely:
The gleam formatter is great, and I appreciate the no-configuration approach. Bikeshedding formatting rules is one of my least favorite things to do!

There is one feature from elm-format that I'd love to see added though, which is that if the developer has manually inserted a line break into a data structure or function call, then the formatter will switch to a multi-line format for that data/function.

Some examples:
```rust
// Input
send_msg_resp(conn,    500,    "error")

// Output
send_msg_resp(conn, 500, "error")
```
```rust
// Input
send_msg_resp(conn,
                500, "error")

// Output
send_msg_resp(
  conn,
  500,
  "error",
)
```
```rust
// Input
let err_msg = ["Job timed out after ",     int.to_string(job_timeout),    "ms"]

// Output
let err_msg = ["Job timed out after ", int.to_string(job_timeout), "ms"]
```
```rust
// Input
let err_msg = ["Job timed out after ", int.to_string(job_timeout),
                 "ms"]

// Output
let err_msg = [
  "Job timed out after ",
  int.to_string(job_timeout),
  "ms",
]
```

lpil:
I like this suggestion a lot! Let's do this one.￼

PR #2929, opened 2024-04-08
One-line pipes

### LINKED ISSUE #1622: Formatter: Permit pipe operators to render on a single line
https://github.com/gleam-lang/gleam/discussions/1619#discussioncomment-2708716

Also related to https://github.com/gleam-lang/gleam/issues/1117 as I think it would be nice to be able to force it to render them on multiples lines.

giacomocavalieri:
This PR closes #1622 

Before going further with the implementation and adding more tests I'd love to get some feedback on the implementation, specifically on the data structure choice. The big picture idea is:
- Add a new `NewLine` token
- Keep track of all new lines like we do with empty lines, comments, etc.
- When we're in a pipe we want to see if there's any new line between its start and end
  - If that's the case then the programmer wants the newline to be split on multiple lines
  - Otherwise, we try and fit it on a single line and split it only if it goes beyond the 80 chars limit

On the data structure choice:
- I went with a `BTreeSet` because it looks like to be the best to check if a value (the newline) exists between a range (pipe start and end positions). If I used an array like empty lines, every time we run into a pipe we would have to walk through the array to find it
- I'm also considering using an array (just like empty lines, comments etc) and implementing a kind of binary search on that one, taking advantage of the fact that the positions would already be sorted in increasing order

What do you think would be best @lpil? I quite like how simple the implementation is in this PR thanks to the `BTreeSet`

Having this kind of data structure and keeping track of newlines would also allow us to extend this behaviour to other stuff in the future (function arguments, list items, etc. like it was suggested in #1117)

---

EDIT: In the end we decided to go for the array of newlines and implement a binary search!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -127,6 +127,31 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Pipes can now be placed on a single line if they are short enough:
+
+  ```gleam
+  [1, 2, 3] |> list.map(int.to_string) |> string.join(with: "\n")
+  ```
+
+  In addition you can also force the formatter to break a pipe on multiple lines
+  by manually breaking it. This:
+
+  ```gleam
+  [1, 2, 3]
+  // By putting a newline here I'm telling the formatter to split the pipeline
+  |> list.map(int.to_string) |> string.join(with: "\n")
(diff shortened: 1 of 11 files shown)


lpil (approved):
This is so nice!!!!!!

PR #2923, opened 2024-04-08
Make sure comments are not moved outside of empty lists

giacomocavalieri:
This PR closes #2910

PR #2933, opened 2024-04-09
Fix formatting of empty bit arrays

### LINKED ISSUE #2932: Gleam format produces invalid code
```diff
diff --git a/test/cbor_gl_test.gleam b/test/cbor_gl_test.gleam
index fc094f4..8645a9d 100644
--- a/test/cbor_gl_test.gleam
+++ b/test/cbor_gl_test.gleam
@@ -99,7 +99,9 @@ pub fn decode_byte_string_test() {
   >>)
   |> should.be_ok()
   |> should.equal(
-    #(<<78:int-size(64), 78:int-size(64), 78:int-size(64), 78:int-size(64)>>, <<>>),
+    #(<<78:int-size(64), 78:int-size(64), 78:int-size(64), 78:int-size(64)>>, <<
+      ,
+    >>),
   )
 }
```

In the case of long lines with an empty `BitString`, `gleam format` tries to format `<<>>` as:
```

<<
,
>>
```
which is syntactically invalid. I've not found a workaround for this issue, as it appears `<<>>` is the only way to represent an empty `BitString` and `gleam format` is very opinionated.

Relevant repo/line [here](https://github.com/chiefnoah/cbor_gl/blob/master/test/cbor_gl_test.gleam#L98). See the failing CI pipeline for additional info.

PgBiel:
1. Made it so empty bit arrays are not broken by the formatter into `<<\n,\n>>`, but kept as `<<>>`, to avoid generating invalid syntax. Fixes #2932
2. Extended https://github.com/gleam-lang/gleam/pull/2923 to also apply to bit arrays, such that comments inside empty bit arrays are not moved outside them. This required a small refactor as we needed a mutable reference to `self` where we previously had no `self`.
3. Added a missing comment in a test I added in a previous PR. I think the reference to the appropriate issue, as is done for other tests, should be there as it is helpful for future readers.

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -401,7 +401,7 @@ impl<'comments> Formatter<'comments> {
                 let segment_docs = segments
                     .iter()
                     .map(|s| bit_array_segment(s, |e| self.const_expr(e)))
-                    .collect::<Vec<_>>();
+                    .collect_vec();
 
                 self.bit_array(
                     segment_docs,
@@ -848,7 +848,7 @@ impl<'comments> Formatter<'comments> {
                 let segment_docs = segments
                     .iter()
                     .map(|s| bit_array_segment(s, |e| self.bit_array_segment_expr(e)))
-                    .collect::<Vec<_>>();
+                    .collect_vec();
 
                 self.bit_array(
                     segment_docs,
@@ -1789,7 +1789,7 @@ impl<'comments> Formatter<'comments> {
                 let segment_docs = segments
                     .iter()
                     .map(|s| bit_array_segment(s, |e| self.pattern(e)))
-                    .collect::<Vec<_>>();
+                    .collect_vec();
 
                 self.bit_array(segment_docs, false, location)
             }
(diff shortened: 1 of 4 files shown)


compiler-core/src/format.rs
```diff
@@ -834,12 +842,20 @@ impl<'comments> Formatter<'comments> {
...
+            UntypedExpr::BitArray {
+                segments, location, ..
+            } => {
+                let segment_docs = segments
                     .iter()
-                    .map(|s| bit_array_segment(s, |e| self.bit_array_segment_expr(e))),
-                segments.iter().all(|s| s.value.is_simple_constant()),
-            ),
+                    .map(|s| bit_array_segment(s, |e| self.bit_array_segment_expr(e)))
+                    .collect::<Vec<_>>();
```
giacomocavalieri, line 851: This one and the others below could be `collect_vec()` I think

giacomocavalieri (approved):
Awesome! I've left just one small comment

compiler-core/src/format.rs
PgBiel: Done :+1:

chiefnoah:
Thanks @PgBiel for this!

lpil (approved):
Amazing! Thank you!

PR #2934, opened 2024-04-09
Oh dear debugger... fixed on JS

### LINKED ISSUE #2931: `debugger` variable name should be overridden in JS
When using `debugger` as a variable name, which is totally fine in gleam, it returns an error in JS, because `debugger` is a reserved keyword.

Example code:

```gleam
msg.Restart(debugger) -> {
  let restart_effect =
    model.debuggers
    |> debugger.get(debugger)
    |> result.then(fn(d) {
      d.steps
      |> list.first()
      |> result.map(fn(item) { d.dispatcher(item.model) })
    })
    |> result.unwrap(effect.none())
```

JS code

```javascript
let debugger = msg[0];
let restart_effect = (() => {
  let _pipe = model.debuggers;
  let _pipe$1 = $debugger.get(_pipe, debugger);
  let _pipe$2 = $result.then$(
    _pipe$1,
    (d) => {
      let _pipe$2 = d.steps;
      let _pipe$3 = $list.first(_pipe$2);
      return $result.map(
        _pipe$3,
        (item) => { return d.dispatcher(item.model); },
      );
    },
  );
  return $result.unwrap(_pipe$2, $effect.none());
})();
```

In the JS code, in strict mode, `debugger` is reserved. `debugger` should be named `debugger$` or something like that.

PgBiel:
Fixes #2931

- Variables named `deoh dear` are no longer escaped (they are awesome and deserve peace)
- Variables named `debugger` are now escaped
-

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Compiler
+
+- Fixed a bug on the JavaScript target where variables named `debugger`, which
+  is a JavaScript keyword, were not being renamed, leading to runtime errors.
+
 ## v1.1.0-rc1 - 2024-04-08
 
 ### Compiler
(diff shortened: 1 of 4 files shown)


lpil (approved):
ahahaha! deoh dear

Issue #2938, opened 2024-04-09
New "private members in public API" check fails for re-export


rawhat:
Hello!

Just testing the v1.1 RC against some of my repos.  In `mist` I am re-exporting some types from an internal module in my public main module.  I added some documentation comments to them just so it's not quite so opaque.

Unfortunately, the new check marks those as a warning.  Example:

```gleam
/// Re-exported type that represents the default `Request` body type. See
/// `mist.read_body` to convert this type into a `BitString`. The `Connection`
/// also holds some additional information about the request. Currently, the
/// only useful field is `client_ip` which is a `Result` with a tuple of
/// integers representing the IPv4 address.
pub type Connection =
  InternalConnection
```

The errors come from public functions that actually _use_ this type.  I've been trying to avoid the whole "make a new type to export and convert between them" where possible.  I suspect this would/will be an issue if we ever get constructor re-exports as well.

lpil:
Thank you. I've reopened the original issue for this and we've reverted this implementation.

PR #2939, opened 2024-04-09
Revert internal warning and build tool complaining

giacomocavalieri:
These pesky gleamlins not putting everything in a single file smh

Jokes aside, we want to revert these changes as it would break both Lustre and Mist (see #2938) and That's Not Good.
These could be added back once we have a better story for reexports of private/internal types but it looks like something for another Gleam version!


lpil (commented):
Can we actually remove the code please 🙏 

I don't like to have commented out code as it gets out of sync with the real code, and we're gunna have to approach this in a whole different way in future with making aliases part of the type system.

lpil:
I will do it :)

giacomocavalieri:
Thanks!! Taking notes for the future

PR #2943, opened 2024-04-09
Warn for redundant assertions

giacomocavalieri:
I think this would be a neat addition:
- people could forget to leave an `assert` when that's not actually needed
- it would help newcomers avoid falling into the common trap of thinking that patterns can check for variable equality:
  ```gleam
  let wibble = 1
  // ... other stuff that maybe is using wibble, so we don't get an
  // unused variable warning for immediately shadowing it.
  let assert wibble = 2
  // I've noticed newcomers sometime think that this would end up checking 2 == 1
  // With a warning saying the assert is redundant it would be harder to make
  // this mistake!
  ```


lpil (approved):
Beautiful, thank you!

PR #2946, opened 2024-04-09
Fix confusing error message for some reserved keywords

### LINKED ISSUE #2945: Confusing error message when using reserved keyword
When I define a function that is called like a keyword the error message could be improved:
```gleam
pub fn echo() { todo }
//     ^^^^ I was expecting a name here
```
It could instead say that `echo` is a keyword and not a valid name.

I'm working on a fix :)

giacomocavalieri:
This PR closes #2945


lpil (approved):
Fab!! Could you rebase on main please 🙏

lpil (approved):
Thank you

PR #2962, opened 2024-04-11
Module type inference supports returning multiple errors and pipes multiple errors to outputs properly

### LINKED ISSUE #2951: Make it possible for analysis to return multiple errors for a single module
Return a data structure of typed code, warnings, and errors. If there are errors then the build is a failure and codegen is expected not to happen.

This ticket does not need to include making all the analysis code actually return multiple errors, this is just a first step to get the mechanisms in place.

Acepie:
Closes #2951

Used leaking of private types as an example to show the multiple errors but we can remove that bit if we'd rather handle it all at once. The core of the change is just changing the return type on infer_modules and changing the error type for type errors to have a list of errors instead of a single error. The majority of the lines changed are formatting in the errors since we now have an error that produces multiple diagnostics

Screenshot 2024-04-11 at 6 55 51 AM
Screenshot 2024-04-11 at 6 56 18 AM

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -111,31 +111,22 @@ impl TargetSupport {
 }
 
 #[derive(Debug)]
-pub struct InferenceResult {
+pub struct InferenceFailure {
     // Right now this is an option because we don't complete the module typing on error
     // Once we add proper type holes and always return a module this should not be an option
     pub ast: Option<TypedModule>,
     // All the inference errors that occurred during type inference
-    pub errors: Vec<Error>,
+    pub errors: Vec1<Error>,
 }
 
-impl InferenceResult {
-    // Used to create InferenceResults from a singe error.
+impl From<Error> for InferenceFailure {
+    // Used to create InferenceFailure from a singe error.
     // Should not be needed once we start actually collecting all the errors.
-    pub fn from_error(error: Error) -> Self {
-        Self {
+    fn from(error: Error) -> Self {
+        InferenceFailure {
             ast: None,
-            errors: vec![error],
+            errors: vec1::vec1![error],
         }
-    }
-
-    // Used by tests to get the underlying ast.
-    // Panics if there are type errors.
-    pub fn expect(self, message: &str) -> TypedModule {
-        if !self.errors.is_empty() {
-            panic!("Inference contains errors: {}", message);
-        }
-        self.ast.expect(message)
     }
 }
 
@@ -156,7 +147,7 @@ pub fn infer_module<A>(
     line_numbers: LineNumbers,
     package_config: &PackageConfig,
     src_path: Utf8PathBuf,
-) -> InferenceResult {
+) -> Result<TypedModule, InferenceFailure> {
     let mut errors = Vec::new();
     let name = module.name.clone();
     let documentation = std::mem::take(&mut module.documentation);
@@ -170,9 +161,7 @@ pub fn infer_module<A>(
         warnings,
         target_support,
     );
-    if let Err(e) = validate_module_name(&name) {
-        return InferenceResult::from_error(e);
-    }
+    validate_module_name(&name)?;
 
     let mut type_names = HashMap::with_capacity(module.definitions.len());
     let mut value_names = HashMap::with_capacity(module.definitions.len());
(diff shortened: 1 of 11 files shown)


compiler-core/src/analyse.rs
```diff
@@ -199,22 +251,35 @@ pub fn infer_module<A>(
             direct_dependencies,
             warnings,
             &env,
-        )?;
-        typed_statements.push(statement);
+        );
+        if let Err(e) = statement {
+            return InferenceResult::from_error(e);
+        }
+        typed_statements.push(statement.expect("statement should exist"));
```
lpil, line 258: Here too

compiler-core/src/analyse.rs
```diff
@@ -199,22 +251,35 @@ pub fn infer_module<A>(
...
     }
     for t in statements.custom_types {
-        let statement = infer_custom_type(t, &mut env)?;
-        typed_statements.push(statement);
+        let statement = infer_custom_type(t, &mut env);
+        if let Err(e) = statement {
+            return InferenceResult::from_error(e);
+        }
+        typed_statements.push(statement.expect("statement should exist"));
```
lpil, line 265: Here too

compiler-core/src/analyse.rs
```diff
@@ -199,22 +251,35 @@ pub fn infer_module<A>(
...
 
     // Sort functions and constants into dependency order for inference. Definitions that do
     // not depend on other definitions are inferred first, then ones that depend
     // on those, etc.
-    let definition_groups = into_dependency_order(statements.functions, statements.constants)?;
+    let definition_groups = into_dependency_order(statements.functions, statements.constants);
+    if let Err(e) = definition_groups {
+        return InferenceResult::from_error(e);
+    }
+    let definition_groups = definition_groups.expect("definition_groups should exist");
```
lpil, line 282: Here too

compiler-core/src/analyse.rs
```diff
@@ -224,12 +289,18 @@ pub fn infer_module<A>(
         for definition in group {
             match definition {
                 CallGraphNode::Function(f) => {
-                    let statement = infer_function(f, &mut env, &mut hydrators, &name)?;
-                    working_group.push(statement);
+                    let statement = infer_function(f, &mut env, &mut hydrators, &name);
+                    if let Err(e) = statement {
+                        return InferenceResult::from_error(e);
+                    }
+                    working_group.push(statement.expect("statement should exist"));
```
lpil, line 296: Here too

compiler-core/src/analyse.rs
```diff
@@ -224,12 +289,18 @@ pub fn infer_module<A>(
...
                 }
                 CallGraphNode::ModuleConstant(c) => {
-                    let statement = infer_module_constant(c, &mut env, &name)?;
-                    working_group.push(statement);
+                    let statement = infer_module_constant(c, &mut env, &name);
+                    if let Err(e) = statement {
+                        return InferenceResult::from_error(e);
+                    }
+                    working_group.push(statement.expect("statement should exist"));
```
lpil, line 303: Here too

compiler-core/src/build/package_compiler.rs
```diff
@@ -435,12 +435,19 @@ fn analyse(
...
+        // so that instead of returning immediately we bubble up and
+        // do the error check prior to code generation.
+        if (!errors.is_empty()) {
+            return Err(Error::Type {
+                path: path.clone(),
+                src: code.clone(),
+                error: errors,
+            });
+        }
+        let ast = ast.expect("Module should have an AST when no errors");
```
lpil, line 450: Here too

compiler-core/src/type_/tests.rs
```diff
@@ -425,7 +425,13 @@ pub fn compile_module_with_opts(
         LineNumbers::new(src),
         &config,
         "".into(),
-    )
+    );
+
+    if errors.is_empty() {
+        Ok(ast.expect("module should have been inferred"))
```
lpil, line 431: Here too

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__leak_multiple_private_types.snap
```diff
@@ -0,0 +1,37 @@
...
+   │
+10 │         pub fn ret_private2() -> Private {
+   │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The following type is private, but is being used by this public export.
+
+    Private
+
+Private types can only be used within the module that defines them.
+error: Private type used in public interface
```
lpil, line 27: Let's put an empty line between errors to make it easier to see where they start and end

compiler-core/src/error.rs
```diff
@@ -2450,300 +2001,762 @@ value, or use `let assert` rather than `let`."
...
+                    hint: None,
+                    level: Level::Error,
+                    location: None,
+                }
+            }
+
+            Error::Type { path, src, error } => self.type_to_diagnostic(
+                path,
+                src,
+                error.first().expect("there should be at least one error"),
```
lpil, line 2757: Let's remove this runtime check and instead use the type system to enforce that it is correct. We could use a Vec1 here and have this function return Vec1<Diagnostic>. This would also simplify it as we won't have the two paths through this code with the first special case for Error::Type

compiler-core/src/analyse.rs
```diff
@@ -152,42 +184,62 @@ pub fn infer_module<A>(
     // Register any modules, types, and values being imported
     // We process imports first so that anything imported can be referenced
     // anywhere in the module.
-    let mut env = imports::Importer::run(origin, env, &statements.imports)?;
+    let env = imports::Importer::run(origin, env, &statements.imports);
+    if let Err(e) = env {
+        return InferenceResult::from_error(e);
+    }
+    let mut env = env.expect("env should exist");
```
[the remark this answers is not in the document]
Acepie: Sounds good :D thank you. I had a feeling this wasn't the right way to do this but blanked on it at the time for some reason

compiler-core/src/error.rs
Acepie: Switching the error within Error::Type to a vec1 makes sense to me. The main reason I had `to_diagnostic` as a separate thing was to avoid wrapping all the non type cases in vec! (mainly to make the initial refactoring lift easier).

compiler-core/src/error.rs
```diff
@@ -58,9 +60,12 @@ pub enum Error {
     Type {
         path: Utf8PathBuf,
         src: EcoString,
-        error: crate::type_::Error,
+        error: Vec1<crate::type_::Error>,
     },
 
+    #[error("type checking module {module} produced no results")]
+    TypeNotGenerated { module: Name },
```
Acepie, line 67: I added this type in case we have a compiler bug that causes errors to be empty but no ast to be generated (instead of panicing like before). lmk if that seems reasonable

compiler-core/src/error.rs
Acepie: Ok between doing this and switching the formatting/structuring to be back to the original the errors are now much closer to the original

Acepie:
Huh looks like there was some unrelated vuln that was released that is causing the check to fail https://rustsec.org/advisories/RUSTSEC-2024-0336

Acepie:
Ok sorry about all the changes but realized the more I thought about what you said that we may be better off switching back to a result but just making a new `InferenceFailure` struct to contain the errors and ast. Once I figured out making the error bubble up properly it actually cleaned up the inference itself to be almost like before. LMK if I'm totally off base and should switch back though

compiler-core/src/analyse.rs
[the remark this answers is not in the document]
lpil: Wicked!

lpil (approved):
Great stuff! Thank you!!!

PR #2980, opened 2024-04-12
Fix generated `mod` property in erlang application

rawhat:
The generated output was just `{mod, 'my_module'}`.  But I'm pretty sure
this needs to include the arguments as well?  This PR updates it to
produce: `{mod, {'my_module', []}}`.  In my local testing, this worked
as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,10 @@
 
 ### Compiler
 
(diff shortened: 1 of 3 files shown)


nerdyworm:
Just wanted to confirm that this change fixes the issue.

lpil (approved):
Thank you!!

Issue #2899, opened 2024-04-05
Requirement overriding


pendletong:
I think it would be a useful addition to the language to have a mechanism by which requirements which conflict can be ignored.

I'm not sure the best way to go about this. Node has the concept of overrides (https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) which strikes me as a possible solution.

I think in my case I just wanted the build to override a single include - gleescript required filepath >=0.1-<1.0 but filepath 1.0 was needed for simplifile >=1.6.

lpil:
Good idea!

darky:
Similar problem
https://github.com/MystPi/pprint/issues/4

darky:
`npm` resolves this problem by installing multiple versions of same packages, which suitable for every parent package, if it can't be flattened by semver of course.
Via that behaviour ☝️  many jokes, that `node_modules` weights more, than black hole, but it's just works!

darky:
Using of `overrides` also is choice, but if something will broken, it's your responsibility :)

lpil:
We will not be duplicating packages, it's an awful design in my opinion. It causes build sizes and compile times to balloon, creates a lot of extra work for the programmers verifying and updating their deps, it would increase the complexity of the compiler, and it removes this pressure to make folks update their deps and to keep their deps tree shallow.

darky:
Totally agree about duplicating packages, but situation, when programmer can't install some package, because some dependency semver conflict exists in tree, is not fine totally. Moreover, with growing ecosystem, this problem can occur very often. 

Need to do something, `overrides` section, for example. But it may broke compilation, may not broke, depends of your lucky. 

Also we can try to ask community not broke semver and backward compatibiltiy in their packages, but I think it's utopia. Nevertheless, Clojure has very stable ecosystem and packages backward compatibility almost never broken. 

Another suggestions?..

lpil:
We are not duplicating dependencies.

lpil:
It wasn't an issue for Elixir, Elm or PureScript, ecosystems I was familiar with during their early stages, I'm sure it won't be a problem for us.

PgBiel:
A related thought (I apologize if this is off-topic): it'd be nice if we could override dependencies in arbitrary ways: not only pin a package to a different version (which overrides the version used for all dependencies), but also being able to pin a package to a local path, or to another Hex package, or (in the future) to an arbitrary Git repository/commit.

[...]

PR #2981, opened 2024-04-13
(POC) Transitive dependency overriding

PgBiel:
Proof of concept for #2899. Basically implements most of the hard parts; this PR should be easy to adjust based on the result of the discussion there.

## Current design

Adds a new `[patch]` section to `gleam.toml`, which works exactly like `[dependencies]`, except that any package in the root project's `[patch]` will override any transitive dependencies to that package. For example, consider this `gleam.toml` in the `test1` project:

```toml
[dependencies]
gleam_stdlib = ">= 0.34.0 and < 2.0.0"
test2 = { path = "../test2" }

[dev-dependencies]

[patch]
test2 = { path = "../test3" }
simplifile = "1.6.0"
filepath = { path = "../filepath" }
```

If we assume that `test3` depends on `simplifile` 1.7.0, what will happen here is that:

1. `test1` will depend on `test3` instead of `test2` (for now assuming the package at `test3` is also named `test2`).
2. `test2` will depend on simplifile 1.6.0 instead of 1.7.0, even if incompatible (the error is silenced).
3. `simplifile` 1.6.0 (depended on by test2) will now depend on filepath at `../filepath` instead of the Hex version it usually depends on (1.0.0).

This `[patch]` section is added to the `manifest.toml` to ensure the manifest updates when the `[patch]` changes.

## Current implementation

1. On `compiler-core`, most changes occur in `dependency`. There, we have a check for version incompatibilities; it is simply skipped when the package is patched by the root config. Additionally, when working with pubgrub's traits for dependency resolving, we ensure we filter packages' versions according to the root config patch, and also update the fetched packages' dependencies accordingly so pubgrub doesn't get confused. We also, of course, add the `patch` section to `PackageConfig`, and also require specifying a `patch` to access a config's dependencies.
2. On `compiler-cli`, most changes are in `dependencies`. Here, we had to refactor providing local and Git packages into a `PackageProvider` struct containing `provide_*_package` as methods instead of top-level functions. This was necessary as the `PackageFetcher` now stores the PackageProvider in a RefCell, such that, if a Hex package depends on a package which was patched to point to a local or Git package instead of a Hex package, the PackageProvider is invoked to provide that package instead of fetching from Hex (or get it from cache if the package was already provided before - hence the need for RefCell, so we can update the cache, which is also later reused to add those local packages to the manifest). Additionally, in the `provide_*_package` methods, we update each package's dependencies according to the root config's patch during traversal of dependencies. Finally, in the appropriate location within `compiler-cli`, we also ensure the manifest is updated when the patch section changes.

## TODO

I will wait for further discussion before progressing on this PR, so here's more or less what still needs to be done.

- [ ] Decide on the final design (via discussion at the relevant issue)
- [ ] Add support for overriding a package with another package, even if it has a different name (could be separate PR)
- [ ] Add unit tests
- [ ] Update CHANGELOG

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -1,3 +1,4 @@
+use std::cell::RefCell;
 use std::{
     collections::{HashMap, HashSet},
     time::Instant,
@@ -510,7 +511,7 @@ fn get_manifest<Telem: Telemetry>(
     // to date so we can return it unmodified.
     if is_same_requirements(
         &manifest.requirements,
-        &config.all_dependencies()?,
+        &config.all_dependencies(&config.patch)?,
         paths.root(),
     )? {
         tracing::debug!("manifest_up_to_date");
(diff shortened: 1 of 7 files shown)


lpil:

Issue #2986, opened 2024-04-14
Confusing bugged error message from dependency resolver


SaphiraKai:
A confusing and buggy-looking error message can be reproduced on at least Gleam `1.0.0` and `1.1.0-rc3` by creating the following files;

`gleam.toml`:
```toml
name = "buggy_error"
version = "1.0.0"

[dependencies]
gleam_stdlib = "~> 0.34 or ~> 1.0"
lustre = "~> 3.1"
```

`manifest.toml`:
```toml
packages = [
  { name = "argv", version = "1.0.2", build_tools = ["gleam"], requirements = [], otp_app = "argv", source = "hex", outer_checksum = "BA1FF0929525DEBA1CE67256E5ADF77A7CDDFE729E3E3F57A5BDCAA031DED09D" },
  { name = "gleam_community_ansi", version = "1.4.0", build_tools = ["gleam"], requirements = ["gleam_community_colour", "gleam_stdlib"], otp_app = "gleam_community_ansi", source = "hex", outer_checksum = "FE79E08BF97009729259B6357EC058315B6FBB916FAD1C2FF9355115FEB0D3A4" },
  { name = "gleam_community_colour", version = "1.4.0", build_tools = ["gleam"], requirements = ["gleam_json", "gleam_stdlib"], otp_app = "gleam_community_colour", source = "hex", outer_checksum = "795964217EBEDB3DA656F5EB8F67D7AD22872EB95182042D3E7AFEF32D3FD2FE" },
  { name = "gleam_json", version = "1.0.0", build_tools = ["gleam"], requirements = ["gleam_stdlib", "thoas"], otp_app = "gleam_json", source = "hex", outer_checksum = "8B197DD5D578EA6AC2C0D4BDC634C71A5BCA8E7DB5F47091C263ECB411A60DF3" },
  { name = "gleam_stdlib", version = "0.36.0", build_tools = ["gleam"], requirements = [], otp_app = "gleam_stdlib", source = "hex", outer_checksum = "C0D14D807FEC6F8A08A7C9EF8DFDE6AE5C10E40E21325B2B29365965D82EB3D4" },
  { name = "glint", version = "0.14.0", build_tools = ["gleam"], requirements = ["gleam_community_ansi", "gleam_community_colour", "gleam_stdlib", "snag"], otp_app = "glint", source = "hex", outer_checksum = "21AB16D5A50D4EF34DF935915FDBEE06B2DAEDEE3FCC8584C6E635A866566B38" },
[...]

PR #3001, opened 2024-04-17
Feat: add annotation LSP code action

raquentin:
Closes #2953. Added LSP code action to add explicit type annotations to assignments.

Closes #2954. Added LSP code action to add explicit type annotations to module functions. 

This implementation is inevitably opinionated, I tried to match precedence where it exists but with only 1 other code action there wasn't much. Any changes lmk

Changes:
- added new code action handler `code_action_add_annotation`
- added `code_action_add_annotation` call in LSP engine
- added function `name_if_self_is_var` to `TypedPattern` in `ast.rs` for use in code action suggestion text
- added function `find_nodes_in_range` to `Module` in `build.rs`
- re-organized LSP code action test file a bit to scale better with more code actions
- added LSP test for both annotation types, tested extensively via lsp

Implementation Details:
- this code action is suggested when the start of the selection/cursor is within a line containing an assignment/func
- assignment var name/func name (if applicable) and type is shown in the suggestion:
image
- the abstraction of two helper funcs out of the code action for use in future code actions (primarily), see changes

PR #3002, opened 2024-04-17
simplify error message for version constraint failure

zahash:
fixes https://github.com/gleam-lang/gleam/issues/2986

the new error message looks like this ("hellogleam" is the current project name)

```sh

An error occurred while determining what dependency packages and
versions should be downloaded.

Unable to find compatible versions for the version constraints in your
gleam.toml. The conflicting packages are:

- hellogleam
- lustre_dev_tools
- glint
```

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -12,9 +12,12 @@ use crate::{
     type_::{pretty::Printer, UnifyErrorSituation},
 };
 use ecow::EcoString;
-use hexpm::version::pubgrub_report::{DefaultStringReporter, Reporter};
 use hexpm::version::ResolutionError;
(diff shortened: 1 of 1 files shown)


zahash:
rustfmt seems to be failing because of a completely different file that has nothing to do with this change

raquentin:
> rustfmt seems to be failing because of a completely different file that has nothing to do with this change

Had this same issue on my pr, it can be fixed easily at merge but if you want to go ahead and fix it in the pr look at https://github.com/gleam-lang/gleam/pull/3001/commits/b17e1b7f53742ecddf70de153d098960c2b4eab7

lpil (approved):
Thank you!!

PR #3005, opened 2024-04-18
Fix variable escaping in constants on JS

### LINKED ISSUE #3004: Variables in constants not being escaped on JS
The code below

```gleam
import gleam/io

pub const class = 5
pub const oh_dear = class

pub fn main() {
  io.println("Hello from mongus!")
}
```

compiles to

```js

export function main() {
  return $io.println("Hello from mongus!");
}

export const class$ = 5;

export const oh_dear = class;
```

which errors at runtime with

```
export const oh_dear = class;
                            ^

SyntaxError: Unexpected token ';'
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:167:18)
    at callTranslator (node:internal/modules/esm/loader:285:14)
    at ModuleLoader.moduleProvider (node:internal/modules/esm/loader:291:30)

Node.js v20.11.1
```

due to a missing call to `maybe_escape_identifier_doc` at

https://github.com/gleam-lang/gleam/blob/99c50dde27f6146a8dbbd6963a74a7dd4fa70ee3/compiler-core/src/javascript/expression.rs#L1289

PgBiel:
Fixes #3004

1. Fixes escaping of variable names in constant expressions
2. Fixes escaping of accessed module members in constant expressions
3. Fixes escaping of variable names in constant expressions in case clause guards
4. Added an extra test related to imports for good measure

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
 - Empty case expressions are no longer parse errors and will instead be
   exhaustiveness errors. ([Race Williams](https://github.com/raquentin))
 
+- Fixed variables in constant expressions not being escaped correctly when
+  exporting to JavaScript. ([PgBiel](https://github.com/PgBiel))
+
 ### Formatter
 
 - Redundant alias names for imported modules are now removed.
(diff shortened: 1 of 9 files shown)


lpil (approved):
Amazing!! Thank you

PR #3007, opened 2024-04-18
Add annotation printer

omprakaash:
Adds an annotation printer class that is to be used for inlay hints/code annotations in the language server. This is similar to the pretty_printer, but also context aware. 
 Ref: #2525


compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    // Mapping of module alias to module name
+    module_aliases: &'a HashMap<EcoString, EcoString>,
+    type_parameters: &'a HashMap<u64, TypeAst>,
+    current_module: EcoString,
+}
+
+impl<'a> AnnotationPrinter<'a> {
+    pub fn new(
+        type_aliases: &'a HashMap<(EcoString, EcoString), EcoString>,
```
omprakaash, line 34: Would it be better if we make the constructor just take in the module as an argument ?  We could then construct the necessary context in the constructor.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+    printed_types: HashMap<EcoString, EcoString>,
+
+    // Mapping of module.type to type alias -
+    // eg: import mod1.{type Cat as c} -> Key: (mod1, Cat), Value: c
+    // eg: type UserId = Int -> Key: (mod, UserId), Value: Int
+    type_aliases: &'a HashMap<(EcoString, EcoString), EcoString>,
+
+    // Mapping of module alias to module name
+    module_aliases: &'a HashMap<EcoString, EcoString>,
+    type_parameters: &'a HashMap<u64, TypeAst>,
```
lpil, line 28: These are always strings as the values, right? It looks like below when we pattern match on them we insist that they are.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+                    }
+                    typ_str.push_str(typ_alias.as_str());
+                } else if module == PRELUDE_MODULE_NAME || *module == self.current_module {
+                    if self.name_clashes_if_unqualified(name, module) {
+                        typ_str.push_str(module);
+                        typ_str.push('.');
+                    }
+                    typ_str.push_str(name.as_str());
+                } else {
+                    // Always qualify types from other modules
```
lpil, line 72: If the type has been imported in an unqualified fashion then we want to use the unqualified constructor, not always qualify it.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    fn name_clashes_if_unqualified(&mut self, type_: &EcoString, module: &str) -> bool {
+        match self.printed_types.get(type_) {
+            None => false,
+            Some(previous_module) if module == previous_module => false,
+            Some(_different_module) => true,
+        }
+    }
+
+    fn args_to_string(&mut self, args: &[Arc<Type>]) -> String {
```
lpil, line 124: Rather than build an extra string all methods should push to the existing string buffer

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+                typ_str.push_str("#(");
+                typ_str.push_str(&self.args_to_string(elems));
+                typ_str.push(')');
+
+                typ_str
+            }
+        }
+    }
+
+    fn name_clashes_if_unqualified(&mut self, type_: &EcoString, module: &str) -> bool {
```
lpil, line 116: What does this function mean? I'm not sure what it is doing looking at the previously printed types.

I think it's here to work out if a prelude type has been shadowed by a type defined in this module, but you can't work that out by looking at what has been printed. Rather, we need to know the names of all the types that are in scope in this module, both defined and imported.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+            uid: u64::default(),
+            printed_types: HashMap::new(),
+            type_aliases,
+            module_aliases,
+            type_parameters,
+            current_module,
+        }
+    }
+
+    pub fn print(&mut self, typ: &Type) -> String {
```
lpil, line 50: This method constructs a great many strings- instead let's have it push strs onto the mutable buffer so it only ever allocates one and then grows that as needed.

I'd do this by having a public method that returns a string, and internally it resets an buffer and then calls a private printing function which recursively calls itself and each call will append to the buffer.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    let mut printer =
+        AnnotationPrinter::new(&type_aliases, &module_aliases, &type_parameters, "".into());
+
+    let typ = Type::Var {
+        type_: Arc::new(std::cell::RefCell::new(TypeVar::Generic { id: 0 })),
+    };
+
+    assert_eq!(printer.print(&typ), "T");
+}
```
lpil, line 218: Nice! Let's have tests for all the other possible paths through the type printer. We want full coverage to make sure that we are generating correct code.

compiler-core/src/language_server/annotation_printer.rs
lpil: Could be good, however I think that would make it a lot harder to write the tests as there would be a great deal more information that would be needed to construct this class.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+            printed_types: HashMap::new(),
+            type_aliases,
+            module_aliases,
+            type_parameters,
+            current_module,
+        }
+    }
+
+    pub fn print(&mut self, typ: &Type) -> String {
+        let mut typ_str = String::new();
```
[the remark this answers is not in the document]
omprakaash: Oh yes, will change the buffer to an EcoString.

compiler-core/src/language_server/annotation_printer.rs
omprakaash: Oh Yes. Will change this. Might have to keep track of all unqualified imports in the module then.

compiler-core/src/language_server/annotation_printer.rs
omprakaash: Gotcha. Will add more tests.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+pub struct AnnotationPrinter<'a> {
+    names: HashMap<u64, EcoString>,
+    uid: u64,
+
+    // A mapping of printd type names to the module that they are defined in.
+    printed_types: HashMap<EcoString, EcoString>,
+
+    // Mapping of module.type to type alias -
+    // eg: import mod1.{type Cat as c} -> Key: (mod1, Cat), Value: c
+    // eg: type UserId = Int -> Key: (mod, UserId), Value: Int
```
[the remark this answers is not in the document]
omprakaash: Oh the field `type_aliases` contains mapping of pairs of types and the module in which they are defined to their alias in the current module. (Sorry the second example is actually flipped since the value has to be UserId).  

So for example, lets say there is an import statement in a module like so `import mod1.{type Tiger as Tig}`. This would result in a map entry with the key being ("mod1", "Tiger") and the value being Tig. -> `type_aliases.insert(("mod1", "Tiger"), "Tig")`

Another example is that of an type alias statement like so in a module named foo: `type UserId = Int`. Since `Int` has been aliased by `UserId`,  an entry in the map is inserted, with they key being a pair of (foo, Int) and the value being UserId. (`type_aliases.insert(("foo", "Int"), "UserId"`). So any occurrences of the above types in a module which are not already annotated by the user, could be annotated with their aliased types.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+    type_aliases: &'a HashMap<(EcoString, EcoString), EcoString>,
+
+    // Mapping of module alias to module name
+    module_aliases: &'a HashMap<EcoString, EcoString>,
+    type_parameters: &'a HashMap<u64, TypeAst>,
+    current_module: EcoString,
```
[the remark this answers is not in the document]
omprakaash: Sry, type_parameters is probably not a clear name for this. What this does it keeps track of user annotations for generic variables in a function. 
eg 1: 
```gleam
fn foo(arg1: gen, arg2){
    arg1 == arg2
}
```
 -> Here the annotation that will be generated for arg2 will be `gen`

eg2: 
```gleam
fn identity(x) {
    let z = x
    let y: value = x
    y
}
```

The type annotations generated for x and z will be 'value'.

compiler-core/src/language_server/annotation_printer.rs
omprakaash: Gotcha Thank you !

compiler-core/src/language_server/annotation_printer.rs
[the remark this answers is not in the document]
lpil: I think that type alias one is going to break for generics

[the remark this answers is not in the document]
omprakaash: Hmm sorry, not quite sure on how this would break generics. Could you give me an example ?

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,218 @@
...
+
+    // A mapping of printd type names to the module that they are defined in.
+    printed_types: HashMap<EcoString, EcoString>,
+
+    // Mapping of module.type to type alias -
+    // eg: import mod1.{type Cat as c} -> Key: (mod1, Cat), Value: c
+    // eg: type UserId = Int -> Key: (mod, UserId), Value: Int
+    type_aliases: &'a HashMap<(EcoString, EcoString), EcoString>,
+
+    // Mapping of module alias to module name
```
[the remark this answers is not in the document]
omprakaash: This is just used to keep track of module aliases. 
eg: `import testmod as test`
So the key would be "testmod" and the vale "test"

omprakaash:
Hey,  I changed the logic for annotation generation based on the earlier review and added some comments as well. I am still a bit unclear on how the current use of `type_aliases` breaks generics (https://github.com/gleam-lang/gleam/pull/3007#discussion_r1575008618). Will add more test cases as soon that is fixed.

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,297 @@
...
+            uid: u64::default(),
+            type_aliases,
+            module_aliases,
+            generic_annotations,
+            unqualified_imports,
+            unqualified_type_names,
+        }
+    }
+
+    pub fn print_type(&mut self, typ: &Type) -> String {
```
lpil, line 63: We use EcoString, not String. If conversion is needed to interact with some library we should do it at the boundary of the system, not internally

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -0,0 +1,297 @@
...
+#[test]
+fn test_type_parameter() {
+    let type_aliases = HashMap::new();
+    let module_aliases = HashMap::new();
+    let unqualified_imports = HashMap::new();
+    let unqualified_type_names = HashSet::new();
+    let mut generic_annotations = HashMap::new();
+
+    let type_var = TypeAst::Var(crate::ast::TypeAstVar {
+        name: EcoString::from("T"),
```
lpil, line 212: This isn't valid Gleam, it should be lowercase for a variable

compiler-core/src/language_server/annotation_printer.rs
lpil: Are you working on these? 🙏

compiler-core/src/language_server/annotation_printer.rs
```diff
@@ -295,3 +297,213 @@ fn test_prelude_type_shadowed() {
...
+
+    let typ = Type::Named {
+        name: EcoString::from("Cat"),
+        args: vec![],
+        module: EcoString::from("mod1"),
+        publicity: crate::ast::Publicity::Public,
+        package: EcoString::from(""),
+    };
+
+    assert_eq!(printer.print_type(&typ), "Animals.Cat");
```
lpil, line 467: This is not valid Gleam. What happened here?

lpil (commented):
Nice! Once we have all the tests we can merge this I think

raquentin:
Is this stable? I can complete #3001 if so

lpil:
Thank you!! Gunna use this as a dev branch while v1.2.0 is worked through.

PR #3022, opened 2024-04-20
fix panic when unused import is first character of file

Acepie:
Currently on main the lsp crashes if there is an unused import as the first character of the file because `start - 1` would be negative and `start` is a u32


Acepie:
Didn't put a changelog since its a bug in a fix that's only in main

lpil (approved):
Thank you!

PR #3028, opened 2024-04-22
Improve error message if importing type using the value import syntax or vice versa

### LINKED ISSUE #2576: Improve error message if importing type using the value import syntax
Give a hint to say that you probably meant the type as the value does not exist

For example, if you have a module named `wibble`, that defines a type named `One` but no value called `One` and you write this code

```gleam
import wibble.{One}
```

the resulting error message should suggest `import wibble.{type One}`

Pi-Cla:
Resolves #2576
Now they look like this:
```
error: Unknown module field
  ┌─ /src/one/two.gleam:1:19
  │
1 │ import gleam/foo.{Bar}
  │                   ^^^ Did you mean `type Bar`?
`Bar` is only a type, it cannot be imported as a value.
```

```
error: Unknown module type
  ┌─ /src/one/two.gleam:1:19
  │
1 │ import gleam/foo.{type Baz}
  │                   ^^^^^^^^ Did you mean `Baz`?
`Baz` is only a value, it cannot be imported as a type.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,28 @@
 
 - Initial support for type analysis returning multiple errors. ([Ameen Radwan](https://github.com/Acepie))
 
+- Improve error message if importing type using the value import syntax or vice versa. ([Pi-Cla](https://github.com/Pi-Cla/))
+
+```
+error: Unknown module field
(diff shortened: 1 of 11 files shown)


lpil (approved):
Beautiful! Thank you

PR #3034, opened 2024-04-23
[Help Wanted] Ensure _ pattern marks all rows beneath it as redundant

### LINKED ISSUE #2651: Unreachable cases not always detected for ints, floats, and strings
Hello, I'm walking through the gleam.run tutorial and I've found what appears to be a bug.
The code below does not give an error or warning about unreachable code.
Perhaps this is specific to the wasm runtime, and not present in the OTP version?

```
import gleam/io
// import gleam/int

pub fn main() {
  let x = 3 // int.random(4)
  io.debug(x)

  let result = case x {
    // Match specific values
    0 -> "Zero"
    1 -> "One"
    _ -> "Other"
    // Match any other value
    3 -> "hi"
    4 -> "no"
  }
  io.debug(result)
}
```

Reordering the cases slightly does tell me there is unreachable code:

```
import gleam/io
import gleam/int

pub fn main() {
  let x = 3 // int.random(4)
  io.debug(x)

  let result = case x {
    // Match specific values
    _ -> "Other"
    0 -> "Zero"
    1 -> "One"
    // Match any other value
    3 -> "hi"
    4 -> "no"
  }
  io.debug(result)
}
```

Pi-Cla:
We did not have thorough enough tests as pointed out in #2651 so none of our existing tests placed the `_` pattern anywhere but on the first or last row. Hence me repurposing an if statement that only checked the first row into a for loop checking every row

Also to get the fancy Github thing saying this PR will close an issue:
Resolves #2651 woohoo

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -299,18 +299,23 @@ impl<'a> Compiler<'a> {
             .map(|row| self.move_unconditional_patterns(row))
             .collect::<Vec<_>>();
 
-        // There may be multiple rows, but if the first one has no patterns
-        // those extra rows are redundant, as a row without columns/patterns
-        // always matches.
-        let first_row_always_matches = rows.first().map(|c| c.columns.is_empty()).unwrap_or(false);
-        if first_row_always_matches {
-            let row = rows.remove(0);
-            self.diagnostics.reachable.push(row.body.clause_index);
+        // There may be multiple rows, but if any row has the `_` pattern or no column/pattern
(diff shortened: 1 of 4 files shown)


Acepie:
I don't think the approach you have here will work in the general case. It seems like the root cause of the issue is that for infinite cases we need to make sure that fallback rows and regular rows are processed in the exact same order as they appear in the case but https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/exhaustiveness.rs#L453-L455 means that fallback rows will always be processed last. I think I have an alternative solution to this by switching up the way we append fallback rows to the rest of the rows

PR #3037, opened 2024-04-24
Only add Hex link to HTML docs when publishing to Hex

### LINKED ISSUE #3020: Only add Hex link to HTML docs when publishing to Hex
For whatever reason, not all packages may be in Hex.pm. Currently my package Gloss cannot be there as that name is squatted, but there may be other reasons, for example using a service like <https://codecodeship.com> that has its own Hex compatible repo.

The docs builder `gleam docs build` forcefully adds a link to Hex even if the package is not there: <https://github.com/gleam-lang/gleam/blob/203801cd5c3a7d4498205a221248231210eddcc4/compiler-core/src/docs.rs#L75>. It could be (rarely, but still) useful to be able to opt out of this link. For me, this causes my docs to link to the squatted package.

Pi-Cla:
Fix #3020

### DIFF
--- a/compiler-cli/src/docs.rs
+++ b/compiler-cli/src/docs.rs
@@ -7,6 +7,7 @@ use gleam_core::{
     analyse::TargetSupport,
     build::{Codegen, Mode, Options, Package},
     config::{DocsPage, PackageConfig},
+    docs::DocContext,
     error::Error,
     hex,
     io::HttpClient as _,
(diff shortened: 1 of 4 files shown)


lpil (commented):
Lovely! Thank you.

Would you mind using an enum rather than a bool? Seeing `true` without any context is less clear than something like `DocsContext::Hex`

PR #3038, opened 2024-04-24
Make sure comments are not moved outside of a function expression

### LINKED ISSUE #3027: Formatter puts comment into wrong block
When formatting the following code:
```gleam
import ct/player
import ct/std
import ct/triggers
import ct/world
import gleam/list
import gleam/string
import gleam/bool.{guard}

pub fn main() {
  let my_name = player.get_name(player.me())

  triggers.register_on_tick(fn() {
    let players_near_me =
      world.get_players()
      |> list.filter(fn(player) {
        player.get_name(player) != my_name
        // && player.distance_to(player, player.me()) <. 5.0
      })

    use <- guard(
      when: players_near_me
        |> list.length
        <= 0,
      return: Nil,
    )

    std.log(
      "players: "
      <> {
        players_near_me
        |> list.map(player.get_name)
        |> string.join(", ")
      },
    )
  })
}

```
The formatter gives back:

```gleam
import ct/player
import ct/std
import ct/triggers
import ct/world
import gleam/list
import gleam/string
import gleam/bool.{guard}

pub fn main() {
  let my_name = player.get_name(player.me())

  triggers.register_on_tick(fn() {
    let players_near_me =
      world.get_players()
      |> list.filter(fn(player) { player.get_name(player) != my_name })

    // && player.distance_to(player, player.me()) <. 5.0
    use <- guard(
      when: players_near_me
        |> list.length
        <= 0,
      return: Nil,
    )

    std.log(
      "players: "
      <> {
        players_near_me
        |> list.map(player.get_name)
        |> string.join(", ")
      },
    )
  })
}
```

The comment is now no longer in it's containing block as it should be.

giacomocavalieri:
This PR closes #3027

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,18 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Comments at the end of an anonymous function are no longer moved out of it.
+  You can now write this:
+
+  ```gleam
+  fn() {
+    todo
+    // A comment here!
+  }
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Language Server
 
 - The code action to remove unused imports now removes the entire line is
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you

PR #3039, opened 2024-04-24
fix off by 1 in completion range from lsp

Acepie:
Import completions were not appearing in Zed. It turns out there was an off by 1 bug in the end range that the lsp was emitting for the completion. Other editors just clip that range to the end of the line but zed https://github.com/zed-industries/zed/blob/e1791b7dd0ff1948d3d302674fda2f659974c0b3/crates/project/src/lsp_command.rs#L1513 throws an exception if the end is past the line and doesn't show the completion. Fixing the off by 1 bug causes the completions to appear again

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,3 +134,6 @@
 
 - Fixed a typo when attempting to publish a package with non-Hex dependencies
   ([inoas](https://github.com/inoas))
+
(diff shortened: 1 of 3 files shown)


Acepie:
We had tests and everything I just can't count apparently 😭

lpil (approved):
These positioning tests are so fiddly

PR #3046, opened 2024-04-26
fix issue with reachability of infinite cases

### LINKED ISSUE #2651: Unreachable cases not always detected for ints, floats, and strings

Acepie:
Closes #2651

Thanks to @Pi-Cla for getting this started and writing up test cases for this. The change ultimately changes infinite cases such that instead of adding all fallback cases at the end we instead add the fallback cases in the original clause order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,4 +194,4 @@
 
 - Fixed a bug where unreachable infinite cases would not be detected when
   after a discard or variable pattern.
-  ([Ameen Radwan](https://github.com/Acepie))
+  ([Ameen Radwan](https://github.com/Acepie)) and ([Pi-Cla](https://github.com/Pi-Cla))


lpil (approved):
Thank you so much  @Acepie and @Pi-Cla !!

PR #3047, opened 2024-04-26
Add Never type to Prelude (WIP on warning for expressions after a function with return type Never)

Pi-Cla:
As discussed in #3044 it would be neat to add a type to the Prelude which could represent functions which never return.

The `Never` type can not be constructed because if it could be constructed then someone could write a function which does return at some point with it.
For example we don't want this to be possible:
```gleam
fn go_forever() -> Never {
    // oh no it actually does stop!
    Never
}
```

WIP on figuring out where to add in the Warning for expressions coming after a function/expression of type `Never`.

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -1971,6 +1971,7 @@ impl<'a> TypePrinter<'a> {
 
     fn print_prelude_type(&self, name: &str, args: &[Arc<Type>]) -> Document<'static> {
         match name {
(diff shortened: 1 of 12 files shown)


lpil (commented):
Hello! Thank you but it cannot be implemented as a named type as it has different semantics to them. It needs to be a super type of all other types, otherwise panic would be too restrictive and not usable in case expressions.

lpil:

Issue #3033, opened 2024-04-23
Improve error message when there's something that's not a function to the left of `use` 


giacomocavalieri:
Consider this snippet of code:
```gleam
pub fn main() {
  use <- 1
  todo
}
```
The resulting error looks like this:
```txt
error: Type mismatch
  ┌─ /src/main.gleam:2:10
  │
2 │   use <- 1
  │          ^

This value is being called as a function but its type is:

    Int
```
But someone who doesn't know how use works might be left wondering _why_ that's an error, I think a more helpful error message could look something like this:
```txt
error: Type mismatch
  ┌─ /src/main.gleam:2:10
  │
2 │   use <- 1
  │          ^

On the right hand side of the `<-` operator should go a function
that takes a callback function as its last argument.
This value has type:

    Int

See: https://tour.gleam.run/advanced-features/use/
```

lpil:
The problem here is that `use` is sugar and the analyser doesn't know anything about it has already been erased. We should promote it to being something more than sugar, then we can have much better error messages.

giacomocavalieri:
Ok! I'd really like to work on this, if I finish my backlog of other issues I need to work on before anyone does this, I'll give it a shot!

Issue #3032, opened 2024-04-23
Improve error message for `use` with function with incorrect arity


giacomocavalieri:
Consider this snippet of code:
```gleam
import gleam/result

pub fn main() {
  use a <- result.try(todo, todo)
  todo
}
```
The error message from the compiler looks like this:
```txt
error: Incorrect arity
  ┌─ /src/main.gleam:4:3
  │  
4 │ ╭   use a <- result.try(todo, todo)
5 │ │   todo
  │ ╰──────^ Expected 2 arguments, got 3

This call accepts these additional labelled arguments:

  - apply
```
Which I think could be quite confusing, especially since it highlight the whole block below use. I think it could be more helpful if it highlighted just the function name and explained you shouldn't supply its last argument:
```txt
error: Incorrect arity
  ┌─ /src/main.gleam:4:3
  │  
4 │     use a <- result.try(todo, todo)
  │              ^^^^^^^^^^

Something something

See: https://tour.gleam.run/advanced-features/use/
```
I'm not sure how to word the error message though 😅

lpil:
This is one place where I feel that two arrows might be useful, one pointing to the arguments, one pointing to `result.try`.  The library we use does support that I believe, but our diagnostic type only supports one.

PR #3048, opened 2024-04-27
Use improvements

### LINKED ISSUE #3031: Improve error message for `use` functions with wrong last argument
Consider the following snippet of code: 
```gleam
pub fn main() {
  use a <- io.println
  1
}
```
The error produced by the compiler looks like this:
```txt
error: Type mismatch
  ┌─ /src/main.gleam:5:3
  │
5 │   1
  │   ^

Expected type:

    String

Found type:

    fn(a) -> Int
```
I think it would be way more useful if the error was something like this:
```txt
error: Type mismatch
  ┌─ /src/main.gleam
  │
4 │   use a <- io.println
  │            ^^^^^^^^^^

This function cannot be used on the right hand side of a `use`.
I was expecting its last argument to be a callback function with type:

  fn(a) -> String

But its actual type is:

  String
  
See: https://tour.gleam.run/advanced-features/use/
```
However, if the last argument is actually a callback function the error message could be worded differently, just highlighting the type mismatch:
```text
error: Type mismatch
  ┌─ /src/main.gleam
  │
4 │   use a, b <- result.try(result)
  │               ^^^^^^^^^^

This function accepts a callback function with a different type. 
Expected type:

  fn(Int) -> Result(a, String)

Found type:

  fn(a, b) -> Int
```

giacomocavalieri:
This PR improves on various use error messages.
Closes #3031, #3032 and #3033

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2730,7 +2730,8 @@ Rename or remove one of them.",
                 TypeError::NotFnInUse { location, typ } => {
                     let mut printer = Printer::new();
                     let text = wrap_format!(
-                        "In a use expression, there should be a function on the right hand side of `<-`, but this value has type:
+                        "In a use expression, there should be a function on the right hand side \
+of `<-`, but this value has type:
 
 {}
 
@@ -2791,16 +2792,19 @@ See: https://tour.gleam.run/advanced-features/use/");
                         _ => format!("{given} arguments"),
                     };
                     let label = format!("Expected {expected_string}, got {given}");
-                    let mut text: String = format!("The function on the right of `<-` here takes {expected_string}.\n");
+                    let mut text: String = format!("The function on the right of `<-` \
+here takes {expected_string}.\n");
 
                     if expected > given {
                         if supplied_arguments == 0 {
                             text.push_str("The only argument that was supplied is the `use` callback function.\n")
                         } else {
-                            text.push_str(&format!("You supplied {supplied_arguments_string} and the final one is the `use` callback function.\n"));
+                            text.push_str(&format!("You supplied {supplied_arguments_string} \
+and the final one is the `use` callback function.\n"));
                         }
                     } else {
-                        text.push_str("All the arguments have already been supplied, so it cannot take the the `use` callback function as a final argument.\n")
+                        text.push_str("All the arguments have already been supplied, \
+so it cannot take the the `use` callback function as a final argument.\n")
                     };
 
                     text.push_str("\nSee: https://tour.gleam.run/advanced-features/use/");
@@ -2824,7 +2828,9 @@ See: https://tour.gleam.run/advanced-features/use/");
 
                 TypeError::UseFnDoesntTakeCallback { location, actual_type: Some(actual) } => {
                     let mut printer = Printer::new();
-                    let text = wrap_format!("The function on the right hand side of `<-` has to take a callback function as its last argument. But the last argument of this function has type:
+                    let text = wrap_format!("The function on the right hand side of `<-` \
+has to take a callback function as its last argument. \
+But the last argument of this function has type:
 
 {}
 
@@ -2861,7 +2867,8 @@ See: https://tour.gleam.run/advanced-features/use/",
                         _ => format!("{given} were provided"),
                     };
 
-                    let text = wrap_format!("This function takes a callback that expects {expected}. But {specified} on the left hand side of `<-`.
+                    let text = wrap_format!("This function takes a callback that expects {expected}. \
+But {specified} on the left hand side of `<-`.
 
 See: https://tour.gleam.run/advanced-features/use/");
                     Diagnostic {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1638,7 +1638,7 @@ But function expects:
                 } => {
                     let mut printer = Printer::new();
                     printer.with_names(annotated_names.clone());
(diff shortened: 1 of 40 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -2344,20 +2398,42 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
                     location,
                     implicit,
                 } = arg;
-                let value = self.infer_call_argument(value, typ.clone())?;
+
+                // If we're typing a `use` call then the last argument is the
+                // use callback and we want to treat it differently to report
+                // better errors.
+                let is_last_argument = i == args_count - 1;
+                let argument_kind = match (kind, is_last_argument) {
```
lpil, line 2421: Can we match on just `kind` and use `is_last_argument` in a guard please

compiler-core/src/type_/expression.rs
```diff
@@ -528,13 +568,22 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            call.arguments,
+            call_location,
+            CallKind::Use {
+                assignments_location,
+                last_statement_location,
+            },
+        )?;
 
-        Ok(Statement::Expression(call))
+        Ok(Use {
```
lpil, line 581: This has radically changed the meaning and data of `Use`- now the callback code is within the call in the `Use`. Can we document this somewhere? Or is there some way to make it clearer?

compiler-core/src/type_/expression.rs
```diff
@@ -2410,7 +2487,19 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                assignments_location,
+                last_statement_location,
+            } => e.into_use_unify_error(
+                function_location,
+                assignments_location,
+                last_statement_location,
+                value.location(),
+            ),
+            ArgumentKind::Regular => convert_unify_error(e, value.location()),
+        })?;
```
lpil, line 2502: Could you extract this into a method please? I'm finding it a bit hard to see the logic as the majority of the code is type conversion boilerplate.

compiler-core/src/type_/expression.rs
```diff
@@ -2705,10 +2794,10 @@ struct UseAssignments {
 }
 
 impl UseAssignments {
-    fn from_use_expression(sugar_assignments: Vec<UseAssignment>) -> UseAssignments {
+    fn from_use_expression(sugar_assignments: &[UseAssignment]) -> UseAssignments {
```
lpil, line 2797: There's a bunch of cloning here which we didn't have to do before. Why is that?

lpil (commented):
Wonderful work! I am very impressed!

I've left some comments in-line, mostly minor code readability things. I've also left one question about some cloning which I couldn't quite figure out.

compiler-core/src/error.rs
```diff
@@ -2725,7 +2725,166 @@ Rename or remove one of them.",
...
+                            path: path.clone(),
+                            src: src.clone(),
+                            extra_labels: vec![],
+                        }),
+                    }
+                },
+
+                TypeError::UseFnDoesntTakeCallback { location, actual_type: Some(actual) } => {
+                    let mut printer = Printer::new();
+                    let text = wrap_format!("The function on the right hand side of `<-` has to take a callback function as its last argument. But the last argument of this function has type:
```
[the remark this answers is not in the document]
giacomocavalieri: Aha! I was wondering why it stopped working all of a sudden 😆

compiler-core/src/type_/environment.rs
```diff
@@ -726,19 +726,49 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
...
+                    .map_err(|_| UnifyError::CouldNotUnify {
+                        expected: t1.clone(),
+                        given: t2.clone(),
+                        situation: Some(UnifyErrorSituation::FunctionsMismatch {
+                            reason: FunctionsMismatchReason::Results {
+                                one: retrn1.clone(),
+                                other: retrn2.clone(),
+                            },
+                        }),
+                    })
```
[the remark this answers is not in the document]
giacomocavalieri: Oops!

compiler-core/src/type_/expression.rs
```diff
@@ -505,6 +537,14 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         }
 
         let statements = Vec1::try_from_vec(statements).expect("safe: todo added above");
+        // We need this to report good error messages in case there's a type error
+        // in use. We consider `use` to be the last statement of a block since
+        // it consumes everything that comes below it and returns a single value.
+        let last_statement_location = statements
+            .iter()
+            .find_or_last(|e| e.is_use())
```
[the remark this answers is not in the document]
giacomocavalieri: The cool part is I discovered completely by accident 😆

compiler-core/src/error.rs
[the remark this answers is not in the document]
lpil: It took me forever to work that one out. It's really hard to find the lines that cause the problem when it happens too

lpil:
I think you're working on this so I've made it a draft so I know not to look yet in my notifications list

compiler-core/src/type_/expression.rs
giacomocavalieri: Ok, I'm adding some comments!

lpil (approved):
Incredible work as per usual! This is going to make such a huge difference to how easy it is to debug `use`

PR #3049, opened 2024-04-27
Fix 3023

### LINKED ISSUE #3023: Warn for unreachable code after todo and panic
Right now any argument given is simply ignored, leading to a bit confusing behavior.
While it may be syntactically possible to give an argument to the two keywords, they will mostly be ignored right now.

The tour teaches that the correct way to give a message is using the `as` keyword, but that is rather late in the tour and easily forgotten later, so it would be nice to give a hint and remind people of that gotcha.

giacomocavalieri:
This PR closes #3023 

This is a work in progress for now, I still have to get to the unreachable code warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,29 @@
 
   ([Pi-Cla](https://github.com/Pi-Cla/))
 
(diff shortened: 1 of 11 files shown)


lpil (approved):
Beautiful!! Thank you!

PR #3051, opened 2024-04-27
Added goto definition support for imports

Acepie:
For qualified import statements it just goes to the top of the file. For unqualified import statements it goes to the definition of the actual type/value. This also sets up unqualified imports as a separate `Located` in the lsp which should help with making implementing hover relatively easy (didn't wanna put it in this pr but will probably have one up for hover after this)

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -2176,8 +2176,9 @@ where
                         as_name: None,
                     };
                     if self.maybe_one(&Token::As).is_some() {
-                        let (_, as_name, _) = self.expect_name()?;
+                        let (_, as_name, end) = self.expect_name()?;
                         import.as_name = Some(as_name);
+                        import.location.end = end;
                     }
                     imports.values.push(import)
                 }
@@ -2191,8 +2192,9 @@ where
                         as_name: None,
                     };
                     if self.maybe_one(&Token::As).is_some() {
-                        let (_, as_name, _) = self.expect_upname()?;
+                        let (_, as_name, end) = self.expect_upname()?;
                         import.as_name = Some(as_name);
+                        import.location.end = end;
                     }
                     imports.values.push(import)
                 }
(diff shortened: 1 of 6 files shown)


compiler-core/src/parse.rs
```diff
@@ -2207,8 +2209,9 @@ where
                         as_name: None,
                     };
                     if self.maybe_one(&Token::As).is_some() {
-                        let (_, as_name, _) = self.expect_upname()?;
+                        let (_, as_name, end) = self.expect_upname()?;
                         import.as_name = Some(as_name);
+                        import.location.end = end;
```
lpil, line 2214: Won't this change how the errors for duplicate imports etc are shown? I would like to keep them as only the first part of the import and not `.{}` etc

lpil (commented):
Thank you, this is really nice! I've left one question inline.

compiler-core/src/parse.rs
Acepie: I made this change so that the `as name` would be included as part of the import so that it would also have the goto work properly. this would change the highlight to include the `as name` part of the import but not the `.{}`. That being said I'm fine with flipping this back and just not having the goto work on the rename part of the import

lpil: Sorry, I'm not following. `as name` comes after `.{}` so wouldn't it be included?

Let's add a test for an invalid import or something where the import is over multiple lines, that'll make it clear what is happening.

Acepie: This change was for unqualified `as name` for an individual type/value like `import example_module.{my_num as num}`. Lemme double check if we have a test for errors with the wider `as name` and the unqualified `as name` and add them if they are missing

Acepie: Ok I added the tests though looking at the output more, I wonder if making this change would be better than the existing behavior for unqualified imports? Cause the actual issue is technically not that the type is imported twice but that the same name is used with the as. Lmk what you think

Screenshot 2024-04-30 at 11 12 17 AM
Screenshot 2024-04-30 at 11 12 26 AM

lpil: Ah yes! Sorry I got confused there.

I agree, the error could be improved here

PR #3077, opened 2024-05-01
fix a bug where publishing to hex would fail if the description in gl…

inoas:
…eam.toml had an utf8 symbol


inoas:
Ref https://github.com/hexpm/hexpm/issues/1250

compiler-cli/src/publish.rs
```diff
@@ -600,7 +600,7 @@ fn release_metadata_as_erlang() {
     let meta = ReleaseMetadata {
         name: "myapp",
         version: &version,
-        description: "description goes here",
+        description: "description goes here 🌈",
```
lpil, line 603: Love it!

lpil (approved):
Thank you so much!

PR #3078, opened 2024-05-01
LSP hover on unqualified support

Acepie:
Hovering over an unqualified import for an type shows the same hover as using the type as an annotation. Hovering over an unqualified import for a value has the same behavior as hovering over the value as an expression. This is true for both the unqualified name as well as the as name


lpil (approved):
Super nice work, thank you!

PR #3079, opened 2024-05-01
Add gleam hex revert

### LINKED ISSUE #3019: gleam publish --revert
mix supports that, gleam could as well?

Pi-Cla:
A new package can be reverted or updated within 24 hours of it's initial publish, a new version of an existing package can be reverted or updated within one hour.

Resolves #3019

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -365,6 +365,20 @@ enum Hex {
     /// - HEXPM_PASS: (optional) The Hex password to authenticate with.
     #[command(verbatim_doc_comment)]
     Unretire { package: String, version: String },
+
+    /// Revert a release from Hex
+    ///
+    /// This command uses this environment variables:
+    ///
+    /// - HEXPM_USER: (optional) The Hex username to authenticate with.
+    /// - HEXPM_PASS: (optional) The Hex password to authenticate with.
+    Revert {
+        #[arg(long)]
+        package: Option<String>,
+
+        #[arg(long)]
+        version: Option<String>,
+    },
 }
 
 #[derive(Subcommand, Debug)]
@@ -469,6 +483,8 @@ fn main() {
         Command::Hex(Hex::Unretire { package, version }) => {
             hex::UnretireCommand::new(package, version).run()
         }
+
+        Command::Hex(Hex::Revert { package, version }) => hex::revertcommand(package, version),
 
         Command::Add { packages, dev } => add::command(packages, dev),
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,11 @@
 
   Hint: Rename this module and try again.
   ```
+- New subcommand `gleam hex revert` added. ([Pi-Cla](https://github.com/Pi-Cla))
+  * You can specify the options like this: `gleam hex revert --package gling --version 1.2.3`
+  * A new package can be reverted or updated within 24 hours of it's initial publish,
+  a new version of an existing package can be reverted or updated within one hour.
+  * You could already update packages even before this release by running: `gleam publish` again.
 
 ### Compiler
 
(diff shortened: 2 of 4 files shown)


inoas:
> I ended up having to make revert a subcommand because if I tried to make it like:
> 
> ```
> gleam publish --revert [VERSION]
> ```
> 
> where --revert and [VERSION] is optional then I end up with the type `Option<Option<String>>` which Clippy does not like.
> 
> This would then require that whenever any part of the code was processing the full double option to add a line to silence clippy else clippy would complain.

Thanks for doing this!
Note: The changelog is now not in sync.

compiler-wasm/Cargo.toml
```diff
@@ -14,6 +14,8 @@ gleam-core = { path = "../compiler-core" }
 console_error_panic_hook = "0"
 serde-wasm-bindgen = "0"
 wasm-bindgen = { version = "0", features = ["serde-serialize"] }
+# See https://docs.rs/getrandom/latest/getrandom/index.html#webassembly-support
+getrandom = { version = "0", features = ["js"] }
```
lpil, line 18: What's this for? No changes have been made to the wasm crate.

compiler-cli/src/main.rs
```diff
@@ -271,6 +273,15 @@ pub enum ExportTarget {
     },
 }
 
+#[derive(Subcommand, Debug)]
+enum PublishRevertCommand {
+    /// Reverts a published release within 24 hours of initial publish or within one hour for an already existing package
```
lpil, line 278: This is too long to be a CLI help message, it needs to fit in around 60 columns

lpil (commented):
Thank you!

Let's make this `gleam hex revert` to match `gleam hex retire`.

compiler-wasm/Cargo.toml
Pi-Cla: This dependency is only there to enable the `js` feature on `getrandom` which is a transitive dependency for a few parts of gleam. Without the `js` feature, getrandom would complain whenever its being compiled for `wasm32-unknown-unknown` which now does happen because I bumped the dependencies in hexpm-rust.

I will make an additional comment briefly explaining this and move this to the shared workspace

compiler-cli/src/main.rs
Pi-Cla: Hmm... then I guess I will have to trim out the information about when a package can be reverted. If instead added a link to the hex docs which talk about the time limit then the link itself would also be too long.

Oh well, I guess the message will just have to say: "Reverts a published release"

compiler-wasm/Cargo.toml
```diff
@@ -22,6 +22,7 @@ itertools.workspace = true
 serde.workspace = true
 termcolor.workspace = true
 tracing.workspace = true
+getrandom.workspace = true
```
lpil, line 25: How come this has been added? This package is unmodified and doesn't use random numbers afaik

Pi-Cla: This is from the required update to the latest version of hexpm-rust. After the dependencies in hexpm-rust were bumped there was a new indirect dependency added `getrandom` which required the `js` flag to be added in order for builds on the `wasm32-unknown-unknown` target to work. I added comments to the root Cargo.toml to explain this:
https://github.com/gleam-lang/gleam/blob/218ebb06827c983631dea8ebf6c0ef1abcd3b8a1/Cargo.toml#L61-L62

Pi-Cla: Also this diff is inaccurate now because the PR that updates the CI got merged which also required this same hexpm-rust bump. So this PR now has no additional effect on this stuff. I have rebased this PR to give an updated diff

lpil (approved):
Wonderful! Thank you!!

PR #3083, opened 2024-05-02
fix: failures in CI after 1.78 Rust release

### LINKED ISSUE #3082: `Some checks were not successful` in CI
This could be a result of Rust's recent `1.78.0` release.

hamirmahal:
fixes #3082.


compiler-core/src/ast_folder.rs
hamirmahal, line 1: An alternative to removing the unused `trait`s could be prepending them with an underscore.

lpil:
Sorry @hamirmahal but these have been fixed by one of the older PRs.

@Pi-Cla what warnings do you see? I don't get any.

Issue #3085, opened 2024-05-03
asdf local gleam version - move compile gleam binary into a bin subdir


inoas:
`asdf gleam uninstall nightly && asdf gleam install nightly` failed today supposedly because the CI on the nightly was broken.

So I checked if I can use `asdf` to specify a local version, and I can.

in `.tool-versions`:

```
gleam path:../gleam-src/target/debug
```

This means I have my gleam project next to the gleam compiler which in this case is in gleam-src.

However, asdf looks for `PATH/bin/gleam`. I have added a dir and symbolic link into target/debug for now and it works. However to make it seamsless to use asdf with locally compiled versions of gleam it would be nice compiled gleam binaries could be put into a bin subdirectory so for instance into `/target/debug/bin/`. Would that be possible?

lpil:
Not possible I'm afraid. asdf needs to adapt to Gleam, not the other way around.

inoas:
> Not possible I'm afraid. asdf needs to adapt to Gleam, not the other way around.

I will bring it up with asdf then.

Is it possible to commit a symbolic link to git? That would solve it for now?

inoas:
If the submission of the symbolic link works, maybe the PR is fine?

PR #3086, opened 2024-05-03
add symbolic link so that using asdf with local gleam binaries works

inoas:
fixes https://github.com/gleam-lang/gleam/issues/3085


inoas:
OK, I read "impossible" but this here means "wontfix" which is fine, but different. If it had been wontfix I'd not attempted this PR. Sorry about it.

lpil:
Sorry I was unclear @inoas !

PR #3092, opened 2024-05-03
Add more warnings for pattern matching on literal values

### LINKED ISSUE #3068: Do not complain about match on literal empty tuple
Right now the compiler provides a wrong suggestion when one tries to pattern match on a literal empty tuple:
```gleam
case #() {
//   ^^^ Redundant tuple wrapper, you can remove this
  _ -> todo
}
```
But following the compiler's advice would lead to a syntax error, despite this pattern matching being useless I think we do not want this warning on empty tuples

giacomocavalieri:
This PR closes #3068 and adds a couple more warning like I suggested in that issue


inoas:
You are a machine!

compiler-core/src/type_/error.rs
```diff
@@ -384,6 +384,14 @@ pub enum EmptyListCheckKind {
     NonEmpty,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LiteralCollectionKind {
+    List,
+    Tuple,
+    BitArray,
```
lpil, line 391: There's a use case for bit arrays! You could be matching on segments that don't align with the segments you used for construction.

giacomocavalieri: You're right, I've removed the warning when matching on non empty literal bit arrays 👍

lpil (approved):
Marvellous! Thank you!!

PR #3095, opened 2024-05-04
Continue analysis on errors for constants

### LINKED ISSUE #3090: Do not exit analysis when a constant is invalid
Register an error and continue analysis of the rest of the module.

If the constant expression is invalid but it has a valid type annotation then we can register the constant into the environment using the type annotation to get the type.

If the constant expression is valid but the type annotation is invalid we can register the constant into the environment using the type inferred from the expression.

If both the constant expression and the type annotation are valid but disagree we can register the constant into the environment using the type annotation.

Split from https://github.com/gleam-lang/gleam/issues/2766

# Questions

If a constant definition is invalid and it had no type annotation then it could produce a large number of error messages from the rest of the module as any place that references this type will get a "undefined value" error. Should we in keep track of there being a failed constant with this name and suppress unknown value error messages relating to it? I think it would be a nice experience to only see the root error.

Acepie:
Closes #3090

First pass at continuing analysis when there are errors in constant definitions and literals.The bulk of the line changes are passing the error vector down throughout all the inference recursion. The core logic of the change is

- pass error vector throughout inference so that errors can be collected as we go
- A new `InvalidConstant` type of constant has been added to act as a hole when inference fails for a constant
  - This new type will only ever be valid during inference and will never appear in UntypedAsts or during code gen
- `infer_module_constant` and `infer_const` now always return Constants and append their errors to error vector
- `infer_function` now returns with its own error as well as the errors in the error vector
  - this is done so constant literal related errors are not lost when there is a function error
  - this is temporary until `infer_function` also no longer returns a result
  - once that is done the normal error flow can be restored
- break `infer_const` into 2 separate methods, a helper that does the inference for the value and the wrapping method that checks the annotation and appends errors/inserts constants


compiler-core/src/type_/expression.rs
```diff
@@ -234,24 +234,28 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         self.environment.new_unbound_var()
     }
 
-    pub fn infer(&mut self, expr: UntypedExpr) -> Result<TypedExpr, Error> {
+    pub fn infer(
+        &mut self,
+        expr: UntypedExpr,
+        errors: &mut Vec<Error>,
```
lpil, line 240: How about we put `errors` on `self`? 

It's not specific to any one particular expr and every single function takes it as an argument, which makes me think it's an argument to the class as a whole, not to its methods. It would make this PR much smaller too!

compiler-core/src/type_/expression.rs
```diff
@@ -2307,27 +2366,88 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                     ValueConstructorVariant::Record { .. } => unreachable!(),
                 }
             }
-        }?;
+
+            Constant::Invalid { .. } => panic!("invalid constants can not be in an untyped ast"),
+        }
+    }
+
+    pub fn infer_const(
```
lpil, line 2374: This function has become quite hard to read and features a very large amount of nesting. How could we restructure it to be easier to follow?

compiler-core/src/type_/tests.rs
```diff
@@ -284,7 +287,16 @@ fn compile_statement_sequence(src: &str) -> Result<Vec1<TypedStatement>, crate::
...
+    .infer_statements(ast, errors);
+    match (res, Vec1::try_from_vec(errors.to_vec())) {
+        (Ok(res), Err(_)) => Ok(res),
+        (Ok(_), Ok(errors)) => Err(errors),
+        (Err(err), Ok(mut errors)) => {
+            errors.push(err);
+            Err(errors)
+        }
+        (Err(err), Err(_)) => Err(Vec1::new(err)),
+    }
```
lpil, line 299: What is this code doing?

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__subject_int_float_guard_tuple.snap
```diff
@@ -1,8 +1,20 @@
 ---
 source: compiler-core/src/type_/tests/errors.rs
-assertion_line: 917
 expression: "type X { X(a: Int, b: Float) }\nfn x() { case X(1, 2.0) { x if x == X(2.0, 1) -> 1 } }"
 ---
+error: Inexhaustive patterns
+  ┌─ /src/one/two.gleam:2:10
+  │
+2 │ fn x() { case X(1, 2.0) { x if x == X(2.0, 1) -> 1 } }
```
lpil, line 8: Let's add another case to avoid making it unclear what is being tested here

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__guard_record_wrong_arity.snap
```diff
@@ -1,8 +1,20 @@
...
 expression: "type X { X(a: Int, b: Float) }\nfn x() {\n  case X(1, 2.0) { x if x == X(1) -> 1 }\n}"
 ---
+error: Inexhaustive patterns
+  ┌─ /src/one/two.gleam:3:3
+  │
+3 │   case X(1, 2.0) { x if x == X(1) -> 1 }
+  │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This case expression does not have a pattern for all possible values.
+If it is run on one of the values without a pattern then it will crash.
```
lpil, line 12: Let's add another case to avoid making it unclear what is being tested here

compiler-core/src/type_/tests/errors.rs
```diff
@@ -1160,6 +1160,33 @@ fn const_annotation_wrong_4() {
...
+const tpl: String = #(Ok(1), MyInvalidType, 3)
+const lst = [1, 2.0]
+const unbound: MyInvalidType = MyInvalidType
+const assignment1: String = num
+const assignment2: String = str1
+const assignment3: String = tpl
+const assignment4: String = lst
+const assignment5: String = unbound"
+    );
+}
```
lpil, line 1188: What are these tests testing? There's a lot happening and I can't tell what the intent is. I don't think I would be able to tell if it is a bug or not if the output of these changes.

Instead we should have many small tests, each testing precisely one edge case, ideally with a comment or descriptive name to state what is being tested.

lpil (commented):
Very nice! Exciting that fault tolerant analysis is happening!

I've left some notes inline, mostly around style. The approach itself looks good to me

compiler-core/src/type_/expression.rs
Acepie: Yeah good call. Originally when I started I didn't realize it would have to be passed everywhere

compiler-core/src/type_/tests.rs
Acepie: There were a couple where we were directly instantiating and calling the exprtyper instead of going through the analysis module. In those tests we were only looking at the return value. This meant that the errors appended to the errors vector were ignores. This bit here is appending the errors vector with the return value so that all errors are accounted for in the returned error

compiler-core/src/type_/expression.rs
```diff
@@ -2300,16 +2303,72 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            // Type annotation is invalid but the inferred value is ok. Use the inferred type.
+            (Some(Err(annotation_err)), Ok(inferred)) => {
+                self.errors.push(annotation_err);
+                inferred
+            }
+            // Type annotation and inferred value are invalid. Place a placeholder constant with an unbound type.
+            // This should limit the errors to only the definition assuming the constant is used consistently.
+            (Some(Err(annotation_err)), Err(value_err)) => {
+                self.errors.push(annotation_err);
+                self.errors.push(value_err.clone());
```
lpil, line 2365: How come this is cloned?

compiler-core/src/type_/expression.rs
lpil: Thank you

Issue #2221, opened 2023-06-13
Linux gnu aarch64 compilation fails on CI


lpil:
```
error: linking with `aarch64-linux-gnu-gcc` failed: exit status: 1
  |
[...]

PR #3098, opened 2024-05-04
The Great CI Revamp

### LINKED ISSUE #1950: Remove actions-rs/* from workflows
They have been abandoned.

@tranzystorek-io says that there is an alternative for /toolchain here https://github.com/dtolnay/rust-toolchain

There is also Swatinem/rust-cache

Unclear what is needed to replace /cargo. We use its cross compilation features currently.

Pi-Cla:
We now no longer use any actions-rs and we can now build aarch64-unknown-linux-gnu again!

The build errors were caused by an old version of `ring` which `hexpm-rust` had still depended on, after I bumped `ring` on `hexpm-rust` it now build again.

Fixes #1950 #2221

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,13 +81,10 @@ jobs:
         if: ${{ matrix.target == 'x86_64-unknown-linux-musl'}}
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: ${{ matrix.toolchain }}
           target: ${{ matrix.target }}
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you!!

.github/workflows/ci.yaml
```diff
@@ -81,13 +80,10 @@ jobs:
         if: ${{ matrix.target == 'x86_64-unknown-linux-musl'}}
 
       - name: Install Rust toolchain
-        uses: actions-rs/toolchain@v1
+        uses: dtolnay/rust-toolchain@stable
```
hamirmahal, line 83: Nice!

PR #3099, opened 2024-05-05
Add basic semantic tokens implementation

u9g:
This pr adds an ast visitor and a semantic token implementation for
- `{module}.{label}`
- `{tuple}.{label}`
- `{tuple}.{index}`
- `{record}.{label}`
- pattern of variable (ie `let x = 9`)
- import coloring depending on whether the import is a type, function, record, or other

### DIFF
--- a/compiler-core/src/ast_visitor.rs
+++ b/compiler-core/src/ast_visitor.rs
@@ -1,0 +1,269 @@
+use std::sync::Arc;
+
+use ecow::EcoString;
+use vec1::Vec1;
+
+use crate::{
+    ast::{
+        Arg, Definition, Import, Module, Pattern, SrcSpan, Statement, TypedDefinition, TypedExpr,
+    },
+    type_::{ModuleInterface, ModuleValueConstructor, Type},
+};
+
+#[derive(Default)]
+pub struct Visitor<'a> {
+    pub owned_visit_import: Option<&'a mut dyn FnMut(&Import<EcoString>)>,
+    pub owned_visit_pattern_of_var: Option<&'a mut dyn FnMut(&SrcSpan, &EcoString, &Type)>,
+    pub owned_visit_record_access:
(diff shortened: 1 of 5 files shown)


lpil (commented):
Hello! What functionality does this provide to the user?

It would be best to open an issue to discuss new features before implementing them as there's no guarantee we can accept any particular change.

u9g:
> Hello! What functionality does this provide to the user?

Here's a before and after in vscode
image
image



> It would be best to open an issue to discuss new features before implementing them as there's no guarantee we can accept any particular change.

it's fine if this isn't accepted, I developed this for my own use but I also think it would be nice to see an implementation land in the standard LSP too.

u9g:
> Sorry, I'm a bit confused. It looks like it broke your syntax

This is using the latest released vscode-gleam, not my syntax.

inoas:
I think it shows function values versus other values in different colours

lpil:
In the after screenshot it seems to inconsistently highlight variables, and not all the keywords are highlighted any more.

That aside, what does this enable for the programmer?

MystPi:
I've been wanting something like this for a long time!

> In the after screenshot it seems to inconsistently highlight variables, and not all the keywords are highlighted any more.
> 
> That aside, what does this enable for the programmer?

It's not highlighting variables inconsistently; they are highlighted based on whether they are a function (yellow) or another value (blue). This is actually really helpful in other languages like JavaScript that do this:
image

The only keyword that looks wrong (from what I can tell) is the `as` in the topmost import. Likely something that would need to be fixed before merging.

nicklimmm:
We have an existing AST visitor implementation here: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/ast/visit.rs

lpil:
Closing due to inactivity. Please feel free to reopen in future.

Issue #3018, opened 2024-04-20
Formatter: Unexpectily moves last comment of custom type constructor args out to its parent type scope


inoas:
```gleam
type Select {
  Select(
    // comment: String,
    // modifier: String,
    // with: String,
    select: String,
    distinct: String,
    from: String,
    join: String,
    where: String,
    group_by: String,
    having: String,
    // window: String,
    order: String,
    limit: String,
    offset: String,
    union: String,
    // epilog: String,
  )
}
```
format:
```gleam
type Select {
  Select(
    // comment: String,
    // modifier: String,
    // with: String,
    select: String,
    distinct: String,
    from: String,
    join: String,
    where: String,
    group_by: String,
    having: String,
    // window: String,
    order: String,
    limit: String,
    offset: String,
    union: String,
  )
  // epilog: String,
}
```

lpil:
Thank you

Axot017:
I'd like to work on this! This is my first time working on this project, so if it takes too long feel free to take over the issue if necessary

lpil:
Awesome! Thank you!

Axot017:
I saw that @nino already created a PR with fix for that, but i investigated issue a little bit more and I found some related issues all caused by the same `wrap_args` function that was not aware of trailing comments that can occur after args. My PR is a little bit bigger because I tried to fix all of this issues, so I will leave it just in case (two possible solutions are better then one :P).

Here are some example issues that i also fixed:
```gleam
type Triple {
  Triple(a: Int, b: Int, c: Int)
}

fn main() {
  let triple = Triple(1, 2, 3)
  let Triple(
    really_really_long_variable_name_a,
    c: really_really_long_variable_name_c,
    ..,
    // comment
  ) = triple
  really_really_long_variable_name_c
}
```
was formatted to
```gleam
type Triple {
  Triple(a: Int, b: Int, c: Int)
}

fn main() {
  let triple = Triple(1, 2, 3)
  let Triple(
    really_really_long_variable_name_a,
    c: really_really_long_variable_name_c,
    ..,
  ) =
// comment 
triple
  really_really_long_variable_name_c
}
```
------------------------
```gleam
type Record {
  Record(
    // comment_line_1: String,
    // comment_line_2: String,
  )
}
```
Was formatted to
```gleam
type Record {
  Record
  // comment_line_1: String,
  // comment_line_2: String,
}
```
----------
```gleam
pub fn main() {
  fn(
    // A comment 1
    a,
[...]

PR #3101, opened 2024-05-05
Fix formatting of trailing comments in arguments

Axot017:
Fixes https://github.com/gleam-lang/gleam/issues/3018 and related issues. (After investigation it turned out that this bug occurs in every place where wrap_args was used so I fixed all of them)


lpil (approved):
Wonderful work!! Thank you!!

PR #3105, opened 2024-05-06
Add hint to use --replace flag when not explicitly replacing a release

### LINKED ISSUE #2271: gleam publish: Show hint about republishing instead of hexpm library error
gleam publish on the same version should show an hint on how to republish.

Also because hexpm talks about republishing maybe --republish is better than --replace, but that's minor.

## Outdated text

~~For a brief time hex.pm allows republishing on the same version tag it seems, but not gleam publish~~

> error: Hex API failure
> 
> There was a problem when using the Hex API.
> 
> This was error from the Hex client library:
> 
>     must explicitly express your intention to replace the release

~~I'd like to see a --force or --republish or --overwrite tag. I'd prefer --republish.
If given the explicit intention as mentioned in the error would be passed through.~~

Pi-Cla:
Resolves #2271

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -270,6 +270,9 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
 
     #[error("The Gleam module {path} would overwrite the Erlang module {name}")]
     GleamModuleWouldOverwriteStandardErlangModule { name: EcoString, path: Utf8PathBuf },
+
+    #[error("Version already published")]
+    HexPublishReplaceRequired { version: String },
 }
 
 impl Error {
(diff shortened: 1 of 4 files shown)


compiler-core/src/error.rs
```diff
@@ -270,6 +270,9 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
 
     #[error("The Gleam module {path} would overwrite the Erlang module {name}")]
     GleamModuleWouldOverwriteStandardErlangModule { name: EcoString, path: Utf8PathBuf },
+
+    #[error("--replace flag not used to replace a release")]
+    ReplaceRequired,
```
lpil, line 275: Let's call this `HexPublishReplaceRequired` as it's a bit cryptic otherwise, and mention Hex in the message above

compiler-core/src/error.rs
```diff
@@ -3365,6 +3368,14 @@ causing confusing errors and crashes.
                 location: None,
                 hint: Some("Rename this module and try again.".into()),
             }],
+
+            Error::ReplaceRequired => vec![Diagnostic {
+                title: "--replace flag not used to replace a release".into(),
```
lpil, line 3373:
```suggestion
                title: "Version already published".into(),
```

compiler-core/src/error.rs
```diff
@@ -3365,6 +3368,14 @@ causing confusing errors and crashes.
...
+                text: "Attempted to publish an already published release without the --replace flag.".into(),
```
lpil, line 3374: Let's make this something like "There is already a version v1.0.0 of this library on Hex. This version has been recently published so you can replace it with new code, or you can publish using a different version number", with the wrap function.

lpil (commented):
Very nice! Thank you

lpil (approved):
Thank you!!

PR #3111, opened 2024-05-08
Try to compile time error when trying to make non byte aligned array on js

### LINKED ISSUE #1591: Currently bitstring segment size must be multiple of 8 when compiling to JavaScript, add compiletime check
```gleam
// ok
<<1:size(8)>>
<<1:size(x)>>

// not ok
<<1:size(7)>>
```
This won't catch all errors, because x could evaluate at runtime to a value that isn't a multiple of 8. However catching the literal cases when the term is explicitly an integer will still be valuable.

Pi-Cla:
Try to detect when someone tries to make a non byte aligned array on Javascript. This compile time error still fails to trigger if the size value is a variable though. So we still need the runtime error in the Javascript template too.

Fixes #1591

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,24 @@
   reserved word in Erlang/OTP 27.
   ([Jake Barszcz](https://github.com/barszcz))
 
+- Non byte aligned arrays that use literals for size are now marked as an
+  Unsupported feature for Javascript since they would already cause
+  a runtime error on Javascript.
+
+  This means if you compile specifically for Javascript you will now recieve this error:
+
+  ```
+  error: Unsupported feature for compilation target
+    ┌─ /src/test/gleam_test.gleam:6:5
+    │
+  6 │   <<1:size(5)>>
+    │     ^^^^^^^^^
+
+  Non byte aligned array is not supported for JavaScript compilation.
+  ```
+
+  Else any functions which rely on this will not be compiled into Javascript.
+
 ### Formatter
 
 ### Language Server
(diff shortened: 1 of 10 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__explicit_sized.snap
```diff
@@ -1,11 +1,5 @@
...
-assertion_line: 71
 expression: "\nfn go() {\n  <<256:size(4)>>\n}\n"
 ---
-import { toBitArray, sizedInt } from "../gleam.mjs";
-
-function go() {
-  return toBitArray([sizedInt(256, 4)]);
-}
-
+import { toBitArray } from "../gleam.mjs";
```
lpil, line 5: What happened to these two tests?

lpil (commented):
Thank you!!

I would put the tests in https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/javascript/tests/bit_arrays.rs

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__explicit_sized.snap
Pi-Cla: This snapshot correspond to this test:
https://github.com/gleam-lang/gleam/blob/f863f0ba57c184b6b50eef012bc2a0f39204b160/compiler-core/src/javascript/tests/bit_arrays.rs#L70-L78
Since the size of this array is being set to 4 it would not be byte aligned and it would cause a runtime error. So instead this code marks that as unsupported and so removes the function it is contained in.

I guess I can rewrite these tests to have a size which is a multiple of 8 instead and also add in some new tests which use not byte aligned bit arrays.

lpil (approved):
Wonderful! And thank you so much for waiting for me

Issue #3122, opened 2024-05-10
Error message says "field" when it should say "label"


lpil:
```
          0 │+error: Duplicate field
          1 │+  ┌─ /src/one/two.gleam:2:22
          2 │+  │
          3 │+2 │ pub fn one(a a: Int, a b: Float) -> Nil {
          4 │+  │                      ^^^^^^^^^^
          5 │+
          6 │+The field `a` has already been defined. Rename this field.
```

It is the label that is named `a`.

PR #3124, opened 2024-05-10
Change 'field' to 'label' in TypeError::DuplicateField definition

aelishRollo:
According to https://github.com/gleam-lang/gleam/issues/3122 there's an error message incorrectly identifying a "label" as a "field". This PR changes the error message to say "label" as shown below:

Before

``The field `a` has already been defined.``

After

``The label `a` has already been defined.``


lpil (approved):
Thank you!

PR #3128, opened 2024-05-11
fix formatting when comment is before first statement in anonymous function

### LINKED ISSUE #3109: Comment put in wrong place by formatter
When formatting the code snippet:

```gleam
import gleam/option.{None}

pub fn main() {
  None
  |> option.map(fn(a) { //
  a })
}
```

The formatted code ends up being:

```gleam
import gleam/option.{None}

pub fn main() {
  None
  |> option.map(
    fn(
      a,
      //
    ) {
      a
    },
  )
}
```

Which I think is wrong, as the comment should end up on the first line of the block.

Acepie:
Closes #3109

The root cause of the issue here is that https://github.com/gleam-lang/gleam/commit/2d56b6a2554c1023f15e432eda5f16831d54c1e2#diff-6e737fd16c3877b93866d0f63709896f2f7b2f1afa64a6543a8344f99f7ebe96R2383 with this change the arg wrapping is dependent on a passed in location to determine when to stop looking for comments but in the AST for functions we use locations inconsistently between top level function definitions and anonymous function definitions. In top level functions, the location refers to only the function head and so https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/format.rs#L753 points to the exact opening bracket of the body. On the other hand for anonymous functions, the location actually include the whole function meaning that the ast does not know where the real location of the body opening bracket is. Instead the current implementation looks at the start of the first statement https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/format.rs#L794 hence the issue.

To solve the issue I added a new field to the AST that would consistently be the location of the head. For top level functions it would be exactly the same as the location whereas for anonymous functions the behavior would match top level.

### DIFF
--- a/compiler-core/src/ast/untyped.rs
+++ b/compiler-core/src/ast/untyped.rs
@@ -31,11 +31,11 @@ pub enum UntypedExpr {
 
     // TODO: create new variant for captures specifically
     Fn {
-        /// The location from the start of the `fn` keyword to the end of the head before the opening bracket
-        head_location: SrcSpan,
         /// For anonymous functions, this is the location of the entire function including the end of the body.
         /// For named functions, this is the location of the function head.
         location: SrcSpan,
+        /// The byte location of the end of the function head before the opening bracket
+        end_of_head_byte_index: u32,
         is_capture: bool,
         arguments: Vec<Arg<()>>,
         body: Vec1<UntypedStatement>,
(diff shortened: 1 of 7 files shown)


compiler-core/src/ast/untyped.rs
```diff
@@ -31,6 +31,10 @@ pub enum UntypedExpr {
 
     // TODO: create new variant for captures specifically
     Fn {
+        /// The location from the start of the `fn` keyword to the end of the head before the opening bracket
+        head_location: SrcSpan,
```
lpil, line 35: This duplicates the start of the `location` srcspan and so the start is redundant. It also isn't used in the code. We can store just the int index of the end instead.

lpil (commented):
Thank you. I've left one note inline

lpil (approved):

PR #3130, opened 2024-05-11
Show helpful error message when calling a function in a constant value

### LINKED ISSUE #2983: Improve error message for function call inside constant
```gleam
pub fn thing() {
  "a"
}

const wibble = thing()
```
This snippet of code results in this error message:
```
error: Syntax error
  ┌─ /src/main.gleam:7:21
  │
7 │ const wibble = thing()
  │                     ^ I was not expecting this

Expected one of: 
An import, const, type, if block, or function.
```
I think that's because the compiler sees it as:
```gleam
const wibble = thing
()
```
And not as a function call

Another example is when you wrap the function call in a constructor/list/etc:
```gleam
const thing = [wibble()]
//                   ^ I was not expecting this
pub fn wibble() { todo }
```

nino:
This should resolve #2983.

Attempting to call a function in a constant definition will now show a more helpful error message:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:3:18
  │
3 │ const wib: Int = wibble(1, "wobble")
  │                  ^^^^^^^^^^^^^^^^^^^ Functions can only be called within other functions
```


nino:
@lpil Turns out, there already was [such a test case](https://github.com/gleam-lang/gleam/pull/3130/commits/94dd468a49613cc18c7e2807f5b5fbb909bc8faf#diff-8d8ee5aff3e9e4044d999def8e48e311d97f035956d499fffe4e4fdcaf7d9ccc). I've updated the code to make the error message look the same whether or not the function is from another module. (i.e., the error highlights the whole function call from beginning of name to end of arguments)

lpil (commented):
I'm a bit unsure about using the expression function args parsing method here as it emits errors relating to expressions, not constants. This will be inaccurate at times.

Previously we stopped when we saw a `(` as this is enough to know that it's a function call. How come you decided to move away from that here?

nino:
@lpil 

> as it emits errors relating to expressions, not constants

That's a good point. I'd decided to parse the function params because that would make the error message nicer, by highlighting the whole function call, rather than just the name and opening paren. (Also hadn't realized that the parser already checks for `(` after imported identifiers (`const a = b.c()` already showed a reasonable error).)

I guess, to get the error message to highlight the whole function call, we'd have to wrap `parse_fn_args` in some custom error handling logic or something. Which is probably overkill.

I've now changed the code for `const a = b()` to match the existing code for `const a = b.c()`. ([79d5a50](https://github.com/gleam-lang/gleam/pull/3130/commits/79d5a50c2a2adf6a3a51ee8934c2d9e6aecf2ccf)) Now the error message only highlights up to the opening paren, but it won't show wrong error messages inside the parens.

lpil (approved):
Ah! OK cool that sounds great.

Thank you once again! This should help people a lot 💜

Issue #3110, opened 2024-05-08
dependencies::is_same_requirements is incorrect


u9g:
`dependencies::is_same_requirements` is incorrect when a requirement is made using an absolute path in the `gleam.toml` 

this is what the left and right look like after going through the maybe canonicalize operations: 

` INFO download_deps: left: \\?\C:\Users\---\Documents\code\5-7-24\stdlib right: C:\Users\---\Documents\code\5-7-24\stdlib`

The solution is to just unconditionally canonicalize both paths.

lpil:
Ah! Thank you.

Francisco-Montanez:
I can take this if not done

lpil:
Thank you @Francisco-Montanez !

PR #3135, opened 2024-05-13
Fix incorrect behavior in dependencies::is_same_requirements

Francisco-Montanez:
This pull request addresses a bug where path comparisons could fail due to discrepancies in path formatting. The solution is to canonicalize both paths.


Francisco-Montanez:
It should fix #3110

lpil (approved):
Lovely! Thank you

Issue #3126, opened 2024-05-10
Inconsistent aliased string prefix matching on Erlang when prefix contains control/unicode characters.


aznashwan:
Matching against a String which starts with a backslash using an alias for the first part of the match leads to inconsistent behaviour on Erlang.

Sample code with intentional redundant patterns to showcase the issue:

```gleam
import gleam/io

pub fn match_leading_slash(str: String) -> String {
  case str {
    "/" as start <> rest -> {
      "Matched starting '/' with alias"
    }
    "/" <> rest -> {
      "Matched starting '/' without alias"
    }
    "\\" as start <> rest -> {
      "Matched starting '\\' with alias"
    }
    "\\" <> rest -> {
      "Matched starting '\\' without alias"
    }
    _ -> "Did not match anything"
  }
}

pub fn main() {
  let _ = io.debug(match_leading_slash("/usr/bin"))
  let _ = io.debug(match_leading_slash("\\\\.\\pipe\\test"))
}
```

Output:
```bash
$ gleam run --target javascript
   Compiled in 0.01s
    Running backslash_alias_test.main
"Matched starting '/' with alias"
"Matched starting '\\' with alias"

$ gleam run --target erlang
   Compiled in 0.01s
    Running backslash_alias_test.main
"Matched starting '/' with alias"
# NOTE: this should have also matched the first clause with the alias like JS.
"Matched starting '\\' without alias"
```

Tested on macOS with following Homebrew-installed tooling:
```bash
$ brew info gleam
==> gleam: stable 1.1.0 (bottled), HEAD

$ brew info erlang
==> erlang: stable 26.2.5 (bottled), HEAD

$ brew info rebar3
==> rebar3: stable 3.23.0 (bottled)

$ brew info node
==> node: stable 22.1.0 (bottled), HEAD
```

Here's the generated Erlang FWIW:

```erlang
%%% $ cat ./build/dev/erlang/backslash_alias_test/_gleam_artefacts/backslash_alias_test.erl

-module(backslash_alias_test).
-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).

-export([match_leading_slash/1, main/0]).

-spec match_leading_slash(binary()) -> binary().
match_leading_slash(Str) ->
    case Str of
        <<Start:1/binary, Rest/binary>> when Start =:= <<"/">> ->
            <<"Matched starting '/' with alias"/utf8>>;

[...]

PR #3139, opened 2024-05-15
Fix String with escapes prefix matching on Erlang.

aznashwan:
Fixes gleam-lang/gleam#3126.

Fixes String prefix matching on Erlang when the prefix contains escape characters or unicode codepoints by ensuring the correct prefix byte length is matched against.


lpil (commented):
Hello! Thank you for this

Could you reuse the code here please? 🙏  https://github.com/gleam-lang/gleam/blob/2c59b9c3f9b00fe0c35726e00e97f7843eff5c1f/compiler-core/src/javascript/pattern.rs#L856

It'll need to be make slightly more flexible to specify whether it's utf8 (Erlang) or utf16 (JavaScript)

aznashwan:
> Could you reuse the code here please?

Done, hadn't noticed it before since I only investigated the Erlang part of the compiler tree but it was basically exactly what I needed.

I've adapted the function to directly return the updated string, properly treat the escape chars Gleam supports, and moved it to a [top-level `crate::strings` module](https://github.com/gleam-lang/gleam/pull/3139/files#diff-c7e6b4e895ded4ed39bf97d94cb55c4c7561637ae8152f76f1e1bbdd2b2228a3R6) for lack of a better place to put it, but please feel free to suggest a new one.

I've also split the specific prefixes tested against to make the transpiled code a little easier to follow.

lpil (approved):
Lovely! Thank you.

Issue #3125, opened 2024-05-10
Error message more helpful when having triple equals instead of double inside lambda fn


hagenek:
[1,2,3] |> list.filter(fn (a) { a === 3 }) will produce an error :

```
error: Syntax error
  ┌─ /src/main.gleam:6:37
  │
6 │ [1,2,3] |> list.filter(fn (a) { a === 3 }) 
  │                                     ^ There must be a 'let' to bind variable to value
```
Hint: Use let for binding.
See: https://tour.gleam.run/basics/assignments/

lpil:
Thank you. We don't need to have a specific `===` error message, but let's at least make this not show this error message in this case.

giacomocavalieri:
It could be cool to have a specific error message for this case, especially for folks used to JS:
```text
error: Syntax error
  ┌─ /src/main.gleam:6:37
  │
6 │ [1,2,3] |> list.filter(fn (a) { a === 3 }) 
  │                                   ^^^ Did you mean `==`?

Gleam uses `==` to check for equality between two values.
```

rabingaire:
Humm, I wonder if this error message is enough?

```
error: Syntax error
  ┌─ /Users/rabingaire/Desktop/gleam/example/test_app/src/test_app.gleam:4:37
  │
4 │   [1,2,3] |> list.filter(fn (a) { a === 3 }) 
  │                                     ^^ This operator has no value on its right side

Hint: Remove it or put a value after it.
```

giacomocavalieri:
Yeah I agree that could be more than enough for all other binary operators, but in case of `===` in particular I think the compiler should try being a bit more helpful and show an error message like the one I showed before

inoas:
Because people are so used to strictly equals why can't the formatter just kill it with fire and replace it with double equals?

giacomocavalieri:
That could also be an option, but a good error message is still required as not everyone runs the formatter

inoas:
> That could also be an option, but a good error message is still required as not everyone runs the formatter

100%

PR #3150, opened 2024-05-17
Better error message for triple equals

rabingaire:
fix: #3125

This is the new error message after the fix:
```
error: Syntax error
  ┌─ /Users/rabingaire/Desktop/test_app/src/test_app.gleam:4:37
  │
4 │   [1,2,3] |> list.filter(fn (a) { a === 3 }) 
  │                                     ^^^ Did you mean `==`?

Gleam uses `==` to check for equality between two values.
See: https://tour.gleam.run/basics/equality
```

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -29,6 +29,7 @@ pub enum LexicalErrorType {
     BadName { name: String },
     BadDiscardName { name: String },
     BadUpname { name: String },
+    InvalidTripleEqual,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -372,6 +373,13 @@ impl LexicalError {
             LexicalErrorType::InvalidUnicodeEscape(InvalidUnicodeEscapeError::InvalidCodepoint) => {
                 ("Invalid Unicode codepoint", vec![])
             }
+            LexicalErrorType::InvalidTripleEqual => (
+                "Did you mean `==`?",
+                vec![
+                    "Gleam uses `==` to check for equality between two values.".into(),
+                    "See: https://tour.gleam.run/basics/equality".into(),
+                ],
+            ),
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -265,6 +265,20 @@
     │                  ^^^^^^^ Functions can only be called within other functions
   ```
 
+- The compiler will now provide more helpful error messages when triple equals
+  are used instead of double equals. ([Rabin Gaire](https://github.com/rabingaire))
+
+  ```
+  error: Syntax error
+    ┌─ /src/parse/error.gleam:4:37
+    │
+  4 │   [1,2,3] |> list.filter(fn (a) { a === 3 })
+    │                                     ^^^ Did you mean `==`?
+
+  Gleam uses `==` to check for equality between two values.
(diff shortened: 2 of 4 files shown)


lpil (commented):
Thank you. Could you add a test to ensure that this code does not result in that helpful error:

```
==     =
```

I think currently it will show the error in this case, which isn't what we want.

It may be better to implement this in the tokeniser rather than in the parser.

rabingaire:
> Thank you. Could you add a test to ensure that this code does not result in that helpful error:
> 
> ```
> ==     =
> ```
> 
> I think currently it will show the error in this case, which isn't what we want.
> 
> It may be better to implement this in the tokeniser rather than in the parser.

Thank you for the recommendation, I made the changes and now the logic is handled in tokenizer, let me know if the error message is good enough, if it is not we can add a special case like these 

```
LexicalErrorType::UnrecognizedToken { tok } if *tok == '\'' => (
    "Unexpected single quote",
    vec!["Hint: Strings are written with double quotes.".into()],
),
```

lpil:
Hello! I gave this a try and I got this error:

```
error: Syntax error
  ┌─ /Users/louis/Desktop/thing/src/thing.gleam:4:5
  │
4 │   ===
  │     ^ I can't figure out what to do with this character

Hint: Is it a typo?
```

This doesn't seem any clearer to me than the previous error. Let's have one like Giacomo suggested in the issue, that'll help the programmer understand what the problem is.

I also tested with `== =` and I got the same error, but it should have the original error and the new error should only be for `===`

rabingaire:
> Hello! I gave this a try and I got this error:
> 
> ```
> error: Syntax error
>   ┌─ /Users/louis/Desktop/thing/src/thing.gleam:4:5
>   │
> 4 │   ===
>   │     ^ I can't figure out what to do with this character
> 
> Hint: Is it a typo?
> ```
> 
> This doesn't seem any clearer to me than the previous error. Let's have one like Giacomo suggested in the issue, that'll help the programmer understand what the problem is.
> 
> I also tested with `== =` and I got the same error, but it should have the original error and the new error should only be for `===`

Thank you for the suggestions, made changes as requested please check 👍🏽

compiler-core/src/parse/error.rs
```diff
@@ -372,6 +373,20 @@ impl LexicalError {
...
+                    "See: https://tour.gleam.run/basics/equality".into(),
+                ],
+            ),
+            LexicalErrorType::InvalidOperator { expected } => (
+                "This is not a valid operator at this context",
+                vec![
+                    "Hint: Is it a typo?".into(),
+                    format!("Try: `{}` operator instead.", expected),
+                ],
+            ),
```
giacomocavalieri, line 389: I'd personally define a specific `InvalidTripleEqual` instead of a general one since we don't currently have any uses for it besides the triple equal

rabingaire: Thank you for the review updated the commit 👍🏽

giacomocavalieri: No, thank you!

lpil (commented):
Thank you! Very nice. Could you update the changelog file too? Thanks

rabingaire:
> Thank you! Very nice. Could you update the changelog file too?

Thank you, added the changelog 👍🏽

lpil (approved):
Wonderful! Thank you

PR #3151, opened 2024-05-17
Fix formatting of commented case branches

### LINKED ISSUE #3140: Wrong formatting in `case` expressions in presence of a long comment
I noticed a weird behaviour when commenting a case with multiple patterns and pointed it out to @giacomocavalieri to assure it wasn't intended behaviour.

Here's an example:
```gleam
fn foo(a: Int, b: Int) -> Bool {
  case a, b {
    // a very long comment a very long comment a very long comment a very long comment
    1, 2 -> True
    _, _ -> False
  }
}
```

After the formatting, the code above becomes the following:
```gleam
fn foo(a: Int, b: Int) -> Bool {
  case a, b {
    // a very long comment a very long comment a very long comment a very long comment
    1, 2
    -> True
    _, _ -> False
  }
}
```

Such behaviour happens only when the comment is quite long (i suppose > 80 characters?) and in case of a shorter one the formatter works as expected.
```gleam
fn foo(a: Int, b: Int) -> Bool {
  case a, b {
    // a not quite long comment
    1, 2 -> True
    _, _ -> False
  }
}
```

### Version info
OS: Windows 11
gleam 1.1.0
rebar 3.22.1 on Erlang/OTP 25 Erts 13.2.2.2

### Logs
[link to pastebin](https://pastebin.com/xzP0wawD)

giacomocavalieri:
This PR closes #3140

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,11 +58,13 @@
 
   Hint: Rename this module and try again.
   ```
+
 - New subcommand `gleam hex revert` added. ([Pi-Cla](https://github.com/Pi-Cla))
-  * You can specify the options like this: `gleam hex revert --package gling --version 1.2.3`
-  * A new package can be reverted or updated within 24 hours of it's initial publish,
-  a new version of an existing package can be reverted or updated within one hour.
-  * You could already update packages even before this release by running: `gleam publish` again.
+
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!

PR #3156, opened 2024-05-18
LSP: ignore completion in comments

### LINKED ISSUE #2161: LSP: do not provide autocompletions within comments
Heres a screenshot of this issue
image

Rust's LSP and pylance do not have this behavior. I think that is what gleam should do.

I am on gleam version nightly-230517 installed through asdf.

nicklimmm:
Closes #2161 

Simply check if the current cursor is in a comment context by comparing comment spans (from Module's extras).

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -18,7 +18,7 @@ use camino::Utf8PathBuf;
 use ecow::EcoString;
 use lsp::CodeAction;
 use lsp_types::{self as lsp, Hover, HoverContents, MarkedString, Url};
-use std::sync::Arc;
+use std::{cmp::Ordering, sync::Arc};
 
 use super::{
     code_action::{CodeActionBuilder, RedundantTupleInCaseSubject},
@@ -203,13 +203,39 @@ where
                 .module_line_numbers
                 .byte_index(params.position.line, params.position.character);
 
-            // Do not suggest in comments
-            if module.extra.is_within_comment(byte_index)
-                || (byte_index > 0
-                    && module
-                        .extra
-                        // Cursor is at the end of the comment
-                        .is_within_comment(byte_index - 1))
+            // Use binary search to find if the byte_index is
+            // in the range [span.start, span.end]
+            //
+            // Do note that the end is inclusive, because we would like to
+            // detect if the cursor is right after the comment
+            // (when we are typing/insert mode in Vim)
+            let comment_cmp = |span: &SrcSpan| {
+                if byte_index < span.start {
+                    Ordering::Less
+                } else if span.end < byte_index {
+                    // inclusive end
+                    Ordering::Greater
+                } else {
+                    Ordering::Equal
+                }
+            };
+
+            // If in comment context, do not provide completions
+            //
+            // We cannot simply use `module.extra.is_within_comment` as it
+            // cannot detect if we are currently typing/inserting text at the
+            // end of the comment
+            if module.extra.comments.binary_search_by(comment_cmp).is_ok()
+                || module
+                    .extra
+                    .doc_comments
+                    .binary_search_by(comment_cmp)
+                    .is_ok()
+                || module
+                    .extra
+                    .module_comments
+                    .binary_search_by(comment_cmp)
+                    .is_ok()
             {
                 return Ok(None);
             }
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -18,7 +18,7 @@ use camino::Utf8PathBuf;
 use ecow::EcoString;
 use lsp::CodeAction;
 use lsp_types::{self as lsp, Hover, HoverContents, MarkedString, Url};
-use std::{cmp::Ordering, sync::Arc};
+use std::sync::Arc;
 
 use super::{
     code_action::{CodeActionBuilder, RedundantTupleInCaseSubject},
@@ -203,40 +203,8 @@ where
                 .module_line_numbers
                 .byte_index(params.position.line, params.position.character);
 
-            // Use binary search to find if the byte_index is
-            // in the range [span.start, span.end]
-            //
-            // Do note that the end is inclusive, because we would like to
-            // detect if the cursor is right after the comment
-            // (when we are typing/insert mode in Vim)
-            let comment_cmp = |span: &SrcSpan| {
-                if byte_index < span.start {
-                    Ordering::Less
-                } else if span.end < byte_index {
-                    // inclusive end
-                    Ordering::Greater
-                } else {
-                    Ordering::Equal
-                }
-            };
-
             // If in comment context, do not provide completions
-            //
-            // We cannot simply use `module.extra.is_within_comment` as it
-            // cannot detect if we are currently typing/inserting text at the
-            // end of the comment
-            if module.extra.comments.binary_search_by(comment_cmp).is_ok()
-                || module
-                    .extra
-                    .doc_comments
-                    .binary_search_by(comment_cmp)
-                    .is_ok()
-                || module
-                    .extra
-                    .module_comments
-                    .binary_search_by(comment_cmp)
-                    .is_ok()
-            {
+            if module.extra.is_within_comment(byte_index) {
                 return Ok(None);
             }
 
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -203,6 +203,17 @@ where
                 .module_line_numbers
                 .byte_index(params.position.line, params.position.character);
 
+            // Do not suggest in comments
+            if module.extra.is_within_comment(byte_index)
+                || (byte_index > 0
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/engine.rs
```diff
@@ -205,6 +205,17 @@ where
             let byte_index =
                 line_numbers.byte_index(params.position.line, params.position.character);
 
+            // Do not suggest in comments
+            if module.extra.is_within_comment(byte_index)
+                || (byte_index > 0
+                    && module
+                        .extra
+                        // Cursor is at the end of the comment
+                        .is_within_comment(byte_index - 1))
```
lpil, line 214: How come this is done here and not within `is_within_comment`?

lpil (commented):
Thank you

compiler-core/src/language_server/engine.rs
nicklimmm: `is_within_comment` simply checks the byte index is in the range `[span.start, span.end)` in any comment.

For example:

```gleam
// comment
          ^cursor is at byte 10 in insert mode (after `t`)
```

The comment above has the span of `[2, 10)`, then `is_within_comment(10)` here returns false. So to ensure it knows it is in the comment context, we `-1` it.

For completeness, only using `byte_index - 1` doesn't cover the case when the following happens:

```gleam
//
  ^cursor is at byte 2 (after the last `/`)
```

The above has the span of `[2, 3)`, then `is_within_comment(2 - 1)` here returns false. So we don't `-1` in this case.

lpil: Right, so why isn't this done inside the method? Seems like without that it doesn't do what it claims to do. Is that right?

nicklimmm: > Right, so why isn't this done inside the method? Seems like without that it doesn't do what it claims to do. Is that right?

`is_within_comment` is working as intended, which is to check if the byte index is in a comment.

If we try to incorporate this `-1` adjustment into the method, we will break things. 

Suppose we have an empty comment, then followed by a newline character (`//\n`). If we have a byte index that points to the newline (2 in this case, in vim's normal mode), we shouldn't include the newline in the comment. This means that we shouldn't `-1` the byte index in the method.


For reference, changing from normal to insert mode in the last character of the line bumps the offset number as shown below,

image

image

nicklimmm: Ignore this section for now, I've found another uncovered edge case on empty comments.

lpil: I don't understand why the logic for whether a byte index is inside a comment is different here than in other situations. Could you please explain with examples for both situations. Thank you

compiler-core/src/language_server/engine.rs
```diff
@@ -205,6 +205,43 @@ where
             let byte_index =
                 line_numbers.byte_index(params.position.line, params.position.character);
 
+            // Use binary search to find if the byte_index is
```
nicklimmm, line 208: @lpil Moving the discussion of different logic to here

compiler-core/src/language_server/engine.rs
```diff
@@ -205,6 +205,43 @@ where
...
+                    .is_ok()
+                || module
+                    .extra
+                    .module_comments
+                    .binary_search_by(comment_cmp)
+                    .is_ok()
+            {
+                return Ok(None);
+            }
+
```
nicklimmm, line 244: @lpil Moving the discussion here after I made some changes to cover all edge cases.

nicklimmm: Instead of "is this cursor in a comment?", we should view this as "if I insert a char in this cursor, would it be included in the comment?". If true, then do not provide completions.

To demonstrate the new POV, suppose that we have a comment with 2 characters in it

```
//ab
  ^^^ don't provide completions if our cursor is in any of these positions
```

Expanded version:

```
//|ab
  ^ cursor is at byte index 2 (no completion)
//a|b
   ^ cursor is at byte index 3 (no completion)
//ab|
    ^ cursor is at byte index 4 (no completion)
```

So, with the new POV, we should check if the byte index is in the range `[start, end]` (note the inclusive end)

nicklimmm: To compare:

```
is_within_comment:

//abc
  ^^^ considers this span
```

```
new logic to avoid completions in comments:

//abc
  ^^^^ considers this span (extended by 1)
```

nicklimmm: If the `is_within_comment` has the same logic as the new one, then this wouldn't make much sense:

Suppose that the source code is only `//abc`:

```
//abc
     ^ is_within_comment(5) -> true (?)
```

But the source code only has 5 characters, so if we index the source code with 5 (e.g. `code[5]`), it would panic

nicklimmm: Once we have a common understanding, I'll put the updated explanation in the implementation

lpil: You've not included an example for when the cursor should not be considered in a comment? All of the examples you've given are within a comment.

nicklimmm: Technically, the definition of "is_within_comment(byte_index)" is `comment_span.start <= byte_index < comment_span.end`

So the cursor in `//abc|` (where the character `|` refers to the cursor) is not in a comment (if we use `is_within_comment`)

lpil: And what is the code example that makes that true? All the examples you've given are evidence that it is not the case, indicating an off-by-one error.

nicklimmm: Examples of correct cases (`is_within_comment` returns true):

```
//|abc
//a|bc
//ab|c
```

lpil: Right, and what's the counterexample that shows that the modifications you've made outside of the function shouldn't be moved inside the function?

nicklimmm: The modifications made outside of the function are in green (added, which is a special case used for ignoring the completion), and the line in red is the one from `SrcSpan::cmp_byte_index`

```diff
  if byte_index < span.start {
      Ordering::Less
- } else if span.end <= byte_index {
+ } else if span.end < byte_index {
+    // inclusive end
      Ordering::Greater
  } else {
      Ordering::Equal
  }
```

lpil: Again, please give the examples that explain why a special case is needed. Thank you.

nicklimmm: So, a special case is needed here to cover a case whenever we ask for completions right after `//`.

Using only `is_within_comment` here triggers completions right after `//`:

https://github.com/gleam-lang/gleam/assets/65026286/63404313-c72f-41c7-a2df-a7ea129b688a


But using the special case doesn't trigger completions right after `//`:

https://github.com/gleam-lang/gleam/assets/65026286/481c3a49-f083-48e5-a7c5-89a1221a2611

lpil: I'm not asking for examples of when this is needed for comments, I'm asking for how it is not needed for the other uses of this function.

nicklimmm: I don't think there are any other uses for this "modified" version for now.

Unless, in the future, we need to detect whether we are in a comment during insert mode (vim) for some LSP-related features.

nicklimmm: Bumping this in case you missed this @lpil

nicklimmm: The existing uses are only for detecting if a char is inside a comment context for deleting redundant tuples in case exprs (https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/language_server/code_action.rs#L242).

The new behaviour doesn't suit the existing uses because of this one specific case:

In raw/escaped chars,

```gleam
// comment\n0
          ^^this is a newline
```

which is basically,

```gleam
// comment
0
```

The comment span only covers `// comment`. But if we use the new behaviour, it would say that the newline (`\n`) is within comment, i.e. `is_within_comment() -> true` (basically took more characters than it should).

lpil: Could you give an example of when the new behaviour is not suitable for the other uses please.

lpil:
Hello! It would be lovely to get this merged! Could you either move that code inside the method OR provide examples that detail why it is unsuitable for the other uses of the function. Thank you

nicklimmm:
> Hello! It would be lovely to get this merged! Could you either move that code inside the method OR provide examples that detail why it is unsuitable for the other uses of the function. Thank you

I believe moving the new logic inside the existing method should (technically) not break anything. I'll test it and update this PR soon

nicklimmm:
It works as intended after moving the new logic! So I guess it is good to go, except for this one test that failed due to network timeout.

We can try to re-test and I think it should be fine.

lpil (approved):
Beautiful!!! Thank you

PR #3160, opened 2024-05-19
Add rust-version

dahlbaek:
The MSRV added here is the version of rustc currently shipped
with Debian Testing (trixie): https://packages.debian.org/testing/rustc.
This should make it clear (in principle) how to add the gleam compiler
to the main Debian component of the next stable Debian release.

Some tweaks are made to the lockfile in order to support this MSRV.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -274,9 +274,9 @@ dependencies = [
 
 [[package]]
 name = "bumpalo"
-version = "3.15.4"
+version = "3.14.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7ff69b9dd49fd426c69a0db9fc04dd934cdb6645ff000864d98f7e2af8830eaa"
+checksum = "7f30e7476521f6f8af1a1c4c0b8cc94f0bee37d91763d0ca2665f299b6cd8aec"
 
 [[package]]
 name = "bytes"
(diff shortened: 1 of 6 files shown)


dahlbaek:
closes #3163

lpil:
What's the motivation for this? The only version we officially support is the latest stable. Wouldn't requiring that cause problems for packaging for Debian as they use an older version?

dahlbaek:
For a binary package, that's a reasonably policy I think. When building, one can always choose to ignore the meta-data (cargo build --ignore-rust-version), but I think the more common case will be that when it's time for a new release, one chooses a version of rustc to package, and then it'll be straightforward to pick the latest release of gleam that supports that version of rustc.

dahlbaek:
I'm looking into upstreaming small tweaks to the impacted dependencies so their latest version will support 1.71.1

- https://github.com/fitzgen/bumpalo/pull/252
- https://github.com/rustonaut/vec1/pull/34
- https://github.com/pulldown-cmark/pulldown-cmark/pull/894

lpil:
>  then it'll be straightforward to pick the latest release of gleam that supports that version of rustc (via rust-version).

This will result in a very poor experience for users of the Gleam package to the point where I don't believe there's any value in having the package.

dahlbaek:
Following discussion in  #3163, I'll close this as well.

PR #3169, opened 2024-05-20
Refresh and cleanup all snapshots

giacomocavalieri:
I noticed there were quite some old snapshots not connected to any test so I deleted all snapshot folders and generated all the snapshots from scratch.
It looks like `insta` is no longer storing the assertion line in the snapshots so it removed it from all 1024 of them 😬


lpil (approved):
Thank you!!!

PR #3173, opened 2024-05-21
Module loader ignores previously matching cache if there have been recent changes in LSP mode

### LINKED ISSUE #3172: Language server does not update module information if the code matches the last successful compile
Was testing out the new fault tolerance support with the language server and ran into a subtle case where the language server's module information is not properly updated. If the code of the last successful compilation exactly matches the current code then the compiler will reuse the old cached version of the module instead of recompiling the module. This causes issues for the language server since we now update the language server with incomplete/partial module information.

Example

```gleam
import gleam/int
import gleam/io

pub fn main() {
  7
  |> int.to_string
  |> io.println
}
```

If you comment out ` |> int.to_string` and then comment it back in, you will notice that language server functionality stops working within main.

Making any change that invalidates the cache (either a lsp reload or even just adding any white space) fixes the issue so its definitely not a high priority but was annoying/confusing when I first ran into it.

Acepie:
Closes #3172

In LSP mode the last cached build is not necessarily the last version of the module that was loaded by the LSP so if there have been source changes we ignore the cache even if the source matches the last successful cached version

### DIFF
--- a/compiler-core/src/build/module_loader.rs
+++ b/compiler-core/src/build/module_loader.rs
@@ -1,7 +1,7 @@
 #[cfg(test)]
 mod tests;
 
-use std::time::SystemTime;
+use std::{collections::HashSet, time::SystemTime};
 
 use camino::{Utf8Path, Utf8PathBuf};
 
@@ -31,6 +31,9 @@ pub(crate) struct ModuleLoader<'a, IO> {
     pub source_directory: &'a Utf8Path,
     pub artefact_directory: &'a Utf8Path,
     pub origin: Origin,
+    /// The set of modules that have had partial compilation done since the last
+    /// successful compilation.
+    pub incomplete_modules: &'a HashSet<EcoString>,
 }
 
 impl<'a, IO> ModuleLoader<'a, IO>
@@ -73,7 +76,7 @@ where
             if meta.fingerprint != SourceFingerprint::new(&source_module.code) {
                 tracing::debug!(?name, "cache_stale");
                 return Ok(Input::New(source_module));
-            } else if self.mode == Mode::Lsp {
+            } else if self.mode == Mode::Lsp && self.incomplete_modules.contains(&name) {
                 // Since the lsp can have valid but incorrect intermediate code states between
                 // successful compilations, we need to invalidate the cache even if the fingerprint matches
                 tracing::debug!(?name, "cache_stale for lsp");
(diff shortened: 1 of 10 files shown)


compiler-core/src/build/module_loader.rs
```diff
@@ -73,6 +73,11 @@ where
             if meta.fingerprint != SourceFingerprint::new(&source_module.code) {
                 tracing::debug!(?name, "cache_stale");
                 return Ok(Input::New(source_module));
+            } else if self.mode == Mode::Lsp {
```
lpil, line 76: Do you think we only want to do this in the instance that we are in LSP mode and the previous cache was for an invalid state? To avoid things like changing git branches performing excess compilation?

Likely not the most impactful optimisation today seeing as Gleam codebases are not huge, and this only happens when lots of files are changed, but it might be simple to add. I'm not sure myself.

lpil (commented):
Nice nice. Left a question inline.

compiler-core/src/build/module_loader.rs
Acepie: I think it would be doable to explicitly store "these modules had partial compilation occur so ignore the fingerprint" rather than always ignoring the fingerprint for the lsp. Just need to wire up some plumbing to store/pass that info down and reset it on successful compile

Acepie: Think I've implemented something that does what we want for this. lmk what you think

compiler-core/src/build/project_compiler.rs
```diff
@@ -562,13 +566,31 @@ where
...
+        let outcome = match outcome {
+            Outcome::Ok(modules) => {
+                // On successful compilation we remove the module from the incomplete set
+                modules.iter().for_each(|m| {
+                    let _ = self.incomplete_modules.remove(&m.name);
+                });
+                Outcome::Ok(modules)
+            }
+            Outcome::PartialFailure(modules, errors) => {
+                // On partial compilation failure we add all the module to the incomplete set
```
lpil, line 586: It's not just these that are incomplete, this also contains modules that have compiles successfully. If we're going to track this we need to make it accurate.

This data structure is being manipulated within the compiler, but then for this part we do it here. Perhaps it should be done inside instead? Though there's not a clear point where it has information about which is the one that failed and as such is incomplete.

Acepie: Good point, yeah I can move the insert to at the point of failure instead (I think the set itself still needs to be stored here so as to make sure the lsp maintains it/doesn't throw it out but the inserts/clears can definitely be moved)

lpil (approved):
Oh neat! That ended up being much cleaner than I expected. Thank you

PR #3177, opened 2024-05-22
fixed end location of tuple annotations

### LINKED ISSUE #3176: Hover not working for types inside tuples
If I hover on `Model` my editor says it could not find the type information. It works if I hover on the tuple itself.

```gleam
fn init(_flags: Nil) -> #(Model, Effect(e)) {
  todo
}
```

Acepie:
Closes #3176 

The end location for a tuple type was previously set to the end of the opening left paren instead of the closing right paren. This meant that hovering inside a tuple type was not properly being captured but hovering on the opening paren was.

### DIFF
--- a/compiler-core/src/language_server/tests/hover.rs
+++ b/compiler-core/src/language_server/tests/hover.rs
@@ -634,6 +634,27 @@ fn append(x: String, y: String) -> String {
 }
 
 #[test]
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you! And hey look, the existing tests caught it and I clearly ignored it 😅

PR #3187, opened 2024-05-24
RC - fix  wrong warning for redundant matches

giacomocavalieri:
This PR fixes a bug in the RC where the compiler would raise a warning for matches on literal values that are used just for their guards. Something like this will no longer raise a warning:

```gleam
case True {
  _ if condition -> todo
  _ if other -> todo
  _ -> todo
}
```


lpil (approved):
Fabulous!!

PR #3189, opened 2024-05-24
Add failing test for trailing comment in bit arrays

adamroyle:
In my use case, commenting out a trailing value moved it out of the bit array which was very confusing.


lpil:
Hello! Please make a new test rather than additional assertions in an existing test. We like to keep each test as narrow as possible. Thank you.

Is there an issue for this bug?

adamroyle:
Hi! There is no associated issue. I have moved into a separate test.

adamroyle:
Sorry! I'll create an issue with it instead.

PR #3207, opened 2024-05-28
Fix Wrong Error Message on Declaring Names Having Capital Letters

### LINKED ISSUE #3204: Wrong Assignment Error Message on Bad Names
When defining a new Variable starting with a capital letter, it throws a wrong error message as in the picture. If not fixed yet, I have a possible solution by defining a new ```TypeError``` and implement its message in ```to_diagnostics()```. Please, inform me if my solution is proper.

Screenshot 2024-05-27 143019

Arian94:
This PR closes #3204.


lpil (commented):
Thank you.

Rather than make a new error variant let's add a "situation" enum to the existing error, much like how we do for unification errors. We also want to only show this extra error information when it is the specific case in the issue, not always.

lpil:
Here's the pattern being used for the unification errors: https://github.com/gleam-lang/gleam/blob/c2a6660134db8b72f4b398df7a82e5a9e393d25d/compiler-core/src/type_/error.rs#L123-L129

We can apply this same one

Arian94:
I made some changes. Please take a look.

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 
 ### Bug Fixes
 
+- Fixed a bug where a wrong error message would be thrown on
+  naming variables starting with uppercase.
+  ([Arian Pourarian](https://github.com/Arian94))
```
lpil, line 17: This isn't a bug, the previous error message was correct. We are adding additional detail to it here.

compiler-core/src/type_/error.rs
```diff
@@ -789,6 +791,20 @@ pub enum UnknownValueConstructorError {
     },
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum UnknownValueConstructorErrorSituation {
+    /// A value using ```let``` is assigned to a variable having a capital case.
+    UnknownConstructorName,
```
lpil, line 797: This name doesn't match the description or the error produced here. Something to do with patterns perhaps?

compiler-core/src/type_/error.rs
```diff
@@ -789,6 +791,20 @@ pub enum UnknownValueConstructorError {
...
+}
+
+impl UnknownValueConstructorErrorSituation {
+    pub fn description(&self) -> Option<&'static str> {
```
lpil, line 801: What's this for? I couldn't see it used anywhere

compiler-core/src/type_/environment.rs
```diff
@@ -401,6 +401,7 @@ impl<'a> Environment<'a> {
                     name: name.clone(),
                     variables: self.local_value_names(),
                     type_with_name_in_scope,
+                    situation: Some(UnknownValueConstructorErrorSituation::UnknownConstructorName),
```
lpil, line 404: This is being added _always_ when trying to get a value constructor from the environment, but this ticket is for only when it is in a pattern. This needs to be moved to the pattern analysis code instead.

lpil (commented):
Nice. I've left some notes inline. Could you also add some more tests, such as when the constructor is nested within some other data structure like a tuple. Thanks

CHANGELOG.md
Arian94: Does it mean it's a feature? Or it is not worth mentioning it in the CHANGELOG file?

compiler-core/src/type_/error.rs
Arian94: To make the coding pattern similar to the Unification Errors, I just implemented a rather raw function. Could be removed if it's not necessary.

CHANGELOG.md
lpil: It's a feature, adding new detail to an error message.

compiler-core/src/type_/error.rs
lpil: If it's not used then don't include it 👍

Arian94:
I made new changes.

I have some confusion in my understanding. Why do you see this as a constructor pattern issue? I see it as an assignment issue.
Is it how it is designed? I mean, do you use the constructor pattern using the ```let``` keyword?

compiler-core/src/type_/error.rs
```diff
@@ -789,6 +791,12 @@ pub enum UnknownValueConstructorError {
     },
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum UnknownValueConstructorErrorSituation {
+    /// A value using ```let``` is assigned to a variable having a capital case.
+    NotFoundPattern,
```
lpil, line 797: The name of this variant doesn't match its meaning or the documentation above. Let's give a descriptive and accurate name.

compiler-core/src/error.rs
```diff
@@ -1843,12 +1843,22 @@ constructing a new record with its values."
                     variables,
                     name,
                     type_with_name_in_scope,
+                    situation
                 } => {
                     let text = if *type_with_name_in_scope {
                         wrap_format!("`{name}` is a type, it cannot be used as a value.")
                     } else {
-                        wrap_format!("The name `{name}` is not in scope here.")
+                        let mut situation_text = wrap_format!("The name `{name}` is not in scope here.");
```
lpil, line 1851: We now have 2 fields that describe the situation, that being `type_with_name_in_scope` and `situation`. Let's merge them together by making `type_with_name_in_scope` a variant of the `situation` enum.

compiler-core/src/type_/environment.rs
```diff
@@ -393,6 +393,7 @@ impl<'a> Environment<'a> {
         &mut self,
         module: Option<&EcoString>,
         name: &EcoString,
+        situation: Option<UnknownValueConstructorErrorSituation>,
```
lpil, line 396: Rather than pass in a situation let's have it default to non and then the calling code can optionally add it using `map_err`, the same as it does for the unify errors.

lpil (commented):
Thank you. Nearly there!

Arian94:
@lpil I made new changes. Have a look. Thx.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Compiler
 
+- Added more detail on the constructor error
+  where a variable starting with an uppercase is defined by let binding.
+  ([Arian Pourarian](https://github.com/Arian94))
```
lpil, line 11: This isn't right. The error message is giving more detail for when a pattern references a value constructor that does not exist.

compiler-core/src/error.rs
```diff
@@ -1842,13 +1842,18 @@ constructing a new record with its values."
...
-                    } else {
-                        wrap_format!("The name `{name}` is not in scope here.")
+                    let text = match situation {
+                        Some(UnknownValueConstructorErrorSituation::TypeWithNameIsInScope) => {
+                            wrap_format!("`{name}` is a type, it cannot be used as a value.")
+                        }
+                        Some(UnknownValueConstructorErrorSituation::NameStartsWithCapitalCase) => {
+                            wrap_format!("The name `{name}` is not in scope here.\n\nThis pattern is matching a constructor named `{name}`, but there is no constructor in scope with that name.")
+                        }
+                        _ =>  wrap_format!("The name `{name}` is not in scope here.")
```
lpil, line 1854: We don't use discard patterns ever as they disable exhaustiveness checking.
```suggestion
                        None =>  wrap_format!("The name `{name}` is not in scope here.")
```

compiler-core/src/type_/pattern.rs
```diff
@@ -425,9 +432,27 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
...
+                                name: name.clone(),
+                                variables: map_error_variables.clone(),
+                                situation: Some(
+                                    UnknownValueConstructorErrorSituation::NameStartsWithCapitalCase,
+                                ),
+                            }
+                        } else {
+                            e
+                        }
+                    })
```
lpil, line 455: Extract this into a helper function like we do with unify errors please, and use pattern matching rather than equality. The equality approach here does a lot of cloning and so is slower.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_capital_case_tuple_variable.snap
```diff
@@ -0,0 +1,14 @@
...
+error: Unknown variable
+  ┌─ /src/one/two.gleam:1:39
+  │
+1 │ let triple = #(1, 2.2, "three") let #(One, _, _) = triple
+  │                                       ^^^
+
+The name `One` is not in scope here.
+
+This pattern is matching a constructor named `One`, but there is no
+constructor in scope with that name.
```
lpil, line 14: This error message says the same thing twice. Is that an improvement on what it said before?

compiler-core/src/type_/pattern.rs
Arian94: I shifted this into the ```convert_get_value_constructor_error``` function. I'd think it's fine and safe to do so.

Arian94:
I suppose this should be the last commit.

PR #3217, opened 2024-05-30
Fix list append error message

giacomocavalieri:
This PR fixes a bug where the compiler would output a confusing error message if a spread was used to append to a list like this: `[..xs, 1]`. The error would complain about the spread being useless instead of saying it can't be used to append to a list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ### Bug Fixes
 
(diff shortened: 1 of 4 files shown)


lpil (commented):
Could you rebase on main please? The changelog is out of date. Thanks!

lpil (approved):
Thank you

PR #3219, opened 2024-05-30
Complete importable module values and types

### LINKED ISSUE #2757: LSP: add new import via autocompletion (after #2756)

Acepie:
Closes #2757

Initial support for completable values and types from modules not already imported. It uses the existing mechanisms from import completion to determine which modules should be importable and gets the values/types from those. the import statement is added above the first import of the file (or the top of the file if it is the first import). LMK if that is a reasonable choice or if we should try to find a better line to add the import to. Since the formatter will auto sort the import to the right spot I figured this was a reasonable behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,6 @@
   reserved word in Erlang/OTP 27.
   ([Jake Barszcz](https://github.com/barszcz))
 
-
 ### Formatter
 
 ### Language Server
@@ -53,6 +52,9 @@
   ```
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- LSP can now suggest completions for values and types from importable modules and adds the import to the top of the file.
+  ([Ameen Radwan](https://github.com/Acepie)
 
 ### Bug Fixes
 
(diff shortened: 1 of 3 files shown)


lpil (commented):
This is so cool! Thank you!

I've been trying this out and I've hit on some small issues relating to multiple modules can have the same last segment of their name.

It can be hard to tell the difference between these modules as the full name isn't displayed anywhere.

If there already is a module with that name it'll result in a compile error when imported due to the name collision.

Do you have thoughts on how we should resolve these? Perhaps the module name could be included at the top of the description.

Acepie:
Hmmm yeah that's fair. Improving the text display to include the full module and relying on the user to change the import might be a good enough solution for now? Cause the only alternative we could do would be generating an as for the user but idk if we should try to be smart about that (like if it made `import mything as mything2` that would work but I'd be annoyed personally)

Acepie:
I'll start with adding the full module name in the details when I get time tonight

lpil:
That sounds good. I'll make an issue for the collision issue

lpil:
Actually I'll add the module name too! I'm too excited to wait 😁

lpil (approved):
Thank you!!!

PR #3220, opened 2024-05-30
Fix bug where the formatter would strip newlines of anonymous function

### LINKED ISSUE #3210: Formatter strips new lines in lambda expressions
When I try and write the following code with a new line between the `process.send` call and the tuple, the formatter always removes the whitespace:

```gleam
  use _, connection, message <- mist.websocket(
    request: req,
    on_init: fn(conn) {
      process.send(
        runtime,
        server_component.subscribe("ui", fn(patch) {
          patch
          |> server_component.encode_patch
          |> json.to_string
          |> mist.send_text_frame(conn, _)
          Nil
        }),
      )
      #(Nil, None)
    },
    on_close: fn(_) { process.send(runtime, lustre.shutdown()) },
  )
```

I think this is a regression

giacomocavalieri:
This PR closes #3210

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@
   trying to use the spread syntax to append to a list.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the formatter would strip empty lines out of the body of an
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!

PR #3222, opened 2024-05-31
Warn for redundant capture

### LINKED ISSUE #2999: Add diagnostic for unnecessary function capture
Thanks for all the work that has gone into the language and the diagnostics, they truly are amazing!

While going through the tutorial, it occurred to me after I wrote this and then removed the `(_)` that the function capture was completely unnecessary, but with no warning I'm sure this error exists in a ton of codebases already, and could be easily dealt with with a diagnostic added to the compiler.

```gleam
  1
  |> int.to_string(_) // <-- (_) is unnecessary and can be removed without changing the meaning of the pipeline
  |> string.repeat(5)
  |> string.drop_left(2)
  |> io.println
```

giacomocavalieri:
This PR closes #2999

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@
 
 ### Compiler
 
+- The compiler now emits a warning for redundant function captures in a
+  pipeline:
+
+  ```
+  warning: Redundant function capture
+    ┌─ /src/warning/wrn.gleam:5:17
+    │
+  5 │     1 |> wibble(_, 2) |> wibble(2)
+    │                 ^ You can safely remove this
+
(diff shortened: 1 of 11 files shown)


lpil:
Linter is failing!

giacomocavalieri:
Oops, should be all good now!

lpil (approved):
Nice! Thank you

nicklimmm:
The current implementation outputs false positives for `_` not in the first argument.

A reproducible example:

image

giacomocavalieri:
Oops my bad, I'll fix it!

PR #3226, opened 2024-05-31
Fix 3108

### LINKED ISSUE #3108: List pattern matching
There's a strange pattern that I don't think should compile when pattern matching on a list:
```gleam
case list {
  [first..rest] -> todo
//^^^^^^^^^^^^^ This compiles as [first, ..rest]
  _ -> todo
}
```
I think it should be a parse error instead

giacomocavalieri:
This PR closes #3108


lpil (commented):
Nice. Could we emit this warning during parsing rather than storing this information in the AST?

giacomocavalieri:
Oh yes! I think it will be quite a big change because we don't have a warning emitter during parsing but it's totally doable, I'll change it

lpil:
Can we use just a vec? I think I want to remove the warning emitter now that we have an errors vec

giacomocavalieri:
I'm not sure I understand how that would work, what do you mean by just using a vec?

lpil:
As in rather than pass in that emitter class pass in a vec and have it append to the vec.

CHANGELOG.md
```diff
@@ -28,6 +28,23 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler will now raise a warning when using deprecated syntax to
+  prepend to a list:
```
lpil, line 32:
```suggestion
- The syntax `[a..b]` is now deprecated in favour of the `[a, ..b]` syntax.
  This was to avoid it being mistaken for a range syntax, which Gleam does
  not have.
```

CHANGELOG.md
```diff
@@ -49,6 +67,24 @@
...
+    │
+  5 │         ["a"..rest] -> rest
+    │             ^^ This spread should be preceded by a comma
+
+  This syntax for pattern matching on a list is deprecated.
+  When matching on the rest of a list it should always be preceded by a
+  comma, like this: `[item, ..list]`.
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
```
lpil, line 87: I think maybe a rebase went wrong here? It's in here twice!

compiler-core/src/parse.rs
```diff
@@ -127,11 +128,16 @@ impl Attributes {
 //
 // Public Interface
 //
-pub fn parse_module(src: &str) -> Result<Parsed, ParseError> {
+pub fn parse_module(src: &str, warnings: &TypeWarningEmitter) -> Result<Parsed, ParseError> {
```
lpil, line 131: I don't think the parser should be emitting type warnings!

compiler-core/src/parse.rs
```diff
@@ -2819,12 +2842,27 @@ where
         parser: &impl Fn(&mut Self) -> Result<Option<A>, ParseError>,
         sep: Option<&Token>,
     ) -> Result<Vec<A>, ParseError> {
+        let (_, res) = self.sep_series_of(parser, sep)?;
+        Ok(res)
+    }
+
+    // Parse a series by repeating a parser, and a separator. Returns true if
+    // the series ends with the trailing separator.
+    fn sep_series_of<A>(
```
lpil, line 2851: What's a sep? No abbreviations please 🙏 

Also `///` for documentation comments.

lpil (commented):
Lovely! thank you. I've left a few small notes

lpil (approved):
Beautiful! Thank you very much!

Issue #3233, opened 2024-06-02
Certain identifiers named `Maybe` generate invalid Erlang code on Erlang/OTP 27


barszcz:
Gleam version: 1.2.1
OS: macOS Monterey (12.4)
Runtime: BEAM (Erlang/OTP 27)

As of Erlang/OTP 27, [`maybe` is now a reserved word](https://www.erlang.org/doc/system/expressions.html#maybe). This means that some (but not all) generated Erlang code attempting to use `maybe` as an identifier no longer compiles. Some example code follows:

```gleam
import gleam/io

// custom type named Maybe doesn't compile
pub type Maybe {
  // variant named Maybe doesn't compile
  Maybe
}

// module level function named maybe doesn't compile
pub fn maybe() {
  io.println("oops")
  "oops"
}

// module level const named maybe compiles
const maybe = "here's my number, so call me maybe"

pub fn main() {
  // local variable binding named maybe compiles
  let maybe = "here's my number, so call me maybe"
  io.println(maybe)
}
```

As I mentioned in the Gleam Discord, I discovered this while working through the Exercism track – [the "Valentine's Day" exercise](https://exercism.org/tracks/gleam/exercises/valentines-day) asks you to create a custom type variant named `Maybe` so newcomers to Gleam are especially likely to get bitten by this. Here's a trace:
<details>
<summary>long log</summary>

```
DEBUG download_deps: locking_build_directory path="/Users/jake/dev/maybe_bug_example/build/packages"
TRACE download_deps: reading_file path="/Users/jake/dev/maybe_bug_example/gleam.toml"
DEBUG download_deps: reading_manifest_toml
TRACE download_deps: reading_file path="/Users/jake/dev/maybe_bug_example/manifest.toml"
DEBUG download_deps: manifest_up_to_date
TRACE download_deps: reading_file path="/Users/jake/dev/maybe_bug_example/build/packages/packages.toml"
DEBUG download_deps: locking_build_directory path="/Users/jake/dev/maybe_bug_example/build/dev/erlang"
DEBUG download_deps: locking_build_directory path="/Users/jake/dev/maybe_bug_example/build/dev/javascript"
DEBUG download_deps: locking_build_directory path="/Users/jake/dev/maybe_bug_example/build/prod/erlang"
DEBUG download_deps: locking_build_directory path="/Users/jake/dev/maybe_bug_example/build/prod/javascript"
[...]

PR #3234, opened 2024-06-02
Add `maybe` as a reserved word for Erlang compile target

barszcz:
`maybe` is reserved as of Erlang/OTP 27.
See https://www.erlang.org/doc/system/expressions.html#maybe for more info.

Let me know if there are any other tests I should add? I'm not super familiar with Rust, but this seemed like a simple enough change.

Closes https://github.com/gleam-lang/gleam/issues/3233,

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,11 @@
   anonymous function passed as an argument.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the compiler would generate invalid Erlang code when
+  declaring an identifier named `Maybe`/`maybe` and compiling against
+  Erlang/OTP 27, where `maybe` is now a reserved word.
+  ([Jake Barszcz](https://github.com/barszcz))
+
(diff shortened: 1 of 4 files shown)


lpil (commented):
Nice! You'll need to run cargo-insta to update the snapshot. Could you also update the changelog too? Thanks

barszcz:
@lpil done and done!

CHANGELOG.md
```diff
@@ -38,6 +38,11 @@
   anonymous function passed as an argument.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the compiler would generate invalid Erlang code when
+  declaring an identifier named `Maybe`/`maybe` and compiling against
+  Erlang/OTP 27, where `maybe` is now a reserved word.
+  ([Jake Barszcz](https://github.com/barszcz))
```
lpil, line 44: This is not a bug, it is that we now have support for OTP27.

lpil (approved):
Thank you!!

Issue #3175, opened 2024-05-22
Helpful error message for trying to pattern match on the end of a list


lpil:
Explain why this isn't a thing!
```gleam
  case somelist {
    [rest.., last] -> 1
    _ -> 2
  }
```

PR #3235, opened 2024-06-02
Add error message for trying to match after a spread

Eingin:
This closes: #3175 

In the case where there is a tail after parsing the a list pattern but the list isn't closed we create a different error. This specifies that the spread operator consumes the remaining elements of the list.

To facilitate displaying this error I lightly refactored the `next_tok_unexpected ` method into a more generic `next_tok_error` that `next_tok_unexpected` now uses but also allows any error to be specified for the token.

I have also added a test case for both error scenarios.

There might be better copy for the error so let me know what you think!


compiler-core/src/parse/tests.rs
```diff
@@ -1002,6 +1002,34 @@ fn main() {
...
+        "
+fn main() {
+    case [1, 2, 3] {
+        [1, 2, *] -> 0
+        _ -> 0
+    }
+}
+"
+    );
+}
```
lpil, line 1031: Remove this test please, it's unrelated to this work 🙏

compiler-core/src/parse/tests.rs
```diff
@@ -1002,6 +1002,34 @@ fn main() {
     );
 }
 
+#[test]
+fn case_list_pattern_after_spread() {
+    assert_module_error!(
+        "
+fn main() {
+    case [1, 2, 3] {
+        [1, .., last] -> 0
```
lpil, line 1011: This isn't the syntax from the issue, could you make it match please 🙏

compiler-core/src/parse/tests.rs
```diff
@@ -1002,6 +1002,34 @@ fn main() {
     );
 }
 
+#[test]
+fn case_list_pattern_after_spread() {
+    assert_module_error!(
+        "
+fn main() {
+    case [1, 2, 3] {
```
lpil, line 1010: let's remove the case expression as we're not testing anything to do with case expressions here.

compiler-core/src/parse/error.rs
```diff
@@ -232,6 +232,14 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                     "See: https://tour.gleam.run/flow-control/case-expressions/".into(),
                 ],
             ),
+            ParseErrorType::ListPatternAfterSpread => (
+            "Invalid pattern", vec![
+                "I was expecting the end of the list.".into(),
+                "The spread pattern precesing this consumes the remaining elements of the list".into(),
+                "therefore there is nothing left to match.".into()
```
lpil, line 239: Let's have the message explain why it is that you cannot append to a list.

lpil (commented):

compiler-core/src/parse/error.rs
Eingin: I've added the explanation from `ListSpreadFollowedByElements` let me know if you anything more specific in mind.

compiler-core/src/parse.rs
```diff
@@ -1163,8 +1163,21 @@ where
...
                 };
-                let (end, rsqb_e) =
-                    self.expect_one_following_series(&Token::RightSquare, "a pattern")?;
+
+                let (end, rsqb_e) = match self.maybe_one(&Token::RightSquare) {
+                    Some((start, end)) => Ok((start, end)),
+                    None => {
+                        if tail.is_some() {
+                            self.next_tok_error(ParseErrorType::ListPatternAfterSpread)
+                        } else {
```
giacomocavalieri, line 1172: Just doing this would produce confusing errors when another more appropriate syntax error would be best. Imagine you write something like this:
```gleam
case list {
  [one, two, ..rest -> todo
}
```
Now this new error would pop up underlining the arrow and that would be quite confusing.
```
case list {
  [one, two, ..rest -> todo
                    ^^ 

I was expecting the end of the list.
The spread pattern precesing this consumes the remaining elements of the list
therefore there is nothing left to match.
```
The best approach here would be to copy what the parser is doing when parsing a list value: there we try and parse another list of items and see if it succeeds or not. So here we'd have to do something similar and parse a list of patterns and see if it succeeds (then we can show this error) or not (all good).

CHANGELOG.md
```diff
@@ -6,6 +6,23 @@
...
 ### Compiler
 
+- Added more a verbose error when a pattern is attempted to be used after a spread.
+
+```
+error: Syntax error
+  ┌─ /src/parse/error.gleam:4:17
+  │
+4 │         [rest.., last] -> 1
+  │                  ^^^^ Invalid pattern
```
giacomocavalieri, line 16: Technically the highlighted bit _is_ a valid pattern. I think it would be nicer if the message was similar to the same you get for trying to append to a list with the spread operator.

When you try to use it to append it looks like this:
```
error: Syntax error
  ┌─ /src/main.gleam:4:10
  │
4 │   [1, ..[1, 2, 3], 3]
  │       ^^^^^^^^^^^ I wasn't expecting elements after this spread

A spread can only be used to prepend elements to lists like this: `[first, ..rest]`.
```

So here we might give something like this:
```
error: Syntax error
  ┌─ /src/main.gleam:4:10
  │
4 │   [1, ..xs, last]
  │       ^^^^ I wasn't expecting other patterns after this spread

A spread can only be used to match on the entire end of a list.
It is not possible to extract items from the end of a list using pattern
matching because that would require walking through the entire list.
```

CHANGELOG.md
```diff
@@ -6,6 +6,23 @@
...
 ### Compiler
 
+- Added more a verbose error when a pattern is attempted to be used after a spread.
+
+```
+error: Syntax error
+  ┌─ /src/parse/error.gleam:4:14
+  │
+4 │         [rest.., last] -> 1
+  │              ^^ I wasn't expecting other patterns after this spread
```
giacomocavalieri, line 16: I think this should be:
```
[..rest, last]
 ^^^^^^ I wasn't expecting other patterns after this spread
```

compiler-core/src/parse.rs
```diff
@@ -1155,16 +1155,48 @@ where
...
+                let tail = if let Some((start, Token::DotDot, end)) = self.tok0 {
+                    dot_dot_location = Some((start, end));
                     self.advance();
                     let pat = self.parse_pattern()?;
-                    let _ = self.maybe_one(&Token::Comma);
+                    if self.maybe_one(&Token::Comma).is_some() {
+                        // See if there's a list of items after the tail,
+                        // like `[..wibble, wobble, wabble]`
+                        let elements =
+                            Parser::series_of(self, &Parser::parse_expression, Some(&Token::Comma));
```
giacomocavalieri, line 1168: I think this should parse a `series_of` patterns and not `expression`

compiler-core/src/parse/tests.rs
```diff
@@ -1002,6 +1002,20 @@ fn main() {
     );
 }
 
+#[test]
+fn case_list_pattern_after_spread() {
+    assert_module_error!(
+        "
+fn main() {
+    case somelist {
+        [rest.., last] -> 1
```
giacomocavalieri, line 1011: This should be `[..rest, last]`

compiler-core/src/parse.rs
```diff
@@ -2879,20 +2911,21 @@ where
     }
 
     // Error on the next token or EOF
-    fn next_tok_unexpected<A>(&mut self, expected: Vec<EcoString>) -> Result<A, ParseError> {
+    fn next_tok_error<A>(&mut self, error: ParseErrorType) -> Result<A, ParseError> {
```
giacomocavalieri, line 2914: I think we can remove this, it doesn't look like it's used anymore

giacomocavalieri (commented):
Well done!! There's still a couple of things I'd change but it's starting to look great 💕

giacomocavalieri (approved):
This looks great to me!

lpil (approved):
Lovely work! Thank you very much! And also thank you for waiting as I slowly get through all my notifications.

PR #3236, opened 2024-06-02
Enable remove redundant tuple action with catch all patterns

giacomocavalieri:
Previously the LSP wouldn't suggest the "Remove redundant tuple" action in a case match that had a catch all ignored pattern. Now it is enable for those cases as well:
```gleam
case #(a, b, c) {
  #(1, 2, 3) -> todo
  _ -> todo
}
```
Becomes:
```gleam
case a, b, c {
  1, 2, 3 -> todo
  _, _, _ -> todo
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,27 @@
 
 ### Language Server
 
+- The language server will now suggest the "Remove redundant tuple" action even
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you!!

PR #3239, opened 2024-06-02
Fix bug with redundant capture warning

giacomocavalieri:
This PR fixes a bug I didn't notice in a PR that got recently fixed. Now the compiler won't complain about _all_ function holes but only those that are truly redundant


lpil (approved):
Thank you

Issue #3215, opened 2024-05-30
Fix autocompletion of unqualified types in imports sometimes duplicating the `type` keyword.


lpil:
```gleam
import wibble.{type ][}
```
Autocompleting a type here can result in this
```gleam
import wibble.{type type Wobble}
```

Seems to happen in all editors

Related to https://github.com/gleam-lang/gleam/issues/2165

PR #3240, opened 2024-06-03
Completions provide by the LSP use the text_edit API

### LINKED ISSUE #2165: Fix VS Code duplicating module name when autocompleting module functions
When testing autocomplete with helix (helix-editor.com), I get some odd behavior:

Entering 

``string.re`` doesn't give me an autocompletion but a live error message ``the module gleam/string does not have a re field.

When completing ``string.repeat``, I get offered an autocompletion, which - when accepted, expands to ``string.string.repeat``.

When not accepting, but deleting characters back to ``string.rep``, I get an autocompletion offered for ``string.replace``, ``string.repeat``, ... (including ``string.is_empty``, which doesn't look like a good match).

However the gist of it is that I need to first completely type a valid field in a module in order to make autocompletion yield anything on that line of code.

Acepie:
Closes #2165 #3270 and #3215 (as far as I can tell from personal testing at least)

Completions now use the text edit API to provide suggestions. This is done by looking up the phrase surrounding the cursor and making an edit from the start of that phrase to the end of that phrase. The phrase includes any name/upname valid characters as well as the dot character. In the special case of unqualified imports, the phrase is instead any name/upname as well as whitespace (not including new line). This is to allow for `type Upname` to be treated as 1 phrase. I could see us having more such cases but for now those are the only 2 I could see.

This pr was a draft because it builds on top of https://github.com/gleam-lang/gleam/pull/3219 and I think that one should probably be merged first. Originally I had started writing a mechanism to filter the importable modules options based on the last phrase and realized that the feature was ultimately too tied to this change and figured I'd separate it out instead.


compiler-core/src/language_server/engine.rs
```diff
@@ -198,10 +198,12 @@ where
             // Check current filercontents if the user is writing an import
             // and handle separately from the rest of the completion flow
             // Check if an import is being written
-            if let Some(value) = this.import_completions(src, &params, module) {
+            let src_line_numbers = LineNumbers::new(src.as_str());
```
lpil, line 201: Sorry, I'm confused- what's `src` here and why is it different from `module.code`?

compiler-core/src/language_server/engine.rs
```diff
@@ -433,15 +452,17 @@ where
         }
 
         // Module types
-        for (name, type_) in &module.ast.type_info.types {
-            completions.push(type_completion(None, name, type_));
+        // Do not complete direct module types if the user has already started typing a module select.
```
lpil, line 455: Could you expand this comment to explain what it means and what would happen if we don't do this. I don't understand what it means.

compiler-core/src/language_server/engine.rs
```diff
@@ -452,43 +473,69 @@ where
                     continue;
                 }
 
-                let module = import.used_name();
-                if module.is_some() {
-                    completions.push(type_completion(module.as_ref(), name, type_));
+                if let Some(module) = import.used_name() {
+                    // If the user has already started a module select then don't show irrelevant modules.
```
lpil, line 477: Could you add more detail to this, such as examples.

compiler-core/src/language_server/engine.rs
```diff
@@ -452,43 +473,69 @@ where
...
+                    if let Some(input_mod_name) = &module_select {
+                        if !module.contains(input_mod_name.as_str()) {
```
lpil, line 479: Contains rather than equals?

compiler-core/src/language_server/engine.rs
```diff
@@ -791,9 +936,11 @@ where
 }
 
 fn type_completion(
-    module: Option<&EcoString>,
+    module: Option<&str>,
     name: &str,
     type_: &TypeConstructor,
+    insert_range: lsp::Range,
+    include_type_in_completion: bool,
```
lpil, line 943: Let's always use an enum over a bool as it's not clear what `false` means outside of context

lpil (commented):
Awesome!! Thank you

I've left a few comments but I think it would be great if we could find a way to avoid the extra nesting and flow control that's scattered around the code now. It's getting harder to understand and there's a lot more logic mixed in than there was previously. Do you have any ideas here? I've not dug into it myself.

Can we also use snapshots for the new tests? Example tests have a very large maintenance burden that I would like to avoid.

Acepie:
I'll swap the tests over to snapshots. As for the logic, I thought about refactoring it since its a bit messy and mostly duplicated but every time I tried it always ended up harder to understand than the original. I'll try one more time to see if I can think of something better but maybe it just needs to be broken out into a sub module at this point?

compiler-core/src/language_server/engine.rs
Acepie: src was the raw buffer content vs module.code being the buffer content based on the last compile. For finding the current cursor point and phrase around the cursor, using the direct src seems more correct since we don't want there to be weird completion location drift if the code is left in a noncompiling state for too long

compiler-core/src/language_server/engine.rs
Acepie: I used contains here to support typing partial pieces of the module name which I could see people doing a lot (e.x. if I had an `unordered_set` module a user could type `set.` and still get relevant results). lmk what you think I can just as easily switch it to equals

Acepie:
Alright snapshots and comments are added. LMK what you think. I'll try to spend a little bit of time seeing if I can refactor the logic a bit in a way that doesn't feel too overly complicated but honestly, I might just need somebody with a better understanding of rust patterns to help with this cause last time I tried I ended up with something truly unreadable

compiler-core/src/language_server/engine.rs
lpil: Is it possible for them to be different now that we have fault tolerant compilation?

compiler-core/src/language_server/engine.rs
lpil: Let's try equality and we can relax it later if it feels bad

compiler-core/src/language_server/engine.rs
Acepie: I think there are still edge cases where it is possible mainly in cases where the currently typed thing fails to parse/lex which will happen a lot for completions where the user just types the dot

```
import gleam/io

pub fn main() {
  io.
}
```

lpil (approved):
You're a superstar!! Thank you

PR #3247, opened 2024-06-06
Fix counting type params in Result type

### LINKED ISSUE #3221: Add qualifiers for type variables in generated Erlang typespecs
Starting from OTP 26 erlang/otp#6915, warnings are emitted for unbound type variables, for example:

```erlang
 -spec try_call_timeout_test() -> {ok, GGA}
```

We need to replace this unbound type variable with `term()` so that no warning is emitted.

*originally reported in gleam-lang/erlang#46*

dvic:
According to https://github.com/erlang/otp/issues/8533#issuecomment-2142485802 a disjunction does not count as being used twice:

```erlang
-spec wibble(atom()) -> {ok, X} | {error, X}.
```

This PR changes the type parameter usage counting for Result types: the counting is deduplicated for the disjunction.

Fixes #3221

I have manually checked and this fixes https://github.com/gleam-lang/erlang/pull/46#issuecomment-2133522860

image

vs gleam 1.2.0:

image


dvic:
The one CI failure is a timeout in `erlef/setup-beam`

Acepie:
That one happens sometimes. You can generally just push an empty commit to force a rebuild

lpil (commented):
Doesn't this break nested types in this case? As the ids are reset as each one is reached?

dvic:
> Doesn't this break nested types in this case? As the ids are reset as each one is reached?

I don't think this patch "resets" the ids, it just increments the counter by 1 instead of the number of time it occurs nested. This should be the correct behaviour if I can wrap my head around this 😅 But it might be wise to add some more sanity check test cases.

I added the following (https://github.com/gleam-lang/gleam/pull/3247/commits/c9eb6db1b83a4dadc7e61973f118c9305e9b4658), did you mean this kind of situation? The generated code is fine according to `erlc` (no warnings).

```gleam
        pub type Nested(a, b) {
            VariantX(field_a: a, field_b: b)
            VariantY(field_a: a, field_b: b)
        }

        pub type Parent(a, b) {
            ParentX(Nested(a,b))
            ParentY(Nested(a,b))
        }

        pub fn foo1() -> Parent(a, a) {
            todo
        }

        pub fn foo2() -> Parent(a, b) {
            todo
        }

        pub fn foo3(val_a: a) -> Parent(a, a) {
            todo
        }

        pub fn foo4(val_a: a, val_b: b) -> Parent(a, b) {
            todo
        }
```

The generated code:

```erlang
-module(my@mod).
-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).

-export([foo1/0, foo2/0, foo3/1, foo4/2]).
-export_type([nested/2, parent/2]).

-type nested(I, J) :: {variant_x, I, J} | {variant_y, I, J}.

-type parent(K, L) :: {parent_x, nested(K, L)} | {parent_y, nested(K, L)}.

-spec foo1() -> parent(any(), any()).
foo1() ->
    erlang:error(#{gleam_error => todo,
            message => <<"This has not yet been implemented"/utf8>>,
            module => <<"my/mod"/utf8>>,
            function => <<"foo1"/utf8>>,
            line => 13}).

-spec foo2() -> parent(any(), any()).
foo2() ->
    erlang:error(#{gleam_error => todo,
            message => <<"This has not yet been implemented"/utf8>>,
            module => <<"my/mod"/utf8>>,
            function => <<"foo2"/utf8>>,
            line => 17}).

-spec foo3(T) -> parent(T, T).
foo3(Val_a) ->
    erlang:error(#{gleam_error => todo,
            message => <<"This has not yet been implemented"/utf8>>,
            module => <<"my/mod"/utf8>>,
            function => <<"foo3"/utf8>>,
            line => 21}).

-spec foo4(W, X) -> parent(W, X).
foo4(Val_a, Val_b) ->
    erlang:error(#{gleam_error => todo,
            message => <<"This has not yet been implemented"/utf8>>,
            module => <<"my/mod"/utf8>>,
            function => <<"foo4"/utf8>>,
            line => 25}).
```

compiler-core/src/erlang/tests/type_params.rs
```diff
@@ -10,3 +10,36 @@ fn named_args_count_once() {
...
+        pub fn foo3(val_a: a) -> Parent(a, a) {
+            todo
+        }
+
+        pub fn foo4(val_a: a, val_b: b) -> Parent(a, b) {
+            todo
+        }
+        "
+    );
+}
```
lpil, line 45: This test does many things. Could you break it out into one test per example please? Thanks

I can't see any nested types that use the same type variables at different levels- I think that's where we may have the issue.

dvic: > This test does many things. Could you break it out into one test per example please? Thanks

Sure!


> I can't see any nested types that use the same type variables at different levels- I think that's where we may have the issue.

Can you give an example of such a type? Then I can see what is generated now vs before this PR.

lpil: Maybe something like `Wibble(a, Wibble(a, b))`?

dvic: Sure!

For types that have two or more variants, I've simplified the test cases to:
- `Result(a, a)` generates `{ok, any()} | {error, any()}`
- `Result(a, Result(a, b))` generates `{ok, any()} | {error, {ok, any()} | {error, any()}}`
- `#(a, Result(a, b))` generates `{I, {ok, I} | {error, any()}}`

I think this is correct?

However, for the case with a single variant, the current patch generates sub-optimal code:
- `Wibble(a, a)` generates `wibble(any(), any())`, while it technically could be `wibble(A, A)`
- `Wibble(a, Wibble(a, b))` generates `wibble(any(), wibble(any(), any()))`, while it could be `wibble(A, wibble(A, any()))`
- `#(a, Wibble(a, b))` generates `{A, wibble(A, any())}`

On the other hand, we've seen from the discussion on https://github.com/erlang/otp/issues/8533 that there's not really a way to express that the two a type variables are the same type, so not sure if we want to generate types like that?

In that case, this PR should take a different solution I think and take into account the `constructors` while counting the parameter occurrences ([here](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/erlang.rs#L265)).

dvic: Now that I think about it, shouldn’t uniquely counting across constructors be the only thing we need? (for custom types)

lpil (commented):
Could you remove all references to "foobar" please, it's a rather grim WW2 reference

dvic:
> Could you remove all references to "foobar" please, it's a rather grim WW2 reference

TIL, will do, thanks!

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__type_params__custom_type_nested_named_args_count_once.snap.new
```diff
@@ -0,0 +1,20 @@
...
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).
+
+-export([foo/0]).
+-export_type([wibble/2]).
+
+-type wibble(I, J) :: {wibble, I, J}.
+
+-spec foo() -> wibble(any(), wibble(any(), any())).
```
lpil, line 14: This should be `wibble(A, wibble(A, any()))`

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__type_params__custom_type_named_args_count_once.snap.new
```diff
@@ -0,0 +1,20 @@
...
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).
+
+-export([wibble/0]).
+-export_type([wibble/2]).
+
+-type wibble(I, J) :: {wibble, I, J}.
+
+-spec wibble() -> wibble(any(), any()).
```
lpil, line 14: Shouldn't this be `wibble(a, a)`?

dvic: yes, that’s still a unaccepted snapshot i think

dvic: Will try to tackle this PR tonight, I think I have the solution: counting of variables should be done by taking max of count for each constructor. Will get back to you if this works out.

dvic:
@lpil I thought this should be ready now. But I now see that the original example from gleam_erlang is not fixed by the current solution. Will look into that before marking the PR as ready again.

The approach I took is a bit simpler than what I first had planned: I deduplicate the counts only for Result types, as this is the only type that we directly print in functions (AFAIK) with the disjunction expanded. The test snapshot should be up-to-date.

dvic:
@lpil this is basically the case from the original issue:

https://github.com/gleam-lang/gleam/pull/3247/commits/40bdae6d14169d8de5c84f899c4bd2023eabcb43

I checked the type by forcing a type error:
image

Any idea why the test is passing but it's not working on the original project? I'm running `cargo build --release` and then `cp target/release/gleam /Users/dvic/Development/gleam_erlang` and then `./gleam build` (first I removed the build folder entirely).

dvic:
Ok, now it should be ready, apologies for the noise, I thought I was there 😅

lpil (approved):
Thank you!!!!

dvic:
> Ah! Could you update the changelog please.
> 
> Note I have rebased your branch onto main to remove the merge commits as we don't permit them.

Done! Feel free to adjust the changelog text if necessary, I was not sure how to describe this change 🙂

lpil (approved):
Thank you!!

PR #3255, opened 2024-06-09
Implement code action to ignore unused Result value

### LINKED ISSUE #2960: LSP code action: add `let _ =` to focused unused value

leah-u:
Closes #2960

My first contribution, please be gentle :3

I had to undo the changes from commit [64924b9](https://github.com/gleam-lang/gleam/commit/64924b949856253603300b47151d37d45d1c5fe7) as my tests would otherwise panic with an integer underflow. I'm not really sure why it was necessary in the first place, maybe @giacomocavalieri can help?

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -54,6 +54,7 @@ fn engine_response(src: &str, line: u32) -> engine::Response<Option<Vec<lsp_type
 
 const REMOVE_UNUSED_IMPORTS_TITLE: &str = "Remove unused imports";
 const REMOVE_REDUNDANT_TUPLES: &str = "Remove redundant tuples";
+const ASSIGN_UNUSED_RESULT: &str = "Assign unused Result value to `_`";
 
 fn apply_first_code_action_with_title(src: &str, line: u32, title: &str) -> String {
     let response = engine_response(src, line)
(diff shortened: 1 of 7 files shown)


compiler-core/src/language_server/tests/action.rs
```diff
@@ -91,13 +92,13 @@ fn apply_code_edit(
...
-                line_numbers.byte_index(edit.range.end.line, edit.range.end.character) - offset;
+            let start = line_numbers.byte_index(edit.range.start.line, edit.range.start.character)
+                as i32
+                - offset;
+            let end = line_numbers.byte_index(edit.range.end.line, edit.range.end.character) as i32
+                - offset;
             let range = (start as usize)..(end as usize);
-            offset += end - start;
-            offset -= edit.new_text.len() as u32;
+            offset += end - start - edit.new_text.len() as i32;
```
lpil, line 101: Revert this please, this old version has it has a bug in it 🙏

lpil (commented):
Thank you! Fab work here!

We can avoid traversing the AST as the location of the discarded result is contained by the warning that would have been emitted during compilation. There's 2 ways we could get this information:

1. Replace the warning emitter trait with just a vec of warnings that gets appended to. Then the language server can iterate over the vec once to see if there's any code actions it should offer. This is the best solution.
2. Have the analyser also record the location of unused warnings in the analysed module data and have the language server use this data. This would be a temporary solution, but we already do something like this for unused imports.

Either of these approached would be fab.

compiler-core/src/language_server/tests/action.rs
leah-u: The way it was before with `offset` being a `u32`, the subtraction on line 100 would panic with an integer underflow if the code action added more characters than it removed. I'm not sure how else to solve this. What exactly is the bug?

lpil: It would not apply the text edit correctly. The underflow likely means you have a bug and are trying to remove more content than exists.

lpil:
Are you still working on this @leah-u ? Thank you

leah-u:
Sorry, I kinda forgot about this. I might give it another shot some time, but I don't currently have any time to work on it.

PR #3260, opened 2024-06-11
Fix compiler crash when defining duplicate types

### LINKED ISSUE #3209: Compiler fatal crash when defining a type with the same name as an imported type.
```gleam
import wisp.{type Request}
type Request { Request }
```

Defining a duplicate type name like this causes a fatal compiler crash.

```
Panic: compiler-core/src/analyse.rs:991
        name uniqueness checked above: DuplicateTypeName { location: SrcSpan { start: 233, end: 245 }, previous_location: SrcSpan { start: 203, end: 215 }, name: "Request" }
Gleam version: 1.2.0-nightly-20240528
Operating system: linux
        ```

GearsDatapacks:
This is a possible solution to #3209, which moves the duplicate name check into `Environment`, meaning it has access to the imported types as well as locally defined types.

Also, should I add a test for this? If so, where is the best place to put that?

Closes #3209

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,10 @@
   anonymous function passed as an argument.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the compiler would crash when a type was defined with
+  the same name as an imported type.
+  ([Gears](https://github.com/gearsdatapacks))
+
 ## v1.2.1 - 2024-05-30
 
 ### Bug Fixes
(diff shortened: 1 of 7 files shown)


lpil (commented):
Thank you! We will need tests, aye. In the type test modules would be best, you can put one in each of the different type definition modules.

GearsDatapacks:
I've added a test for each of the type definition cases

lpil (approved):
Wonderful work! And thank you for waiting for me to review this, I have had a short holiday. 💜

Issue #2956, opened 2024-04-11
LSP: rename local variable


lpil:
Ideally we implement this using a call graph and not by traversing the AST

nicklimmm:
I'd like to work on this. I had quite a hard time refactoring manually 😅 

Just wondering, why should we use the call graph instead of the AST? 

I believe we can still find all the nodes to rename while respecting scopes, since we can access the span of the definition from `TypedExpr`'s `ValueConstructorVariant::LocalVariable`.

nicklimmm:
> It's possible to implement by tree walking but implementing it with a call-graph would put us one step closer to renaming other things, avoiding future rework. I would happily accept a PR with either approach.

Got it. Do we have something that we can reuse from the compiler to get the call-graph? I noticed that we have something like `CallGraphNode` during the untyped stage, but I don't think we can do much with that.

nicklimmm:
I'll check around if we can pass that to the lsp engine without a lot of changes.

nicklimmm:
> During analysis we build a call graph, and I believe it does cover all expressions. We'd need to find a good way to be able to keep it around and use it later in the LSP.
> 
> https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/call_graph.rs

CMIIW here, but the call graph only traverses all expressions without adding edges for usages/references. Also, it is for determining the compilation order of functions and module constants.

If the above is true, I think building a usages-references graph from the typed AST or just collecting the necessary usages-references by walking the AST might be a better option here.

lpil:
It does add edges for references. If it didn't we wouldn't be able to order the nodes in dependency order.

https://github.com/gleam-lang/gleam/blob/f81f4066a237409a3d4774b6e193457613e9dc58/compiler-core/src/call_graph.rs#L172-L175

There's likely a bunch of missing information that we will need to store at this stage, but it does at least have that part.

nicklimmm:
> It does add edges for references.
> 
> https://github.com/gleam-lang/gleam/blob/f81f4066a237409a3d4774b6e193457613e9dc58/compiler-core/src/call_graph.rs#L172-L175
> 
> There will more more information that we will need to store at this stage, that'll be a good portion of the work here.
> 
> This work will be done, so while I will accept an AST walking solution it will be deleted and replaced with a call graph approach when we implement further rename functionality in the not-so-distant future.

Got it. The graph currently only contains nodes for functions and module constants, so we need to add support for others such as local variables.

[...]

PR #3273, opened 2024-06-13
Convert variable names to snake_case

### LINKED ISSUE #2978: Code action: Convert variable names to snake_case
Coming from other languages like Elm and JavaScript where the conventions are to have variables named in camel-case.

In Gleam, camel-case is forbidden and results in syntax errors, which I can very much understand as it should make for very consistent naming patterns.
My force of habit is strong, and I spend quite some time converting the camel-case names I write to snake-case, which is distracting from my goal at hand.

A suggestion I have that could make that easier, is have the formatter automatically convert names to snake-case.

For instance, the formatter could change the code like this
```gleam
fn doSomething() { todo }
-->
fn do_something() { todo }
```

The compiler error already suggests this name change in the error message.

Same idea for references to consts/functions in expressions.

If this can be done reliably, then I can't think of many downsides (but I may not be familiar enough with the syntax).

One downside I can imagine is that if you've never written Gleam, and you don't ever see the error message that variables should be snake_case, then it will feel quite jarring to see a rename. Once you know it, I think it will be helpful though.

GearsDatapacks:
Closes #2978 
Let me summarise what I've done:

- BadName errors are now no longer reported in the lexer. The lexer treats all previous bad names as valid.
- The analyser now is responsible for validating names. It checks the name of variables (`let`, `use` and `const`), functions, parameters, types and type constructors.
- Finding an error in a variable name still declares it, to allow any references to that variable to still type-check correctly.
- When the analyser finds an invalid name, it adds it to a `bad_names` array, just like `unused_imports`.
- I've had to add more location information to certain ast nodes to allow proper renaming of variables without overwriting other parts of the source code 

The main problem I see with this approach is that it would be easy to miss validation of names in certain places, but I can't think of an easy way to check all names in the analyser.
This implementation also doesn't rename any references to invalid variables, but #2956 may need to be closed before that's possible/easy to add.


GearsDatapacks:
I've decided to mark this ready for review as I would like feedback on it and it seems more likely to get reviewed this way

compiler-core/src/error.rs
```diff
@@ -2927,10 +2928,79 @@ See: https://tour.gleam.run/advanced-features/use/");
...
+                        },
+                        path: path.clone(),
+                        src: src.clone(),
+                        extra_labels: vec![],
+                    }),
+                }
+                },
+
+                TypeError::BadUpName { location, name } => {
+                    let label = "This is not a valid upname";
```
lpil, line 2978: "upname" doesn't mean anything to a Gleam programmer, it's an internal name to the compiler and could change at any time. 

Let's merge these 3 new error variants into one `BadName` variant and have an enum to say what the context is:

```rust
pub enum BadNameKind {
    Type,
    TypeVariable,
    CustomTypeVariant,
    Variable,
    Function,
}
```

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__rename_invalid_bit_array_pattern.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "apply_first_code_action(code, 1)"
```
lpil, line 3: Please put the original src as the expression so we can tell what the snapshot is for. It is quite challenging to debug quickly if you can't see what the input is!

compiler-core/src/type_/expression.rs
```diff
@@ -1315,6 +1349,17 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            for pattern in clause.pattern.iter() {
+                let errors = self.check_valid_pattern_names(pattern);
+                self.errors.extend(errors);
+            }
+            for multi_pattern in clause.alternative_patterns.iter() {
+                for pattern in multi_pattern {
+                    let errors = self.check_valid_pattern_names(pattern);
+                    self.errors.extend(errors);
+                }
+            }
```
lpil, line 1361: We should not traverse the patterns an extra time for this, rather it should be done during the existing pattern analysis.

lpil (commented):
Lovely work!! Thank you very much. I've left a handful of notes inline

GearsDatapacks:
Thanks for all the feedback! I know it's quite a big PR, and I'm still pretty new to this codebase. I'll go through and implement these as soon as possible.

compiler-core/src/error.rs
GearsDatapacks: > "upname" doesn't mean anything to a Gleam programmer, it's an internal name to the compiler and could change at any time.

Right, I figured that but I stuck with "upname" since that was used in error messages previously

GearsDatapacks:
I've made some changes:
- I've removed `BadNames` from the capnp schema
- I've extracted duplicate name-checking code into methods
  - Since these checks occur in three different structs, I wrote a trait to generalise that. I'm not sure if that was the best way to go about it, but it's the one that I thought of.
- The test files should now contain the source code :)

GearsDatapacks:
Thinking about this more, it might be better to implement `NameChecker` as a struct, which just holds references to the `errors` and `bad_names` vectors.

compiler-core/src/type_/tests/errors.rs
```diff
@@ -1130,6 +1130,174 @@ const foo = 2"
     );
 }
 
+#[test]
+fn invalid_const_name() {
```
lpil, line 1134: Beautiful tests!!!

compiler-core/src/type_/name.rs
```diff
@@ -0,0 +1,81 @@
+use ecow::EcoString;
+
+use crate::ast::{ArgNames, SrcSpan};
+
+use super::{BadNameKind, Error};
+use heck::{ToSnakeCase, ToUpperCamelCase};
+
+pub trait NameChecker {
```
lpil, line 8: We don't define new traits, please remove this.

Could you move all the logic for deciding whether names are valid elsewhere- it's not specific to types so it should go in the type_ module namespace.

lpil (commented):
Looking great!!

GearsDatapacks:
I've now moved the name checking code into function in `name.rs` inside the `analyse` module

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                errors.push(error);
+                bad_names.push(bad_name);
+            }
+        }
+    }
+    (errors, bad_names)
+}
+
+fn check_bad_name(location: SrcSpan, name: &EcoString, kind: BadNameKind) -> Option<Error> {
```
lpil, line 107: This name implies that checks it is a valid "bad name", which I found confusing. Could we give it a clearer name please

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,116 @@
+use ecow::EcoString;
```
lpil, line 1: None of the functions in this module check that the names are of the formats they are intended to check for. Using regexes here would be good.

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,116 @@
...
+
+use crate::ast::{ArgNames, SrcSpan};
+
+use super::{BadNameKind, Error};
+use heck::{ToSnakeCase, ToUpperCamelCase};
+
+type BadName = (SrcSpan, EcoString);
```
lpil, line 8: This name isn't right, it's not a bad name. It's a correction or such. Let's use a struct for this.

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,116 @@
...
+        .map(|error| (error, (location, EcoString::from(name.to_snake_case()))))
+}
+
+pub fn check_valid_discard_name(location: SrcSpan, name: &EcoString) -> Option<(Error, BadName)> {
+    check_bad_name(location, name, BadNameKind::Discard).map(|error| {
+        (
+            error,
+            (
+                location,
+                EcoString::from(format!("_{}", name.to_snake_case())),
```
lpil, line 25: Favour `eco_format!` rather than building a string and then converting it to an ecostring

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,116 @@
...
+            },
+            (location, EcoString::from(name.to_upper_camel_case())),
+        ));
+    }
+    None
+}
+
+pub fn check_valid_argument(names: &ArgNames) -> (Vec<Error>, Vec<BadName>) {
+    let mut errors = Vec::new();
+    let mut bad_names = Vec::new();
```
lpil, line 51: Pass in these as mutable references to avoid allocating these vectors and then immediately emptying them into another vector

lpil (commented):
Thank you!

GearsDatapacks:
I've now cleaned up everything that you suggested

CHANGELOG.md
```diff
@@ -21,6 +21,21 @@
   being attempted.
   ([Ameen Radwan](https://github.com/Acepie))
 
+- LSP now suggests a code a action to rename variables, types and functions when they
+  don't match the gleam naming conventions:
```
lpil, line 25:
```suggestion
- The language server now suggests a code a action to rename variables, types and
  functions when they don't match the Gleam naming requirements:
```

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,139 @@
...
+}
+
+pub fn check_valid_name(
+    location: SrcSpan,
+    name: &EcoString,
+    kind: BadNameKind,
+) -> Option<(Error, NameCorrection)> {
+    // Some of the internally generated variables (such as `_capture` and `_use0`)
+    // start with underscores, so we allow underscores here.
+    let valid_name_regex = Regex::new("^_?[a-z][a-z0-9_]*$").expect("Regex is correct");
```
lpil, line 22: This should be a static value rather than recomputing the regex on every name. See other uses of regex in the compiler for how we do this

compiler-core/src/analyse/name.rs
```diff
@@ -0,0 +1,139 @@
...
+use super::{BadNameKind, Error};
+use heck::{ToSnakeCase, ToUpperCamelCase};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct NameCorrection {
+    pub location: SrcSpan,
+    pub correction: EcoString,
+}
+
+pub fn check_valid_name(
```
lpil, line 15: These functions do more than just check the name, they also return a correction. Let's move that out of these and keep them doing one thing.

Let's also make the error case `Err`.

compiler-core/src/analyse.rs
```diff
@@ -957,6 +1004,13 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
         // could improve our approach here somewhat.
         environment.assert_unique_type_name(name, *location)?;
 
+        if let Some((error, correction)) =
+            check_valid_upname(*name_location, name, BadNameKind::Type)
+        {
+            self.errors.push(error);
+            self.name_corrections.push(correction);
+        }
```
lpil, line 1012: This 
```rust
if let Some((error, correct)) = .... {
    self.errors.push(error)
    self.name_corrections.push(correction)
}
```
is duplicated several times.

Let's make a helper for it:

```rust
self.check_name_case(name, location, NameKind::Type)
```

compiler-core/src/type_/error.rs
```diff
@@ -449,6 +463,46 @@ pub enum LiteralCollectionKind {
     Record,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BadNameKind {
```
lpil, line 467:
```suggestion
pub enum Named {
```

compiler-core/src/type_/error.rs
```diff
@@ -449,6 +463,46 @@ pub enum LiteralCollectionKind {
...
+    Variable,
+    Argument,
+    Label,
+    Constant,
+    Function,
+    Discard,
+}
+
+impl BadNameKind {
+    pub fn to_string(self) -> EcoString {
```
lpil, line 480:
```suggestion
    pub fn as_str(&self) -> &'static str {
```

compiler-core/src/error.rs
```diff
@@ -2949,9 +2950,39 @@ See: https://tour.gleam.run/advanced-features/use/");
                         }),
                     }
                 },
+
+                TypeError::BadName { location, name, kind } => {
+                    let kind_str = kind.to_string();
+                    let label = format!("This is not a valid {kind_str} name");
+                    let text = if kind.is_discard() {
```
lpil, line 2957: Use pattern matching instead of `if/else`

lpil (commented):
Nearly there! Thank you

GearsDatapacks:
All done now.
- I've made the various refactors you suggested
- The `check_valid_name` functions have now been reworked into two functions: `check_name_case` and `correct_name_case`
  - `check_name_case` returns an error if the given name is invalid
  - `correct_name_case` returns a `NameCorrection` for the given name

lpil (approved):
This is fantastic!!! Thank you!

Issue #3252, opened 2024-06-08
LSP: Add Document Symbol support


leonqadirie:
Suggestion: Add LSP support for [Document Symbols](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentSymbol).

From the LSP spec:
> They represents programming constructs like variables, classes, interfaces etc. that appear in a document. Document symbols can be hierarchical and they have two ranges: one that encloses its definition and one that points to its most interesting range, e.g. the range of an identifier.

They provide a (subjectively) great way to quickly conceptualise/grasp and navigate code in editors that support them. An example of a buffer's Document Symbols in Rust is attached.

![LSP Document Symbols - Example](https://github.com/gleam-lang/gleam/assets/39130739/78fa40b5-7819-40fa-96ed-2b07c3acb194)

lpil:
Oooh very nice! Sounds like it should be straightforward to add too.

PgBiel:
Working on it!

![image](https://github.com/gleam-lang/gleam/assets/9021226/4bc24819-fcaf-4ee4-ada7-7807b4cdf25c)

By the way, it seems that Zed already supports this feature for Gleam out of the box, but this will still matter for VSCode and (potentially) other editors.

leonqadirie:
Very nice, thank you!

Zed added support [in this PR](https://github.com/zed-industries/zed/pull/6733), without diving into the code I suspect they use Gleam's tree-sitter to implement the symbol view. It definitely matters for other editors!

PR #3275, opened 2024-06-13
LSP: Support document symbols

PgBiel:
Fixes https://github.com/gleam-lang/gleam/issues/3252

The main implementation is mostly done. All that is left is:

1. **DONE:** For `selection_range`, I imagine that the correct range would be one pointing to the identifier for each symbol (e.g. a range around `Name` in `pub type Name`); right now I just duplicated the full symbol's range. I will look into this soon (draft in the meantime).
2. Code style nitpicks might be in order. Feel free to ask for any changes in this regard.

It may also be desirable to write the full function signature alongside function symbols instead of just their types (at least, that's how `rust-analyzer` does it), but, in principle, the type has been working fine for an MVP. We could change this later if needed.

By the way, the `#[allow(deprecated)]` are because `DocumentSymbol` contains a deprecated field, but we have to specify it anyway to construct it.

## Showcase

The source code below:

```gleam

pub type A

pub type B {
    C

    D(Option(Int))

    E(something: Option(Int))

    F(Int, something: Option(Int))
}

pub const my_const = 5

pub fn super_func(a: Int) -> Int {
    a + 5
}
```

produces the following outline on VSCode:

image


PgBiel:
I've fixed symbol locations such that:

1. They now properly span the whole symbol, including a custom type's constructors or a function's body, so VSCode can now properly show the full path of symbols to your current cursor location.
    image

2. The symbol's `selection_range` now only spans its name (whenever available, which should be always for top-level definitions). This is consistent with what the LSP standards expect, and also with what `rust-analyzer` does. This ensures clicking the symbol on the outline takes you to its name and not to `pub fn` or similar.
    - To do this, I had to add `name_location` / `label_location` / `alias_location` fields to a few symbols and properly populate them in the parser. Perhaps it could be more consistent to have a `selection_location` field common to those types, like rust-analyzer does, which is always present and is used to (later) instruct the LSP on what to do. But I think the current, less generalized impl will do for now.

PgBiel:
Included doc comments as part of the symbol ranges as well now. Seems to match the LSP spec's recommendation, as well as what `rust-analyzer` does (for comparison).

image

PgBiel:
By the way, let me know if I should have used `ast::visit` for any part of the implementation. In particular, I wanted to avoid introducing those more fine-grained location fields to the AST structures to be filled during parsing, but I couldn't think of a better way. I'm open to suggestions regarding any possible improvements here.

compiler-core/src/ast.rs
```diff
@@ -219,10 +220,12 @@ pub type TypedRecordConstructorArg = RecordConstructorArg<Arc<Type>>;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct RecordConstructorArg<T> {
     pub label: Option<EcoString>,
+    pub label_location: Option<SrcSpan>,
     pub ast: TypeAst,
     pub location: SrcSpan,
     pub type_: T,
     pub doc: Option<EcoString>,
+    pub doc_position: Option<u32>,
```
lpil, line 228: Same here, group it with the documentation.

compiler-core/src/ast.rs
```diff
@@ -467,13 +470,15 @@ pub struct Function<T, Expr> {
     pub location: SrcSpan,
     pub end_position: u32,
     pub name: EcoString,
+    pub name_location: Option<SrcSpan>,
```
lpil, line 473: How come the name is optional?

compiler-core/src/ast.rs
```diff
@@ -570,9 +576,11 @@ pub struct CustomType<T> {
     pub location: SrcSpan,
     pub end_position: u32,
     pub name: EcoString,
+    pub name_location: SrcSpan,
     pub publicity: Publicity,
     pub constructors: Vec<RecordConstructor<T>>,
     pub documentation: Option<EcoString>,
+    pub doc_position: Option<u32>,
```
lpil, line 583: Group with the documentation

compiler-core/src/ast.rs
```diff
@@ -605,11 +613,13 @@ pub type UntypedTypeAlias = TypeAlias<()>;
 pub struct TypeAlias<T> {
     pub location: SrcSpan,
     pub alias: EcoString,
+    pub alias_location: SrcSpan,
```
lpil, line 616: I was a bit confused by this as it's `alias.alias_location`. Is it the name, or the whole alias definition?

compiler-core/src/ast.rs
```diff
@@ -605,11 +613,13 @@ pub type UntypedTypeAlias = TypeAlias<()>;
...
     pub parameters: Vec<EcoString>,
     pub type_ast: TypeAst,
     pub type_: T,
     pub publicity: Publicity,
     pub documentation: Option<EcoString>,
+    pub doc_position: Option<u32>,
```
lpil, line 622: Group with the documentation please

compiler-core/src/language_server/engine.rs
```diff
@@ -267,6 +270,107 @@ where
...
+        self.respond(|this| {
+            let mut symbols = vec![];
+            let Some(module) = this.module_for_uri(&params.text_document.uri) else {
+                return Ok(symbols);
+            };
+            let line_numbers = LineNumbers::new(&module.code);
+
+            for definition in &module.ast.definitions {
+                match definition {
+                    #[allow(deprecated)]
```
lpil, line 286: We can not specify the `deprecated` field to avoid needing to allow deprecated fields?

compiler-core/src/ast.rs
PgBiel: It's optional for anonymous functions. When it's missing, the parser just assigns an empty string to `name` (which weirded me out a bit). I'd say we could either change `name` to `Option<EcoString>`, or maybe keep `name_location` as a `SrcSpan` but assigning `0` when there's no name. (Regardless, for top-level functions, they won't ever be anonymous, so we can't hit that case.)

compiler-core/src/ast.rs
```diff
@@ -206,6 +206,7 @@ pub struct RecordConstructor<T> {
     pub name: EcoString,
     pub arguments: Vec<RecordConstructorArg<T>>,
     pub documentation: Option<EcoString>,
+    pub doc_position: Option<u32>,
```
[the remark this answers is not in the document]
PgBiel: Do you think we should introduce a separate type to make it more semantically suggestive? I.e. something like `Spanned<EcoString>` (which is what Typst uses) instead of `(u32, EcoString)`.

compiler-core/src/language_server/engine.rs
PgBiel: I think the problem is that not specifying the field translates to specifying it as `None` in the Rust side, whereas `Some(...)` would be (wrongly, since it is deprecated) specifying it. But their usage of `#[deprecated(...)]` doesn't distinguish between both cases, since it applies to any usage of the field...

Is there an alternative you're thinking of?

compiler-core/src/ast.rs
PgBiel: It's just the name! I wanted it to be consistent with the `alias` field, so I guess it suffers from the same problem haha. Maybe we could rename both to `name`?

compiler-core/src/ast.rs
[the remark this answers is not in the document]
PgBiel: It turns out

> This should always be present if the documentation is present

doesn't appear to be true for this particular type due to the `put_doc` method below, which allows adding documentation outside of parsing. This appears to only be used in `build.rs`:

https://github.com/gleam-lang/gleam/blob/17fa7b8a218358e60685298d3cfb5af0a4d7f31c/compiler-core/src/build.rs#L263-L270

Not sure what's the best way forward here. Either we keep the field separate for this type, or we set the position to be the same as `location.start` in such a case.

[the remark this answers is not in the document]
PgBiel: Actually, I found out that `build.rs` had enough information to provide the documentation position, so I changed `put_doc` to force a position to be specified alongside the documentation, and everything went well. Carry on!

compiler-core/src/analyse.rs
```diff
@@ -532,7 +533,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
         }
 
         let variant = ValueConstructorVariant::ModuleFn {
-            documentation: doc.clone(),
+            documentation: doc.clone().map(|(_, doc)| doc),
```
lpil, line 536: Let's only clone the string as we immediately discard the span

compiler-core/src/analyse.rs
```diff
@@ -361,7 +361,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             publicity,
             deprecation: deprecation.clone(),
             variant: ValueConstructorVariant::ModuleConstant {
-                documentation: doc.clone(),
+                documentation: doc.clone().map(|(_, doc)| doc),
```
lpil, line 364: Let's only clone the string as we immediately discard the span

compiler-core/src/analyse.rs
```diff
@@ -874,7 +876,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 _ => fn_(args_types.clone(), typ.clone()),
             };
             let constructor_info = ValueConstructorVariant::Record {
-                documentation: constructor.documentation.clone(),
+                documentation: constructor.documentation.clone().map(|(_, doc)| doc),
```
lpil, line 879: Let's only clone the string as we immediately discard the span

compiler-core/src/ast.rs
lpil: Cool! That makes sense, thank you.

It would be good to also add an example of an anon function in the doc comment for the struct and add a doc comment to the field saying it's optional for anonymous functions

compiler-core/src/call_graph/into_dependency_order_tests.rs
```diff
@@ -29,6 +29,7 @@ fn parse_and_order(
                 .collect_vec(),
             body: crate::parse::parse_statement_sequence(src).expect("syntax error"),
             location: Default::default(),
+            name_location: (!name.is_empty()).then_some(SrcSpan::default()),
```
lpil, line 32: I don't understand this code. Could it be written with `match` or `if` please

compiler-core/src/language_server/engine.rs
lpil: Isn't the deprecated field the one called `deprecated`, which we do set to None?

If it's some other field could we put the attribute on the field itself to avoid making this entire clause? With a comment explaining the situation too

compiler-core/src/package_interface.rs
```diff
@@ -406,7 +406,7 @@ impl ModuleInterface {
                     let _ = types.insert(
                         name.clone(),
                         TypeDefinitionInterface {
-                            documentation: documentation.clone(),
+                            documentation: documentation.clone().map(|(_, doc)| doc),
```
lpil, line 409: Clone just the string please 🙏

compiler-core/src/package_interface.rs
```diff
@@ -415,13 +415,16 @@ impl ModuleInterface {
                                 constructors
                                     .iter()
                                     .map(|constructor| TypeConstructorInterface {
-                                        documentation: constructor.documentation.clone(),
+                                        documentation: constructor
+                                            .documentation
+                                            .clone()
+                                            .map(|(_, doc)| doc),
```
lpil, line 421: Clone just the string please 🙏

compiler-core/src/package_interface.rs
```diff
@@ -415,13 +415,16 @@ impl ModuleInterface {
...
                                         name: constructor.name.clone(),
                                         parameters: constructor
                                             .arguments
                                             .iter()
                                             .map(|arg| ParameterInterface {
-                                                label: arg.label.clone(),
+                                                label: arg.label.clone().map(|(_, label)| label),
```
lpil, line 427: Clone just the string please 🙏

compiler-core/src/package_interface.rs
```diff
@@ -448,7 +451,7 @@ impl ModuleInterface {
                     let _ = type_aliases.insert(
                         alias.clone(),
                         TypeAliasInterface {
-                            documentation: documentation.clone(),
+                            documentation: documentation.clone().map(|(_, doc)| doc),
```
lpil, line 454: Clone just the string please 🙏

compiler-core/src/package_interface.rs
```diff
@@ -476,7 +479,7 @@ impl ModuleInterface {
                             ),
                             type_: TypeInterface::from_type(type_.as_ref()),
                             deprecation: DeprecationInterface::from_deprecation(deprecation),
-                            documentation: documentation.clone(),
+                            documentation: documentation.clone().map(|(_, doc)| doc),
```
lpil, line 482: Clone just the string please 🙏

compiler-core/src/package_interface.rs
```diff
@@ -505,7 +509,7 @@ impl ModuleInterface {
                                 implementations,
                             ),
                             deprecation: DeprecationInterface::from_deprecation(deprecation),
-                            documentation: documentation.clone(),
+                            documentation: documentation.clone().map(|(_, doc)| doc),
```
lpil, line 512: Clone just the string please 🙏

lpil (commented):
Lovely work! I've left a few notes inline, and we'll need tests for the document symbols part.

compiler-core/src/language_server/engine.rs
PgBiel: The attribute being in the entire clause was a mistake, thanks for pointing that out. The clause used to just return a `DocumentSymbol` directly, which is why the attribute was added above it, but I added more stuff afterwards and forgot to move the attribute. I've fixed that now.

Unfortunately, it seems that adding the attribute to the field itself doesn't work, so I kept it above the `DocumentSymbol` constructor. I added a comment on the first one explaining the situation.

compiler-core/src/ast.rs
PgBiel: This field no longer exists and was merged with `alias`, so I'll resolve the thread. Feel free to unresolve if further action should be taken.

compiler-core/src/ast.rs
PgBiel: Done! I kept the fields there and documented the two cases (anonymous and named functions).

PgBiel:
I'll look into adding some tests soon :)

compiler-core/src/ast.rs
```diff
@@ -457,25 +457,32 @@ impl Publicity {
 #[derive(Debug, Clone, PartialEq, Eq)]
 /// A function definition
 ///
+/// Note that, for an anonymous function, the `name` field will be an empty string,
+/// and the `name_location` field will be `None`. Conversely, a named function
+/// will have a non-empty `name` field, and its `name_location` field will be `Some`.
```
lpil, line 462: This should be one option containing both, as with the others

compiler-core/src/ast.rs
```diff
@@ -633,7 +638,9 @@ pub enum Definition<T, Expr, ConstantRecordTag, PackageName> {
 impl TypedDefinition {
     pub fn main_function(&self) -> Option<&TypedFunction> {
         match self {
-            Definition::Function(f) if f.name == "main" => Some(f),
+            Definition::Function(f) if f.name.as_ref().is_some_and(|(_, name)| name == "main") => {
```
lpil, line 641: TIL is_some_and!!!

lpil (commented):
Lovely work!! Could you add some tests and also add that comment to each of the `#[allow` attributes please 🙏

PgBiel:
Sure, I'll do it soon!
Sorry, had run out of time 😅
But I definitely want to move this PR across the finish line :)

PgBiel:
Some initial notes from tests:

1. Initial `///` from first line of doc comments isn't being included in symbol spans;
2. `pub const` isn't being included in the location (it seems to be generally expected that the constant's location field corresponds to its name);
3. Constructor over multiple lines is not spanning the end parens, but rather stopping at the start of the line containing it;
4. Constructor selection range seems to be across the whole constructor instead of just its name

PgBiel:
@lpil There we go, had to make several changes when rebasing on main and upon adding tests, but I think it should be fine now. Feel free to take a look :)

compiler-core/src/ast.rs
```diff
@@ -696,10 +700,12 @@ pub type UntypedModuleConstant = ModuleConstant<(), ()>;
...
 /// ```
 pub struct ModuleConstant<T, ConstantRecordTag> {
-    pub documentation: Option<EcoString>,
+    pub documentation: Option<(u32, EcoString)>,
+    /// The location of the constant, starting at the "(pub) const" keywords and
+    /// ending after the ": Type" annotation, or (without an annotation) after its name.
     pub location: SrcSpan,
     pub publicity: Publicity,
-    pub name: EcoString,
+    pub name: (SrcSpan, EcoString),
```
lpil, line 708: This should be two named fields as it's not an optional field which forces them to be grouped together.

What's the srcspan used for? I couldn't see where it is used.

compiler-core/src/ast.rs
```diff
@@ -727,12 +733,11 @@ pub type UntypedCustomType = CustomType<()>;
 /// ```
 pub struct CustomType<T> {
     pub location: SrcSpan,
-    pub name_location: SrcSpan,
     pub end_position: u32,
-    pub name: EcoString,
+    pub name: (SrcSpan, EcoString),
```
lpil, line 737: This should be two fields, as above

compiler-core/src/parse.rs
```diff
@@ -209,7 +209,7 @@ pub struct Parser<T: Iterator<Item = LexResult>> {
     tok0: Option<Spanned>,
     tok1: Option<Spanned>,
     extra: ModuleExtra,
-    doc_comments: VecDeque<(u32, String)>,
+    doc_comments: VecDeque<(u32, u32, String)>,
```
lpil, line 212: It's not clear to me what these values are. Could you make a struct for them so that they have names. Thank you

compiler-core/src/parse/extra.rs
```diff
@@ -7,7 +7,9 @@ use crate::ast::SrcSpan;
 #[derive(Debug, PartialEq, Eq, Default)]
 pub struct ModuleExtra {
     pub module_comments: Vec<SrcSpan>,
-    pub doc_comments: Vec<SrcSpan>,
+    /// Contains location information for each documentation comment, including the position
+    /// right before its '///' marker starts, as well as the span of its contents (after the marker).
+    pub doc_comments: Vec<(u32, SrcSpan)>,
```
lpil, line 12: Isn't this u32 always the same as the `start` of the SrcSpan?

compiler-core/src/parse/token.rs
```diff
@@ -10,7 +10,7 @@ pub enum Token {
     Int { value: EcoString },
     Float { value: EcoString },
     String { value: EcoString },
-    CommentDoc { content: String },
+    CommentDoc { marker_start: u32, content: String },
```
lpil, line 13: Why does this token have positional information? Positional information is handled at the level above, as ou can see with how none of the other tokens have any.

lpil (commented):
Thank you. I've left some more notes inline.

I think something has gone awry with the rebase, there's conflicts which mean I cannot rebase this branch into main.

compiler-core/src/parse/extra.rs
PgBiel: The `start` of the SrcSpan points to the `x` in `///x`, whereas the u32 points to the first `/` in `///x`. I just wanted to distinguish the marker start from the content start, basically, as comments always use the content spans as their own spans, not considering their markers.

I guess we could switch to just taking the content start and subtracting 3 instead of storing the marker start if a doc comment can never exist without `///` anyway.

compiler-core/src/parse/token.rs
PgBiel: I mostly just wanted a way to return the `marker_start` as well as the `start` without having to change a ton of signatures in the lexer (reasoning here https://github.com/gleam-lang/gleam/pull/3275#discussion_r1685511724 ). We could attempt the simplification mentioned in that reply instead if you'd prefer.

compiler-core/src/ast.rs
PgBiel: Name spans are needed for symbols. This one is used here: https://github.com/gleam-lang/gleam/blob/95e11ffd80b7ba9eaeda22c45445dd7d10535033/compiler-core/src/language_server/engine.rs#L405

PgBiel:
> I think something has gone awry with the rebase, there's conflicts which mean I cannot rebase this branch into main.

That's probably due to merge commits; you'll have to squash and merge instead. (I tried to manually squash locally, but it's a bit difficult at this point :p)

compiler-core/src/parse/extra.rs
lpil: It doesn't matter which we go for, but we shouldn't duplicate the data here. One is always the other + 3, so it can be calculated when needed.

PgBiel:
There we go!

The CI is failing as it seems you've added an explicit check against merge commits, but other than that it's fine. I believe I owe a bit of an explanation: The reason I've used merge commits to keep the branch up-to-date with Gleam's `main` is that it was much simpler (for me, as a contributor): for each update, I only had to resolve the conflicts considering the latest version of the PR; I didn't have to update each and every commit since the PR was created (in a traditional rebase), saving a lot of time (as I had to update the branch multiple times). Then, by squashing and merging at the end (at least up to the latest merge commit, but just pressing "Squash and merge" is of course easier), no merge commits will make it to Gleam's main branch, avoiding any conflicts and applying the patch in the same way. (This is the policy we generally follow at [Typst](https://github.com/typst/typst), for instance.)

With that said, it may of course be preferable to be able to rebase the PR to `main` and keep the commit history depending on your preferred policies, though in principle I don't see problems with either model, so long as the contributor can deliver their changes.

Regarding this PR, you should be able to squash and merge as needed, given the above. Let me know if you face any problems or have any concerns.

lpil (approved):
Thank you!!!!

Issue #3271, opened 2024-06-12
LSP: When auto adding import, if it's the first import, a blank line should follow it


johtso:
With the new tab completion auto adding of imports, if it's the first import in the file, there's no blank line between it and the first line of code.

lpil:
Good idea. Thanks!

codemonkey76:
I've had a go at this, see #3277

Acepie:
This can be closed now right?

lpil:
Yes, thank you

PR #3277, opened 2024-06-14
Add blank line following when auto importing as first import, #3271

codemonkey76:


compiler-core/src/language_server/engine.rs
```diff
@@ -471,6 +471,7 @@ where
 
         // Importable modules
         let import_location = self.first_import_line_in_module(module, line_numbers);
+        let has_any_imports = module.ast.definitions.iter().find_map(get_import).is_none();
```
Acepie, line 474: Am I reading this wrong or should this be `is_some`

codemonkey76: ah, i had my variable called has_no_imports before, lemme fix it

codemonkey76: I'm a little confused i think... because it works and the tests pass, but my logic seems backwards:
```rust
        let has_any_imports = module.ast.definitions.iter().find_map(get_import).is_none();
```
and
```rust
                let new_lines = if has_any_imports { "\n" } else { "\n\n" };
```

Acepie: It looks like your logic is right for values but reversed for types based on the code and tests in place now

Acepie:
Thank you :D this looks good to me

lpil (commented):
Rather than looking for the first import could we look for the first statement please 🙏 So imports are always added at the top.

codemonkey76:
I'm checking if there is NO import, and adding an extra newline in the completion if true.

lpil:
Aye, but we still want an extra new line if there is an import but it is not at the top, no?

lpil:
Thank you. In which case please move this into the existing method that find the import location, to avoid traversing the module twice and to make the code more concise.

PR #3278, opened 2024-06-14
Surface diagnostic errors for import cycles

### LINKED ISSUE #3274: Surface diagnostic errors for import cycles
It would be helpful to surface some/all of the import statements causing an import cycle via a diagnostic error. This way the editor does not appear to have started failing to report all diagnostic data from the gleam build tool.

Currently import cycles result in compilation errors like the following:

```
error: Import cycle

The import statements for these modules form a cycle:

    ┌─────┐
    │     thing3
    │     ↓
    │     thing2
    │     ↓
    │     thing
    └─────┘
Gleam doesn't support dependency cycles like these, please break the
cycle to continue.
```

In some editors (nvim, vscode) this error is logged with a popup but results in no other editor response, in zed there is no popup for this output 

It may also be useful for the compiler message to include the code locations of the imports involved

Acepie:
Closes #3274 

Looks up the import location within the module when a cycle error occurs and adds the location to the error. In addition, in order to make this change render, there were updates to the lsp feedback and diagnostic mechanisms to allow for cross file feedback and the rendering of extra labels.

### DIFF
--- a/compiler-core/src/build/package_loader.rs
+++ b/compiler-core/src/build/package_loader.rs
@@ -15,6 +15,7 @@ use itertools::Itertools;
 use vec1::Vec1;
 
 use crate::{
+    ast::SrcSpan,
     build::{module_loader::ModuleLoader, package_compiler::module_name, Module, Origin},
     config::PackageConfig,
     dep_tree,
@@ -120,7 +121,7 @@ where
             .sorted_by(|(a, _), (b, _)| a.cmp(b))
             .collect();
         let sequence = dep_tree::toposort_deps(deps)
-            .map_err(|e| convert_deps_tree_error(e, dep_location_map))?;
+            .map_err(|e| self.convert_deps_tree_error(e, dep_location_map))?;
 
         // Now that we have loaded sources and caches we check to see if any of
         // the caches need to be invalidated because their dependencies have
@@ -271,6 +272,63 @@ where
             mtime,
             self.warnings.clone(),
         )
+    }
+
+    fn convert_deps_tree_error(
+        &self,
+        e: dep_tree::Error,
+        dep_location_map: HashMap<EcoString, &Input>,
+    ) -> Error {
+        match e {
+            dep_tree::Error::Cycle(modules) => {
+                let modules = modules
+                    .iter()
+                    .enumerate()
+                    .map(|(i, module)| {
+                        // cycles are in order of reference so get next in list or loop back to first
+                        let ind = if i == modules.len() - 1 { 0 } else { i + 1 };
+                        let next = modules.get(ind).expect("next module must exist");
+                        let input = dep_location_map.get(module).expect("dependency must exist");
+                        let location = match input {
+                            Input::New(module) => {
+                                let (_, location) = module
+                                    .dependencies
+                                    .iter()
+                                    .find(|d| &d.0 == next)
+                                    .expect("import must exist for there to be a cycle");
+                                ImportCycleLocationDetails {
+                                    location: *location,
+                                    path: module.path.clone(),
+                                    src: module.code.clone(),
+                                }
+                            }
+                            Input::Cached(cached_module) => {
+                                let (_, location) = cached_module
+                                    .dependencies
+                                    .iter()
+                                    .find(|d| &d.0 == next)
+                                    .expect("import must exist for there to be a cycle");
+                                let src = self
+                                    .io
+                                    .read(&cached_module.source_path)
+                                    .expect("failed to read source")
+                                    .into();
+                                ImportCycleLocationDetails {
+                                    location: *location,
+                                    path: cached_module.source_path.clone(),
+                                    src,
+                                }
+                            }
+                        };
+                        (module.clone(), location)
+                    })
+                    .collect_vec();
+                Error::ImportCycle {
+                    modules: Vec1::try_from(modules)
+                        .expect("at least 1 module must exist in cycle"),
+                }
+            }
+        }
     }
 }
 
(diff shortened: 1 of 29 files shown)


compiler-core/src/build/package_loader.rs
```diff
@@ -1547,9 +1554,35 @@ fn ensure_gleam_module_does_not_overwrite_standard_erlang_module(input: &Input)
...
+) -> Error {
     match e {
-        dep_tree::Error::Cycle(modules) => Error::ImportCycle { modules },
+        dep_tree::Error::Cycle(modules) => {
+            let mut locations = vec![];
+            for (i, module) in modules.iter().enumerate() {
+                // cycles are in order of reference so get next in list or loop back to first
+                let ind = if i == modules.len() - 1 { 0 } else { i + 1 };
+                let next = modules.get(ind).expect("next module must exist");
+                // HACK: for cached modules we do not have the location of the import
```
lpil, line 1568: We will need to store this information then in this case.

compiler-core/src/diagnostic.rs
```diff
@@ -19,18 +21,37 @@ pub struct Label {
     pub span: SrcSpan,
 }
 
+trait ToCodespanLabel {
```
lpil, line 24: No new traits please.

compiler-core/src/diagnostic.rs
```diff
@@ -19,18 +21,37 @@ pub struct Label {
...
+        match &self.text {
+            None => label,
+            Some(text) => label.with_message(text.clone()),
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Location {
+pub struct ExtraLocation {
```
lpil, line 43: It looks like the vast majority of the time the path is the same, so I think location information should be optional.

compiler-core/src/error.rs
```diff
@@ -81,7 +88,10 @@ pub enum Error {
     DuplicateSourceFile { file: String },
 
     #[error("cyclical module imports")]
-    ImportCycle { modules: Vec<EcoString> },
+    ImportCycle {
+        modules: Vec<EcoString>,
+        locations: Vec1<ImportCycleLocationDetails>,
```
lpil, line 93: Shouldn't this be one vector? It seems they must always be the same length.

compiler-core/src/language_server/feedback.rs
```diff
@@ -123,15 +123,29 @@ impl FeedbackBookKeeper {
...
-                    feedback.append_diagnostic(path, diagnostic);
-                }
-
-                None => {
-                    feedback.append_message(diagnostic);
+            if let Some(l) = &diagnostic.location {
+                _ = self.files_with_errors.insert(l.path.clone());
+                feedback.append_diagnostic(l.path.clone(), diagnostic.clone());
+
+                for extra in &l.extra_labels {
```
lpil, line 130: This renders the additional information as entirely new diagnostics rather than being part of the first one. This may be OK for imports cycles, but in the duplicate import example there is now a diagnostic that just says "First imported here" with no other context.

It should instead be a nested diagnostic, possibly of a different colour like how hints are displayed.

lpil (commented):
Thank you! I've left some notes on how we're approaching the extra information. I think the way it is presented currently is a bit unclear for the user.

compiler-core/src/diagnostic.rs
Acepie: oh good call I don't need this anymore. In an old version of this code I had this implemented on the locations

compiler-core/src/language_server/feedback.rs
Acepie: Yeah that makes sense. The reason I had it like this was that it looks like sub-diagnostics can't be across files in the lsp protocol if I am reading correctly. Lemme change this so the ones in different files are separate diagnostics but the ones in the same file are sub diagnostics

Acepie: Oh nevermind I'm totally wrong and misread the protocol 🤦🏽 lemme fix this

compiler-core/src/language_server/server.rs
```diff
@@ -455,15 +456,47 @@ fn diagnostic_to_lsp(diagnostic: Diagnostic) -> Vec<lsp::Diagnostic> {
...
+        .extra_labels
+        .iter()
+        .map(|extra| {
+            let message = extra.label.text.clone().unwrap_or_default();
+            let location = if let Some(path) = &extra.path {
+                let line_numbers = LineNumbers::new(
+                    &extra
+                        .src
+                        .clone()
+                        .expect("different path must have source attached"),
```
lpil, line 472: If these should be given together than they should be in the same `Option`

compiler-core/src/error.rs
```diff
@@ -1335,9 +1348,13 @@ Names in a Gleam module must be unique so one will need to be renamed."
                             },
                             path: path.clone(),
                             src: src.clone(),
-                            extra_labels: vec![Label {
-                                text: Some("First imported here".into()),
-                                span: *previous_location,
+                            extra_labels: vec![ExtraLocation {
```
lpil, line 1351: Perhaps this name as it's not always a location
```suggestion
                            extra_labels: vec![ExtraLabel {
```

lpil (commented):
Oh this is looking really very nice!!!

compiler-core/src/language_server/server.rs
Acepie: Yeah makes sense I'll combine them into 1 tuple

compiler-core/src/error.rs
Acepie: works for me

Acepie:
Alright this should be all good now 👍🏽

lpil (approved):
Beautiful work!!!

PR #3282, opened 2024-06-16
Update content width of built docs to avoid horizontal scrolling

### LINKED ISSUE #3276: Horizontal scrolling in code blocks in built docs for lines 79-80 characters long
In documentation built with `gleam docs build`, code blocks with lines of length 79 or greater cause horizontal scrolling of the block, which is a bit unsightly. Gleam wraps code at 80 characters, so it would be good if 80 characters could be used in a code block in the docs without triggering a horizontal scrollbar. A scrollbar would still appear for lines longer than Gleam's 80 character convention.

This can be seen here: https://hexdocs.pm/ieee_float/0.1.0/index.html

Browsers affected are Safari (v17.5) and Chrome (v125), but Firefox (v126) doesn't seem to be affected.

An easy workaround is to adjust code blocks in documentation to only be max 78 characters long.

Gleam version: 1.2.1

richard-viney:
Fixes #3276.

The actual required value to avoid scrolling in my testing on macOS was 771px, but I haven't tested on Windows or Linux. I rounded up to 772px, but could consider rounding up further to be more on the safe side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,10 @@
   the same name as an imported type.
   ([Gears](https://github.com/gearsdatapacks))
 
+- Fixed a bug where a horizontal scrollbar would appear on code blocks in built
+  documentation when they contained lines 79 or 80 characters long.
+  ([Richard Viney](https://github.com/richard-viney))
+
 ## v1.2.1 - 2024-05-30
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you

Acepie:
Looks like this got stuck on a lint issue that is resolved on main. A rebase should fix this

richard-viney:
Rebased.

lpil (approved):
Thank you

PR #3285, opened 2024-06-17
Determine removed package exists for `gleam remove`

### LINKED ISSUE #3279: `gleam remove` always reports a removal, even if the package did not exist in `gleam.toml`
To reproduce, run `gleam remove <package>` using any arbitrary input, such as `gleam remove doesntexist`:
```
vxern@Dorian tatoeba % gleam remove doesntexist
    Removed doesntexist
```

Gleam version: `v1.2.1`

hnlcf:
Fix #3279

### DIFF
--- a/compiler-cli/src/cli.rs
+++ b/compiler-cli/src/cli.rs
@@ -139,6 +139,10 @@ pub(crate) fn print_removed(text: &str) {
     print_colourful_prefix("Removed", text)
 }
 
+pub(crate) fn print_not_exist(text: &str) {
+    print_colourful_prefix("Not Exist", text)
+}
+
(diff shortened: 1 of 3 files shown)


markholmes:
What does the output look like with this fix? Do you have an example?

hnlcf:
For non-existent packages, the `print_colourful_prefix` function is currently used to print `Not Exist`

lalala

lpil (commented):
Thank you.

Could you return an Error variant instead? And put the printing back to the end.

hnlcf:
OK, I return error instead for non-existent packages. 

image

lpil (approved):
Thank you

PR #3292, opened 2024-06-18
Fix #3288

### LINKED ISSUE #3288: Compiler panics when a custom type variant doesn't exist
Ran into this while working on the [Forth exercise](https://exercism.org/tracks/gleam/exercises/forth) on the exercism track.

It popped up when adding a custom type variant that held data of a type I had not yet created. Here's the full error message:

```
error: Fatal compiler bug!

Panic: compiler-core/src/exhaustiveness.rs:746
	Custom type variants must exist: Type { name: "Entry", hint: AlternativeTypes(["Float", "Bool", "BitArray", "Result", "Forth", "ForthError", "Nil", "UtfCodepoint", "Int", "List", "String", "Entry"]) }
Gleam version: 1.2.0
Operating system: macos

variable set.
```

Here is the code prior to receiving the error:

```gleam
import gleam/int
import gleam/list
import gleam/result
import gleam/string

pub type Forth {
  Forth(stack: List(Int))
}

pub type ForthError {
  DivisionByZero
  StackUnderflow
  InvalidWord
  UnknownWord
}

type Entry {
  Value(Int)
}


pub fn new() -> Forth {
  Forth([])
}

pub fn format_stack(f: Forth) -> String {
  case f {
    Forth([]) -> ""
    Forth(stack) ->
      stack
      |> list.reverse
      |> list.map(int.to_string)
      |> string.join(" ")
  }
}

pub fn eval(f: Forth, prog: String) -> Result(Forth, ForthError) {
  let #(entry, rest) = result.unwrap(string.split_once(prog, " "), #(prog, ""))

  case f, prog {
    Forth(stack), "" -> Ok(Forth(stack))
    Forth(stack), _ -> {
      use entry <- result.try(parse_entry(entry))
      case entry {
        Value(num) -> eval(Forth([num, ..stack]), rest)
      }
    }
  }
}

fn parse_entry(entry: String) -> Result(Entry, ForthError) {
  case int.parse(entry) {
    Ok(num) -> Ok(Value(num))
    _ ->
      case entry {
        "+" -> Ok(Value(0))
        _ -> Error(InvalidWord)
      }
  }
}

```

...And the code w/ the line (`Word(Operator)`) that causes the error:

```gleam
import gleam/int
import gleam/list
import gleam/result
import gleam/string

pub type Forth {
  Forth(stack: List(Int))
}

pub type ForthError {
  DivisionByZero
  StackUnderflow
  InvalidWord
  UnknownWord
}

type Entry {
  Value(Int)
  Word(Operator)
}


pub fn new() -> Forth {
  Forth([])
}

pub fn format_stack(f: Forth) -> String {
  case f {
    Forth([]) -> ""
    Forth(stack) ->
      stack
      |> list.reverse
      |> list.map(int.to_string)
      |> string.join(" ")
  }
}

pub fn eval(f: Forth, prog: String) -> Result(Forth, ForthError) {
  let #(entry, rest) = result.unwrap(string.split_once(prog, " "), #(prog, ""))

  case f, prog {
    Forth(stack), "" -> Ok(Forth(stack))
    Forth(stack), _ -> {
      use entry <- result.try(parse_entry(entry))
      case entry {
        Value(num) -> eval(Forth([num, ..stack]), rest)
      }
    }
  }
}

fn parse_entry(entry: String) -> Result(Entry, ForthError) {
  case int.parse(entry) {
    Ok(num) -> Ok(Value(num))
    _ ->
      case entry {
        "+" -> Ok(Value(0))
        _ -> Error(InvalidWord)
      }
  }
}

```

It makes sense that my code would not compile, but it's a little surprising that it causes the compiler to panic.

Hope this report is helpful!

GearsDatapacks:
Fixes #3288 
This just makes the analyser more fault tolerant for custom types that reference other types that don't exist, and still registers the custom type in that case.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -844,7 +844,13 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 constructor.arguments.iter().enumerate()
             {
                 // Build a type from the annotation AST
-                let t = hydrator.type_from_ast(ast, environment)?;
+                let t = match hydrator.type_from_ast(ast, environment) {
+                    Ok(t) => t,
+                    Err(e) => {
+                        self.errors.push(e);
+                        continue;
+                    }
+                };
 
                 fields.push(TypeValueConstructorField { type_: t.clone() });
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,10 @@
   aliasing it and then using it in a constant expression would generate invalid
   JavaScript.
   ([Louis Pilfold](https://github.com/lpil))
+  
+- Fixed a bug where the compiler would crash because types weren't registered if
+  they referenced a non-existent type.
+  ([Gears](https://github.com/gearsdatapacks))
 
 ## v1.2.1 - 2024-05-30
 
--- a/compiler-core/src/type_/tests/exhaustiveness.rs
+++ b/compiler-core/src/type_/tests/exhaustiveness.rs
@@ -966,3 +966,23 @@ pub fn main(x: something) {
 "#
     );
 }
+
+#[test]
+fn reference_absent_type() {
+    // This test is here because this code previously caused the compiler
+    // to crash, and we want to make sure that it doesn't break again
+    assert_module_error!(
+        "
+type Foo {
+    Bar(Int)
+    Baz(Qux)
+}
+
+pub fn main(foo) {
+    case foo {
+        Bar(x) -> x
+    }
+}
+"
+    );
+}
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__exhaustiveness__reference_absent_type.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__exhaustiveness__reference_absent_type.snap
@@ -1,0 +1,38 @@
+---
+source: compiler-core/src/type_/tests/exhaustiveness.rs
+expression: "\ntype Foo {\n    Bar(Int)\n    Baz(Qux)\n}\n\npub fn main(foo) {\n    case foo {\n        Bar(x) -> x\n    }\n}\n"
+---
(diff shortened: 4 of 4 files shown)


GearsDatapacks:
There are probably other cases where a similar crash happens; I haven't really checked for places that might happen, but it might be worth looking into

compiler-core/src/analyse.rs
```diff
@@ -839,8 +839,14 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             for (i, RecordConstructorArg { label, ast, .. }) in
                 constructor.arguments.iter().enumerate()
             {
-                // Build a type from the annotation AST
-                let t = hydrator.type_from_ast(ast, environment)?;
+                // Build a type from the annotation AST// Build a type from the annotation AST
```
lpil, line 842: Funny comment here

lpil (commented):
Hello! Thank you for this fix 💜 

For my understanding: how does this fix the crash? What was the cause?

compiler-core/src/analyse.rs
GearsDatapacks: Oops. must have accidentally copy-pasted.

GearsDatapacks:
> For my understanding: how does this fix the crash? What was the cause?

Essentially, when the compiler ran into an error analysing a constructor, for example if it referenced a non-existent type, it would exit the function early, not getting to the piece of code that properly registers the type and constructors. Then, in the exhaustiveness checker, there was a `.expect` which relied on that registering, which then caused the crash.

I can't remember the exact details of what bit it's relying on, but this just prevents that early return.

compiler-core/src/analyse.rs
GearsDatapacks: I've now fixed that comment.

lpil (approved):
Ah that makes sense. Thank you very much

PR #3293, opened 2024-06-19
Support more bit array options for ints and floats on JavaScript

richard-viney:
This PR adds support for the following options for floats and ints in bit array patterns & expressions on the JavaScript target:

- Explicit size on floats: 64-bit and 32-bit. 16-bit floats aren't included, though they are supported by Erlang.
- Explicit endianness on floats and ints via `big` and `little` options. `native` endianness hasn't been added, but potentially could be. `big` remains the default.
- Explicit signedness on ints via `signed` and `unsigned`.

Implementation notes:

1. There was some pre-existing code duplication in `expression.rs` for bit arrays in constant expressions. This PR hasn't aimed to fix this, and there's now a bit more code duplication of a similar nature in `expression.rs` as a result.
2. There was one place where it was incorrectly assumed that the JavaScript platform is little endian (in the use of `Float64Array`). Most clients these days are indeed little endian, so it wasn't much of a problem in practice, but I've changed the code to use `DataView` instead.
3. There are non-backwards compatible changes to function signatures in the JavaScript prelude, but affected functions were all tagged with `// @internal` so I assumed that it's ok to change these.

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -6,6 +6,7 @@ use super::{
 };
 use crate::{
     ast::*,
+    javascript::endianness::Endianness,
     line_numbers::LineNumbers,
     pretty::*,
     type_::{ModuleValueConstructor, Type, ValueConstructor, ValueConstructorVariant},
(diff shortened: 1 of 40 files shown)


compiler-core/src/javascript/expression.rs
```diff
@@ -222,57 +222,112 @@ impl<'module> Generator<'module> {
...
+                            });
+                        }
+
+                        self.not_in_tail_position(|gen| gen.wrap_expression(size))?
+                    }
+                    _ => {
+                        let default_size = if segment.type_ == crate::type_::int() {
+                            8usize
+                        } else {
+                            64usize
```
lpil, line 274: Shouldn't this only be for floats?

compiler-core/src/javascript/expression.rs
```diff
@@ -222,57 +222,112 @@ impl<'module> Generator<'module> {
...
-                    if size_int > 0 && size_int % 8 != 0 {
-                        return Err(Error::Unsupported {
-                            feature: "Non byte aligned array".into(),
-                            location: segment.location,
-                        });
-                    }
-                    self.tracker.sized_integer_segment_used = true;
-                    let size = self.not_in_tail_position(|gen| gen.wrap_expression(size))?;
-                    Ok(docvec!["sizedInt(", value, ", ", size, ")"])
+            if segment.type_ == crate::type_::int() || segment.type_ == crate::type_::float() {
```
lpil, line 226: This whole function has become hard to read and understand, and I'm not confident I can review it well enough to catch any bugs with the design here.

Could we instead split into a parsing step which turns it into a struct or an enum, and then use that to perform code generation. Thank you

compiler-core/src/javascript/pattern.rs
```diff
@@ -14,8 +14,8 @@ enum Index<'a> {
     Int(usize),
     String(&'a str),
     ByteAt(usize),
-    IntFromSlice(usize, usize),
-    FloatAt(usize),
+    IntFromSlice(usize, usize, bool, bool),
+    FloatFromSlice(usize, usize, bool),
```
lpil, line 18: Convert these to struct variants `IntFromSlice { a: ..., b: ...}` please as I don't understand what the 4 unnamed fields are here

compiler-core/src/javascript/pattern.rs
```diff
@@ -88,12 +88,19 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
...
+        is_big_endian: bool,
+        is_signed: bool,
+    ) {
+        self.path
+            .push(Index::IntFromSlice(start, end, is_big_endian, is_signed));
     }
 
-    fn push_float_at(&mut self, i: usize) {
-        self.path.push(Index::FloatAt(i));
+    fn push_float_from_slice(&mut self, i: usize, end: usize, is_big_endian: bool) {
```
lpil, line 102: Let's use an enum for endianness

compiler-core/src/javascript/pattern.rs
```diff
@@ -470,78 +495,111 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
...
                 let mut offset = Offset::new();
                 for segment in segments {
-                    match segment.options.as_slice() {
-                        [] | [Opt::Int { .. }] => {
-                            self.push_byte_at(offset.bytes);
-                            self.traverse_pattern(subject, &segment.value)?;
-                            self.pop();
-                            offset.increment(1);
-                            Ok(())
+                    if segment.type_ == crate::type_::int()
```
lpil, line 498: Same here as above with doing it in two passes

compiler-core/src/javascript/tests/bit_arrays.rs
```diff
@@ -50,6 +50,24 @@ fn float() {
         r#"
 fn go() {
   <<1.1:float>>
+  <<1.1:float-big>>
+  <<1.1:float-little>>
```
lpil, line 54: Each test should only test one thing, so make a new test for each of these new examples please :pray:

compiler-core/templates/prelude.mjs
```diff
@@ -156,38 +156,76 @@ export function toBitArray(segments) {
...
-  let value = int;
+export function sizedInt(value, size, isBigEndian) {
   if (size < 0) {
     return new Uint8Array();
   }
-  if (size % 8 != 0) {
-    const msg = `Bit arrays must be byte aligned on JavaScript, got size of ${size} bits`;
-    throw new globalThis.Error(msg);
-  }
+
```
lpil, line 163: Put back the byte alignment check please 🙏

compiler-core/templates/prelude.mjs
```diff
@@ -156,38 +156,76 @@ export function toBitArray(segments) {
...
+export function byteArrayToFloat(byteArray, start, end, isBigEndian) {
+  const view = new DataView(byteArray.buffer);
+
+  const byteSize = end - start;
+
+  if (byteSize === 8) {
+    return view.getFloat64(start, !isBigEndian)
+  } else if (byteSize === 4) {
+    return view.getFloat32(start, !isBigEndian)
+  }
```
lpil, line 228: This function returns `undefined` when byteSize is not 8 or 4. It must always return a valid value.

Rather than have an end we could have a second bool for whether it is 32 or 64 bits.

compiler-core/templates/prelude.mjs
```diff
@@ -201,8 +239,23 @@ export function codepointBits(codepoint) {
 }
 
 // @internal
-export function float64Bits(float) {
-  return new Uint8Array(Float64Array.from([float]).buffer).reverse();
+export function sizedFloat(float, size, isBigEndian) {
+  if (size !== 32 && size !== 64) {
```
lpil, line 243: Let's use a bool as that forces it to be one or the other.

lpil (commented):
Awesome! Thank you! I've left some notes inline.

compiler-core/templates/prelude.mjs
richard-viney: I don't think that's an option because the size can be specified dynamically:

```gleam
  let a = int.random(1000)
  let b = <<1.0:float-size(a)>>
```

Both Erlang and JS throw an error if the size at runtime isn't supported for floats.

compiler-core/src/javascript/pattern.rs
richard-viney: Done as per previous comment

compiler-core/src/javascript/tests/bit_arrays.rs
richard-viney: Done!

richard-viney (commented):
Thanks for the review, all comments have been addressed or responded to, ready for another look 👍

compiler-core/templates/prelude.mjs
lpil: This function is currently invalid and can return undefined, it must be fixed. No code should depend on always being used correctly as then mistakes can result in cryptic and hard to debug incorrect behaviour instead of an easy to fix error.

lpil (commented):
Thank you for waiting! Sorry this is taking me so long

compiler-core/templates/prelude.mjs
richard-viney: For sure, did you see the throw that was added to this function last month to avoid returning undefined?

image

richard-viney:
Thanks, I've responded to your comment and rebased to fix a small merge conflict.

Whenever suits, please just let me know what else you think's needed here 👍

compiler-core/templates/prelude.mjs
lpil: Oh sorry, I did not. Thank you

compiler-core/src/pretty.rs
```diff
@@ -39,6 +39,12 @@ pub trait Documentable<'a> {
     fn to_doc(self) -> Document<'a>;
 }
 
+impl<'a> Documentable<'a> for bool {
+    fn to_doc(self) -> Document<'a> {
+        Document::Str(if self { "true" } else { "false" })
```
lpil, line 44: Sorry, one last thing. Please remove this as bools do not have a universal format.

lpil: I will do it!

lpil (approved):
Heroic effort!!! Thank you so much!

PR #3295, opened 2024-06-19
[LSP] Add parentheses to functions

Zhomart:
image

This is how Rust works. And I believe it also speeds up coding.
image

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -653,10 +653,14 @@ fn value_completion(
     value: &crate::type_::ValueConstructor,
     insert_range: Range,
 ) -> CompletionItem {
-    let label = match module_qualifier {
+    let mut label = match module_qualifier {
         Some(module) => format!("{module}.{name}"),
         None => name.to_string(),
     };
+    // Append "()" if the value is a function.
+    if let ValueConstructorVariant::ModuleFn { .. } = value.variant {
+        label.push_str("()");
+    }
 
     let type_ = Printer::new().pretty_print(&value.type_, 0);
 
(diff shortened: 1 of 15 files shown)


lpil (commented):
Doesn't this leave the cursor after the () instead of inside them?

movva-gpu:
> Doesn't this leave the cursor after the () instead of inside them?

I mean, isn't it possible to leave the cursor inside the ()?

Zhomart:
Tested on VSCode, it moves the cursor at the end.

Zhomart:
> I mean, isn't it possible to leave the cursor inside the ()?

This is good point, but I think we should add argument names and preselect them on autocomplete. Example: 

2024-06-19 at 22 25 21 - Red Eagle

movva-gpu:
Well, then the difference is that instead of pressing the `(` key, you will need to press the `←` key, so maybe it's not worth it?

movva-gpu:
> This is good point, but I think we should add argument names...

Ahhh maybe, I'm still sceptical

movva-gpu:
> Ahhh maybe, I'm still sceptical

The more I think about it, the more I like it tbh

Olian04:
> > I mean, isn't it possible to leave the cursor inside the ()?
> 
> This is good point, but I think we should add argument names and preselect them on autocomplete. Example from rust:
> 
> 2024-06-19 at 22 25 21 - Red Eagle

If we are to include this in the gleam LSP then we need to only add the arguments that needs to be filled in manually. In other words, for functions in a `|>` we should not add the first argument, and for functions in a `use` expression we should not add the final argument.

Zhomart:
What about this PR? Should we merge it or make it part of the "args" proposal ?

movva-gpu:
Why did you close it?

I think you should make it into some sort of beta, so multiple person uses it, then poll on the Discord to see what people think.

Zhomart:
> Why did you close it?
> 
> I think you should make it into some sort of beta, so multiple person uses it, then poll on the Discord to see what people think.

I found a bug :)

Also, I decided to impl [Olian04](https://github.com/Olian04)'s suggestion above, which makes more sense imho

movva-gpu:
Okay!

Issue #3286, opened 2024-06-17
Say "I was not expecting this keyword" in error when there is an unexpected keyword


mscharley:
I recently ran into the following ambiguous situation:

![image](https://github.com/gleam-lang/gleam/assets/336509/b31c5f16-55dd-4b8d-81fa-127976ba5e7a)

Here, `as` is being interpreted as a placeholder pattern by the syntax highlighter but the compiler is interpreting it as the `as` keyword and throwing it's hands up because that's not a valid place for that keyword. It would be great if the compiler message could include what it thought was there, not just what it expected to see.

mscharley:
Perhaps ambiguity is the wrong word, but as a human this could be either a variable name or the `as` keyword but the compiler (correctly) interprets it as the `as` keyword.

It would be nice if it said something like `I was not expecting an as keyword.` instead of `I was not expecting this.`. It isn't always clear from context what 'this' actually means.

mscharley:
More specifically, it says "Expected one of: a pattern" but to me, that's exactly what I wrote there - a placeholder pattern.

lpil:
Ah! Good suggestion. Thank you

PR #3299, opened 2024-06-20
Add Information About Found Token in UnexpectedToken Error Type

rdghosal:
Addresses #3286.

This commit introduces:
- A `found` member to `ParserErrorType::UnexpectedToken` of type struct `UnexpectedTokenInfo`
- An impl of std::fmt::Display in `UnexpectedTokenInfo` that specifies whether the unexpected token is a reserved word (i.e., keyword).

This commit changes:
- The `details` of `ParserErrorType::UnexpectedToken` to read `"I was not expecting this. Found {.found}, expected..."`
- In the case of a keyword, `details` of `ParserErrorType::UnexpectedToken` read as `"I was not expecting this. Found the keyword {.found}, expected..."` (see updated snapshots).

Would appreciate any and all feedback.
Thanks!

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -2813,7 +2813,7 @@ where
                 Token::UpName { .. } => {
                     parse_error(ParseErrorType::IncorrectName, SrcSpan { start, end })
                 }
-                _ if is_reserved_word(tok) => parse_error(
+                _ if tok.is_reserved_word() => parse_error(
                     ParseErrorType::UnexpectedReservedWord,
                     SrcSpan { start, end },
                 ),
@@ -3543,90 +3543,6 @@ fn parse_error<T>(error: ParseErrorType, location: SrcSpan) -> Result<T, ParseEr
 //
 // Misc Helpers
 //
-
-/// Returns whether the given token is a reserved word.
-///
-/// Useful for checking if a user tried to enter a reserved word as a name.
-fn is_reserved_word(tok: Token) -> bool {
-    match tok {
-        Token::As
-        | Token::Assert
-        | Token::Case
-        | Token::Const
-        | Token::Fn
-        | Token::If
-        | Token::Import
-        | Token::Let
-        | Token::Opaque
-        | Token::Pub
-        | Token::Todo
-        | Token::Type
-        | Token::Use
-        | Token::Auto
-        | Token::Delegate
-        | Token::Derive
-        | Token::Echo
-        | Token::Else
-        | Token::Implement
-        | Token::Macro
-        | Token::Panic
-        | Token::Test => true,
-
-        Token::Name { .. }
-        | Token::UpName { .. }
-        | Token::DiscardName { .. }
-        | Token::Int { .. }
-        | Token::Float { .. }
-        | Token::String { .. }
-        | Token::CommentDoc { .. }
-        | Token::LeftParen
-        | Token::RightParen
-        | Token::LeftSquare
-        | Token::RightSquare
-        | Token::LeftBrace
-        | Token::RightBrace
-        | Token::Plus
-        | Token::Minus
-        | Token::Star
-        | Token::Slash
-        | Token::Less
-        | Token::Greater
-        | Token::LessEqual
-        | Token::GreaterEqual
-        | Token::Percent
-        | Token::PlusDot
-        | Token::MinusDot
-        | Token::StarDot
-        | Token::SlashDot
-        | Token::LessDot
-        | Token::GreaterDot
-        | Token::LessEqualDot
-        | Token::GreaterEqualDot
-        | Token::LtGt
-        | Token::Colon
-        | Token::Comma
-        | Token::Hash
-        | Token::Bang
-        | Token::Equal
-        | Token::EqualEqual
-        | Token::NotEqual
-        | Token::Vbar
-        | Token::VbarVbar
-        | Token::AmperAmper
-        | Token::LtLt
-        | Token::GtGt
-        | Token::Pipe
-        | Token::Dot
-        | Token::RArrow
-        | Token::LArrow
-        | Token::DotDot
-        | Token::At
-        | Token::EndOfFile
-        | Token::CommentNormal
-        | Token::CommentModule
-        | Token::NewLine => false,
-    }
-}
 
 // Parsing a function call into the appropriate structure
 #[derive(Debug)]
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -241,12 +241,12 @@ where
         parse_result: Result<A, ParseError>,
     ) -> Result<A, ParseError> {
         let parse_result = self.ensure_no_errors(parse_result)?;
-        if let Some((start, found, end)) = self.next_tok() {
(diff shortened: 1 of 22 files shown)


compiler-core/src/parse.rs
```diff
@@ -3486,6 +3489,19 @@ fn is_reserved_word(tok: Token) -> bool {
...
+/// the token is a reserved word (i.e., keyword).
+fn format_found_token(tok: Token) -> String {
+    let base = tok.to_string();
+    if is_reserved_word(tok) {
+        format!("{base} (a keyword)")
+    }
+    else {
+        format!("{base}")
+    }
+}
```
lpil, line 3502: This is error presentation logic so it needs to live in the warning printer, and the data in the warning needs to be designed data structures rather than strings. Perhaps it could contain the token itself.

compiler-core/src/parse/snapshots/gleam_core__parse__tests__assignment_pattern_invalid_bit_segment.snap
```diff
@@ -8,6 +8,6 @@ error: Syntax error
 3 │     let <<b1, pub>> = <<24, 3>>
   │               ^^^ I was not expecting this
 
-Expected one of: 
+Found "pub" (a keyword), expected one of:
```
lpil, line 11:
```suggestion
Found the keyword `pub`, expected one of: 
```

lpil (commented):
Lovely! Thank you! I've left some small notes inline

compiler-core/src/parse/snapshots/gleam_core__parse__tests__assignment_pattern_invalid_bit_segment.snap
rdghosal: @lpil Thanks for the suggestion. Backticks are definitely better on the eyes here.

Should this be the case for non-keywords as well, or shall we keep double quotes? I ask because currently the `expected` token that gets printed is also wrapped in double quotes.

rdghosal: @lpil Pushed some updates per your feedback and kept the behavior of non-keywords the same. Lmk if that works... thanks!

compiler-core/src/parse/error.rs
```diff
@@ -397,3 +405,14 @@ impl LexicalError {
...
+impl fmt::Display for UnexpectedTokenInfo {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let base = self.display.replace("\"", "");
+        if self.is_keyword {
+            write!(f, "the keyword `{}`", base)
+        } else {
+            write!(f, "\"{}\"", base)
+        }
+    }
+}
```
lpil, line 418: Do not implement this using traits please. We avoid using traits, and this specifically is a misuse of Display as this isn't a general purpose to-string impl, it's specific to that one function.

compiler-core/src/parse/snapshots/gleam_core__parse__tests__capture_with_name.snap
```diff
@@ -8,6 +8,6 @@ error: Syntax error
 3 │   add(_name, 1)
   │       ^^^^^ I was not expecting this
 
-Expected one of: 
+Found "_name", expected one of:
```
lpil, line 11: Let's say "a name". They can already see what the name is 2 lines above

compiler-core/src/parse.rs
```diff
@@ -241,11 +241,15 @@ where
...
     ) -> Result<A, ParseError> {
         let parse_result = self.ensure_no_errors(parse_result)?;
-        if let Some((start, _, end)) = self.next_tok() {
+        if let Some((start, tok, end)) = self.next_tok() {
             // there are still more tokens
             let expected = vec!["An import, const, type, or function.".into()];
             return parse_error(
                 ParseErrorType::UnexpectedToken {
+                    found: UnexpectedTokenInfo {
+                        display: tok.to_string(),
```
lpil, line 250: Use an enum rather than a string here. It could be the token. I would probably put the entire token as `found` and not have this struct. The `is_reserved_word` can be moved to be a method on Token

rdghosal:
@lpil Thanks for all the detailed feedback! Pushed some updates per your comments... Do let me know if anything else appears off

compiler-core/src/parse.rs
```diff
@@ -3606,7 +3525,9 @@ pub fn make_call(
                 if name != "_" {
                     return parse_error(
                         ParseErrorType::UnexpectedToken {
+                            found: ParserErrorCause::Descriptor("a name".into()),
```
lpil, line 3528: We can get rid of the `ParserErrorCause` enum by making a token here!
```suggestion
                            found: Token::Name(name),
```

compiler-core/src/parse/snapshots/gleam_core__parse__tests__const_invalid_bit_array_segment.snap
```diff
@@ -8,6 +8,6 @@ error: Syntax error
 2 │ const a = <<1, 2, <->>
   │                   ^^ I was not expecting this
 
-Expected one of: 
+Found "<-", expected one of:
```
lpil, line 11: Where do these double quotes come from? I couldn't find them in the codebase 😅 

Could you replace them all with backticks please, to match the keywords.

compiler-core/src/parse/snapshots/gleam_core__parse__tests__function_invalid_signature.snap
```diff
@@ -0,0 +1,13 @@
...
+source: compiler-core/src/parse/tests.rs
+expression: "\nfn f(a, \"b\") -> String {\n    a <> b\n}\n"
+---
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:9
+  │
+2 │ fn f(a, "b") -> String {
+  │         ^^^ I was not expecting this
+
+Found "b", expected one of:
```
lpil, line 11: For any token with content can you say "a string", "a float", "an int" etc rather than display the content a second time in the error. Thank you

lpil (commented):
Looking great!!

compiler-core/src/parse.rs
rdghosal: Oh, man! This was so obvious... thanks for pointing that out!

rdghosal:
@lpil Pushed up a few more updates per your feedback. Thank you!

lpil (commented):
Beautiful work, thank you. Would you mind updating the changelog file also?

lpil (approved):
Beautiful work, thank you. Would you mind updating the changelog file also?

lpil:
We don't permit merge commits in this repo. Would you like to rebase to remove the ones from this branch, or should I squash-merge this PR?

lpil (approved):
Wonderful!! Thank you very much!!

lpil:
I think something may have gone a bit wrong with the rebase, there's multiple copies of each commit in the history, and there's conflicts preventing a rebase into main. I'll squash merge for this one. Thank you

Issue #2957, opened 2024-04-11
LSP code action: add missing patterns to case expression


lpil:
The exhaustiveness checker has this information so we need to expose it to the language server

It may be challenging to work out in the exhaustiveness checker if the variant names need to be qualified or not. It would be annoying to have to fix it every time.

Perhaps this needs to come after https://github.com/gleam-lang/gleam/issues/2952

GearsDatapacks:
I'll try and implement this once #3393 is merged, although we should probably fix #2426 first too

PR #3302, opened 2024-06-20
LSP Code Action: Convert `let assert` into a case expression

### LINKED ISSUE #2958: LSP code action: convert `let assert Ok` into case
```gleam
let assert Ok(x) = blah
```
into
```gleam
let x = case blah {
  Ok(x) -> x
  Error(e) -> panic as "value was not ok"
}
```

GearsDatapacks:
Closes #2958 
Implements a code action which converts:
```gleam
let assert Ok(value) = get_result()
```
into:
```gleam
let value = case get_result() {
  Ok(value) -> value
  _ -> panic
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,13 @@
 
   ([Gears](https://github.com/gearsdatapacks))
 
+## v1.3.2 - 2024-07-11
+
+### Language Server
+
+- The language server no longer shows completions when inside a literal string.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Bug Fixes
 
 - Fixed a bug where the compiler would report errors for duplicate `@external`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,7 +116,7 @@
 - The language server no longer shows completions when inside a literal string.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
-- LSP can now suggest a code action to convert `let assert Ok` into a case expression:
+- LSP can now suggest a code action to convert `let assert` into a case expression:
 
   ```
   let assert Ok(foo) = bar()
@@ -127,7 +127,7 @@
   ```
   let foo = case bar() {
     Ok(foo) -> foo
-    Error(e) -> panic as "value was not ok"
+    _ -> panic
   }
   ```
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -119,14 +119,14 @@
 - LSP can now suggest a code action to convert `let assert` into a case expression:
 
   ```
-  let assert Ok(foo) = bar()
+  let assert Ok(value) = get_result()
   ```
 
   Becomes:
 
   ```
-  let foo = case bar() {
-    Ok(foo) -> foo
+  let value = case get_result() {
+    Ok(value) -> value
     _ -> panic
   }
   ```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@
 
   For example, provided a User type:
 
-  ```
+  ```gleam
   pub type User {
     name: String
   }
@@ -84,14 +84,30 @@
 - The language server now suggests a code a action to rename variables, types and
   functions when they don't match the Gleam naming requirements:
 
-  ```
+  ```gleam
   let myNumber = 10
   ```
 
   Becomes:
 
-  ```
+  ```gleam
   let my_number = 10
+  ```
+
+- The language server can now suggest a code action to convert `let assert` into
+  a case expression:
+
+  ```gleam
+  let assert Ok(value) = get_result()
+  ```
+
+  Becomes:
+
+  ```gleam
+  let value = case get_result() {
+    Ok(value) -> value
+    _ -> panic
+  }
   ```
 
   ([Gears](https://github.com/gearsdatapacks))
@@ -108,30 +124,6 @@
 
 - Fixed a bug where a private type could be leaked from an internal module.
   ([Ameen Radwan](https://github.com/Acepie))
-
-## v1.3.2 - 2024-07-11
-
-### Language Server
-
-- The language server no longer shows completions when inside a literal string.
-  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
-
-- LSP can now suggest a code action to convert `let assert` into a case expression:
-
-  ```
-  let assert Ok(value) = get_result()
-  ```
-
-  Becomes:
-
-  ```
-  let value = case get_result() {
-    Ok(value) -> value
-    _ -> panic
-  }
-  ```
-
-  ([Gears](https://github.com/gearsdatapacks))
 
 ## v1.3.2 - 2024-07-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,23 @@
 - The language server no longer shows completions when inside a literal string.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- LSP can now suggest a code action to convert `let assert Ok` into a case expression:
+
+  ```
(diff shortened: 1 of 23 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -276,3 +276,89 @@ impl<'a> RedundantTupleInCaseSubject<'a> {
...
+            name,
+            arguments,
+            type_,
+            ..
+        } = &assignment.pattern
+        else {
+            return;
+        };
+
+        if type_.is_result() && name == "Ok" {
```
lpil, line 309: I think we would like this to work for any pattern rather than just `Ok`. Do you think you'd be up for implementing that?

You could get the span of the pattern and copy the string contents of that range to make the edit.

We could either for now we could put a `todo` on the right hand side of the `->`, or do the work to figure out all the variables that get bound and then make a tuple of them. Might not be too tricky using another visitor applied to the pattern.

lpil (commented):
Very nice!! Thank you. I've left a note inline

compiler-core/src/language_server/code_action.rs
GearsDatapacks: Yep, I can do that. Can you give an example of how it would work with capturing all the variables? I don't completely get what you mean by that.

GearsDatapacks: Oh I think I see. You just mean for the pattern that's asserted? Not for the other possible cases

lpil: Some examples:
---
```gleam
let assert Ok(x) = y
```
would need to generate
```gleam
let x = case y {
  Ok(x) -> x
  _ -> panic
}
```
---
```gleam
let assert [x, 1, y] = z
```
would need to generate
```gleam
let #(x, y) = case z {
  [x, 1, y] -> #(x, y)
  _ -> panic
}
```

GearsDatapacks: Right. So we would just do the `_` case for other options, no inference of other patterns? That makes it much easier to implement. Maybe once #2957 is complete, we can add that functionality.

GearsDatapacks:
I've made some changes:
- This code action now applies to any pattern
- There didn't seem to be a `visit_pattern` method in the ast visitor, nor seemingly is the information of assigned variables stored in the AST, so I wrote my own pattern visitor function, like is done in various other places. Please let me know if I should implement it it differently
- Some of the location information of various patterns (namely alias and list) was incorrect, I've now fixed those.
  - The only tested behaviour that changes from this is the location of error reporting, which doesn't really affect behaviour, so I accepted the changes.

CHANGELOG.md
```diff
@@ -130,6 +130,23 @@
   better/more accurate insertions.
   ([Ameen Radwan](https://github.com/Acepie)
 
+- LSP can now suggest a code action to convert `let assert` into a case expression:
+
+  ```
+  let assert Ok(foo) = bar()
```
lpil, line 136: Please remove all foobar stuff, it's a rather grim ww2 reference.

compiler-core/src/language_server/code_action.rs
```diff
@@ -276,3 +276,148 @@ impl<'a> RedundantTupleInCaseSubject<'a> {
...
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+        self.actions
+    }
+
+    // Recursively traverses a pattern and finds which variables get bound
+    fn traverse_pattern(&mut self, pattern: &TypedPattern) {
+        println!("{pattern:#?}");
```
lpil, line 372: Debug logging here!

compiler-core/src/language_server/code_action.rs
```diff
@@ -276,3 +276,148 @@ impl<'a> RedundantTupleInCaseSubject<'a> {
...
+        }
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+        self.actions
+    }
+
+    // Recursively traverses a pattern and finds which variables get bound
+    fn traverse_pattern(&mut self, pattern: &TypedPattern) {
```
lpil, line 371: How about we make a trait for visiting patterns? Then we wouldn't need to duplicate this code each time

lpil (commented):
Looking fantastic!! Couple small notes but nearly there

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__case20.snap
```diff
@@ -6,7 +6,7 @@ error: Type mismatch
   ┌─ /src/one/two.gleam:1:18
   │
 1 │ case [1] { [x] | [] as x -> 1 }
-  │                  ^
+  │                  ^^
```
lpil, line 9: Good find!

compiler-core/src/language_server/code_action.rs
GearsDatapacks: Should I add the pattern visiting code to the ast visitor, or create a separate trait?

GearsDatapacks:
I have now added the pattern visiting methods to the AST visitor trait

compiler-core/src/language_server/tests/action.rs
```diff
@@ -338,6 +340,122 @@ pub fn main() {
     );
 }
 
+#[test]
+fn test_convert_let_assert_to_case() {
+    insta::assert_snapshot!(apply_first_code_action_with_title(
+        "
+pub fn main() {
+  let assert Ok(foo) = Ok(1)
```
lpil, line 348: Remove all the foobar stuff please, it's a rather grim WW2 reference

compiler-core/src/language_server/code_action.rs
```diff
@@ -276,3 +276,131 @@ impl<'a> RedundantTupleInCaseSubject<'a> {
...
+
+        let edit = TextEdit {
+            range,
+            new_text: format!(
+                "let {assigned} = case {expr} {{
+{indent}  {pattern} -> {assigned}
+{indent}  _ -> panic
+{indent}}}",
+                // "_" isn't a valid expression, so we just return Nil from the case expression
+                assigned = if assigned == "_" { "Nil" } else { assigned }
```
lpil, line 340: I don't think `_` is ever possible to be used as a variable name, and we're missing the case where no variables are assigned at all, e.g. `let assert [] = x`. Could we have a test for that please

lpil (commented):
Super close! Couple tiny things below

GearsDatapacks:
With the case of no variables, I somehow overlooked a mistake in my implementation. That should now be fixed.

CHANGELOG.md
```diff
@@ -2,15 +2,24 @@
 
 ## Unreleased
 
-### Build tool
+### Language Server
```
lpil, line 5: Could you put back the other headings and rebase on main to resolve the conflict please 🙏

lpil (commented):
💜 💜 💜

CHANGELOG.md
GearsDatapacks: Done :)

GearsDatapacks:
The tests failing on macos seems to be a network issue?

lpil (approved):
Thank you!!!!!

oops, I think you've deleted some of the changelog accidentally. I'll fix it

Yeah idk why macOS fails to work on GitHub some times

lpil:
Sorry, I've found an issue while testing.

If I have this code:
```gleam
  let assert [z] = {
    let assert [y] = {
      x
    }
    y
  }
```

And I try to expand the inner one it instead expands the outer one, and then it doesn't offer the code action for the inner one at all.

I think to resolve ambiguity it should only offer the code action if the section/cursor is between `let` and `=`.

I'm not sure why it doesn't offer the code action the second time round.

lpil:
Wow I made a mess of this rebase 😅

PR #3305, opened 2024-06-21
Add else to Erlang's reserved keywords

### LINKED ISSUE #3304: "Else" type variant causes erlang troubles
It looks like I hit an edge case. 

```rs
pub type Example {
  Else
} 
```

I don't know if this will minimally repoduce the problem, but erlang doesn't seem to like using the "else" token.

giacomocavalieri:
This PR closes #3304

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,11 @@
   reserved word in Erlang/OTP 27.
   ([Jake Barszcz](https://github.com/barszcz))
 
+- Functions, types and constructors named `maybe` and `else` are now
+  escaped in generated Erlang to avoid clashing with Erlang's keywords.
+  ([Jake Barszcz](https://github.com/barszcz)) and
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 - Non byte aligned arrays that use literals for size are now marked as an
   Unsupported feature for Javascript since they would already cause
   a runtime error on Javascript.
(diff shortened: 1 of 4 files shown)


compiler-core/src/erlang/tests.rs
```diff
@@ -632,3 +632,19 @@ pub fn main() {
 "
     );
 }
+
+// https://github.com/gleam-lang/gleam/issues/3304
+#[test]
+fn type_named_else() {
```
lpil, line 638: Another test for functions too please

CHANGELOG.md
```diff
@@ -132,6 +132,10 @@
   JavaScript.
   ([Louis Pilfold](https://github.com/lpil))
 
+- Fixed a bug where calling a type or a constuctor `Else` would generate invalid
+  erlang code.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
```
lpil, line 137: Could you merge this with the `maybe` note above please 🙏

lpil (commented):
Thank you!

compiler-core/src/erlang/tests.rs
giacomocavalieri: `else` is a keyword in gleam as well so we can't have a function named like that

CHANGELOG.md
giacomocavalieri: Ok!

compiler-core/src/erlang/tests.rs
lpil: oh yes!!!

lpil (approved):
Thank you again!

PR #3307, opened 2024-06-21
Implemented extra newline after first import Issue #3271

codemonkey76:


compiler-core/src/language_server/completer.rs
```diff
@@ -586,30 +596,28 @@ where
...
-    fn first_import_line_in_module(&'a self) -> Position {
-        let import_location = self
-            .module
-            .ast
-            .definitions
-            .iter()
-            .find_map(get_import)
-            .map_or(0, |i| i.location.start);
+    fn first_import_line_in_module(&'a self) -> (bool, Position) {
+        let import_location = self.module.ast.definitions.iter().find_map(get_import);
```
lpil, line 600: Rather than iterate through all the definitions please just check if the first one is an import, and always place the import as the first thing in the module. Thank you.

GearsDatapacks:
I think this PR should be closed, since the issue in question has already been fixed by another PR

lpil:
Thank you both!

Issue #2993, opened 2024-04-16
Display qualified type names on LSP hover


giacomocavalieri:
When you hover over a variable the LSP can show its type. In this example you can see that it shows `List(Error)`, I think the LSP should show the `Error` type as qualified since it's not defined in the same module and I didn't import It qualified from the module it's defined in. So what I'd like to see here is `List(rule.Error)`
Not sure how feasible this is and we would probably want to make exceptions for prelude types but I think it could make the type on hover more helpful

![The error type should be qualified here since I didnt import unqulified and is not defined here](https://github.com/gleam-lang/gleam/assets/20598369/eda34e8b-14da-4c15-a77b-75dcd550dea2)

Acepie:
Posting here for whoever works on this (I might work on this but wanna see if others have thoughts). The main difficulty with implementing this is that the types that the LSP prints are created using the same type pretty printer that the docs and error messages use. [Currently](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/type_/pretty.rs#L59) that pretty printer only qualifies types if there would be a collision with an existing type that has already been printed. Making this change would require changing the behavior in a way that still lets those work the same as today (maybe by allowing the caller to pass in modules to never qualify and defaulting to qualify unless the module is explicitly excepted)

lpil:
There are multiple LSP related tickets open that want a context aware printer, so making one and switching over all the LSP stuff sounds good to me.

ascandone:
I'd like to give it a try. I've got a WIP here https://github.com/gleam-lang/gleam/pull/3312

PR #3290, opened 2024-06-18
Lsp/inlay hints on pipe

ascandone:
Implements https://github.com/gleam-lang/gleam/issues/3291

screen

This PR shows inlay hints with the type of a pipeline step (_only_ for pipeline expressions that span across multiple lines).  Hints won't be shown for literal values (ints, floats, strings, bitstrings
The functionality is toggled via the `gleam.inlayHints.pipelines` client config option

The PR is meant to be read as a whole (not commit by commit - however the config functionality is implemented in a single commit: https://github.com/gleam-lang/gleam/pull/3290/commits/9e12629f2b9461f140fcb476cac937a906f087b3)

Here's the vscode client PR: https://github.com/gleam-lang/vscode-gleam/pull/82

#### Vscode
Once the [vscode client pr](https://github.com/gleam-lang/vscode-gleam/pull/82) is merged, it will be possible to toggle the option via the vscode settings

#### Sublime text
Screenshot 2024-07-04 at 09 25 56

The user will need to enable the functionality in the settings of the LSP sublime extension:

```json
{
  "clients": {
    "gleam": {
      "settings": {
        "gleam.inlayHints.pipelines": true
      }
    }
  },
}
```

#### Nvim

image

The user will need to enable the functionality in the `init.lua` file:
```lua
require'lspconfig'.gleam.setup{
  settings = {
    gleam = {
      inlayHints = {
        pipelines = true,
      },
    },
  },
}
```

### Future development
* handle hover on inlay hints (e.g. show type doc)
* handle click on inlay hints (e.g. goto definition)

### Credits
The part of the pr about config is partially based upon this previous pr https://github.com/gleam-lang/gleam/pull/2393

PR #3312, opened 2024-06-22
Context aware type pprint

ascandone:
Implements https://github.com/gleam-lang/gleam/issues/2993

## TODOS
* [ ] handle type aliases

### Open questions
Should hovered exported types be context-aware pprinted? e.g.
```gleam
// Suppose X has type T
import {type T as Renamed, X}
// What should we show when hovering on `X` ?
```

### Specs
Handled cases (assume the current file is `my_module.gleam`:

**1) types in imported modules**
```gleam
import some_module
some_module.Type // <- pprinted type
```

**2) types in module that are not imported**
```gleam
// let's say that the type is in the `some_module` module
some_module.Type // <- pprinted type
```

**3) types in module that are imported and renamed**
```gleam
import some_module as renamed_module
renamed_module.Type // <- pprinted type
```

**4) types in modules that are imported and exposed as unqualified imports**
```gleam
import some_module.{type Type}
Type // <- pprinted type
```

**5) types in modules that are imported and exposed as unqualified imports, and renamed**
```gleam
import some_module.{type Type as RenamedType}
RenamedType // <- pprinted type
```

**6) types defined locally**
```gleam
type Type {}
Type // <- pprinted type
```

**7) types from the gleam prelude**
```gleam
Int // <- pprinted type
```

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -478,7 +478,7 @@ where
                     Some(hover_for_function_head(fun, lines, module))
                 }
                 Located::ModuleStatement(Definition::ModuleConstant(constant)) => {
-                    Some(hover_for_module_constant(constant, lines, &module))
+                    Some(hover_for_module_constant(constant, lines, module))
                 }
                 Located::ModuleStatement(_) => None,
                 Located::UnqualifiedImport(UnqualifiedImport {
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -747,7 +747,7 @@ fn hover_for_pattern(pattern: &TypedPattern, line_numbers: LineNumbers) -> Hover
     let documentation = pattern.get_documentation().unwrap_or_default();
 
     // Show the type of the hovered node to the user
-    let type_ = Printer::new().pretty_print(pattern.type_().as_ref(), 0);
+    let type_ = printer_from_module(module).pretty_print(pattern.type_().as_ref(), 0);
     let contents = format!(
         "```gleam
 {type_}
@@ -791,8 +791,12 @@ fn hover_for_function_head(
     }
 }
 
-fn hover_for_function_argument(argument: &TypedArg, line_numbers: LineNumbers) -> Hover {
-    let type_ = Printer::new().pretty_print(&argument.type_, 0);
+fn hover_for_function_argument(
+    argument: &Arg<Arc<Type>>,
+    module: &Module,
+    line_numbers: LineNumbers,
+) -> Hover {
+    let type_ = printer_from_module(module).pretty_print(&argument.type_, 0);
     let contents = format!("```gleam\n{type_}\n```");
     Hover {
         contents: HoverContents::Scalar(MarkedString::String(contents)),
@@ -821,12 +825,6 @@ fn hover_for_annotation(
         contents: HoverContents::Scalar(MarkedString::String(contents)),
         range: Some(src_span_to_lsp_range(location, &line_numbers)),
     }
-}
-
-fn printer_from_module(module: &Module) -> Printer {
-    let mut printer = Printer::new();
-    printer.with_imports_context(module.name.clone(), (&module.ast).into());
-    printer
 }
 
 fn hover_for_module_constant(
@@ -863,7 +861,7 @@ fn hover_for_expression(
         .unwrap_or("".to_string());
 
     // Show the type of the hovered node to the user
-    let type_ = Printer::new().pretty_print(expression.type_().as_ref(), 0);
+    let type_ = printer_from_module(module).pretty_print(expression.type_().as_ref(), 0);
     let contents = format!(
         "```gleam
 {type_}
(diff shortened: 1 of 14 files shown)


lpil:

ascandone:
> Hello! Are you still working on this?

👋 I am blocked due to the seemingly complex problem of printing aliases of types instead of the original definition.
Here's an example of the issue I'm referring to: https://github.com/gleam-lang/gleam/pull/3312/commits/2e4409f4790fa06a5715eb191f4db55943a0cb9d

I think the desidered behaviour would be to pprint the alias instead of the actual type definition. This is useful for the pattern of re-exporting private types as alias, like in lustre:
Screenshot 2024-09-20 at 22 35 45
As you can see, the attribute will be pprinted as `lustre/internals/vdom.Attribute` instead of `lustre/attributes.Attribute`

However, fixing this doesn't look trivial:
here's the value of the `AliasedType` of the commit that I linked before:
`Named { publicity: Public, package: "app", module: "html/internal", name: "Attribute", args: [] }`
This value is passed as type in the `fn print<'a>(&mut self, type_: &Type) -> Document<'a> ` function so I don't even know if it's possible to workaround it

### Proposed solution

`pretty::print(t)` returns `TypeAlias(p1', .., pn')` if exists a type (either defined locally or imported)
```gleam
type TypeAlias(p1, .., pn) = MyType(a1, .., an)
```
such that `unify(MyType(a1, .., an), t)` yields the substitution `{ p_i ' => t_i forall i in 1, .., n }` (note that Gleam currently forbids `p_i`not to occur inside `MyType(a1, .., an)`).
If such substitution does not exist, just print `t` normally

**example 1**
```gleam
// internal
pub type PubAttribute(a) =
  internal.Attribute(a)

pub fn checked() -> PubAttribute(String) { todo }
```
when printing `PubAttribute(String)`, we actually receive `internal.Attribute(a)`. This value can be unified with the value of the `PubAttribute` alias, with the `a => String` substitution. Therefore we'll print `PubAttribute(String)`, as expected.

**example 1 bis**
```gleam
// internal
pub type PubAttribute(a) =
  internal.Attribute(a)

pub fn checked() -> internal.Attribute(String) { todo } // <- note aren't returning the alias
```
This behaves in the same way as previous example. This might likely be counter-intuitive; however it looks like ocaml behaves similarly, and in some cases it might even be handy

**example 2**
```gleam
// internal
pub type PubAttribute(a, b) =
  internal.Attribute(Result(b, a))

pub fn checked() -> PubAttribute(Int, Bool) { todo }
```
Applying the alias definition, we get `internal.Attribute(Result(Bool, Int))`. This value can be unified with the value of the `PubAttribute` alias, with the `a => Bool, b => Int` substitution. Therefore we'll print `PubAttribute(Bool, Int)`

**example 3**
```gleam
// internal
pub type PubAttribute(a) =
  internal.Attribute(Option(a))

pub fn checked() -> internal.Attribute(Int) { todo }
```
In this case, `internal.Attribute(Int)` cannot be unified with the alias definition, therefore is printed normally

ascandone:
Oh and btw, would it be possible to try to merge [this one](https://github.com/gleam-lang/gleam/pull/3290) first? It's done except for this couple of doubts that I mentioned, could you give me an hint with that?

ascandone:
Another solution could also consist in actually adding a "TypeAlias" variant to the [Type](https://github.com/gleam-lang/gleam/blob/ab8cb13c6bd031ed348bdacee764405e6e4f7cde/compiler-core/src/type_.rs#L50) enum
Roughly like
```rust
enum Type {
  // ...
  TypeAlias {
       // same as "Named"
        publicity: Publicity,
        package: EcoString,
        module: EcoString,
        name: EcoString,
        args: Vec<Arc<Type>>,
     
        // but also have the type it resolves to
        // so that unify() can recur on it
        value: Arc<Type>,
    },
}
```

I think that's the way Elm manages to have very good messages regarding type aliases
https://github.com/elm/compiler/blob/2f6dd29258e880dbb7effd57a829a0470d8da48b/compiler/src/Type/Type.hs#L93

lpil:
> I think the desidered behaviour would be to pprint the alias instead of the actual type definition.

We don't want to do this as it results in very poor error messages and hover information- the programmer can't see what the type is so they don't know how to debug their program.

Context aware type printing is implemented [here](https://github.com/gleam-lang/gleam/blob/ab8cb13c6bd031ed348bdacee764405e6e4f7cde/compiler-core/src/type_/printer.rs) so that code can be reused here.

ascandone:
> > I think the desidered behaviour would be to pprint the alias instead of the actual type definition.
> 
> We don't want to do this as it results in very poor error messages and hover information- the programmer can't see what the type is so they don't know how to debug their program.
> 
> Context aware type printing is implemented [here](https://github.com/gleam-lang/gleam/blob/ab8cb13c6bd031ed348bdacee764405e6e4f7cde/compiler-core/src/type_/printer.rs) so that code can be reused here.

Got it! Just another thing, isn't this other [PR](https://github.com/gleam-lang/gleam/pull/3599) implementing the same thing? Should I close mine?

GearsDatapacks:
The other PR has now been merged, so I guess this one should be closed now

ascandone:
yeah sorry about the mess :)

Issue #3262, opened 2024-06-11
Allow `gleam add` to take a version specifier


mscharley:
It would be really helpful if `gleam add` could take a version specifier along with a package name and use that to limit what it installs.

My initial syntax proposal would be `gleam add gleam_json@1.0.1`. This would be a non-breaking change since you could just not specify a version and things would work as before - it is a purely additive change.

What I propose this should do is replace the version that gleam determines. eg, the current latest version of `gleam_json` is `2.0.0`. Instead of selecting the latest version, gleam would instead select the version specified. The initial version written into `manifest.toml` would be the version specified, and the version guard written into `gleam.toml` would be the same derived version used currently, but based on the specified version instead of the latest version, eg. `>= 1.0.1 and < 2.0.0`.

lpil:
Sounds good! It would be nice to be able to specify just a major or major and minor version too.

```
gleam add gleam_json@1
gleam add gleam_json@1.2
```

GearsDatapacks:
I think #3316 should have closed this

lpil:
Thank you

PR #3316, opened 2024-06-22
Handle Version Specifiers at Package Download

rdghosal:
Addresses #3262.

### Adds:
- A helper to parse an optionally provided version specifier that is applied at package download.
- Unit tests to validate whether the specifier is PubGrub-compliant.

### Notes:
The solution here deviates from the original suggestion, in terms of what delimiters are expected as follows:

- Suggested:
```bash
$ gleam add my_package@1.0.0
```

- Implemented:
```bash
$ gleam add my_package==1.0.0
```

I chose this implementation as an enhancement to the original proposal that allows the user to provide a more flexible version specifier (e.g., `<~ 1.0.0`, `<= 1.0.0`).

Needless to say, this PR is open to feedback, and the implementation here can also be made to align with the original suggestion.
Careful review of the testing methodology would especially be appreciated.

Thank you!

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -151,7 +151,9 @@ fn get_hexpm_version(package: &str) -> Result<Requirement> {
                 3 => Ok(Requirement::hex(version)),
                 n_parts => Err(Error::InvalidVersionFormat {
                     input: version.to_string(),
-                    error: format!("Expected up to 3 version numbers in specifier (MAJOR.MINOR.PATH), found {n_parts}"),
+                    error: format!(
+                        "Expected up to 3 numbers in version specifier (MAJOR.MINOR.PATCH), found {n_parts}"
+                    ),
                 }),
             }
         }
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -118,7 +118,7 @@ pub fn update() -> Result<()> {
     Ok(())
 }
 
-fn get_hexpm_version(package: &str) -> Result<Requirement> {
+fn parse_hex_requirement(package: &str) -> Result<Requirement> {
     match package.find('@') {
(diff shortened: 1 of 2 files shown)


lpil:
Hello! Please implement the syntax in the issue. I think it should have this behaviour:

```shell
gleam add lustre@1.2.3  # "1.2.3"
gleam add lustre@1.2   # ">= 1.2.0 and < 2.0.0"
gleam add lustre@1     # ">= 1.0.0 and < 2.0.0"
```

rdghosal:
@lpil Thanks for the feedback and clear examples. The PR has been updated accordingly. 

Please let me know if there's anything else we should address.
Thank you!

compiler-cli/src/dependencies.rs
```diff
@@ -118,6 +118,77 @@ pub fn update() -> Result<()> {
...
+    assert!(get_hexpm_version("package_1@1.2.3.4").is_err());
+    assert!(get_hexpm_version("package_2@not_a_version").is_err());
+
+    // Good versions.
+    let packages = vec![
+        ("package_3", ">= 0.0.0"),
+        ("package_4@1", ">= 1.0.0 and < 2.0.0"),
+        ("package_5@1.2", ">= 1.2.0 and < 2.0.0"),
+        ("package_6@1.2.3", "1.2.3"),
+    ];
```
lpil, line 178: One example per test please 🙏

rdghosal:
@lpil Thanks for the great feedback! Updated the PR per your comments with an update to the changelog as well.

lpil (approved):
Beautiful!! Thank you so much!

rdghosal:
@lpil Thanks for the review! I'm seeing a lint job fail... shall I push up another commit to this PR? Seems like a straightforward fix.

lpil:
I fixed it on main :)

PR #3318, opened 2024-06-23
Support constant string concatenation

### LINKED ISSUE #3208: Allow string concatenation when defining const values
This is currently not valid syntax:

```gleam
pub const foo = "foo"
pub const foo_bar = foo <> "bar"
```

It would be great if this would work!

DeviousStoat:
Implements support for string concatenation in module constants' expressions.

```gleam
const foo = "foo"
const foobar = foo <> "bar"
const foofoobarbaz = foo <> foobar <> "baz"
```

closes #3208


compiler-core/src/erlang.rs
```diff
@@ -512,6 +512,37 @@ fn tuple<'a>(elems: impl IntoIterator<Item = Document<'a>>) -> Document<'a> {
...
+            match &constructor.variant {
+                ValueConstructorVariant::ModuleConstant { literal: Constant::String { value, ..}, .. } => docvec!['"', string_inner(value), "\"/utf8"],
+                ValueConstructorVariant::ModuleConstant { literal: concat_value @ Constant::StringConcatenation { .. }, .. } => docvec![const_inline(concat_value, env), "/binary"],
+                _ => panic!("Constant string concatenation argument can only be a string or const var at code generation time"),
+            }
+        }
+
+        Constant::StringConcatenation { .. } => docvec![const_inline(value, env), "/binary"],
+
+        _ => panic!("Constant string concatenation argument can only be a string or const var at code generation time"),
```
lpil, line 542: Please remove these panics, code generation is not the place to do type checking, and other variants could be suitable in future. Calling `const_inline` on whatever the `Constant::*` is is good.

compiler-core/src/erlang/tests/strings.rs
```diff
@@ -438,3 +438,32 @@ pub fn main(x) {
 "#,
     );
 }
+
+#[test]
+fn assert_const_concat() {
```
lpil, line 443: Could you remove all "foobar" please, it's a rather depressing ww2 reference.

compiler-core/src/erlang/tests/strings.rs
```diff
@@ -438,3 +438,32 @@ pub fn main(x) {
...
+    );
+}
+
+#[test]
+fn assert_const_concat_multiple() {
+    assert_erl!(
+        r#"
+const foo = "foo"
+const foobar = foo <> "bar"
+const foofoobarbaz = foo <> foobar <> "baz"
```
lpil, line 462: Split this into multiple tests please, we only test one example per test.

Could we also have a test for a much larger number of concatenations in one constant too please.

compiler-core/src/format/tests.rs
```diff
@@ -6156,3 +6156,19 @@ fn newlines_are_not_stripped_if_two_consecutive_anonymous_function_are_passed_as
...
+        r#"const long_string = "some"
+  <> " very"
+  <> " long"
+  <> " string"
+  <> " indeed"
+  <> " please"
+  <> " break"
+"#
+    );
+}
```
lpil, line 6174: Can we have tests for unbroken concatenation expressions and also something like `const x = "some long string 1" <> "some long string 2" <> ["here is a list", "with several elements", "in order to make it be too long to fit on one line", "so we can see how it breaks", "onto multiple lines"] <> "and a last string`

compiler-core/src/type_/tests/errors.rs
```diff
@@ -1905,3 +1905,13 @@ fn leak_multiple_private_types() {
         "
     );
 }
+
+#[test]
+fn const_string_concat_invalid_type() {
+    assert_module_error!(
+        "
+const foo = 5
+const bar = foo <> \"bar\"
```
lpil, line 1914: Remove the foobar please 🙏

compiler-core/src/javascript/tests/strings.rs
```diff
@@ -224,3 +224,32 @@ pub fn go(x) {
 "#,
     )
 }
+
+#[test]
+fn const_concat() {
+    assert_js!(
+        r#"
+const foo = "foo"
```
lpil, line 232: Remove the foobar please 🙏

lpil (commented):
Lovely work! I've left some notes inline. Thank you

DeviousStoat:
I think I addressed your comments.
Btw, don't know if it matters too much as I haven't worked on too many big rust projects but I think this codebase is really well made, it's a joy to work on it and easy to pickup, akin to the gleam language itself

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__assert_const_concat_many_strings.snap
```diff
@@ -0,0 +1,37 @@
...
+expression: "\nconst big_concat = \"a\" <> \"b\" <> \"c\" <> \"d\" <> \"e\" <> \"f\" <> \"g\" <> \"h\" <> \"i\" <> \"j\" <> \"k\" <> \"l\" <> \"m\" <> \"n\" <> \"o\" <> \"p\" <> \"q\" <> \"r\" <> \"s\" <> \"t\" <> \"u\" <> \"v\" <> \"w\" <> \"x\" <> \"y\" <> \"z\"\n\npub fn main() {\n  big_concat\n}\n"
+---
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).
+
+-export([main/0]).
+
+-spec main() -> binary().
+main() ->
+    <<"a"/utf8,
```
lpil, line 12: This should render like this:

```erlang
main() ->
    <<"a"/utf8,
      "b"/utf8,
      "c"/utf8,
      ...
```

compiler-core/src/parse/tests.rs
```diff
@@ -1121,3 +1121,22 @@ fn newline_tokens() {
...
+const bar = foo <> \"bar\"
+"
+    );
+}
+
+#[test]
+fn const_string_concat_naked_right() {
+    assert_module_error!(
+        "
+const foo = \"foo\" <>
```
lpil, line 1139: Remove the foobar from this file please 🙏

lpil (commented):
Oh that you, you are very kind! 💜 

This is looking great! Thank you. 2 small thing in the notes

compiler-core/src/parse/tests.rs
DeviousStoat: oops sorry missed this one

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__assert_const_concat_many_strings.snap
DeviousStoat: ooh right, it's not just format problem, we don't need to `/binary` the concatenation, we can build a bit array of `/utf8` strings. This is a lot nicer now, I am starting to understand what you meant with "string concatenation doesn't require compile time code execution"

compiler-core/src/erlang.rs
```diff
@@ -512,6 +512,71 @@ fn tuple<'a>(elems: impl IntoIterator<Item = Document<'a>>) -> Document<'a> {
...
+    elems: impl IntoIterator<Item = Document<'a>>,
+) -> Document<'a> {
+    join(elems, break_(",", ", "))
+        // We want to align the utf8 string parts of the bit array
+        // <<"a"/utf8,
+        //   "b"/utf8>>
+        // instead of
+        // <<"a"/utf8,
+        // "b"/utf8>>
+        .set_nesting(INDENT + 2)
```
lpil, line 525: This will be wrong if the indent level is supposed to be anything other than 1!

For example:

```gleam
const x = [
  <<
    ...,
    ...,
  >>
]
```

I don't mind so much about the exact format used, but the indentation should increase as there's more nesting

DeviousStoat: oh right, I pushed a change with a test case for a long concatenation in a list.
Let me know what you think, it doesn't exactly has the format that you specified before but the indentation should be good now.

I think to render like that:
```erl
main() ->
    <<"a"/utf8,
      "b"/utf8,
      "c"/utf8,
      ...
```
would require changes in the nesting logic or a special case I am not sure, I could look into it if you want

giacomocavalieri: Great job @DeviousStoat!! I think I can help with the formatting here. The code could look something like this (I haven't tested it but this should at least get you closer to a working solution 😁)
```rs
fn const_string_concatenate_bit_array<'a>(
    elems: impl IntoIterator<Item = Document<'a>>,
) -> Document<'a> {
  let elems = elems.collect_vec();
  if elems.is_empty {
    return "<<>>".to_doc();
  }
  
  let open = "<<".to_doc().append(break_("", "").nest(INDENT));
  let close = break_("", "").append(">>".to_doc());
  
  join(elems, break_(", ", ",").nest(INDENT))
    .surround(open, close)
    .group()
}
```

The look will be something like this if the bit array is split:
```erl
<<
  "wibble"/utf8,
  "wobble"/utf8
>>
```

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__assert_const_concat_many_strings_in_list.snap
```diff
@@ -0,0 +1,37 @@
...
+            "q"/utf8,
+            "r"/utf8,
+            "s"/utf8,
+            "t"/utf8,
+            "u"/utf8,
+            "v"/utf8,
+            "w"/utf8,
+            "x"/utf8,
+            "y"/utf8,
+            "z"/utf8>>].
```
giacomocavalieri, line 37: I think the preferred look would be:
```erl
<<
    "a"/utf8,
    "b"/utf8,
    ...
>>
```
You can have a look at [this comment](https://github.com/gleam-lang/gleam/pull/3318#discussion_r1664311024) to get an idea how to do it

lpil (approved):
Bravo!! Thank you

PR #3322, opened 2024-06-23
LSP: Show fn/value signature instead of a module name

Zhomart:
RFC: What do you think about this proposal?

**Motivation:** I want to see the function or value signatures when autocompleting, it allows quickly understand what function does and what does it return.  The module name can be deduced by the prefix before ".".

| Before | After |
|--------|--------|
| image | image | 

| Before | After |
|--------|--------|
| image |  image | 

| Before | After |
|--------|--------|
| image | image|

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -246,13 +246,7 @@ where
             if already_imported_values.contains(name) {
                 continue;
             }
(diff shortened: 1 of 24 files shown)


lpil:
How does the programmer know which module each item comes from after this case? The type information was also already included in the completion information, typically shown for the selected option.

PR #3325, opened 2024-06-23
LSP: Add extra new line after 1st import autocomplete

Zhomart:
Not sure what happened to this PR: https://github.com/gleam-lang/gleam/pull/3277

But here's a fix for https://github.com/gleam-lang/gleam/issues/3271

Changes:

1. Use the 1st definition instead of the 1st import
2. Add extra newline after the import if the 1st definition is not import

Demo (update Jul 13):

2024-07-13 at 18 30 15 - Azure Cuckoo

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -425,8 +425,9 @@ where
         }
 
         // Importable modules
-        let (import_location, has_imports) = self.first_import_line_in_module();
-        let after_import_newlines = self.add_newlines_after_import(import_location, has_imports);
+        let (first_import_pos, first_is_import) = self.first_import_in_module();
+        let after_import_newlines =
+            self.add_newlines_after_import(first_import_pos, first_is_import);
         for (module_full_name, module) in self.completable_modules_for_import() {
             // Do not try to import the prelude.
             if module_full_name == "gleam" {
@@ -461,7 +462,7 @@ where
                 );
                 add_import_to_completion(
                     &mut completion,
-                    import_location,
+                    first_import_pos,
                     module_full_name,
                     &after_import_newlines,
                 );
@@ -547,8 +548,9 @@ where
         }
 
         // Importable modules
-        let (import_location, has_imports) = self.first_import_line_in_module();
-        let after_import_newlines = self.add_newlines_after_import(import_location, has_imports);
+        let (first_import_pos, first_is_import) = self.first_import_in_module();
+        let after_import_newlines =
+            self.add_newlines_after_import(first_import_pos, first_is_import);
         for (module_full_name, module) in self.completable_modules_for_import() {
             // Do not try to import the prelude.
             if module_full_name == "gleam" {
@@ -578,7 +580,7 @@ where
 
                 add_import_to_completion(
                     &mut completion,
-                    import_location,
+                    first_import_pos,
                     module_full_name,
                     &after_import_newlines,
                 );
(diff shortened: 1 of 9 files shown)


compiler-core/src/language_server/completer.rs
```diff
@@ -586,30 +592,46 @@ where
...
-        Position::new(import_location - 1, 0)
+        (
+            Position::new(import_location - 1, 0),
+            first_import.is_some(),
+        )
+    }
+
+    // Returns true if an extra newline should be added after the import statement.
+    // * ``import_location`` - The position of the first import statement in the source code.
+    fn add_newline_after_import(&'a self, import_location: Position, has_imports: bool) -> bool {
```
lpil, line 607: Create a new enum rather than using a bool please, as you can't tell what `true` and `false` mean without reading the whole function for what it does with them.

compiler-core/src/language_server/completer.rs
```diff
@@ -447,7 +447,7 @@ where
                     insert_range,
                     TypeCompletionForm::Default,
                 );
-                add_import_to_completion(&mut completion, import_location, module_full_name);
+                add_import_to_completion(&mut completion, import_location, module_full_name, false);
```
lpil, line 450: How come this is always false?

lpil (commented):
Thank you. Could you update the changelog file and add some tests please?

compiler-core/src/language_server/completer.rs
Zhomart: Good catch, I assumed it's import. Fixed it.

compiler-core/src/language_server/completer.rs
Zhomart: I'm kind of new to rust, not sure what's the best way... But added an enum, does this work?

Zhomart:
> Thank you. Could you update the changelog file and add some tests please?

The existing tests cover these cases.

lpil (commented):
This will not be merged until all the PR feedback has been resolved. Dedicated tests that demonstrate specifically the behaviour is not optional, with comments if they do not express the intent clearly enough by themselves.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__importable_adds_extra_new_line_if_no_imports.snap
```diff
@@ -0,0 +1,69 @@
...
+                        start: Position {
+                            line: 0,
+                            character: 0,
+                        },
+                        end: Position {
+                            line: 0,
+                            character: 0,
+                        },
+                    },
+                    new_text: "import dep\n\n",
```
Zhomart, line 60: Double new line here, but single in other testcases

Zhomart:
> This will not be merged until all the PR feedback has been resolved. Dedicated tests that demonstrate specifically the behaviour is not optional, with comments if they do not express the intent clearly enough by themselves.

Done! Added tests.

compiler-core/src/language_server/completer.rs
```diff
@@ -588,30 +605,57 @@ where
...
-    fn first_import_line_in_module(&'a self) -> Position {
-        let import_location = self
-            .module
-            .ast
-            .definitions
-            .iter()
-            .find_map(get_import)
-            .map_or(0, |i| i.location.start);
+    fn first_import_line_in_module(&'a self) -> (Position, bool) {
+        let first_import = self.module.ast.definitions.iter().find_map(get_import);
```
lpil, line 609: Sorry- we always want imports to be added at the top, so could you make this check the first definition rather than searching for the first import? Thank you

Zhomart: Hey, no worries. I used the first defnition. Please take a look. 

It's weird that the first definition is always "import" statement. So it didn't change the behavior 🤷 .

Zhomart: Ok, apparently we can just use the `location.start` to find the 1st definition :)

lpil (approved):
Beautiful, thank you. I'll update the changelog

lpil:
I love your profile picture @Zhomart ! 😁

PR #3326, opened 2024-06-23
Annotate javascript constants as pure

giacomocavalieri:


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__consts__literal_list_does_not_get_constant_annotation.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: compiler-core/src/javascript/tests/consts.rs
+expression: "pub const a = [1, 2, 3]"
+---
+import { toList } from "../gleam.mjs";
+
+export const a = toList([1, 2, 3]);
```
lpil, line 7: This should be marked as pure I think

giacomocavalieri: Oh right, I missed this. It should be fixed now :)

lpil (approved):
Thank you

PR #3331, opened 2024-06-26
Fix invalid erlang when shadowing the bit pattern's size variable

### LINKED ISSUE #3315: Invalid Erlang generated with shadowed variables in bit patterns
```gleam
pub fn go(code) {
  let pre = 1
  case code {
    <<pre:bytes-size(pre), _:bytes>> -> pre
    _ -> panic
  }
}
```

Generates this invalid Erlang:

```erlang
go(Code) ->
    Pre = 1,
    case Code of
        <<Pre@1:Pre@1/binary, _/binary>> ->
            Pre@1;

        _ ->
            erlang:error(#{gleam_error => panic,
                    message => <<"panic expression evaluated"/utf8>>,

                    function => <<"go"/utf8>>,
                    line => 5})
    end.
```

Note how `Pre@1` is being used for the size, not `Pre`.

Eingin:
This closes #3315

When using the `size(x)` option on a bit array pattern. If the size option is passed a variable, and that variable is shadowed by the pattern segment the compiler will use the new shadowed name. This happens since the the usage of the already defined variable comes after the definition of the pattern's variable. It will therefore cause the compiler to use the new shadowed name instead of the intended previous name.

```gleam
let pre = 1
..
<<pre:bytes-size(pre), _:bytes>>
```
This currently creates
```erlang
<<Pre@1:Pre@1/binary, _/binary>>
```

To fix this I have moved the creation of the erlang output for the size portion before the value part. This will ensure that the previous variable name is used.

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -669,21 +669,21 @@ fn const_segment<'a>(
     options: &'a [TypedConstantBitArraySegmentOption],
     env: &mut Env<'a>,
 ) -> Document<'a> {
-    let mut value_is_a_string_literal = false;
-    let document = match value {
-        // Skip the normal <<value/utf8>> surrounds
-        Constant::String { value, .. } => {
-            value_is_a_string_literal = true;
-            value.to_doc().surround("\"", "\"")
-        }
+    let value_is_a_string_literal = matches!(value, Constant::String { .. });
 
-        // As normal
-        Constant::Int { .. } | Constant::Float { .. } | Constant::BitArray { .. } => {
-            const_inline(value, env)
-        }
+    let create_document = |env: &mut Env<'a>| {
+        match value {
+            // Skip the normal <<value/utf8>> surrounds
+            Constant::String { value, .. } => value.to_doc().surround("\"", "\""),
 
-        // Wrap anything else in parentheses
-        value => const_inline(value, env).surround("(", ")"),
+            // As normal
+            Constant::Int { .. } | Constant::Float { .. } | Constant::BitArray { .. } => {
+                const_inline(value, env)
+            }
+
+            // Wrap anything else in parentheses
+            value => const_inline(value, env).surround("(", ")"),
+        }
     };
 
     let size = |value: &'a TypedConstant, env: &mut Env<'a>| match value {
@@ -697,7 +697,7 @@ fn const_segment<'a>(
     let unit = |value: &'a u8| Some(eco_format!("unit:{value}").to_doc());
 
     bit_array_segment(
-        document,
+        create_document,
         options,
         size,
         unit,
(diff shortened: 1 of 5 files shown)


lpil (commented):
Thank you!

Rather than iterating over the options and cloning the size how about passing in the value as a function that takes no arguments? Then it'll be computed after the options without doing extra work, and it'll be a bit simpler

Eingin:
Yeah lets do that! I've updated the PR so now the value is passed as a function. I had to refactor the way `bit_array_segment` function works. Unfortunately now that there are multiple closures. One for the value and one for the size. The borrow checker doesn't like the multiple mutable refs with the printing. So it now takes in a printer function since the various uses of it have different requirements/dependencies for the output.

compiler-core/src/erlang.rs
```diff
@@ -610,67 +614,72 @@ fn expr_segment<'a>(
...
-fn bit_array_segment<'a, Value: 'a, SizeToDoc, UnitToDoc>(
-    mut document: Document<'a>,
+fn bit_array_segment<'a, Value: 'a, ValueType, ValueToDoc, SizeToDoc, UnitToDoc>(
+    mut value_to_doc: ValueToDoc,
     options: &'a [BitArrayOption<Value>],
     mut size_to_doc: SizeToDoc,
     mut unit_to_doc: UnitToDoc,
     value_is_a_string_literal: bool,
-    env: &mut Env<'a>,
+    print_value: &mut dyn FnMut(&'a ValueType) -> Document<'a>,
```
lpil, line 677: How come these are using dynamic dispatch now?

compiler-core/src/erlang.rs
```diff
@@ -610,67 +614,72 @@ fn expr_segment<'a>(
...
 ) -> Document<'a>
 where
-    SizeToDoc: FnMut(&'a Value, &mut Env<'a>) -> Option<Document<'a>>,
+    ValueToDoc: FnMut(&mut dyn FnMut(&'a ValueType) -> Document<'a>) -> Document<'a>,
+    SizeToDoc:
+        FnMut(&'a Value, &mut dyn FnMut(&'a ValueType) -> Document<'a>) -> Option<Document<'a>>,
```
lpil, line 682: Please revert these to the previous signature and add an addition function type here rather than using dynamic dispatch for `print_value`.

Eingin: @lpil 
Unfortunately we cant remove the dynamic dispatch. Since the `ValueToDoc` & `SizeToDoc` are closures they don't let us use generics for passing the `print_value` into them. If we try what your proposing, the borrow checker gives us [E0782](https://doc.rust-lang.org/error_codes/E0782.html) at the `ValueToDoc` & `SizeToDoc` definitions. 

Im not sure how best to solve this, I'll try think about it today and come up with something. Buy any recommendations would be helpful.

lpil: `print_value` should not be passed to them, it should continue to work as before.

lpil:

Eingin:
@lpil Sorry for the slow response.
I have pushed a new version of the changes but looks like I somehow accidentally closed this PR. Can you Re-open it up?
 
I've changed it to wrap the the creation of the document in a closure. The only thing I'm not totally sure on are the `RefCell`s i had to add for the parameters used in the `pattern_segment` part. We cant follow the current pattern of passing in the variables into the `bit_array_segment` since there are exclusively used only by this function. If you have a better idea on how to let the borrow checker be happy do let me know.

GearsDatapacks:
I'm pretty sure you can re-open it yourself

Eingin:
> I'm pretty sure you can re-open it yourself

Nope, the re-open button is disabled for me unfortunately.

lpil (approved):
Beautiful, thank you very much

PR #3332, opened 2024-06-26
Show ignored fields on `..` hover

giacomocavalieri:


compiler-core/src/build.rs
```diff
@@ -301,6 +301,10 @@ pub struct UnqualifiedImport<'a> {
 #[derive(Debug, Clone, PartialEq)]
 pub enum Located<'a> {
     Pattern(&'a TypedPattern),
+    PatternSpread {
+        spread_location: SrcSpan,
+        unused_fields: Vec<(Option<EcoString>, std::sync::Arc<type_::Type>)>,
```
lpil, line 306: Prefer references to taking ownership here, to avoid cloning. All the arguments without modification sounds best to me.

How to use the information can then be up to the engine

compiler-core/src/type_/pattern.rs
```diff
@@ -472,8 +509,8 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                                         type_: (),
                                     },
                                     location: spread_location,
-                                    label: None,
-                                    implicit: false,
+                                    label: missing_labels.next(),
```
lpil, line 512: Clever!

lpil (commented):
Very nice! Left a small note

compiler-core/src/build.rs
giacomocavalieri: Ok! I should have fixed it :)

lpil (approved):
Beautiful!!!

PR #3338, opened 2024-06-29
Better error message for call expression in case clause guard

### LINKED ISSUE #2509: Error message for using a function in `case` guard clause is unhelpful

DeviousStoat:
closes  #2509

Gave it a try since the other PR was closed and I think this would be very helpful

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1432,9 +1432,8 @@ where
         }
     }
 
-    // checks if we have a function call after we find an invalid left paren
-    // Warning: this function will consume the tokens
-    // Returns the location of the right paren if we have a function call
+    /// Checks if we have an unexpected left parenthesis and returns appropriate
+    /// error if it is a function call.
     fn parse_function_call_in_clause_guard(&mut self, start: u32) -> Result<(), ParseError> {
         if let Some((l_paren_start, l_paren_end)) = self.maybe_one(&Token::LeftParen) {
             if let Ok((_, end)) = self
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1319,6 +1319,13 @@ where
         }
     }
 
+    fn add_multi_line_clause_hint(&self, mut err: ParseError) -> ParseError {
+        if let ParseErrorType::UnexpectedToken { ref mut hint, .. } = err.error {
+            *hint = Some("Did you mean to wrap a multi line clause in curly braces?".into());
+        }
+        err
+    }
+
     // examples:
     //   pattern -> expr
     //   pattern, pattern if -> expr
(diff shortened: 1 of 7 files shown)


compiler-core/src/parse.rs
```diff
@@ -1395,6 +1395,21 @@ where
         }
     }
 
+    // checks if we have a function call after we find an invalid left paren
+    // Warning: this function will consume the tokens
+    // Returns the location of the right paren if we have a function call
+    fn is_invalid_call_in_case_clause_guard(&mut self) -> Option<u32> {
+        if self.parse_fn_args().is_err() {
```
lpil, line 1402: As the comment says this could consume some characters but then fail, at which point the parser is left in an "ok" state at the level up. This could result in some invalid code being "successfully" parsed.

The parser must always remain in an error state once it encounters an error.

How about making this function return a parse error instead? It would ensure it is always correct, and it removes some of the duplicated logic from the sites where this function is called

compiler-core/src/parse.rs
```diff
@@ -1395,6 +1395,21 @@ where
         }
     }
 
+    // checks if we have a function call after we find an invalid left paren
+    // Warning: this function will consume the tokens
+    // Returns the location of the right paren if we have a function call
+    fn is_invalid_call_in_case_clause_guard(&mut self) -> Option<u32> {
```
lpil, line 1401: This name doesn't match how the other methods are named in the parser

compiler-core/src/parse.rs
```diff
@@ -1395,6 +1395,21 @@ where
         }
     }
 
+    // checks if we have a function call after we find an invalid left paren
+    // Warning: this function will consume the tokens
+    // Returns the location of the right paren if we have a function call
```
lpil, line 1400: Doc comments use `///`. Capital letters and punctuation too please 🙏

lpil (commented):
Thank you!

lpil (approved):
Lovely work!! Thank you very much

lpil:
Would you mind rebasing to be up-to-date with main and also updating the changelog? Thank you

DeviousStoat:
I rebased on main and updated the changelog

lpil (approved):
Beautiful!!! Thank you!!

Issue #3324, opened 2024-06-23
'Expected one of : "}" a record constructor' -- Error message improvement for custom types


sajorobinson:
I was having trouble creating a custom type and was having trouble understanding the error message.

```gleam
Expected one of:
"}"
a record constructor
```

I later learned I was trying to make the type incorrectly. I needed to make a constructor for it. Knowing that now, the error message makes a lot more sense, but I didn't understand it then.

The error message could possibly be improved to be clearer. Possibly:

- Better formatting?
- More explicitly stating what the problem is?

A screenshot is below.

![Screenshot_2024-06-22_at_18 54 18](https://github.com/gleam-lang/gleam/assets/26798946/870c1b2c-b2a2-426a-bb1c-aa1e6c7a7c65)

Thanks!

lpil:
Thank you!

rdghosal:
I'm taking a look! 👀

lpil:
Thank you! What do you think the error ought to be?

rdghosal:
@lpil I think that "a record constructor" and the message contents are pretty descriptive, but I think a change in formatting could improve readability.

I'm currently working on changing this error message to look something like the following... What do you think?
(Note, the expected tokens will indeed be inlined as below and the message will read "Expected" when there's only one expected token.)

```
Expected one of "}" or a record constructor
```

sajorobinson:
Something I noticed is the way the error is presented, it seems like it's supposed to be an unordered list of possible things it's expecting.

I think I would have interpreted its meaning more quickly if it looked like:

```
Expected one of:

- "}"
- A record constructor
```

or, if markdown unordered lists aren't desirable, something using indents like:

```
Expected one of:

    "}"
    A record constructor
```

rdghosal:
Agreed; if we go with a list, there needs to be something that denotes list items (the capital "A" in "A record" is a nice touch too).

Also, I noticed that my proposed solution results in horizontal growth of the error message, though, in practice, this shouldn't be an issue. I suppose it's a matter of preference...

lpil:
Seeing as this is a common mistake perhaps we could show what the correct code should be. We could use the name of the type as the name of the constructor we display

lpil:
```gleam
pub type User {
  name: String
}
```
```
error: Syntax error
   ┌─ /Users/louis/src/lpil/puck/src/puck.gleam:23:3
   │
23 │   name: String
   │   ^^^^ I was not expecting this

[...]

PR #3339, opened 2024-06-29
Improve Error Messaging for Incorrect Custom Type Defs

rdghosal:
Addresses #3324.

This commit changes:
- Error messaging on an incorrect record constructor definition to one that explains the expected syntax for a record constructor and presents and example thereof.


compiler-core/src/ast.rs
```diff
@@ -413,6 +413,47 @@ impl TypeAst {
             TypeAst::Var(_) | TypeAst::Hole(_) => Some(Located::Annotation(self.location(), type_)),
         }
     }
+
+    /// Consumes a `TypeAst` to return the corresponding annotation.
+    pub fn to_annotation(self) -> EcoString {
```
lpil, line 418: This doesn't need to take ownership, let's have it take a reference.

Could we call it `print` too please, to match more what we have in other parts of the codebase.

Ideally this would call an internal helper function and append to a mutable EcoString rather than making lots of Strings

compiler-core/src/ast.rs
```diff
@@ -413,6 +413,47 @@ impl TypeAst {
...
+                let return_ = func.return_.to_annotation();
+                let args = func
+                    .arguments
+                    .into_iter()
+                    .map(|a| a.to_annotation())
+                    .collect::<Vec<_>>()
+                    .join(", ");
+                format!("fn({args}) -> {return_}").into()
+            }
+            TypeAst::Constructor(constructor) => {
```
lpil, line 439: Include the module qualifier if it is present in the AST please 🙏

compiler-core/src/parse/error.rs
```diff
@@ -232,6 +232,41 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
...
             ),
+            ParseErrorType::ExpectedRecordConstructor { type_name, field, type_ } => {
+                let (constructor, label, annotation) = match (field, type_) {
+                    (Some(f), Some(t)) => (
+                        type_name.to_owned(),
+                        format!("{f}: "),
+                        t.clone().to_annotation(),
+                    ),
+                    (Some(f), None) => (
+                        f.to_pascal_case().as_str().into(),
```
lpil, line 243: How come this doesn't use `type_name` when the above does?

compiler-core/src/parse/tests.rs
```diff
@@ -1014,6 +1014,21 @@ type A {
...
+#[test]
+fn type_invalid_variant_constructor() {
+    assert_module_error!(
+        "
+pub type User {
+    name: String,
+}
+"
+    );
+}
```
lpil, line 1030: Could we have tests for each of the syntaxes that this feature supports please 🙏  I can see there's at least 4 variants

lpil (commented):
Very nice! Would you mind updating the changelog also?

rdghosal:
@lpil Thanks, as always, for the thorough review. I've pushed updates per your feedback, including more tests, an 
updated changelog, and an approach that tightens the scope to cover the use case in the issue.

rdghosal:
@lpil A friendly ping in case you had any additional feedback on this PR!

lpil (approved):
Thank you!!!

PR #3358, opened 2024-07-03
Turn LS tests into snapshot tests

giacomocavalieri:
This PR replaces most of the old LS tests (especially the hover ones) with snapshot tests.

A huge time saver is that in a hover test you'll no longer have to guess if the highlighted span is correct, since it will be shown directly in the source code; previously one would have to manually write down the expected highlighted span.

Another nice thing is now there is a handy dsl to write hover tests and avoid hard coding positions, for example:
```rs
let code = "fn wibble() { todo }"
assert_hover!(..., Position::new(0, 4))
assert_hover!(..., find_position_of("wibble").under_char('i'))
```

Screenshot 2024-07-03 alle 13 31 18

### DIFF
--- a/compiler-core/src/language_server/tests/hover.rs
+++ b/compiler-core/src/language_server/tests/hover.rs
@@ -136,6 +136,11 @@ fn show_hover(code: &str, range: Range, pointer: Position) -> String {
 
 #[macro_export]
 macro_rules! assert_hover {
+    ($code:literal, $pointer:expr $(,)?) => {
+        let project = TestProject::for_source($code);
+        assert_hover!(project, $pointer);
+    };
+
     ($project:expr, $pointer:expr $(,)?) => {
         let src = $project.src;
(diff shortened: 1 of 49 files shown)


compiler-core/src/language_server/tests/hover.rs
```diff
@@ -51,10 +136,11 @@ fn show_hover(code: &str, range: Range, pointer: Position) -> String {
 
 #[macro_export]
 macro_rules! assert_hover {
-    ($project:expr, $position:expr $(,)?) => {
+    ($project:expr, $pointer:expr $(,)?) => {
```
lpil, line 144: Remove references to "pointer" please 🙏

giacomocavalieri: Oops missed that!

lpil (approved):
Thank you!!!

PR #3359, opened 2024-07-03
De-foo-ify code

giacomocavalieri:
Since we wanted to avoid using "foo-bar-baz" I've gone and replaced all those with the much cuter "wibble-wobble-wubble". And hopefully new contributors will follow the example not seeing any foo-bar in the code :)


lpil (approved):
Wow! Thank you!!

PR #3363, opened 2024-07-05
Field access completions

### LINKED ISSUE #3298: LSP: Suggestion on record accessors
Given the following record:
```gleam
pub type Model {
  Model(count: Int, cats: List(String))
}
```
I'd like to be able to auto-complete the accessors:
```gleam
let model = Model(0, [])
model. // <- should suggest `count` and `cats`
```

Acepie:
Closes #3298

Change ultimately comes down to 4 relatively distinct parts

- fieldaccess with a missing label now parses correctly with an empty label. instead, they fail at analysis.
- fault tolerance has been added to record accesses so they don't turn the entire expression invalid. In addition, if a record access is being attempted and the record itself is valid, rather than just placing an invalid expr into the ast, a dummy record access expr will be inserted so we can detect the attempt
- find_node now takes in an arg to make the ast search end inclusive. This is important for giving good suggestions on the last character of an expression.
  - I made this a bool arg to find_node instead of changing the behavior since I think exclusive is reasonable/correct for non-completion cases but I could see the argument either way so lmk what you think on this
  - This is the majority of the actual LoC change since its just passing the bool through all the recursive calls
- plumbing in the lsp so that if the found expression is a record access to add accessors for a type to the completions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,9 +9,17 @@
 - The warning for the deprecated `[..]` pattern has been improved.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Record accessors are now fault tolerant. This means an invalid label can be
+  properly detected and won't invalidate the rest of the expression.
+  ([Ameen Radwan](https://github.com/Acepie))
+
 ### Formatter
 
 ### Language Server
+
+- The language server can now show completions for fields if a record access is
+  being attempted.
+  ([Ameen Radwan](https://github.com/Acepie))
 
 ### Bug Fixes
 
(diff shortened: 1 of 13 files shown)


lpil:
Is there a reason not to always make it inclusive?

Acepie:
I think switching everything to be inclusive would fine? but I would guess there could be some edge cases that I'm not thinking of off the top of my head. I'll switch it to always be inclusive for now

lpil:
Let's give it a try. We can always add more complexity later if we find a place it doesn't work the way we want.

Acepie:
Ok that cleans up the pr a bit :D

lpil (approved):
omg this feels amazing to use 😍

Issue #3323, opened 2024-06-23
LSP: Show auto-complete types instead of the module names


Zhomart:
Gleam: 
<img width="572" alt="image" src="https://github.com/gleam-lang/gleam/assets/830536/40cac882-f8d8-4373-96f1-662e67c11833">

How other languages do:

Rust: 
![image](https://github.com/gleam-lang/gleam/assets/830536/e955d6bb-09a4-4211-98f3-bca3c9f75002)

Typescript (only on select):
<img width="733" alt="image" src="https://github.com/gleam-lang/gleam/assets/830536/b5af248e-0474-404b-997d-bae4bc575539">

Golang (only on select):
<img width="579" alt="image" src="https://github.com/gleam-lang/gleam/assets/830536/be483fa3-d948-413c-894b-e99ada3afa60">

I've created a PR to showcase this (https://github.com/gleam-lang/gleam/pull/3322), what do you think?

lpil:
The examples you've given here don't support removing the module in my eyes.

Rust: it prominently shows the module first of all. When there is not enough space for the module and the type it shows only the module.

TypeScript: The module is used in the code to filter the suggestions so they all belong to the same module. In this case it is safe to only show the type.

Go: This is completion on an object rather than generally so no modules are included.

It is vital that there is a way for the programmer to be able to tell what module each item comes from. Today we show both the module and the type, with the type being in the further details section, which you have collapsed in VSCode so it does not show. We could adopt the Rust approach where it shows the type information inline in the instance there is enough space after the module, or we could take some other approach so long as the module is always clear.

Zhomart:
I believe gleam should show type information. This helps engineers during coding.

It seems like we can use field to show type informaiton.  Let me try create PR and see if it works.

<img width="646" alt="image" src="https://github.com/gleam-lang/gleam/assets/830536/3b1fc8d7-c5ca-4d0e-a5f3-df98070b0973">

Zhomart:
Hi @lpil, please take a look at this (https://github.com/gleam-lang/gleam/pull/3364) example. VSCode shows it nicely. What do you think?

lpil:
Hello! Please reply to my comments above 🙏

Zhomart:
> The examples you've given here don't support removing the module in my eyes.
> 
> Rust: it prominently shows the module first of all. When there is not enough space for the module and the type it shows only the module.
> 
[...]

PR #3364, opened 2024-07-06
LSP: Set type to label_details.detail

Zhomart:
LSP Docs: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItemLabelDetails

As discussed in https://github.com/gleam-lang/gleam/issues/3323, this is a PR to show types right after the completion value name.

## Examples:

Import:

image

Import 2:

image

Value:

image

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -681,9 +681,9 @@ fn value_completion(
     CompletionItem {
         label: label.clone(),
         kind,
-        detail: Some(type_),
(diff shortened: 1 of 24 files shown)


lpil:
Hello! Can you show what it looks like when there is not enough space to show both the type and the module in full, and what it looks like in other editors. Thank you

lpil:
This PR was opened prior to the approach being decided in the issue. Please see the issue for discussion and unresolved questions. Once resolved this PR can be worked on

PR #3372, opened 2024-07-08
Fix error spans for duplicate `@external` attributes

### LINKED ISSUE #3370: Error span differs between erlang and javascript for duplicate @external attributes
There is a small inconsistency when reporting errors for duplicate `@external` attributes between erlang and javascript.

This code...
```gleam
@external(erlang, "wibble", "wobble")
@external(erlang, "wibble", "wobble")
fn wibble() -> Int
```

...results in this error
```
error: Syntax error
   ┌─ /src/main.gleam:13:1
   │
13 │ @external(erlang, "wibble", "wobble")
   │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Duplicate attribute

This attribute has already been given.
```

Whereas this code...
```gleam
@external(javascript, "wibble", "wobble")
@external(javascript, "wibble", "wobble")
fn wibble() -> Int
```

...results in this error
```
error: Syntax error
  ┌─ /src/main.gleam:8:1
  │
8 │ @external(javascript, "wibble", "wobble")
  │ ^^^^^^^^^ Duplicate attribute

This attribute has already been given.
```

Note that in erlang, the span is the entire attribute whereas in javascript the span is only the `@external` keyword.

cszczepaniak:
Closes #3370 

This PR addresses a small inconsistency in parse error spans for duplicate `@external` attributes between erlang and javascript.


compiler-core/src/parse.rs
```diff
@@ -3139,37 +3153,25 @@ where
...
+            _ => return parse_error(ParseErrorType::UnknownAttribute, SrcSpan::new(start, end)),
+        };
 
-            "javascript" => {
-                let _ = self.expect_one(&Token::Comma)?;
-                let (_, module, _) = self.expect_string()?;
-                let _ = self.expect_one(&Token::Comma)?;
-                let (_, function, _) = self.expect_string()?;
-                let _ = self.maybe_one(&Token::Comma);
-                let _ = self.expect_one(&Token::RightParen)?;
```
cszczepaniak, line 3163: Here's the subtle bug: we weren't pulling off the `end` for the RightParen token in this branch, but we were in the erlang branch. I decided to commonize as much of these two branches as possible to avoid such inconsistencies.

lpil (commented):
Thank you! Could you update the changelog also please 🙏

cszczepaniak:
@lpil Done!

lpil (approved):
Lovely work, thank you

PR #3381, opened 2024-07-10
Do not show completions when typing strings

giacomocavalieri:
The LS currently shows completions even when writing literal strings, I find that to be a bit distracting and not particularly useful. This PR removes completions when writing inside a String

Export-1720618609383


lpil (approved):
Lovely, thank you

PR #3384, opened 2024-07-10
Do not report `[..]` as being deprecated

### LINKED ISSUE #3383: The [..] pattern shouldn't show a deprecation warning
With Gleam v1.3.0+

```gleam
case {
  [_, _, ..] -> "this is fine"
  [..] -> panic as "deprecated"
}
```

The `[..]` pattern is currently triggering the following warning:

> Deprecated list pattern matching syntax
> 
> This spread should be preceded by a comma.
> 
> This syntax for pattern matching on a list is deprecated. When matching on the rest of a list it should always be preceded by a comma, like this: `[item, ..list]`.

The `[..]` was fine with Gleam v1.2.1

giacomocavalieri:
This PR closes #3383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Compiler
 
+- The `[..]` pattern has been deprecated in favour of the `_` catch all pattern.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Formatter
 
 ### Language Server
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1204,12 +1204,15 @@ where
 
                 let mut elements_after_tail = None;
                 let mut dot_dot_location = None;
-                let tail = if let Some((start, Token::DotDot, end)) = self.tok0 {
-                    dot_dot_location = Some((start, end));
-                    if !elements_end_with_comma {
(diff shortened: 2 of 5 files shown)


giacomocavalieri:
Really? I thought the only deprecated one was `[item..rest]`, not `[..]`. In that case the current message is wrong anyway because it asks to put a comma before the `..`. What would be the suggested replacement in the warning in this case?

lpil:
`[..]` is deprecated as it's confusing (is it empty or at least one or any?) and it's a second way of writing the same thing, which we don't do in Gleam.

giacomocavalieri:
I still think the current error message is rather confusing as it suggests putting a comma before `..`, resulting in invalid syntax. Any ideas on how it could be improved?

giacomocavalieri:
Ok! I can change that

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Compiler
 
+- The `[..]` pattern has been deprecated in favour of the `_` catch all pattern.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
```
lpil, line 10: It was deprecated previously, this is improving the error message.

lpil (approved):
Thank you!

Issue #1953, opened 2023-01-12
Flag for gleam new that can be used to create a project targeting Javascript


CrowdHailer:
As it stands `gleam new` creates a project targeting erlang. including gh actions etc.
It would be good if there was a way to create projects targeting JS.

Note I see there is a template option. Maybe using that is the right thing to do, but if so we probably don't want `lib` to mean erlang. And, if using the template option it would still be good to have the option for creating a js project documented.
Currently the info is:
```
--template <TEMPLATE>          [default: lib] [possible values: lib]
```

lpil:
Sounds good. We could have `erlang` and `javascript` as the templates.

aadarsh10c:
Hey , I would love to work on this issue . Let me know how can I help, looking forward for your response.

lpil:
Brilliant, thank you @aadarsh10c ! Let me know if I can give you any pointers at any time.

Tar-Tarus:
Hi there 👋 
This issue seems to be a bit stale but if it's still up for grabs, I'd like to work on it.

Also I have some initial questions regarding the required changes for this:

> Sounds good. We could have `erlang` and `javascript` as the templates.
- From the following response, is it fair to assume that the `lib` option should be replaced with `erlang` (and remain the default)?

What are the essential changes to the template when the user chooses `javascript`? 
- My initial idea would be to add the `target` flag to `gleam.toml` to specify the JS target 
- Also I was wondering whether adding the [`gleam_javascript`](https://github.com/gleam-lang/javascript) dependency in the `gleam.toml` template makes sense. From what I can see though, this library doesn't seem like a must have for simple JS-target projects. 
- From what I gather from the [`gleam.toml`](https://gleam.run/writing-gleam/gleam-toml/) documentation, there's also a variety of configuration options for running on a JS runtime. Would it make sense to add those as well? Or possibly add a comment to the `gleam.toml` file to hint the user at those parameters?

- Any pointers as to what would also be essential to adapt/adjust depending on the `template` flag?

lpil:
Hello there!

> From the following response, is it fair to assume that the lib option should be replaced with erlang (and remain the default)?

Sounds good!

> My initial idea would be to add the target flag to gleam.toml to specify the JS target

Perfect. Let's do that.

> Any pointers as to what would also be essential to adapt/adjust depending on the template flag?

I've not looked into this feature at all I'm afraid.

Tar-Tarus:
[...]

PR #3395, opened 2024-07-12
CLI: Add javascript as option for template flag to `new`

Tar-Tarus:
This PR resolves: #1953


compiler-cli/src/new.rs
```diff
@@ -24,9 +24,14 @@ const ELIXIR_VERSION: &str = "1.15.4";
 #[derive(
     Debug, Serialize, Deserialize, Display, EnumString, VariantNames, ValueEnum, Clone, Copy,
 )]
-#[strum(serialize_all = "kebab_case")]
+#[strum(serialize_all = "lowercase")]
 pub enum Template {
-    Lib,
+    #[strum(serialize = "erlang", serialize = "erl")]
```
lpil, line 29: Could you put back `Lib` please? To remove it would be a breaking change. Don't include it in the help text though

compiler-cli/src/new.rs
```diff
@@ -142,6 +151,7 @@ pub fn hello_world_test() {
             Self::GleamToml => Some(format!(
                 r#"name = "{project_name}"
 version = "1.0.0"
+{target}
```
lpil, line 154: This outputs an extra empty line for Erlang. Let's keep the Erlang format as it was before.

compiler-cli/src/new/tests.rs
```diff
@@ -33,6 +33,54 @@ fn new() {
...
+        },
+        "1.0.0-gleam",
+    )
+    .unwrap();
+    creator.run().unwrap();
+
+    assert!(path.join("gleam.toml").exists());
+
+    let toml = crate::fs::read(path.join("gleam.toml")).unwrap();
+    assert!(!toml.contains("target = \"javascript\""));
```
lpil, line 57: Use snapshots for the tests please

lpil (commented):
Great stuff! Thank you! I've left some comments inline

compiler-cli/src/new.rs
Tar-Tarus: I've corrected the template in aa39651a7ab41df0a1acb7754c4ab00b06df393c and 19a4695e75f89e215c2fb3f332705eb8747dcbd4 and the template now looks as expected.

compiler-cli/src/new/tests.rs
Tar-Tarus: Thanks for the hint, learning something new 👌 
I've set up snapshot assertions for those two tests in 4d18cd55f243c7517eab17e6482be5ae60ee76a2 .
Let me know if they need more work or if something is missing.

compiler-cli/src/new.rs
lpil: Thank you. Remove the deprecation notice for this variant please.

The code for how these small string is super complicated now and very hard to understand. Can we simplify this please, perhaps just have a regular array/vector of values, or a string.

lpil (commented):
Nice! Could you make the snapshots test the whole project rather than just the gleam.toml please? Thank you.

compiler-cli/src/new.rs
Tar-Tarus: Thanks for the feedback!
I've made the changes in the latest commit e8336a0c6382d2e87e0138d8a5d634a10603f6ca , only keeping the `clap` annotation for the help documentation.

Tar-Tarus:
> Nice! Could you make the snapshots test the whole project rather than just the gleam.toml please? Thank you.

I've had a look at this and here's my conclusion/question: I couldn't find a similar case of testing the entire project with `insta` instead of just a single file (most uses of `insta` are for single compilation artefacts or simple strings). The `insta` documentation does include the possibility of using [glob patterns to snapshot an entire directory](https://insta.rs/docs/advanced/) which I believe would be the most fitting in this case as future changes to the template could also be captured with this. I've implemented this in commit e46dfd06a9d1748d718dbd8501f0f1ceb25b2731
The use of this `glob!()` macro does require the feature to be selected with the package dependency. Therefore I've also made the according change in the root `Cargo.toml`. (1011b2f5fcfe0757f8a6827dc44dded85ca8e039)

Let me know if this is unclear or possibly misunderstood on my part and there's a better/smarter way to snapshot the whole project instead...

lpil (approved):
Great! Thank you. Could you update the changelog too please 🙏

compiler-cli/src/new/tests.rs
```diff
@@ -33,6 +34,62 @@ fn new() {
...
+            name: None,
+            skip_git: false,
+            skip_github: true,
+        },
+        "1.0.0-gleam",
+    )
+    .unwrap();
+    creator.run().unwrap();
+
+    glob!(path, "my_project/*.*", |file_path| {
```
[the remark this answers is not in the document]
Tar-Tarus: Corrected in 19745c4ddda599bfe38732b3035c2f944b72a98e

Tar-Tarus:
> Great! Thank you. Could you update the changelog too please 🙏

I've added an entry in commit 58c18f4740f1832775c93859526054dffe1c19fe

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
   platform for the generated documentation.
   ([Jiangda Wang](https://github.com/frank-iii))
 
+- `gleam new` now takes an optional `--template` flag (with options `erlang`
+  and `javascript`) to specify the template and target for the generated
+  project.
```
lpil, line 13: The flag already existed, it's that there are new values and "erlang" replaces "lib"

lpil (approved):
Thank you

PR #3400, opened 2024-07-13
Function call fault tolerance

Acepie:
This improves the fault tolerance by making it so that infer_call always returns a valid TypedExpr. If the function itself is invalid (i.e. not a function) then we just return an Invalid expression. Otherwise, we try to infer each argument and replace invalid arguments with invalid expressions.

For context, the primary motivation for this was to improve the completions when functions have errors. So now with this change if you are writing a function you will get completions even if the number of arguments is wrong or one of the arguments doesn't type check)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,10 @@
   expressions.
   ([Thomas](https://github.com/DeviousStoat))
 
+- Function calls are now fault tolerant. This means that errors in the function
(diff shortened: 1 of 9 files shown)


lpil (approved):
Oh this is super nice!! Thank you!!

lpil:
@Acepie I've discovered an issue with this relating to labels https://github.com/gleam-lang/gleam/issues/3405

PR #3402, opened 2024-07-14
dont allow private types to leak out of internal modules

### LINKED ISSUE #3387: Can use private type in signature of public function when in a very specific directory
Normally the following code would result in an error.

```gleam
type Foo {
  Foo
}

pub fn bar(arg: Foo) -> Foo {
  case arg {
    Foo -> Foo
  }
}
```

```error
Private type used in public interface

The following type is private, but is being used by this public export.

    Foo

Private types can only be used within the module that defines them.
```

However if you put that code inside of a module located in a very specific directory, then it doesn't cause an error. On top of that, if you're generating `.d.ts` files, then those files will refer to types that don't exists.

```typescript
export function bar(arg: Foo$): Foo$;
```

```error
Cannot find name 'Foo$'.ts(2304)
```

## Steps to reproduce

```sh
gleam new type_bug
cd type_bug
mkdir src/type_bug # this folder must be named the same as the project
mkdir src/type_bug/internal # this folder must be named internal
cat > src/type_bug/internal/bug.gleam << EOF
type Foo {
  Foo
}

pub fn bar(arg: Foo) -> Foo {
  case arg {
    Foo -> Foo
  }
}
EOF
```

## Versions

**Gleam:** gleam 1.3.1
**OS:** macOS 14.5

## Gleam Logs

https://gist.github.com/Olian04/fb5cf7ebd58d4c9a89d9823a1696ccf5

Acepie:
Closes #3387

Removed the special casing for internal modules. If the type of a nonprivate module value ever contains anything private then give the error

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1165,15 +1165,6 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             return;
         }
 
-        // We also don't want to raise a warning if we're inside an internal
-        // module ourselves, the type wouldn't actually be publicly exposed.
-        if self
-            .package_config
-            .is_internal_module(self.module_name.as_str())
-        {
-            return;
-        }
-
         // If a private or internal value references a private type
         if let Some(leaked) = value.type_.find_private_type() {
             self.errors.push(Error::PrivateTypeLeak {
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -100,6 +100,14 @@ macro_rules! assert_module_error {
 }
 
 #[macro_export]
+macro_rules! assert_internal_module_error {
+    ($src:expr) => {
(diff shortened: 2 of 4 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1165,15 +1165,6 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
         }
 
-        // We also don't want to raise a warning if we're inside an internal
-        // module ourselves, the type wouldn't actually be publicly exposed.
-        if self
-            .package_config
-            .is_internal_module(self.module_name.as_str())
-        {
-            return;
-        }
```
lpil, line 1175: Oh weird, why did I add this

lpil (approved):
Thank you!

lpil:
Would you mind updating the changelog please

lpil (approved):

PR #3404, opened 2024-07-14
Case fault tolerance

Acepie:
Makes case inference fault tolerant by individually checking each subject, pattern, guard, and then clause. This ultimately enables better LSP behavior even when a case is inexhaustive or 1 clause/subject has an issue

Screenshot 2024-07-14 103548
Screenshot 2024-07-14 103912
Screenshot 2024-07-14 104055


lpil (approved):
Bravo!! Fantastic work!

Issue #3405, opened 2024-07-15
Function fault tolerance results in confusing errors when using labels and arity is incorrect


lpil:
When a function has not enough supplied arguments and labels are used the remaining ones get positionally rewritten incorrectly, which results in incorrect "expected X, got Y" errors.

I think a solution here may be to not assert the type of the arguments when the arity is incorrect?

<img width="954" alt="image" src="https://github.com/user-attachments/assets/2bd4d611-f2ad-4742-9123-8c480e3daa5b">

lpil:
I wouldn't spend much time on it myself, it's ok to lose some errors in the presence of others.

PR #3406, opened 2024-07-15
Add signature help

### LINKED ISSUE #2242: LSP: Display function signature when passing arguments
In instances like such:
```gleam
fn main() {
  do_something(
}
```
it would be nice to have the function signature for `do_something` displayed as you pass in the arguments so as to know the order and type information.

giacomocavalieri:
This PR closes #2242 

I tried to make the snapshots as pretty as possible, but this is just a small example of how it looks if those were not clear enough:

https://github.com/user-attachments/assets/9be7d986-0420-407e-acb8-5dbbacd31427


compiler-core/src/language_server/signature_help.rs
```diff
@@ -0,0 +1,241 @@
...
+use super::{compiler::LspProjectCompiler, files::FileSystemProxy};
+
+pub fn for_expression<IO>(
+    compiler: &LspProjectCompiler<FileSystemProxy<IO>>,
+    expr: &TypedExpr,
+) -> Option<SignatureHelp>
+where
+    IO: CommandExecutor + FileSystemWriter + FileSystemReader + Clone,
+{
+    match expr {
```
lpil, line 27: Favour `let ... = ... else { return None }` or similar over nesting

compiler-core/src/language_server/signature_help.rs
```diff
@@ -0,0 +1,241 @@
...
+    let args_len = args.len() as u32;
+    if args_len == 0 {
+        return None;
+    }
+
+    let index_to_label = match field_map {
+        Some(field_map) => field_map
+            .fields
+            .iter()
+            .map(|(name, index)| (*index, name.clone()))
```
lpil, line 112: Can avoid the clone here seeing as it's never consumed

compiler-core/src/language_server/signature_help.rs
```diff
@@ -0,0 +1,241 @@
...
+            documentation: fun.get_documentation().map(|d| {
+                Documentation::MarkupContent(MarkupContent {
+                    kind: MarkupKind::Markdown,
+                    value: d.into(),
+                })
+            }),
+            parameters: Some(parameters),
+            active_parameter: None,
+        }],
+        active_signature: Some(0),
```
lpil, line 138: What does this mean?

compiler-core/src/language_server/tests/hover.rs
```diff
@@ -14,51 +14,51 @@ fn hover(tester: TestProject<'_>, position: Position) -> Option<Hover> {
     })
 }
 
-struct PositionFinder {
+pub struct PositionFinder {
```
lpil, line 17: If this is used by other tests let's move it out of the hover specific tests. Could be in `tests.rs`

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__signature_help__help_for_aliased_qualified_call.snap
```diff
@@ -0,0 +1,16 @@
...
+  wibble.example_fn()
+                    ↑
+}
+
+
+----- Signature help -----
+wibble.example_fn(Int, String) -> Nil
+                  ▔▔▔
+
+No documentation
```
lpil, line 16: Oh my gosh these are such beautiful snapshots

lpil (commented):
Oh my gosh Jak this is amazing!!! Fantastic work!

Left some super tiny comments inline

compiler-core/src/language_server/signature_help.rs
giacomocavalieri: This is the index of the signature to use from the list of signatures we include in the response. Since I always include just a single signature I'm telling the LS to use that

lpil (approved):
Thank you!!!

lpil:
ah! Looks like more is needed for pipes to be supported

compiler-core/src/type_/pipe.rs
```diff
@@ -98,14 +98,13 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
...
-                        }
-
+                    match fun.type_().fn_types() {
                         // Rewrite as right(..args)(left)
-                        _ => self.infer_apply_to_call_pipe(fun, arguments, location)?,
+                        Some((_args, return_)) if return_.fn_arity() == Some(1) => {
+                            self.infer_apply_to_call_pipe(fun, arguments, location)
+                        }
+                        // Rewrite as right(left, ..args)
+                        _ => self.infer_insert_pipe(fun, arguments, location),
```
giacomocavalieri, line 107: As we discussed earlier this is needed so that in case of a call with the wrong arity the pipe insertion `right(left, ..args)` is tried instead of `right(..args)(left)` to get better signature suggestions when someone is writing something like this: `[] |> list.map()`. This way, to the LS it looks like someone is calling `list.map([], )` and can highlight the second argument instead of being `list.map()([])` and having the first list argument highlighted

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__correct_pipe_arity_error_location.snap
```diff
@@ -6,4 +6,4 @@ error: Incorrect arity
   ┌─ /src/one/two.gleam:2:18
   │
 2 │ fn main() { 1 |> x() }
-  │                  ^^^ Expected 2 arguments, got 0
+  │                  ^^^ Expected 2 arguments, got 1
```
giacomocavalieri, line 9: Both would be correct but I believe that the new one is going to be a bit more useful, my hunch is that the case where people meant `left |> right(..args)` to be `right(left, ..args)` are more common than `right(..args)(left)`

giacomocavalieri:
Oh and here's a little preview showing it for pipes and regular function calls

https://github.com/user-attachments/assets/91374203-585a-4e8c-bba9-73836d4e93c7

giacomocavalieri:
Oops once again marking this as WIP because I noticed it doesn't work with `use` expressions, I'd like to sort that out before merging

compiler-core/src/type_/expression.rs
```diff
@@ -2918,17 +2916,19 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
                         // Then we keep the error but still want to continue analysing the arguments that were passed.
                         MatchFunTypeError::IncorrectArity {
-                            args, return_type, ..
+                            args,
+                            return_type,
+                            expected,
+                            given,
+                            ..
                         } => {
+                            missing_args = expected.saturating_sub(given);
```
Acepie, line 2925: Just wanted to get thoughts on this. Was about to start on https://github.com/gleam-lang/gleam/issues/3405 but I think this change here might just make this issue obsolete since if i'm reading this right you are adding the missing args in which would clear up any confusion?

giacomocavalieri: I only do that for use calls because I know the callback must be the last one, so I'm filling up the empty spaces with placeholder variables. For all other calls this wouldn't work so I'm guessing this doesn't solve the problem in that issue.
But now that I think about it we could do the same for calls with the wrong arity and labels to fill up the gaps and give better error messages, though it would be a little bit trickier to figure out where to put the placeholder args in case labels are used

lpil:
While testing I've found a behaviour which seems like it might be a bug. If the function being used does not exist it doesn't report seem to report that, and it shows incorrect signature help for this non-existent function.

image

lpil:
Can we have a test for that case ?

giacomocavalieri:
Oops, I totally forgot about it 😁 I've added one!

compiler-core/src/type_/pipe.rs
lpil: I think this may have changed the semantics of the language and may break existing programs.

Previously if a function was missing 1 argument and returned a function that took no arguments it would rewrite to `right(left, ...args)`, but now it rewrites to `right(...args)(left)`, which is a type error.
I think this should be `if arity == arguments.len() + 1 && return_.fn_arity() == Some(1)`?

Could you add a test for this case please, to ensure the behaviour has not changed. I could be reading this wrong, so having a test to check it and to ensure it doesn't happen in future by mistake would be good.

lpil (approved):
Heroic!!!!

PR #3413, opened 2024-07-19
Add action to move imports to top of module

### LINKED ISSUE #3186: LSP Action: Move all imports to the head for the module
This would speed up adding modules while typing.

If there are multiple import groups it could be consistently added to either the head or tailed group.

giacomocavalieri:
This PR closes #3186 

I also moved this action and the tuple one to their own module because it was starting to feel a bit overwhelming to have everything in the actions module. If you don't like this I can put everything back to just that file though!


compiler-core/src/language_server/engine.rs
```diff
@@ -688,11 +695,13 @@ fn code_action_fix_names(
...
 
-            CodeActionBuilder::new(&format!("Rename to {}", correction))
+            let action = CodeActionBuilder::new(&format!("Rename to {}", correction))
                 .kind(lsp_types::CodeActionKind::QUICKFIX)
                 .changes(uri.clone(), vec![edit])
                 .preferred(true)
-                .push_to(actions);
+                .build();
+
+            actions.push(action);
```
lpil, line 704: How come this has changed?

compiler-core/src/language_server/engine.rs
```diff
@@ -274,6 +278,7 @@ where
             code_action_fix_names(module, &params, &mut actions);
             actions.extend(LetAssertToCase::new(module, &params).code_actions());
             actions.extend(RedundantTupleInCaseSubject::new(module, &params).code_actions());
+            actions.extend(MoveImportsToTop::new(module, &params).code_actions());
```
lpil, line 281: If you're going to move them to module could you move all of them please. There are still two in this module. 🙏 

This bit here could look like this:

```rust
code_actions::all(module, &params, &actions)
```

compiler-core/src/language_server/tests/action.rs
```diff
@@ -751,6 +763,69 @@ fn test_convert_outer_let_assert_to_case() {
...
+    let code = r#"import list
+
+pub fn main() {}
+
+import result
+"#;
+
+    insta::assert_snapshot!(apply_first_code_action_with_title(code, 5, MOVE_IMPORTS_UP));
+}
+
```
lpil, line 828: Could we have a test with unqualified items please 🙏

compiler-core/src/language_server/tests/action.rs
```diff
@@ -751,6 +763,69 @@ fn test_convert_outer_let_assert_to_case() {
...
+    insta::assert_snapshot!(apply_first_code_action_with_title(code, 4, MOVE_IMPORTS_UP));
+}
+
+#[test]
+fn move_imports_to_the_top_of_the_module_moves_all_imports_when_on_import_below_a_definition_2() {
+    let code = r#"import list
+
+pub fn main() {}
+
+import result
```
lpil, line 823: What happens if the imports are lower but in groups? Do the groups get preserved?

What about comments?

lpil (commented):
Thanks! I've left a few notes inline as I think we need to cover things like import grouping and comments

compiler-core/src/language_server/engine.rs
giacomocavalieri: I'm going back to the single module for now!

compiler-core/src/language_server/engine.rs
giacomocavalieri: Oops my bad I was experimenting with a different API and forgot to undo it 😁

compiler-core/src/language_server/tests/action.rs
giacomocavalieri: Mmh this is quite tricky, basically I'd have to reimplement part of the pretty printer here but I'm not a fan of that... I'll have to think a bit more how to approach this

PR #3414, opened 2024-07-19
incorrect arity fault tolerance works better with labels

### LINKED ISSUE #3405: Function fault tolerance results in confusing errors when using labels and arity is incorrect

Acepie:
Closes #3405

Fixes double arity reporting for functions with labelled arguments and stops type checking of arguments before first labelled argument if arity is incorrect. This still leads to a good experience with signature help but prevents incorrect type checking of the labelled arguments. As far as I can tell this is the best we can do since putting placeholder arguments would break signature help but not putting placeholder arguments makes doing the reordering of arguments impossible in some cases


lpil (approved):
Thank you!!!

PR #3418, opened 2024-07-21
LSP Code Action: Import Module

### LINKED ISSUE #3258: LSP code action: import module
Right now, if I use a module that exists but was not imported yet I receive an error message:
```
Unknown module

No module has been found with name `<name>`.
```

It would be great if we could import directly using a code action

GearsDatapacks:
Closes #3258
This PR implements a code action to import modules.
- Unknown module errors now also report an `ImportSuggestion`, which checks for any modules whose last path component matches the name.
- I have moved some code which was previously only used in import completion into the main `engine.rs` file of the language server, so it can be used in this code action too.
- I had some trouble getting the tests working, since the tests I wrote don't actually import the modules which they rely on, and that means that the function to sort the modules into dependency order doesn't work, because it doesn't realise that the tested file depends on those modules existing already to be able to suggest them as imports. I've temporarily managed to fix this by renaming the file to `test_app`, and that seems to put it after the others because of alphabetical order. This isn't a good solution, and I'm not even sure that alphabetical order is guaranteed, so please let me know how I can fix this issue better.
- I have also changed my name on the changelog, I was unsure before as to whether to use my real name or internet name, and I've since changed my mind. I hope that's OK.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1663,7 +1663,7 @@ fn get_type_dependencies(type_: &TypeAst) -> Vec<EcoString> {
             ..
         }) => {
             deps.push(match module {
-                Some(module) => format!("{}.{}", name, module).into(),
+                Some((module, _)) => format!("{}.{}", name, module).into(),
                 None => name.clone(),
             });
 
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -292,7 +292,6 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             module_types_constructors: types_constructors,
             module_values: values,
             accessors,
-            import_suggestions,
             ..
         } = env;
 
@@ -803,7 +802,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             )
             .collect();
         let typed_parameters = environment
-            .get_type_constructor(&None, &name, location)
+            .get_type_constructor(&None, &name)
             .expect("Could not find preregistered type constructor ")
             .parameters
             .clone();
@@ -1340,7 +1339,7 @@ fn analyse_type_alias(t: UntypedTypeAlias, environment: &mut Environment<'_>) ->
     // analysis aims to be fault tolerant to get the best possible feedback for
     // the programmer in the language server, so the analyser gets here even
     // though there was previously errors.
-    let type_ = match environment.get_type_constructor(&None, &alias, location) {
+    let type_ = match environment.get_type_constructor(&None, &alias) {
         Ok(constructor) => constructor.type_.clone(),
         Err(_) => environment.new_generic_var(),
     };
@@ -1716,33 +1715,4 @@ fn sorted_type_aliases(aliases: &Vec<UntypedTypeAlias>) -> Result<Vec<&UntypedTy
         .iter()
         .sorted_by_key(|alias| sorted_deps.iter().position(|x| x == &alias.alias))
         .collect())
-}
-
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ImportSuggestion {
-    pub location: SrcSpan,
-    pub suggestion: EcoString,
-}
-
-pub fn suggest_imports(
-    module_name: &str,
-    location: SrcSpan,
-    importable_modules: &[&EcoString],
-    import_suggestions: &mut Vec<ImportSuggestion>,
-) {
-    import_suggestions.extend(importable_modules.iter().filter_map(|&option| {
-        if option
-            .split('/')
-            .last()
-            .unwrap_or(option)
-            .eq_ignore_ascii_case(module_name)
-        {
-            Some(ImportSuggestion {
-                location,
-                suggestion: option.clone(),
-            })
-        } else {
-            None
-        }
-    }));
 }
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -292,6 +292,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             module_types_constructors: types_constructors,
             module_values: values,
             accessors,
+            import_suggestions,
             ..
         } = env;
 
@@ -802,7 +803,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             )
             .collect();
         let typed_parameters = environment
-            .get_type_constructor(&None, &name)
+            .get_type_constructor(&None, &name, location)
             .expect("Could not find preregistered type constructor ")
             .parameters
             .clone();
@@ -1339,7 +1340,7 @@ fn analyse_type_alias(t: UntypedTypeAlias, environment: &mut Environment<'_>) ->
     // analysis aims to be fault tolerant to get the best possible feedback for
     // the programmer in the language server, so the analyser gets here even
     // though there was previously errors.
-    let type_ = match environment.get_type_constructor(&None, &alias) {
+    let type_ = match environment.get_type_constructor(&None, &alias, location) {
         Ok(constructor) => constructor.type_.clone(),
         Err(_) => environment.new_generic_var(),
     };
@@ -1715,4 +1716,33 @@ fn sorted_type_aliases(aliases: &Vec<UntypedTypeAlias>) -> Result<Vec<&UntypedTy
         .iter()
         .sorted_by_key(|alias| sorted_deps.iter().position(|x| x == &alias.alias))
         .collect())
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ImportSuggestion {
+    pub location: SrcSpan,
+    pub suggestion: EcoString,
+}
+
+pub fn suggest_imports(
+    module_name: &str,
+    location: SrcSpan,
+    importable_modules: &[&EcoString],
+    import_suggestions: &mut Vec<ImportSuggestion>,
+) {
+    import_suggestions.extend(importable_modules.iter().filter_map(|&option| {
+        if option
+            .split('/')
+            .last()
+            .unwrap_or(option)
+            .eq_ignore_ascii_case(module_name)
+        {
+            Some(ImportSuggestion {
+                location,
+                suggestion: option.clone(),
+            })
+        } else {
+            None
+        }
+    }));
 }
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1,7 +1,7 @@
 #![allow(clippy::unwrap_used, clippy::expect_used)]
 use crate::build::{Outcome, Runtime, Target};
 use crate::diagnostic::{Diagnostic, ExtraLabel, Label, Location};
-use crate::type_::error::{MissingAnnotation, ModuleSuggestion, UnknownTypeHint};
+use crate::type_::error::{MissingAnnotation, UnknownTypeHint};
 use crate::type_::error::{Named, RecordVariants};
 use crate::type_::{error::PatternMatchKind, FieldAccessUsage};
 use crate::{ast::BinOp, parse::error::ParseErrorType, type_::Type};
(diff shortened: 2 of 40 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1671,3 +1673,32 @@ fn sorted_type_aliases(aliases: &Vec<TypeAlias<()>>) -> Result<Vec<&TypeAlias<()
...
+    location: SrcSpan,
+    importable_modules: &[&EcoString],
+    import_suggestions: &mut Vec<ImportSuggestion>,
+) {
+    import_suggestions.extend(importable_modules.iter().filter_map(|&option| {
+        if option
+            .split('/')
+            .last()
+            .unwrap_or(option)
+            .eq_ignore_ascii_case(module_name)
```
lpil, line 1694: use `==` please

compiler-core/src/language_server/engine.rs
```diff
@@ -674,6 +677,46 @@ fn code_action_unused_imports(
         .push_to(actions);
 }
 
+fn code_action_import_module(
```
lpil, line 680: Put this in the code action module please

compiler-core/src/language_server/engine.rs
```diff
@@ -674,6 +677,46 @@ fn code_action_unused_imports(
...
+    let line_numbers = LineNumbers::new(&module.code);
+    let (first_import_pos, first_is_import) = first_import_in_module(module, &line_numbers);
+    let after_import_newlines = add_newlines_after_import(
+        first_import_pos,
+        first_is_import,
+        &line_numbers,
+        &module.code,
+    );
+
+    for suggestion in suggestions {
```
lpil, line 701: This may produce an overwhelming amount of code actions. Instead let's have one code action and then it presents a menu of options, like Rust Analyzer does

compiler-core/src/language_server/tests/action.rs
```diff
@@ -7,8 +7,8 @@ use lsp_types::{
 use super::*;
 
 const TEST_FILE_PATH: &str = match cfg!(target_family = "windows") {
-    true => r"\\?\C:\src\app.gleam",
-    false => "/src/app.gleam",
+    true => r"\\?\C:\src\test_app.gleam",
```
lpil, line 10: Revert this please, having the tests depend on this ordering would make them super brittle.

lpil (commented):
Nice, thank you.

I'm in the process removing all the LS specific code from the analyser as I don't think it should do anything to create code actions, rather the LS will iterate across errors and warnings to decide what actions to offer. Once it is merged this will need to be reworked to fit it: https://github.com/gleam-lang/gleam/pull/3415

compiler-core/src/language_server/tests/action.rs
GearsDatapacks: I agree. How do you suggest I fix this problem?

lpil: I don't have any suggestions here, I've not looked into it I'm afraid.

compiler-core/src/language_server/engine.rs
GearsDatapacks: Looking into Rust Analyzer, it seems that this behaviour is not part of the Language Server Protocol. Instead, they implement custom code action handling on the client. Should I try and figure that out? It would need to be implemented separately for each editor we support, as far as I can tell.

compiler-core/src/build/package_loader/tests.rs
```diff
@@ -57,6 +57,7 @@ fn write_cache(
         accessors: Default::default(),
         unused_imports: Vec::new(),
         name_corrections: Vec::new(),
+        import_suggestions: Vec::new(),
```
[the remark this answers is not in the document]
GearsDatapacks: I think so, although currently the errors store a list of currently imported modules to suggest, rather than importable modules, which is what is needed for this. There are two ways I could fix this, either we keep the error message the same, and store additional data in the error, or we change the error message to something like:
```
io.println("Hello")
^^ Unknown module io.

Hint: Did you mean to import gleam/io?
```
That's just written off the top of my head, I would rewrite it to be consistent with the other error messages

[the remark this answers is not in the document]
lpil: Having a suggestion in the error is a fantastic idea. We should do this for sure. We could look for a module that has the `println` member in this case.

GearsDatapacks:
I've made some changes:
- The diagnostic for an unknown module now suggests importing a module if one exists with the same name and exports a type/value with the same name as the accessed one.
- The code action now looks through the errors emitted, and suggests import based on those.
  - I've stored this error in the `LanguageServerEngine` because I wasn't sure where to put it. Please tell me where the code action functions should get the errors from
- It looks like the recent refactor of the code action tests has fixed the issue I was previously having, so I don't need a hacky solution any more

compiler-core/src/language_server/server.rs
```diff
@@ -252,6 +252,7 @@ where
                     }
                     Err(e) => {
                         let feedback = project.feedback.build_with_error(e, compilation, warnings);
+                        eprintln!("{feedback:#?}");
```
lpil, line 255: Debug printing here

compiler-core/src/language_server/engine.rs
```diff
@@ -141,11 +143,14 @@ where
...
 
-        outcome
+        let result = outcome
             // Register which modules have changed
             .map(|modules| self.modules_compiled_since_last_feedback.extend(modules))
             // Return the error, if present
-            .into_result()
+            .into_result();
+
+        self.error = result.clone().err();
```
lpil, line 152: Only clone if it is an error please

lpil (commented):
Nice!

Although with this change we no longer get a suggestion for when the user makes a typo with name of the module. I think we should search both the imported and not-yet imported modules so we can make suggestions from both.

GearsDatapacks:
> Is that something we'd want to do here, or is that for a later PR?
I can implement that, yep

GearsDatapacks:
Alright, I've added a check in case the user made a typo, but the code action is only offered if there is an exact matching module

compiler-core/src/type_/error.rs
```diff
@@ -71,6 +71,47 @@ pub enum RecordVariants {
...
+/// A suggestion for an unknown module
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum ModuleSuggestion {
+    /// A module which which has a matching name, and an
+    /// exported value matching the one being accessed
+    Matching(EcoString),
+    /// A module already imported in the current scope
+    Imported(EcoString),
+    /// A module which can be imported
+    Importable(EcoString),
```
lpil, line 83: What's the difference between importable and matching?

lpil (commented):
Thank you! Looking great! I've left 1 question inline

compiler-core/src/type_/error.rs
GearsDatapacks: Yeah, I was a little unsure how to implement this, because initially you suggested checking for a member with a name matching what was accessed. But then we also want to suggest in case the user made a typo. At the moment, `Matching` is a module with a matching name, and a matching export, whereas `Importable` is just any module which can be imported. `Matching` takes precedence over others, and  is the only one suggested in the code action. I don't really like how this works; do you have any suggestions as to how it should?

lpil: If the module doesn't have the value being accessed then does it make sense to suggest importing it at all? It would be incorrect after the import still.

GearsDatapacks: That makes sense. I'll only suggest for typos or unimported modules with matching names. Thanks!

GearsDatapacks:
Ok, I've had a think and reworked this code. The language server will now suggest a code action to correctly fix an unknown module error when:
- There's an importable module with a matching or similar name, with a public type/value of the same name
- There's an imported module with a similar name, with a public type/value of the same name

compiler-core/src/language_server/imports.rs
```diff
@@ -0,0 +1,77 @@
...
+    let first_definition = module
+        .ast
+        .definitions
+        .iter()
+        .min_by(|a, b| a.location().start.cmp(&b.location().start));
+    let import = first_definition.and_then(get_import);
+    let import_start = import.map_or(0, |i| i.location.start);
+    let import_line = line_numbers.line_number(import_start);
+    (Position::new(import_line - 1, 0), import.is_some())
+}
```
lpil, line 24: This function doesn't seem to do what it says. I think it should return `Option<Position>` and be called `position_of_first_definition_if_import` or something like that

compiler-core/src/language_server/imports.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub fn get_import(statement: &TypedDefinition) -> Option<&Import<EcoString>> {
```
lpil, line 26: Is this used anywhere?

compiler-core/src/language_server/imports.rs
```diff
@@ -0,0 +1,77 @@
+use ecow::EcoString;
```
lpil, line 1: Can we call this `edits` or something? Seems not to be able imports

compiler-core/src/type_/environment.rs
```diff
@@ -667,6 +670,63 @@ impl<'a> Environment<'a> {
...
+        suggestions
+            .into_iter()
+            .sorted()
+            .filter_map(|suggestion| {
+                edit_distance_with_substrings(module, suggestion.last_name_component(), threshold)
+                    .map(|distance| (suggestion, distance))
+            })
+            .sorted_by_key(|&(_, distance)| distance)
+            .map(|(suggestion, _)| suggestion)
+            .collect()
```
lpil, line 717: Only one is used so we can use min by rather than sort and make a vector

lpil (commented):
Thank you!

compiler-core/src/error.rs
```diff
@@ -619,7 +619,7 @@ fn edit_distance(a: &str, b: &str, limit: usize) -> Option<usize> {
     (distance <= limit).then_some(distance)
 }
 
-fn edit_distance_with_substrings(a: &str, b: &str, limit: usize) -> Option<usize> {
+pub fn edit_distance_with_substrings(a: &str, b: &str, limit: usize) -> Option<usize> {
```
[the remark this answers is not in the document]
GearsDatapacks: Good point. I just copied the code from `did_you_mean`, so I didn't really think about it

compiler-core/src/error.rs
```diff
@@ -1957,15 +1957,15 @@ Private types can only be used within the module that defines them.",
                 TypeError::UnknownModule {
                     location,
                     name,
-                    imported_modules,
+                    suggestions
                 } => Diagnostic {
                     title: "Unknown module".into(),
                     text: format!("No module has been found with the name `{name}`."),
-                    hint: None,
+                    hint: suggestions.first().map(|suggestion| suggestion.suggestion(name)),
```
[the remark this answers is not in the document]
GearsDatapacks: We do actually use the other suggestions in the code action, so we can suggest `gleam/io` and `some/other/io` for `io.println`, for example

compiler-core/src/language_server/imports.rs
GearsDatapacks: Several places. Much of this code was originally part of the import completion system, which has been moved to a separate file to be also used in this code action.

compiler-core/src/language_server/imports.rs
GearsDatapacks: I can do that. As noted below though, I didn't write any of the code in this file, simply moved it to a more accessible place.

compiler-core/src/language_server/imports.rs
lpil: Oh yes, I see now. Thanks

compiler-core/src/error.rs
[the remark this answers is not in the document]
lpil: Thank you

lpil (approved):
Super nice work! Feels great to use, thank you!!!

PR #3419, opened 2024-07-21
Introduce field punning

### LINKED ISSUE #3411: Field punning
```gleam
let day = something()
let month = something()
let year = something()
let date = Date(day:, month:, year:)
```

And `Date(day:, month:, year:)` expands to `Date(day: day, month: month, year: year)`.

Should work in:
- Record constructors, both patterns and values
- Labelled functions

https://github.com/gleam-lang/gleam/discussions/3146

giacomocavalieri:
This PR closes #3411 by introducing field punning for:
- function calls `fun(label:)`
- record constructors (both in constants and in functions) `Wibble(label:)`
- record constructors patterns `let Wibble(label:) = todo`

I've tried adding all the test cases I could think of, let me know if I've missed something!

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1666,10 +1666,6 @@ pub enum Pattern<Type> {
         location: SrcSpan,
         name: EcoString,
         type_: Type,
-        /// We need to keep track of this at the variable pattern level in order
-        /// to report better error messages for unused variables coming from
-        /// punned patterns.
-        is_punned: bool,
     },
 
     /// A reference to a variable in a bit array. This is always a variable
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1171,19 +1171,25 @@ pub struct CallArg<A> {
     pub label: Option<EcoString>,
     pub location: SrcSpan,
     pub value: A,
-    // This is true if this argument is given as the callback in a `use`
-    // expression. In future it may also be true for pipes too. It is used to
-    // determine if we should error if an argument without a label is given or
-    // not, which is not permitted if the argument is given explicitly by the
-    // programmer rather than implicitly by Gleam's syntactic sugar.
     pub implicit: Option<ImplicitCallArgOrigin>,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum ImplicitCallArgOrigin {
+    /// The implicit callback argument passed as the last argument to the
+    /// function on the right hand side of `use`.
+    ///
     Use,
+    /// An argument added by the compiler when rewriting a pipe `left |> right`.
+    ///
     Pipe,
+    /// An argument added by the compiler to fill in all the missing fields of a
+    /// record that are being ignored with the `..` syntax.
+    ///
     PatternFieldSpread,
+    /// An argument used to fill in the missing args when a function on the
+    /// right hand side of `use` is being called with the wrong arity.
+    ///
     IncorrectArityUse,
 }
 
@@ -1237,6 +1243,16 @@ impl CallArg<UntypedExpr> {
     }
 }
 
+impl<T> CallArg<T>
+where
+    T: HasLocation,
+{
+    #[must_use]
+    pub fn is_punned(&self) -> bool {
+        self.label.is_some() && self.location == self.value.location()
+    }
+}
+
 impl<T> HasLocation for CallArg<T> {
     fn location(&self) -> SrcSpan {
         self.location
@@ -1254,6 +1270,13 @@ pub struct UntypedRecordUpdateArg {
     pub label: EcoString,
     pub location: SrcSpan,
     pub value: UntypedExpr,
+}
+
+impl UntypedRecordUpdateArg {
+    #[must_use]
+    pub fn is_punned(&self) -> bool {
+        self.value.location() == self.location
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1666,6 +1666,10 @@ pub enum Pattern<Type> {
         location: SrcSpan,
         name: EcoString,
         type_: Type,
+        /// We need to keep track of this at the variable pattern level in order
+        /// to report better error messages for unused variables coming from
+        /// punned patterns.
+        is_punned: bool,
     },
 
     /// A reference to a variable in a bit array. This is always a variable
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -1704,29 +1704,67 @@ impl<'comments> Formatter<'comments> {
     }
 
     fn call_arg<'a>(&mut self, arg: &'a CallArg<UntypedExpr>, arity: usize) -> Document<'a> {
-        match &arg.label {
-            Some(s) => commented(
-                s.to_doc().append(": "),
-                self.pop_comments(arg.location.start),
-            ),
-            None => nil(),
+        self.format_call_arg(arg, expr_call_arg_formatting, |this, value| {
+            this.comma_separated_item(value, arity)
+        })
+    }
+
+    fn format_call_arg<'a, A, F, G>(
+        &mut self,
+        arg: &'a CallArg<A>,
+        figure_formatting: F,
+        format_value: G,
+    ) -> Document<'a>
+    where
+        F: Fn(&'a CallArg<A>) -> CallArgFormatting<'a, A>,
+        G: Fn(&mut Self, &'a A) -> Document<'a>,
+    {
+        match figure_formatting(arg) {
+            CallArgFormatting::Unlabelled(value) => format_value(self, value),
+            CallArgFormatting::Punned(label) => {
+                let comments = self.pop_comments(arg.location.start);
+                let label = label.as_ref().to_doc().append(":");
+                commented(label, comments)
+            }
+            CallArgFormatting::Labelled(label, value) => {
+                let comments = self.pop_comments(arg.location.start);
+                let label = label.as_ref().to_doc().append(": ");
+                let value = format_value(self, value);
+                commented(label, comments).append(value)
+            }
         }
-        .append(self.comma_separated_item(&arg.value, arity))
     }
 
     fn record_update_arg<'a>(&mut self, arg: &'a UntypedRecordUpdateArg) -> Document<'a> {
         let comments = self.pop_comments(arg.location.start);
-        let doc = arg
-            .label
-            .as_str()
-            .to_doc()
-            .append(": ")
-            .append(self.expr(&arg.value));
+        match arg {
+            // Punned argument.
+            _ if arg.is_punned() => commented(arg.label.as_str().to_doc().append(":"), comments),
 
-        if arg.value.is_binop() || arg.value.is_pipeline() {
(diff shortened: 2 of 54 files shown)


giacomocavalieri:
The parse failure messages are exactly as they are now for similar cases but I think those could be improved, what do you think?

compiler-core/src/ast.rs
```diff
@@ -1643,6 +1666,10 @@ pub enum Pattern<Type> {
         location: SrcSpan,
         name: EcoString,
         type_: Type,
+        /// We need to keep track of this at the variable pattern level in order
```
lpil, line 1669: What's a variable pattern level?

compiler-core/src/format.rs
```diff
@@ -1704,29 +1704,67 @@ impl<'comments> Formatter<'comments> {
...
+            _ if arg.is_punned() => commented(arg.label.as_str().to_doc().append(":"), comments),
+
+            // A call arg that wasn't explicitly punned but could be. We want to
+            // enforce a consistent style and only one way to do things so if
+            // something can be punned it should be.
+            UntypedRecordUpdateArg {
+                label,
+                value: UntypedExpr::Var { name, .. },
+                ..
+            } if label == name => commented(arg.label.as_str().to_doc().append(":"), comments),
```
lpil, line 1751: I checked other formatters and asked around what people think of this and it is unpopular. Let's let the programmer pick which syntax to use.

compiler-core/src/format.rs
```diff
@@ -2807,3 +2842,90 @@ fn is_breakable_argument(expr: &UntypedExpr, arity: usize) -> bool {
...
+            CallArgFormatting::Punned(arg.label.as_ref().expect("punned arg with no label"))
+        }
+        // A call arg that wasn't explicitly punned but could be. We want to
+        // enforce a consistent style and only one way to do things so if
+        // something can be punned it should be.
+        CallArg {
+            label: Some(label),
+            value: UntypedExpr::Var { name, .. },
+            ..
+        } if label == name => CallArgFormatting::Punned(label),
```
lpil, line 2865: As above, remove this please 🙏

compiler-core/src/format.rs
```diff
@@ -2807,3 +2842,90 @@ fn is_breakable_argument(expr: &UntypedExpr, arity: usize) -> bool {
...
+            CallArgFormatting::Punned(arg.label.as_ref().expect("punned arg with no label"))
+        }
+        // A call arg that wasn't explicitly punned but could be. We want to
+        // enforce a consistent style and only one way to do things so if
+        // something can be punned it should be.
+        CallArg {
+            label: Some(label),
+            value: UntypedPattern::Variable { name, .. },
+            ..
+        } if label == name => CallArgFormatting::Punned(label),
```
lpil, line 2892: As above, remove this please 🙏

compiler-core/src/format.rs
```diff
@@ -2807,3 +2842,90 @@ fn is_breakable_argument(expr: &UntypedExpr, arity: usize) -> bool {
...
+        // A call arg that wasn't explicitly punned but could be. We want to
+        // enforce a consistent style and only one way to do things so if
+        // something can be punned it should be.
+        CallArg {
+            label: Some(label),
+            value: Constant::Var {
+                name, module: None, ..
+            },
+            ..
+        } if label == name => CallArgFormatting::Punned(label),
```
lpil, line 2921: As above, remove this please 🙏

compiler-core/src/format/tests.rs
```diff
@@ -2711,23 +2711,23 @@ fn pattern_constructor() {
...
+  let Person(name: the_name, age: the_age) = 1
   Nil
 }
 "#
     );
 
     assert_format!(
         r#"fn main() {
-  let Person(age: age, name: name) = 1
+  let Person(age: the_age, name: the_name) = 1
```
lpil, line 2730: Revert the above test changes please

compiler-core/src/format/tests/binary_operators.rs
```diff
@@ -136,8 +136,8 @@ fn labelled_field_with_binary_operators_are_not_broken_if_they_can_fit() {
         r#"pub fn main() {
   Ok(wibble(
     name: names.name,
-    text: text,
-    code: code,
+    text:,
+    code:,
```
lpil, line 140: Revert these please

compiler-core/src/format/tests/cases.rs
```diff
@@ -96,9 +96,9 @@ fn case_pattern_split_on_multiple_lines_is_not_needlessly_nested() {
         r#"pub fn main() {
   case thing {
     CannotSaveNewSnapshot(
-      reason: reason,
-      title: title,
-      destination: destination,
+      reason: _reason,
+      title: _title,
+      destination: _destination,
```
lpil, line 101: Revert these please

compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_punning.snap
```diff
@@ -0,0 +1,12 @@
+---
+source: compiler-core/src/parse/tests.rs
+expression: "\npub fn main() {\n  wibble(:)\n}\n"
+---
+error: Syntax error
+  ┌─ /src/parse/error.gleam:3:10
+  │
+3 │   wibble(:)
+  │          ^ There must be a 'let' to bind variable to value
```
lpil, line 9: This seems like an odd error

compiler-core/src/parse/tests.rs
```diff
@@ -1213,3 +1213,91 @@ case my_string {
...
+
+#[test]
+fn invalid_pattern_punning_3() {
+    assert_module_error!(
+        "
+pub fn main() {
+  let Wibble(arg::) = todo
+}
+"
+    );
```
lpil, line 1302: Could we have a test for `Wibble(arg: arg:)` please. I'm curious as to what error it prompts

compiler-core/src/type_/error.rs
```diff
@@ -697,7 +698,22 @@ pub enum Warning {
 }
 
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
+pub enum VariableKind {
```
lpil, line 701: This doesn't seem right. Punning is not a kind of variable, rather it's property of the container a level above. Can this information be stored there instead?

compiler-core/src/type_/tests/errors.rs
```diff
@@ -829,6 +829,11 @@ fn unexpected_labelled_arg() {
     assert_module_error!(r#"fn id(x) { x } fn y() { id(x: 4) }"#);
 }
 
+#[test]
+fn unexpected_punned_arg() {
+    assert_module_error!(r#"fn id(x) { x } fn y() { let x = 4 id(x:) }"#);
```
lpil, line 834: I'm struggling to read this, please write code on multiple lines

compiler-core/src/parse.rs
```diff
@@ -1638,13 +1640,18 @@ where
...
-                            start: col_s,
-                            end: col_e,
+                    // Punned argument.
+                    Ok(Some(CallArg {
+                        implicit: None,
+                        location: SrcSpan { start, end },
+                        label: Some(name.clone()),
+                        value: UntypedPattern::Variable {
+                            name,
+                            location: SrcSpan { start, end },
```
lpil, line 1650: This span doesn't include the whole AST node, the `:` is outside of it

compiler-core/src/parse.rs
```diff
@@ -1681,7 +1689,15 @@ where
                     value,
                 }))
             } else {
-                self.next_tok_unexpected(vec!["An expression".into()])
+                // A punned argument.
+                Ok(Some(UntypedRecordUpdateArg {
+                    label: label.clone(),
+                    location: SrcSpan { start, end },
```
lpil, line 1695: This leaves the `:` outside

compiler-core/src/parse.rs
```diff
@@ -1934,13 +1950,11 @@ where
     //   a: _
     //   a: expr
     fn parse_fn_arg(&mut self) -> Result<Option<ParserArg>, ParseError> {
-        let mut start = 0;
         let label = match (self.tok0.take(), &self.tok1) {
-            (Some((s, Token::Name { name }, _)), Some((_, Token::Colon, _))) => {
+            (Some((start, Token::Name { name }, end)), Some((_, Token::Colon, _))) => {
```
lpil, line 1954: Same here

compiler-core/src/parse.rs
```diff
@@ -2767,8 +2808,19 @@ where
                     label: None,
                 }))
             }
-        } else if name.is_some() {
-            self.next_tok_unexpected(vec!["a constant value".into()])?
+        } else if let Some((start, label, end)) = label {
+            Ok(Some(CallArg {
+                implicit: None,
+                location: SrcSpan { start, end },
```
lpil, line 2814: Here too

lpil (commented):
Thank you! I've left various comments inline

compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_punning.snap
giacomocavalieri: Yeah it's really strange, it was not introduced by this PR and is already what the compiler is producing today. Do you think I could open an issue and address it in a different PR?

compiler-core/src/parse/tests.rs
giacomocavalieri: Added! <https://github.com/gleam-lang/gleam/pull/3419/commits/6665617b3ec565b00eb72b6d8212a6521ea4d89f>

compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_punning.snap
lpil: Sounds good to me

compiler-core/src/type_/error.rs
giacomocavalieri: So this is needed specifically to track a variable's origin and report better error messages (maybe `VariableOrigin` would be a better name) on how to deal with unused variables coming from pattern matching on record constructors:
```rs
hint: match kind {
  VariableKind::Punned => "You can ignore it with an underscore: `{name}: _`.",
  VariableKind::Regular => "You can ignore it with an underscore: `_{name}`.",
}
```
Maybe I'm misunderstanding your question but I don't see how I could move it elsewhere since the warning needs this kind of information to display different hints

lpil: Yeah I understand the need, but it's not correct as it's both misnamed and it's stored on the wrong AST node. Data about the layer above needs to be in the layer above.

compiler-core/src/ast.rs
giacomocavalieri: Oops poor wording on my part, what I meant to say is we need* to store this information directly on the `Pattern::Variable` node instead on the `CallArg` container to give better error messages

\* Actually there could be a way to do this without duplicating this information on the pattern node as well but it would be a pretty big signature change to the `unify` function and as I was trying that, it made the code much harder to read. So I opted for the smallest change here.
But if you prefer I can show you the other approach as well!

lpil: It's not data about the variable so it should not be on the variable node.

lpil (approved):
Thank you!!

PR #3426, opened 2024-07-22
Fix 2015

### LINKED ISSUE #2015: Formatter relocates docs when function is commented out
Example:

```gleam
/// Does something.
pub fn func1() {
  todo
}

/// Does something else.
pub fn func2() {
  todo
}
```

I'm working on `func1`, but in the middle of that, realize I want to make changes somewhere else first. The code in `func1` is (temporarily) invalid syntax and will make the formatter throw an error, so I comment it out:

```gleam
/// Does something.
//pub fn func1() {
//  some half-baked code in here
//}

/// Does something else.
pub fn func2() {
  todo
}
```

Then, while making those other changes somewhere else, I run the formatter:

```gleam
//pub fn func1() {
//  some half-baked code in here
//}

/// Does something.
/// Does something else.
pub fn func2() {
  todo
}
```

My intended doc comment for `func1` has been "merged" into the doc for `func2`. This also happens if `func2` has no doc; the `func1` doc is relocated to just above `func2`.

After seeing this happen I can understand why it happens, but it's still pretty unexpected, and could introduce mistakes into published docs if you don't notice it happening.

My initial suggestion would be that a "floating" doc comment like this should be treated as a normal comment when parsing, and that maybe `gleam docs` should output a warning about it ("function doc ignored: not connected to a function"). This seems like it might be a bit complicated to implement, though...

There is also the workaround of only commenting out the function _body_ and not the whole definition, which will cause the formatter to insert a `todo` but otherwise not mess with it. I'd be fine if the solution is "just remember to do it that way"; this is not something I expect to come up very often.

giacomocavalieri:
This PR closes #2015 

This would happen if one commented out a documented definition:
```gleam
/// Wibble's
// pub fn wibble() {}

/// Wobble's
pub fn wobble() {}
```
The compiler would merge wibble's documentation into wobble's one resulting in some confusing behaviour (the formatter would clump those doc strings together, ignoring the comment; the generated doc for wobble would also include wibble's doc; hovering over wobble would also show wibble's doc; ...).

Now if a doc comment comes before a regular comment it is considered as not belonging to any definition. In addition I've implemented some changes in the formatter to make this more evident in case someone messed up:

```gleam
/// Docstring
// some random comment
pub fn wibble() {}
```
Is formatted into:
```gleam
/// Docstring
// some random comment

pub fn wibble() {}
```
I think the additional space makes it clearer that the doc string is not related to the `wibble` function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -118,7 +118,53 @@
   of the expression.
   ([Ameen Radwan](https://github.com/Acepie))
 
+- Documentation comments that come before a regular comment are no longer
+  clumped together with the documentation of the following definition.
+  Now commenting out a definition won't result in its documentation merging with
+  the following one's.
+
+  ```gleam
+  /// This doc comment will be ignored!
+  // a commented definition
+  // fn wibble() {}
+
+  /// Wibble's documentation.
+  fn wibble() { todo }
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Formatter
+
+- The formatter will no longer move a documentation comment below a regular
+  comment following it. This snippet of code is left as it is by the formatter:
+
+  ```gleam
+  /// This doc comment will be ignored!
+  // a commented definition
+  // fn wibble() {}
+
+  /// Wibble's documentation.
+  fn wibble() {
+    todo
+  }
+  ```
+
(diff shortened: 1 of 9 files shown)


lpil (approved):
Bravo!!

giacomocavalieri:
Done!

PR #3427, opened 2024-07-22
Label completions

Acepie:
Gives completions for labels when calling a function or constructing a record. Hopefully this should encourage people to use labels more/make using labels easier especially with the new punning syntax allowing people to not need local variable completion as much

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -645,9 +645,9 @@ where
     pub fn completion_labels(
         &'a self,
         fun: &TypedExpr,
-        existing_args: &Vec<CallArg<TypedExpr>>,
+        existing_args: &[CallArg<TypedExpr>],
     ) -> Vec<CompletionItem> {
-        let fun_type = fun.type_().fn_types().and_then(|(args, _)| Some(args));
+        let fun_type = fun.type_().fn_types().map(|(args, _)| args);
         let already_included_labels = existing_args
             .iter()
             .filter_map(|a| a.label.clone())
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -10,13 +10,13 @@ use lsp_types::{
 use strum::IntoEnumIterator;
 
 use crate::{
-    ast::{Definition, Import, Publicity, TypedDefinition},
+    ast::{CallArg, Definition, Import, Publicity, TypedDefinition, TypedExpr},
     build::Module,
     io::{CommandExecutor, FileSystemReader, FileSystemWriter},
     line_numbers::LineNumbers,
     type_::{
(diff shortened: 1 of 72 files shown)


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__completions_for_record_labels.snap
```diff
@@ -0,0 +1,148 @@
+---
+source: compiler-core/src/language_server/tests/completion.rs
+expression: "completion(TestProject::for_source(code), Position::new(6, 23))"
```
lpil, line 3: I can't tell what is happening in these snapshots. It's a lot of data and the expression doesn't show what is being tested at all. Could we make them readable please 🙏

compiler-core/src/language_server/completer.rs
```diff
@@ -622,6 +623,60 @@ where
...
+        &'a self,
+        fun: &TypedExpr,
+        existing_args: &[CallArg<TypedExpr>],
+    ) -> Vec<CompletionItem> {
+        let fun_type = fun.type_().fn_types().map(|(args, _)| args);
+        let already_included_labels = existing_args
+            .iter()
+            .filter_map(|a| a.label.clone())
+            .collect_vec();
+        if let Some(field_map) =
```
lpil, line 655: Prefer guards over nesting, for example:
```rust
let Some(field_map) = ... else {
  return vec![]
};
```

lpil (commented):
Oh nice! That was a lot easier than I expected. Fab stuff

I've left a note inline about finding the snapshots opaque. I think they would be hard to debug or change going forward

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__completions_for_record_labels.snap
Acepie: I'm not entirely sure how to make this more readable honestly. I'll add a comment in the code itself saying something like " // completion inside parens below includes labels" so that at least it is a bit more clear what the point of the test is

Acepie: oh actually lemme see if I can make the expression at least show the actual code

Acepie: lmk if this helps or if there was something in particular you had in mind

Acepie:
Ok completions now have a nice little cursor rendered where the completion takes place. I think this should be nicer to debug now

lpil (approved):
Oh that is super nice, thank you!!

PR #3428, opened 2024-07-23
Fix 3421

### LINKED ISSUE #3421: Code action to pun a fields where possible
It could apply to all punnable fields of a call or update.

giacomocavalieri:
This PR closes #3421.
The action will update all labelled args under its selection that could use the label shorthand syntax. In addition, to have a consistent naming and a bit less jargon I've turned "pun" into "shorthand label"


lpil (commented):
Very nice!!

compiler-core/src/language_server/engine.rs
```diff
@@ -618,11 +621,25 @@ pub fn overlaps(a: lsp_types::Range, b: lsp_types::Range) -> bool {
...
+        || within_including_end(b.end, a)
+}
+
+// Returns true if a position is within a range.
 fn within(position: lsp_types::Position, range: lsp_types::Range) -> bool {
     position >= range.start && position < range.end
 }
 
+// Returns true if a position is within a range, including its end.
+fn within_including_end(position: lsp_types::Position, range: lsp_types::Range) -> bool {
```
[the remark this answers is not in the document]
giacomocavalieri: I think we should! I wasn't sure though

compiler-core/src/format.rs
```diff
@@ -1734,13 +1734,13 @@ impl<'comments> Formatter<'comments> {
         G: Fn(&mut Self, &'a A) -> Document<'a>,
     {
         match figure_formatting(arg) {
-            CallArgFormatting::Unlabelled(value) => format_value(self, value),
-            CallArgFormatting::Punned(label) => {
+            CallArgFormatting::WithoutLabel(value) => format_value(self, value),
```
lpil, line 1737: I think this is worse 😅  Can we stick with the original naming please for the other two and have something that fits for the new variant.

ShorthandLabelled ?

giacomocavalieri: Ok!

lpil (approved):
Thank you!!

PR #3429, opened 2024-07-23
Label shorthand bug

giacomocavalieri:
This fixes a bug @inoas found where the formatter would treat a labelled function hole as a shorthand label:
```gleam
[] |> list.map(with: fun, over: _)
// Would incorrectly be formatted as
[] |> list.map(with: fun, over:)
```

This PR is based on #3428 so that should be merged first


lpil (approved):
Thank you

PR #3434, opened 2024-07-24
Action to fill labelled args

giacomocavalieri:
This PR:
- Adds an action to fill in the missing labelled arguments of a function call
  - It only works on calls that haven't been explicitly supplied args
  - It still plays nicely with pipes and the use syntax
- Fixes a bug where the LS would give the wrong signature hint for calls with labelled arguments
- Changes the code action snapshot:
  - Now the snapshots include the original code and highlight the selected code that triggered the action
  - All the tests have been rewritten to use the `PositionFinder` api to allow more granular control over the pieces of code to select to test if an action can be triggered. Before it was only possible to select entire lines

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -2,18 +2,25 @@ use std::{iter, sync::Arc};
 
 use crate::{
     ast::{
-        self, visit::Visit as _, AssignName, AssignmentKind, CallArg, Pattern, SrcSpan, TypedExpr,
+        self,
+        visit::{visit_typed_expr_call, Visit as _},
+        AssignName, AssignmentKind, CallArg, ImplicitCallArgOrigin, Pattern, SrcSpan, TypedExpr,
         TypedPattern, TypedRecordUpdateArg,
     },
     build::Module,
     line_numbers::LineNumbers,
     parse::extra::ModuleExtra,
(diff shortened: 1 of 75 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -508,3 +515,114 @@ impl<'ast> ast::visit::Visit<'ast> for LabelShorthandSyntax<'_> {
...
+                ),
+                new_text: missing_labels
+                    .iter()
+                    .sorted_by_key(|(position, _label)| *position)
+                    .map(|(_, label)| format!("{label}: todo"))
+                    .join(", "),
+            };
+
+            let mut action = Vec::with_capacity(1);
+            CodeActionBuilder::new("Fill in missing labelled arguments")
```
lpil, line 583: The name is a bit long and I think it implies it works when there's already some arguments. "Fill labels" or something? Rust calls it "Fill struct fields"

lpil (commented):
Very nice!!

One small note inline, and changelog please

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Ok! Sounds good to me

lpil (approved):
Beautiful!

PR #3435, opened 2024-07-24
LSP: Assign unused Result value to `_`

### LINKED ISSUE #2960: LSP code action: add `let _ =` to focused unused value

Frank-III:
hopefully resolve #2960

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -914,7 +914,7 @@ fn code_action_unused_values(
     actions: &mut Vec<CodeAction>,
 ) {
     let uri = &params.text_document.uri;
-    let unused: Vec<&SrcSpan> = module
+    let mut unused_values: Vec<&SrcSpan> = module
         .ast
         .type_info
         .warnings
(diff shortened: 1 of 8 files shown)


compiler-core/src/language_server/tests/action.rs
```diff
@@ -845,6 +846,113 @@ pub type Wibble { Wibble(arg1: Int, arg2: Int) }
...
+        Nil
+    }
+    Nil
+}
+"
+    );
+}
+
+// #[test]
+// fn test_assign_unused_result_on_block_start() {
```
Frank-III, line 881: do you have any suggestions on how to test code actions that only apply when on a specific line

compiler-core/src/language_server/engine.rs
```diff
@@ -68,6 +68,9 @@ pub struct LanguageServerEngine<IO, Reporter> {
     // the usual request-response loop.
     progress_reporter: Reporter,
 
+    // Keep track of unused value from warnings
+    unused_values: Vec<SrcSpan>,
```
lpil, line 72: Rather than storing unused values locations in the engine the engine can on-demand traverse the warnings in the module interface to find the locations.

compiler-core/src/language_server/tests/action.rs
lpil: Sorry, what do you mean? The first argument is the line

lpil (commented):
Nice!

compiler-core/src/language_server/engine.rs
Frank-III: Thank you for the suggestion! It looks way cleaner now!

compiler-core/src/language_server/tests/action.rs
Frank-III: Sorry, I should make it clearer. This is what I am trying to achieve(similar to what `rust-analyzer` does):
```gleam
pub fn main() {
//  ↓ will show the trigger to turn it into: let _ = { Ok(0)\nOk(0)\n };
    {
        Ok(0) // will only show one action to turn it into: let _= Ok(0) but no the one for outer block
        Ok(0)
    }
//  ↑ will show the trigger to turn it into: let _ = { Ok(0)\nOk(0) }
    Nil
}
```
and the `engine_response` function use `range: Range::new(Position::new(0, 0), Position::new(line + 1, 0))` for code actions which will fail if I want to test if the code action only show up on a specific line like the above example. I hope it is clear this time🙏.

lpil: The existing behaviour seems OK to me?

Frank-III: Roger, I made it only for the start and end line because when I was testing this feature, showing 2 actions confused me a little bit. If you feel okay with that, I will do it.
image

lpil: I think it should only show for the inner-most one

compiler-core/src/language_server/engine.rs
```diff
@@ -908,6 +909,62 @@ fn position_within(position: lsp_types::Position, range: lsp_types::Range) -> bo
...
+            .unwrap_or_default();
+
+        // only show actions for the most inner span
+        if (start_char == "{"
+            && params.range.start.line != hover_range.start.line
+            && params.range.end.line != hover_range.end.line)
+            || !overlaps(params.range, hover_range)
+        {
+            continue;
+        }
```
lpil, line 953: What does this do? It says "only show actions for the most inner span", but it seems to be checking if the first character is a `{`, not if it is the inner-most span.

lpil (commented):
Nice, thank you. I've left a note inline, and the changelog will need updating.

compiler-core/src/language_server/engine.rs
Frank-III: fixed now!

lpil (approved):
Hero! Thank you so much!!!

PR #3442, opened 2024-07-27
Make sure shorthand label code action triggers in all cases

giacomocavalieri:
I was testing the compiler on one of my projects and noticed the "Use label shorthands" action didn't work properly for:
- nested calls `wibble(wibble(arg: arg))`
- nested record updates `Wibble(..todo, arg: Wibble(..todo, arg: arg))`
- nested patterns `Ok(Wibble(arg: arg))`
- alternative patterns `Wibble(arg: arg) | Wibble(arg: arg)`

I've fixed all those bugs and now it should work for all possible cases!


lpil (approved):
Thank you

PR #3444, opened 2024-07-28
right-align and restyle opaque tag

### LINKED ISSUE #3340: HTML documentation: "opaque" indicator is mispositioned when there are source links
image

This does not look good

graphiteisaac:
small and simple PR! this just right aligns the `opaque` tag for generated documentation. I've also removed the border and darkened the background to keep it standing out but make it a bit less intense, since it is not something that needs to be so in-your-face. This fixes #3340 

Pasted image


lpil (approved):
Thank you!!

PR #3445, opened 2024-07-28
Completion sorting

Acepie:
Completions now include a sort text and all the places where we create CompletionItems (except for module imports where there will only ever be module names with no sort preferences) now call `sort_text` with a particular priority. Priority is currently only based on where a completion came from but in the future this can be extended further to include other variables like if the type is correct or not


compiler-core/src/language_server/completer.rs
```diff
@@ -25,6 +25,29 @@ use super::{
     compiler::LspProjectCompiler, files::FileSystemProxy, DownloadDependencies, MakeLocker,
 };
 
+#[derive(Copy, Clone)]
+enum SortPriority {
```
lpil, line 29: This is called "sort priority", but it's not a sort priority, it's a description of the item which is used to determine the priority. CompletionKind or something perhaps?

compiler-core/src/language_server/completer.rs
```diff
@@ -25,6 +25,29 @@ use super::{
...
+    // Types defined in the prelude
+    Prelude,
+    // Types defined in a module that has not been imported
+    ImportableModule,
+}
+
+// Gives the sort text for a completion item based on the priority and label.
+// This ensures that higher priority items are placed before lower priority items.
+fn sort_text(priority: SortPriority, label: &str) -> String {
+    let priority = priority as u8;
```
lpil, line 47: This seems a bit cryptic and error prone. Perhaps a big bold comment is needed on the definition to say that the ordering matter, which is a unique property for this type.

lpil (commented):
Oh very nice!! Thank you

compiler-core/src/language_server/completer.rs
Acepie: Yeah fair enough, I changed it to CompletionKindPriority since the priority is part of the type definition itself

compiler-core/src/language_server/completer.rs
Acepie: Yeah totally fair, I added a nice doc comment to the type definition. lmk if that's enough or if i should add something here as well

lpil:
It still says it a priority, but it is not.

Acepie:
hmmm i disagree/think changing this makes it less clear but I'll defer to you on this since I don't think its worth spending time on

lpil:
I agree it is less clear. It needs to be both correct and clear.

Perhaps not using the int casting would help? The implicit mapping between constructors and ordering is still quite cryptic to me.

Or there could be a priority value passed in, as the previous name stated.

Acepie:
I can turn it into an explicit match if you'd prefer though I think using the enum as an int here is fine

Acepie:
Lmk if this seems better to you. Kept the kind naming but switched to an explicit match

lpil (approved):
Brilliant! Thanks for going back and forth with me on this one too

Issue #3294, opened 2024-06-19
Imports renamed with `as` then used in consts generate invalid JavaScript code


richard-viney:
On Gleam 1.2.1, the following code:

```gleam
import gleam/order.{Gt as GreaterThan}

const my_const = GreaterThan

pub fn main() {
  my_const
}
```

Works fine on Erlang, but on JavaScript generates invalid code that gives the following runtime error:

```
file:///./build/dev/javascript/tmp/tmp.mjs:8
const my_const = new Gt();
                ^

ReferenceError: Gt is not defined
    at file:///./build/dev/javascript/tmp/tmp.mjs:8:17
    at ModuleJob.run (node:internal/modules/esm/module_job:262:25)
    at async ModuleLoader.import (node:internal/modules/esm/loader:475:24)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:109:5)

Node.js v22.2.0
```

lpil:
Oh dear! Thank you

PR #3447, opened 2024-07-28
Fix record constructors in guards using non-aliased names in JS

PgBiel:
It appears the fix for https://github.com/gleam-lang/gleam/issues/3294 was incomplete (only fixed aliased constructors in top-level constants, not in guards). This PR fixes this.


lpil (approved):
Thank you

PR #3449, opened 2024-07-30
Remove unused import

sobolevn:
It is already there: https://github.com/gleam-lang/gleam/blob/f84b4ac2db55f73a33e1cfbcfa30a5aead1c51d7/compiler-core/src/language_server/engine.rs#L23


lpil (approved):
Thank you

Issue #3424, opened 2024-07-22
Confusing error message for function call with colon


giacomocavalieri:
```gleam
pub fn main() {
  wibble(:)
}
```
Results in a confusing syntax error:
```text
error: Syntax error
  ┌─ /src/main.gleam:4:10
  │
4 │   wibble(:)
  │          ^ There must be a 'let' to bind variable to value

Hint: Use let for binding.
See: https://tour.gleam.run/basics/assignments/
```

lpil:
Thank you

PR #3450, opened 2024-07-30
Fix confusing `There must be a let to bind variable to value` error

sobolevn:
I don't think that it was safe to check for `:` and `=` in expression parsing. We should instead validate statements that look like assignments start with `let`.

Instead `wobble(:)` now generates `I was not expecting this`, which is expected by me 😆 

Closes https://github.com/gleam-lang/gleam/issues/3424

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -996,7 +996,7 @@ where
         // Better error: name definitions must start with `let`
         if let Some((_, Token::Name { .. }, _)) = self.tok0.as_ref() {
             match self.tok1 {
-                Some((start, Token::Equal, end)) | Some((start, Token::Colon, end)) => {
+                Some((start, Token::Equal | Token::Colon, end)) => {
                     return parse_error(ParseErrorType::NoLetBinding, SrcSpan { start, end })
                 }
 
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -201,6 +201,15 @@ pub fn parse_const_value(src: &str) -> Result<Constant<(), ()>, ParseError> {
 }
 
 //
+// Parser's state: do we bind a name at the moment?
+//
+#[derive(Debug)]
+enum NameBindingState {
+    Empty,
+    Let,
+}
+
+//
 // Parser
 //
 #[derive(Debug)]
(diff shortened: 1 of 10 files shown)


compiler-core/src/parse/tests.rs
```diff
@@ -427,7 +427,11 @@ fn no_let_binding3() {
         "[x] = [2]",
         ParseError {
             location: SrcSpan { start: 4, end: 5 },
-            error: ParseErrorType::NoLetBinding
+            error: ParseErrorType::UnexpectedToken {
```
giacomocavalieri, line 430: This should stay as a `NoLetBinding` error

compiler-core/src/parse.rs
```diff
@@ -599,11 +610,21 @@ where
...
-                UntypedExpr::List {
+                let list = UntypedExpr::List {
                     location: SrcSpan { start, end },
                     elements,
                     tail,
+                };
+
+                // We also need to check for a special case: `[x] = ...`
+                // It only should be used in `let` context.
+                if !matches!(self.name_binding_state, NameBindingState::Let) {
```
lpil, line 621: Prefer a `is_` method over `matches!`, we generally don't use matches.

lpil (commented):
Thank you. I'm not sure about this state that has been added. Could it instead be a parameter to parse_expression? As that's where it is used.

> Migrating to a full grammar parser one day might be a very good solution, because parser hacks will be increasing in the future :)

What do you mean by this? Gleam has a fair simple and straightforward grammar.

> Current limitation of two tokens per step - gives very little context.

This is a design goal. Back tracking is the sign of a poorly designed grammar and has a large impact on parser performance.

sobolevn:
@lpil thanks a lot for the quick feedback!

> I'm not sure about this state that has been added. Could it instead be a parameter to parse_expression? As that's where it is used.

Yes, I fully agree, current solution is better.

> What do you mean by this? Gleam has a fair simple and straightforward grammar.

I am a big fan of formal grammars 😊 
Like we have in https://github.com/python/cpython/blob/main/Grammar/python.gram

It allows us to create declarative cases for common errors with ease:
- `def some[](): ...` will match: https://github.com/python/cpython/blob/8fb88b22b7a932ff16002dd19e904f9cafd59e9f/Grammar/python.gram#L1445-L1449
- `print 1` will match: https://github.com/python/cpython/blob/8fb88b22b7a932ff16002dd19e904f9cafd59e9f/Grammar/python.gram#L1163-L1166

But, this is more like a thinking out loud, not anything else :)

sobolevn:
Done! 👍

lpil (approved):
Thank you!!!

PR #3456, opened 2024-07-31
Do not create intermediate vector in `pipe.infer_each_expression`

sobolevn:
Looks like that line is no longer needed. I am not sure if compiler optimizes this or not, so why not be specific?


lpil (approved):
Thank you! I've no idea why that was there.

PR #3459, opened 2024-07-31
add `GLEAM_CACERTS_PATH` env variable

### LINKED ISSUE #3246: Add support to custom CA certificate stores
Many corporate environments are protected by firewalls that act as [SSL Forward Proxy](https://docs.paloaltonetworks.com/pan-os/9-1/pan-os-admin/decryption/decryption-concepts/ssl-forward-proxy), effectively breaking the SSL connection and replacing the root certificate for one owned by the organization. For the general browsing use case that is not an issue, since the organization root CA is added to the trusted certificate store of all clients OS.

Unfortunately, command line tools (such as gleam) are generally not aware about those custom trusted certificates, and therefore the SSL connection fails when trying to obtain Hex packages:
`Unable to determine package versions: error sending request for url
(https://repo.hex.pm/packages/gleeunit): error trying to connect: invalid
peer certificate: UnknownIssuer`

The most "convenient" approach for Windows users would be to configure Gleam to be able to use the OS certificate store, so any changes on the corporate certificate configuration can immediately be used by Gleam without any configuration changes. For instance, in Git this is solved by [choosing the SChannel SSL backend mechanism](https://stackoverflow.com/questions/16668508/how-do-i-configure-git-to-trust-certificates-from-the-windows-certificate-store):
`git config --global http.sslBackend schannel`

However this solution would require the usage of a Windows specific API, and will not solve the problem for other non-Windows users.

Alternatively, an environment variable or a configuration setting can be used to specify the location of the trusted custom certificate store.

For instance, when using Elixir in a corporate environment, in order to retrieve Hex dependencies using mix you have the following two options to [specify the path to the CA certificate store PEM file](https://hexdocs.pm/hex/Mix.Tasks.Hex.Config.html):

1. Set the HEX_CACERTS_PATH environment variable to point to the custom certificate store.
2. Use "mix hex.config cacerts_path /your/certificate/store.pem"

More background about this issue can be found on the related [Github discussion page](https://github.com/gleam-lang/gleam/discussions/3152).

mdjohns:
Resolves #3246

This PR adds a `GLEAM_CACERTS_PATH` environment variable to allow users to configure a custom certificate for HTTP requests. This follows the established naming convention of `HEX_CACERTS_PATH`, which is used in [the Elixir world](https://hexdocs.pm/hex/Mix.Tasks.Hex.Config.html#module-config-keys).  The issue above does a great job with a more thorough explanation. 

I encountered this issue myself when trying to set up a new Gleam app on my work machine and thought I'd take a stab at this. 

This is my first Rust contribution, so please don't hesitate to request any changes! 
I also imagine that we'd want to document this change, so let me know the best place for that.

<details>
  <summary>Before and After</summary>

Screenshot 2024-07-31 at 12 09 48 PM

</details>

### DIFF
--- a/compiler-cli/src/http.rs
+++ b/compiler-cli/src/http.rs
@@ -6,6 +6,7 @@ use gleam_core::{Error, Result};
 use http::{Request, Response};
 
 static REQWEST_CLIENT: OnceLock<reqwest::Client> = OnceLock::new();
+static CERTS_ENV_VAR: &str = "GLEAM_CACERTS_PATH";
 
 #[derive(Debug)]
 pub struct HttpClient;
@@ -27,7 +28,7 @@ impl gleam_core::io::HttpClient for HttpClient {
             .try_into()
             .expect("Unable to convert HTTP request for use by reqwest library");
         let mut response = REQWEST_CLIENT
-            .get_or_init(reqwest::Client::new)
+            .get_or_init(init_client)
             .execute(request)
             .await
             .map_err(Error::http)?;
@@ -40,5 +41,34 @@ impl gleam_core::io::HttpClient for HttpClient {
         builder
             .body(response.bytes().await.map_err(Error::http)?.to_vec())
             .map_err(Error::http)
+    }
+}
+
+fn init_client() -> reqwest::Client {
+    if let Some(cert) = get_certificate() {
+        return reqwest::Client::builder()
+            .add_root_certificate(cert)
+            .build()
+            .expect("Unable to initialize a reqwest HTTP client");
+    } else {
+        return reqwest::Client::new();
+    }
+}
+
+fn get_certificate() -> Option<reqwest::Certificate> {
+    match std::env::var(CERTS_ENV_VAR) {
+        Ok(certs_path) => {
+            let data = std::fs::read(certs_path).expect(&format!(
+                "Unable to read certs file set as `{}`",
+                CERTS_ENV_VAR
+            ));
+            let cert = reqwest::Certificate::from_pem(&data).expect(&format!(
+                "Unable to construct a certificate from certs file set as `{}`",
+                CERTS_ENV_VAR
+            ));
+
+            Some(cert)
+        }
+        _ => None,
     }
 }


msanchezdev:
I think it'd be safe to default GLEAM_CACERTS_PATH to the variable HEX_CACERTS_PATH if not set

compiler-cli/src/http.rs
```diff
@@ -42,3 +43,32 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+            .add_root_certificate(cert)
+            .build()
+            .expect("Unable to initialize a reqwest HTTP client");
+    } else {
+        return reqwest::Client::new();
+    }
+}
+
+fn get_certificate() -> Option<reqwest::Certificate> {
+    match std::env::var(CERTS_ENV_VAR) {
```
lpil, line 59: This can be written as `let certs_path = std::env::var(CERTS_ENV_VAR)?` to get rid of the nesting

compiler-cli/src/http.rs
```diff
@@ -42,3 +43,32 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+        Ok(certs_path) => {
+            let data = std::fs::read(certs_path).expect(&format!(
+                "Unable to read certs file set as `{}`",
+                CERTS_ENV_VAR
+            ));
+            let cert = reqwest::Certificate::from_pem(&data).expect(&format!(
+                "Unable to construct a certificate from certs file set as `{}`",
+                CERTS_ENV_VAR
+            ));
```
lpil, line 68: The two `expect`s here cause the build tool to crash in the event that the user provides an invalid path or certs. We don't crash on invalid input, so instead errors will need to be returned from this function.

Our file io module has a file read function which will do this for you, and you'll need to construct our file IO error manually for the parsing of the certs.

lpil (commented):
Thank you.

It looks like these would be used for all HTTP requests rather than just for Hex. Is that what we want or would that be too restrictive if we wanted to make requests to other domains too?

lpil:

mdjohns:
Hey all, sorry for the delay on this one, I got wrapped up in some other things.
I'm happy to re-open with @lpil's changes (thanks for those btw!).

Regarding this question:
> It looks like these would be used for all HTTP requests rather than just for Hex. Is that what we want or would that be too restrictive if we wanted to make requests to other domains too?
>
> How might we test this change?

I'm not sure I have a good answer for this. My assumption is that this is the expected behavior. For example, cURL would behave the same way. But @kitoid (since you wrote the detailed original issue!) or anyone else could confirm?

I'm also happy to test it if we can think of anything in particular to test.

PR #3460, opened 2024-07-31
Help to infer unannotated `fn` args from the call context

sobolevn:
This is a little first step in better annotated `fn` inference. Right now I teached it how to infer the arg types from the call context. Something similar to `infer_lambda_type_using_context` https://github.com/python/mypy/blob/8b74b5ab03742ba86348dfe14d0468a995f74ee7/mypy/checkexpr.py#L5318-L5393 in MyPy / CPython. 

This test still fails:

```rust
#[test]
fn pipe_into_annonymous_annotated_function() {
    assert_module_infer!(
        r#"
pub fn main() {
  let a = 1
     |> fn (x) { #(x, x + 1) }
     |> fn (x) { x.0 }
     |> fn (x) { x }
}
"#,
        vec![("main", "Int")]
    );
}
```

```
---- type_::tests::pipes::pipe_into_annonymous_annotated_function stdout ----
thread 'type_::tests::pipes::pipe_into_annonymous_annotated_function' panicked at compiler-core/src/type_/tests.rs:324:6:
should successfully infer: [NotATupleUnbound { location: SrcSpan { start: 77, end: 78 } }]
```

But, let's do it incrementally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,16 @@
   was imported then the alias would be shown.
   ([Surya Rose](https://github.com/gearsdatapacks))
 
+- Improves how inference works for anonymous functions followed by call
+  arguments. For example:
+
+  ```gleam
+  fn(x) { x.0 }(#(1, 2))
+  ```
+
+  would be infered as `fn() -> Int` in this context.
+  ([sobolevn](https://github.com/sobolevn))
+
 ### Formatter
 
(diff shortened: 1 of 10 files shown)


lpil (approved):
This is super nice! Thank you! Would you mind updating the changelog please? 🙏

lpil (approved):
Thank you

PR #3468, opened 2024-08-01
refactor: Use `ast` type aliases where possible

sobolevn:
It caught my eye that there are places where established type aliases are duplicated.


lpil (approved):
Thank you!

Issue #3216, opened 2024-05-30
Add `let assert ... as` syntax


isaacharrisholt:
Originally mentioned in #2958.
Support some sort of `as` syntax when using `let assert` to add additional context as for why we want to panic.

For example, if querying for a DOM element that we need to be present:

```gleam
import plinth/document

fn main() {
  let assert Ok(el) = document.query_selector("#my-el")
  // ...
}
```

It would be nice to be able to provide some context with the crash about why we need that specific element to be there. The current alternative is:

```gleam
let el = case document.query_selector("#my-el") {
  Ok(el) -> el
  Error(_) -> panic as "Element needs to be present"
}
```

However, this is quite verbose and quite common, which is the reason for the `let assert` syntax in the first place.

There are a few options for syntax that I can come up with, in order of personal preference:

## 1. `as` at end

```gleam
let assert Ok(_) = my_result as "Element needs to be present"
```

I personally prefer to have the `as` clause at the end of the line, similar to `panic` and `todo`. The message could be varying length, and keeping it towards the end makes the line easier to read. However, to newer Gleam users, this may seem more like the `my_result as "X"` is one item, which could lead to confusion.

## 2. `as` after variable

```gleam
let assert Ok(_) as "Element needs to be present" = my_result
```

This option solves the potential confusion above, but it could also push the value we're asserting against quite far from the values we're trying to assign, e.g.:

```gleam
let assert Ok(_) as "Some really loooooooooooooooooooooooooooong message" = my_result
```

## 3. `as` after `assert`

```gleam
let assert as "Element needs to be present" Ok(_) = my_result
```

Again, this solves the confusion issue in 1, and keeps the values together, but having a string next to the `Ok(_)` just feels icky. Can't explain why. Sorry.

---

If you have any alternative syntax suggestions, let me know and I'll chuck 'em above :)

giacomocavalieri:
I don't think 1 is what we should go for because it could lead to some confusion if you write this:
```gleam
let assert Ok(_) = todo as "error message"
```
Here you can't really tell if `as` is part of the first assertion or the `todo`.

I think 3 would be the best choice if in the future we also want some kind of `assert` keyword to support it as well in a similar way:
```gleam
// 3.
assert as "wibble" [1, 2, 3] = list.reverse([3, 2, 1])
assert as "wibble" is_empty([])
let assert as "wibble" Ok(result) = maybe_fail()
```

[...]

PR #3469, opened 2024-08-01
Implement `let assert as "msg"` syntax

sobolevn:
This is a draft of https://github.com/gleam-lang/gleam/issues/3216

## Design

- I went with option `3.` from https://github.com/gleam-lang/gleam/issues/3216#issue-2326145589 since option `2.` means something different and option `1.` was very confusing, `let assert as msg` is similar to `todo as msg` and `panic as msg`
- I remove `AssignmentKind` enum and replaced it with `AssertAssignment` struct which contains both location and optional message of `assert`
- In `let assert as MSG` our `MSG` can only be a unit expression: basically a name or `String`
- I added a inference of `MSG` to only be `String` and nothing else

## Further steps

I understand that this feature was not fully discussed and not fully specified.
So, please: feel free to close this PR if you don't want this.

I did it just for fun :)

## Impressions

I really had fun working on this! 
The compiler is very simple and is pleasant to work with. Tests are really intuitive with the help of `insta`.

Thanks a lot for writting Gleam, it is so refreshing and cool!

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1942,27 +1942,6 @@ impl<A> HasLocation for Pattern<A> {
     }
 }
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
-pub enum AssignmentKind {
-    // let x = ...
-    Let,
-    // let assert x = ...
-    Assert { location: SrcSpan },
-}
-
-impl AssignmentKind {
-    /// Returns `true` if the assignment kind is [`Assert`].
-    ///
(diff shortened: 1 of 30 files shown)


lpil:
The design hasn't been settled on in the issue.

GearsDatapacks:
Oh yeah I see, thanks. I'll look at the issue for this then

lpil:
Please feel to reopen if you wish to continue with this PR!

sobolevn:
From what I remember this was completed, other than the design discussion 😅 
Feel free to take over, I am very busy with the work stuff right now.

Issue #828, opened 2020-11-03
Render imported types with module qualifiers in documentation


michallepicki:
Given a module that imports code from another module, and public types (or functions) using that type, e.g.
```
import gleam/otp/node.{Node}
import gleam/otp/actor.{StartError}

pub type ApplicationStartMode {
  Normal
  Takeover(Node)
  Failover(Node)
}

pub fn worker(
  start: fn(argument) -> Result(Sender(msg), StartError),
) -> ChildSpec(msg, argument, argument) {
// ...
```

Currently types in documentation are rendered as present in code. For someone reading the module documentation without broader knowledge of the library, it is hard to see where a type comes from without reaching into module source code. For the library author, it is an inconvenience to always specify a module even if code compiles.

The `gleam docs` command could automatically add module qualifiers when rendering documentation. (Some time in the future, it could also render a clickable link)

lpil:
Great suggestion! Thank you.

lpil:
One question I have, how do we show namespaced modules in type sigs?

```rust
import one/two/three.{Type}

pub type MyType = Type
```

Here the type is `one/two/three.Type` and we'd want that qualified in some fashion, but  `one/two/three.Type` isn't valid Gleam syntax.

Perhaps that's OK but it would be good to see some mock-up screenshots (can just edit the HTML in browser) before we commit to an implementation.

michallepicki:
What about something like this? Last part of the imported module in the type, full module name on mouse hover (span `title` attribute)
![image](https://user-images.githubusercontent.com/4972756/98029346-a5cdc200-1e0f-11eb-9ea8-9ff361dd35e9.png)

lpil:
That's a great idea! Let's do that

JohnDoneth:
I've been working on rendering clickable links to types in the documentation as mentioned here but have hit a couple blockers.

- We don't currently have the library name in the type information. 
- When it comes to the stdlib types, they don't contain the module paths.. 

If `Float` had the module path `["gleam", "float"]` with the library name `"gleam_stdlib"`. Then we'd have all the information necessary to generate links. Assuming everything is on Hex we could generate: `https://hexdocs.pm/gleam_stdlib/gleam/float/#Float`

Would it be fine to hard-code those module paths in the static type information and add an additional field to specify the library name to types?

michallepicki:
Louis will decide but here are my 2 cents:
[...]

PR #3471, opened 2024-08-01
WIP: Links in type signatures

### LINKED ISSUE #3461: (hex)docs: types in function signatures should be clickable links to their documentation
When looking at the documentation of a function on hexdocs such as this one

screenshot

... I want to click on types (`Request`, `Connection`, `FileError`) and see their documentation. As it is now, I always have to select, copy and paste them into the search bar. Even more critically, sometimes the types are not defined in the same package (as is the case for `FileError`) and then you have to go to the source code to find out more (it turns out `FileError` is from `simplifile`!).

joshi-monster:
I feel like this has already gotten quite big, so I thought I'd stop at the early point before you say it's all unreasonable anyways and I waste a bunch of time :D

First: I cannot "just" generate links, since [highlight.js will not allow them.](https://github.com/highlightjs/highlight.js/wiki/security).

Instead, all code relevant to type/function/constant signatures is highlighted at compile time, and types are always printed using the `type_::pretty::Printer` in the docs. It introduces a new variant to the Document, and a new DocumentWriter trait that can react on them. Since it looked like the documents formatter goes to great length to avoid being recursive, I've also  tried to do that. This introduces another new `Cmd` type that gets used instead of the list of `(indent, mode, doc)` triples, and consequently changes a lot of types around in format. Just doing recursion would basically get rid of all of that.

Overall, this allows for all signatures (with small extensions all parseable Gleam code) to be highlighted by the compiler directly, without relying on Javascript. This could potentially also be used to add colors/links to error messages, or annotate all examples in the docs as well.

If a resolved type (ignoring aliases) is publicly exposed, a link to its documentation will be generated as well. Versions are currently ignored. Ignoring aliases means that a common pattern doesn't work currently (lustre for example does this with every type), where an internal type is re-exported using a type alias.

### Questions

- What about tests? There are existing tests for the docs output which I updated, but generally, I'm a bit lost what you would expect.
- What to do about the alias thing?

fix #3461 #828

### DIFF
--- a/compiler-cli/src/docs.rs
+++ b/compiler-cli/src/docs.rs
@@ -11,6 +11,7 @@ use gleam_core::{
     error::Error,
     hex,
     io::HttpClient as _,
+    manifest::Manifest,
     Result,
 };
 
@@ -68,6 +69,7 @@ pub fn build(options: BuildOptions) -> Result<()> {
     crate::fs::delete_directory(&paths.build_directory_for_target(Mode::Prod, config.target))?;
 
     let out = paths.build_documentation_directory(&config.name);
+    let manifest = crate::build::download_dependencies()?;
     let mut built = crate::build::main(
         Options {
             mode: Mode::Prod,
@@ -76,9 +78,14 @@ pub fn build(options: BuildOptions) -> Result<()> {
             warnings_as_errors: false,
             root_target_support: TargetSupport::Enforced,
         },
-        crate::build::download_dependencies()?,
+        manifest.clone(),
     )?;
-    let outputs = build_documentation(&config, &mut built.root_package, DocContext::Build)?;
+    let outputs = build_documentation(
+        &config,
+        &mut built.root_package,
+        &manifest,
+        DocContext::Build,
+    )?;
 
     // Write
     crate::fs::delete_directory(&out)?;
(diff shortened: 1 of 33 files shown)


lpil (commented):
Thank you!

We don't want to modify the formatter at all for this. As the comment on the documentation generation code says we want to stop using the same code for both, so this would be the point where we split it so we can have two specialised bits of code rather than 1 lowest-common-denominator.

The pretty module must not know anything about Gleam or HTML or code, so the annotation and to-html method must be removed. The thing to add is a ZeroWidthText which would work the same as text but have contents which do not contribute to the size. This would be a very small change to the code.

Loops are used over recursion as Rust doesn't have TCO so the printer would (and used to!) crash if recursion is used.

The tests as you have them look great!

joshi-monster:
Thank you so much for taking a look at this! I kind of put this on hold, since I wanted to wait what your thoughts in the discussion thread were first :slightly_smiling_face: 

I had a version with a `ZeroWidthString` constructor in the first, but decided against it since it would have added very HTML/doc-specific things to the formatter, and I didn't see the comment, sorry. I ended up with those annotations since they were kind of general (following the highlightjs classes) and also allowed me to experiment with syntax highlighting in the CLI output :smile:

A special formatter for printing docs and a `ZeroWithString` would also solve this, and be indeed a much simpler change to the  the pretty module, I'll do that then instead!

I understand why the loop  is done this way; I just felt like for sure noone is gonna stack-overflow with the depth of the  tree :smile:

lpil:

PR #3472, opened 2024-08-02
Do not show completions when writing inside a comment

giacomocavalieri:
This PR fixes a bug in the release candidate where the LS would show completions when writing inside a comment.


lpil (approved):
Thank you

Issue #3455, opened 2024-07-30
LSP: Add local variable suggestions


Bkeinn:
Add lsp support for suggesting local variables

Expect: 
![Screenshot from 2024-07-29 21-36-43](https://github.com/user-attachments/assets/ff18cc48-6827-4c9d-87c7-170f86ba54b8)
it to suggest the add_one variable

lpil:
Thank you

GearsDatapacks:
This can be closed now that #3480 has been merged

PR #3480, opened 2024-08-02
LSP: add function argument and local variables suggestions

ezegros:
This draft PR addresses issue #3455 by adding suggestions for both arguments and variables declared within the same function as the cursor's position.

## Description
- Module Filtering: Iterate over the module definitions, filtering out those where the cursor is not within a function.
- Argument Suggestions: Gather suggestions for arguments by iterating through the function's arguments list.
- Variable Suggestions: Traverse the function body to collect suggestions for variables assigned before the current cursor position.

This approach was chosen to minimize extensive refactoring and to avoid modifying existing functionality. However, there may be more efficient methods or existing functions that could simplify this implementation. Feedback and suggestions for improvement are highly appreciated.


Acepie:
FYI a previous attempt using the same approach was done https://github.com/gleam-lang/gleam/pull/2755 as far as I can tell the implementation you have would not work for when the cursor is in an anonymous function

ezegros:
> FYI a previous attempt using the same approach was made #2755 as far as I can tell the implementation you have would not work when the cursor is in an anonymous function

Yes, you are right It doesn't, I'll keep working on it. Also thanks for showing me that PR I didn't see it before and although the code seems quite old It is probably useful (sad it was abandoned).

ezegros:
Hey @Acepie! just added a fix so it works with anonymous functions too, let me know what you think

Acepie:
From a process perspective, you would need Louis to review/approve this. If you look at the conversation in the previous implementation you'll notice a discussion about the general approach that was taken. Granted that was months ago so idk if his thoughts have changed on this at all.

From a pure implementation perspective, one thing I want to emphasize is that a general solution for this that traverses the AST would need to be recursive. Imagine a case where the anonymous function was more complicated and had another anonymous function inside of it. Theoretically, nothing stops somebody from making N nested anonymous functions. Additionally, there are some other edge cases you would need to consider that would cause an issue here (what if the anonymous function is returned directly as an expression instead of assigned, what if the pattern is complex/has more than 1 assignment)

ezegros:
Well, a lot of things were missing so thanks for the insights. I completely refactored and followed what was done in the mentioned PR which made a lot of sense. I imported some tests and added a few more

lpil:
Hello! Is this ready for review? It's marked as a draft :)

compiler-core/src/language_server/completer.rs
```diff
@@ -535,6 +536,38 @@ where
...
+                    mod_name,
+                    insert_range,
+                    cursor,
+                    &fun.body,
+                ));
+            }
+
+            for definition in &self.module.ast.definitions {
+                // Stop if we've passed the cursor position
+                if definition.location().start > cursor {
```
lpil, line 561: Are definitions guaranteed to be ordered by location at this point? I think they are ordered by dependency here.

compiler-core/src/language_server/completer.rs
```diff
@@ -535,6 +536,38 @@ where
...
+                    break;
+                }
+
+                // Only proceed if the definition contains the cursor
+                if definition.find_node(cursor).is_none() {
+                    continue;
```
lpil, line 567: There's no side effects in this loop? I think it does nothing?

compiler-core/src/language_server/completer.rs
```diff
@@ -881,3 +965,183 @@ fn field_completion(label: &str, type_: Arc<Type>) -> CompletionItem {
...
+                extract_from_expression(mod_name, insert_range, cursor, expr, &mut completions)
+            }
+
+            Statement::Use(_) => (),
+        });
+
+    completions
+}
+
+fn extract_from_expression(
```
lpil, line 1005: We have a visitor class for traversing AST like this, best use that

compiler-core/src/language_server/tests/completion.rs
```diff
@@ -712,6 +712,184 @@ pub fn wibble(
...
+type Foo { Bar(str: String) }
+fn wibble() {
+  let _a = 10
+  let a = 10
+  let b as c = 5.
+  let assert #([d, e] as f, g) = #([0, 1], 2)
+  let assert <<h:1>> as i = <<1:1>>
+  let assert "a" <> j = "ab"
+  let Bar(k) = Bar("baz")
+
```
lpil, line 813: Only test one thing per test please, split these into multiple

lpil (commented):
Thank you for this. I've left some notes and questions inline.

Gosh these snapshots are hard to read. I'll have to improve how they are printed

ezegros:
I've addressed your comments, snapshots improvement is great!

compiler-core/src/language_server/completer.rs
```diff
@@ -916,3 +967,120 @@ fn field_completion(label: &str, type_: Arc<Type>) -> CompletionItem {
...
+            cursor,
+            completions: Vec::new(),
+        }
+    }
+
+    pub fn completions(&self) -> Vec<CompletionItem> {
+        self.completions.clone()
+    }
+
+    pub fn visit_statements(&mut self, statements: &'a [Statement<Arc<Type>, TypedExpr>]) {
```
lpil, line 992: Could you use detailed comments to document why this function does this please 🙏  I didn't understand it at first.

compiler-core/src/language_server/completer.rs
```diff
@@ -916,3 +967,120 @@ fn field_completion(label: &str, type_: Arc<Type>) -> CompletionItem {
...
+            self.mod_name,
+            name,
+            type_,
+            self.insert_range,
+        ));
+    }
+}
+
+impl<'ast> ast::visit::Visit<'ast> for LocalCompletion<'_> {
+    fn visit_typed_assignment(&mut self, assignment: &'ast ast::TypedAssignment) {
```
lpil, line 1024: Could you use detailed comments to document why this function does this please 🙏  I didn't understand it at first.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__local_variable_all_assignment_patterns.snap
```diff
@@ -0,0 +1,734 @@
+---
```
lpil, line 1: I think this is outdated or unused?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__local_variable_bit_string.snap
```diff
@@ -0,0 +1,34 @@
...
+expression: "\nfn wibble() {\n  let assert <<h:1>> as i = <<1:1>>\n\n}\n"
+---
+fn wibble() {
+  let assert <<h:1>> as i = <<1:1>>
+|
+}
+
+
+----- Completion content -----
+h
```
lpil, line 12: I think this is out of date or unused

compiler-core/src/language_server/completer.rs
```diff
@@ -916,3 +967,120 @@ fn field_completion(label: &str, type_: Arc<Type>) -> CompletionItem {
...
+
+    fn visit_typed_expr_fn(
+        &mut self,
+        _: &'ast ast::SrcSpan,
+        _: &'ast Arc<Type>,
+        _: &'ast bool,
+        args: &'ast [ast::TypedArg],
+        body: &'ast [ast::TypedStatement],
+        _: &'ast Option<ast::TypeAst>,
+    ) {
```
lpil, line 1040: How come this method has been defined? Isn't this what the base implementation does?

compiler-core/src/language_server/completer.rs
```diff
@@ -570,6 +571,24 @@ where
...
         if module_select.is_none() {
+            let cursor = self
+                .src_line_numbers
+                .byte_index(self.cursor_position.line, self.cursor_position.character);
+
+            // Function arguments and variables
+            // Find the function that the cursor is in
+            // and push the arguments and local variables
+            if let Some(fun) = self.module.ast.definitions.iter().find_map(|d| match d {
+                Definition::Function(f) if f.full_location().contains(cursor) => Some(f),
```
lpil, line 582: Will this suggest completions for the values within the header of the function? Could we have a test to verify it does not please 🙏 

```gleam
pub fn main(x, y, | z) {
  Nil
}

lpil (commented):
Wonderful! Really nice work. Some very small notes inline

lpil:
Oh! And could you update the changelog too please?

I'm excited for this one!! 💜

compiler-core/src/language_server/completer.rs
ezegros: The base implementation does not iterate over the function's arguments, so using the base will lead to missing suggestions.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__local_variable_bit_string.snap
ezegros: I'm afraid I do not understand, what exactly is out of date or unused?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__local_variable_all_assignment_patterns.snap
ezegros: Removed!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__completion__local_variable_bit_string.snap
ezegros: Removed!

compiler-core/src/language_server/completer.rs
ezegros: Added documentation! Let me know what you think if it is clear or not

compiler-core/src/language_server/completer.rs
ezegros: I refactored a little bit to make it clearer. Now it receives the Function instead of the body statements so you don't have to check the usage to know what it does! This also means that suggestions for the function argument are added here. Of course, with all these changes I changed the name to clarify and added some comments to document it. Let me know what you think!

compiler-core/src/language_server/completer.rs
lpil: Wonderful!

compiler-core/src/language_server/completer.rs
lpil: Thank you

lpil (approved):
Beautiful work!! Thank you!!!

PR #3481, opened 2024-08-02
Adds initial plumbing for computing external usages of a module

Acepie:
ModuleInterfaces will now include all the locations of usages of external values and types. Since gleam does not allow cyclical module imports, this would allow us to find all references of another module used in the module without having to recompile the module or do a post processing step on all the modules as a whole later. Using these values for doing "find all references" in the lsp would come next but didn't want this pr to get too long

### DIFF
--- a/compiler-core/generated/schema_capnp.rs
+++ b/compiler-core/generated/schema_capnp.rs
@@ -478,7 +478,7 @@ pub mod module {
       !self.reader.get_pointer_field(6).is_null()
     }
     #[inline]
-    pub fn get_external_value_usages(self) -> ::capnp::Result<::capnp::struct_list::Reader<'a,crate::schema_capnp::property::Owned<::capnp::struct_list::Owned<crate::schema_capnp::property::Owned<::capnp::struct_list::Owned<crate::schema_capnp::src_span::Owned>>>>>> {
+    pub fn get_external_value_usages(self) -> ::capnp::Result<::capnp::struct_list::Reader<'a,crate::schema_capnp::property::Owned<crate::schema_capnp::external_usages::Owned>>> {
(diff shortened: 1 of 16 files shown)


chikoYEAT (approved):
1,192 lines added and 64 lines removed, everything here seems correct to me. No unnecessary code added

lpil:
Sorry, could you explain the approach taken here? I see if is using the usage system which is due to be deleted soon.

Acepie:
The core premise is that if we want to keep module compilation individual then we need to know what references/usages to other modules are present and store those in the module interface so that we can make use of that information later when showing references/allowing for refactors. 

This change only relies on the previous usage counting mechanisms for knowing if a variable is an unqualified import so we can include those. We could remove that logic and deal with unqualified imports whenever a new usage system is worked on though the core idea of seeing that a variable is referencing an unqualified imported value/type would be the same

lpil:
I'm not a huge fan of how it's using two things we want to delete (the usage system and capnp). I think it should touch neither of these systems ideally.

How would one find all the references for a function? At first to me it seems that you'd need to traverse all the interfaces to do this with this design, can that be avoided?

Acepie:
I don't see how we'd ever be able to not use capnp since in order to do anything useful across multiple modules we need enough information in the ModuleInterface which means updating capnp or whatever new mechanism we replace capnp with (otherwise we'd have to do tree walking on each call which we've already said we don't want to do). Yeah we would need to look at every ModuleInterface to see if it ever references the module the located node is defined in. I don't see how we could possibly avoid doing that since there is no way we could attach reference information at module compilation time.

I'll just close this if you think this approach is entirely a nonstarter

PR #3484, opened 2024-08-03
Store formatted variables in a String

PhilTaken:
This PR fixes a compilation error when compiling with slightly less recent rust compilers, e.g. 1.78.0 and 1.77.1.

The performance overhead of formatting the string here is negligible in any case since the string is only discarded if there are zero or one variables, in which case the extra call to `join` is very cheap.

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -344,10 +344,11 @@ impl<'ast> ast::visit::Visit<'ast> for LetAssertToCase<'_> {
         self.visit_typed_pattern(&assignment.pattern);
         let variables = std::mem::take(&mut self.pattern_variables);
 
+        let formatted_all = format!("#({})", variables.join(", "));
         let assigned = match variables.len() {
             0 => "_",
             1 => variables.first().expect("Variables is length one"),
-            _ => &format!("#({})", variables.join(", ")),
+            _ => &formatted_all,
         };
 
         let edit = TextEdit {


compiler-core/src/language_server/code_action.rs
```diff
@@ -344,10 +344,11 @@ impl<'ast> ast::visit::Visit<'ast> for LetAssertToCase<'_> {
         self.visit_typed_pattern(&assignment.pattern);
         let variables = std::mem::take(&mut self.pattern_variables);
 
+        let formatted_all = format!("#({})", variables.join(", "));
         let assigned = match variables.len() {
             0 => "_",
             1 => variables.first().expect("Variables is length one"),
```
Deepfriedice, line 350: Is this line doing anything after this change?

sobolevn (approved):
I also faced this problem, compiler update helped. I guess it is fine to include.

lpil:
Hello! What's the motivation for this? We don't support other Rust versions and we provide precompiled binaries for people to use.

PhilTaken:
Personally, I am not a fan of precompiled tools especially when the resulting binaries are not fully reproducible but that is a matter of taste.

This is the only change that would enable retaining compatibility with a rustc version that was just released this May. I'd totally understand if the final decision is not to merge and just to raise the MSRV.

lpil:
I'm guessing this is due to Nix? That build tool produces more support requests and takes up more time than all other build tools and operating systems combined. It's clearly documented what we support, and we provide precompiled binaries that work on NixOS without modification to the operating system or the binary, so I don't understand why there are frequent requests to change things in Gleam to facilitate Nix.

PR #3486, opened 2024-08-03
Fix issue where record access completions wouldn't work for private records in the same module

### LINKED ISSUE #3485: Language Server: Label access completion only works with public types
The language server doesn't show any suggestion for fields when a record's type isn't public:

image

If we set `Person` as public, suggestions show up.

image

Acepie:
Closes #3485

Looks like we were previously removing the private accessors from the environment prior to creating the ModuleInterface. This meant that even if we were in the module the type was defined in, looking up the accessor would return no result. Seeing as the accessor includes the publicity anyway, I just removed the filtering logic and did it in the lsp/inference itself instead

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,3 +11,7 @@
 ### Language Server
 
 ### Bug Fixes
+
+- Fix a bug that caused record accessors for private types to not be completed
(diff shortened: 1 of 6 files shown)


lpil (approved):
Thank you

PR #3487, opened 2024-08-04
Report non-byte aligned bit array patterns as compile errors for JS target

richard-viney:
This PR contains three small changes:

- Non-byte aligned integers in bit array patterns are now compile time errors on JS
- A couple of existing errors related to unsupported bit array features on JS are now more descriptive
- Small optimisation to `sizedInt()` in the JS prelude that avoids a `.reverse()` call

Can split these out to separate PRs if that's preferred

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -316,6 +316,11 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- On the JavaScript target, non-byte aligned integers in bit array patterns are
+  now reported as a compile-time error.
+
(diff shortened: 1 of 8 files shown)


lpil (commented):
Very nice! Could you add some tests please 🙏

richard-viney:
Tests have been added, should be good to go. Pre-existing tests cover the changes made in `prelude.mjs`.

richard-viney:
I believe the CI failure is caused by minor linter incompatibilities with Rust 1.81 that was released a couple of days ago, not by the changes in this PR.

lpil (approved):
Thank you!!

PR #3494, opened 2024-08-07
Improving readability in build.rs

chikoYEAT:
instead of using `to_string`, i used `String::from` for initial module assignment for better readability.
..


lpil:
Hello! The formatter is failing here

chikoYEAT:
ok let me check, this should work ig

lpil (approved):
Thanks

PR #3505, opened 2024-08-11
fix REBAR_BARE_COMPILER_OUTPUT_DIR path

### LINKED ISSUE #3067: fail to compile ezstd
platfrom: max m1
gleam: v 1.1.0, installed with brew

`gleam add ezstd & gleam run`

```
LINK obj/conf_713d974d1b5639d588bafad52bd377d1/zstd
zstd build completed
~/WebstormProjects/gleam/e/build/dev/erlang/ezstd/_build/deps ~/WebstormProjects/gleam/e/build/dev/erlang/ezstd
~/WebstormProjects/gleam/e/build/dev/erlang/ezstd
 CPP    ezstd_nif.cc
 CPP    nif_utils.cc
 LD     ezstd_nif.so
===> Analyzing applications...
===> Compiling ezstd
===> Missing artifact priv/ezstd_nif.so
error: Shell command failure
```
if I clone repo: https://github.com/silviucpp/ezstd and `rebar3 compile`
same working if try rebar3 lib with `rebar3 new lib myapp` & `rebar3 compile` after adding ezstd as dep

try same with mix also looks like working

gusinacio:
This solves an issue where it could not compile https://github.com/silviucpp/ezstd.

It was using relative paths instead of absolute paths.

Fixes #3067


lpil (commented):
Thank you. Could you update one of the test projects in `test` to replicate the problem and verify the fix, and also update the changelog. Thank you

test/project_erlang/gleam.toml
```diff
@@ -23,6 +23,7 @@ ssl_verify_fun = "~> 1.1"
 # TODO: replace this with a package with a nif that compiles super fast. Perhaps
 # just a hello world.
 bcrypt = "~> 1.1"
+ezstd = "~> 1.1"
```
lpil, line 26: Document what this dep is here for please 🙏

lpil (commented):
I wasn't able to rebase this PR as it was using the "main" branch and my git-fu is too weak 😁 

Could you rebase the branch please

lpil (approved):
Thank you!!

lpil:
Looks like CI is failing due to that package not supporting windows. I think you're going to have to test in a different project and only run it on non-windows CI.

gusinacio:
Hello @lpil, I updated the CI step that was causing problems so now we have two different steps: one for windows and other for non-windows.

lpil (approved):
Thank you!!

Issue #3489, opened 2024-08-04
lsp crash


harcma:
To reproduce 
Create a new project with `gleam new name` 
in vscode
Open src/name.gleam
Paste in my code
> popup apears:
> The Gleam Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information. 

<details>  <summary>Crash logs from output</summary>

 INFO language_server_starting
Hello human!

This command is intended to be run by language server clients such
as a text editor rather than being run directly in the console.

Many editors will automatically start the language server for you
when you open a Gleam project. If yours does not you may need to
look up how to configure your editor to use a language server.

You can exit this program by pressing ctrl+c.

 INFO creating_new_language_server_engine path="/home/harcma/Ram/g1eam_lsp"
 INFO compile{package=gleam_stdlib}: analysing_modules count=0
 INFO compile{package=gleeunit}: analysing_modules count=0
 INFO compile{package=g1eam_lsp}: analysing_modules count=2
error: Fatal compiler bug!

This is a bug in the Gleam compiler, sorry!

Please report this crash to https://github.com/gleam-lang/gleam/issues/new
and include this error message with your report.

Panic: compiler-core/src/exhaustiveness.rs:746
	Custom type variants must exist: Type { name: "Token", hint: AlternativeTypes(["String", "UtfCodepoint", "Result", "Float", "Int", "BitArray", "CalcError", "Token", "Bool", "Nil", "List"]) }
Gleam version: 1.4.0
Operating system: linux

If you can also share your code and say what file you were editing or any
steps to reproduce the crash that would be a great help.

You may also want to try again with the `GLEAM_LOG=trace` environment
variable set.

[Error - 3:30:41 PM] The Gleam Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.

</details>
<details><summary>My code</summary>
import gleam/int.{to_string}
import gleam/io
import gleam/list
import gleam/string

const al = [
  "A", "B", "C", "D", "E", "F", "G", "H", "I", "J", "K", "L", "M", "N", "O", "P",
  "Q", "R", "S", "T", "U", "V", "W", "X", "Y", "Z", "a", "b", "c", "d", "e", "f",
  "g", "h", "i", "j", "k", "l", "m", "n", "o", "p", "q", "r", "s", "t", "u", "v",
  "w", "x", "y", "z",
]

[...]

PR #3508, opened 2024-08-13
Fix another compiler crash for duplicate constructor definition

### LINKED ISSUE #3506: Compiler crash when checking exhaustiveness of a pattern on a type with duplicate variants
Hello, thank you for all the wonderful work on Gleam! On v1.4.1 and the current main branch, but not on v1.2.1, running on Ubuntu and compiling to BEAM, I noticed the following small program will crash the compiler:

```gleam
pub type Bug {
  Ant
  Bee
  Bee
  Caterpillar
}

pub fn get_bugs(bug: Bug) {
  case bug {
    Ant -> "And"
    Bee -> "Bee"
    Caterpillar -> "Caterpillar"
  }
}
```
The error msg given by the compiler on v1.4.1 is:
```
Panic: compiler-core/src/exhaustiveness.rs:746
        Custom type variants must exist: Type { name: "Bug", hint: AlternativeTypes(["Bool", "Int", "List", "UtfCodepoint", "Result", "BitArray", "Nil", "String", "Bug", "Float"]) }
Gleam version: 1.4.1
Operating system: linux
```
The error msg given by the compiler on the current main branch is:
```
Panic: compiler-core/src/exhaustiveness.rs:539
        Case must exist
Gleam version: 1.4.1
Operating system: linux
```

On v1.2.1 and v1.4.1:
This code will compile and throw an error about the duplicate Caterpillar definition:
```gleam
pub type Bug {
  Ant
  Bee
  Caterpillar
  Caterpillar
}

pub fn get_bugs(bug: Bug) {
  "Bug!"
}
```
On v1.2.1 and v1.4.1 (but not the current main branch) this code will crash the compiler:
```gleam
pub type Bug {
  Ant
  Bee
  Caterpillar
  Caterpillar
}

pub fn get_bugs(bug: Bug) {
  case bug {
    Ant -> "And"
    Bee -> "Bee"
    Caterpillar -> "Caterpillar"
  }
}
```
on v1.4.1 it will give the following error:
```
Panic: compiler-core/src/exhaustiveness.rs:746
        Custom type variants must exist: Type { name: "Bug", hint: AlternativeTypes(["Bool", "Int", "List", "UtfCodepoint", "Result", "BitArray", "Nil", "String", "Bug", "Float"]) }
Gleam version: 1.4.1
Operating system: linux
```

So it seems that when the exhaustiveness check is run for the case statement in get_bugs, it will crash when there is _some_ set of duplicate variants. It seems like #3490 fixed a subset of the issue, but there is still a case where this pattern crashes the compiler.

Hope this helps! Let me know if any more info is needed.

GearsDatapacks:
Fixes #3506 

This is a similar issue to #3489; this crash was caused by a mismatch of `constructor_index`s and actual registered constructors. If a duplicate constructor existed, it would not be registered but the index was still incremented, which broke an assumption made in another place in the compiler. I've just changed it to manually increment an index instead of using `Iterator::enumerate`.

I didn't add another changelog entry since this is just another edge-case of the aforementioned issue, and doesn't really require a separate entry in my opinion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,11 @@
 
 - Fixed a bug where incorrect syntax error message were shown,
   when using `:` or `=` in wrong possitions in expressions.
+  ([Ankit Goel](https://github.com/crazymerlyn))
+
+- Fixed a bug where the compiler would crash when pattern matching on a type
+  which had constructors of duplicate names.
+  ([Surya Rose](https://github.com/gearsdatapacks))
 
 ## v1.4.1 - 2024-08-04
 
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -866,7 +866,8 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
 
         let mut constructors_data = vec![];
 
-        for (index, constructor) in constructors.iter().enumerate() {
+        let mut index = 0;
+        for constructor in constructors.iter() {
             if let Err(error) = assert_unique_name(
                 &mut self.value_names,
                 &constructor.name,
@@ -926,6 +927,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 module: self.module_name.clone(),
                 constructor_index: index as u16,
             };
+            index += 1;
 
             // If the contructor belongs to an opaque type then it's going to be
             // considered as private.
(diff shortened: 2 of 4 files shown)


lpil (approved):
Thank you! Could you update the changelog please 🙏

PR #3511, opened 2024-08-14
Allow to omit `utf8` option for bit array string literals

### LINKED ISSUE #3495: Allow Strings in a BitArray literal without having to specify utf8 encoding
I think it would be a neat little ux improvement to allow strings in a bit array without specifying the encoding, defaulting to utf8.

```gleam
// instead of
<<"a":utf8, 1, 2, 3>>
// just write
<<"a", 1, 2, 3>>
```

giacomocavalieri:
This PR closes #3495 

I haven't made any changes to the formatter but it could remove any `utf8` redundant option from string literals to enforce a single style.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,21 @@
 
 ### Compiler
 
+- It is now possible to omit the `:utf8` option for literal strings used in a
+  `BitArray` segment.
+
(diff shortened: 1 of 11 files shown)


lpil (approved):
Thank you!!

PR #3513, opened 2024-08-14
Do not compile root when module comes from dependencies

### LINKED ISSUE #2322: Permit `gleam run -m mod_from_some_dep` even when the root package does not compile
Currently the full package must compile for `gleam run -m` to compile. Make it so it is possible to run modules from dependencies even when the root package does not compile.

This can be done by skipping compilation of the root package when running a command from outside the root package.

giacomocavalieri:
This PR closes #2322

This is still a WIP because I need to figure out a way to test this that doesn't require me installing the compiler locally 😁


compiler-cli/src/run.rs
```diff
@@ -74,8 +74,17 @@ pub fn command(
 
     let target = target.unwrap_or(mod_config.target);
 
+    println!("{:#?}", package_kind);
```
lpil, line 77: Debug left in here

compiler-core/src/build.rs
```diff
@@ -86,6 +86,12 @@ impl Target {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Compile {
+    All,
+    DepsOnly,
+}
```
lpil, line 93: Maybe a comment to explain why we have this?

compiler-core/src/build/package_compiler.rs
```diff
@@ -75,6 +82,7 @@ where
             target,
             write_metadata: true,
             perform_codegen: true,
+            skip_analysis: false,
```
lpil, line 85: This approach means that it still loads caches and emits warnings. How about instead we don't load any modules at all? That's what I did when I implemented this. I wish I knew what happened to my version of this 😅

test/root_package_not_compiled_when_running_dep/.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
+name: test
```
lpil, line 1: Remove this please

test/root_package_not_compiled_when_running_dep/README.md
```diff
@@ -0,0 +1,24 @@
+# root_package_not_compiled_when_running_dep
```
lpil, line 1: Document what this project is for please

test/root_package_not_compiled_when_running_dep/test/root_package_not_compiled_when_running_dep_test.gleam
```diff
@@ -0,0 +1,12 @@
...
+
+pub fn main() {
+  gleeunit.main()
+}
+
+// gleeunit test functions end in `_test`
+pub fn hello_world_test() {
+  1
+  |> should.equal(1)
+}
```
lpil, line 12: Remove this please

lpil (commented):
Thank you!

compiler-core/src/build/package_compiler.rs
```diff
@@ -42,6 +42,13 @@ pub struct PackageCompiler<'a, IO> {
     pub ids: UniqueIdGenerator,
     pub write_metadata: bool,
     pub perform_codegen: bool,
+    /// If set to true the compiler won't try and analyse any of the package's
+    /// modules and always succeed compilation returning no compile modules.
+    ///
+    /// Code generation is still carried out so that a root package will have an
+    /// entry point nonetheless.
+    ///
+    pub skip_analysis: bool,
```
lpil, line 51: I don't think this name matches what it does now.

compiler-core/src/build/package_compiler.rs
```diff
@@ -105,68 +113,74 @@ where
             return e.into();
         }
 
-        let artefact_directory = self.out.join(paths::ARTEFACT_DIRECTORY_NAME);
-        let codegen_required = if self.perform_codegen {
-            CodegenRequired::Yes
+        let outcome = if self.skip_analysis {
```
lpil, line 116: There's lots of nesting here and the second branch is very long, making it hard to read. How about this instead?

```rust
// some docs here
let loaded = if self.compile_modules {
  loader.run()
} else {
  Loaded::empty()
};
```

There's more setup done in the case when it's not used, but I think it'd be easier to follow.

compiler-core/src/build/package_compiler.rs
giacomocavalieri: I've renamed it to `compile_modules` and switched the default value, hope it is better now

compiler-core/src/build/package_compiler.rs
giacomocavalieri: Yeah agreed!

lpil (approved):
Thank you!!!

Issue #3503, opened 2024-08-11
Inconsistency with tag name format expected for source links in docs based on repository type


Nicd:
1. Create project with version `1.0.0` and configure `gleam.toml` to point to GitHub/GitLab/Bitbucket as the repository. Observe that generated source links in the docs use `v1.0.0` format with "v" prefix to link to the correct source.
2. Change repository type to e.g. `forgejo` (`{ type = "forgejo",   host = "example.com", user = "example", repo = "my_project" }`). After regeneration of docs, observe that source links now point to `1.0.0` without "v" prefix.

From here on in the code we can see that GitLab/GitHub/Bitbucket have the "v" prefix in the URL and Gitea (Forgejo), CodeBerg, and SourceHut do not: https://github.com/gleam-lang/gleam/blob/8743158ef7ac8c8bebd045fadb4f4775a0e7cc8d/compiler-core/src/docs/source_links.rs#L30

Note: Fixing this would break existing users of the latter three who depend on there _not_ being a "v" prefix.

lpil:
Thank you

sobolevn:
@Nicd would you like to work on this? If not, I can take this over :)

Nicd:
Go right ahead, I don't know when I would have time to take a look at it (I'm supposed to be in bed right now too!). It's a tiny change but need to check the tests and so on.

Nicd:
An additional thing I just noticed: the "Repository" link that is generated has a double slash. So `repository = { type = "forgejo", host = "git.ahlcode.fi", user = "nicd", repo = "scriptorium" }` results in a link to `https://git.ahlcode.fi//nicd/scriptorium`.

PR #3514, opened 2024-08-15
Fix tag generation URL in docs for some VCSs

sobolevn:
The change itself is rather simple, however, I am not really comfortable with it.

Imagine this use-case:
1. Some user has SourceHut repo configured. They released `1.0.0` version and their doc links work as they expect to.
2. With gleam version update their links will change, so their tag naming scheme would also have to change: from `1.0.0` to `v1.0.0`
3. Docs does not say anything about this (at least ones I was able to find): https://gleam.run/writing-gleam/command-line-reference/#docs
4. It is basically a breaking change (despite it being a bug fix probably). So, we cannot release it in a patch release. And we need to communicate the change correctly. In CPython we provide a `DeprecationWarning` for 2 minor releases before the behavior change itself.

However, I guess that there would not be a lot of real affected users, because platforms are not very popular. So, I guess it is better to do this sooner than later.

Closes https://github.com/gleam-lang/gleam/issues/3503

### DIFF
--- a/compiler-core/src/docs/source_links.rs
+++ b/compiler-core/src/docs/source_links.rs
@@ -51,21 +51,21 @@ impl SourceLinker {
             )),
             Repository::CodeBerg { user, repo } => Some((
                 format!(
-                    "https://codeberg.org/{}/{}/src/tag/{}/{}#L",
+                    "https://codeberg.org/{}/{}/src/tag/v{}/{}#L",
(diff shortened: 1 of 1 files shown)


lpil (approved):
I think this is OK given the CLI tells you to push the tag with the `v` at the start, the platforms are niche, it's easy to fix, and we don't have an alternative approach.

PR #3516, opened 2024-08-16
LSP: Include braces in type import completions

### LINKED ISSUE #3498: LSP: autocompleting type imports should add brackets if needed
Right now importing types becomes:

`import foo.type Bar` instead of `import foo.{type Bar}`

crazymerlyn:
Fixes #3498

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -62,6 +62,8 @@ fn sort_text(kind: CompletionKind, label: &str) -> String {
 enum TypeCompletionForm {
     // The type completion is for an unqualified import.
     UnqualifiedImport,
+    // The type completion is for an unqualified import within braces.
+    UnqualifiedImportWithinBraces,
     Default,
 }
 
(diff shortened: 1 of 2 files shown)


lpil (commented):
Looks good! Could you add some tests and update the changelog please 🙏

crazymerlyn:
I tried to add tests but ran into an issue. With or without my change, the tests don't generate any completions for `import dep.|` while working just fine for `import dep.{|}`. Since the behavior hasn't really changed for the second case, it's hard to add a test for it.

I did test it by installing it locally and trying it out in my editor but can't seem to replicate that behavior in the tests. Regardless of whether i use `assert_completion!` or `assert_completion_with_prefix` i can't get the completion to be generated in the case of `import dep.|`. Not idea why that is considering the lsp generates those completions just fine when actually using it.

lpil:
Are you still working on this @crazymerlyn ?

crazymerlyn:
Haven't looked at it since my last comment. Don't really have an idea for what to do next in terms of testing this change.

Frank-III:
> I tried to add tests but ran into an issue. With or without my change, the tests don't generate any completions for `import dep.|` while working just fine for `import dep.{|}`. Since the behavior hasn't really changed for the second case, it's hard to add a test for it.
> 
> I did test it by installing it locally and trying it out in my editor but can't seem to replicate that behavior in the tests. Regardless of whether i use `assert_completion!` or `assert_completion_with_prefix` i can't get the completion to be generated in the case of `import dep.|`. Not idea why that is considering the lsp generates those completions just fine when actually using it.

I think this happens because `import dep.|` results in a parse error and make the completion failed at this stage:
```gleam            
let module = match this.module_for_uri(&params.text_document.uri) {
                Some(m) => m,
                None => return Ok(None),
}; // can't get the module
```
probably need to figure out how to test this properly

lpil:

Frank-III:
Hi @lpil, I would like to help finish this one as the logic is completed, we just need a way to test it. since `import dep.|` is not a valid syntax, the current way of testing does not apply to this one, do you have suggestions on how to test it? Appreciated!

lpil:
You would need to compile it with valid syntax, update the file in the filesystem and then ask for compilations!

giacomocavalieri:
Hello! Since #3963 was merged, I think this one can be closed

PR #3519, opened 2024-08-16
Fix bit array codegen bug

giacomocavalieri:
This PR fixes a bug with Erlang code generation.

This bug cannot happen unless the list of typed options is empty and it never is since the compiler always adds one, so I can't really add a test for it. Still I think it was worth fixing in case this invariant changes in the future


lpil (commented):
Could you add a test for this please 🙏

lpil (approved):
Thank you

PR #3520, opened 2024-08-16
Replace `typ` with `type_`

giacomocavalieri:
I remove all leftover occurrences of `typ` in favour of `type_` so everything should be consistent now (and in the capnp schema there where two leftover `typ` fields that I renamed to `type` like all the other ones).


lpil (approved):
Thank you!!

PR #3522, opened 2024-08-17
Improve lambda functions inference in pipes

sobolevn:
This is the second step after https://github.com/gleam-lang/gleam/pull/3460

Now I reuse the function inference we have in `expression.rs` to infer individual `fn` expressions in pipes.

Closes https://github.com/gleam-lang/gleam/issues/2504

### DIFF
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -2356,6 +2356,20 @@ pub fn main() {
 }
 
 #[test]
+fn pipe_with_annonymous_unannotated_functions_wrong_arity3() {
+    assert_module_error!(
+        r#"
+pub fn main() {
+  let a = 1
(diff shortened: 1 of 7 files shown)


compiler-core/src/type_/tests.rs
```diff
@@ -2311,3 +2311,100 @@ fn assert_suitable_main_function_javascript_not_supported() {
...
+#[test]
+fn pipe_with_annonymous_mixed_functions() {
+    assert_module_infer!(
+        r#"
+pub fn main() {
+  let a = "abc"
+     |> fn (x) { #(x, x <> "d") }
+     |> fn (x) { x.0 }
+     |> fn (x: String) { x }
+}
```
lpil, line 2381: In Gleam we don't put a space between `fn` and `(`

lpil (approved):
Thank you so much!!

PR #3523, opened 2024-08-17
refactoring: Use `SpannedString` alias where possible

sobolevn:
This alias is quite useful, because we have a lot of similar annotations.


lpil (approved):
Thank you

Issue #3515, opened 2024-08-16
pipeline issue in 1.4.1


nohzafk:
Following code can be compile in gleam 1.3.2 but not in 1.4.1

Here is the code that can reproduce the issue

```gleam
import gleam/list

pub fn relu(t) {
  fn(theta: List(Int)) {
    // use t and theta and return a Float
    0.0
  }
}

pub fn k_relu(k: Int) {
  fn(t: Float) {
    fn(theta: List(Int)) {
      case k {
        0 -> t
        _ -> {
          // following code is OK on gleam 1.3.2 and gleam 1.4.1               
          let next_layer = k_relu(k - 1)(relu(t)(theta))
          next_layer(list.drop(theta, 2))
          // following code is OK on gleam 1.3.2, but raise error on gleam 1.4.1
          // let next_layer = theta |> relu(t) |> k_relu(k - 1)
          // theta |> list.drop(2) |> next_layer
        }
      }
    }
  }
}
```

this is the error on gleam 1.4.1
```
error: Type mismatch
   ┌─ /Users/randall/projects/the_gleam_learner/src/demo.gleam:11:3
   │
11 │ ╭   fn(t: Float) {
12 │ │     fn(theta: List(Int)) {
13 │ │       case k {
14 │ │         0 -> t
   · │
22 │ │     }
23 │ │   }
   │ ╰───^

The type of this returned value doesn't match the return type
annotation of this function.

Expected type:

    fn(List(Int)) -> Float

Found type:

    fn(Float) -> fn(List(Int)) -> Float

error: Type mismatch
   ┌─ /Users/randall/projects/the_gleam_learner/src/demo.gleam:18:37
   │
18 │           let next_layer = theta |> relu(t) |> k_relu(k - 1)
   │                                     ^^^^^^^

Expected type:

    Int

Found type:

    Float

error: Incorrect arity
   ┌─ /Users/randall/projects/the_gleam_learner/src/demo.gleam:18:48
   │
18 │           let next_layer = theta |> relu(t) |> k_relu(k - 1)
   │                                                ^^^^^^^^^^^^^ Expected 1 argument, got 2
```

lpil:
Thank you

sobolevn:
I might have an idea on how to solve it! PR is coming soon.

giacomocavalieri:
Hero! I had no idea where to start 😆

sobolevn:
@giacomocavalieri me neither, that's why I used this script.

```bash
[...]

PR #3525, opened 2024-08-18
Fix function calls inference in pipes

sobolevn:
So, why did I make this change?

In some cases `return_.fn_arity()` was returning `None`, because `return_` was not a function type, but `Type::Var { type_ }`. Which does not have an arity.

Why is it safe not to check the arity here? Because it will be checked in the second `self.expr_typer.do_infer_call_with_known_fun` call here: https://github.com/gleam-lang/gleam/blob/72f4ac9cf1e3bfe5d188076b95561acd9876dd6b/compiler-core/src/type_/pipe.rs#L249-L260

Closes https://github.com/gleam-lang/gleam/issues/3515

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,6 +93,11 @@
   arrays.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug which affected inference of function
+  calls in pipe expressions.
+
+  ([sobolevn](https://github.com/sobolevn))
+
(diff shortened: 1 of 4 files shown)


sobolevn:
CC @nohzafk @giacomocavalieri

sobolevn:
Oups, somehow unrelated changes got into this PR, removing and updating.

lpil (approved):
Thank you

Issue #2426, opened 2023-11-27
Inexhaustive warning message is incorrect for case expressions with multiple subjects


lpil:
```
warning: Inexhaustive patterns
  ┌─ /src/warning/wrn.gleam:3:3
  │  
3 │ ╭   case x, y {
4 │ │     True, _ -> 1
5 │ │   }
  │ ╰───^

This case expression does not have a pattern for all possible values.
If is run on one of the values without a pattern then it will crash.

The missing patterns are:

    False
```

The missing pattern should be `False, _`. The fact there are multiple patterns has been lost.

TanklesXL:
Heya @lpil , this may or may not be related to the following warning I got with exhaustiveness and guards. Given the following snippet:

```gleam
  case 1 {
    x if x < 0 -> panic
    x if x > 0 -> panic
    x if x == 0 -> panic
  }
```

Even though a given integer must be either greater, equal to or lower than 0, I get an exhaustiveness warning.

```text
warning: Inexhaustive patterns
  ┌─ .../code/gleam/exhaust/src/exhaust.gleam:4:3
  │  
4 │ ╭   case 1 {
5 │ │     x if x < 0 -> panic
6 │ │     x if x > 0 -> panic
7 │ │     x if x == 0 -> panic
8 │ │   }
  │ ╰───^

This case expression does not have a pattern for all possible values.
If is run on one of the values without a pattern then it will crash.

The missing patterns are:

    _

In a future version of Gleam this will become a compile error.
```

Maybe I'm misunderstanding how the exhaustiveness checking currently works and this is totally fine though, it just seemed a bit odd to me 😄

michallepicki:
An interesting example:
```gleam
import gleam/io

pub fn main() {
  let x = False
  let y = True
  case x, y {
    True, True -> 1
  }
}
```

The error is:
```
warning: Inexhaustive patterns
  ┌─ /src/main.gleam:6:3
  │  
6 │ ╭   case x, y {
7 │ │     True, True -> 1
8 │ │   }
  │ ╰───^

This case expression does not have a pattern for all possible values.
If is run on one of the values without a pattern then it will crash.

The missing patterns are:

    False
    True

In a future version of Gleam this will become a compile error.
```

michallepicki:
Copying my notes from Discord to here:

[...]

PR #3537, opened 2024-08-19
LSP Code Action: Add missing patterns to inexhaustive case expression

### LINKED ISSUE #2957: LSP code action: add missing patterns to case expression

GearsDatapacks:
Closes #2957 
I have added a code action which fills in the missing patterns for an inexhaustive case expression. The formatting of the resulting syntax isn't great at the moment, but it works.
Let me know if I should work on adapting the generated code actions to produce nicer looking code as per the situation.

NOTE: This currently doesn't work properly with multi-patterns due to #2426. But I think it's useful enough to have now, and it will work fine as soon as that issue is fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -259,6 +259,26 @@
 
   ([Surya Rose](https://github.com/gearsdatapacks))
 
+- The Language Server can now suggest a code action to fill in the missing
+  patterns of a case expression:
+
+  ```gleam
+  let a = True
+  case a {}
+  ```
+
+  Becomes:
+
+  ```gleam
+  let a = True
+  case a {
+    False -> todo
+    True -> todo
+  }
+  ```
+
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Bug Fixes
 
 - Fixed a bug where the warnings were printed above the errors without any new line between them.
@@ -273,7 +293,7 @@
   ([Surya Rose](https://github.com/gearsdatapacks))
 
 - Fixed a bug where incorrect syntax error message were shown, when using `:` or
-  `=` in wrong possitions in expressions.
+  `=` in wrong positions in expressions.
   ([Ankit Goel](https://github.com/crazymerlyn))
 
 - Fixed a bug where the compiler would crash when pattern matching on a type
--- a/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_bool.snap
+++ b/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_bool.snap
@@ -1,0 +1,21 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\npub fn main() {\n  let bool = True\n  case bool {}\n}\n"
+---
+----- BEFORE ACTION
+
+pub fn main() {
+  let bool = True
+  case bool {}
+  ▔▔▔▔▔↑      
+}
+
+
+----- AFTER ACTION
+
+pub fn main() {
+  let bool = True
+  case bool {
+    _ -> todo
+}
+}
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -105,6 +105,28 @@
 
 ### Language Server
 
+- The Language Server can now suggest a code action to fill in the missing
+  patterns of a case expression:
+
+  ```gleam
+  case #(True, False) {
```
lpil, line 112: Could you use a different example please 🙏  We don't want to recommend using tuples like this

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_bool.snap
```diff
@@ -0,0 +1,21 @@
...
+  ▔▔▔▔▔↑      
+}
+
+
+----- AFTER ACTION
+
+pub fn main() {
+  let bool = True
+  case bool {
+    _ -> todo
```
lpil, line 19: This isn't very helpful. I would be very frustrated if the action gave me this.

Is there a way to add the constructors here?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_custom_type.snap
```diff
@@ -0,0 +1,38 @@
...
+  Wibble
+  Wobble
+  Wubble
+}
+
+pub fn main() {
+  let wibble = Wobble
+  case wibble {
+    Wobble -> Nil
+
```
lpil, line 34: Remove this blank line please 🙏

lpil (commented):
Thank you! Some notes inline

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_bool.snap
GearsDatapacks: Unfortunately this is more a problem with the exhaustiveness checker. For this code:
```gleam
pub fn main() {
  case True {
    
  }
}
```

The compiler produces this error:
```
error: Inexhaustive patterns
  ┌─ /home/gears/projects/temp/testing_gleam/src/testing_gleam.gleam:5:3
  │  
5 │ ╭   case True {
6 │ │     
7 │ │   }
  │ ╰───^

The missing patterns are:

    _
```

I was planning to open another issue after this about improving these suggestions, but should I open that now and leave this on hold until that is resolved?

lpil: Aye, I understand what the codebase does today. What I mean is that to be a useful addition this code action needs to be able to add constructors for empty case expressions.

GearsDatapacks:
I've updated the parts which you suggested and are directly related to this PR. I guess I'll leave this as a draft for now, and open an issue to improve the exhaustiveness checker before continuing with this.

lpil:
I've found a bug!

This code results in extra spacing
```rust
pub fn handle_request(request: Request) {
  case request.method {
    
  }
}
```
```rust
  case request.method {
    
  
    http.Connect -> todo
    http.Delete -> todo
    http.Get -> todo
    http.Head -> todo
    http.Options -> todo
    http.Other(_) -> todo
    http.Patch -> todo
    http.Post -> todo
    http.Put -> todo
    http.Trace -> todo
  }
```

GearsDatapacks:
Alright, that case should be fixed now

GearsDatapacks:
Actually there's one more case I want to fix before this is merged:
```gleam
let value = case
  some_very_long_variable_name_which_causes_the_formatter_to_wrap
 {}
```

Since it determines the indentation based on the start of the case expression, it thinks all the `let value = ` is indentation, and indents the patterns way too much

GearsDatapacks:
That's all taken care of now

lpil (approved):
Thank you!!! I love this so much!

Issue #3531, opened 2024-08-19
Improve error message for defining a `fn` inside `type`


sobolevn:
Improve error message for defining a function inside a type.
I think that this is important because many people will try to do that because of OOP background and `class` / methods mindset.

This is the code I am writting:

```gleam
pub type SchoolPerson {
  Student(String)

  fn some() {}
}
```

Results in:

```
error: Syntax error
  ┌─ /src/main.gleam:4:3
  │
4 │   fn some() {}
  │   ^^ I was not expecting this

Found the keyword `fn`, expected one of: 
- `}`
- a record constructor
```

The same message in Rust:

```rust
struct Some {
    field: u32,

    fn do_some() {
        1
    }
}
```

Results in:

```
error: functions are not allowed in struct definitions
  --> compiler-core/src/docs.rs:41:5
   |
38 |   struct Some {
   |          ---- while parsing this struct
...
41 | /     fn do_some() {
42 | |         1
43 | |     }
   | |_____^
   |
   = help: unlike in C++, Java, and C#, functions are declared in `impl` blocks
   = help: see https://doc.rust-lang.org/book/ch05-03-method-syntax.html for more information
```

My proposal:

```
error: Syntax error
  ┌─ /src/main.gleam:4:3
  │
4 │   fn some() {}
  │   ^^ I was not expecting this

Found the keyword `fn`, expected one of: 
- `}`
- a record constructor
Hint: unlike some other languages, functions are declared separately from types.
```

lpil:
I've not seen this before, is this something we've had reports of people doing?

sobolevn:
This is something that I tried doing my first time :)

lpil:
Let's add this then!

PR #3539, opened 2024-08-20
Add a hint about placing functions into types

sobolevn:
Closes https://github.com/gleam-lang/gleam/issues/3531


lpil (approved):
Thank you!

PR #3540, opened 2024-08-20
Fix several bit_array tests on js to expect a compilation error on supported features

sobolevn:
There are several parts to it.
1. Errors are produced together with the invalid code.
2. They were not checked, invalid code was asserted instead.
3. gleam.run probably also does not raise these compilation errors to user for some reason


lpil (approved):
Thank you

PR #3543, opened 2024-08-20
CHANGELOG

giacomocavalieri:
- I've added a couple of examples in the CHANGELOG to showcase some of the new features
- I removed the duplicated entry _"Adds a hint to syntax error when defining a function inside a type."_ (it was already described in depth by the entry before it)
- I made sure all entries have the same style (at most 80 chars long, filling each line as much as possible before going on a new one)


lpil (approved):
So nice!!

PR #3545, opened 2024-08-20
Better errors

lpil:


GearsDatapacks:
Sorry if this is not the place to put this but:
Would it make sense to make these configurable somehow? I really like how in the gleam compiler, you have the "This is a bug in the gleam compiler, sorry!" whenever the program unexpectedly crashes. I wonder if it would be possible to implement a way of configuring it like that in gleam.
I haven't actually looked into how it is done in rust, or how this is handled on the erlang side, so I have no idea if this is even possible. Just an idea I had

lpil:
If you want to print something when there's a runtime error then you would catch the exception and print, the same as this code does.

PR #3547, opened 2024-08-20
Fix warning new line bug

### LINKED ISSUE #3535: Missing empty line between last warning and first error
When there's both warnings and errors in a Gleam project the last warning and the first error are printed with no empty line separating them
Screenshot 2024-08-19 alle 18 28 23

vkobinski:
Fixes #3535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,9 @@
 
 ### Bug Fixes
 
+- Fixed a bug where the warnings were printed above the errors without any new line between them.
+([Victor Kobinski](https://github.com/vkobinski))
+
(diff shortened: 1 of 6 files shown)


lpil (commented):
Hello! Thank you, but this seems a bit complex. How about we move the newline at the start of the printed warning to the end of it instead?

vkobinski:
I made this change to make the new line be in the end of the block of warnings. Because every warning has a new line at the end of it, but the block just have the new line of the last warning.

lpil:
Let's go with the approach of moving the newline to the end, a 1 line change and no additions to the IO injection system is good 👍

vkobinski:
Makes sense, it's better this way. I will change it!

PR #3553, opened 2024-08-21
Improve import error suggestions

sobolevn:
This is the first PR in the series. 

This case is not covered yet:

```rust
#[test]
fn import_submodule_as_value() {
    assert_with_module_error!(
        ("one/two", "pub fn whatever() {}"),
        "
        import one.{two}
        "
    );
}
```

It requires a bit tricky logic to get right, so I decided to split it out.

Better wordings are more than welcome! 👍

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,7 +113,9 @@
                ^^
                I was not expecting this.
 
-  Hint: did you mean `import gleam/io`?
+  This syntax for import is not correct. Probably, you meant:
+  - `import gleam/io` to import module `io` from package `gleam`
+  - `import gleam.{io}` to import value `io` from module `gleam`
   ```
   ([sobolevn](https://github.com/sobolevn))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,6 +106,12 @@
 - Adds a hint to syntax error when defining a function inside a type.
   ([sobolevn](https://github.com/sobolevn))
 
+- Improve import error messages.
+
+  Now we recommend fixes for `import a.b`
+  and have better suggestions for `import a/b` when `b` is acutally a value.
+  ([sobolevn](https://github.com/sobolevn))
+
 ### Formatter
 
 ### Language Server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -106,10 +106,15 @@
 - Adds a hint to syntax error when defining a function inside a type.
   ([sobolevn](https://github.com/sobolevn))
 
-- Improve import error messages.
+- Improve import error messages. Now this code produces:
 
-  Now we recommend fixes for `import a.b`
-  and have better suggestions for `import a/b` when `b` is acutally a value.
+  ```gleam
+  import gleam.io
+               ^^
+               I was not expecting this.
+
+  Hint: did you mean `import gleam/io`?
+  ```
   ([sobolevn](https://github.com/sobolevn))
 
 ### Formatter
--- a/compiler-core/src/analyse/imports.rs
+++ b/compiler-core/src/analyse/imports.rs
@@ -1,4 +1,5 @@
 use ecow::EcoString;
+use itertools::Itertools;
 
 use crate::{
     ast::{SrcSpan, UnqualifiedImport, UntypedImport},
@@ -48,9 +49,28 @@ impl<'context, 'problems> Importer<'context, 'problems> {
 
         // Find imported module
         let Some(module_info) = self.environment.importable_modules.get(&name) else {
+            // Improve error message when users confuse `import one/two`
+            // with `import one.{two}`.
+            let mut modpath = name.split("/").collect_vec();
+            let last_part = match modpath.last() {
+                Some(name) => name,
+                None => "",
+            };
+            modpath.truncate(modpath.len() - 1);
+            let basename = EcoString::from(modpath.join("/"));
+
+            let mut hint = None;
+            if let Some(module) = self.environment.importable_modules.get(&basename) {
+                if module.get_public_value(last_part).is_some() {
+                    hint = Some(format!(
+                        "Maybe you meant `import {basename}.{{{last_part}}}`?"
+                    ));
+                }
+            }
             self.problems.error(Error::UnknownModule {
                 location,
                 name: name.clone(),
+                hint,
                 imported_modules: self.environment.imported_modules.keys().cloned().collect(),
             });
             return;
--- a/compiler-core/src/analyse/imports.rs
+++ b/compiler-core/src/analyse/imports.rs
@@ -4,6 +4,7 @@ use itertools::Itertools;
 use crate::{
     ast::{SrcSpan, UnqualifiedImport, UntypedImport},
     build::Origin,
+    error::wrap,
     type_::{
         EntityKind, Environment, Error, ModuleInterface, Problems, UnusedModuleAlias,
         ValueConstructorVariant,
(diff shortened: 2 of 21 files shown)


compiler-core/src/parse/snapshots/gleam_core__parse__tests__import_dotted_name.snap
```diff
@@ -0,0 +1,13 @@
...
+---
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:12
+  │
+2 │ import one.two
+  │            ^^^ I was not expecting this
+
+This syntax for import is not correct. Probably, you meant:
+- `import one/two` to import module `two` from package `one`
+- `import one.{two}` to import value `two` from module `one`
```
giacomocavalieri, line 13: I think we could have a shorter message if we can't figure out what the programmer meant, something along the lines of 
"To import a value from a module you have to write it like this: `import one.{Two}`" and then have a link to the relevant piece of the gleam tour for more detail on imports (there's other error messages that do this you could take as reference, like the ones for a wrong use usage)

And then we can provide an hint right under the import if we can figure out a good suggestion that wouldn't result in an error. So all in all it would be something like this:

- If the thing they're trying to import is actually a module value
  ```txt
  error: Syntax error
    ┌─ /src/parse/error.gleam:2:12
    │
  2 │ import one.two
    │        ^^^^^^^ Did you mean `one.{two}`

  This is not a valid import, to import a value from a module you
  have to write: `import one.{two}`.
  
  See: https://tour.gleam.run/basics/modules/
  ```
- If the thing they're trying to import is actually a module type
  ```txt
  error: Syntax error
    ┌─ /src/parse/error.gleam:2:12
    │
  2 │ import one.Two
    │        ^^^^^^^ Did you mean `one.{type Two}`?

  This is not a valid import, to import a type from a module you
  have to write: `import one.{type Two}`.

  See: https://tour.gleam.run/basics/type-imports/
  ```
- If the thing they're trying to use is another module:
  ```txt
  error: Syntax error
    ┌─ /src/parse/error.gleam:2:12
    │
  2 │ import gleam.io
    │        ^^^^^^^^ Did you mean `gleam/io`?

  This is not a valid import, to use the `io` module you have to write
  `import gleam/io`.

  See: https://tour.gleam.run/basics/modules/
  ```
- If it's none of the above then we can have something a bit more generic with
  more options like the one you're displaying here
  
What do you think @lpil?

CHANGELOG.md
```diff
@@ -106,6 +106,12 @@
 - Adds a hint to syntax error when defining a function inside a type.
   ([sobolevn](https://github.com/sobolevn))
 
+- Improve import error messages.
+
+  Now we recommend fixes for `import a.b`
+  and have better suggestions for `import a/b` when `b` is acutally a value.
+  ([sobolevn](https://github.com/sobolevn))
```
giacomocavalieri, line 113: Could you add an example with a bit of Gleam code, something like this:
```gleam
import gleam.io
```
Showcasing the corresponding error message suggesting the correct import? I think it would be nicer to see examples for people reading the changelog

compiler-core/src/parse/snapshots/gleam_core__parse__tests__import_dotted_name.snap
sobolevn: `import one.Two` is not a valid syntax, so I can only produce error messages in the parser. Parser does not have enough info to tell if this type actually exists. So, we can recommend an invalid code. I am not very confident about that.

But it general - I understood your idea and it indeed looks better! I will rework my PR to make errors shorter and provide a link to the docs.

giacomocavalieri: Ah right I had missed that this is a syntax error, that makes sense

sobolevn:
Ok, I failed to make suggestions shorter 🙈 They are now longer.

Changes:
1. I modified how `did_you_mean` suggestions are applied, because we can literally have this use-case:

```rust
#[test]
fn import_value_as_submodule3() {
    // Two possible suggestions: `import one/wobble` or `import one.{wibble}`
    // One via hint, one via `did_you_mean`.
    assert_with_module_error!(
        ("one", "pub fn wibble() {}"),
        ("one/wobble", ""),
        "import one/wibble"
    );
}
```

We should suggest both:
- `import one/wobble`
- `import one.{wibble}`

Because it is not clear what was meant. Now we do both suggestions.

- I changed `imported_modules` to `importable_modules`, because `imported_modules` were always empty in `analyze`, so we can have better suggestions.
- I provided a link to the docs, since this topic is rather complex.

Sadly, we cannot apply suggestions from https://github.com/gleam-lang/gleam/pull/3553#discussion_r1725421247 which are great, because of syntax error limitations :(
Ideally there should be no syntax error here and all errors should come from `analyze` step, but it is a very complex and probably-not-justified change.

Please, take a look!

CHANGELOG.md
sobolevn: Done!

compiler-core/src/analyse/imports.rs
```diff
@@ -48,10 +50,51 @@ impl<'context, 'problems> Importer<'context, 'problems> {
...
+                }
+            }
+            let importable_modules = self
+                .environment
+                .importable_modules
+                .keys()
+                .cloned()
+                .collect_vec();
+            // If we have a single module here, it means that it is `gleam` module.
+            // We don't want to suggest that.
```
lpil, line 87: Why not?

compiler-core/src/parse/error.rs
```diff
@@ -358,6 +388,11 @@ pub enum ParseErrorType {
         field_type: Option<TypeAst>,
     },
     CallInClauseGuard, // case x { _ if f() -> 1 }
+    InvalidImportSyntax {
```
lpil, line 391: I'm not sure what this means. Can you add documentation and a clearer name please.

compiler-core/src/type_/error.rs
```diff
@@ -958,6 +959,7 @@ pub fn convert_get_value_constructor_error(
             location,
             name,
             imported_modules,
+            hint: None,
```
lpil, line 962: There should be no error formatting in the analysis modules, instead errors should be represented using types and then they are printed in the error module.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__import_value_as_submodule.snap
```diff
@@ -0,0 +1,14 @@
...
+expression: import one/wibble
+---
+error: Unknown module
+  ┌─ /src/one/two.gleam:1:1
+  │
+1 │ import one/wibble
+  │ ^^^^^^^^^^^^^^^^^
+
+No module has been found with the name `one/wibble`.
+Hint: Did you mean `import one.{wibble}`?
```
lpil, line 12: Can you add a test like this but with a module name that's longer than 80 characters please.

CHANGELOG.md
```diff
@@ -106,6 +106,19 @@
...
 
+- Improve import error messages. Now this code produces:
+
+  ```gleam
+  import gleam.io
+               ^^
+               I was not expecting this.
+
+  This syntax for import is not correct. Probably, you meant:
+  - `import gleam/io` to import module `io` from package `gleam`
```
lpil, line 117: `gleam` isn't a package, this would be to import the module `gleam/io`

compiler-core/src/type_/error.rs
```diff
@@ -109,6 +109,13 @@ pub enum Error {
         imported_modules: Vec<EcoString>,
     },
 
+    UnknownModuleWithRichSuggestions {
+        location: SrcSpan,
+        name: EcoString,
+        name_parts: (EcoString, EcoString),
+        importable_modules: Vec<EcoString>,
+    },
```
lpil, line 117: Merge this into the existing unknown module error please

compiler-core/src/type_/error.rs
```diff
@@ -109,6 +109,13 @@ pub enum Error {
         imported_modules: Vec<EcoString>,
     },
 
+    UnknownModuleWithRichSuggestions {
+        location: SrcSpan,
+        name: EcoString,
+        name_parts: (EcoString, EcoString),
```
lpil, line 115: What are name parts?

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__import_value_as_submodule_very_long_name.snap
```diff
@@ -0,0 +1,16 @@
...
+error: Unknown module
+  ┌─ /src/one/two.gleam:1:1
+  │
+1 │ import this_is_a_very_long_name_for_a_module_even_hard_to_read/this_is_a_very_long_submodule_name
+  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Did you mean `this_is_a_very_long_name_for_a_module_even_hard_to_read`?
+
+No module has been found with the name `this_is_a_very_long_name_for_a_module_even_hard_to_read/this_is_a_very_long_submodule_name`.
+Hint: Did you mean `import
+this_is_a_very_long_name_for_a_module_even_hard_to_read.
+{this_is_a_very_long_submodule_name}`?
```
lpil, line 14: This isn't the correct formatting. Wrapping shouldn't be applied to code.

lpil (commented):
I'm finding it confusing the way that 2 different suggestions are presenting in 2 different ways now. I think it should only present 1 suggestion and it should always be presented in the same way.

lpil:

sobolevn:
Sorry, no :(
I got busy on my work projects.

lpil:
Thank you

PR #3556, opened 2024-08-22
skip optional dependencies

### LINKED ISSUE #1964: Do not download optional dependencies by default

### LINKED ISSUE #3555: Skip optional dependencies
Elixir packages allows to mark a dependency as optional, which reduces compile time and also allows you to select better runtime support, algorithms and additional features.

Currently, [`rustler`](https://github.com/rusterlium/rustler/tree/master) (an FFI package between Rust and Erlang) compiles the packages with `rustler` as an optional dependency in case you don't want to use the prebuilt binaries.

Rustler is currently one of the most important packages for FFI with Erlang and it's crucial that Gleam supports it well.
By skipping the compilation of the optional dependencies and letting the users add the optional dependency by themselves, rustler is supported without effort.

gusinacio:
Fixes #1964

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,8 @@
 - Improved the error message for unknown record fields, displaying an additional
   note on how to have a field accessor only if it makes sense.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+- The compiler now ignores `optional` dependencies when resolving versions unless explicitly specified.
+  ([Gustavo Inacio](https://github.com/gusinacio))
 
 ### Formatter
 
--- a/compiler-core/src/dependency.rs
+++ b/compiler-core/src/dependency.rs
@@ -595,6 +595,22 @@ mod tests {
     }
 
     #[test]
+    fn resolution_with_optional_deps_incompatible() {
+        let result = resolve_versions(
+            make_remote(),
+            HashMap::new(),
+            "app".into(),
+            vec![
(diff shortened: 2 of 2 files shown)


lpil (commented):
Thank you!

The code looks good, but it's not quite the behaviour we need. The dependency needs to only be included if another package requires it, but the constraint still needs to taken into account when this is the case. The first need is met here, but not the second.

gusinacio:
Hello @lpil, I added back the constraints. Now in case it was optional and a later dep is required, we take into account the new version. If it was required and we find an optional dep, we also add the requirement so we can resolve the versions correctly.

lpil (approved):
Lovely work! Thank you very much!

PR #3557, opened 2024-08-22
Infer minimum Gleam version

### LINKED ISSUE #3492: Infer minimum Gleam version
Currently you can specify the minimum required Gleam version, but there's nothing to ensure that you have got it correct.

When analysing the code keep track of what features are used and use that to determine what the minimum Gleam version would be, and then complain if the version in `gleam.toml` is too permissive.

We don't need to worry about v0.* versions.

We could automatically add it to the package config before publishing if none is set.

giacomocavalieri:
This PR closes #3492

It's still a WIP because I have some questions before I can move forward:
1. I used the `Version` type that comes from `hexpm::version`, should we use our own minimal custom type instead of relying on this one?
2. If someone specifies an insufficient version (let's say `gleam.toml` has `1.2` but we inferred the minimum one to be `1.5`) how should the error be presented to the user?
   Do we want the error message to include an example of a usage of a feature that explains why we inferred that number? For example something that looks like this:
   ```txt
   error: invalid Gleam version
   
   1 | tuple.0.1
            ^^^^

   Nested tuple access requires a Gleam version >= 1.2 but the version specified in
   your `gleam.toml` is 1.1!
   ```
   Even if someone used multiple features I think it would be better to show just one, I don't think there's much value in making the error message any longer
3. What if the `gleam.toml` doesn't specify any gleam version? Do we want to manually add the constraint when publishing it or have an error and force the programmer to explicitly write the inferred version down? Is there any downside to doing it implicitly?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -818,6 +818,7 @@ dependencies = [
  "lsp-types",
  "opener",
  "pretty_assertions",
+ "pubgrub",
  "regex",
  "reqwest",
  "rpassword",
--- a/compiler-cli/Cargo.toml
+++ b/compiler-cli/Cargo.toml
@@ -35,6 +35,8 @@ fslock = "0"
 same-file = "1"
 # Open generated docs in browser
 opener = "0"
+# Pubgrub dependency resolution algorithm
+pubgrub = "0"
 camino = { workspace = true, features = ["serde1"] }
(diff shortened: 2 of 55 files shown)


lpil (approved):
Beautiful work as per usual!!!! 💜

PR #3561, opened 2024-08-24
Improve suggestions for inexhaustive case expression error

### LINKED ISSUE #3544: Improve suggestions for inexhaustive case expression error
Currently, for this code:
```gleam
case True {}
```

The compiler produces this error:

```
error: Inexhaustive patterns
  ┌─ FILE
  │  
5 │ ╭   case True {
6 │ │     
7 │ │   }
  │ ╰───^

The missing patterns are:

    _
```

This is not super helpful; it would be much better to suggest `True` and `False` as missing patterns instead. Also, #2957 cannot be implemented in a particularly useful manner until this is improved. (See [here](https://github.com/gleam-lang/gleam/pull/3537#discussion_r1723230130))

Edit: There was some confusion about the above code. The matching on the literal value `True`, has nothing to do with this. It's merely to demonstrate this error.

GearsDatapacks:
Closes #3544.
Instead of just returning `Failure` when there are no rows, it runs exhaustiveness checking on the first subject of the case expression to give better suggestions to improve it. 
In future we should maybe check other subjects too, to further improve this. Although this could produce a big decision tree, and we don't do that at the moment for other patterns so I left it.
For example, at the moment, for this code:
```gleam
case True, False {}
```
It will suggest `True, _` and `False, _`. (Actually it won't, because of #2426, but that's a separate issue)

However, if we check both subjects, it will suggest all four possible combinations.
However, consider this possibility:
```
type Wibble { A B C D }
case A, B, C, D {}
```
This would produce `4^4 = 256` possible options, which is a bit much. So I've left it with just the one for now.

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -265,7 +265,7 @@ impl<'a> Compiler<'a> {
 
     pub fn compile(mut self, rows: Vec<Row>) -> Match {
         Match {
-            tree: self.compile_rows(rows),
+            tree: self.check_empty_rows(rows),
             diagnostics: self.diagnostics,
         }
     }
@@ -297,7 +297,7 @@ impl<'a> Compiler<'a> {
         }
     }
 
-    fn compile_rows(&mut self, rows: Vec<Row>) -> Decision {
+    fn check_empty_rows(&mut self, rows: Vec<Row>) -> Decision {
         // If there are no rows, we get an immediate decision failure,
         // which gives a pretty unhelpful error message. Instead, we
         // run one pass of compiling to ensure we don't just suggest `_`
@@ -305,7 +305,7 @@ impl<'a> Compiler<'a> {
             // Even though we run a compile pass, an empty case expression is always
             // invalid, so we make sure to report that here
             self.diagnostics.missing = true;
-            let tree = self.compile_cases(
+            let tree = self.compile_rows_for_variable(
                 self.subject_variables
                     .first()
                     .expect("Must have at least one subject")
@@ -320,11 +320,11 @@ impl<'a> Compiler<'a> {
                 _ => tree,
             }
         } else {
-            self.do_compile_rows(rows)
+            self.compile_rows(rows)
         }
     }
 
-    fn do_compile_rows(&mut self, rows: Vec<Row>) -> Decision {
+    fn compile_rows(&mut self, rows: Vec<Row>) -> Decision {
         if rows.is_empty() {
             self.diagnostics.missing = true;
             return Decision::Failure;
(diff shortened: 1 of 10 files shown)


compiler-core/src/exhaustiveness.rs
```diff
@@ -290,6 +298,33 @@ impl<'a> Compiler<'a> {
     }
 
     fn compile_rows(&mut self, rows: Vec<Row>) -> Decision {
+        // If there are no rows, we get an immediate decision failure,
+        // which gives a pretty unhelpful error message. Instead, we
+        // run one pass of compiling to ensure we don't just suggest `_`
+        if rows.is_empty() {
+            // Even though we run a compile pass, an empty case expression is always
+            // invalid, so we make sure to report than here
```
sobolevn, line 306:
```suggestion
            // invalid, so we make sure to report that here
```

typo

compiler-core/src/exhaustiveness.rs
```diff
@@ -309,12 +344,35 @@ impl<'a> Compiler<'a> {
...
+        let variable = rows
+            .first()
+            .expect("Must have at least one row, otherwise we don't reach this point")
+            .columns
+            .iter()
+            .map(|col| col.variable.clone())
+            .max_by_key(|var| counts.get(&var.id).copied().unwrap_or(0))
+            .expect("The first row must have at least one column");
+
+        self.compile_cases(variable, rows)
```
lpil, line 371: What are cases? Is this something like `compile_rows_for_variable` or something?

compiler-core/src/exhaustiveness.rs
```diff
@@ -290,6 +298,33 @@ impl<'a> Compiler<'a> {
     }
 
     fn compile_rows(&mut self, rows: Vec<Row>) -> Decision {
+        // If there are no rows, we get an immediate decision failure,
```
lpil, line 301: There's still a `if rows.is_empty() {` in do_compile. How about we move the true branch from this if/else into a new method (handle_empty_row or something?) and then merge `compile_rows` and `do_compile_rows` back together.

compiler-core/src/type_/tests/exhaustiveness.rs
```diff
@@ -1129,3 +1131,38 @@ pub fn main() {
...
+
+#[test]
+fn empty_case_of_list() {
+    assert_error!(
+        "
+let list = []
+case list {}
+"
+    );
+}
```
lpil, line 1168: Can we have a test for the infinite types of ints, floats, and strings? 🙏

lpil (commented):
Oh this is lovely work, thank you very much! Left some small notes below

compiler-core/src/exhaustiveness.rs
GearsDatapacks: The entire algorithm sort of relies on that `Failure` being returned in `do_compile_rows`, it terminates the recursive process. If we try to continue compiling at that point, we run into an infinite loop of failing and trying to add more info, but failing again. This only really works if we do it once at the beginning. I'm not really sure if there's a way to rework this without looping forever

lpil: Ah I see, thank you. Given these are non-trivial functions can we give them better names?

GearsDatapacks:
Alright, I've addressed all of the points you raised.

lpil (approved):
Thank you!!!

PR #3563, opened 2024-08-25
Default to latest major if no version constraint given.

### LINKED ISSUE #3482: Gleam package manager seems to default to oldest possible version instead of most recent
I've been trying to pick up Gleam, part of that was going through this tutorial for a simple web app: https://gleaming.dev/articles/building-your-first-gleam-web-app/ 

I got stuck on the second step which was to run the line:
`gleam add wisp lustre mist gleam_http gleam_erlang dot_env gleam_json`

This adds the listed dependencies to the gleam.toml, with the widest version requirement range possible: 
```
gleam_stdlib = ">= 0.34.0 and < 2.0.0"
wisp = ">= 0.1.0 and < 1.0.0"
lustre = ">= 4.3.0 and < 5.0.0"
mist = ">= 1.2.0 and < 2.0.0"
gleam_http = ">= 3.6.0 and < 4.0.0"
gleam_erlang = ">= 0.25.0 and < 1.0.0"
dot_env = ">= 1.1.0 and < 2.0.0"
gleam_json = ">= 2.0.0 and < 3.0.0"
```
It then fetches the dependencies and creates a manifest.toml with the solved version numbers. 

I noticed a problem when trying to use the wisp library, and it not having code in it apart from the main function. I figured out that the package manager had defaulted to using the wisp version 0.1.0, with the hello world. I manually changed the requirements to force wisp version 0.16.0 and that fixed my problem. 

Digging deeper, it looks like the package manager calls PubGrub to solve it's version requirements, and PubGrub is suggesting the lowest possible version for each requirement. I looked into this, here's what I think so far. 

On this line: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/dependency.rs#L244 we are calling `choose_package_with_fewest_versions`. The [function docs](https://docs.rs/pubgrub/latest/src/pubgrub/solver.rs.html#272) say that it will greedily choose the first allowed version returned by `list_available_versions`. `list_available_versions` seems to be returning functions in order of oldest to newest. This seems like where it's going wrong. My instinct is that if we can change the sort order in this function it will resolve this.

Alternatively, this may be limited to my machine/runtime. Let me know if this is helpful, I have reproduced this a couple of times already. 

Gleam Version: 1.3.2
OS: Manjaro Linux Kernel: 6.2.16-2-MANJARO

joshi-monster:
I don't like this patch at all, but I guess it kinda solves the problem... 

If you also don't like it feel free to just close it :smile: 

fix #3482

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -121,7 +121,10 @@ pub fn update() -> Result<()> {
 pub fn parse_gleam_add_specifier(package: &str) -> Result<(EcoString, Requirement)> {
     let Some((package, version)) = package.split_once('@') else {
         // Default to the latest version available.
-        return Ok((package.into(), Requirement::hex(">= 0.0.0")));
+        return Ok((
+            package.into(),
+            Requirement::hex(dependency::RANGE_UNCONSTRAINED),
+        ));
     };
 
     // Parse the major and minor from the provided semantic version.
@@ -182,7 +185,7 @@ fn parse_gleam_add_specifier_non_numeric_version() {
 #[test]
 fn parse_gleam_add_specifier_default() {
     let provided = "some_package";
-    let expected = ">= 0.0.0";
+    let expected = dependency::RANGE_UNCONSTRAINED;
     let (package, version) = parse_gleam_add_specifier(provided).unwrap();
     match &version {
         Requirement::Hex { version: v } => {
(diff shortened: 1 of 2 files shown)


lpil (commented):
This helps with the one specific combo of deps you have in mind but it breaks other combinations. It is not overall a better approach and will fail to find a way to add the new package for projects that it should be possible to add to.

If you want to add a specific major version you should use a version specifier, as the default README instructs you to.

joshi-monster:
I think it's a little bit better, since if I as a user didn't specify a version I usually mean "give me the latest"; 
I don't think there is a good way to define "the set of versions that minimizes some total "oldness" score across all packages"

Right now it would give me any possible version, which is either the latest or not what I want / unexpected. I think having to be explicit in the unusual case (_give me an old version, please_) is still slightly better.

But as I said, I hate this change anyways :smile:

lpil:
The command is not "give me the latest", it's "give me the dep". A command that fails to give you the dep because the latest isn't acceptable with your given constraints is broken.

If you want the latest there is a command for exactly that.

joshi-monster:
Sure, that's fair! :slightly_smiling_face:

PR #3564, opened 2024-08-25
cleanup unused import

Frank-III:
similar to what we did with `unused_values` in #3435, it seems that we could remove `unused_imports` from module interface.


lpil (approved):
Love this, thank you!

PR #3565, opened 2024-08-25
Use a persistent Erlang/Elixir build server.

joshi-monster:
For me, just compiling and running the `escript` every time takes over half a second. This change only starts the `escript` once per build, after which the same process is reused, saving that overhead when compiling multiple packages.

I haven't found a good way to fit the `BeamCompiler` trait into the IO abstraction, so I just added that one special trait + method instead. The more generic process abstractions I came up with don't really capture the state keeping that's going on. If you know a better place/way to structure things, I'll happily change it!

**Trying to break things**

- When Rust `panic!()`s destructors seem to still get called
- When compiler exits abruptly (I tried `kill` or `process::exit(1)`), the escript receives `eof` from `io:get_line`, after which it exits.
- If the escript exits after reporting completion, it will be restarted. If it exits beforehand, the buffer will be consumed after which we return an error; this causes the `BeamCompiler` to be dropped and the (potential) zombie process to be killed.
- The escript is also a child process of the compiler, so the OS should kill it as well if something goes wrong.


compiler-core/src/build/beam_compiler.rs
```diff
@@ -0,0 +1,144 @@
...
+        let escript_path = out
+            .join(paths::ARTEFACT_DIRECTORY_NAME)
+            .join("gleam@@compile.erl");
+
+        let escript_source = std::include_str!("../../templates/gleam@@compile.erl");
+        io.write(&escript_path, escript_source)?;
+
+        tracing::trace!(escript_path=?escript_path, "spawn_beam_compiler");
+
+        let mut process = std::process::Command::new("escript")
```
lpil, line 111: Core never performs IO, this would need to be injected in from CLI and not be provided at all for wasm.

compiler-core/templates/gleam@@compile.erl
```diff
@@ -5,7 +5,18 @@
...
 
-main(Args) ->
+main(_) -> main().
+
+main() ->
+    case io:get_line("") of
+        eof -> ok;
+        Line ->
+            Args = string:split(string:trim(Line), [31], all), % Unit Separator
+            erlang:display(run(Args)),
```
lpil, line 15: Debugging left in here?

compiler-core/templates/gleam@@compile.erl
```diff
@@ -5,7 +5,18 @@
...
 -record(arguments, {lib = "./", out = "./", modules = []}).
 
-main(Args) ->
+main(_) -> main().
+
+main() ->
+    case io:get_line("") of
+        eof -> ok;
+        Line ->
+            Args = string:split(string:trim(Line), [31], all), % Unit Separator
```
lpil, line 14: What's `[31]`? Can it be written more clearly?

compiler-core/templates/gleam@@compile.erl
```diff
@@ -5,7 +5,18 @@
 
 -record(arguments, {lib = "./", out = "./", modules = []}).
 
-main(Args) ->
+main(_) -> main().
+
+main() ->
```
lpil, line 10: How about a more clear name such as "compile_package_loop"? I didn't realise this ran more than once at first as it's called "main"

lpil (commented):
Nice! I've left some things inline. Would you mind updating the changelog please 🙏

compiler-core/templates/gleam@@compile.erl
joshi-monster: No; this is the output the build tool waits for to know that the erlang process has finished.
I've made it more explicit by using `io:put_chars` instead.

joshi-monster:
done! 🙂 

I've noticed something weird going on with the `kill` in my drop implementation - sometimes the last BEAM file would not be written, or there would instead exist a `.bea#` file... I'm partially guessing here, but I _think_ that's because the Erlang compiler uses the `delayed_write` option.

Instead I now close the stdin instead of forcefully killing the process, which causes the build script to exit normally, hopefully flushing any pending writes first. From a quick Google there does not seem to be a good way to manually wait for all pending writes (https://github.com/erlang/otp/issues/5220)

lpil (approved):
Wow! Thank you!! I'll test this for a bit before merging

lpil:
Ah! Looks like it doesn't work on Windows!

lpil (changes requested):
Ah! Looks like it's broken on Windows

joshi-monster:
I literally tried everything, and `read_line` really just eats control characters on windows :'(

So I went ahead and changed the argument handling to instead format an Erlang term, using the same simple path escape that you used in the 1.5.1 hotfix.

lpil (approved):
Fantastic!! Thank you!! The performance improvement is really noticeable

PR #3566, opened 2024-08-27
Print correct missing patterns for inexhaustive case expressions with multiple subjects

### LINKED ISSUE #2426: Inexhaustive warning message is incorrect for case expressions with multiple subjects

GearsDatapacks:
Fixes #2426.
This ensures that a pattern is printed for every subject of the expression, in the correct order.


lpil (approved):
Super impressive work!!!

PR #3571, opened 2024-08-27
Fixed a bug when importing modules from pattern constructors

GearsDatapacks:
When #3418 was implemented, it looks like I overlooked a case:
```gleam
case option.None {
  option.Some(_) -> Nil
  // ^ Use "import module" code action here
  option.None -> Nil
}
```
Would become:
```gleam
import gleam/option

case option.None {
  option -> Nil
  option.None -> Nil
}
```

That is now fixed.


CHANGELOG.md
```diff
@@ -306,6 +306,10 @@
   inexhaustive case expressions matching on more than one subject.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where the language server would remove the name of a constructor
+  pattern when importing an unknown module via a code action.
+  ([Surya Rose](https://github.com/GearsDatapacks))
```
lpil, line 311: I think this code action wasn't released yet, so we can skip this changelog entry

lpil (approved):
Thank you!!

CHANGELOG.md
GearsDatapacks: Ah right, yes. That's still unreleased. I'm away at the moment so I'll have to wait until I get back to remove this and fix merge conflicts

GearsDatapacks:
I have now removed the changelog entry.

lpil (approved):
Champ! Thank you

Issue #3567, opened 2024-08-27
Improve error message for using `@external` with unknown target


tcoopman:
Gleam version: 1.4.1

This code

```gleam
@external(erl, "decode", "index")
```

results in this error:
![image](https://github.com/user-attachments/assets/ccbb0ca9-b0a9-4e24-a28d-fe6f82b7a4a8)

Instead of saying that the attribute is not recognized, valid values for the external should be shown.

lpil:
Thank you

PR #3576, opened 2024-08-29
fix: better external target error

Frank-III:
resolve: #3567


lpil (commented):
Lovely, thank you. Could you add a snapshot test for this please.

lpil (approved):
Fabulous, thank you!

PR #3591, opened 2024-09-05
improve error title for unknown module value

### LINKED ISSUE #3534: Improve "Unknown module field" error message
I think the error message we get when trying to access a module function that doesn't exist could be improved. Right now it looks like this:

Screenshot 2024-08-19 alle 20 13 49

Calling it "module field" strikes me as a bit odd, I've never heard this term before. Maybe it could be worded a bit differently:

```diff
+error: Unknown module function
-error: Unknown module field
   ┌─ /Users/giacomocavalieri/Documents/progetti/ssg/build/packages/jot/src/jot.gleam:20:11
   │
20 │       dict.update(attributes, key, fn(previous) {
   │           ^^^^^^^

+The module `gleam/dict` does not have a `update` function.
-The module `gleam/dict` does not have a `update` field.
```

giacomocavalieri:
This closes #3534, as suggested by Louis I went with "unknown module value" since it's also how it is referred to in the error message itself instead of "module field"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -233,8 +233,11 @@
   Whereas before, it would suggest `_` as the only missing pattern.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
-- Improve error message for using @external with unknown target
+- Improve error message for using `@external` with unknown target
   ([Jiangda Wang](https://github.com/frank-iii))
+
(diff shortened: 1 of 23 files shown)


lpil (approved):
Thank you

PR #3599, opened 2024-09-08
Display qualified type names on LSP hover

### LINKED ISSUE #2993: Display qualified type names on LSP hover

GearsDatapacks:
Closes #2993 
This uses the `Printer` implemented in #3007 to print correctly qualified or aliased type names when hovering with the language server. I've had to slightly rework the printer so it doesn't take a mutable reference to `TypeNames`, because that wasn't really feasible in the context of the language server.

**TODO**
<details>
<summary>Done</summary>

- [x] Fix the case where a type alias's generics don't map one-to-one with the aliased type, for example:
  ```gleam
  type LocalResult = Result(LocalType, LocalError)
  let a: LocalResult = local_fallible_function()
  //  ^ hovering here shows `LocalResult(LocalType, LocalError)`
  ```

- [x] Write tests for these new cases
- [x] Merge the `TypeNames` struct with the `ValueNames` struct for pattern-printing, as they have very similar semantics and functionality
- [x] Store information about type variables in the `TypeNames` struct when they are created, so we can print these properly
- [x] Correctly qualify prelude types when shadowed
- [x] When importing a type alias, treat is as a local alias rather than a local custom type
- [x] Don't alias types when hovering over the annotation for an alias itself:
```gleam
type Number = Int
//            ^ Hovering here shows `Number`, because we've aliased it, but it should probably print `Int`
```
</details>

**Questions**
1. Currently, this change is for hover and document symbols, which are the two things which use the pretty printer in `engine.rs`. There's also signature help, which could probably benefit from this. Any other places we would want this? Maybe in the little hint for value completions?
2. How should aliases other than the trivial case of `type A = B` be handled? Should we try to print alias names for complex types like `#(fn(Int) -> Float, Result(String, BitArray))`, or just leave it as is?

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1181,27 +1181,9 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             hydrator.disallow_new_type_variables();
             let type_ = hydrator.type_from_ast(resolved_type, environment, &mut self.problems)?;
 
-            if let Type::Named {
-                module,
-                name: type_name,
-                args,
-                ..
-            } = type_.as_ref()
-            {
-                // If the alias or type itself have any parameters, it requires a
-                // lot of work to reconstruct and correctly print the alias,
-                // so we just don't register it.
-                //
-                // Maybe in future we could do this extra work, but for the time
-                // being, it's simpler not to.
-                if parameters.is_empty() && args.is_empty() {
-                    environment.names.named_type_in_scope(
-                        module.clone(),
-                        type_name.clone(),
-                        name.clone(),
-                    );
-                }
-            }
+            environment
+                .names
+                .type_in_scope(name.clone(), type_.as_ref());
 
             // Insert the alias so that it can be used by other code.
             environment.insert_type_constructor(
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1200,14 +1200,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                         type_name.clone(),
                         name.clone(),
                     );
-                } else {
-                    // We need to make sure we register the type as existing, even if we don't define
-                    // the underlying type for printing, because these types do still affect printing,
-                    // for shadowing prelude types, even if they aren't printed at all.
-                    environment.names.type_exists_in_scope(name);
                 }
-            } else {
-                environment.names.type_exists_in_scope(name);
             }
 
             // Insert the alias so that it can be used by other code.
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1184,14 +1184,23 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             if let Type::Named {
                 module,
                 name: type_name,
+                args,
                 ..
             } = type_.as_ref()
             {
-                environment.names.named_type_in_scope(
-                    module.clone(),
-                    type_name.clone(),
-                    name.clone(),
-                );
+                // If the alias or type itself have any parameters, it requires a
+                // lot of work to reconstruct and correctly print the alias,
+                // so we just don't register it.
+                //
+                // Maybe in future we could do this extra work, but for the time
+                // being, it's simpler not to.
+                if parameters.is_empty() && args.is_empty() {
+                    environment.names.named_type_in_scope(
+                        module.clone(),
+                        type_name.clone(),
+                        name.clone(),
+                    );
+                }
             }
 
             // Insert the alias so that it can be used by other code.
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1200,7 +1200,14 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                         type_name.clone(),
                         name.clone(),
                     );
+                } else {
+                    // We need to make sure we register the type as existing, even if we don't define
+                    // the underlying type for printing, because these types do still affect printing,
+                    // for shadowing prelude types, even if they aren't printed at all.
+                    environment.names.type_exists_in_scope(name);
                 }
+            } else {
+                environment.names.type_exists_in_scope(name);
             }
 
             // Insert the alias so that it can be used by other code.
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -335,7 +335,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 warnings,
                 minimum_required_version: self.minimum_required_version,
             },
-            extra: type_names,
+            names: type_names,
         };
 
         match Vec1::try_from_vec(self.problems.take_errors()) {
@@ -1181,17 +1181,17 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             hydrator.disallow_new_type_variables();
             let type_ = hydrator.type_from_ast(resolved_type, environment, &mut self.problems)?;
 
-            match type_.as_ref() {
-                Type::Named {
-                    module,
-                    name: type_name,
-                    ..
-                } => environment.type_names.named_type_in_scope(
+            if let Type::Named {
+                module,
+                name: type_name,
+                ..
+            } = type_.as_ref()
+            {
+                environment.type_names.named_type_in_scope(
                     module.clone(),
                     type_name.clone(),
                     name.clone(),
-                ),
-                _ => {}
+                );
             }
 
             // Insert the alias so that it can be used by other code.
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -298,6 +298,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             module_types_constructors: types_constructors,
             module_values: values,
             accessors,
+            type_names,
             ..
         } = env;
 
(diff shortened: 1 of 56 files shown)


compiler-core/src/type_/printer.rs
```diff
@@ -135,63 +202,82 @@ impl TypeNames {
...
+            return NameQualifier::Qualified(module, name.as_str());
+        };
+
+        NameQualifier::Unimported(name.as_str())
     }
 }
 
 #[derive(Debug)]
-pub enum NamedTypeNames<'a> {
+pub enum NameQualifier<'a> {
```
lpil, line 280: I'm not sure this name quite fits, only one of the 3 variants has a qualifier

compiler-core/src/type_/printer.rs
```diff
@@ -49,6 +46,11 @@ pub struct TypeNames {
     ///
     local_types: HashMap<(EcoString, EcoString), EcoString>,
 
+    /// Types which exist in the prelude, and haven't been shadowed by a local type.
+    /// These are a special case, because they are unqualified by default, but can be
+    /// shadowed and then must be qualified.
+    unshadowed_prelude_types: HashSet<EcoString>,
```
lpil, line 52: Why do prelude types get a special case instead of being inserted into the scope like any other?

compiler-core/src/type_/printer.rs
```diff
@@ -92,18 +94,67 @@ pub struct TypeNames {
...
+    pub fn new() -> Self {
         Self {
-            uid: Default::default(),
-            current_module,
             local_types: Default::default(),
+            unshadowed_prelude_types: Default::default(),
             imported_modules: Default::default(),
             type_variables: Default::default(),
-            type_variable_names: Default::default(),
+            local_constructors: Default::default(),
```
lpil, line 156: I think this is value constructors rather than type constructors so let's call them `local_value_constructors`

compiler-core/src/type_/printer.rs
```diff
@@ -200,35 +286,70 @@ pub enum NamedTypeNames<'a> {
     Qualified(&'a str, &'a str),
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum PrintMode {
+    /// Prints the context-specific representation of a type
+    Normal,
+    /// Prints the raw type, always qualified. Useful for providing
+    /// more detail to the user.
+    Raw,
```
lpil, line 295: I'm not sure this documentation described how this is actually used, and the name isn't super clear either. I think it's whether to show aliases?

Normal and ExpandAliases ?

compiler-core/src/analyse.rs
```diff
@@ -1102,6 +1110,35 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
+            if let Type::Named {
+                module,
+                name: type_name,
+                args,
+                ..
+            } = type_.as_ref()
+            {
+                // If the alias or type itself have any parameters, it requires a
+                // lot of work to reconstruct and correctly print the alias,
+                // so we just don't register it.
```
lpil, line 1122: What are the implications of this?

compiler-core/src/analyse.rs
```diff
@@ -1102,6 +1110,35 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
+                if parameters.is_empty() && args.is_empty() {
+                    environment.names.named_type_in_scope(
+                        module.clone(),
+                        type_name.clone(),
+                        name.clone(),
+                    );
+                } else {
+                    // We need to make sure we register the type as existing, even if we don't define
+                    // the underlying type for printing, because these types do still affect printing,
+                    // for shadowing prelude types, even if they aren't printed at all.
```
lpil, line 1135: Sorry, what does this mean? How does it affect printing?

compiler-core/src/analyse/imports.rs
```diff
@@ -248,8 +259,12 @@ impl<'context, 'problems> Importer<'context, 'problems> {
...
 
             self.environment
-                .value_names
-                .imported_module(import.module.clone(), used_name)
+                .names
+                .imported_module(import.module.clone(), used_name.clone());
+
+            self.environment
+                .names
+                .imported_module(import.module.clone(), used_name);
```
lpil, line 267: Is this intentionally here twice?

compiler-core/src/ast.rs
```diff
@@ -36,16 +37,17 @@ pub trait HasLocation {
...
 
-pub type TypedModule = Module<type_::ModuleInterface, TypedDefinition>;
+pub type TypedModule = Module<type_::ModuleInterface, TypedDefinition, Names>;
 
-pub type UntypedModule = Module<(), TargetedDefinition>;
+pub type UntypedModule = Module<(), TargetedDefinition, ()>;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct Module<Info, Statements> {
+pub struct Module<Info, Statements, Names> {
```
lpil, line 45: I think we could just have an empty one to avoid the extra type parameter

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__hover__hover_contextual_type_annotation_aliased_module.snap
```diff
@@ -0,0 +1,18 @@
+---
+source: compiler-core/src/language_server/tests/hover.rs
+expression: "\nimport wibble/wobble as wubble\n\nfn main(wibble: wubble.Wibble) {\n  wibble\n}\n"
+---
+import wibble/wobble as wubble
+
+fn main(wibble: wubble.Wibble) {
```
lpil, line 7: Very nice

lpil (commented):
Super nice work!!! I've left a bunch of stylistic notes and questions

compiler-core/src/type_/printer.rs
GearsDatapacks: Because unlike other types, types in the prelude can be shadowed, and need to be printed differently. The `local_types` map stores a map from module + type names to their local alias. This means that if we have `("gleam", "Int")` map to `"Int"`, but also `("some_mod", "Int")` map to `"Int"`, both will be printed as `Int`. I can do this in a different way if you would like; this is just the first way that came to mind.

This is different from the value equivalent of this, since any value can be shadowed, not just prelude values. But I could implement a similar system as there.

compiler-core/src/analyse.rs
GearsDatapacks: If the user makes a non-trivial type alias like:
```gleam
type MyResult(a) = Result(a, MyError)
```
We don't print `MyResult(Int)`, for example, when a value has that type, and instead expand the alias (`Result(Int, MyError)`).
Since alias information is lost, we would have to match the `MyError` type with the error type of said value, and place the ok type into the type parameter of `MyResult`. 
This is possible, but requires storing a lot of extra information in the `Names` struct, and doing more computations when printing. So I decided to leave it out for now.

compiler-core/src/analyse.rs
GearsDatapacks: This has to do with the way prelude types are handled, which are discussed above. I'll wait until further feedback on that to elaborate, because this may become redundant.

compiler-core/src/analyse/imports.rs
GearsDatapacks: No, that must have been left over from the merger of `TypeNames` and `ValueNames`

compiler-core/src/type_/printer.rs
lpil: Why have the special case for the prelude at all? Could they be in `local_types` by default?

GearsDatapacks: They can, but unlike other types, prelude types can be shadowed, meaning the types `gleam.Result` and `current_module.Result` would be printed as the same. We need some way to distinguish those.

lpil: Why would they be printed the same? If prelude result has been overridden with the locally defined result then the code would print the prelude result as `gleam.Result` and the local one as `Result`, wouldn't it?

GearsDatapacks: Well, they won't be printed the same in the final result. But if we just stick prelude types in `local_types`, there's nothing stopping them from being printed the same. That's the entire point of this piece of code. We need some way to keep track of which prelude types have been shadowed, so that we can print them differently.

GearsDatapacks: Ok I'll remove the special case and we'll see

GearsDatapacks:
I have removed the special case for prelude types and as you can see, the tests are failing. Hopefully that will demonstrate the problem, because I can't seem to explain it properly with words

GearsDatapacks:
As for solutions to this, I went for an explicit special case for prelude types, because unlike values, which can be shadowed anywhere, types (except for prelude types) cannot.
Perhaps the best way is to do it in the same way as the values track them (see [here](https://github.com/gleam-lang/gleam/blob/5d4847e2d0ee8ee6bd078f4f73316017fcde929e/compiler-core/src/exhaustiveness/printer.rs#L82)). This would remove the special case for prelude types, although it would store some redundant information since most types cannot be shadowed at all.

lpil (approved):
Ah! I see now, thanks, that was is helpful. I've added eviction on shadowing which covers the prelude without a special case for that one module.

Really excellent work here! Thanks again!

Issue #1562, opened 2022-03-27
Permit FFI modules in subdirectories


lpil:
i.e. `src/one/two/three.js` and `src/one/two/three.erl`.

We will need to track Erlang module names to ensure there are no clashes. For example `src/one/themodule.erl` and `src/two/themodule.erl` in one project is an error.

- [ ] Erlang
- [ ] Elixir
- [ ] JavaScript

hayleigh-dot-dev:
Adding some extra context at @erikareads (on discord)'s request:

You can only write ffi code directly inside src/, but we'd find it a lot more useful if i could co-locate ffi files along with the modules that need it:

```
src/
 wibble/
  wobble.gleam
  wobble_ffi.mjs
```

It'd mean inside wobble.gleam we could write
```gleam
@external(javascript, "./wobble_ffi.mjs", "thing")
```
instead of
```gleam
@external(javascript, "../wobble_ffi.mjs", "thing")
```
and so on.

---

For a concrete use-case [lustre](https://github.com/lustre-labs/lustre) has a `lustre/try.gleam` module that spins up a http server for either erlang or js and there is some ffi code to do that in `http_ffi.erl` and `http.ffi.mjs` accordingly. That ffi code is not relevant to the rest of the codebase so keeping it in the project root places undue importance on the code and also makes it minorly annoying to import from.

As lustre gains more CLI commands more of these one-off FFI modules are likely to be necessary so it'd be really great if they could be organised away in a way that makes more sense than how it is currently.

erikareads:
For Javascript this seems to be the path creation function:

https://github.com/gleam-lang/gleam/blob/4396f9a38a2439d5c7961f952adeec93bd9db624/compiler-core/src/javascript.rs#L324-L341

It currently assumes that it will be in the top of the project.

erikareads:
There seems to be a similar (identical?) function in the typescript.rs file:

https://github.com/gleam-lang/gleam/blob/4396f9a38a2439d5c7961f952adeec93bd9db624/compiler-core/src/javascript/typescript.rs#L279-L297

winterqt:
Has anyone tried to take a stab at implementing this yet? If not, I'd like to give it a shot, I just don't want to step on anyone's toes.

isaacharrisholt:
For the Erlang name clashing, is there a reason you couldn't use an absolute name, like the compiled Gleam does anyway? e.g.

```text
src/
 wibble/
  wobble.gleam
[...]

PR #3601, opened 2024-09-08
Permit FFI files in subdirectories

PgBiel:
Closes https://github.com/gleam-lang/gleam/issues/1562

~~Still WIP (finishing the details regarding the duplicate module check)~~

Changes:
- FFI files in subdirectories are now copied. For JavaScript specifically, they can be used as `@external(javascript, "./file.mjs", "function")`. For Erlang, you'd use the module's name as usual.
- An additional check was included to ensure no `.erl` files with the same name exist within the project, as Erlang cannot have duplicate modules.
- Added support for directories to the in-memory filesystem. (Otherwise, the `is_directory` function would return `true` for files and `read_dir` would also list the directory itself, causing tests to enter infinite recursion and crash.)

Potential improvements:
- [x] I'd like to add explicit integration tests for using `@external` with relative paths in JS (I tested locally with `.mjs` with node.js, Deno and Bun and it worked, but could be worth it to test other potential edge cases). Let me know where would be the best spot in the codebase to do so.
- [x] Consider having an error when a native file would overlap with a `.gleam` file's generated code (`.mjs` or `.erl`).


compiler-core/src/io/memory.rs
```diff
@@ -96,7 +115,49 @@ impl FileSystemWriter for InMemoryFileSystem {
...
+    fn mkdir(&self, path: &Utf8Path) -> Result<(), Error> {
+        let mut parent_counter = 0;
+
+        // Traverse ancestors from parent to root
+        // Create each missing ancestor
+        for ancestor in path.ancestors() {
+            if ancestor == "" || ancestor.ends_with(".") {
+                continue;
+            }
+            if ancestor.ends_with("..") {
```
lpil, line 127: What's this for? I thought we always used absolute paths.

compiler-core/src/io/memory.rs
```diff
@@ -236,6 +323,30 @@ impl FileSystemReader for InMemoryFileSystem {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum InMemoryFileRepr {
```
lpil, line 327: What's a repr? We don't abbreviate in this codebase so a full name should be used.

compiler-core/src/io/memory.rs
```diff
@@ -247,30 +358,45 @@ impl FileSystemReader for InMemoryFileSystem {
 //
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InMemoryFile {
-    buffer: Rc<RefCell<Vec<u8>>>,
+    repr: InMemoryFileRepr,
```
lpil, line 361: Here too. What's a repr?

lpil (commented):
Thank you!

You'll see there's integration tests in the test directory which can be updated with nested FFI files

compiler-core/src/io/memory.rs
PgBiel: It was just because we're using camino which handles `..` and `.` differently, so I wanted to ensure correctness, but I guess this isn't respected elsewhere either, so I'll remove it

compiler-core/src/io/memory.rs
PgBiel: Okay! I meant "representation" here so I'll probably just change to that, though I admit this might not be the best name either way; I'm open to other ideas.

PgBiel:
Problem: My handmade recursive directory walker is too naive and can enter infinite loops. I'll probably add a new method to `FileSystemReader` where we can address this at the IO boundary, similarly to how it's done for `gleam_source_files`.

PgBiel:
Alright, I've made the implementation more robust (it also now supports copying nested Erlang modules), added detection of conflicting `.gleam` and `.mjs` files (but not of conflicting `.gleam` and `.erl` files yet due to the issue outlined here: https://github.com/gleam-lang/gleam/issues/1562#issuecomment-2347939465 ), added detection of conflicting `.erl` files in separate subpaths, added more unit tests and added integration tests. :smile:

inoas:
In case this is desired and-or not yet considered:

Can we enforce some namespacing on Erlang, Elixir and JavaScript modules that maps to the directory structure and else throw a compiler error?

inoas:
> I'm not sure that would be desirable since the main point of FFI is to have full control over what the final transpiled code does. Besides, we would have to parse all Erlang files, though I guess we could do with some regex in this case. But I think it's more a matter of flexibility, and the error just tells you if you ever mess it up somehow.
> 
> The error is also not perfect since I guess it's theoretically possible for you to shadow some module from some other package. But it's an attempt.

maybe it could warn.
I don't know... I think these FFI files will all be new and setting up strictness is nice there. You can still build any erlang/elixir library as a deprendency.

PgBiel:
I think this is fine to be honest... Forcing people to name their modules like `folder@folder2@something` in every folder doesn't sound good or useful realistically. And it can clash either way.

lpil (commented):
Very nice, thank you.

It seems now that there's a concept of directories in the in memory file system and also the OS file system, and each is responsible for implementing traversal. I think it would make sense to remove this duplication and instead to have the trait know nothing about directory walking and then the walking logic depends on the trait and is shared between all implementations.

PgBiel:
> I think it would make sense to remove this duplication and instead to have the trait know nothing about directory walking and then the walking logic depends on the trait and is shared between all implementations.

If I understood correctly, you are proposing making a dir walking function which uses only the trait's `read_dir` function (and other trait functions if needed), right? Though I'm not sure if we would want to reinvent the wheel here, given we already have a package which implements graph traversal logic to avoid loops etc. What if we compromise and simply add a `walk_dir` function to the trait and use that where applicable?

lpil:
Yes, that's right. We should not have two implementations of the same thing, and removing a dependency is beneficial for maintenance. Having one less dependency to verify for the EU Cyber Resilience Act is nice too!

I'm not worried about correctness seeing as this is a trivial algorithm, will be fine.

PgBiel:
By the way, I believe this is very close to completion (the dir walker algorithm was already pushed, but I haven't tested it well yet) - I'll be picking this up again very soon, once I have a bit more time :)

PgBiel:
Ready for another look :smile:

lpil (approved):
Thank you!!

Issue #2170, opened 2023-05-25
Add a command to update indivudual dependencies


NthTensor:
It would be nice to be able to use `gleam deps update <package_1> <package_2>` to update specific packages.

Use case: When local dependencies get merged we will have to update them with `gleam deps update` or similar. Unfortunately, this also updates all the other packages.

lpil:
Agreed! This would be a fab addition. I like the suggested `gleam deps update <package_1> <package_2>` CLI.

SnakeDoc:
I have an interest in seeing this feature implemented. Before I take a swing, are there any "gotchas" I need to be aware of, in regard to how the update process currently works, or perhaps some history behind the way the `update` cli command is currently implemented?

Motivation behind this feature: https://github.com/renovatebot/renovate/pull/31002

SnakeDoc:
Couple thoughts after taking a look under the hood:

1. Gleam is now 1.x, so we should avoid breaking/changing the existing CLI for `gleam update` and `gleam deps update`.
2. We should strive to make the command immune to common scripting "gotchas", such as undefined/unresolved variables.

The previously proposed command was:

    gleam update <package names>
    gleam deps update <package names>

This retains compatibility with the existing 1.x CLI, and allows the original `gleam deps update` command to continue updating all available dependencies. However, this proposal exposes some potential problems for script writers.

Consider the following scenario:

    gleam deps update ${PACKAGES}

Where `${PACKAGES}` is undefined. This will result in the Gleam CLI interpreting this command as-if the user intended to update all available dependencies. Instead, we should throw an error of some kind.

This leads me to propose the following CLI:

    # updates all available deps
    # retained for compatibility, perhaps to be deprecated and eventually removed?
    gleam deps update

    # explicitly updates all available deps
    gleam deps update --all

    # update the list of packages
    gleam deps update --packages <list of packages>

Effectively, `gleam deps update` would become an alias of the more explicit command, `gleam deps update --all`.

Then, `gleam deps update --packages <packages>` would allow users to safely do something like `gleam deps update --packages ${PACKAGES}`. In the event `${PACKAGES}` is undefined or empty, we would throw an error instead of proceeding against the user's intentions.

@lpil Thoughts?

lpil:
Let's go with the API extension NthTensor suggested.

This would be implemented by unlocking the dependencies specified along with any dependencies they bring in.

GearsDatapacks:
Can this issue be closed now that #3602 is merged?

[...]

PR #3602, opened 2024-09-09
Add support for updating specific packages and their dependencies

SnakeDoc:
In reference to #2170 

Add support for updating specific packages and their dependencies via the Gleam CLI.

Updated commands:
```
gleam update <packages>
gleam deps update <packages>
```

Help output:
```
gleam deps update --help
Update dependency packages to their latest versions

Usage: gleam deps update [PACKAGES]...

Arguments:
  [PACKAGES]...  (optional) Names of the packages to update
                 If omitted, all dependencies will be updated

Options:
  -h, --help  Print help
```

The updated commands can be used like:
```
gleam deps update package_a package_b package_c
```

To update all packages, the existing functionality was kept:
```
gleam deps update
```

This is my first time with rust - let me know if I missed anything.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -43,7 +43,7 @@ pub fn list() -> Result<()> {
         &config,
         &cli::Reporter::new(),
         UseManifest::Yes,
-        None,
+        Vec::new(),
     )?;
     list_manifest_packages(std::io::stdout(), manifest)
 }
@@ -118,11 +118,11 @@ pub fn update(packages: Vec<String>) -> Result<()> {
 
     if packages.is_empty() {
         // Update all packages
-        _ = download(&paths, cli::Reporter::new(), None, None, UseManifest::No)?;
+        _ = download(&paths, cli::Reporter::new(), None, Vec::new(), UseManifest::No)?;
     } else {
         // Update specific packages
         let packages_to_update = packages.into_iter().map(EcoString::from).collect();
-        _ = download(&paths, cli::Reporter::new(), None, Some(packages_to_update), UseManifest::Yes)?;
+        _ = download(&paths, cli::Reporter::new(), None, packages_to_update, UseManifest::Yes)?;
     }
 
     Ok(())
@@ -253,7 +253,7 @@ pub fn download<Telem: Telemetry>(
     paths: &ProjectPaths,
     telemetry: Telem,
     new_package: Option<(Vec<(EcoString, Requirement)>, bool)>,
-    packages_to_update: Option<Vec<EcoString>>,
+    packages_to_update: Vec<EcoString>,
     // If true we read the manifest from disc. If not set then we ignore any
     // manifest which will result in the latest versions of the dependency
     // packages being resolved (not the locked ones).
@@ -299,7 +299,7 @@ pub fn download<Telem: Telemetry>(
         &config,
         &telemetry,
         use_manifest,
-        packages_to_update.as_ref(),
+        packages_to_update,
     )?;
     let local = LocalPackages::read_from_disc(paths)?;
 
@@ -615,7 +615,7 @@ fn get_manifest<Telem: Telemetry>(
     config: &PackageConfig,
     telemetry: &Telem,
     use_manifest: UseManifest,
-    packages_to_update: Option<&Vec<EcoString>>,
+    packages_to_update: Vec<EcoString>,
 ) -> Result<(bool, Manifest)> {
     // If there's no manifest (or we have been asked not to use it) then resolve
     // the versions anew
@@ -632,7 +632,7 @@ fn get_manifest<Telem: Telemetry>(
     };
 
     if should_resolve {
-        let manifest = resolve_versions(runtime, mode, paths, config, None, telemetry, None)?;
+        let manifest = resolve_versions(runtime, mode, paths, config, None, telemetry, Vec::new())?;
         return Ok((true, manifest));
     }
 
@@ -640,7 +640,7 @@ fn get_manifest<Telem: Telemetry>(
 
     // If there are no requested updates, and the config is unchanged
     // since the manifest was written then it is up to date so we can return it unmodified.
-    if packages_to_update.is_none() && is_same_requirements(
+    if packages_to_update.is_empty() && is_same_requirements(
         &manifest.requirements,
         &config.all_drect_dependencies()?,
         paths.root(),
@@ -803,16 +803,16 @@ fn resolve_versions<Telem: Telemetry>(
     config: &PackageConfig,
     manifest: Option<&Manifest>,
     telemetry: &Telem,
-    packages_to_update: Option<&Vec<EcoString>>,
+    packages_to_update: Vec<EcoString>,
 ) -> Result<Manifest, Error> {
     telemetry.resolving_package_versions();
     let dependencies = config.dependencies_for(mode)?;
     let mut locked = config.locked(manifest)?;
 
-    if let (Some(packages_to_update), Some(manifest)) = (packages_to_update, manifest) {
-        locked = unlock_packages(packages_to_update, &locked, Some(&manifest.packages))?;
+    if !packages_to_update.is_empty() {
+        let manifest_packages = manifest.map(|m| &m.packages[..]);
+        unlock_packages(&mut locked, &packages_to_update, manifest_packages)?;
     }
-    let locked = locked.clone();
 
     // Packages which are provided directly instead of downloaded from hex
     let mut provided_packages = HashMap::new();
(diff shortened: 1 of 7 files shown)


compiler-cli/src/dependencies.rs
```diff
@@ -603,6 +615,7 @@ fn get_manifest<Telem: Telemetry>(
     config: &PackageConfig,
     telemetry: &Telem,
     use_manifest: UseManifest,
+    packages_to_update: Option<&Vec<EcoString>>,
```
lpil, line 618: Let's use an empty vector as the empty value rather than wrapping it

compiler-cli/src/dependencies.rs
```diff
@@ -790,10 +803,16 @@ fn resolve_versions<Telem: Telemetry>(
     config: &PackageConfig,
     manifest: Option<&Manifest>,
     telemetry: &Telem,
+    packages_to_update: Option<&Vec<EcoString>>,
```
lpil, line 806: I think we can take ownership of the vector as we make a copy to consume it, and it's not used after outside this function

compiler-cli/src/dependencies.rs
```diff
@@ -790,10 +803,16 @@ fn resolve_versions<Telem: Telemetry>(
...
 ) -> Result<Manifest, Error> {
     telemetry.resolving_package_versions();
     let dependencies = config.dependencies_for(mode)?;
-    let locked = config.locked(manifest)?;
+    let mut locked = config.locked(manifest)?;
+
+    if let (Some(packages_to_update), Some(manifest)) = (packages_to_update, manifest) {
+        locked = unlock_packages(packages_to_update, &locked, Some(&manifest.packages))?;
+    }
+    let locked = locked.clone();
```
lpil, line 815: Similar ownership things for this, can pass a mutable reference to `unlock_packages`

compiler-cli/src/dependencies.rs
```diff
@@ -975,6 +994,29 @@ fn provide_package(
...
 
+/// Unlocks specified packages and optionally their dependencies
+/// If `manifest_packages` is provided, the dependencies of the packages are also unlocked
+fn unlock_packages(
+    packages_to_unlock: &[EcoString],
+    locked: &HashMap<EcoString, Version>,
+    manifest_packages: Option<&[ManifestPackage]>,
+) -> Result<HashMap<EcoString, Version>> {
+    let mut new_locked = locked.clone();
+    let mut queue = packages_to_unlock.to_owned();
```
lpil, line 1005: Let's keep the name `packages_to_unlock` as it's a bit clearer

compiler-cli/src/dependencies.rs
```diff
@@ -1415,3 +1457,272 @@ fn verified_requirements_equality_with_canonicalized_paths() {
...
+    ];
+
+    let packages_to_unlock = vec!["nonexistent_package".into()];
+    let result = unlock_packages(&packages_to_unlock, &locked, Some(&manifest_packages)).unwrap();
+
+    assert_eq!(result, locked, "Locked packages should remain unchanged");
+}
+
+#[test]
+fn test_unlock_circular_dependencies() {
```
lpil, line 1575: We don't support circular deps FYI

compiler-cli/src/dependencies.rs
```diff
@@ -1415,3 +1457,272 @@ fn verified_requirements_equality_with_canonicalized_paths() {
...
+                outer_checksum: Base16Checksum(vec![]),
+            },
+        },
+    ];
+
+    let packages_to_unlock: Vec<EcoString> = vec![];
+    let result = unlock_packages(&packages_to_unlock, &locked, Some(&manifest_packages)).unwrap();
+
+    assert_eq!(result, locked, "Locked packages should remain unchanged when no packages are specified to unlock");
+}
```
lpil, line 1728: Could we have a test for packages A, B, and C where A and B both depend on C, and A is being unlocked. B and C should remain locked.

Thank you.

compiler-cli/src/main.rs
```diff
@@ -93,6 +93,14 @@ use clap::{
 };
 use strum::VariantNames;
 
+#[derive(Args, Debug, Clone)]
+struct UpdateOptions {
+    /// (optional) Names of the packages to update
+    /// If omitted, all dependencies will be updated
+    #[arg(verbatim_doc_comment)]
```
lpil, line 100: What does this mean?

lpil (commented):
Lovely!!! Could you update the changelog also?

SnakeDoc:
Thank you, I will make the requested changes and request a new review once completed.

compiler-cli/src/main.rs
SnakeDoc: If you are asking about the `verbatim_doc_comment` - this tells `clap-rs` to use the comment verbatim, ie. respect the newline between the two comment lines. I thought this was better visually, and emphasized the default a bit better than one long line.

A visual:
```
Arguments:
  [PACKAGES]...  (optional) Names of the packages to update
                 If omitted, all dependencies will be updated
```
versus:
```
Arguments:
  [PACKAGES]...  (optional) Names of the packages to update. If omitted, all dependencies will be updated.
```

`clap-rs` docs regarding preprocessing: https://docs.rs/clap/latest/clap/_derive/index.html#pre-processing 

There are two existing uses of `verbatim_doc_comment` in the codebase: https://github.com/search?q=repo%3Agleam-lang%2Fgleam%20verbatim_doc_comment&type=code

I can rewrite the comment if you'd prefer it to be a single line. It's also possible I overlooked a more elegant way to communicate this to users?

compiler-cli/src/dependencies.rs
SnakeDoc: My thoughts here was it might actually be desirable to unlock dependency C in this scenario.

By unlocking dependency C, we allow it to update alongside dependency A, while considering the requirements of both dependencies A and B. This should mean C, as part of A's tree, gets an opportunity to update to it's latest allowed version that satisfies all requirements.

I believe that is what is going on at [`compiler-core/src/dependency.rs#L56`](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/dependency.rs#L56) since dependency C would not be found in `locked` and therefore would be rediscovered via [`compiler-core/src/dependency.rs#L273`](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/dependency.rs#L273).

Without allowing dependency C to update in this scenario, users would have two choices:

1. Run `gleam deps update dependency_c` (repeating for any downstream dependencies that remain locked)
2. Run `gleam deps update` to update everything.

It's also very possible users may not be aware dependency C exists, if it is a transient and only found within `manifest.toml`, meaning they might not be aware it needs updating.

This led me to believe the proposed changes might be in-line with user's expectations for this command. Perhaps if this is undesirable, we can add an optional flag to direct this behavior? Such as:
```
gleam deps update --no-conflicts <package names>
```

It is possible I've overlooked and/or overthought aspects of this issue. I am happy to add the requested test + changes if you think that is the correct approach.

SnakeDoc:
Requested changes added and relevant conversations resolved. I have left open two conversations that might need some clarification.

compiler-cli/src/main.rs
lpil: That's great, thank you for the explanation!

compiler-cli/src/dependencies.rs
lpil: I don't think that `gleam update package1 package2` should impact `package3`.

RE the discovery issue, there's issues for commands to learning about your dependency tree

SnakeDoc: Hmm, one last clarification before I get back to work on this:

Are you wanting `gleam deps update package_a` to only ever unlock/update `package_a`, or is it ok to also unlock/update any of `package_a`'s dependencies so long as no other packages also depend on them?

SnakeDoc: @lpil, I hope you don't mind me following up. I have some time this weekend to work on the changes, but I want to ensure I understand the requirements correctly.

From your last comment, I'm inferring that `gleam deps update <package names>` should never unlock or update packages the user didn't explicitly specify (ie. we would never unlock any package not explicitly named in the command). This would be a significant change from the current implementation of `unlock_packages()` and its associated tests.

To clarify, should we only ever unlock/update the explicitly named packages?

I appreciate your valuable time and guidance on this! Once I have confirmation, I'm ready to implement these changes and update the relevant tests.

lpil: Sorry, I'm not sure what you mean by this one. You can't tell solely from the list of locked packages like that what is to be unlocked, you need to know what the specified dependencies of the root package are.

Any package that is a dependency of the root package _excluding asked to be upgraded_ should be locked, the rest should be unlocked.

SnakeDoc: I have pushed some changes and additional tests that I think satisfy the requirements. Please review and let me know what you think.

lpil:
Thanks for waiting, I'm back from the CodeBEAM conference now. Could you rebase on main to remove the merge commit and so the tests will run?

SnakeDoc:
I hope the conference was a blast!

Merge commit should be gone now.

SnakeDoc:
Thanks for the guidance on this PR! :rocket:

PR #3607, opened 2024-09-10
Warn about unused qualified constructors

giacomocavalieri:
This PR fixes a bug where an unused qualified constructor wouldn't be reported as unused


lpil (approved):
Thank you!

PR #3613, opened 2024-09-11
Allow to jump to definition of external functions defined in the same module

### LINKED ISSUE #3610: LSP: "Go to definition" does not work on external functions declared in the same file
How to reproduce:

1. Create this gleam file and a corresponding `wibble_ffi.mjs`.

```gleam
pub fn main() {
  external_wibble()
}

@external(javascript, "./wibble_ffi.mjs", "wibble")
fn external_wibble() -> Nil
```

2. Hover over `external_wibble()` and use the "go to definition" feature from your editor.
3. Nothing happens.

Expected behavior: the cursor should move to the function declaration of `external_wibble()`.

giacomocavalieri:
This PR closes #3610


compiler-core/schema.capnp
```diff
@@ -156,6 +158,11 @@ struct ValueConstructorVariant {
   }
 }
 
+struct External {
+    module @0 :Text;
+    function @1 :Text;
+}
```
lpil, line 164:
```suggestion
struct External {
  module @0 :Text;
  function @1 :Text;
}
```

compiler-core/src/analyse.rs
```diff
@@ -1285,17 +1286,13 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
-        );
-        let (impl_module, impl_function) = implementation_names(external, &self.module_name, name);
         let variant = ValueConstructorVariant::ModuleFn {
             documentation: documentation.as_ref().map(|(_, doc)| doc.clone()),
-            name: impl_function,
+            name: name.clone(),
             field_map,
-            module: impl_module,
+            external_erlang: external_erlang.clone().map(|(m, f, _)| (m, f)),
+            external_javascript: external_javascript.clone().map(|(m, f, _)| (m, f)),
```
lpil, line 1294: Map then clone. Cloning then mapping results in more being cloned than is needed to be.

compiler-core/src/language_server/engine.rs
```diff
@@ -175,6 +175,11 @@ where
                 None => return Ok(None),
             };
 
+            let a =
+                node.definition_location(this.compiler.project_compiler.get_importable_modules());
+            tracing::info!("{:#?}", node);
+            tracing::info!("{:#?}", a);
```
lpil, line 181: Debugging left in here.

Use the debug API `dbg!` for debugging rather than the logging API `tracing` as then then CI will let you know if you left it in.

compiler-core/src/language_server/tests.rs
```diff
@@ -392,6 +393,34 @@ impl<'a> TestProject<'a> {
...
+                .find(|(name, _)| *name == module_name)
+                .map(|(_, src)| *src)
+        };
+
+        find_module(&self.root_package_modules)
+            .or(find_module(&self.dependency_modules))
+            .or(find_module(&self.test_modules))
+            .or(find_module(&self.hex_modules))
+            .or(find_module(&self.dev_hex_modules))
+            .or(find_module(&self.indirect_hex_modules))
```
lpil, line 421: This isn't lazy so it does way more work than it needs to due to not short circuiting.

What's the difference between all of these? Why isn't it always `root_package_modules` or `dependency_modules`?

lpil (commented):
Thank you! Few small notes

compiler-core/src/language_server/engine.rs
giacomocavalieri: Ah I had no idea 😆 thanks for the heads up!

compiler-core/src/language_server/tests.rs
giacomocavalieri: I'm not sure what the purpose of each one is in the `TestProject` but since one could jump to the definition of something defined in any of those modules I extended the search to each one of those.

I've made it lazy!

lpil: Oh sorry I just realised this is in the test code, I'm a fool. Sorry about that

lpil (approved):
Thank you!!!

Issue #3611, opened 2024-09-11
`gleam fix` fixes warnings that have fix code actions


lpil:
Some warnings have code actions to fix them such as "unused import" which can always be safely applied.

Make `gleam fix` apply all of these edits for people not using the LS at that time.

ankddev:
Came through this very old ticket 😄. This should be done in several stages/PRs, I think:

1. Move quick fixes to their own module/crate
2. Integrate this new system to the LS
3. Integrate this new system to the `gleam fix`
4. I think that it would be good if diagnostics would show that quick fixes are available

I think that it would be better to open separate issues for each

lpil:
Sounds reasonable! I think it would be consistent to move all the refactorings out rather than just the quick fixes. I presume none of them are too hard-coupled to the language server in any particular way.

> I think that it would be good if diagnostics would show that quick fixes are available

In the diagnostics could be a bit verbose and would be not the right recommendation for the LS user, so perhaps it could be a single message at the end of CLI commands if there has been any fixable warnings.

PR #3615, opened 2024-09-12
Fix minimum required version with `gleam fix`

giacomocavalieri:
As we've decided the other day, before drafting a release we wanted `gleam fix` to fix the `gleam.toml`'s `gleam` version constraint. I've implemented the fix in this PR, this doesn't close the issue <https://github.com/gleam-lang/gleam/issues/3611> yet but all the machinery needed for it should already be in place after this gets merged!

### DIFF
--- a/compiler-cli/src/fix.rs
+++ b/compiler-cli/src/fix.rs
@@ -91,5 +91,5 @@ fn minimum_required_version_from_warnings(warnings: Vec<Warning>) -> Option<Vers
             _ => None,
         })
         .reduce(std::cmp::max)
-        .map(|version| version.clone())
+        .cloned()
 }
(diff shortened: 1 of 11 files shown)


compiler-cli/src/fix.rs
```diff
@@ -1,13 +1,57 @@
...
+        warnings.clone(),
+    )?;
+    let warnings = warnings.take();
+
+    let result = fix_minimum_required_version(warnings)?;
+
+    match result {
+        FixResult::FixedSomething => println!("Your Gleam code has been fixed!"),
+        FixResult::NothingToFix => println!("Nothing to fix."),
+    };
```
giacomocavalieri, line 46: Do you think we should be more precise about what has been fixed?

lpil: Yeah good idea!

lpil (approved):
Lovely!!!

PR #3616, opened 2024-09-12
Fix minimum version inference bug

giacomocavalieri:
With this change the version lower bound is `0.1.0` by default.
- If someone is not using any feature that would warrant increasing the lower bound, this won't cause any problems when publishing even if you have a version constraint like `>= 0.x`, and `gleam fix` won't change it.
- However, when no constraint is explicitly added and the compiler adds the inferred one it makes sure that is still `>= 1.x`, even if the inferred one might be lower.


lpil (approved):
Thank you!!!

PR #3620, opened 2024-09-14
Fix nested tuple access formatting

### LINKED ISSUE #3619: Incorrect formatting of nested tuple access
When formatting one of these cases:
```gleam
tuple.0.field
```
```gleam
tuple.0.1
```

It produces:
```gleam
{ tuple.0 }.field
```

Instead, it should leave them as is.

giacomocavalieri:
This PR closes #3619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you

Issue #3614, opened 2024-09-11
Record with function param in guard causes compile error


Neilerino:
Bumped into this case in a project I was working on last night. If I try to use a record which includes a function as a parameter I get an erlang compilation error. Running with latest release of gleam. Here's my basic repro steps:

```
// test_gleam.gleam
import gleam/io
import other_file

pub fn main() {
  let case_var = ["test", "test2"]
  case case_var {
    [head, ..] if head == other_file.case_test.input_string -> {
      io.println("test")
    }
    _ -> {
      io.println("not test")
    }
  }
}
```

```
// other_file.gleam
pub type Test {
  Test(input_string: String, func: fn(Int) -> Int)
}

pub fn add_one(x: Int) -> Int {
  x + 1
}

pub const case_test = Test("test", add_one)

```

`gleam run` output snippet with trace logging enabled:
```
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/packages"
TRACE download_deps: reading_file path="/Users/Neilerino/Programming/test_gleam/gleam.toml"
DEBUG download_deps: reading_manifest_toml
TRACE download_deps: reading_file path="/Users/Neilerino/Programming/test_gleam/manifest.toml"
DEBUG download_deps: manifest_up_to_date
TRACE download_deps: reading_file path="/Users/Neilerino/Programming/test_gleam/build/packages/packages.toml"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/dev/erlang"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/dev/javascript"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/prod/erlang"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/prod/javascript"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/lsp/erlang"
DEBUG download_deps: locking_build_directory path="/Users/Neilerino/Programming/test_gleam/build/lsp/javascript"
TRACE download_deps: writing_file path="/Users/Neilerino/Programming/test_gleam/build/packages/packages.toml"
TRACE reading_file path="/Users/Neilerino/Programming/test_gleam/gleam.toml"
TRACE reading_file path="/Users/Neilerino/Programming/test_gleam/gleam.toml"
[...]

PR #3621, opened 2024-09-15
Added handling to guard logic for const record

Neilerino:
Added compiler logic so that if you're referencing a constant record in a guard clause it compiles the referenced field instead of creating a tuple accessor for that specific field.

I'm very new to rust, gleam, and erlang so please leave any suggestion you might have on the code 

Fixes: https://github.com/gleam-lang/gleam/issues/3614

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- Fix a bug that caused the compiler to crash when accessing a field of a const record, that had a function param in a guard clause.
+  ([Neil Wadden](https://github.com/Neilerino))
+
 ## v1.5.0-rc1 - 2024-09-14
 
 ### Build tool
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -1387,11 +1387,38 @@ fn bare_clause_guard<'a>(guard: &'a TypedClauseGuard, env: &mut Env<'a>) -> Docu
 
         ClauseGuard::FieldAccess {
             container, index, ..
-        } => tuple_index_inline(container, index.expect("Unable to find index") + 1, env),
+        } => handle_guard_field_access(container, index.expect("Unable to find index"), env),
 
         ClauseGuard::ModuleSelect { literal, .. } => const_inline(literal, env),
 
         ClauseGuard::Constant(constant) => const_inline(constant, env),
+    }
+}
+
+fn handle_guard_field_access<'a>(
+    container: &'a TypedClauseGuard,
+    index: u64,
+    env: &mut Env<'a>,
+) -> Document<'a> {
+    /// In Erlang, including function references directly in guard expressions is invalid.
+    /// This function ensures that when we have a field access in a guard, and the container
+    /// is a constant record that may contain function references, we avoid inlining the entire
+    /// record into the guard.
+    ///
+    /// Specifically, if the `container` is a module-selected constant record (`ClauseGuard::ModuleSelect`),
+    /// and the literal is a `TypedConstant::Record`, we extract the specific field value from the record's
+    /// arguments and generate code for that value. This avoids inlining the entire
+    /// record (which may contain function references) into the guard.
+    match container {
+        ClauseGuard::ModuleSelect { literal, .. } => match literal {
+            TypedConstant::Record { args, .. } => {
+                let index = index as usize;
+                let field = args.get(index).expect("Unable to find field");
+                const_inline(&field.value, env)
+            }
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -409,3 +409,6 @@
 - Fix a bug that caused record accessors for private types to not be completed
   by the LSP, even when in the same module.
   ([Ameen Radwan](https://github.com/Acepie))
+
```
GearsDatapacks, line 412: Looks like you added this to the wrong place. It's currently under the v1.4.1 changelog, it should be under `Unreleased` at the top.

compiler-core/src/erlang.rs
```diff
@@ -1375,14 +1375,32 @@ fn bare_clause_guard<'a>(guard: &'a TypedClauseGuard, env: &mut Env<'a>) -> Docu
...
         ClauseGuard::Constant(constant) => const_inline(constant, env),
     }
 }
 
+fn handle_guard_field_access<'a>(
+    container: &'a TypedClauseGuard,
+    index: u64,
+    env: &mut Env<'a>,
+) -> Document<'a> {
+    match container {
```
GearsDatapacks, line 1391: Maybe add a comment here, explaining what you're doing? It might not be clear why this function exists

GearsDatapacks (commented):
You'll need to wait for Louis, the maintainer, to review this before it gets merged, but I'll just leave a couple of notes I came up with here.

compiler-core/src/erlang.rs
Neilerino: I may have gone overboard with the comment. Lemme know if I should tone it back a bit lol

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Bug Fixes
+
+- Fix a bug that caused the compiler to crash when accessing a field of a const record, that had a function param in a guard clause.
+  ([Neil Wadden](https://github.com/Neilerino))
+
```
Neilerino, line 9: I wasn't super sure about this: where another release version was cut I added a new section for `unreleased`, and `bug fixes`. Lemme know if this wasn't correct though, and I'll fix it 👍

compiler-core/src/erlang.rs
GearsDatapacks: Probably the first paragraph is fine on its own

CHANGELOG.md
GearsDatapacks: Yeah that's correct. There's just been a release, so you needed to add that. Thanks!

GearsDatapacks (commented):
Thanks! It looks good so far. The thing is, this only fixes a subset of cases caused by this issue. For example:
```gleam
type Wrapper {
  Wrapper(fn() -> Nil)
}
const wrapper = Wrapper(main)
pub fn main() {
  case wrapper {
    w if w == wrapper -> todo
    _ -> todo
  }
}
```

It would probably be worth fixing all related issues in one go, instead of some specific cases being fixed.

Neilerino:
> Thanks! It looks good so far. The thing is, this only fixes a subset of cases caused by this issue. For example:
> 
> ```gleam
> type Wrapper {
>   Wrapper(fn() -> Nil)
> }
> const wrapper = Wrapper(main)
> pub fn main() {
>   case wrapper {
>     w if w == wrapper -> todo
>     _ -> todo
>   }
> }
> ```
> 
> It would probably be worth fixing all related issues in one go, instead of some specific cases being fixed.

Yeah that makes sense to me! I'll have a crack at this tomorrow after work and see what I can cook up :+1:

Neilerino:
I had a crack at this (it's been a really fun way for me to practice rust!). I'm currently working on a solution that checks if the compiler can determine the boolean evaluation of a guard during compilation. If it can, it compiles the guard to true or false, instead of rendering the tuple for the imported const record. 

This would also compile any un-necessary guards too, which might "technically" be a bit more performant. But that would be a splitting hairs type of performance difference haha.

I'm still picking away at this, so I'm moving this PR to draft, and I'll re-open it once I've got something ready for review 🙂

lpil:

Neilerino:
> Hello! Are you still working on this one? Thanks

@lpil hey! Sorry!! Work got pretty hectic, and I hit a slight wall with the solution I was trying to implement here, where I'm going to need to include a little bit more logic in the const types during parsing. I'll have a few weeks off soon, and I was planning on having another crack at this then. In the meantime though I can close down this PR, and re-open once I have something. Sorry for leaving the stale PR!!!!

lpil:
Thank you 💜

PR #3630, opened 2024-09-18
Properly format lines with big grapheme clusters

### LINKED ISSUE #3627: Formatter wraps lines with big grapheme clusters
The following code snippet gets wrapped onto multiple lines. The string has been taken from the [unicode-segmenter](https://github.com/cometkim/unicode-segmenter/blob/main/test/grapheme.js#L103) javascript library tests and consists of 6 grapheme clusters and 75 codepoints.

```gleam
pub fn main() {
  sw("ZALGO!", [])
}
```

gets formatted as

```gleam
pub fn main() {
  sw(
    "ZALGO!",
    [],
  )
}
```

tested using Gleam 1.4.1 and 1.5.0-rc1 :purple_heart: 

**Edit:** It seems like Github is stripping out most of the codepoints, which I guess is reasonable since stacks of modifiers are used in spam; the link to the source file above points to the line with that string.

giacomocavalieri:
This PR closes #3627
The pretty printer used `str.len()` as the width of a string; but what a human would intuitively think of as the length of a string is its number of grapheme clusters. This would result in some lines looking short but being split nonetheless by the formatter.

### DIFF
--- a/compiler-core/src/pretty.rs
+++ b/compiler-core/src/pretty.rs
@@ -42,79 +42,79 @@ pub trait Documentable<'a> {
 
 impl<'a> Documentable<'a> for char {
     fn to_doc(self) -> Document<'a> {
-        Document::from_string(format!("{self}"))
+        Document::string(format!("{self}"))
     }
 }
 
 impl<'a> Documentable<'a> for String {
     fn to_doc(self) -> Document<'a> {
-        Document::from_string(self)
+        Document::string(self)
     }
 }
 
 impl<'a> Documentable<'a> for &'a str {
     fn to_doc(self) -> Document<'a> {
-        Document::from_str(self)
+        Document::str(self)
     }
 }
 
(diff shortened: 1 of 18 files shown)


compiler-core/src/pretty.rs
```diff
@@ -41,73 +42,79 @@ pub trait Documentable<'a> {
...
     fn to_doc(self) -> Document<'a> {
-        Document::String(format!("{self}"))
+        Document::string(format!("{self}"))
     }
 }
 
 impl<'a> Documentable<'a> for u8 {
     fn to_doc(self) -> Document<'a> {
-        Document::String(format!("{self}"))
+        Document::string(format!("{self}"))
```
lpil, line 117: Could you make all these formats into eco_format please. 🙏  Ideally we remove String support altogether

lpil (approved):
Thank you!!!

PR #3640, opened 2024-09-23
Improve resolution error messages

giacomocavalieri:
This PR tries to improve some messages displayed if version resolution fails.
This only works in a limited case and keeps defaulting to the current error message; solving the problem in the most general way possible seems a really hard problem so I think we could try doing this and keep on improving incrementally as we identify and handle recurring failure reasons.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -897,6 +897,7 @@ dependencies = [
  "tracing",
  "unicode-segmentation",
  "vec1",
+ "version-ranges",
  "xxhash-rust",
 ]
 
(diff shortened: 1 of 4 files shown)


lpil:
Hey hey Jak! Is this branch still wanted?

giacomocavalieri:
Oops I haven't worked on this for quite a while but I'd say yeah it's very much wanted. I'll have to go through the new version resolving library's api to implement this, it might be a challenging piece of work

lpil:
Jak said he will start this from scratch in future!

PR #3647, opened 2024-09-25
Variant inference in case expressions

### LINKED ISSUE #2060: Type narrowing in case expressions
Was wondering if this could be added to the compiler:

```gleam
pub type Being {
  Person(name: String, age: Int, is_happy: Bool)
  // imagine this type having a gazillion fields
  Dog(name: String, age: Int, is_happy: Bool)
}

pub fn have_birthday(being) {
  // It's a birthday, so increment their age and make them happy
  case being {
    Dog(..) as dog -> Dog(..dog, age: dog.age + 1, is_happy: True)
    Person(..) as person ->
      Person(..person, age: person.age + 1, is_happy: True)
  }
}
```

This leads to 

```
[gleam] error: Unsafe record update
   ┌─ ./src/playground.gleam:10:23
   │
10 │     Dog(..) as dog -> Dog(..dog, age: dog.age + 1, is_happy: True)
   │                       ^^^ I can't tell this is always the right constructor

This type has multiple constructors so it cannot be safely updated.
If this value was one of the other variants then the update would be
produce incorrect results.

Consider pattern matching on it with a case expression and then
constructing a new record with its values.
```

Now I already pattern-matched on it and the compiler could know it's a dog.  If the Dog would have lot's of fields then writing out every field becomes tedious.

GearsDatapacks:
Closes #2060
This PR allows the compiler to infer the variant of values when pattern matching. This allows you to perform safe record updates when the variant is known, and access fields which aren't common to all variants


inoas:
@GearsDatapacks wow these two changes coming from you are worth to use nightly all the time ASAP! Thank you so much!

compiler-core/src/type_.rs
```diff
@@ -63,6 +63,7 @@ pub enum Type {
         module: EcoString,
         name: EcoString,
         args: Vec<Arc<Type>>,
+        constructor_index: Option<u16>,
```
lpil, line 66: What's this field? It's not clear to me how a type has a constructor index.

GearsDatapacks: This represents which variant a certain value has when the variant is known, so for example:
```gleam
type Wibble { Wibble Wobble }

fn wibble(w: Wibble) {
  //      ^ We have no idea which variant this is, so constructor_index is None
  case w {
    Wibble as w -> todo
    //        ^ We know that this is the Wibble variant, so constructor_index would be Some(0)
    Wobble as w -> todo
  }
}
```

This perhaps isn't the best way to implement this, but it's the way I came up with that made fairly minimal changes to the way that the type system is represented internally. If you have another idea, I'm open to suggestions

lpil: I don't have any particular suggestions for the design, I think this is quite clever.

I'm wondering if there's a clearer name that we can use here as it wasn't obvious to me what this means. At the least it needs a good deal of documentation.

compiler-core/src/type_.rs
```diff
@@ -1142,6 +1195,7 @@ pub fn generalise(t: Arc<Type>) -> Arc<Type> {
                 package: package.clone(),
                 name: name.clone(),
                 args,
+                constructor_index: *constructor_index,
```
lpil, line 1198: I think this may be a bug? Shouldn't generalising a type discard what variant it is?

GearsDatapacks: Oops, I must have missed that one

GearsDatapacks:
Hopefully the functionality of the `constructor_index` field should be clearer now

compiler-core/src/type_/tests/errors.rs
```diff
@@ -2434,8 +2434,8 @@ pub type Wibble {
   Wobble(not_field: String, field: Int)
 }
 
-pub fn main() {
-  Wibble(1).field
+pub fn main(wibble: Wibble) {
+  wibble.field
```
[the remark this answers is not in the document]
GearsDatapacks: See above

compiler-core/src/type_/tests/errors.rs
```diff
@@ -2471,3 +2471,38 @@ pub fn main() {
...
+
+pub fn b_to_a(value: MyRecord) {
+  case value {
+    A(..) -> value
+    B(..) as b -> A(..b, other: "Hi")
+  }
+}
+"#
+    );
+}
```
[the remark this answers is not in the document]
GearsDatapacks: I should be able to make that compile, although it doesn't currently. Either way, I'll add a test later

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__record_update_unknown_variant.snap
```diff
@@ -0,0 +1,16 @@
...
+  │
+8 │   Wibble(..value, wubble: True)
+  │   ^^^^^^ I can't tell this is always the right constructor
+
+This type has multiple constructors so it cannot be safely updated. If
+this value was one of the other variants then the update would be produce
+incorrect results.
+
+Consider pattern matching on it with a case expression and then
+constructing a new record with its values.
```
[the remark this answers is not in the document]
GearsDatapacks: This error message hasn't changed in this PR; it's always been like that, but now is probably a good time to improve it, thanks!

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__record_update_wrong_variant.snap
```diff
@@ -0,0 +1,14 @@
...
+error: Mismatched record update
+   ┌─ /src/one/two.gleam:10:19
+   │
+10 │     B(..) as b -> A(..b, other: "Hi")
+   │                   ^ This is never the right constructor
+
+This value is never the same constructor as the one being updated.
+
+If you want to change one variant of a type into another, each field of the
+new type must be specified explicitly.
```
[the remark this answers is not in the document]
GearsDatapacks: Good idea!

GearsDatapacks:
Thanks for the review Jak! I'll get to those points as soon as I have time

giacomocavalieri:
Playing around with it I found two small bugs.
```gleam
pub fn main() {
  let a = todo

  case a {
    Wibble(..) -> Wibble(..a, b: 1)
//                ^^^^^^^^^^^^^^^^^ No complaints here
    Wobble(..) -> Wobble(..a, c: 1)
//                ^^^^^^ Mismatched record update
  }
}
```
I'd expect the compiler to complain on both, instead here it looks like it's narrowing the `todo` to be a `Wibble`.

And the second one is:
```gleam
pub type Wibble {
  Wibble(a: Int, b: Int)
  Wobble(a: Int, c: String)
}

pub fn main() {
  let a = case todo {
    Wibble(..) as b -> Wibble(..b, b: 1)
    Wobble(..) as b -> Wobble(..b, c: "a")
  }

  a.b
//^^^ This shouldn't compile but it does
}
```

lpil:
@GearsDatapacks Note I renamed some stuff to make naming a bit more consistent as `main` currently is inconsistent

GearsDatapacks:
I've addressed all the points in your review now. This now compiles:
```gleam
case some_wibble {
  Wibble(..) -> some_wibble.field_only_present_on_wibble
  _ -> todo
}
```

Both bugs you found are now also fixed.
I just made it so whenever a type with a narrowed variant is unified with an unbound type variable, it discards the narrowed variant. That seemed to solve all the potentially problems that I could think of. It does very slightly reduce the amount which we can infer about narrowed type variants, but it was the only way I could think of to fix these problems without making big changes to the unification process.

GearsDatapacks:
I also have a question about an error message which could potentially be confusing to the user. For example:
```gleam
type Wibble {
  Wibble(a: Int, b: Int, c: Int)
  Wobble(a: Int, d: Int)
}

pub fn a(wibble: Wibble) {
  wibble.wobble // Error: The type `Wibble` only has the `a` field
}

pub fn b() {
  let wibble = Wibble(a: 1, b: 2, c: 3)
  wibble.wobble // Error: The type `Wibble` only has the fields `a`, `b` and `c`
}
```

In the second case the compiler has inferred that `wibble` is of the `Wibble` variant, so it says that it has all three of those fields. However, this extra information isn't reflected anywhere in the error message. Should we print that somehow?

lpil:
Nice work!

I've found a bug where the narrowing is escaping out of the case expression. This invalid code compiles and crashes at runtime

```gleam
import gleam/io

pub type Thingy {
  A(a: Int)
  B(x: Int, b: Int)
}

pub fn main() {
  A(1) |> fun |> io.debug
}

fn fun(x) {
  case x {
    A(..) -> x.a
    B(..) -> x.b
  }
  x.b
}
```

GearsDatapacks:
Interesting, I wonder why that's happening. I'll look into it, thanks!

GearsDatapacks:
Found the issue. It was to do with `TypeVar::Link`, which allows shared mutability with `Arc<RefCell<Type>>`. This meant that when I was changing the `narrowed_variant` field of a type in the current scope, it was also mutating the type in the parent scope as they were linked together. To prevent this, I just called `collapse_links` when creating the type in the inner scope.

lpil:
Hello! I'm back from CodeBEAM now.

> It was to do with TypeVar::Link, which allows shared mutability with Arc<RefCell<Type>>.

This is intentional, the mutability of types is used to propagate inferred type information to linked type variables. We'll need to carefully test this to ensure that types are still inferred correctly.

I'll try and review this soon.

CHANGELOG.md
```diff
@@ -109,6 +109,25 @@
   the Erlang target if any of their arguments were discarded.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler now narrows the variant of custom types when pattern matching,
+  allowing for safe record updates if the variant is known:
```
giacomocavalieri, line 113: Talking with Louis and Hayleigh at code beam we agreed that we should be really clear in framing this as "safe record updates" instead of _narrowing_ of any type to avoid misleading people into thinking this is type narrowing. This could be reworded as:
```suggestion
- The compiler can now allow you to perform record updates for specific
  variants of a type after pattern matching on it:
```

GearsDatapacks: Sounds good!

GearsDatapacks:
I've added a couple of tests to make sure this change doesn't break type inference, but I honestly couldn't think of that many cases to test. Anyone who is more familiar with the type inference system, please suggest more cases if you come up with any

compiler-core/src/error.rs
```diff
@@ -1881,31 +1882,57 @@ assigned variables to all of them."
...
-                            },
-                            path: path.clone(),
-                            src: src.clone(),
-                            extra_labels: vec![],
-                        }),
+                TypeError::UnsafeRecordUpdate { location, reason } =>
+                    match reason {
+                        UnsafeRecordUpdateReason::UnknownVariant {constructed_variant} => {
+                            let text = wrap_format!("
+I cannot use this value to build an updated `{constructed_variant}` as it could be
```
lpil, line 1889: Remove this hard-coded newline by adding back the \ at the end please

lpil (commented):
This looks super nice! I've been testing it and it seems to work great!

One tiny note inline. It would be great if other folks could test this out and try and find any bugs, in case Gears and I have missed anything. @giacomocavalieri says he will be able to test tomorrow or so.

💜

compiler-core/src/error.rs
GearsDatapacks: Done!

GearsDatapacks:
I have now implemented exhaustiveness checking for narrowed types, as discussed on Discord.
This allows stuff like:
```gleam
type Wibble {
  Wibble(Bool)
  Wobble(Int)
}

pub fn main(b: Bool) {
  let always_wibble = Wibble(b)
  case always_wibble {
    Wibble(True) -> 1
    Wibble(False) -> 0
  }
  // ^ This case expression is exhaustive!
}
```

GearsDatapacks:
I should add version tracking to this as well. I was thinking about it before, and wasn't really sure if there was a good way to do it, but it should actually be pretty simple.

giacomocavalieri:
With some type system changes (like inference of anonymous functions in pipes) we decided to exclude those from the tracking. Here it seems like it wouldn't be too hard to check if someone is relying on narrowing when updating/accessing fields so I think it would be nice to track it

GearsDatapacks:
Yep. I'm not sure I'll be able to track the exhaustiveness checking for narrowed types, but record access and record updates should be doable

compiler-core/src/type_.rs
```diff
@@ -1279,6 +1279,9 @@ pub fn generalise(t: Arc<Type>) -> Arc<Type> {
 pub enum FieldAccessUsage {
     /// Used as `thing.field()`
     MethodCall,
+    /// Used internally when trying to access a field when
+    /// performing record updates
+    RecordUpdate,
```
giacomocavalieri, line 1284: It's not clear to me what this means, just by the name and description I'd guess it's this
```gleam
Wobble(..wobble, field: other.field)
//                            ^^^^^
```
Did you mean access to a field that is inferred to be safe after narrowing? I'd change the name in that case

compiler-core/src/type_/error.rs
```diff
@@ -841,6 +841,7 @@ pub enum FeatureKind {
     NestedTupleAccess,
     InternalAnnotation,
     AtInJavascriptModules,
+    TypeNarrowing,
```
giacomocavalieri, line 844: I'd define two different variants to produce nicer messages:
- one to talk about narrowing for safe record updates
- one to talk about narrowing for safe accessors

compiler-core/src/type_.rs
GearsDatapacks: Yeah sorry I'll clarify that

compiler-core/src/type_/error.rs
GearsDatapacks: Good idea!

lpil (commented):
This is looking super nice! I'm trying to give this a thorough test at the moment as I'm nervous about the change to type mutation. I'd appreciate others doing the same and using this branch for lots of programming.

compiler-core/src/type_/environment.rs
```diff
@@ -812,7 +835,11 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
 
         return match action {
             Action::Link => {
-                *type_.borrow_mut() = TypeVar::Link { type_: t2 };
+                let mut t2 = t2.deref().clone();
+                t2.generalise_custom_type_variant();
+                *type_.borrow_mut() = TypeVar::Link {
+                    type_: Arc::new(t2),
+                };
```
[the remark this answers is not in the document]
GearsDatapacks: Why would it? All the change does is create a new `Arc`, instead of cloning the existing `t2` one.

GearsDatapacks:
All your suggested wording changes have been implemented

CHANGELOG.md
```diff
@@ -126,6 +126,26 @@
 - Optimised creation of bit arrays on the JavaScript target.
   ([Richard Viney](https://github.com/richard-viney))
 
+- The compiler can now infer the variant of custom types when pattern matching,
+  allowing for safe record updates if the variant is known, or access fields
+  which are only present on one variant:
```
giacomocavalieri, line 131: Instead of cramming it into a single entry I'd split it in two: one for the record updates and one for the record access, so the example can be more focused

GearsDatapacks: Good idea, I wasn't super happy with how the entry looked

compiler-core/src/type_/environment.rs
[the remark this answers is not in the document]
lpil: Because it no longer mutates, which is how I made sure that linked types are updated together when there's more context.

CHANGELOG.md
lpil: It'd also need to bring up pattern matching!

lpil (approved):
Thank you!!

Issue #3603, opened 2024-09-09
LSP code action: Toggle between qualified and unqualified import


george-grec:
Currently, when you take a suggested auto import from the LSP for a type like here:
```gleam
pub fn wibble() {
  Som|
}
```
and select the `Some` variant from the `gleam/option` module you get the following code:

```gleam
import gleam/option

pub fn wibble() {
  option.Some|
}
```

In many cases, you actually want to use an unqualified import for a type or variant because you (plan to) use it heavily, e.g. Option, Some, None, lustre elements, Dynamic. To do that right now you have to go to the top of the file and find your way back to where you were.
This breaks your developer flow so it would be great if there was a code action when hovering the type or variant to change the import to an unqualified one or vice versa.

AcnoSaga:
I'd love to start hacking on this. Where can I find the LSP code? Still familiarising myself with the codebase.

lpil:
In here! https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/language_server.rs

giacomocavalieri:
Do we want to allow this just for record constructors or for all kind of functions? I think making it easier to use functions unqualified would be a net loss given we want to encourage the opposite style

lpil:
Just types and constructors sounds good

Frank-III:
> I'd love to start hacking on this. Where can I find the LSP code? Still familiarising myself with the codebase.

Hi @AcnoSaga, are you still looking into this? if not, I would like to give it a shot!

PR #3653, opened 2024-09-27
LSP Code Action: Convert to unqualified import

Frank-III:
hopefully resolve part of #3603

I am not very sure if `visit_typed_ast` should live in the visitor trait, please review my code 🙏.


compiler-core/src/type_.rs
```diff
@@ -597,6 +597,14 @@ impl ModuleValueConstructor {
             | ModuleValueConstructor::Constant { documentation, .. } => documentation.as_deref(),
         }
     }
+
+    pub fn name(&self) -> Option<&str> {
```
lpil, line 601: Constants have names too, so I don't think this is a well named function. You could inline the match expression rather than trying to come up with an appropriate generic name for this niche behaviour if that's easier.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_nested_type_outer.snap
```diff
@@ -0,0 +1,21 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport option\nimport wobble\npub fn main(x) -> option.Option(wobble.Wibble) {\n    todo\n}\n"
+---
+----- BEFORE ACTION
+
+import option
+import wobble
+pub fn main(x) -> option.Option(wobble.Wibble) {
+           ↑
```
lpil, line 10: Why does this suggest the code action? The cursor is not within the type constructor.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_nested_constructor_outer.snap
```diff
@@ -0,0 +1,21 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport option\nimport wobble\npub fn main(x) -> option.Option(wobble.Wibble) {\n    option.Some(wobble.Wobble(1))\n}\n"
+---
+----- BEFORE ACTION
+
+import option
+import wobble
+pub fn main(x) -> option.Option(wobble.Wibble) {
+           ↑
```
lpil, line 10: Same here

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_nested_constructor_inner.snap
```diff
@@ -0,0 +1,23 @@
...
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport option\nimport wobble\n\npub fn main(x) -> option.Option(wobble.Wibble) {\n    option.Some(wobble.Wobble(1))\n}\n"
+---
+----- BEFORE ACTION
+
+import option
+import wobble
+
+pub fn main(x) -> option.Option(wobble.Wibble) {
+           ↑
```
lpil, line 11: And here

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_multiple_imports.snap
```diff
@@ -0,0 +1,25 @@
...
+expression: "\nimport option\nimport wobble\n\npub fn main() {\n  option.Some(1)\n  wobble.Wibble(1)\n}\n"
+---
+----- BEFORE ACTION
+
+import option
+import wobble
+
+pub fn main() {
+  option.Some(1)
+             ↑
```
lpil, line 12: This one has applied the action to the wrong constructor?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_function.snap
```diff
@@ -0,0 +1,21 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport result\n\npub fn main() {\n  result.is_ok()\n}\n"
+---
+----- BEFORE ACTION
+
+import result
+
+pub fn main() {
+           ↑
```
lpil, line 10: This one too

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_function.snap
```diff
@@ -0,0 +1,21 @@
...
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport result\n\npub fn main() {\n  result.is_ok()\n}\n"
+---
+----- BEFORE ACTION
+
+import result
+
+pub fn main() {
+           ↑   
+  result.is_ok()
```
lpil, line 11: This code action should never be offered for functions. Unqualified functions are not supported.

lpil (commented):
Nice! I've left some notes inline.

Could you also add these tests:

- The constructor is used multiple times (all should be updated)
- The import comes after the constructor usage
- Record constructors without any arguments
- Type constructors without any arguments
- Record constructor patterns with arguments
- Record constructor patterns without arguments

Thank you. Please un-draft this PR when you're ready for another review.

Frank-III:
> Nice! I've left some notes inline.
> 
> Could you also add these tests:
> 
> * The constructor is used multiple times (all should be updated)
> * The import comes after the constructor usage
> * Record constructors without any arguments
> * Type constructors without any arguments
> * Record constructor patterns with arguments
> * Record constructor patterns without arguments
> 
> Thank you. Please un-draft this PR when you're ready for another review.

Thank you for the review! I would fix all others but regarding the first one could you elaborate more: 
- since we are traversing the module top down, to update all occurrences we need to keep track of all possible constructors until we find the overlapped one (and continue) and update those that are the same, do you have better ideas?
- when the selected span overlaps multiple constructors do we want to offer actions for all of them

Thank you again!

lpil:
I'd probably pass over the AST multiple times.

All of them sounds good, so long as it is clear from the message which one it is for. Otherwise the inner-most one is fine.

Frank-III:
I go for the inner-most one, all questions are resolved and I also add support when a module is aliased, please review it again!

compiler-core/src/language_server/code_action.rs
```diff
@@ -909,3 +909,449 @@ pub fn code_action_add_missing_patterns(
...
+            .find_in_code(import_start, import_end, |s| s.rfind('}'))
+            .is_some();
+
+        let trailing_comma_pos = if has_brace {
+            self.find_trailing_comma_before_closing_brace(import_start, import_end)
+        } else {
+            None
+        };
+
+        let (insert_pos, new_text) = match (
```
lpil, line 1029: Favour guards or some other structure over matching on 2 bools in a tuple as these patters are hard to understand.

compiler-core/src/language_server/code_action.rs
```diff
@@ -909,3 +909,449 @@ pub fn code_action_add_missing_patterns(
...
+            break;
+        }
+        None
+    }
+
+    fn edit_import(&mut self, name: String) {
+        let qualified_constructor = self
+            .qualified_constructor
+            .as_ref()
+            .expect("import should be set");
```
lpil, line 1014: Is this expectation correct? It says import but the property is called constructor.

compiler-core/src/language_server/code_action.rs
```diff
@@ -909,3 +909,449 @@ pub fn code_action_add_missing_patterns(
...
+                location.start + 1,
+            )
+        };
+        self.edits.push(TextEdit {
+            range: src_span_to_lsp_range(span, &self.line_numbers),
+            new_text: "".to_string(),
+        });
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
lpil, line 1150: Never turn off the linter, it is there for a reason.

compiler-core/src/language_server/code_action.rs
```diff
@@ -909,3 +909,449 @@ pub fn code_action_add_missing_patterns(
...
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+        self.second_pass = true;
+        self.visit_typed_module(&self.module.ast);
+        if self.edits.is_empty() {
+            return vec![];
+        }
+        let mut action = Vec::with_capacity(1);
+        CodeActionBuilder::new("Convert to unqualified import")
```
lpil, line 954: "Unqualify $NAME"

compiler-core/src/language_server/tests/action.rs
```diff
@@ -1560,6 +1561,435 @@ pub fn main(res) {
...
+"#;
+    assert_code_action!(
+        CONVERT_TO_UNQUALIFIED_IMPORT,
+        TestProject::for_source(src)
+            .add_hex_module("option", "pub type Option(v) { Some(v) None }")
+            .add_hex_module("wobble", "pub type Wibble { Wobble(Int) }"),
+        find_position_of("wobble.").select_until(find_position_of("Wibble")),
+    );
+}
+
```
lpil, line 1992: Could you add tests for other module names, for other constructor names, for the import being over multiple lines, and for the import to come after the usages. Thank you

lpil (commented):

compiler-core/src/language_server/code_action.rs
```diff
@@ -909,3 +909,449 @@ pub fn code_action_add_missing_patterns(
...
+            edits: vec![],
+            qualified_constructor: None,
+            should_stop: false,
+            second_pass: false,
+        }
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+        self.second_pass = true;
```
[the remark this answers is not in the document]
Frank-III: Thank you for the suggestion! it is indeed cleaner!

compiler-core/src/language_server/tests/action.rs
Frank-III: More tests are added including other all above and bad formated import, usage inside function call, list or tuple and within complex patterns

compiler-core/src/language_server/code_action.rs
```diff
@@ -1043,3 +1043,566 @@ impl<'a> AddAnnotations<'a> {
...
+        location: &'ast SrcSpan,
+        name: &'ast EcoString,
+        arguments: &'ast Vec<CallArg<TypedPattern>>,
+        module: &'ast Option<(EcoString, SrcSpan)>,
+        constructor: &'ast crate::analyse::Inferred<type_::PatternConstructor>,
+        spread: &'ast Option<SrcSpan>,
+        type_: &'ast Arc<Type>,
+    ) {
+        let range = src_span_to_lsp_range(*location, &self.line_numbers);
+        if overlaps(self.params.range, range) {
```
lpil, line 1209: I think we could avoid the extra work if we've found the constructor already, seeing as it does another module traversal within the `if` body.
```suggestion
        if self.qualified_constructor.is_none() && overlaps(self.params.range, range) {
```

lpil (approved):
Beautiful work!!! Thank you so much!

Could you update the changelog please 🙏

compiler-core/src/language_server/code_action.rs
Frank-III: I think for record value constructor yes, but for type constructors or pattern record constructor say `option.Option(wobble.Wobble)` the whole `option.Option(wobble.Wobble)` is valid range so when trigger at `wobble.Wobble` it would go through `option.Option` first which will set the qualified_constructor.

lpil (approved):
Beautiful work! Thank you very much!!!!

PR #3654, opened 2024-09-29
Suggest to pattern match on a `Result(a, _)` when `a` is expected

### LINKED ISSUE #3644: Suggest unwrapping when passing `Result(a, _)` instead of `a`
It would be nice if the compiler could suggest how to deal with a result when it's being used where the wrapped `Ok` value is expected instead. What I mean is:
```gleam
pub fn main() {
  let response =
    request.new()
    |> request.set_host("pokeapi.co")
    |> request.set_path("api/v2/pokemon/ditto")
    |> httpc.send

  parse_response(response)
}

fn parse_response(response: Response(String)) {
  todo
}
```
Have an error that looks like this:
```txt
error: Type mismatch
   ┌─ /Users/giacomocavalieri/Desktop/goto/src/goto.gleam:13:18
   │
13 │   parse_response(response)
   │                  ^^^^^^^^

Expected type:

    Response(String)

Found type:

    Result(Response(String), Dynamic)

Hint: to get a value out of a `Result` you can pattern match on it

    case response {
      Ok(response) -> parse_response(response)
      Error(_) -> todo as "deal with error"
    }
```
We could also have a code action to apply this fix too.

giacomocavalieri:
This PR closes #3644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,47 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The compiler can now suggest to pattern match on a `Result(a, b)` if it's
+  being used where a value of type `a` is expected. For example, this code:
+
+  ```gleam
+  import gleam/list
+  import gleam/int
+
+  pub fn main() {
+    let not_a_number = list.first([1, 2, 3])
+    int.add(1, not_a_number)
+  }
+  ```
+
+  Results in the following error:
+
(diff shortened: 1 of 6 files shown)


lpil:
Sorry, closed accidentally.

giacomocavalieri:
Oops sorry I forgot to accept the snapshot! 😁

lpil (approved):
Thank you!

PR #3663, opened 2024-10-02
Suggest wrapping value in `Ok` or `Error` if it makes sense 

giacomocavalieri:
As suggested by @GearsDatapacks the compiler can now suggest to wrap a value into an `Ok`/`Error` if it would make the types line up:
```gleam
pub fn main() {
  wibble(1)
//       ^ Did you mean to wrap this in an `Ok`?
}

pub fn wibble(value: Result(Int, String)) { todo }
```

This is a WIP because I still have to figure out how to pretty print this information in pattern matching branches, right now it looks quite bad 😁

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,11 +26,9 @@
   solve a type mismatch error:
 
   ```gleam
-  pub fn first(list: List(a)) -> Result(a, Nil) {
-    case number {
-      [] -> Error(Nil)
-      [first, ..rest] -> first
-    }
+  pub fn greet_logged_user() {
+    use <- bool.guard(when: !logged_in, return: Error(Nil))
+    "Hello!"
   }
   ```
 
@@ -38,15 +36,10 @@
 
   ```txt
   error: Type mismatch
-    ┌─ /src/one/two.gleam:5:5
+    ┌─ /main.gleam:7:3
     │
-  5 │     [first, ..rest] -> first
-    │     ^^^^^^^^^^^^^^^^^^^^^^^^
-    │                        │
-    │                        Did you mean to wrap this in an `Ok`?
-
-  This case clause was found to return a different type than the previous
-  one, but all case clauses must return the same type.
+  7 │   "Hello!"
+    │   ^^^^^^^^ Did you mean to wrap this in an `Ok`?
 
   Expected type:
 
@@ -54,7 +47,7 @@
 
   Found type:
 
-      a
+      String
   ```
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,43 @@
   allows the Language Server to provide completion hints for `case` subjects.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The compiler can now suggest to wrap a value in an `Ok` or `Error` if that can
+  solve a type mismatch error:
+
+  ```gleam
+  pub fn first(list: List(a)) -> Result(a, Nil) {
+    case number {
+      [] -> Error(Nil)
+      [first, ..rest] -> first
+    }
+  }
+  ```
+
+  Results in the following error:
+
+  ```txt
+  error: Type mismatch
+    ┌─ /src/one/two.gleam:5:5
+    │
+  5 │     [first, ..rest] -> first
+    │     ^^^^^^^^^^^^^^^^^^^^^^^^
+    │                        │
+    │                        Did you mean to wrap this in an `Ok`?
+
+  This case clause was found to return a different type than the previous
+  one, but all case clauses must return the same type.
+
+  Expected type:
+
+      Result(a, Nil)
+
+  Found type:
+
+      a
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
 - `gleam new` now has refined project name validation - rather than failing on
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -4066,7 +4066,7 @@ fn hint_unwrap_result(
     // If the got type is `Result(a, _)` and the expected one is
     // `a` then we can display the hint.
     let wrapped_type = given.result_ok_type()?;
-    if !wrapped_type.same_as(&expected) {
+    if !wrapped_type.same_as(expected) {
         None
     } else {
         Some(wrap_format!(
@@ -4076,7 +4076,7 @@ fn hint_unwrap_result(
       Ok(value) -> todo
       Error(error) -> todo
     }}",
-            printer.print_type(&expected),
+            printer.print_type(expected),
             printer.print_type(given),
         ))
     }
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2007,6 +2007,29 @@ But function expects:
                     text.push_str(&printer.print_type(expected));
                     text.push_str("\n\nFound type:\n\n    ");
                     text.push_str(&printer.print_type(given));
+
+                    let (main_message_location, main_message_text, extra_labels) = match situation {
+                        Some(UnifyErrorSituation::CaseClauseMismatch { clause_location }) =>
+                            match hint_wrap_value_into_result(expected, given) {
+                                None => (clause_location, None, vec![]),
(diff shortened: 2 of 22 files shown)


GearsDatapacks:
Looks like you forgot to accept the snapshot again :)

giacomocavalieri:
Ah yeah this time it's intentional, I'm not really satisfied with how the error looks for case expressions 😆

compiler-core/src/error.rs
```diff
@@ -3699,6 +3729,20 @@ fn hint_alternative_operator(op: &BinOp, given: &Type) -> Option<String> {
     }
 }
 
+fn hint_wrap_value_into_result(expected: &Arc<Type>, given: &Arc<Type>) -> Option<String> {
+    let expected = collapse_links(expected.clone());
+    let (expected_ok_type, expected_error_type) = expected.result_types()?;
+
+    let given = collapse_links(given.clone());
+    if collapse_links(expected_ok_type) == given {
```
lpil, line 3737: This won't work some times due to the inner types also being links. Good enough for some simple cases though.

Maybe can drop a TODO here? Or implement a `equal` method which is like unify without the mutation? Would remove the need to collapse links too.

lpil (commented):
Nice! Left 1 comment inline.

Could we have some tests for in `use` and returned from a function annotated as being result? Those will be the most common I think.

compiler-core/src/error.rs
giacomocavalieri: Done! I've also replace another piece of code that was using `collapse_links` to check for equality of two types

joshi-monster:
Hey! 💜 

I tried your `same_as` function to see if it could work for me to detect my record update type changes, and I noticed 2 problems with it where it might be too strict 🙂 

- Record (`Named`) types are currently considered to be different types if their inferred variant differs. The way it works also affects types with only one variant.
- Unbound and generic type variables are only considered equal if they refer to the same type variable.

The last one means that your new hint doesn't work for these cases for example:

```gleam
pub fn wibble() -> Result(List(String), Nil) {
  [] // List(a) `same_as` List(String)
}
pub fn wobble() -> Result(List(b), Nil) {
  [] // List(a) `same_as` List(b)
}
```

I feel like it would be more useful if unbound type variables where allowed to match any type on the right-hand side; Something like `is_assignable_to` or `can_unify` might maybe be more useful?

~ yoshi

lpil:

giacomocavalieri:
Oh my I hadn't noticed the conflicts, once those are solved I think this is ready for review!

giacomocavalieri:
This is ready now!

compiler-core/src/type_.rs
```diff
@@ -422,6 +434,121 @@ impl Type {
...
+                    package,
+                    module,
+                    name,
+                    ..
+                },
+                Type::Named {
+                    package: other_package,
+                    module: other_module,
+                    name: other_name,
+                    ..
```
lpil, line 460: There's a bug here! The parameters must also be the same

compiler-core/src/type_.rs
```diff
@@ -422,6 +434,121 @@ impl Type {
...
+                        .all(|(one, other)| one.same_as(other))
+                    && retrn.same_as(other_retrn)
+            }
+
+            (Type::Var { type_ }, other) => type_.as_ref().borrow().same_as_other_type(other),
+
+            (Type::Tuple { .. }, Type::Fn { .. } | Type::Named { .. }) => false,
+            (one @ Type::Tuple { .. }, Type::Var { type_ }) => {
+                type_.as_ref().borrow().same_as_other_type(one)
+            }
```
lpil, line 488: Could you merge all the ones with var on the right hand side and move the new merged clause along with clause with var on the left up to the top please.

compiler-core/src/type_.rs
```diff
@@ -422,6 +434,121 @@ impl Type {
...
+}
+
+impl TypeVar {
+    #[must_use]
+    fn same_as_other_type(&self, other: &Type) -> bool {
+        match (self, other) {
+            (
+                TypeVar::Unbound { .. },
+                Type::Named { .. } | Type::Fn { .. } | Type::Tuple { .. },
+            ) => false,
```
lpil, line 507: How come this is false?

lpil (commented):
Wonderful, thank you! Left a few small notes

compiler-core/src/type_.rs
giacomocavalieri: I thought an unbound variable is never meant to survive the type inference process, so I chose a safe default there

compiler-core/src/type_.rs
giacomocavalieri: Not sure I understand how you want it, I tried changing it, let me know if it looks how you intended!

giacomocavalieri: Changed it! Not sure if it's the look you wanted, let me know if it's better now

compiler-core/src/type_.rs
giacomocavalieri: Oops thanks for catching this!

giacomocavalieri:
This should be ready now!

CHANGELOG.md
```diff
@@ -82,8 +82,49 @@
...
+  Results in the following error:
+
+  ```txt
+  error: Type mismatch
+    ┌─ /src/one/two.gleam:5:5
+    │
+  5 │     [first, ..rest] -> first
+    │     ^^^^^^^^^^^^^^^^^^^^^^^^
+    │                        │
+    │                        Did you mean to wrap this in an `Ok`?
```
lpil, line 106: This error is wrong! `Ok([first, ..rest] -> first)` would not be valid

compiler-core/src/error.rs
```diff
@@ -1879,19 +1881,49 @@ But function expects:
...
+                                    clause_location,
+                                    None,
+                                    vec![
+                                        ExtraLabel {
+                                            src_info: None,
+                                            label: Label {
+                                                text: Some(hint),
+                                                span: *location,
+                                            }
+                                        },
```
lpil, line 1904: I'm finding this quite hard to read with all the nesting, it's easy to miss that it's returning a tuple. Could you assign this bit to a variable pleaes

compiler-core/src/type_.rs
lpil: That's only in the interface, there can be many unbound variables within.

CHANGELOG.md
giacomocavalieri: The message is referring to the `first` bit but the way the two overlapping spans are rendered it's not very clear; do you think we should stop highlighting the entire case branch in this case? However that would mean that the error message is not very precise now:
```
[first, ..rest] -> first
                   ^^^^^ Did you mean to wrap this in an `Ok`?

This case clause was found to return a different type than the previous
one, but all case clauses must return the same type.
``` 
And it would be more confusing when the part on the right of `->` is on its own line as the clause would not be displayed in the error!
```
  first
  ^^^^^ Did you mean to wrap this in an `Ok`?

This case clause was found to return a different type than the previous
one, but all case clauses must return the same type.
```

lpil: I think we should continue to highlight the entire cause as the source of the type error, but a label about wrapping an expression should only point to the code that is to be wrapped.

giacomocavalieri: The tricky thing is it is already pointing to just the code that needs to be wrapped but the way the library renders it it's not immediately obvious. If the then clause were not on the same line what you'd see is this:
```gleam
     case wibble {
       A -> Error("a")
  ->   wobble ->
 |     1
 |     ^ Did you mean to wrap this ...
 |
 incorrect branch
```
But when the two are on the same line the two look like this
```gleam
wobble -> 1
^^^^^^^^^^^
          |
          Did you mean to wrap this
incorrect branch
```
The way it gets rendered is not the best, maybe I could try and remove it for branches when the then clause is on the same line or for case matches entirely

lpil: Ah, I see, that is annoying.

CHANGELOG.md
```diff
@@ -4,6 +4,43 @@
...
```
lpil, line 28: This changelog entry needs updating I think? It no longer makes this suggestion

giacomocavalieri: Updated, thank you!

lpil (approved):
Thank you!!

PR #3667, opened 2024-10-04
Fix invalid TypeScript definitions for opaque types that use a private type

richard-viney:
This code:

```gleam
type Foo {
  Foo
}

pub opaque type Bar {
  Bar(Foo)
}
```

was generating an invalid `.d.mts` file that referenced a non-existent type `Foo$`:

```typescript
import type * as _ from "./gleam.d.mts";

class Bar extends _.CustomType {
  constructor(argument$0: Foo$);
  
  0: Foo$;
}

export type Bar$ = Bar;
```

This PR fixes this by:

- Emitting TypeScript definitions for all types, regardless of visibility
- Only `export`ing relevant parts when they have non-private visibility

Variants of opaque types remain non-exported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,10 @@
   JavaScript prelude.
   ([Richard Viney](https://github.com/richard-viney))
 
+- Fixed a bug where generated TypeScript definitions were invalid for opaque
+  types that use a private type.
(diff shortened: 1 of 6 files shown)


lpil (approved):
Thank you!!

Issue #2315, opened 2023-09-01
Typescript: imported type alias produces invalid declaration


vonagam:
```gleam
// a
import gleam/dynamic.{Dynamic}

pub type Input =
  List(Dynamic)
```
```gleam
// b
import b

pub fn process(input: b.Input) -> Int {
  todo
}
```
Produces:
```ts
// a, valid
import type * as dynamic from "../../gleam_stdlib/gleam/dynamic.d.mts";
import type * as _ from "../gleam.d.mts";

export type Input = _.List<dynamic.Dynamic$>;
```
```ts
// b, invalid
import type * as a from "../a.d.mts"; // imported, not used
import type * as _ from "../gleam.d.mts";

export function process(input: _.List<dynamic.Dynamic$>): number; // dynamic is not imported
```

So, either `input` in `process` should become `a.Input` or import for `dynamic` should be added.

lpil:
Thank you

richard-viney:
I've opened a PR that fixes this: https://github.com/gleam-lang/gleam/pull/3668

PR #3668, opened 2024-10-04
Fix TS definitions when importing type aliases that reference unimported modules

richard-viney:
Fixes https://github.com/gleam-lang/gleam/issues/2315.

The approach taken is to look at the entire module's contents, recurse through all referenced types, and add imports for everything that's encountered.

This doesn't take the possibly preferable approach of importing the relevant type alias directly because type aliases don't appear to be tracked in the type information, and I decided not to go down the path of seeing how much work it would be to change that.

Questions:

1. Note that the original `import` statements in the Gleam code no longer get converted 1:1 to TypeScript imports. Does this seem reasonable?

### DIFF
--- a/compiler-core/src/javascript/tests/type_alias.rs
+++ b/compiler-core/src/javascript/tests/type_alias.rs
@@ -23,3 +23,33 @@ pub opaque type OpaqueType {
 "#,
     );
 }
+
+#[test]
+fn import_indirect_type_alias() {
+    assert_ts_def!(
+        (
+            "wibble",
+            "wibble",
+            r#"
+pub type Wibble {
+  Wibble(Int)
+}
+"#
+        ),
+        (
+            "wobble",
+            "wobble",
+            r#"
+import wibble
+pub type Wobble = wibble.Wibble
+"#
+        ),
+        r#"
+import wobble
+
+pub fn main(x: wobble.Wobble) {
+  Nil
+}
+"#,
+    );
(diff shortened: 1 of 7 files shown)


lpil (commented):
Thank you! This looks fab. It does need tests, you'll need to make a new test macro if there's no existing TS one for multiple modules.

richard-viney:
Tests have been added. Ready for review.

compiler-core/src/javascript/tests/type_alias.rs
```diff
@@ -23,3 +23,33 @@ pub opaque type OpaqueType {
...
+            "wobble",
+            r#"
+import wibble
+pub type Wobble = wibble.Wibble
+"#
+        ),
+        r#"
+import wobble
+
+pub fn foo(x: wobble.Wobble) {
```
lpil, line 50: No "foo" please, it's a morbid WW2 reference

lpil (commented):
Very nice! Thank you. One small note inline

compiler-core/src/javascript/tests/type_alias.rs
richard-viney: Changed to "main"

richard-viney:
FYI the CI failures are due to Rust 1.82 that was just released.

lpil (approved):
Champion!!! Thank you

PR #3669, opened 2024-10-04
Fixed export package-definitions not including cached modules

### LINKED ISSUE #2898: `gleam export package-interface` produces empty modules object
When I run `gleam export package-interface --out <some_path>.json` on any of my packages (tested with simplifile and gsv),
the file produced has an empty object for the modules field as though they were empty packages.
```json
{"name":"gsv","version":"1.4.0","gleam-version-constraint":">= 0.32.0","modules":{}}
```

I checked the package-interface.json in the build directory and it's being correctly produced, so the module info get's lost somehow exporting the file.

I'm on an M2 Mac running the main branch of Gleam. The logs from after the compile stage are pretty sparse
```
Compiled in 0.01s
TRACE writing_file path="/Users/benjaminpeinhardt/Development/Projects/gsv/./file.json"
INFO Successfully completed
```

I'll try and look into this as I bet it's something small or I'm doing something wrong, but if someone beats me to it I won't be upset haha.

mine-tech-oficial:
This PR closes #2898


compiler-core/src/build/package_compiler.rs
```diff
@@ -594,28 +603,28 @@ pub(crate) fn module_name(package_path: &Utf8Path, full_module_path: &Utf8Path)
 #[derive(Debug)]
 pub(crate) enum Input {
     New(UncompiledModule),
-    Cached(CachedModule),
+    Cached((CachedModule, CacheMetadata)),
```
aslilac, line 606: this doesn't need to be wrapped in a tuple, you're already in an enum tuple variant

```suggestion
    Cached(CachedModule, CacheMetadata),
```

compiler-core/src/build/package_compiler.rs
```diff
@@ -594,28 +603,28 @@ pub(crate) fn module_name(package_path: &Utf8Path, full_module_path: &Utf8Path)
...
 }
 
 impl Input {
     pub fn name(&self) -> &EcoString {
         match self {
             Input::New(m) => &m.name,
-            Input::Cached(m) => &m.name,
+            Input::Cached(m) => &m.0.name,
```
aslilac, line 613: ...which will also enable nicer pattern matching here

```suggestion
            Input::Cached(module, ..) => &module.name,
```

compiler-core/src/build/package_compiler.rs
```diff
@@ -594,28 +603,28 @@ pub(crate) fn module_name(package_path: &Utf8Path, full_module_path: &Utf8Path)
...
         }
     }
 
     pub fn source_path(&self) -> &Utf8Path {
         match self {
             Input::New(m) => &m.path,
-            Input::Cached(m) => &m.source_path,
+            Input::Cached(m) => &m.0.source_path,
```
aslilac, line 620: same thing here

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Thx for the suggestion! Fixed in commit #4789222ac8db5fe157576681547f5458f1508e76

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Thanks! Fixed in commit 4789222ac8db5fe157576681547f5458f1508e76

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Thanks! Fixed in commit 4789222ac8db5fe157576681547f5458f1508e76

mine-tech-oficial:
Do I need to write any tests? I didn't write any, just fixed the snapshots. I guess I could duplicate the already existing tests to take account cached modules.

compiler-core/src/build/package_compiler.rs
```diff
@@ -304,11 +309,13 @@ where
             let name = format!("{}.cache_meta", &module_name);
             let path = artefact_dir.join(name);
             let info = CacheMetadata {
+                name: module.name.clone(),
                 mtime: module.mtime,
                 codegen_performed: self.perform_codegen,
                 dependencies: module.dependencies.clone(),
                 fingerprint: SourceFingerprint::new(&module.code),
                 line_numbers: module.ast.type_info.line_numbers.clone(),
+                interface: package_interface::ModuleInterface::from_module(module),
```
lpil, line 318: The module interface is now always being generated. It should only be built as needed and it must not be part of the build cache.

mine-tech-oficial: So the goal is to not increase the cache size, correct? If so, when `gleam export package-definitions` is ran, we'd somehow convert the `CachedModule` or `CacheMetadata` into a `ModuleInterface`, correct?

mine-tech-oficial: Although to generate a `ModuleInterface` we need the AST, so we'd need to analyze (at least partially) the cached modules. So I guess it's a tradeoff between more time spent building + a bigger cache v.s. more time spent exporting.

mine-tech-oficial: Makes sense. Fixed in commit 20132488dac1195825fb68fd84d4d85822e2192d

compiler-core/src/build.rs
```diff
@@ -184,6 +185,7 @@ pub enum Mode {
     Dev,
     Prod,
     Lsp,
+    PackageInterface,
```
lpil, line 188: Package interface is not a compiler mode, this must be removed.

compiler-core/src/build/package_loader.rs
```diff
@@ -157,11 +157,18 @@ where
 
                 // A cached module with no stale dependencies can be used as-is
                 // and does not need to be recompiled.
-                Input::Cached(info) => {
-                    tracing::debug!(module = %info.name, "module_to_load_from_cache");
-                    let module = self.load_cached_module(info)?;
-                    loaded.cached.push(module);
-                }
+                Input::Cached(info) => match self.mode {
+                    Mode::PackageInterface => {
```
lpil, line 161: This always compiles modules, even if they have already been compiled.
This isn't the best experience as it repeats already done work, it would be better to build the interface from the cache I think.

mine-tech-oficial:
The problem is that, to build the interface from the cache, we'd need to save the typed AST, which I think is less space-efficient than just saving the interface. That's why I rebuild the modules.
Em quarta-feira, 9 de outubro de 2024 às 09:00, Louis Pilfold ***@***.***> escreveu:

lpil:
Neither of those approaches are what we want.

Does the cache not have enough information to build the module interface JSON from? What's missing?

mine-tech-oficial:
As the package interface is built from various module interfaces, each of these being built from all the definitions, it needs to be built from the typed AST, which isn't cached

lpil:
Why does it need to be built from typed AST? What information do they hold that the interface JSON needs that the cache does not have?

mine-tech-oficial:
I tried adapting the `from_module` function of the `ModuleInterface` struct from the `package_interface` module to work with `UncompiledModule`s (untyped ASTs) but, as it turns out, it needs the type info.
For example, here's the error cargo gives:
```
error[E0308]: mismatched types
   --> compiler-core/src/package_interface.rs:615:73
    |
615 | ...                   type_: from_type_helper(&arg.type_, &mut id_map),
    |                              ---------------- ^^^^^^^^^^ expected `&Type`, found `&()`
    |                              |
    |                              arguments to this function are incorrect
    |
    = note: expected reference `&type_::Type`
               found reference `&()`
```

lpil:
The cache has all the type information for a compiled module, so you can use that.

GearsDatapacks:
From when I looked into implementing this, it seems the main limitation is type aliases. Type alias information isn't stored in the cached interface, so that would need to be accessed somehow

mine-tech-oficial:
After trying to adapt the `from_module` function of the `ModuleInterface` struct from the `package_interface` module to work with `UncompiledModule` and `ModuleInterface` (from the `type_` module) and, as Gears pointed out, it seems that the only thing missing in the cache is the type aliases.

GearsDatapacks:
You don't need `UncompiledModule`. If you add type aliases to it, `ModuleInterface` should be enough

mine-tech-oficial:
I think I might need add a HashMap of TypeAlias to the ModuleInterface struct. Is that okay?

mine-tech-oficial:
I've been doing this and I've stumbled onto a problem: somehow, when the code will encode the modules to the cache, somehow the documentations have an extra newline? Do I have to somehow fix it or can I leave it like so?

GearsDatapacks:
I would say if you can reproduce the bug without your changes, then open a separate issue and let that be fixed separately. If this PR introduces that bug, then please fix it

mine-tech-oficial:
I can actually reproduce it separately, will open an issue for it then

mine-tech-oficial:
Issue opened at #3712

mine-tech-oficial:
Now it reuses as much of the cache as possible, although there is a test failing and I don't know why (need to fix). I'll open for review just to make sure any of the overall logic isn't wrong.

compiler-core/src/build/project_compiler.rs
```diff
@@ -288,7 +292,7 @@ where
         // longer need to have the package borrowed from self.packages.
         let package = self.packages.get(name).expect("Missing package").clone();
         let result = match usable_build_tools(&package)?.as_slice() {
-            &[BuildTool::Gleam] => self.compile_gleam_dep_package(&package),
+            &[BuildTool::Gleam] => self.compile_gleam_dep_package(&package).map(|_| vec![]),
```
lpil, line 295: Why is the result discarded?

compiler-core/src/build/package_compiler.rs
```diff
@@ -104,7 +106,7 @@ where
         stale_modules: &mut StaleTracker,
         incomplete_modules: &mut HashSet<EcoString>,
         telemetry: &dyn Telemetry,
-    ) -> Outcome<Vec<Module>, Error> {
+    ) -> Outcome<(Vec<Module>, Vec<type_::ModuleInterface>), Error> {
```
lpil, line 109: Prefer structs to tuples so it's clear what they are

compiler-core/src/package_interface.rs
```diff
@@ -553,6 +614,159 @@ impl ModuleInterface {
             functions,
         }
     }
+
+    pub fn from_untyped_module(interface: &type_::ModuleInterface) -> ModuleInterface {
```
lpil, line 618: This is not an untyped module!

compiler-core/src/package_interface.rs
```diff
@@ -372,11 +377,67 @@ impl PackageInterface {
...
                 .map(|module| (module.name.clone(), ModuleInterface::from_module(module)))
+                .chain(
+                    package
+                        .cached_modules
+                        .iter()
+                        .filter(|module| !package.config.is_internal_module(module.name.as_str()))
+                        .map(|interface| {
+                            (
+                                interface.name.clone(),
+                                ModuleInterface::from_untyped_module(interface),
```
lpil, line 388: Let's always use the interface and then delete the code for doing it from the AST.

lpil (commented):
Looking good so far! Thank you

compiler-core/build.rs
```diff
@@ -1,7 +1,7 @@
...
-    // capnpc::CompilerCommand::new()
-    //     .file("schema.capnp")
-    //     .output_path("generated/")
-    //     .run()
-    //     .expect("compiling schema.capnp");
+    capnpc::CompilerCommand::new()
+        .file("schema.capnp")
+        .output_path("generated/")
+        .run()
+        .expect("compiling schema.capnp");
```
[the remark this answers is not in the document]
mine-tech-oficial: Yeah, forgot to remove that! Thanks for the heads-up!

compiler-core/src/build/package_compiler.rs
```diff
@@ -646,7 +651,7 @@ pub(crate) struct CachedModule {
 }
 
 #[derive(Debug, serde::Serialize, serde::Deserialize)]
-pub(crate) struct CacheMetadata {
+pub struct CacheMetadata {
```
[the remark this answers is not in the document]
mine-tech-oficial: This is probably some leftover from my first two attempts, will change it back!

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Should I call the struct `Compiled`, `CompileResult` or something else?

compiler-core/src/package_interface.rs
mine-tech-oficial: So I delete the current `ModuleInterface::from_module` and rename the `ModuleInterface::from_untyped_module` to `ModuleInterface::from_interface`?

compiler-core/src/analyse.rs
```diff
@@ -1191,14 +1202,27 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 TypeConstructor {
                     origin: *location,
                     module: self.module_name.clone(),
-                    parameters,
-                    type_,
+                    parameters: parameters.clone(),
```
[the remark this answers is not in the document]
mine-tech-oficial: Will remove the `parameters `clone and make so it only clones `type_` once, instead of twice.

compiler-core/src/analyse.rs
```diff
@@ -965,7 +973,10 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                     }
                 };
 
-                fields.push(TypeValueConstructorField { type_: t.clone() });
+                fields.push(TypeValueConstructorField {
+                    label: label.clone().map_or(None, |(_, label)| Some(label.clone())),
```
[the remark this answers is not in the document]
mine-tech-oficial: Will change the first `clone` to a `as_ref`, so I only clone the EcoString (which should be efficient)

compiler-core/src/build.rs
mine-tech-oficial: Okay, removed that!

compiler-core/src/build/package_loader.rs
mine-tech-oficial: Fixed in commit 11c1d18!

compiler-core/schema.capnp
```diff
@@ -21,15 +21,17 @@ struct Option(Value) {
...
+  types @2 :List(Property(TypeConstructor));
+  typeAliases @3 :List(Property(TypeAliasConstructor));
+  values @4 :List(Property(ValueConstructor));
+  accessors @5 :List(Property(AccessorsMap));
+  package @6 :Text;
+  typesConstructors @7 :List(Property(TypesVariantConstructors)); 
+  lineNumbers @8 :LineNumbers;
+  srcPath @9 :Text;
+  isInternal @10 :Bool;
+  requiredVersion @11 :Version;
```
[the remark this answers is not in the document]
mine-tech-oficial: To be fair what I did was add documentation and type aliases in between the other fields and changed the `@N` annotation to keep it tidy, although I just read in the Cap'n Proto docs that I shouldn't do this, or else I ditch backwards-compatibility. Will change it back to what it was!

mine-tech-oficial:
Uhm, apparently this kind of broke the LSP, as it isn't considering the type aliases anymore. Guess I'll also need to tweak it

compiler-core/src/type_.rs
```diff
@@ -641,6 +643,11 @@ impl ModuleInterface {
     pub fn contains_todo(&self) -> bool {
         self.warnings.iter().any(|warning| warning.is_todo())
     }
+
+    pub fn remove_duplicated_type_aliases(&mut self) {
+        self.types
+            .retain(|name, _| !self.type_aliases.contains_key(name));
+    }
```
lpil, line 650: What is this function for? Seems to be called a lot.

compiler-core/src/package_interface.rs
```diff
@@ -370,183 +375,169 @@ impl PackageInterface {
...
-                            deprecation: DeprecationInterface::from_deprecation(deprecation),
-                            documentation: documentation.as_ref().map(|(_, doc)| doc.clone()),
+                            deprecation: DeprecationInterface::from_deprecation(&value.deprecation),
+                            documentation,
                             parameters: arguments
                                 .iter()
                                 .map(|arg| ParameterInterface {
-                                    label: arg.names.get_label().cloned(),
-                                    type_: from_type_helper(arg.type_.as_ref(), &mut id_map),
+                                    //TODO: Find label
```
lpil, line 504: Is this unfinished?

compiler-core/src/build/package_compiler.rs
```diff
@@ -146,7 +153,7 @@ where
         };
 
         // Load the cached modules that have previously been compiled
-        for module in loaded.cached.into_iter() {
+        for module in loaded.cached.clone().into_iter() {
```
lpil, line 156: It seems rather wasteful to clone all this information, especially since it's only going to be sometimes used, and it already stored elsewhere. Could instead we reference the copy that is stored in the existing_modules collection?

compiler-core/src/build/package_compiler.rs
```diff
@@ -279,7 +302,7 @@ where
         Ok(())
     }
 
-    fn encode_and_write_metadata(&mut self, modules: &[Module]) -> Result<()> {
+    fn encode_and_write_metadata(&mut self, modules: &Vec<Module>) -> Result<()> {
```
lpil, line 305: How come this has changed?

compiler-core/src/language_server/completer.rs
```diff
@@ -16,7 +16,8 @@ use crate::{
     line_numbers::LineNumbers,
     type_::{
         self, collapse_links, pretty::Printer, AccessorsMap, FieldMap, ModuleInterface,
-        PreludeType, Type, TypeConstructor, ValueConstructorVariant, PRELUDE_MODULE_NAME,
+        PreludeType, Type, TypeAliasConstructor, TypeConstructor, ValueConstructorVariant,
```
lpil, line 19: How come there are changes in this file? I wouldn't have expected there to be changes to language server code to enable exporting the package interface from cache

compiler-core/src/language_server/completer.rs
```diff
@@ -397,11 +398,18 @@ where
             });
         }
 
-        // Module types
+        // Module types and type aliases
```
lpil, line 401:
```suggestion
        // Module types
```

lpil (commented):
Hello! Looking good. I've left some questions inline

compiler-core/src/analyse.rs
```diff
@@ -1192,9 +1204,24 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                     origin: *location,
                     module: self.module_name.clone(),
                     parameters,
-                    type_,
+                    type_: type_ref,
                     deprecation: deprecation.clone(),
                     publicity: *publicity,
+                    // TODO: Find if the type is opaque
```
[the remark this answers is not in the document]
mine-tech-oficial: Forgot about this todo 😅. Is there a problem for the code to register the type alias as a non-opaque type, just for type analysis?

compiler-core/src/build.rs
```diff
@@ -225,7 +227,7 @@ impl Package {
     }
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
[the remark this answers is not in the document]
mine-tech-oficial: That's true. It's probably some remainder of my previous attempts. Will remove!

compiler-core/src/type_.rs
mine-tech-oficial: As during the type analysis aliases are registered as normal types, it needs to be removed later. I think I tried removing it only after analysis but there were still type aliases registered as types. I guess there's a better way to do this, although I haven't found how.

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: I'm sorry, but I can't understand exactly what you mean. Do you mean reusing the modules stored in the `existing_modules` further down the line, as the cached modules were moved into it?

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Oh, that was a remainder of some (unpublished) attempts. Will change it back!

compiler-core/src/language_server/completer.rs
mine-tech-oficial: It's because, as types and type aliases are now stored separately, we need to account for them separately in the completion

compiler-core/src/language_server/completer.rs
mine-tech-oficial: I don't understand the proposed change. That section takes care of registering the module types and it's aliases.

compiler-core/src/package_interface.rs
mine-tech-oficial: Oh, yeah, forgot about this part. I couldn't find how to retrieve the parameter label from the `ModuleInterface`. Maybe the arguments in the `Fn` variant of `Type` needs to be converted from a `Vec<Arc<Type>>` to a ``Vec<(Option<EcoString>, Arc<Type>)>`?

mine-tech-oficial:
Hey @lpil, left some answers to your questions. Waiting for some answers!

compiler-core/src/analyse.rs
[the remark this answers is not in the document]
lpil: I'm not aware of any problems, but I'm not sure what you mean here.

compiler-core/src/type_.rs
lpil: They should not be removed, no. They are types in the module.

compiler-core/src/language_server/completer.rs
lpil: Thank you

compiler-core/src/language_server/completer.rs
lpil: Type aliases are types.

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: I guess I could store the names of the cached modules, and retrieve those from the `existing_modules` collection further down the line, although that would require cloning anyway, right?

compiler-core/src/language_server/completer.rs
mine-tech-oficial: Uhm, makes sense. Will revert it!

mine-tech-oficial:
Hey @lpil, I've been trying to rewrite the type analysis so that it doesn't need to register the type aliases as types just for the analysis, but the `analyze_custom_type` function returns a `TypeConstructor`. Is it ok to return the `TypeAliasConstructor` (in case it's a type alias and not a "normal" type) as a non-opaque `TypeConstructor`?

mine-tech-oficial:
Actually I still need to discover what the parameters of the `TypeConstructor` would be...

mine-tech-oficial:
Hey @lpil, do you think it's ok for me to leave the PR with the odd way of registering type aliases as "normal types" just for the analysis, and removing it later (as I still couldn't find a good way to consider it separately), or do you want me to try and fix it anyway?

GearsDatapacks:
Hi @mine-tech-oficial, are you still working on this PR? If not, I'm happy to pick up the work as it would be cool to get this fixed. Thanks!

mine-tech-oficial:
@GearsDatapacks, actually for the past few weeks I didn't do much progress due to not having much time, although I was writing some code to get the labels of the parameters

compiler-core/src/build/package_compiler.rs
mine-tech-oficial: Also should be fixed now

mine-tech-oficial:
@lpil, now I think I only need to find the documentation (not sure how though)

compiler-core/schema.capnp
```diff
@@ -90,6 +90,11 @@ struct RecordAccessor {
   label @2 :Text;
 }
 
+struct FunctionArgument {
+  name @0 :Text;
```
lpil, line 94: Function types don't contain names! Please remove this.

It would also be called a parameter. Things have parameters, and for those parameters they accept arguments.

mine-tech-oficial: So you mean that a parameter has a name/label, while an argument doesn't? In that case, I'd reuse the `TypeValueConstructorParameter` struct?

mine-tech-oficial: To be fair I wasn't happy with the approach I took 😅 . Tomorrow I'll try tackling this again.

mine-tech-oficial:
Hey @lpil, I've left a comment on your answers. Waiting for you to confirm it! (no rush by the way)

compiler-core/schema.capnp
lpil: Neither has names at the type level, but parameters can have names locally within a definition.

mine-tech-oficial:
I think I won't be working on this PR on the near future, so I'll leave it open for anyone who want's to take it. (cc @lpil and @GearsDatapacks). If anyone takes it, I'd consider not using the latest commit, as the only thing it changes (what Louis mentioned above) should be removed.

GearsDatapacks:
I believe this PR can be closed now as the related issue has now been fixed.

lpil:
Thank you all!

PR #3670, opened 2024-10-05
Cleanup code related to try expressions

giacomocavalieri:
I noticed there was some error reporting left related to try-expressions. Since those are no longer part of the language I removed it


lpil (approved):
Thank you

Issue #2249, opened 2023-06-29
Sourcemaps


lpil:
So that exceptions and debugging tools can show Gleam syntax and line numbers.

l1f:
I am interested in this issue. But it is a year old. Is any planing or research on this subject done?

lpil:
Not that anyone has shared with me.

AlisCode:
Hello there, i've started work on this over at my fork on branch `feat/sourcemap-javascript`.

It's not in a place where I want to submit it for review, but so far I've managed to produce a basic sourcemap for a test Gleam module. This is it, visualized.

![image](https://github.com/user-attachments/assets/7fa3d09d-95e5-4e0e-8272-b0e02325485e)

I was able to make debugging work through `vscode-js-debug` on `neovim`, using the following DAP config : 

```lua
{
    type = "pwa-node",
    request = "launch",
    name = "Debug gleam program",
    -- NOTE: This wont work if the project's name doesnt match the folder name
    -- This should ideally be the name variable in the Gleam.toml manifest
    program = "${workspaceFolder}/build/dev/javascript/${workspaceFolderBasename}/gleam.main.mjs",
    cwd = "${workspaceFolder}",
    outFiles = {"${workspaceFolder}/**/*.mjs"},
    resolveSourceMapLocations = {"${workspaceFolder}/build/dev/javascript/**/*.map"},
    sourceMaps = true,
    pauseForSourceMap = true,
}
```

Is anyone actively interested in this, and willing to provide feedback ?

lpil:
I think people are interested for sure!

Hau-Hau:
Hey all, do you have any ETA for this PR?

GearsDatapacks:
I don't think the author of that PR is planning to continue it, so nobody is working on this at the moment

AlisCode:
Indeed I am no longer actively working on this, although I do believe that my draft PR can serve as a starting point maybe. 

I'm not sure the approach is good, one major piece of feedback was that the pretty-printer is affected by those changes. I dont know how it would be possible to implement sourcemaps without touching the pretty-printer in Gleam's current architecture, so the work kinda stopped there.

ankddev:
@lpil This one could be closed!

Issue #3149, opened 2024-05-17
Code Generation: Include file and line attribute


inoas:
This would allow jumping from the generated source code to the gleam source file.

On target Erlang the file attribute could be used: 

> Setting File and Line
>
> The same syntax as for module attributes is used for changing the pre-defined macros ?FILE and ?LINE:
>
> -file(File, Line).
>
> This attribute is used by tools, such as Yecc, to inform the compiler that the source program is generated by another tool. It also indicates the correspondence of source files to lines of the original user-written file, from which the source program is produced.

Source: https://www.erlang.org/doc/reference_manual/modules#setting-file-and-line

As one effect, this would help tools like [stacky](https://hexdocs.pm/stacky/index.html) to allow jumping into the gleam source file from the stack trace.

- On target JS if there is any standard solution, that could be used? Otherwise we can just setup a convention?
- Optional: Maybe the compiler could be told to not include such details for security reasons, by a compiler flag, or is this a no-issue if the pathes are relative not absolute?

lpil:
> On target JS if there is any standard solution, that could be used?

On JS they use sourcemaps, which is a much better solution and will give perfect results. We have an issue for this.

PR #3675, opened 2024-10-06
Add sourcemap generation to JavaScript codegen target 

AlisCode:
Here goes!

* Resolves https://github.com/gleam-lang/gleam/issues/2249
* Partially adresses https://github.com/gleam-lang/gleam/issues/3149 - i don't know anything about Erlang so generating sourcemaps is out of my zone of comfort. 

The PR is still kind of WIP - mainly, I'd like to talk about how to approach testing (and actually add way more snapshot tests before this gets a chance to be merged).

I also have 2 expects on calls to `sourcemap` - I'm sure this is not the right way to do it, so I'm wondering what you guys think is the best approach ? Is it fine to have a "sourcemap error" variant in the big Gleam `Error` enum ?

# Context 

I highly recommend watching the following talk if you're unfamiliar with Sourcemaps : https://www.youtube.com/watch?v=6LI0BJIiamg

Long story short, a sourcemap is a file that defines mappings from specific places in a generated file (the generated javascript code - which needs to be referring to its sourcemap file) to a source file (the gleam module).

It's used for example : 
* By the browser to be able to display the actual line of the Gleam source instead of the generated JS code 
* In tools like Sentry (I'm not affiliated, but they're huge in the industry) to be able to point to the actual source instead of the generated Javascript code
* By debuggers such as [vscode-js-debug](https://github.com/microsoft/vscode-js-debug) to be able to provide debugging straight from your Gleam sources : this PR enables using a debugger on Gleam, which I'm sure some people are excited about (I am! :tada:).

# Content

This PR adds sourcemap generation for the JavaScript codegen target. 
Sourcemap generation is disabled by default, and needs to be manually enabled using the following config in the gleam.toml manifest : 

```toml
[javascript]
sourcemap = true
```

This is done out of symmetry with other languages and tools - for example TypeScript does not automatically generate them. I imagine as your codebase grows large, it can end up being quite a heavy process. Also I'm assuming if this ever gets released, it should maybe be marked experimental.

### DIFF
--- a/compiler-core/src/codegen.rs
+++ b/compiler-core/src/codegen.rs
@@ -6,9 +6,11 @@ use crate::{
     io::FileSystemWriter,
     javascript,
     line_numbers::LineNumbers,
+    sourcemap::{SourceMapEmitter, SourceMapSupport},
     Result,
 };
 use itertools::Itertools;
+use sourcemap::SourceMapBuilder;
 use std::fmt::Debug;
 
 use camino::Utf8Path;
@@ -160,6 +162,7 @@ pub struct JavaScript<'a> {
     output_directory: &'a Utf8Path,
     prelude_location: &'a Utf8Path,
     typescript: TypeScriptDeclarations,
+    sourcemap: SourceMapSupport,
     target_support: TargetSupport,
 }
 
@@ -167,6 +170,7 @@ impl<'a> JavaScript<'a> {
     pub fn new(
         output_directory: &'a Utf8Path,
         typescript: TypeScriptDeclarations,
+        sourcemap: SourceMapSupport,
         prelude_location: &'a Utf8Path,
         target_support: TargetSupport,
     ) -> Self {
@@ -175,16 +179,26 @@ impl<'a> JavaScript<'a> {
             output_directory,
             target_support,
             typescript,
+            sourcemap,
         }
     }
 
     pub fn render(&self, writer: &impl FileSystemWriter, modules: &[Module]) -> Result<()> {
         for module in modules {
             let js_name = module.name.clone();
+            let js_filename = format!("{js_name}.mjs");
             if self.typescript == TypeScriptDeclarations::Emit {
                 self.ts_declaration(writer, module, &js_name)?;
             }
-            self.js_module(writer, module, &js_name)?
+            let source_map_emitter = match self.sourcemap {
+                SourceMapSupport::None => SourceMapEmitter::null(),
+                SourceMapSupport::Emit => {
+                    let mut builder = SourceMapBuilder::new(Some(&js_filename));
+                    let _ = builder.add_source(module.input_path.as_str());
+                    SourceMapEmitter::Emit(builder)
+                }
+            };
+            self.js_module(writer, module, source_map_emitter, &js_name)?;
         }
         self.write_prelude(writer)?;
         Ok(())
@@ -230,10 +244,13 @@ impl<'a> JavaScript<'a> {
         &self,
         writer: &impl FileSystemWriter,
         module: &Module,
+        mut source_map_emitter: SourceMapEmitter,
         js_name: &str,
     ) -> Result<()> {
         let name = format!("{js_name}.mjs");
+        let sourcemap_name = format!("{js_name}.mjs.map");
         let path = self.output_directory.join(name);
+        let sourcemap_path = self.output_directory.join(sourcemap_name);
         let line_numbers = LineNumbers::new(&module.code);
         let output = javascript::module(
             &module.ast,
@@ -242,8 +259,31 @@ impl<'a> JavaScript<'a> {
             &module.code,
             self.target_support,
             self.typescript,
+            &mut source_map_emitter,
         );
         tracing::debug!(name = ?js_name, "Generated js module");
-        writer.write(&path, &output?)
+        writer.write(&path, &output?)?;
+
+        match source_map_emitter {
+            SourceMapEmitter::Null => Ok(()),
+            SourceMapEmitter::Emit(builder) => {
+                // NOTE: This is a bit inefficient
+                // * we first write to a buffer
+                // * then construct a String based on the output
+                // * then write to the output
+                let sourcemap = builder.into_sourcemap();
+                let mut output = Vec::new();
+
+                // TODO: proper error handling
+                sourcemap
+                    .to_writer(&mut output)
+                    .expect("Failed to write sourcemap to memory");
+                // TODO: proper error handling
+                let output =
+                    String::from_utf8(output).expect("Failed to write sourcemap to memory");
+
+                writer.write(&sourcemap_path, &output)
+            }
+        }
     }
 }
--- a/compiler-core/src/codegen.rs
+++ b/compiler-core/src/codegen.rs
@@ -248,9 +248,7 @@ impl<'a> JavaScript<'a> {
         js_name: &str,
     ) -> Result<()> {
         let name = format!("{js_name}.mjs");
-        let sourcemap_name = format!("{js_name}.mjs.map");
         let path = self.output_directory.join(name);
-        let sourcemap_path = self.output_directory.join(sourcemap_name);
         let line_numbers = LineNumbers::new(&module.code);
         let output = javascript::module(
             &module.ast,
@@ -267,20 +265,19 @@ impl<'a> JavaScript<'a> {
         match source_map_emitter {
             SourceMapEmitter::Null => Ok(()),
             SourceMapEmitter::Emit(builder) => {
+                let sourcemap_name = format!("{js_name}.mjs.map");
+                let sourcemap_path = self.output_directory.join(sourcemap_name);
+                tracing::debug!(path = ?sourcemap_path, name = ?js_name, "Emitting sourcemap for module");
+
                 // NOTE: This is a bit inefficient
                 // * we first write to a buffer
                 // * then construct a String based on the output
                 // * then write to the output
                 let sourcemap = builder.into_sourcemap();
-                let mut output = Vec::new();
 
-                // TODO: proper error handling
-                sourcemap
-                    .to_writer(&mut output)
-                    .expect("Failed to write sourcemap to memory");
-                // TODO: proper error handling
-                let output =
-                    String::from_utf8(output).expect("Failed to write sourcemap to memory");
+                let mut output = Vec::new();
+                sourcemap.to_writer(&mut output).expect("Failed to write sourcemap to memory. This is a bug in sourcemap, please report.");
+                let output = String::from_utf8(output).expect("Sourcemap did not generate valid UTF-8. This is a bug in sourcemap, please report.");
 
                 writer.write(&sourcemap_path, &output)
             }
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -9,9 +9,12 @@ use crate::{
     build::Target,
     docvec,
     io::Utf8Writer,
-    parse::extra::{Comment, ModuleExtra},
-    parse::SpannedString,
+    parse::{
+        extra::{Comment, ModuleExtra},
+        SpannedString,
+    },
     pretty::{self, *},
+    sourcemap::SourceMapEmitter,
     type_::{self, Type},
     warning::WarningEmitter,
     Error, Result,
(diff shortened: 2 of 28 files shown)


compiler-core/src/codegen.rs
```diff
@@ -242,8 +257,30 @@ impl<'a> JavaScript<'a> {
...
+
+                // NOTE: This is a bit inefficient
+                // * we first write to a buffer
+                // * then construct a String based on the output
+                // * then write to the output
+                let sourcemap = builder.into_sourcemap();
+
+                let mut output = Vec::new();
+                sourcemap.to_writer(&mut output).expect("Failed to write sourcemap to memory. This is a bug in sourcemap, please report.");
+                let output = String::from_utf8(output).expect("Sourcemap did not generate valid UTF-8. This is a bug in sourcemap, please report.");
```
AlisCode, line 280: Really not a fan of those `expect` - opened to any suggestion here

compiler-core/src/io.rs
```diff
@@ -62,6 +62,62 @@ impl Utf8Writer for String {
     }
 }
 
+/// A wrapper for any writer that keeps track of what line/column it is
+/// currently on.
+///
+/// Used to produce SourceMaps.
+#[derive(Debug)]
+pub struct CursorPositionWriter<'a, W> {
```
AlisCode, line 70: Or move this to `compiler-core/src/io/cursor_position_writer.rs` ?
Also not a fan of the name

compiler-core/src/sourcemap.rs
```diff
@@ -0,0 +1,49 @@
+use sourcemap::SourceMapBuilder;
```
lpil, line 1: Document the code in this module please

compiler-core/src/codegen.rs
lpil: Please remove the bit about opening a bug ticket, all panics have this information printed for them

compiler-core/src/codegen.rs
```diff
@@ -242,8 +257,30 @@ impl<'a> JavaScript<'a> {
...
+            SourceMapEmitter::Null => Ok(()),
+            SourceMapEmitter::Emit(builder) => {
+                let sourcemap_name = format!("{js_name}.mjs.map");
+                let sourcemap_path = self.output_directory.join(sourcemap_name);
+                tracing::debug!(path = ?sourcemap_path, name = ?js_name, "Emitting sourcemap for module");
+
+                // NOTE: This is a bit inefficient
+                // * we first write to a buffer
+                // * then construct a String based on the output
+                // * then write to the output
```
lpil, line 275: Why do we do this? The comment doesn't say

compiler-core/src/codegen.rs
```diff
@@ -242,8 +257,30 @@ impl<'a> JavaScript<'a> {
...
             self.typescript,
+            &mut source_map_emitter,
         );
         tracing::debug!(name = ?js_name, "Generated js module");
-        writer.write(&path, &output?)
+        writer.write(&path, &output?)?;
+
+        match source_map_emitter {
+            SourceMapEmitter::Null => Ok(()),
+            SourceMapEmitter::Emit(builder) => {
```
lpil, line 267: Extract this clause to a helper module please

compiler-core/src/format.rs
```diff
@@ -26,7 +29,12 @@ use camino::Utf8Path;
 
 const INDENT: isize = 2;
 
-pub fn pretty(writer: &mut impl Utf8Writer, src: &EcoString, path: &Utf8Path) -> Result<()> {
+pub fn pretty(
+    writer: &mut impl Utf8Writer,
+    src: &EcoString,
+    path: &Utf8Path,
+    source_map_emitter: &mut SourceMapEmitter,
```
lpil, line 36: Formatting never uses a source map emitter, there should be no change to this API.

compiler-core/src/io.rs
```diff
@@ -62,6 +62,62 @@ impl Utf8Writer for String {
...
+
+impl<'a, W: std::fmt::Write> std::fmt::Write for CursorPositionWriter<'a, W> {
+    fn write_str(&mut self, s: &str) -> std::fmt::Result {
+        self.writer.write_str(s)
+    }
+}
+
+impl<'a, W: Utf8Writer + std::fmt::Write> Utf8Writer for CursorPositionWriter<'a, W> {
+    /// A wrapper around `fmt::Write` that has Gleam's error handling.
+    fn str_write(&mut self, str: &str) -> Result<()> {
```
lpil, line 101: This iterates the string twice. Can we do it once instead?

The pretty printer also iterates over and counts the characters, can that work be reused?

compiler-core/src/io.rs
```diff
@@ -62,6 +62,62 @@ impl Utf8Writer for String {
...
+        if str.is_empty() {
+            return Ok(());
+        }
+        let newline_count = str.rmatches('\n').count();
+        self.line += newline_count;
+        if newline_count > 0 {
+            let lastline = str.lines().last().expect("Should have at least one line");
+            self.column = lastline.len();
```
lpil, line 109: This is number of bytes and not unicode graphemes or characters. Is what sourcemaps expect?

compiler-core/src/javascript/tests/sourcemaps.rs
```diff
@@ -0,0 +1,59 @@
...
+        insta::assert_snapshot!(insta::internals::AutoName, sourcemap_viz);
+    };
+}
+
+#[test]
+fn sourcemap_function_definition() {
+    assert_sourcemap!(
+        "fn add_2(x) {
+  x + 2
+}"
```
lpil, line 57: Is this single case enough to test all the given functionality? I presume it is not only implemented for addition.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_function_definition.snap
```diff
@@ -0,0 +1,50 @@
+---
+source: compiler-core/src/javascript/tests/sourcemaps.rs
+expression: sourcemap_viz
+---
+note:
```
lpil, line 5: This snapshot is rather verbose, to the extent of being hard to read. A more concise format would be very helpful

lpil (commented):
Thank you! I've left a bunch of notes inline

compiler-core/src/io.rs
AlisCode: A sourcemap works with byte indices, in my understanding this is correct. I'll add a test with emojis inside the code, see if that breaks anything.

compiler-core/src/codegen.rs
AlisCode: Could you clarify what is a helper module ? Do you mean that this code belongs in `sourcemap.rs` ?

compiler-core/src/pretty.rs
```diff
@@ -187,6 +192,13 @@ pub enum Document<'a> {
 
     /// A string that is cheap to copy
     EcoString { string: EcoString, graphemes: isize },
+
+    /// Nests the given document and attach a link from the module at the given
+    /// location to the printed document through the sourcemap
+    WithSourceMapLocation {
+        document: Box<Self>,
+        start: LineColumn,
+    },
```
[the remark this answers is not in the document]
AlisCode: When building a sourcemap, for each mapping we want to add, we need two things : 
* The line+column of the mapping in the **generated** (JS) file
* The line+column inside the **source**  (Gleam) file it maps to. 

In my implementation, the field `start` in this enum variant stores the information of where the `Document` is in the source file (a `SrcSpan` passed through the `LineNumbers` util). The only place we know the line+column on the generated JS file is when the pretty printer is writing the `Document` inside a file - then we know **as we're writing** "The document is being printed at the line x, column y", and this info is lost afterwards.

I agree that conceptually the `pretty` module shouldn't know anything about JS, but given that the JS files that Gleam produces are pretty-printed, I don't see how it's possible to generate sourcemaps without hooking into the pretty-printer itself. Any idea I could explore ?

compiler-core/src/io.rs
AlisCode: I've reused `LineColumn` inside of this now. As long as LineNumber does not change behavior (the TODOs mention graphemes which confuses me a bit), then this should work. 

Whenever the behavior of `LineColumn` changes, the test should break :)

compiler-core/src/javascript.rs
```diff
@@ -172,22 +194,29 @@ impl<'a> Generator<'a> {
         // Put it all together
 
         if imports.is_empty() && statements.is_empty() {
-            Ok(docvec![type_reference, "export {}", line()])
+            Ok(docvec![
+                type_reference,
+                "export {}",
+                line(),
+                sourcemap_reference,
```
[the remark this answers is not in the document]
AlisCode: The declaration of the SourceMap URL does not have to be at a specific location, it can be anywhere in the JS code. It is something that's separate from the type reference (which is just TS-specific in my understanding), so I think it's best to keep them as separate declaration blocks for symmetry. 

I take it from this comment that you'd rather have the sourcemap declaration at the top, so I moved it :)

compiler-core/src/javascript/tests/sourcemaps.rs
AlisCode: Before I add more test cases, let's agree on how to test this feature

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_function_definition.snap
AlisCode: I completely agree, but discussions with Gleam folks on discord didn't really lead to a precise conclusion. What exactly would you be happy with? 

Maybe each test could be "what is the Gleam location of this specific location in the JS file" - then the snapshot would contain the whole JS + Gleam code on top of each other, with the one position you're trying to assert? 

As an example, say you write a test case that asks "What is the position in the Gleam code of the character 1 line 1 in the generated JS file ?", the snapshot would look like this: 

```
  ┌─ original.gleam:1:1
  │
1 │ fn add_2(x) {
  │ ^ This code
   |    x + 2
   |  }
  │
  ┌─ generated.js:2:1
  │
2 │ function add_2(x) {
  │ ^ Gets mapped to this
   |     return x + 2;
   |  }
}
```

AlisCode:
I think I need your opinion on stuff before I can write a bunch of unit tests or refactor the PR, so I'm marking this as ready.

lpil:
Hello! Are you waiting on something from me? I can't see anything here but wasn't sure if I missed anything.

AlisCode:
Hi @lpil, 

this PR got a bit old and so it has a few merging conflicts that I'll solve today after work.

There are 2 points which needs clarification before I proceed to add more unit tests : 

https://github.com/gleam-lang/gleam/pull/3675#discussion_r1790679243 
About this comment : The approach I took in this PR is that we're adding a variant to the pretty.rs module, in order for the information on Line + Column to stay relevant even after a prettification pass. Indeed, sourcemaps work by referencing columns+lines in generated code, so if the pretty module is not aware of sourcemaps, I don't see how this is implementable since it would be referencing code that may no longer be in the same place.   

https://github.com/gleam-lang/gleam/pull/3675#discussion_r1815028347
About this comment : Is this approach good enough ? If it is, then I'll add some more test cases :)

cdaringe:
I noticed that if i already have a compiled build cache, then latest i switch to this compiler with sourcemaps turned on, maps arent emitted. i have to run a clean first, presumably to bypass a cache hit that does not consider this flag as input to the cache?

cdaringe:
Im on mobile, thus didn’t search the source, but iff the build cache algo can consider new inputs, im suggesting that the config options added here are missing in that input. It may ir may not be open for extension. I sorta kinda would think that a hash on gleam.toml would invalidate the build cache?

lpil:
Hello! Are you still working on this? Thank you

AlisCode:
Hi @lpil, 

Unfortunately I'm no longer actively working on this. It's become clear to me that I should start with smaller contribs because I don't see how to implement this feature based on the feedback provided in the PR - that, and the path towards actually writing tests is a bit unclear. I also lack the bandwidth, the next few months are probably gonna be a bit busy for me.

lpil:
Thank you, sorry this didn't make it in!

PR #3676, opened 2024-10-07
Add `echo` keyword

giacomocavalieri:
This PR adds an implementation for the `echo` keyword.

Following from this discussion (<https://github.com/gleam-lang/gleam/discussions/3625>) here's a TODO list of things I have to do:
- [ ] parse `echo`
- [ ] pretty print `echo`
- [ ] type `echo`
  - [ ] special handle an `echo` in the middle of a pipeline
  - [ ] make `echo` cannot be used without an expression unless it's as a step of a pipeline
- [ ] generate code for `echo`
  - [ ] copy `io.debug`'s look
- [ ] make sure code with `echo` doesn't get published
- [ ] deprecate `io.debug` from `stdlib`

And here's a list of things I've intentionally left out, that could be added in future PRs (for which we could open separate issues):
- [ ] allow to add a tag with `as "tag"`
- [ ] improve the pretty printing generated code based on type information (include labels, tell apart tuples from custom types, and strings from bitarrays)
- [ ] add other information like the location where the code comes from


inoas:
Can we get the info from echo without actually pushing it to stdout/stderr for libs like pprint?

lpil:
No.

inoas:
> No.

will string.inspect stay and just io.debug be gone?
- if string.inspect stays - do we have two different implementations to inspect at runtime?
- if string.inspect is also removed, would pprint require to impl string.inspect, pprint is very useful with birdie.

lpil:
There will be no changes to `string.inspect` coming from this.

giacomocavalieri:
As far as I'm aware there's no plan to remove `string.inspect`, just `io.debug` since we don't want two ways to do the same things and just sticking to `echo` will give you guarantees like "the build tool won't let you publish a package with debug printing".

Also `string.inspect` and `echo` cover two different cases: the former is for getting a string from any piece of data, while the latter is for debug printing and nothing else.
It's output might change or diverge from what `string.inspect` is doing by taking advantage of type informations from the compiler, the discussion where we talked about it is here <https://github.com/gleam-lang/gleam/discussions/3625>

inoas:
> As far as I'm aware there's no plan to remove `string.inspect`, just `io.debug` since we don't want two ways to do the same things and just sticking to `echo` will give you guarantees like "the build tool won't let you publish a package with debug printing".
> 
> Also `string.inspect` and `echo` cover two different cases: the former is for getting a string from any piece of data, while the latter is for debug printing and nothing else. It's output might change over time or diverge from what `string.inspect` is doing by taking advantage of type informations from the compiler, the discussion where we talked about it is here #3625

Yes. It would just be nice to utilize what echo does in the long run for things like pprint etc instead of keeping 2 implementations around, imho.

Thank you for working on this <3

lpil:
They will always have to be 2 implementations as they do 2 different things.

graphiteisaac:
wondering if it would be worth adding something to either the LSP or formatter to replace io.debug calls with echo, to ease the transition post deprecation? not a necessary thing by any means but could be a nice to have DX addition

lpil:
Probably not worth the effort given `io.debug` calls tend to be removed after debugging is finished

compiler-core/src/parse/tests.rs
```diff
@@ -466,6 +466,52 @@ fn no_eq_after_binding1() {
     );
 }
 
+#[test]
```
lpil, line 469: Can we also have a test for `echo 1 + 1` please 🙏

giacomocavalieri: Should that be the same as `echo { 1 + 1 }` or `{ echo 1 } + 1`?

giacomocavalieri:
This is still missing some integration tests to check the actual echo's output but I'd like to get some feedback on the code so far

compiler-core/src/parse/tests.rs
GearsDatapacks: My intuition says `echo { 1 + 1 }` personally

giacomocavalieri: Mmh I would have said the opposite

Olian04: What about the operator precedence with other operators? Such as `echo 1 * 1` or `echo 1 == 1`?

GearsDatapacks: I would imagine echo to work a bit like let, where it has the lowest precedence. But I guess they are different things, since let is never used in sub-expressions. I would say for minimum confusion it should either have lowest precedence or highest precedence, so you can easily predict its behaviour

Olian04: Agreed. Either **highest** or **lowest**, but always consistent. My argument for **highest** is that it simplifies use cases. For example: Consider A) `echo 1 + 1` and B) `1 + echo 1`. If the echo keyword has the **lowest** precedence, the case A would output 2, equivalent to `{echo { 1 + 1 }}` while case B would output 1, equivalent to `1 + {echo { 1 }}`. However if it has the lowest precedence, then these cases would both output 1 and would both be equivalent to `{echo { 1 }} + 1` and `1 + {echo { 1 }}`.

GearsDatapacks: I guess if echo is considered to be a prefix operator, like `!` or `-`, it would make more sense to have higher precedence. But as a keyword, it feel like it should have lower precedence (e.g. `return` in other languages)

Olian04: I don't think it really makes sense comparing `echo` to `return`, seeing as `echo` is an expression and `return` in a statement.

lpil: It should have the lowest so that you can easily add and remove `echo` to the start of the line and print the whole line. Having to wrap expressions in `{}` would be irritating.

> My argument for highest is that it simplifies use cases in sub-expressions. For example: Consider A) echo 1 + 1 and B) 1 + echo 1.

My intention was for `1 + echo 1` to be a syntax error, but perhaps I can be persuaded otherwise.

compiler-cli/src/publish.rs
```diff
@@ -303,18 +303,27 @@ fn do_build_hex_tarball(paths: &ProjectPaths, config: &mut PackageConfig) -> Res
...
+        } else if module.ast.type_info.contains_echo {
+            modules_containing_echo.push(module.name.clone());
+        }
+    }
+
+    if !modules_containing_todo.is_empty() {
+        return Err(Error::CannotPublishTodo {
+            unfinished: modules_containing_todo,
+        });
+    } else if !modules_containing_echo.is_empty() {
```
lpil, line 323: No need for `else` with a block that returns.

```suggestion
    }
    
    if !modules_containing_echo.is_empty() {
```

compiler-core/src/ast.rs
```diff
@@ -2331,3 +2331,22 @@ pub struct UseAssignment {
     pub pattern: UntypedPattern,
     pub annotation: Option<TypeAst>,
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct TypedPipelineAssignment {
```
lpil, line 2336: Document what this is please 🙏

compiler-core/src/erlang/tests/echo.rs
```diff
@@ -0,0 +1,124 @@
+macro_rules! assert_generic_echo {
```
lpil, line 1: What's a generic echo? This is called "assert Erlang" in the other modules. Any reason to not to have this functionality in all codegen tests?

compiler-core/src/javascript.rs
```diff
@@ -169,25 +169,49 @@ impl<'a> Generator<'a> {
...
+                &Some((
+                    AssignName::Variable("stdlib$dict".into()),
+                    SrcSpan::default(),
+                )),
+                &[],
+            );
+            self.register_prelude_usage(&mut imports, "BitArray", None);
+            self.register_prelude_usage(&mut imports, "List", None);
+            self.register_prelude_usage(&mut imports, "UtfCodepoint", None);
+            self.register_prelude_usage(&mut imports, "CustomType", None);
```
lpil, line 188: What if the programmer has defined types with these names?

compiler-core/src/javascript/tests/echo.rs
```diff
@@ -0,0 +1,121 @@
+macro_rules! assert_generic_echo {
```
lpil, line 1: Same here as with the Erlang

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__echo__echo_in_pipeline_printing_a_function_2.snap
```diff
@@ -0,0 +1,17 @@
+---
+source: compiler-core/src/type_/tests/echo.rs
+expression: "\npub fn main() {\n  [1, 2, 3]\n  |> echo wibble\n}\n\nfn wibble(_: List(Float)) -> List(String) { todo }\n"
+---
+error: Type mismatch
+  ┌─ /src/one/two.gleam:4:6
+  │
+4 │   |> echo wibble
+  │      ^^^^^^^^^^^ This function does not accept the piped type
```
lpil, line 9: I don't think we should be calling `echo` a function.

compiler-core/src/warning.rs
```diff
@@ -957,6 +957,8 @@ Your code will crash before reaching this point.",
                         PanicPosition::LastFunctionArgument =>
                             "This function call is unreachable because its last argument always panics. \
 Your code will crash before reaching this point.",
+                        PanicPosition::EchoExpression =>
+                            "This `echo` won't print anything because the expression it should be printing always panics.",
```
lpil, line 961: This line seems like it might be long enough to break the formatting in this code

compiler-core/templates/echo.mjs
```diff
@@ -0,0 +1,100 @@
...
+    const args = [];
+    for (const i of Array(v.length).keys()) args.push(String.fromCharCode(i + 97));
+    return `//fn(${args.join(", ")}) { ... }`;
+  }
+  return $inspectObject(v);
+}
+
+function $isDict(value) {
+  try {
+    return value instanceof $stdlib$dict.default;
```
lpil, line 96: Perhaps add a comment to explain what's happening here. It's a little cryptic

lpil (commented):
Looks great!!! I've left some comments

compiler-core/src/parse/tests.rs
GearsDatapacks: I think it could be nice for echo to be allowed as an expression, like rust's `dbg!` macro

compiler-core/src/erlang/tests/echo.rs
giacomocavalieri: It's the specific echo implementation that works on any generic type by doing runtime inspection, in the future the compiler could cut down on this code by only adding the implementation needed for a specific type if it can tell it at compile time:
```gleam
pub fn main() {
  echo 1
}
```
Could just become (it might be wrong but you get the idea):
```erl
main() ->
  echo(1, fun inspect@int/1).

echo(N, Inspect) ->
  String = Inspect(N),
  io:put_chars(standard_error, [String, $\n]),
  N.
  
inspect@int(N) ->
  erlang:integer_to_list(N).
```
So here I would want to just `assert_erl!` and look at the entirety of the code unlike in the generic case where I can just check that it contains that big chunk of code.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__echo__echo_in_pipeline_printing_a_function_2.snap
giacomocavalieri: Ooh this slipped me, it should only point to `wibble` in this case I think

compiler-core/src/parse/tests.rs
giacomocavalieri: Ok, I made `echo` lower precedence than binary operations!
Also @lpil we do have a test for `echo 1 + 1` here <https://github.com/gleam-lang/gleam/pull/3676/files#diff-35d6f2cca3b4773466d26264b7d213f97615c5de54ce0a5e7725593aca7e5207R485>

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__echo__echo_in_pipeline_printing_a_function_2.snap
lpil: oooooooh so that's what the error is

giacomocavalieri: Ah wait you can't this is an old snapshot that I forgot to delete I think!

lpil: Thank you.

It would be cool if insta could clean up old snapshots

compiler-core/src/parse/tests.rs
Olian04: Ok, so the expected behaviour is for `echo` to be an expression with the lowest possible precedence?

giacomocavalieri: Yeah, `echo 1 + 1` is the same as `echo { 1 + 1 }`

Olian04: Would then also `let a = echo 1 + 1` be equivalent too `let a = 1 + 1` then `echo a`?

GearsDatapacks: Yes it will

Cargo.toml
```diff
@@ -1,12 +1,13 @@
...
-  "test-helpers-rs",
-  "test-package-compiler",
-  "test-project-compiler",
+    "compiler-cli",
+    "compiler-core",
+    "compiler-wasm",
+    "test-helpers-rs",
+    "test-package-compiler",
+    "test-project-compiler",
+    "test-echo-output",
```
lpil, line 10: Rather than making this specific to echo let's make this a set of tests related to what running a program prints.

lpil (commented):
I found a bug! I've pushed a couple of test cases.

The integration test lag is too much, we'll need to do something else there. I'm thinking we invoke the compiler-cli entrypoint directly from within Rust code rather than in a different OS process

GearsDatapacks:
It does currently

giacomocavalieri:
@lpil did you manage to have a look at those slow integration tests?

giacomocavalieri:
> Lots of test failures- do you know what has happened there?

It's the integration snapshot tests failing because stdlib has many warnings, I'll fix it once we change the way we run the integration tests, otherwise any work done now would be thrown away anyway once we run the compiler directly instead of shelling out.

> Not yet, sorry. I have a plan but have not had time yet.

If you want to share a bit more I could try and implement it myself.

(I've also rebased and fixed the new conflicts)

lpil:
I should have done that ages ago! Sorry!

1. Split the `gleam run` shell command building from the actual execution
1. Call the project build function in each test
1. Use the now-split command building in each test
1. Run the command using the Rust stdlib exec API and capture IO
2. Snapshot output

giacomocavalieri:
I'm not sure I fully understand how point 1-3 will work but I can give this a try in the meantime if you don't have the time

lpil:
> Also where would the split part of gleam run end up? One thing that was blocking me is that I couldn't use the compiler-cli module as a dependency, so does that mean that that part should end up in compiler-core?

Probably easiest to put the tests in compiler-cli?

giacomocavalieri:
> Probably easiest to put the tests in compiler-cli?

Ah! That makes sense, I don't really know how to have a crate inside a crate (if that's even possible) or what that would look like with the autogenerated tests. I'll just keep resolving conflicts until you can give this a shot

compiler-core/src/javascript/expression.rs
```diff
@@ -515,12 +540,45 @@ impl<'module> Generator<'module> {
...
+            match assignment.value.as_ref() {
+                // An echo in a pipeline won't result in an assignment, instead it
+                // just prints the previous variable assigned in the pipeline.
+                TypedExpr::Echo {
+                    expression: None,
+                    location,
+                    ..
+                } => documents.push(self.not_in_tail_position(|gen| {
+                    let var = gen
+                        .latest_local_var
```
lpil, line 552: Why is this state on the generator when it's specific to a pipeline? This would be the first time that function specific state is kept at the top level.

lpil (commented):
Exciting!! I've left a few notes below

lpil (approved):
Thank you Jak!!!!

And it only took 4 months! Easy!

PR #3682, opened 2024-10-09
Implement code action to add type annotations

### LINKED ISSUE #2953: LSP code action: add annotation to assignment

### LINKED ISSUE #2954: LSP code action: add annotations to module function
Follows on from https://github.com/gleam-lang/gleam/issues/2953

GearsDatapacks:
Closes #2953, closes #2954
This PR adds a code action which automatically adds type annotations to functions, constants and local variables

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,23 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The Language Server now suggests a code action to add type annotations to
+  local variables, constants and functions:
+
+  ```gleam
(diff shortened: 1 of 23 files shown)


GearsDatapacks:
Actually I should make this work with `use` too

GearsDatapacks:
Actually, due to the way that `use` is currently implement, that won't be so easy. Should I rework the AST a bit to allow that or just leave it for now?

lpil (approved):
Beautiful work!!!

`use` would be cool but seems trickier. Can be a later addition

PR #3692, opened 2024-10-11
Fix wrong error message for values in pipe

### LINKED ISSUE #3686: Wrong error message for values in a pipe
If I write the following code:
```gleam
pub fn main() {
  wibble
  |> "a"
}

fn wibble(_: Int) -> String { todo }
```
I get the following error message:
```txt
error: Type mismatch
  ┌─ /src/main.gleam:5:6
  │
5 │   |> "a"
  │      ^^^

Expected type:

    String

Found type:

    fn(fn(Int) -> String) -> a
```
It looks like it has the expected and found types swapped.

MarkusPettersson98:
This PR fixes #3686.

I added a test which is very similar to the case in the reported issue. I am a bit unsure about my last commit which basically unconditionally applies the technique from https://github.com/gleam-lang/gleam/pull/3453. It seems fine since all tests still pass, but please tell me if I'm wrong ¯\\_(ツ)_/¯.

### DIFF
--- a/compiler-core/src/type_/pipe.rs
+++ b/compiler-core/src/type_/pipe.rs
@@ -295,16 +295,11 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
             fn_(vec![self.argument_type.clone()], return_type.clone()),
         )
         .map_err(|e| {
-            if self.check_if_pipe_function_mismatch(&e) {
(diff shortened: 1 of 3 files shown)


lpil (approved):
If it works, it works!

Could you update the changelog please. Thank you

MarkusPettersson98:
Done!

lpil (approved):
Thank you!!

Issue #3691, opened 2024-10-11
Compiler error about "unknown module value" could highlight existence of type with the same name


michaeljones:
*As reported on [Discord](https://discord.com/channels/768594524158427167/768594524158427170/1294276312541691935).*

I'm currently getting an error:

```
error: Unknown module value
   ┌─ /home/michael/root/projects/gleam-lang/plug/src/wisp/plug.gleam:57:7
   │
57 │   wisp.Request(
   │       ^^^^^^^^

The module `wisp` does not have a `Request` value.
```

It might be helpful it also said something like "There is a Request **type** in wisp but that cannot be constructed directly. You need to import a type constructor..." Or something like this? With the right words? And better?

Or maybe that is just too complex in this case as wisp.Request is a type alias for something that isn't normally created directly?

lpil:
Something like "Xyz is a type/type constructor, it cannot be used as a value"?

PR #3696, opened 2024-10-12
feat: check if module have type with the same name

Frank-III:
resolve: #3691 

the compiler already suggests something like, do we want to change it?
```
Did you mean `type Request`?

`Request` is only a type, it cannot be imported as a value.
```

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -2151,7 +2151,7 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                         },
                         module_name: module.name.clone(),
                         value_constructors: module.public_value_names(),
-                        type_with_same_name: false,
+                        type_with_same_name: module.get_public_type(&label).is_some(),
                     })?;
 
             // Emit a warning if the value being used is deprecated.
@@ -2465,7 +2465,7 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                         module_name: module_name.clone(),
                         name: name.clone(),
                         value_constructors: module.public_value_names(),
-                        type_with_same_name: false,
+                        type_with_same_name: module.get_public_type(name).is_some(),
                     })?
             }
         };


giacomocavalieri:
I don't think the suggestion is right here: "Did you mean type X" would produce invalid code outside of imports, the rest of the error message looks good to me though :)

Frank-III:
> I don't think the suggestion is right here: "Did you mean type X" would produce invalid code outside of imports, the rest of the error message looks good to me though :)

This is what I thought! maybe we need to differentiate when we are importing or accessing it🤔

giacomocavalieri:
Yeah I think that's the right call

lpil:

Frank-III:
Hi! sorry for kinda forget about this one, as discussed with @giacomocavalieri I replace `type_with_same_name` with a context to differentiate `usage` or `import`, but I am not very sure about the new error message for the opaque constructor now

compiler-core/src/type_/error.rs
```diff
@@ -200,7 +200,7 @@ pub enum Error {
         name: EcoString,
         module_name: EcoString,
         value_constructors: Vec<EcoString>,
-        type_with_same_name: bool,
+        context: ModuleValueErrorContext,
```
lpil, line 203: Let's split this field into two and keep the previous `type_with_same_name`, rather than having the same bool in each.

```rust
pub enum ModuleValueUsageContext {
    UnqualifiedImport
    ModuleAccess
}
```

compiler-core/src/error.rs
```diff
@@ -2189,12 +2190,14 @@ Private types can only be used within the module that defines them.",
...
                 } => {
-                    let text = if *imported_value_as_type {
-                        format!("`{name}` is only a type, it cannot be imported as a value.")
-                    } else {
-                        format!("The module `{module_name}` does not have a `{name}` value.")
+                    let text = match context {
+                        ModuleValueErrorContext::Import { type_with_same_name: true } =>
+                            format!("`{name}` is only a type, it cannot be imported as a value."),
+                        ModuleValueErrorContext::Usage { type_with_same_name: true } =>
+                            format!("{module_name}.{name} is a type/type constructor, it cannot be used as a value"),
```
lpil, line 2199:
```suggestion
                            format!("{module_name}.{name} is a type, it cannot be used as a value"),
```

compiler-core/src/error.rs
```diff
@@ -2189,12 +2190,14 @@ Private types can only be used within the module that defines them.",
...
                     module_name,
                     value_constructors,
-                    type_with_same_name: imported_value_as_type,
+                    context,
                 } => {
-                    let text = if *imported_value_as_type {
-                        format!("`{name}` is only a type, it cannot be imported as a value.")
-                    } else {
-                        format!("The module `{module_name}` does not have a `{name}` value.")
+                    let text = match context {
```
lpil, line 2195: Use the `wrap_format` macro to make these please, or call the wrap function.

compiler-core/src/error.rs
```diff
@@ -2189,12 +2190,14 @@ Private types can only be used within the module that defines them.",
...
+                        ModuleValueErrorContext::Import { type_with_same_name: true } =>
+                            format!("`{name}` is only a type, it cannot be imported as a value."),
```
lpil, line 2197: How about we have it show how to import it as a type, using the `type` keyword?

lpil (commented):
Very nice! I've left some comments inline.

lpil (approved):
Excellent work again! Thank you Frank!!

PR #3699, opened 2024-10-14
Fix prelude re-export in generated TypeScript definitions

richard-viney:
When using the generated `gleam.d.mts` I was seeing this error:

```bash
gleam.d.mts:1:15 - error TS2846: A declaration file cannot be imported without 'import type'. Did you mean to import an implementation file '../prelude.mjs' instead?

1 export * from "../prelude.d.mts";
 ```

The fix seems to be changing the prelude re-export to this:

```ts
export * from "../prelude.mjs";
export type * from "../prelude.d.mts";
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,9 @@
   types that use a private type.
   ([Richard Viney](https://github.com/richard-viney))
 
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you.

PR #3701, opened 2024-10-14
Fix TypeScript definitions for records with identical structures not being branded

richard-viney:
This PR addresses two issues with the generated TypeScript typings:

1. `pub type Dict(key, value)` generating the TypeScript `export type Dict<A, B> = any`. This isn't ideal because it uses `any` which (a) loses type information and (b) triggers downstream linters set to warn on any kind of interaction with an `any` type.

2. Gleam types with identical structure having structurally identical TypeScript types. This leads to being able to accidentally pass incorrect and invalid types to functions when relying on the TypeScript typings, with no errors or warnings being shown.

The proposed solution implemented here is to add branding to all classes in the generated TypeScript definitions. See https://www.learningtypescript.com/articles/branded-types for details.

The following Gleam code:

```gleam
pub type Foo

pub type TestWithVariants {
  TestOne
  TestTwo
}
```

now generates this TypeScript:

```typescript
declare class Foo extends _.CustomType {
  private __brand: void;
}

export type Foo$ = Foo;

export class TestOne extends _.CustomType {
  private __brand: void;
}

export class TestTwo extends _.CustomType {
  private __brand: void;
}

export type TestWithVariants$ = TestOne | TestTwo;
```

### DIFF
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bools__shadowed_bools_and_nil_typescript.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bools__shadowed_bools_and_nil_typescript.snap
@@ -4,11 +4,17 @@ expression: "\npub type True { True False Nil }\npub fn go(x, y) {\n  let assert
 ---
 import type * as _ from "../gleam.d.mts";
 
-export class True extends _.CustomType {}
+export class True extends _.CustomType {
+  private __brand: void;
+}
 
-export class False extends _.CustomType {}
+export class False extends _.CustomType {
+  private __brand: void;
+}
 
-export class Nil extends _.CustomType {}
+export class Nil extends _.CustomType {
+  private __brand: void;
+}
 
 export type True$ = True | False | Nil;
 
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__records__record_with_no_variants_typescript.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__records__record_with_no_variants_typescript.snap
@@ -1,0 +1,9 @@
+---
+source: compiler-core/src/javascript/tests/records.rs
+expression: "\npub type Dict(key, value)\n"
+---
+declare class Dict<I, J> extends _.CustomType {
+  private __brand: void;
+}
+
+export type Dict$<I, J> = Dict<I, J>;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,10 @@
 - Fixed the prelude re-export in generated TypeScript definitions.
   ([Richard Viney](https://github.com/richard-viney))
 
+- Fixed TypeScript definitions for records with identical structures not being
+  nominally typed.
+  ([Richard Viney](https://github.com/richard-viney))
+
 ## v1.5.1 - 2024-09-26
 
 ### Bug Fixes
--- a/compiler-core/src/javascript/tests/records.rs
+++ b/compiler-core/src/javascript/tests/records.rs
@@ -1,4 +1,4 @@
-use crate::assert_js;
+use crate::{assert_js, assert_ts_def};
 
 #[test]
 fn record_accessors() {
@@ -66,5 +66,26 @@ pub type Person {
 }
 pub fn get_name(person: Person) { person.name }
 pub fn get_age(person: Person) { person.age }"
+    );
+}
+
+#[test]
+fn record_with_no_variants_typescript() {
+    assert_ts_def!(
+        r#"
+pub type Dict(key, value)
+"#,
+    );
+}
+
+#[test]
+fn record_with_variants_typescript() {
+    assert_ts_def!(
+        r#"
+pub type Test {
+  Test1()
+  Test2()
+}
+"#,
     );
 }
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__custom_types__const_with_fields_typescript.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__custom_types__const_with_fields_typescript.snap
@@ -5,6 +5,8 @@ expression: "\npub type Mine {\n  Mine(a: Int, b: Int)\n}\n\npub const labels = 
 import type * as _ from "../gleam.d.mts";
 
 export class Mine extends _.CustomType {
+  private __brand: void;
+
   constructor(a: number, b: number);
   
   a: number;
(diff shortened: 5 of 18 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bools__shadowed_bools_and_nil_typescript.snap
```diff
@@ -4,11 +4,17 @@ expression: "\npub type True { True False Nil }\npub fn go(x, y) {\n  let assert
 ---
 import type * as _ from "../gleam.d.mts";
 
-export class True extends _.CustomType {}
+export class True extends _.CustomType {
+  private __gleam__my__mod__True: never;
```
lpil, line 8: These values do not have this field, so this would be incorrect. Is there a way to make these types nominal without making them incorrect?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__records__record_with_no_variants_typescript.snap
```diff
@@ -0,0 +1,7 @@
+---
+source: compiler-core/src/javascript/tests/records.rs
+expression: "\npub type Dict(key, value)\n"
+---
+export type Dict$<I, J> = {
+  __gleam__my__mod__Dict: never;
```
lpil, line 6: This is not correct. The dict external type does not have this field.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bools__shadowed_bools_and_nil_typescript.snap
richard-viney: Hmm yes, this is sort of the point though, i.e. the desired type safety is then enforced by TypeScript _without_ the runtime overhead of actually creating these unused fields for real.

Making the types `private` and of type `never` is intended to stop anyone ever trying to access them. They'd get `undefined` were they to do so. Maybe a type of `undefined` instead of `never` would therefore be better? Or a name that more explicitly indicates not to access the field?

I'll have a think about whether there's an alternative approach here that doesn't involve creating these fields on the underlying object.

lpil: It's not the value of the field, it's that there is a field at all. There is no field with that name on that type.

Why it is that TypeScript considers them the same? It woudn't be the same in TypeScript at runtime as they have different constructors and so would not pass instanceof checks.

richard-viney: 100% there's no field with that name on the type. Here's a hopefully clearer illustration of the problem that's trying to be addressed by this change.

If I write this Gleam code:

```gleam
pub type Thing {
  Thing
}
```

Its TypeScript will be:

```ts
import type * as _ from "../gleam.d.mts";

export class Thing extends _.CustomType {}

export type Thing$ = Thing;
```

Note also that Gleam's `option.Option` type compiles to this TypeScript:

```ts
export class Some<I> extends _.CustomType {
  constructor(argument$0: I);
  0: I;
}

export class None extends _.CustomType {}

export type Option$<I> = Some<I> | None;
```

Now, because `option.None` and `thing.Thing` are structurally identical, they're interchangeable in TypeScript. The fact that they come from different constructors doesn't matter. It's true that `instanceof` checks are all correct, but this is about type-checking arguments to functions which is based on a type's structure, not its constructor (which is what `instanceof` checks).

Right now in TypeScript, an instance of _any_ variant of _any_ type can be passed to an argument of type `option.Option`, because they are _all_ structurally equivalent with `option.None`. This really undermines type-safety when working with the generated TypeScript definitions.

E.g. the following shouldn't type-check and run in TypeScript, but it does:

```ts
import * as thing from "../thing/thing.mjs";
import * as option from "../gleam_stdlib/gleam/option.mjs";

console.log(`Is thing a none? ${option.is_none(new thing.Thing())}`); // Prints false
console.log(`Is thing a some? ${option.is_some(new thing.Thing())}`); // Prints true
```

The relevant docs are here: https://www.typescriptlang.org/docs/handbook/type-compatibility.html.

richard-viney:
The type-only private field is now a constant string which is a bit simpler:

```ts
export class Test1 extends _.CustomType {
  private __gleam_type: "my/mod.Test1";
}
```

Also types with no constructors now generate a non-exported class so that they can use a private member:

```ts
class Dict extends _.CustomType {
  private __gleam_type: "my/mod.Dict";
}

export type Dict$ = Dict;
```

richard-viney:
Turns out that this technique is more commonly referred to as "branding" in the TypeScript ecosystem:

1. https://www.learningtypescript.com/articles/branded-types
2. https://egghead.io/blog/using-branded-types-in-typescript
3. https://effect.website/docs/guides/style/branded-types

Also, with the change in this PR to have Gleam's TS types always declare classes, it's possible to rely on the fact that different classes with the same structure but with a _private_ member are _not_ structurally equivalent (see https://www.typescriptlang.org/docs/handbook/type-compatibility.html#private-and-protected-members-in-classes). This means the only branding code needed in each type is:

```ts
  private __brand: void
```

I've updated the PR to use this more minimal approach.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bools__shadowed_bools_and_nil_typescript.snap
lpil: Those are different classes with differing constructors and as such are not structurally equivalent. Specifically the prototype property differs.

richard-viney: Unfortunately the prototype isn't considered as part of structural typing in TypeScript! If it was then indeed this issue wouldn't exist. See [this TS playground](https://www.typescriptlang.org/play/?#code/MYGwhgzhAEDCYBdoG8BQ1oDswFsCm0AvNAEQBmIArmWQJ4moC+qqokMAIgPYDmK6WXAWLkAlgBMuDZqjKVMwBKK6ZowABZ5gAa3gIAFMEQAuOIgCU-DMBUQuIPADoQvfSSMJH2fEVLQA1GqIXkLmANwCNph2Ds6u7ojQotEIYAp4XGRmSCIB0IaJyRCp6ZnZ5uGRtvZOLjxuHkkpacAZWdx8uYEFSEUlrWUdFREyGlq6iPqYeADu2frDrJo6elOz0B0LlUA) for a simple example. I was surprised by this behaviour initially, I suspect many TypeScript devs would be.

To try and simplify things I've opened a separate PR that only tries to address the presence of the `any` type in the generated typings: https://github.com/gleam-lang/gleam/pull/3707. It uses a somewhat different approach to achieve the desired result in that case.

If that PR works out then I'll look into adapting this one, converting to a draft for now.

PR #3704, opened 2024-10-16
Fix bug with labelled argument order

### LINKED ISSUE #3671: Unexpected argument order on calls with labelled parameters
Hi!

I encounter what I believe to be a bug if the two following properties hold:
- Unlabelled argument order doesn’t differ from function’s parameter order
- Labelled argument order doesn’t matters

Here is a reduction of my original code that motivated me to digg into this issue. The bug is kind of sneaky because two String parameters are silently reversed when a third (labelled) argument corresponding to the first function parameter is added.

```gleam
// This compiles fine
pub fn heading_test() {
  heading("Hello", "/world", size: 1)
}

fn heading(size size: Int, text text: String, href href: String) {
  // Crashes here
  let assert "Hello" = text
  let assert "/world" = href
}
```

----

Using different types to make the issue more obvious. Every line below with a `Error expecting XXX` comment doesn’t type check.

```gleam
type A { A }
type B { B }
type C { C }
type D { D }

fn foo(a a: A, b b: B, c c: C, d d: D) {
  Nil
}

pub fn labelled_call_issue_test() {
  // Calls with a single labelled argument
  foo(A, B, C, d: D)
  foo(A, B, D, c: C)
  foo(A, C, D, b: B) // Error expecting A, D, C, b: B
  foo(B, C, D, a: A) // Error expecting D, B, C, a: A

  // Calls with two labelled arguments
  foo(A, B, c: C, d: D)
  foo(A, B, d: D, c: C)
  foo(A, C, b: B, d: D)
  foo(A, C, d: D, b: B)
  foo(A, D, b: B, c: C)
  foo(A, D, c: C, b: B)
  foo(B, C, a: A, d: D) // Error expecting C, B, a: A, d: D
  foo(B, C, d: D, a: A) // Error expecting C, B, d: D, a: A
  foo(B, D, a: A, c: C) // Error expecting D, B, a: A, c: C
  foo(B, D, c: C, a: A) // Error expecting D, B, c: C, a: A
  foo(C, D, a: A, b: B)
  foo(C, D, b: B, a: A) // Error expecting D, C, b: B, a: A -> labelled argument order matters here

  // No issue with any permutation of a 3 labelled arguments call
  foo(A, b: B, c: C, d: D)
  foo(A, b: B, d: D, c: C)
  foo(A, c: C, b: B, d: D)
  foo(A, c: C, d: D, b: B)
  foo(A, d: D, b: B, c: C)
  foo(A, d: D, c: C, b: B)
  foo(B, a: A, c: C, d: D)
  foo(B, a: A, d: D, c: C)
  foo(B, c: C, a: A, d: D)
  foo(B, c: C, d: D, a: A)
  foo(B, d: D, a: A, c: C)
  foo(B, d: D, c: C, a: A)
  foo(C, a: A, b: B, d: D)
  foo(C, a: A, d: D, b: B)
  foo(C, b: B, a: A, d: D)
  foo(C, b: B, d: D, a: A)
  foo(C, d: D, a: A, b: B)
  foo(C, d: D, b: B, a: A)
}

```

```shell
❯ gleam --version && uname -v
gleam 1.5.0
Darwin Kernel Version 23.6.0: Wed Jul 31 20:48:52 PDT 2024; root:xnu-10063.141.1.700.5~1/RELEASE_ARM64_T6020
```

_Sonoma 14.7 (23H124)_

GearsDatapacks:
Fixes #3671 
This fixes a bug with the reordering of arguments where reordering was just done using `[T]::swap`. This works for simple cases of argument reordering, but breaks down in quite a few cases as shown in that issue.
This implements a more complex reordering method, which removes the labelled arguments first, then reinserts them in order.

A couple of questions:
- I've just put all the erroneous cases from that issue into one test, should I separate them?
- This new method is not super efficient, since is has to iterate the arguments, remove any labelled ones, then iterate again to insert them back. I can think of a few other ways to do this, although they are a bit obscure and could make the code difficult to understand. Let me know if I should change it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -172,6 +172,10 @@
 - Fixed the prelude re-export in generated TypeScript definitions.
   ([Richard Viney](https://github.com/richard-viney))
 
+- Fixed a bug where the compiler would incorrectly type-check and compile
+  calls to functions with labelled arguments in certain cases.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.5.1 - 2024-09-26
 
 ### Bug Fixes
--- a/compiler-cli/src/panic.rs
+++ b/compiler-cli/src/panic.rs
@@ -1,13 +1,13 @@
 #![allow(clippy::unwrap_used)]
-use std::panic::PanicInfo;
+use std::panic::PanicHookInfo;
 
 pub fn add_handler() {
-    std::panic::set_hook(Box::new(move |info: &PanicInfo<'_>| {
+    std::panic::set_hook(Box::new(move |info: &PanicHookInfo<'_>| {
         print_compiler_bug_message(info)
     }));
 }
 
-fn print_compiler_bug_message(info: &PanicInfo<'_>) {
+fn print_compiler_bug_message(info: &PanicHookInfo<'_>) {
     let message = match (
         info.payload().downcast_ref::<&str>(),
         info.payload().downcast_ref::<String>(),
--- a/compiler-core/src/erlang/tests/functions.rs
+++ b/compiler-core/src/erlang/tests/functions.rs
@@ -109,3 +109,31 @@ pub fn main() {
 "#
     );
 }
+
+#[test]
+fn labelled_argument_ordering() {
+    // https://github.com/gleam-lang/gleam/issues/3671
+    assert_erl!(
+        "
+type A { A }
+type B { B }
+type C { C }
+type D { D }
+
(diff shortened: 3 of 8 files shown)


lpil (commented):
Nice! Could you add an Erlang and JavaScript test too please? I think that'll make it easier to see what is happening when debugging if this breaks in future.

GearsDatapacks:
Tests have now been added.

lpil (approved):
Beautiful, thank you.

PR #3707, opened 2024-10-17
TypeScript type safety on records with no variants

richard-viney:
Currently the generated TypeScript declaration for

```gleam
pub type Dict(key, value)
```

as used by stdlib is

```ts
export type Dict$<A, B> = any;
```

Two issues arise from this:

1. The `any` loses type information and safety, e.g. in this case anything can be passed as a dict to stdlib's `dict.*` functions.
2. A TypeScript app consuming this type can't prohibit the use of `any`, which is a best-practice in strict TypeScript.

The proposed solution this PR implements is to change the generated code to be:

```ts
declare const __brand_Dict: unique symbol;
type Dict = { [__brand_Dict]: void };

export type Dict$<A, B> = Dict;
```

This code defines a type with one property named using a non-exported [unique symbol](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html#unique-symbol), meaning that it can't be looked up, isn't present at runtime, and doesn't appear in editor autocomplete. This effectively creates an empty nominal type, which addresses the two issues above.

This approach is a slight variant of the 'branded types' pattern in TypeScript that's well-described here:

- https://www.learningtypescript.com/articles/branded-types
- https://blog.logrocket.com/leveraging-typescript-branded-types-stronger-type-checks/

### DIFF
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
@@ -2,6 +2,6 @@
 source: compiler-core/src/javascript/tests/externals.rs
 expression: "pub type Queue(a)\n\n@external(javascript, \"queue\", \"new\")\npub fn new() -> Queue(a)\n"
 ---
-export type Queue$<I> = any;
+export type Queue$<I> = unknown;
 
 export function new$(): Queue$<any>;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -109,6 +109,10 @@
   the Erlang target if any of their arguments were discarded.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Records with no variants now use the TypeScript type `unknown` instead of
+  `any`.
+  ([Richard Viney](https://github.com/richard-viney))
+
 ### Formatter
 
 ### Language Server
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__tf_type_name_usage.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__tf_type_name_usage.snap
@@ -2,6 +2,6 @@
 source: compiler-core/src/javascript/tests/externals.rs
 expression: "\npub type TESTitem\n\n@external(javascript, \"it\", \"one\")\npub fn one(a: TESTitem) -> TESTitem\n"
 ---
-export type TESTitem$ = any;
+export type TESTitem$ = unknown;
 
 export function one(a: TESTitem$): TESTitem$;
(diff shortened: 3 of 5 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
```diff
@@ -2,6 +2,9 @@
 source: compiler-core/src/javascript/tests/externals.rs
 expression: "pub type Queue(a)\n\n@external(javascript, \"queue\", \"new\")\npub fn new() -> Queue(a)\n"
 ---
-export type Queue$<I> = any;
+declare const __brand_Queue: unique symbol;
+type Queue = { [__brand_Queue]: void };
```
lpil, line 6: This is incorrect, it is not an accurate description of the type.

We will not accept any inaccurate type definitions, even if the inaccuracy is a popular TS inference hack.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
```diff
@@ -2,6 +2,6 @@
 source: compiler-core/src/javascript/tests/externals.rs
 expression: "pub type Queue(a)\n\n@external(javascript, \"queue\", \"new\")\npub fn new() -> Queue(a)\n"
 ---
-export type Queue$<I> = any;
+export type Queue$<I> = unknown;
```
lpil, line 5: What are the implications of using `unknown` here?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
richard-viney: Hmm, if the requirement is to have a truly accurate TypeScript type then I think the author would need to provide it manually, perhaps via a type-based external like this:

```gleam
@external(typescript, "my_ffi.d.mts", "Dict<A, B>")
pub type Dict(A, B)
```

I'm not meaning to propose the above as a new language feature though, just thinking out loud.

In the meantime, changing the `any` to `unknown` helps somewhat. I've reduced this PR to only do that as detailed in the reply below.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__externals__external_type_typescript.snap
richard-viney: It's a simpler change that improves some of the type-safety issues related to the current `any` type.

It's an improvement because:

- It stops instances of these types, e.g. a `Dict`, from being passed freely anywhere.
- It fixes the problem of strict downstream linters emitting warnings about implicit use of `any`.

But it doesn't address the following:

- Anything can be passed where one of these types is expected. This is because all types are assignable/convertible to `unknown`.
- It's not unique, so e.g. `Dict`, `Regex`, `StringBuilder` are all `unknown` and TS can't tell them apart.

Neither of these are regressions from the current status quo of `any` though, they just aren't improved by it.

The only potential downside I can see with `unknown` is that there _could_ theoretically be TS code using the previous `any` in an incorrect way that will now report a TS error, but that's arguably a positive change to now be able to detect any such mistakes.

lpil: They're not unknown though, the programmer knows what the external type is in most instances.

lpil:
Thank you but I'm going to close this as we cannot accept this change. Please open a discussion if you have other proposals for language changes, thank you.

PR #3708, opened 2024-10-18
Add ability to specify erlang node names.

brandonpollack23:
This allows you to specify the erl cli parameters name or sname in the beam shell or run commands.

This makes it easier to attach to a running instance to launch observer, debug, inspect stuff, etc.

### DIFF
--- a/compiler-cli/src/main.rs
+++ b/compiler-cli/src/main.rs
@@ -182,7 +182,19 @@ enum Command {
     Fix,
 
     /// Start an Erlang shell
-    Shell,
+    Shell {
+        /// Node short name.
+        #[clap(long)]
+        sname: Option<String>,
+
+        /// Node long name.
+        #[clap(long)]
+        name: Option<String>,
+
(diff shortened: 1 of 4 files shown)


lpil (commented):
Hello! Thank you for this.

Gleam is a multi-runtime language, so any configuration for runtimes needs to be designed in such a fashion such that it is scoped to just that one runtime, rather than globally as the config is in this PR.

Due to the huge size of all the APIs across all the supported versions of all the supported runtimes it is preferred to use configuration methods that do not require special support within Gleam. For the BEAM specifically it is recommended to use environment variables, as it can be entirely configured that way. https://www.erlang.org/doc/apps/erts/erl_cmd.html#environment-variables

brandonpollack23:
Awesome didn't realize this. Thanks for the reply. Can you put this in an
FAQ?

brandonpollack23:
Well. Maybe not so frequent aq

lpil:
We would like to have documentation for the BEAM in future, but it's not something I have time for presently.

PR #3714, opened 2024-10-18
Fix compiler crash on duplicate fields

### LINKED ISSUE #3713: compiler crash when a variant has duplicate labels in specific circumstance
```gleam
pub type X {
  A
  B(e0: Int, e0: Int)
}

fn compiler_crash(x: X) {
  case x {
    A -> todo
    _ -> todo
  }
}
```

1. One variant has a duplicate field
2. The other variant is used in a pattern match

```rust
Panic: compiler-core/src/exhaustiveness.rs:780
	Custom type variants must exist: Type { name: "X", hint: AlternativeTypes(["UtfCodepoint", "Bool", "Node", "X", "BitArray", "Nil", "Int", "List", "String", "Result", "Float"]) }
Gleam version: 1.5.0
Operating system: linux
```

GearsDatapacks:
Fixes #3713 
This is similar to several other bugs that arose from fault tolerance, where a type variant wouldn't be defined if it didn't have duplicate fields because the whole function would short-circuit from the `?` operator. This PR simply reports the error in `Problems` instead of returning.

### DIFF
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_duplicate_record_fields.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_duplicate_record_fields.snap
@@ -1,0 +1,11 @@
+---
+source: compiler-core/src/type_/tests/errors.rs
+expression: "\npub type X {\n  A\n  B(e0: Int, e0: Int)\n}\n\nfn compiler_crash(x: X) {\n  case x {\n    A -> todo\n    _ -> todo\n  }\n}\n  "
+---
+error: Duplicate label
+  ┌─ /src/one/two.gleam:2:1
+  │
+2 │ pub type X {
+  │ ^^^^^^^^^^
+
+The label `e0` has already been defined. Rename this label.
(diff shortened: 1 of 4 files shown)


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_duplicate_record_fields.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: compiler-core/src/type_/tests/errors.rs
+expression: "\npub type X {\n  A\n  B(e0: Int, e0: Int)\n}\n\nfn compiler_crash(x: X) {\n  case x {\n    A -> todo\n    _ -> todo\n  }\n}\n  "
+---
+error: Duplicate label
+  ┌─ /src/one/two.gleam:2:1
+  │
+2 │ pub type X {
```
lpil, line 8: Can we make this point to the label rather than the type header?

lpil (commented):
Thank you!

GearsDatapacks:
The error now points to a better location.

lpil (approved):
Thank you!!

PR #3718, opened 2024-10-19
Don't remove the first argument in piped functions when it has a label

GearsDatapacks:
This was discussed on Discord, although it looks like no one made an issue of it.
This PR introduces a change to the formatter whereby it will no longer remove the first argument of a function in a pipeline if it is a function capture and has a label, as users may want to still annotate the label of the captured argument, even if it comes lexically first. This also removes the warning for an unnecessary capture in that case.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,34 @@
 
 ### Formatter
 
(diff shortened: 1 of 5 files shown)


lpil (approved):
Beautiful! Thank you

PR #3719, opened 2024-10-20
Consistent colouring of runtime errors on Erlang

### LINKED ISSUE #3638: Small inconsistency in error display
There's a small inconsistency between the look of compiler errors/warnings and the new pretty runtime errors:
Screenshot 2024-09-20 alle 22 51 08
Screenshot 2024-09-20 alle 22 51 54

In runtime errors `error:` is all red, while in the compiler error reports the `:` after `error`/`warning` isn't red/yellow like the text preceding it.

It's super small, but it would be nice if the display were consistent here

richard-viney:
Fixes #3638 with a one line change so the colon isn't coloured, which matches the compiler output:

Screenshot 2024-10-20 at 10 36 52 PM

### DIFF
--- a/compiler-core/templates/gleam@@main.erl
+++ b/compiler-core/templates/gleam@@main.erl
@@ -31,7 +31,7 @@ run_module(Module) ->
 
 print_error(Class, Error, Stacktrace) ->
     Printed = [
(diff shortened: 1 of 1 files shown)


lpil (approved):
Thank you!!

PR #3721, opened 2024-10-20
Fix formatting of lists of constant records

### LINKED ISSUE #3720: No newline inserted between records inside a const list if the first record spans multiple lines
```gleam
type Command {
  Command(name: String, description: String)
}

const commands = [
  Command(
    name: "dev",
    description: "Start a file watcher that automatically re-compiles your app on all file changes.",
  ), Command(name: "help", description: "Show this help text."),
]
```

when writing the same literal inside a function, the second command will be on its own line ~ :purple_heart:

giacomocavalieri:
This PR closes #3720

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -203,6 +203,7 @@
     Some(1)
   }
   ```
+
   ([Jiangda Wang](https://github.com/Frank-III))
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you

PR #3723, opened 2024-10-20
Return error from WASM compiler when unsupported features are used

### LINKED ISSUE #3310: wasm version of compiler not rejecting unsupported code for JS target
I have extracted this reproduction from my project but not yet managed to reduce the function much.
The following can be copied in the tour code playground an it breaks as follows.

Type checking passes, but the underline function is no in scope at runtime.
```gleam
import gleam/io
import gleam/bit_array
import gleam/list

fn underline(code, errors) {
  let code = bit_array.from_string(code)
  let #(_, _, acc) =
    list.fold(errors, #(code, 0, []), fn(state, error) {
      let #(remaining, offset, acc) = state
      let #(#(start, end), reason) = error
      let pre = start - offset
      let emp = end - start
      let offset = end
      case code {
        <<pre:bytes-size(pre), emp:bytes-size(emp), remaining:bytes>> -> {
          let assert Ok(pre) = bit_array.to_string(pre)
          let acc = case pre {
            "" -> acc
            content -> [content,..acc]
          }
          let assert Ok(emp) = bit_array.to_string(emp)
          let acc = case emp {
            "" -> acc
            content -> [content,..acc]
          }
          #(remaining, offset, acc)
        }
        _ -> panic
      }
    })
  list.reverse(acc)
}

pub fn main() {
    underline  
}
```

You can check that this is weird by replacing main with.
```gleam
pub fn main() {
  fn() { underline }  
}
```

richard-viney:
Fixes #3310.

When tested in `gleam-lang/playground`, the correct JS compilation error is now shown:

---

image

---

It might be nice for the playground to still be able to show the Erlang output even when the JS compilation fails. Is this worth having or too niche to bother about now the JS compilation error is surfaced?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,6 +243,10 @@
   which was defined with duplicate fields in one of its variants.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where the WASM compiler would return incomplete JavaScript when
+  unsupported features were used. It now returns a compilation error.
+  ([Richard Viney](https://github.com/richard-viney))
+
 ## v1.5.1 - 2024-09-26
 
 ### Bug Fixes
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!

PR #3725, opened 2024-10-21
Optimise bit array creation in the JavaScript prelude

richard-viney:
### Summary

This PR makes some optimisations to the `toBitArray()` function in the JavaScript prelude.

The optimised version is between 2x and 40x faster in my benchmarks, depending on the shape of the input.

### Details

The specific changes are:

- Special case when there is exactly zero or one segment(s).
- When there is one segment and it's a `Uint8Array`, pass it straight to the `BitArray` constructor without copying.
- When there are multiple segments, if none are `Uint8Array`s then pass `segments` straight to the Uint8Array constructor without copying.
- When there are multiple segments of different types, don't use a `DataView` and instead write directly into a `Uint8Array`.

> [!NOTE]
> I didn't include a changelog entry in because it didn't seem to fit exactly under any of the headings, but can certainly add one somewhere if desired.

### DIFF
--- a/compiler-core/templates/prelude.mjs
+++ b/compiler-core/templates/prelude.mjs
@@ -138,21 +138,53 @@ export class UtfCodepoint {
 
 // @internal
 export function toBitArray(segments) {
-  let size = (segment) =>
-    segment instanceof Uint8Array ? segment.byteLength : 1;
-  let bytes = segments.reduce((acc, segment) => acc + size(segment), 0);
-  let view = new DataView(new ArrayBuffer(bytes));
+  if (segments.length === 0) {
+    return new BitArray(new Uint8Array());
+  }
+
+  if (segments.length === 1) {
+    // When there is a single Uint8Array segment, pass it directly to the bit
+    // array constructor to avoid a copy
+    if (segments[0] instanceof Uint8Array) {
+      return new BitArray(segments[0]);
+    }
+
+    return new BitArray(new Uint8Array(segments));
+  }
+
+  // Count the total number of bytes, and check if there are any Uint8Array
+  // segments
+  let bytes = 0;
+  let hasUint8ArraySegment = false;
+  for (const segment of segments) {
+    if (segment instanceof Uint8Array) {
+      bytes += segment.byteLength;
+      hasUint8ArraySegment = true;
+    } else {
+      bytes++;
+    }
+  }
+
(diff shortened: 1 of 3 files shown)


compiler-core/templates/prelude.mjs
```diff
@@ -138,21 +138,53 @@ export class UtfCodepoint {
...
+  let u8Array = new Uint8Array(bytes);
   let cursor = 0;
   for (let segment of segments) {
     if (segment instanceof Uint8Array) {
-      new Uint8Array(view.buffer).set(segment, cursor);
+      u8Array.set(segment, cursor);
       cursor += segment.byteLength;
     } else {
-      view.setInt8(cursor, segment);
+      u8Array[cursor] = segment;
```
lpil, line 182: What is this doing?

lpil (commented):
Could you add tests in test/javascript_prelude for all the different types of arguments this should accept please 🙏

richard-viney:
I've added tests for all the different cases of input values and paths through the function. These were written against `main` then committed to this branch to check for regressions (none found). I've also double-checked that integer wrapping behaviour on JS matches Erlang.

giacomocavalieri:
> I didn't include a changelog entry in because it didn't seem to fit exactly under any of the headings, but can certainly add one somewhere if desired.

You could add this to the "compiler" section of the changelog! Something along the lines of "optimised creation of bit arrays on the JavaScript target"

compiler-core/templates/prelude.mjs
richard-viney: This is handling a segment that's a `number` when the input segments contain both numbers and `Uint8Array`s, e.g. when constructing something like `<<"A", 123>>`.

Previously `DataView.setInt8()` was being used, but now a numeric `segment` value is set directly on the final `Uint8Array`. If `segment` is outside [0 - 255] it will be wrapped into that range by the assignment, similar to how previously `segment` was being wrapped into the [-128 - 127] range by `DataView.setInt8()`.

The end result is the same, and also the updated tests now explicitly cover all of the integer wraparound points.

richard-viney:
> You could add this to the "compiler" section of the changelog! Something along the lines of "optimised creation of bit arrays on the JavaScript target"

Thanks, changelog updated!

lpil (approved):
Thank you!!

PR #3726, opened 2024-10-21
Improve "Add Type Annotations" Code Action

GearsDatapacks:
I realise that when I implemented #3682, I forgot to implement it for anonymous functions. I have now done that.
I also have limited the action to only between `let` and `=` for local variables, as it can lead to some confusing cases:
```gleam
let a = {
  let b = 10
  //  ^ Use this code action here
  b + 1
}
```
Would become:
```gleam
let a: Int = {
  let b: Int = 10
  b + 1
}
```
Which is often undesirable.

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -47,6 +47,9 @@ pub enum TypedExpr {
 
     Fn {
         location: SrcSpan,
+        /// The location of the function head, if this function was defined
+        /// using anonymous function syntax, as opposed to a capture.
+        head_location: Option<SrcSpan>,
         type_: Arc<Type>,
         is_capture: bool,
         args: Vec<TypedArg>,
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -47,11 +47,8 @@ pub enum TypedExpr {
 
     Fn {
         location: SrcSpan,
-        /// The location of the function head, if this function was defined
-        /// using anonymous function syntax, as opposed to a capture.
-        head_location: Option<SrcSpan>,
         type_: Arc<Type>,
-        is_capture: bool,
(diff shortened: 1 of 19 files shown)


compiler-core/src/ast/typed.rs
```diff
@@ -47,6 +47,9 @@ pub enum TypedExpr {
 
     Fn {
         location: SrcSpan,
+        /// The location of the function head, if this function was defined
+        /// using anonymous function syntax, as opposed to a capture.
+        head_location: Option<SrcSpan>,
```
lpil, line 52: This now duplicates the `is_capture` property. Let's replace both with a `kind: FunctionLiteralKind::{Capture, Anonymous}`, or something like that. The variant for not-a-capture would hold the location of the head.

compiler-core/src/ast/visit.rs
```diff
@@ -129,9 +129,11 @@ pub trait Visit<'ast> {
         visit_typed_expr_var(self, location, constructor, name);
     }
 
+    #[allow(clippy::too_many_arguments)]
```
lpil, line 132: This can be removed once we merge the two fields

lpil (commented):
Great! Thank you! I've added a note inline

compiler-core/src/ast/typed.rs
GearsDatapacks: Sounds good! I didn't really like this method anyway

GearsDatapacks:
I have now implemented your note, and the nice thing about this is: It gives us annotations on `use` for free! This now works:
```gleam
use some_value <- result.map(Ok(10))
//  ^ Use code action here
some_value + 1
```
Becomes:
```gleam
use some_value: Int <- result.map(Ok(10))
some_value + 1
```

lpil (approved):
Super nice work!!!

PR #3727, opened 2024-10-21
Fix Changelog

GearsDatapacks:
There were a few grammar mistakes, and a couple of bug fixes were in the `Compiler` section rather than `Bug Fixes`.


lpil (approved):
Thank you

PR #3732, opened 2024-10-21
Implement Maths operations in constant expressions

### LINKED ISSUE #1488: Allow basic math/... expressions in the const syntax
Context: [Discord discussion](https://discord.com/channels/768594524158427167/811254431051808778/939533231936659527)

Screenshot 2022-02-05 at 16 00 06

GearsDatapacks:
Closes #1488 
This PR implements support for binary operations in constants.

A few notes:
- It's still missing tests. This probably needs a lot of tests.

- This is implemented differently on Erlang and Javascript, due to the different languages being somewhat different.
  For example, this code:
  ```gleam
  const minutes_per_hour = 60
  const hours_per_day = 24
  const minutes_per_day = minutes_per_hour * hours_per_day
  
  pub fn main() {
    minutes_per_day
  }
  ```
  Compiles to this Javascript:
  ```js
  const minutes_per_hour = 60
  const hours_per_day = 24
  const minutes_per_day = minutes_per_hour * hours_per_day
  
  export function main() {
    return minutes_per_day;
  }
  ```
  And this erlang:
  ```erlang
  main() ->
      1440.
  ```
  Although this constant folding process could be done in the analysis stage instead and be applied to javascript as well.

- I've been having some trouble trying to get the formatter  to play nicely with this; I still don't really understand how it works. If someone who is more familiar with it could point me in the right direction, that would be amazing.
- I'm not super happy with the parser code for constant binary operations: I had to copy a lot of the expression-parsing code. It might be good to merge this code somehow.


giacomocavalieri:
One more thing that's missing is the version tracking. We want to track that if you're using this your code is only compatible with `1.6` or later (given this is merged before the 1.6 release), you can have take inspiration from the way it works with arithmetic in guards <https://github.com/gleam-lang/gleam/blob/9ee71a11393de9ebb1eb4a43dbebdea2def182f0/compiler-core/src/type_/expression.rs#L1881>

lpil:
I wonder if there's a way to make sure we always remember to add that. I forgot!

PR #3733, opened 2024-10-21
LSP Code Action: Convert to Qualified import 

Frank-III:
close: #3603

also, I realize there could be two constructors with the same name, and applying the `Unqualify constructor` action could cause a conflict. Should we check that before offering the code action  or let the user check?


lpil:
>  Should we check that before offering the code action or let the user check?

Oh dear! Yes please

lpil (approved):
Wow! Super impressive work! Thank you so much!

I'm really going to enjoy this one, really useful

PR #3743, opened 2024-10-23
fix: check if constructor exist when offer `Unqualify` action

Frank-III:
as discussed in #3733

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1112,15 +1112,15 @@ pub struct QualifiedConstructor<'a> {
     module_aliased: bool,
     used_name: EcoString,
     constructor: EcoString,
-    is_type: bool,
+    layer: ast::Layer,
 }
 
 impl<'a> QualifiedConstructor<'a> {
     fn constructor_import(&self) -> String {
-        if self.is_type {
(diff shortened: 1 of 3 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -1141,15 +1141,35 @@ impl<'a> QualifiedToUnqualifiedImportFirstPass<'a> {
...
-            .iter()
-            .find_map(|def| match def {
-                ast::Definition::Import(import) if import.module == *module_name => Some(import),
-                _ => None,
-            })
+    fn get_module_import(
+        &self,
+        module_name: &EcoString,
+        constructor: &EcoString,
+        is_type: bool,
```
lpil, line 1148: Tiny thing, rather than using a bool can we the the `Layer` type here. It has `Type` and  `Value` variants

lpil (commented):
Lovely! One small note

lpil:

compiler-core/src/language_server/code_action.rs
lpil: If you're up for making the larger change that would be very cool!

lpil (approved):
Thank you!

PR #3744, opened 2024-10-23
Fix PHONY in test Makefiles

PgBiel:
`.phony` was written in lowercase, which appears to not have any effect.


PgBiel:
The new CI error seems to be because the test was never run before.

lpil:
Oh dear!! Thank you very much.

PR #3738, opened 2024-10-22
Improve snapshot format for error snapshots

GearsDatapacks:
Part of #3736 
This PR adds the source code causing the error to the snapshots for error tests.

PR #3745, opened 2024-10-23
Display source code for other modules in snapshots

GearsDatapacks:
As discussed in #3738 
This PR prints the source code for the other modules in the `assert_with_module_error` snapshots, and other similar ones.

### DIFF
--- a/compiler-core/src/javascript/tests.rs
+++ b/compiler-core/src/javascript/tests.rs
@@ -40,7 +40,11 @@ macro_rules! assert_js_with_multiple_imports {
     ($(($name:literal, $module_src:literal)),+; $src:literal) => {
         let compiled =
             $crate::javascript::tests::compile_js($src, vec![$((CURRENT_PACKAGE, $name, $module_src)),*]).expect("compilation failed");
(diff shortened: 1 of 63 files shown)


compiler-core/src/javascript/tests.rs
```diff
@@ -40,7 +40,11 @@ macro_rules! assert_js_with_multiple_imports {
     ($(($name:literal, $module_src:literal)),+; $src:literal) => {
         let compiled =
             $crate::javascript::tests::compile_js($src, vec![$((CURRENT_PACKAGE, $name, $module_src)),*]).expect("compilation failed");
-        let output = format!("----- SOURCE CODE\n{}\n\n----- COMPILED JAVASCRIPT\n{}", $src, compiled);
+            let mut output = String::from("----- SOURCE CODE\n");
+            for (name, src) in [$(($name, $module_src)),*] {
```
GearsDatapacks, line 44: I don't know Rust macros very well, so I've just made this for loop. Is there a way to do this loop statically?

lpil: Probably but I don't care

Issue #3737, opened 2024-10-22
Remove references to "spread" in the codebase


lpil:
Internally we call record updates "spread" in some places. Rename this to use the record update, as that's the official name.

PR #3748, opened 2024-10-23
refactor: replace "spread" with "record update" naming (gh-3737)

george-grec:
closes https://github.com/gleam-lang/gleam/issues/3737

I searched for `spread` in the project and replaced what seemed to me was related to record updates. It was a bit hard to distinguish from pattern spreads.


compiler-core/src/ast.rs
```diff
@@ -1286,7 +1286,7 @@ impl<T> HasLocation for CallArg<T> {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct RecordUpdateSpread {
+pub struct RecordUpdateRecord {
```
lpil, line 1289: search-replace accident? 😁

george-grec: No, no, I went through all occurrences manually and used the LSP for renaming. I thought this represents the _record_ part of the record update, e.g. in `let new_wibble = Wibble(..old_wibble, wobble_factor: 10)` this would be `..old_wibble`. Is that assumption correct? Do you have a better name?

george-grec: Sounds good, I'll fix that after work. Let me know if I missed anything else until then :)

lpil: Thank you!

lpil (approved):
Thank you!

PR #3757, opened 2024-10-29
Fix bug with type alias printing

### LINKED ISSUE #3749: Language server reports wrong type for variables
I ran into a strange behaviour on main. If I write the following code:
```gleam
pub type Wibble =
  List(Int)

pub fn main() {
  []
}
```
The inferred type for `main` (and what I get if I let the LS add the type annotation) is `fn() -> Wibble(a)` which is wrong since `Wibble` is not generic and I wouldn't really want it to use the alias even if it were.

It also happens with aliases imported from other modules:
```gleam
// type DecodeErrors = List(DecodeError)
import gleam/dynamic.{type DecodeErrors}

// Inferred as `fn() -> DecodeErrors(a)`
pub fn main() {
  []
}
```

---

edit: another really strange thing is if I add the type explicitly it still adds the type variable. For example:

```gleam
pub type Wibble =
  List(Int)

// Inferred as `fn(Wibble(Int)) -> Nil`
fn wibble(a: Wibble) {
  Nil
}
```

GearsDatapacks:
Fixes #3749 
This PR changes the `Names` class to only register aliases when the type parameters map exactly to those of the type they alias. For example:
```gleam
type MyResult(a, b) = Result(a, b) // Result will now be printed as MyResult
type OptionalInt = option.Option(Int) // This is ignored as it doesn't map nicely
```

In future we could implement a more complex system, which could map the arguments of the type to the parameters of the alias, but I started implementing that and it got pretty complex pretty fast, so I just went with this minimal approach which fixes the bug.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1182,7 +1182,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
 
             environment
                 .names
-                .type_in_scope(name.clone(), type_.as_ref());
+                .type_in_scope(name.clone(), type_.as_ref(), &parameters);
 
             // Insert the alias so that it can be used by other code.
             environment.insert_type_constructor(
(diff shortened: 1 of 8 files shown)


lpil (approved):
I like this solution a lot! Thank you!

PR #3764, opened 2024-10-31
Warn about unsafe integer values on JavaScript

### LINKED ISSUE #3746: Emit a warning if an int literal is too large to be represented on JavaScript
To avoid confusing runtime behaviour.

richard-viney:
This PR adds a warning when targeting JavaScript and Int literals or constants outside JavaScript's safe range are encountered.

Fixes #3746.

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -17,6 +17,7 @@ use hexpm::version::Version;
 use id_arena::Arena;
 use im::hashmap;
 use itertools::Itertools;
+use num_bigint::BigInt;
 use vec1::Vec1;
 
 #[derive(Clone, Copy, Debug, Eq, PartialOrd, Ord, PartialEq, Serialize)]
@@ -309,8 +310,14 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             UntypedExpr::Var { location, name, .. } => self.infer_var(name, location),
 
             UntypedExpr::Int {
-                location, value, ..
-            } => Ok(self.infer_int(value, location)),
+                location,
+                value,
+                int_value,
+                ..
+            } => {
+                self.check_javascript_int_safety(&int_value, location);
+                Ok(self.infer_int(value, int_value, location))
+            }
 
             UntypedExpr::Block {
                 statements,
@@ -490,10 +497,11 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         }
     }
 
-    fn infer_int(&mut self, value: EcoString, location: SrcSpan) -> TypedExpr {
+    fn infer_int(&mut self, value: EcoString, int_value: BigInt, location: SrcSpan) -> TypedExpr {
         TypedExpr::Int {
             location,
             value,
+            int_value,
             type_: int(),
         }
     }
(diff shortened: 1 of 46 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -3507,6 +3513,37 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+        if let Some(i) = num_bigint::BigInt::parse_bytes(bytes, radix) {
+            let js_min_safe_integer = -9007199254740991i64;
+            let js_max_safe_integer = 9007199254740991i64;
+
+            if i < js_min_safe_integer.into() || i > js_max_safe_integer.into() {
+                self.problems
+                    .warning(Warning::JavaScriptIntUnsafe { location });
+            }
+        }
+    }
```
lpil, line 3546: Seeing as we want to use the int values in a few places should we do this string-to-int during parsing? To avoid each step of the compiler parsing it over and over

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -2419,3 +2419,83 @@ pub fn main(a) {
 ",
     );
 }
+
+#[test]
+fn javascript_unsafe_int() {
+    assert_js_warning!(
+        r#"
+pub fn go() {
```
lpil, line 2427: Split these into multiple tests please, each test should only test one case.

compiler-core/src/warning.rs
```diff
@@ -1046,6 +1046,22 @@ See: https://tour.gleam.run/functions/pipelines/",
                         }),
                     }
                 }
+
+                type_::Warning::JavaScriptIntUnsafe { location } => Diagnostic {
+                    title: "Int is outside JavaScript's safe integer range".into(),
```
lpil, line 1051: It's not clear what this means, could you add more information to the test please 🙏

lpil (commented):
Very nice! Thank you

compiler-core/src/type_/tests/warnings.rs
richard-viney: Have made the tests more granular

compiler-core/src/warning.rs
richard-viney: Have filled in the `text` member, the output is now:

```
warning: Int is outside the safe range on JavaScript
  ┌─ /Users/richard/Desktop/int_test/src/int_test.gleam:1:15
  │
1 │ pub const i = 9_007_199_254_740_992
  │               ^^^^^^^^^^^^^^^^^^^^^ This is not a safe integer on JavaScript

This integer value is too large to be represented accurately by
JavaScript's number type. To avoid this warning integer values must be in
the range -(2^53 - 1) - (2^53 - 1).

See JavaScript's Number.MAX_SAFE_INTEGER and Number.MIN_SAFE_INTEGER
properties for more information.
```

compiler-core/src/type_/expression.rs
richard-viney: The most recent commit to this branch attempts this. I'm not 100% sure it's what you were meaning, but it might be!

- `Token::Int` gets an `int_value: Option<BigInt>` field, which flows through to `UntypedExpr::Int`, `TypedExpr::Int`, `Pattern::Int`, and `Constant::Int`.
- This field is then used for the new warning.

compiler-core/src/warning.rs
lpil: That's really nice!

compiler-core/src/type_/expression.rs
lpil: Very nice! Why is it optional? Shouldn't it always be present?

richard-viney: Yes you're right, I've made it always present and fixed handling of leading underscores on parsing as bigint.

There are now three `.expect("int value to parse as bigint")` calls, but I believe these to be safe due to previous validation done during parsing.

richard-viney:
Rebased to fix merge conflicts

lpil (approved):
Fantastic!! Thank you!

PR #3780, opened 2024-11-02
Reduces compiler-wasm binary file size

malbarbo:
Add the `compiler-wasm` feature `reduce-bin-size` which, when enabled, reduces the generated wasm file from 3.7 MB to 2.2 MB.

We achieved this by:
- Adding the `compiler-core` feature `disable-erlang`, which disables erlang code generation
- Replacing the use of regular expressions with iterator code
- Adding the `compiler-core` feature `disable-hide-internal`, which disables hiding of internal methods, which requires the use of the `globset` crate, which in turn requires the use of the `regex` crate

Nothing changes by default, but when the `compiler-wasm` crate is built with the `reduce-bin-size` feature enabled, the compiler do it's magic and no code from the erlang backend and the `globset` and `regex` crates appears in the wasm file.

### DIFF
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -7,6 +7,7 @@ license-file = "LICENCE"
 
 [features]
 disable-erlang = []
+disable-hide-internal = []
 
 [dependencies]
 # Error message and warning formatting
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -5,6 +5,9 @@ authors = ["Louis Pilfold <louis@lpil.uk>"]
 edition = "2021"
 license-file = "LICENCE"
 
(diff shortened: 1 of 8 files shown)


malbarbo:
I'm incorporating the gleam compiler into a project, and the binary size is import.

I made this PR in case it's an interesting aspect of the gleam project. But if it's not desirable, no problem.

lpil (commented):
Hello! Thank you for this, but this introduces a substantial maintenance overhead which I'm not in a position to support. Gleam is a community project and we about about 0.8 developers worth of money, so we have to be as efficient as possible with our time.

malbarbo:
Thank you for your feedback. I understand the concerns and appreciate the work you do on Gleam. Thanks.

PR #3787, opened 2024-11-04
Skip existing files when skip git option is enable

spk:
even with the --skip-git option the CLI ask to remove the files

tests ensure we can skip files that are none or without content:

```
---- new::tests::skip_existing_git_files_when_skip_git_is_true stdout ----
thread 'new::tests::skip_existing_git_files_when_skip_git_is_true' panicked at compiler-cli/src/new/tests.rs:321:6:
called `Result::unwrap()` on an `Err` value: OutputFilesAlreadyExist { file_names: ["/tmp/.tmpRvY1Nb/my_project/.gitignore"] }
```

### DIFF
--- a/compiler-cli/src/new/tests.rs
+++ b/compiler-cli/src/new/tests.rs
@@ -297,3 +297,31 @@ fn conflict_with_existing_files() {
         })
     );
 }
+
+#[test]
+fn skip_existing_git_files_when_skip_git_is_true() {
+    let tmp = tempfile::tempdir().unwrap();
+    let path = Utf8PathBuf::from_path_buf(tmp.path().join("my_project")).expect("Non Utf8 Path");
+
+    crate::fs::mkdir(&path).unwrap();
+    let file_path = PathBuf::from(&path).join(".gitignore");
+
+    let _ = std::fs::File::create(file_path).unwrap();
+
+    let creator = super::Creator::new(
+        super::NewOptions {
+            project_root: path.to_string(),
+            template: super::Template::Erlang,
+            name: None,
+            skip_git: true,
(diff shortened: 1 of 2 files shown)


compiler-cli/src/new/tests.rs
```diff
@@ -297,3 +297,31 @@ fn conflict_with_existing_files() {
...
+
+#[test]
+fn skip_existing_git_files_when_skip_git_is_true() {
+    let tmp = tempfile::tempdir().unwrap();
+    let path = Utf8PathBuf::from_path_buf(tmp.path().join("my_project")).expect("Non Utf8 Path");
+
+    crate::fs::mkdir(&path).unwrap();
+    let file_path = PathBuf::from(&path).join(".gitignore");
+
+    let _ = std::fs::File::create(file_path);
```
lpil, line 309: Use `.unwrap();` rather than discarding the result as otherwise this operation could fail and we wouldn't know. 🙏

lpil (commented):
Thank you! Left a small note inline

compiler-cli/src/new/tests.rs
spk: Updated checking that the result is ok

lpil (commented):
Use `.unwrap()` please. Using `assert!` in the error not being shown to the user in the event of a failure.

spk:
> Use `.unwrap()` please. Using `assert!` in the error not being shown to the user in the event of a failure, so they would need to modify the code to use `.unwrap()` and run it again to know what the problem is.

Ok seems fair updated.

lpil (approved):
Champion, thank you!

PR #3792, opened 2024-11-05
Warn about unsafe integers in bit array patterns on JavaScript

richard-viney:
The case of unsafe integer literals in bit array patterns was overlooked by https://github.com/gleam-lang/gleam/pull/3764.

This PR makes them emit the new warning as well.


lpil (approved):
Thank you

PR #3793, opened 2024-11-05
Fix constructor functions in JavaScript guards

PgBiel:
The fix to https://github.com/lpil/decode/pull/6 was only applied to constructor functions in constants, but not in guards. This PR fixes that.


lpil (approved):
Thank you!

PR #3794, opened 2024-11-05
chore: improve file create in case of error

spk:
follow up on https://github.com/gleam-lang/gleam/pull/3787#pullrequestreview-2415286300


lpil (approved):
oh yes! Thank you very much

PR #3795, opened 2024-11-05
Report duplicate label error in record updates

### LINKED ISSUE #3783: No error generated for duplicate labels in record updates
When supplying a labeled argument to a function or a record constructor twice, you get a duplicate label error:

```gleam
type Wibble { Wibble(thing: Int, other: Int) }
```

```
error: Duplicate argument
   ┌─ /home/arkan/Projects/oss/gleam/probe/src/probe.gleam:19:29
   │
19 │   let b = Wibble(thing: 10, thing: 20)
   │                             ^^^^^^^^^

The labelled argument `thing` has already been supplied.
```

when using the record update syntax, no such error is produced. Instead, the compiler will emit a "reduntant record update" warning if you repeat a labelled argument often enough:

```
warning: Redundant record update
   ┌─ /home/arkan/Projects/oss/gleam/probe/src/probe.gleam:19:11
   │
19 │   let b = Wibble(..a, thing: 10, thing: 20)
   │           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This record update specifies all fields
```

It just so happens that because of the way codegen works for both targets, the last specified value will be kept.

I'd fix that as part of implementing #745 in #3773 since I need to keep track of all the labels I've seen already anyways, but I wanted to make this issue to document and to ask if emitting an error is actually the desired behaviour? Making this an error could also be seen as a breaking change.

What should I do here?

GearsDatapacks:
Fixes #3783
It sounds like this will be fixed when record updates are reworked anyway, but I wanted to fix it as soon as possible in case we release before that change occurs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -394,6 +394,10 @@
   wouldn't emit a warning when the minimum specified Gleam version was < 1.4.0.
   ([yoshi](https://github.com/joshi-monster))
 
+- Fixed a bug where no error would be reported when duplicate labelled
+  arguments were supplied in a record update.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.5.1 - 2024-09-26
 
 ### Bug Fixes
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -2470,6 +2470,8 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             }
         }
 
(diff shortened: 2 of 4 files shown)


joshi-monster:
So this should be an error, ok! Then I'll steal your test and finish my PR :purple_heart:

GearsDatapacks:
That's what I inferred based on Louis' comment on your issue

lpil (approved):
Thank you

PR #3799, opened 2024-11-06
Fix 3796

### LINKED ISSUE #3796: () is accept after a type name that is not generic
The following code

```gleam
pub fn main() {
  let a: Int() = 0
}
```

compiles fine.

I'm using gleam 1.5.1

giacomocavalieri:
This PR fixes #3796 and a similar bug where `pub type Wibble() { ... }` would be accepted (but rewritten as `pub type Wibble { ... }` by the formatter)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -15,7 +15,7 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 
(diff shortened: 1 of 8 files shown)


lpil (approved):
Thank you!

PR #3800, opened 2024-11-06
Fix bug with variant inference in tuples

### LINKED ISSUE #3797: Type variables preserve record variant inference when inside tuples
(Gleam nightly)

```gleam
type Either(a, b) {
  Left(value: a)
  Right(value: b)
}

fn always_second(_first: value, second: value) -> value {
  second
}

pub fn main() {
  let #(right) = always_second(#(Left(5)), #(Right("hello")))
  Left(..right, inner: 10)
}
```

compiles, except for a "redundant record update" warning. It is important that the values here are wrapped in tuples - the size of the tuple doesn't matter, and a `Box` record works correctly.

There might be a second bug here about type unification not dropping the inferred variant, but it also works without having 2 different values for the compiler to infer:

```gleam
fn twice(state: state, fun: fn(state) -> state) -> state {
  fun(fun(state))
}

fn main() {
  twice(#(Left(5)), fn(value) {
    case value.0 {
      Left(_) -> #(Right("hello"))
      Right(_) -> value
    }
  })
}
``` 

produces

```txt
warning: Unreachable case clause
   ┌─ /home/arkan/Projects/oss/gleam/probe/src/probe.gleam:25:7
   │
25 │       Right(_) -> value
   │       ^^^^^^^^^^^^^^^^^

This case clause cannot be reached as it matches on a variant of a type
which is never present.
```

I'm sorry Gears ~ :purple_heart:

GearsDatapacks:
Fixes #3797

### DIFF
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -285,7 +285,12 @@ impl Type {
                     Arc::make_mut(element).generalise_custom_type_variant();
                 }
             }
-            Type::Fn { .. } => {}
+            Type::Fn { args, retrn } => {
+                for argument in args {
+                    Arc::make_mut(argument).generalise_custom_type_variant();
+                }
+                Arc::make_mut(retrn).generalise_custom_type_variant();
+            }
         }
     }
 
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -280,7 +280,12 @@ impl Type {
                 inferred_variant, ..
             } => *inferred_variant = None,
             Type::Var { type_ } => type_.borrow_mut().generalise_custom_type_variant(),
-            Type::Fn { .. } | Type::Tuple { .. } => {}
+            Type::Tuple { elems } => {
+                for element in elems {
+                    Arc::make_mut(element).generalise_custom_type_variant();
+                }
+            }
+            Type::Fn { .. } => {}
         }
     }
 
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -2147,7 +2147,7 @@ pub fn fun(x) {
 fn type_unification_removes_inferred_variant_in_tuples() {
     assert_module_error!(
         r#"
-type Either(a, b) {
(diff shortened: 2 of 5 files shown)


compiler-core/src/type_.rs
```diff
@@ -280,7 +280,12 @@ impl Type {
                 inferred_variant, ..
             } => *inferred_variant = None,
             Type::Var { type_ } => type_.borrow_mut().generalise_custom_type_variant(),
-            Type::Fn { .. } | Type::Tuple { .. } => {}
+            Type::Tuple { elems } => {
+                for element in elems {
+                    Arc::make_mut(element).generalise_custom_type_variant();
```
GearsDatapacks, line 285: I don't really like how this has to clone every tuple element, even if they don't need to be modified at all. I can't really find a better way to do it though

compiler-core/src/type_/tests.rs
```diff
@@ -2142,6 +2142,28 @@ pub fn fun(x) {
...
+  second
+}
+
+pub fn main() {
+  let #(right) = a_or_b(#(Left(5)), #(Right("hello")))
+  Left(..right, inner: 10)
+}
+"#
+    );
+}
```
lpil, line 2165: Can we have a test for in a function too? Just to ease my mind!

GearsDatapacks: I'm not exactly sure what you mean by that

lpil: You written a test to ensure this problem doesn't happen for tuples, but I can't see a test that shows the same wouldn't happen with anonymous functions.

GearsDatapacks: Ah I see thanks, I'll add that shortly

GearsDatapacks:
There did seem to be a bug with anonymous functions as well. I guess I thought there wouldn't be because it's not a problem with regular functions, but I guess it's because anonymous functions are not generalised.

lpil (approved):
Thank you!!!

PR #3807, opened 2024-11-08
Implement `let assert ... as` syntax

### LINKED ISSUE #3216: Add `let assert ... as` syntax

GearsDatapacks:
Closes #3216 
This PR adds the option to add a custom panic message when using `let assert`.


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__convert_let_assert_with_message_to_case.snap
```diff
@@ -0,0 +1,22 @@
...
+  inner
+}
+
+
+----- AFTER ACTION
+
+pub fn expect(value, message) {
+  let inner = case value {
+    Ok(inner) -> inner
+    _ -> panic as message
```
lpil, line 19: Oh that is super nice

lpil (approved):
Thank you!!

PR #3814, opened 2024-11-10
Escape property names with special meaning on JavaScript

### LINKED ISSUE #3813: Invalid JS generated when compiling a record that has a field named `constructor`, crashes at runtime
The crash seems to be due to the call to `withFields` during a record update, as that function calls the constructor of the type

Reproduction: 
```gleam
pub type Thing {
  Thing(constructor: Nil)
}

pub fn main() {
  let a = Thing(constructor: Nil)
  let b = Thing(..a)
}
```

Resulting JS for type:
```js
export class Thing extends $CustomType {
  constructor(constructor) {
    super();
    this.constructor = constructor;
  }
}
```

Gleam version: 1.5.1

Crash:
```text
TypeError: this.constructor is not a constructor
    at Thing.withFields (file:///Users/rattard/code/gleam/constructor/build/dev/javascript/prelude.mjs:9:12)
    at main (file:///Users/rattard/code/gleam/constructor/build/dev/javascript/constructor/constructor.mjs:12:13)
    at file:///Users/rattard/code/gleam/constructor/build/dev/javascript/constructor/gleam.main.mjs:2:1
    at ModuleJob.run (node:internal/modules/esm/module_job:268:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:543:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:116:5)

Node.js v23.1.0
```

joshi-monster:
Escape `constructor`, `then`, `prototype` and `__proto__`, since they all have special meaning.

fix #3813

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Fixed a bug where some reserved field names were not properly escaped in
+  custom types on the JavaScript target.
+  ([yoshi](https://github.com/joshi-monster))
+
 ## v1.6.0-rc1 - 2024-11-10
 
 ### Build tool
--- a/compiler-core/src/javascript.rs
+++ b/compiler-core/src/javascript.rs
@@ -293,7 +293,9 @@ impl<'a> Generator<'a> {
                 let var = parameter((i, arg));
                 match &arg.label {
                     None => docvec!["this[", i, "] = ", var, ";"],
-                    Some((_, name)) => docvec!["this.", name, " = ", var, ";"],
+                    Some((_, name)) => {
+                        docvec!["this.", maybe_escape_property_doc(name), " = ", var, ";"]
+                    }
                 }
             }),
             line(),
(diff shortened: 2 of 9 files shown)


giacomocavalieri:
There's a few other places that need to be updated:
- field access in guards
- field access when pattern matching

And I think consts should be updated too, iirc they're using a different function so that would not escape fields in constant records

joshi-monster:
I know, I remembered as soon as I pressed open PR :D

joshi-monster:
Thanks for noticing so quickly, guards where indeed broken!! :purple_heart: 

As far as I can tell, constants just call the constructor, which does not use the labels, and record updates are not yet supported in constants.

lpil (commented):
Can we remove the special case for "then" in is_usable_js_identifier with this?

joshi-monster:
No, the special case for `then` in is_usable_js_identifier` never affected records and is there to make sure that constants or functions called `then` would not automatically be called when awaiting dynamic imports. (so using `await import(...)` syntax on a Gleam module)

lpil:
Oh yes I forgot about imports. Thank you!

lpil (approved):
Thank you!!

PR #3817, opened 2024-11-10
Fix incorrect warning about unsafe ints on JS when there's an @external

richard-viney:
In v1.6.0-rc1 the following code was warning about an unsafe int on JS, but shouldn't have been because of the `@external` implementation.

```gleam
@external(javascript, "./test.mjs", "go")
pub fn go() -> Int {
  9_007_199_254_740_992
}
```

### DIFF
--- a/compiler-core/src/type_/tests/warnings.rs
+++ b/compiler-core/src/type_/tests/warnings.rs
@@ -1,7 +1,7 @@
 use super::*;
 use crate::{
-    assert_js_warning, assert_no_warnings, assert_warning, assert_warnings_with_gleam_version,
-    assert_warnings_with_imports,
+    assert_js_no_warnings, assert_js_warning, assert_no_warnings, assert_warning,
+    assert_warnings_with_gleam_version, assert_warnings_with_imports,
 };
 
 #[test]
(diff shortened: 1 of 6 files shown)


compiler-core/src/type_/tests/warnings.rs
```diff
@@ -2671,3 +2671,27 @@ pub fn go() {
...
+fn javascript_unsafe_int_with_external_implementation() {
+    assert_js_no_warnings!(
+        r#"
+@external(javascript, "./test.mjs", "go")
+pub fn go() -> Int {
+  9_007_199_254_740_992
+}
+"#
+    );
+}
```
lpil, line 2685: Could we have a test that this code emits a warning please

```gleam
pub fn main() {
  9_007_199_254_740_992 + helper()
}

@external(javascript, "a", "b")
fn helper() -> Int
```

lpil (commented):
Good catch!

compiler-core/src/type_/tests/warnings.rs
richard-viney: Added extra test for this

lpil (approved):
Thank you!!

PR #3820, opened 2024-11-12
Fix invalid TOML example in changelog

richard-viney:
The example of how to use the new `repository.path` option was altered in 9e97867, but the updated version isn't actually valid TOML as inline tables have to be on one line: https://toml.io/en/v1.0.0#inline-table.

This PR switches to normal TOML table syntax, which is probably what was intended.


lpil (approved):
Thank you

PR #3825, opened 2024-11-13
Specify Erlang v27 in the CI workflow for new projects

richard-viney:
This increases the Erlang version used in the GitHub Actions workflow created by `gleam new` from v26.0.1 to v27.1.2.

Motivation is that lots of folks use `gleam_json`, the latest version of which needs Erlang 27, so this change will avoid them running into a CI failure.

If this is wanted, but isn't eligible for v1.6 RC 2, then just lmk and I'll rebase once v1.6 is released.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+### Build tool
+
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you

PR #3831, opened 2024-11-14
gleam dev idea

lpil:
What do you think?

[Rendered](https://github.com/gleam-lang/gleam/blob/1c31b4e419525a4c9288a87693e35b382333f4db/docs/ideas/gleam-dev.md)

### DIFF
--- a/docs/ideas/gleam-dev.md
+++ b/docs/ideas/gleam-dev.md
@@ -1,0 +1,33 @@
+# Gleam `dev`
+
+Currently Gleam code has to reside in one of two directories.
+
+- `src`: The directory for production code that is part of the application or
+  library itself.
+- `test`: The directory for non-production code. This is likely mostly test
+  code, but it could also include development helpers, code generation, etc.
+
+The non-production directory being called `test` has been a repeat source of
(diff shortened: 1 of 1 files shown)


schurhammer:
> gleam fix could rename test to dev if it contains Gleam code.

I think we better not do this.

1. It can break the project
2. Its very easy to do manually

Other than that, sounds good to me.

strawmelonjuice:
Oh yes this makes that a lot clearer --must say the proposal cleared it up for me...--, if this is doable without considerable load I'd say this is a great idea?

lpil:
That seems reasonable @schurhammer. It is easy for them to do, as you say.

Guria:
Is there a reason not to support both test and dev at the same time?

massivefermion:
Why get rid of `test`? I think test code is the most probable thing this directory is going to be used for, so it deserves a directory of its own. You can just make the compiler recognize the `dev` directory if it's there.

lpil:
It's not a problem in other mainstream ecosystems such as Java and JavaScript, so I don't think it would be a problem for Gleam.

PR #3838, opened 2024-11-16
feat: check for exposing multiple modules on publish

### LINKED ISSUE #3834: validate package with multiple top level module names on publish
# Problem
Since package names don't necessarily match module names, library authors can inadvertently pollute global module names. This is especially problematic as it's easy to introduce multiple top-level modules in a single library.

Given that we have a single namespace for all dependencies, unresolvable transitive conflicts can arise for package consumers.

# Suggested Solution
Restrict packages from publishing more than one top-level module name unless explicitly defined in `gleam.toml` to indicate the library author's intent for potential valid use cases.

The CLI can optionally offer automatic `gleam.toml` updates. For stricter enforcement, we could require manual updates.

Related discussion:
https://discord.com/channels/768594524158427167/1281967484731265044/1281967484731265044

Guria:
fixes #3834

reusing existing `--yes` flag to bypass new warning.


Guria:
rebased to solve conflict in ci workflow file

test/multi_namespace/src/foo.gleam
```diff
@@ -0,0 +1,3 @@
+pub fn main() {
```
lpil, line 1: Rename this file to something else please, "foo" is a morbid world-war-2 reference.

compiler-cli/src/publish.rs
```diff
@@ -123,6 +124,42 @@ fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
...
+            top_level_module_names.join(", ")
+        );
+        println!(
+            "Defining multiple top-level modules can lead to namespace pollution \
+and potential conflicts for consumers."
+        );
+
+        let should_publish =
+            i_am_sure || cli::confirm("\nDo you wish to continue publishing this package?")?;
+        println!();
```
lpil, line 152: Can we add that they should all go in $PACKAGE_NAME to the message please 🙏

lpil (commented):
Very nice! Thank you! Couple small things inline

compiler-cli/src/publish.rs
lpil: The error message needs to show the user how to correct the mistake as the user doesn't know how to do it, otherwise they'll be stuck.

I think telling them to put them all in their package name is good enough, but you could come up with some heuristic for determining what namespace to suggest that would be good also.

Guria:
@lpil updated

Do you expect me to update `changelog` already?

Guria:
I also raised an [related question](https://github.com/gleam-lang/gleam/pull/3893#issuecomment-2504526202) for `confirm_with_text` implementation.

compiler-cli/src/publish.rs
```diff
@@ -123,6 +124,53 @@ fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
...
+        println!(
+            "To fix this, move all your modules under a single top-level module of your choice."
+        );
+        println!();
+        println!("For example:");
+        println!(
+            "  src/{}.gleam\n  src/{}/module1.gleam\n  src/{}/module2.gleam",
+            &package.config.name, &package.config.name, &package.config.name
+        );
+        println!();
```
lpil, line 159: Could you use a multi-line string rather than multiple println statements, and call the `wrap` function to ensure the lines are not too long please

lpil (commented):
That's really nice! Left one comment inline

Guria:
@lpil updated

Question about `cli::confirm_with_text` is still relevent. Should I use it here or relying on `--yes` flag is ok here?

lpil (approved):
Thank you!!

Guria:
@lpil looks like we miss changelog entry for this one

lpil:
Oh dang, sorry I missed that. Could you add one please 🙏

Guria:
@lpil  #3940

PR #3840, opened 2024-11-17
Improve error message when `gleam add`-ing package that doesn't exist

### LINKED ISSUE #3839: Provide actionable information when a package cannot be found with multiple packages and `gleam add`
I recently tried to add two packages to my project with the following command:

```
gleam add --dev globlin globin_fs
```

There's a typo here, it should say **globlin_fs**. This causes the package resolution to error out with the following error:

```
  Resolving versions

An error occurred while determining what dependency packages and
versions should be downloaded.

Unable to determine package versions: resource was not found
```

But the error does not tell me which package could not be found, presumably because we didn't consider the case where multiple packages may be added at once.

The DX is bad here, if I were adding many packages at once it might be a bit of a hassle to spot a subtle typo like this. It'd be great if the error message also told me what specific resource could not be found!

giacomocavalieri:
This PR fixes #3839

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
   exist with `gleam add`.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
-### Language Server
+### Language server
 
 ### Formatter
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### Build tool
 
+- Improved the error message you get when trying to add a package that doesn't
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -151,8 +159,6 @@
       }
   ```
 
-  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
```
lpil, line 154: Did you mean to remove this?

giacomocavalieri: Oops, fixed it!

Issue #3836, opened 2024-11-16
Hint at tuple syntax when struggling with parenthesis


michaeljones:
From a [discord interaction](https://discord.com/channels/768594524158427167/768594524158427170/1307378142440325181), the error message:
```
error: Syntax error
    ┌─ /home/michael/root/projects/contact-stack/src/contact_stack.gleam:134:34
    │
134 │       let urls = dict.from_list([("html", "https://www.contact-stack.com/friends/" <> int.to_string(row.id))])
    │                                  ^ This parenthesis cannot be understood here

Hint: To group expressions in gleam use "{" and "}".
```
Could benefit from the additional text

> and tuples are created with `#(` and `)`.

I don't know if that applies to all instances of that error but for me it was the issue and I think tuples without `#` are common enough in other languages that a hint is valid.

Louis seemed to think this was a reasonable suggestion.

lpil:
Thank you

PR #3845, opened 2024-11-17
refactor(parser): add hint of how tuples are created

vit0rr:
Closes https://github.com/gleam-lang/gleam/issues/3836

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -65,7 +65,10 @@ impl ParseError {
             ),
             ParseErrorType::ExprLparStart => (
                 "This parenthesis cannot be understood here",
-                vec!["Hint: To group expressions in gleam use \"{\" and \"}\".".into()],
+                vec![
+                    "Hint: To group expressions in gleam use \"{\" and \"}\".".into(),
+                    "Hint: Tuples are created with `#(` and `)`.".into(),
+                ],
             ),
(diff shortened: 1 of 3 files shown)


compiler-core/src/parse/error.rs
```diff
@@ -65,7 +65,10 @@ impl ParseError {
             ),
             ParseErrorType::ExprLparStart => (
                 "This parenthesis cannot be understood here",
-                vec!["Hint: To group expressions in gleam use \"{\" and \"}\".".into()],
+                vec![
+                    "Hint: To group expressions in gleam use \"{\" and \"}\".".into(),
+                    "Hint: Tuples are created with `#(` and `)`.".into(),
```
GearsDatapacks, line 70: Instead of putting "hint" twice, I think it would be better to use one sentence.

vit0rr: Agree.
Something like this?

```rust
ParseErrorType::ExprLparStart => (
    "This parenthesis cannot be understood here",
    vec![
        "Hint: To group expressions in Gleam, use \"{\" and \"}\"; tuples are created with `#(` and `)`.".into(),
    ]
),
```

GearsDatapacks: Yeah, or just "and"

lpil (approved):
Thank you!

Issue #2680, opened 2024-03-12
Linux deletion error on `gleam test/run/build` etc.


BenStigsen:
**Gleam Version:** v1.0.0
**Operating System:** Fedora 39

This is what happens:
```
$ gleam new website
$ cd website
$ gleam test
  Resolving versions
Downloading packages
 Downloaded 2 packages in 0.18s
error: File IO failure

An error occurred while trying to delete this directory:

/mnt/sda1/Projects/Gleam/website/build/dev/erlang

The error message from the file IO library was:
    Directory not empty (os error 39)
```

 `sda1` is NTFS.

This was originally posted in the Discord, but I have decided to open an issue for it here. I'll update this issue within the next 3 hours and try to create a new project on my main drive.

lpil:
Oh dear! Do you know what is happening here to cause the problem? I don't have a good idea.

Caleb-o:
As I mentioned in the discord, I've had this issue as well and it seems to fail between drives. If I move the project to my main drive, then it works fine. It seems to be an odd issue with deleting across drives. I haven't tested on Windows to see if it has the same problem, but I'm running on Arch Linux.

BenStigsen:
Okay, so, it seems that am also able to perform these commands on the main drive, which is `btrfs`.

MedLabs:
I have exactly the same issue, using linux and trying this on a NTFS partition

lpil:
Closing this to use the other issue to track

srackham:
I have the same issue on both NixOS and Manjaro Linux.

To reproduce the error create and run a new project on a Linux NFS mount.

- Environments: NixOS 24.05 running gleam 1.4.1 and on Manjaro Linux running gleam 1.5.1.
- NFS mount:

      nuc1:/srackham on /home/srackham/share type nfs4 (rw,relatime,vers=4.2,rsize=524288,wsize=524288,namlen=255,hard,proto=tcp,timeo=600,retrans=2,sec=sys,clientaddr=192.168.1.125,local_lock=none,addr=192.168.1.243)

To reproduce the error:
    
```
$ gleam --version
gleam 1.4.1

$ gleam new vars
Your Gleam project vars has been successfully created.
The project can be compiled and tested by running these commands:

        cd vars
        gleam test


$ cd vars

$ gleam run
  Resolving versions
Downloading packages
 Downloaded 2 packages in 1.01s
error: File IO failure

An error occurred while trying to delete this directory:

[...]

PR #3846, opened 2024-11-17
Prevent deleting locked build directory during version check

### LINKED ISSUE #2700: Deleting Error \build\dev\erlang
Related to https://github.com/gleam-lang/gleam/issues/2680
Gleam Version: v1.0.0
Operating System: Windows 10

Problem:
If i try to run `gleam run`, `gleam build`,  `gleam test` or `gleam check` i get Following error:
```
error: File IO failure
An error occurred while trying to delete this directory:
    Z:\...\build\dev\erlang
    Die Datei kann nicht geöffnet werden, weil sie gerade gelöscht wird. (os error 303)
```

ivanjermakov:
Fixes #2700 and #2680.

The cause of `Directory not empty (os error 39)` when recreating build dir within `check_gleam_version()` is that called `delete_directory` is trying to remove a directory while lock file is still open by `BuildLock::lock()`. This works ok on ext4, but fails with error on NTFS ([and perhaps BRTFS](https://github.com/gleam-lang/gleam/issues/2700#issuecomment-2317087702)) partitions.

Solution I came up with is to lock directory only once compiler is sure that recreating build dir is not needed, thus checking build version before locking.

`cargo test` is passing, by local test of `gleam run` and `gleam lsp` on NTFS partition fix is working as expected, `cargo fmt` and `cargo clippy` are happy. Let me know if any changes needed.


compiler-cli/src/build.rs
```diff
@@ -48,7 +48,6 @@ pub(crate) fn main_with_warnings(
 
     tracing::info!("Compiling packages");
     let result = {
-        let _guard = lock.lock(telemetry);
         let compiler = ProjectCompiler::new(
```
ivanjermakov, line 51: Is it ok that dir is not locked during this call?

compiler-core/src/error.rs
```diff
@@ -3626,6 +3630,8 @@ or you can publish it using a different version number"),
                 location: None,
                 hint: Some("Please add the --replace flag if you want to replace the release.".into()),
             }],
+            Error::BuildVersionMismatch => vec![],
```
ivanjermakov, line 3633: I feel like these two not fit well to the rest of the values, because they are manually handled as a result of a `check_gleam_version` call. Is there a better way of doing it?

lpil (commented):
Fantastic detective work!

Unfortunately we need to lock the build directory here. We could resolve the issue by deleting the containing files (excluding the lock) rather than deleting the parent directory?

ivanjermakov:
@lpil Makes sense, then I'll update `FileSystemWriter::delete_directory` so that deleting everything but locked files gives `Ok` response.

lpil:
@ivanjermakov ah! or how about we move the lock to the top level? That would be easier

ivanjermakov:
What about the lock for packages?

https://github.com/gleam-lang/gleam/blob/9e2977fa6e924246bcef65ddf48f94ef97959c34/compiler-cli/src/dependencies.rs#L276

Should it stay `gleam.lock` in the project root?

PR #3848, opened 2024-11-19
fix cli update use_manifest logic

SnakeDoc:
Fixes the `use_manifest` logic for the `update()` function. It is currently flipped to the opposite of what it should be.


SnakeDoc:
I should have caught this in the rc's, but didn't find the time to play with it. :disappointed:

PR #3850, opened 2024-11-19
Hovering for argument labels

### LINKED ISSUE #3849: Hovering for argument labels
Currently, hovering on the label of an argument shows the type of the whole function/pattern. I would expect it to behave the same as hovering over the argument value itself.

For example:
```gleam
fn add(wibble a, wobble b) {
  a + b
}

fn main() {
  add(wibble: 1, wobble: 2)
  //          ^ Hovering here correctly shows `Int`
  add(wibble: 1, wobble: 2)
  //  ^ Hovering here shows `fn(Int, Int) -> Int`, but it should probably also show `Int`
}
```

I already have a mostly working implementation for this but I thought I'd make an issue first

GearsDatapacks:
Closes #3849 
This PR implements hovering for argument labels.
I wasn't sure whether this is considered a bug fix or a new feature, but other similar changes were put under bug fixes so I did the same for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 ### Language Server
 
+- The language server now provides type information when hovering over argument
+  labels.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Formatter
 
 ### Bug fixed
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you!

PR #3851, opened 2024-11-19
Deprecate shorthand target names

### LINKED ISSUE #3826: Small inconsistency in target strings
In Gleam v1.6.0-rc-1, the following code is valid and compiles:
```gleam
@target(js)
fn wibble() {
  todo
}
```
But this doesn't:
```gleam
@external(js, "./wibble.mjs", "wobble")
fn wibble() -> Nil
```

The shorthands, `js` and `erl` are allowed in various places such as `@target` and the `--target` flag for `gleam build` and `gleam run`. However, they are not allowed in `@external`. Small point, but I would expect the same set of values to be accepted in all such places.

GearsDatapacks:
Closes #3826

### DIFF
--- a/compiler-core/src/warning.rs
+++ b/compiler-core/src/warning.rs
@@ -1,5 +1,6 @@
 use crate::{
     ast::{SrcSpan, TodoKind},
+    build::Target,
     diagnostic::{self, Diagnostic, Location},
     error::wrap,
     type_::{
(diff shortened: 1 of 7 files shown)


compiler-core/src/warning.rs
```diff
@@ -201,6 +201,11 @@ pub enum DeprecatedSyntaxWarning {
     /// ```
     ///
     DeprecatedRecordSpreadPattern { location: SrcSpan },
+
+    DeprecatedTargetShorthand {
+        full_name: &'static str,
```
lpil, line 206: This should be `target: Target` as it's always one of those two values

lpil (commented):
One small comment inline. Thank you!

lpil (approved):
Wonderful, thank you!

PR #3855, opened 2024-11-19
Move build locks to `build` directory so open files do not interrupt clearing old builds

### LINKED ISSUE #2700: Deleting Error \build\dev\erlang

ivanjermakov:
Fixes #2700 and #2680.

The cause of `Directory not empty (os error 39)` when recreating build dir within `check_gleam_version()` is that called `delete_directory` is trying to remove a directory while lock file is still open by `BuildLock::lock()`. This works ok on ext4, but fails with error on NTFS ([and perhaps BRTFS](https://github.com/gleam-lang/gleam/issues/2700#issuecomment-2317087702)) partitions.

Solution is to move lock files up in the file tree from the directory that is being removed.

New lock file paths:

```
$ROOT/build/packages/gleam.lock
$ROOT/build/gleam-dev-erlang.lock
$ROOT/build/gleam-dev-javascript.lock
$ROOT/build/gleam-prod-erlang.lock
$ROOT/build/gleam-prod-javascript.lock
$ROOT/build/gleam-lsp-erlang.lock
$ROOT/build/gleam-lsp-javascript.lock
```

### DIFF
--- a/compiler-cli/src/build_lock.rs
+++ b/compiler-cli/src/build_lock.rs
@@ -10,32 +10,41 @@ use strum::IntoEnumIterator;
 #[derive(Debug)]
 pub(crate) struct BuildLock {
     directory: Utf8PathBuf,
+    filename: String,
 }
 
 impl BuildLock {
     /// Lock the build directory for the specified mode and target.
     pub fn new_target(paths: &ProjectPaths, mode: Mode, target: Target) -> Result<Self> {
-        let build = paths.build_directory_for_target(mode, target);
-        crate::fs::mkdir(&build)?;
-        Ok(Self { directory: build })
+        let directory = paths.build_directory();
+        crate::fs::mkdir(&directory)?;
+        Ok(Self {
+            directory,
+            filename: format!("gleam-{mode}-{target}.lock"),
+        })
     }
 
     /// Lock the packages directory.
     pub fn new_packages(paths: &ProjectPaths) -> Result<Self> {
-        let packages = paths.build_packages_directory();
-        crate::fs::mkdir(&packages)?;
+        let directory = paths.build_packages_directory();
+        crate::fs::mkdir(&directory)?;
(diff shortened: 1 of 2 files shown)


compiler-cli/src/build_lock.rs
```diff
@@ -8,34 +8,59 @@ use gleam_core::{
 use strum::IntoEnumIterator;
 
 #[derive(Debug)]
-pub(crate) struct BuildLock {
-    directory: Utf8PathBuf,
+pub(crate) enum BuildLock {
```
ivanjermakov, line 11: Is it ok for it to be enum? Or we can construct `lock_filename` right in the constructor

lpil: Making the path in the constructor sounds good to me

lpil (commented):
Wonderful work, thank you! Would you mind updating the CHANGELOG file also? Thanks

GearsDatapacks:
Fyi merge commits are not allowed in this repo. You'll need to rebase to remove them before this PR will be merged

ivanjermakov:
@GearsDatapacks thanks, I missed that when checking `CONTRIBUTING.md`.

GearsDatapacks:
Scanning through, it doesn't look like it's mentioned. I want to improve upon the contributing documents at some point, so maybe I'll try and add that

lpil (approved):
Perfect, thank you!

PR #3857, opened 2024-11-20
Report error for missing exponents in float literal

### LINKED ISSUE #3856: Compilation fails when no digits are specified for exponent
Currently, the compiler allows things like `1.5e`, with an empty exponent, as a valid float literal. However, this is not valid syntax in either erlang or javascript, and so the erlang compiler and javascript parser both throw errors if this is used.
I think we should disallow this syntax. From my understanding it would not be considered a breaking change as you cannot write valid code containing such syntax.

GearsDatapacks:
Fixes #3856

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@
 
 ### Bug fixed
 
+- The compiler now throws an error when a float literal ends with an `e` and
+  is missing an exponent.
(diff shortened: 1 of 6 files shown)


lpil (approved):
Thank you!!!! Super fast as always 🥷

GearsDatapacks:
Ah, looks like I forgot to accept the snapshot

PR #3860, opened 2024-11-20
Fix variant inference for prelude types

GearsDatapacks:
One small thing I noticed while playing around:
```gleam
pub type Wibble {
  Wibble
  Wobble
}

pub fn main() {
  let always_wibble = Wibble
  let always_ok = Ok(Nil)

  case always_wibble {
    Wibble -> todo
  }
  // ^ This compiles fine
  case always_ok {
    Ok(Nil) -> todo
  }
  // ^ This throws an error: "Missing pattern: Error(_)"
}
```

It's unlikely to be much of a problem in real code, but for consistency this PR fixes it by adding the `inferred_variant` field to prelude values

### DIFF
--- a/compiler-core/src/type_/prelude.rs
+++ b/compiler-core/src/type_/prelude.rs
@@ -87,14 +87,20 @@ pub fn bool() -> Arc<Type> {
     bool_with_variant(None)
 }
 
-pub fn bool_with_variant(variant_index: Option<u16>) -> Arc<Type> {
+pub fn bool_with_variant(variant: Option<bool>) -> Arc<Type> {
+    let variant = match variant {
+        Some(true) => Some(0),
+        Some(false) => Some(1),
+        None => None,
+    };
+
     Arc::new(Type::Named {
         args: vec![],
         publicity: Publicity::Public,
         name: BOOL.into(),
         module: PRELUDE_MODULE_NAME.into(),
         package: PRELUDE_PACKAGE_NAME.into(),
-        inferred_variant: variant_index,
+        inferred_variant: variant,
     })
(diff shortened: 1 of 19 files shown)


GearsDatapacks:
I've had to modify several existing tests as this change made some of them obsolete, for example checking exhaustiveness errors:
```gleam
case True {
  True -> todo
}
// ^ This now compiles, rendering the test useless
```

compiler-core/src/type_/prelude.rs
```diff
@@ -84,13 +84,17 @@ pub fn float() -> Arc<Type> {
 }
 
 pub fn bool() -> Arc<Type> {
+    bool_with_variant(None)
+}
+
+pub fn bool_with_variant(variant_index: Option<u16>) -> Arc<Type> {
```
lpil, line 90: Let's pass in a bool as there's 2 variants

lpil (commented):
Nice, thank you. One small comment

compiler-core/src/type_/prelude.rs
```diff
@@ -84,13 +84,23 @@ pub fn float() -> Arc<Type> {
 }
 
 pub fn bool() -> Arc<Type> {
+    bool_with_variant(None)
+}
+
+pub fn bool_with_variant(variant: Option<bool>) -> Arc<Type> {
```
lpil, line 90: It says with variant so always take the variant please 🙏 
```suggestion
pub fn bool_with_variant(variant: bool) -> Arc<Type> {
```

lpil (approved):
Thank you!

PR #3866, opened 2024-11-20
Fix compiler crash with variant inference

### LINKED ISSUE #3861: Compiler panic with variant inference
I found another bug with variant inference, while implementing #3860. It's pretty unlikely to occur in real code, but it causes a compiler crash in a very specific case:
```gleam
pub type Wibble {
  Wibble
  Wobble
}

pub fn main() {
  case Wibble, Wobble {
    Wibble, Wibble -> todo
  }
}
```

This causes a crash, because the variant inference algorithm thinks that `Wibble` is a local variable, and narrows it in the scope of the `case` arm. However, this causes `Wibble` to be treated as a local variable, and that goes against what the compiler expects (local variables can't have capital letters). This causes the compiler to crash.

GearsDatapacks:
Fixes #3861

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,10 @@
   is missing an exponent.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where the compiler would crash when pattern matching on multiple
+  subjects and one of them being a constant record.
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you

PR #3871, opened 2024-11-21
Refactor code actions

giacomocavalieri:
This PR does two things:
- I create a `LineNumbers` value that is passed in to all code action builders instead of rebuilding it from scratch every time. This means we will iterate through the source code just once instead of one time for each kind of code action there is
- I added a small `LineNumbersHelper` struct to make it a bit easier to produce code edits, especially ones meant to delete a specific srcspan that were quite hard to understand at a glance

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -87,12 +87,12 @@ impl<'a> LineNumbersHelper<'a> {
     }
 
     fn src_span_to_lsp_range(&self, location: SrcSpan) -> Range {
-        src_span_to_lsp_range(location, &self.line_numbers)
+        src_span_to_lsp_range(location, self.line_numbers)
     }
 
     fn edit_src_span(&self, location: SrcSpan, new_text: String) -> TextEdit {
         TextEdit {
-            range: src_span_to_lsp_range(location, &self.line_numbers),
+            range: src_span_to_lsp_range(location, self.line_numbers),
             new_text,
         }
     }
@@ -327,7 +327,7 @@ impl<'ast> ast::visit::Visit<'ast> for LetAssertToCase<'_> {
         // we only check for the code action between the `let` and `=`.
         let code_action_location =
             SrcSpan::new(assignment.location.start, assignment.value.location().start);
-        let code_action_range = src_span_to_lsp_range(code_action_location, &self.line_numbers);
+        let code_action_range = src_span_to_lsp_range(code_action_location, self.line_numbers);
 
         self.visit_typed_expr(&assignment.value);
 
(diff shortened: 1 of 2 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -75,6 +75,37 @@ impl CodeActionBuilder {
     }
 }
 
+/// A little wrapper around LineNumbers to make it easier to build text edits.
+///
+struct LineNumbersHelper<'a> {
```
lpil, line 80: This isn't a great name as `Helper` doesn't mean anything. I think this class is for constructing `TextEdits`? If so can we have a name relating to that.

compiler-core/src/language_server/code_action.rs
```diff
@@ -75,6 +75,37 @@ impl CodeActionBuilder {
...
+impl<'a> LineNumbersHelper<'a> {
+    fn new(line_numbers: &'a LineNumbers) -> Self {
+        LineNumbersHelper { line_numbers }
+    }
+
+    fn src_span_to_lsp_range(&self, location: SrcSpan) -> Range {
+        src_span_to_lsp_range(location, self.line_numbers)
+    }
+
+    fn edit_src_span(&self, location: SrcSpan, new_text: String) -> TextEdit {
```
lpil, line 93: The names of the method in this class could be more consistent. Perhaps `insert`, `replace`, and `delete`?

```rust
self.edits.delete(some_span);
```
```rust
self.edits.replace(some_span, some_content);
```

The class could also hold the edits vector.

Could you make the methods used outside this class public please? To make the intended API more clear.

compiler-core/src/language_server/engine.rs
```diff
@@ -298,19 +298,42 @@ where
...
-            code_action_fix_names(module, &params, &this.error, &mut actions);
-            code_action_import_module(module, &params, &this.error, &mut actions);
-            code_action_add_missing_patterns(module, &params, &this.error, &mut actions);
-            actions.extend(LetAssertToCase::new(module, &params).code_actions());
-            actions.extend(RedundantTupleInCaseSubject::new(module, &params).code_actions());
-            actions.extend(LabelShorthandSyntax::new(module, &params).code_actions());
-            actions.extend(FillInMissingLabelledArgs::new(module, &params).code_actions());
-            actions.extend(DesugarUse::new(module, &params).code_actions());
-            AddAnnotations::new(module, &params).code_action(&mut actions);
+            let line_numbers = LineNumbers::new(&module.code);
```
lpil, line 301: How about we call this `lines` to reduce wrapping below?

lpil (commented):
Thank you!

lpil (approved):
Oh that's very nice! Thank you

PR #3872, opened 2024-11-21
Add canonical link

### LINKED ISSUE #3864: Add canonical URL links to HTML documentation to prevent search engines serving old versions of docs
Search engines are surfing links to outdated versions of package documentation. This is a problem as it misleads users, especially for core libraries such as the stdlib.

If the documentation had `canonical` links in the `head` element then search engines would hopefully suggest the current version instead.

```html
<link rel="canonical" href="https://hexdocs.pm/package_name/path/to/file.html" />
```

Add this link to docs when publishing.

rockerBOO:
- Initially adding canonical links if project/module name are available.
- add head block to `<head>`

There are snapshots as well but wanted to get this committed and reviewed. 

Fixes #3864 

Thanks.


lpil (commented):
Did you forgot to add the snapshots here? 😁 

Could you rebase on main please? To fix the new lint issues.

compiler-core/src/docs/snapshots/gleam_core__docs__tests__canconical_link.snap
```diff
@@ -0,0 +1,361 @@
...
+    <title>app · test_canconical · v0.1.0</title>
+    <meta name="description" content=""/>
+    <meta name="theme-color" content="#ffaff3" media="(prefers-color-scheme: light)"/>
+    <meta name="theme-color" content="#33384d" media="(prefers-color-scheme: dark)"/>
+    <link rel="stylesheet" href="./css/index.css?v=GLEAM_VERSION_HERE" type="text/css"/>
+    <!-- The docs_config.js file is provided by HexDocs and shared
+         between multiple versions of the same package. -->
+    <script src="./docs_config.js"></script>
+    <link id="syntax-theme" rel="stylesheet" href="./css/atom-one-light.min.css?v=GLEAM_VERSION_HERE"/>
+    <link rel="canconical" href="./test_canconical/app.html" />
```
lpil, line 21: Canonical URLs need to be absolute!

compiler-core/src/docs/snapshots/gleam_core__docs__tests__markdown_code_from_standalone_pages_is_not_trimmed.snap
```diff
@@ -18,6 +18,7 @@ expression: "compile_with_markdown_pages(config, vec![], pages)"
          between multiple versions of the same package. -->
     <script src="./docs_config.js"></script>
     <link id="syntax-theme" rel="stylesheet" href="./css/atom-one-light.min.css?v=GLEAM_VERSION_HERE"/>
+
```
lpil, line 21: Could you configure all the tests to have them please 🙏

compiler-core/src/docs/snapshots/gleam_core__docs__tests__canconical_link.snap
rockerBOO: How do we know the absolute URL here? Since users could host it themselves; I am not aware of where this setting is.

lpil: It should be set via configuration when publishing to Hex. We can omit otherwise for now

compiler-core/src/docs/snapshots/gleam_core__docs__tests__markdown_code_from_standalone_pages_is_not_trimmed.snap
rockerBOO: These are not displayed because the project_name is blank.

lpil: I am asking you to change that 🙏

compiler-core/templates/documentation_page.html
```diff
@@ -7,3 +7,9 @@
 {% block content %}
 {{ content|safe }}
 {% endblock %}
+
+{% block head -%}
+{% if !host.is_empty() && !project_name.is_empty() -%}
+<link rel="canonical" href="{{ host }}/{{ project_name|safe }}/{{ file_path|safe }}" />
+{%- endif %}
+{%- endblock %}
```
lpil, line 15: Is there a reason not to have this in the layout? It seems to be the same for both templates.

compiler-core/src/docs/snapshots/gleam_core__docs__tests__markdown_code_from_standalone_pages_is_not_trimmed.snap
lpil: Yes! I'm asking you to add it to all of them 🙏  Thank you.

A test case for rendering not-for-hex would also be great.

lpil (commented):
Fab! Couple tiny notes inline. Could you update the changelog also please

compiler-core/templates/documentation_page.html
rockerBOO: Originally was more direct to put it where I knew it was being used, but has converged enough that I was able to move it directly into the layout.

lpil (approved):
Beautiful!! Thank you

PR #3874, opened 2024-11-21
Add proper hovering for patterns in `use` expressions

### LINKED ISSUE #3847: Hovering over a field, when destructuring records, should produce the field type
```gleam
use ParseData(is_positive, next_index, tokens) <- result.try(parse_data)
```

I would like to be able to hover over `tokens` and get its type, but currently, we get the overall `ParseData` record type on hover:

SCR-20241118-mkiz

Hovering on `ParseData` should produce `ParseData(Bool)` and hovering on tokens should produce `List(Tokens)`

GearsDatapacks:
Fixes #3847 
This PR implements `find_node` for `use` expressions, allowing them to be used as expected with hover and goto definition (and anything else that uses `find_node`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The language server now provides correct information when hovering over
+  patterns in use expressions.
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you

PR #3875, opened 2024-11-22
Implement code action to convert inexhaustive `let` to `case`

### LINKED ISSUE #3868: LSP Code Action: Convert inexhaustive `let` to `case`
For example, you have this code:
```gleam
pub type Wibble {
  Wibble(wibble: Int)
}

pub fn main(wibble: Wibble) {
  let Wibble(inner) = wibble
  do_something_with(inner)
}
```
But then you change the `Wibble` type:
```gleam
pub type Wibble {
  Wibble(wibble: Int)
  Wobble(wobble: Int)
}
```
The `let` expression now fails to compile. It could be useful to have a code action to turn:
```gleam
let Wibble(inner) = wibble
```
into:
```gleam
let inner = case wibble {
  Wibble(inner) -> inner
  _ -> todo
}
```
This would be very similar to the [code action to convert](https://github.com/gleam-lang/gleam/blob/31966a0b3503d9429cc016d29b1ba88569e22fc6/compiler-core/src/language_server/code_action.rs#L299) `let assert` to `case`, except it would only be offered if the `let` expression is not exhaustive.

GearsDatapacks:
Closes #3868

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -982,12 +982,6 @@ pub fn code_action_add_missing_patterns(
             indent_size += 1;
         }
 
-        // TODO: It would be nice to use the `print_case_expression` function here,
-        // however that causes things like comments to be removed, which is undesirable.
-        // In future we should work out a way to use that function, as the code below is
-        // pretty convoluted and difficult to understand, because it accounts for a set of
-        // very specific edge-cases. A generic formatting function would be much better.
-
         let indent = " ".repeat(indent_size);
 
         // Insert the missing patterns just after the final clause, or just before
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -586,10 +586,10 @@ struct CaseClause<'a> {
     expression: &'a str,
 }
 
-fn print_case_expression<'a>(
+fn print_case_expression(
     indent_size: usize,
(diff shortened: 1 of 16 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -813,6 +969,12 @@ pub fn code_action_add_missing_patterns(
             indent_size += 1;
         }
 
+        // TODO: It would be nice to use the `print_case_expression` function here,
+        // however that causes things like comments to be removed, which is undesirable.
+        // In future we should work out a way to use that function, as the code below is
+        // pretty convoluted and difficult to understand, because it accounts for a set of
+        // very specific edge-cases. A generic formatting function would be much better.
```
lpil, line 976: Don't add open ended todos comments please

lpil (approved):
Thank you!!

lpil:
Ah! Could you rebase on main please

compiler-core/src/language_server/code_action.rs
GearsDatapacks: Alright. Just a thought I had. Should I leave the comment there without the `TODO`, or remove it entirely?

lpil: Remove it entirely, it's not actionable and work to go should go in the issue tracker

GearsDatapacks:
Rebased and comment removed.

Issue #3843, opened 2024-11-17
Notify of new major versions available when running `gleam update`


lpil:
### Discussed in https://github.com/gleam-lang/gleam/discussions/3841

<div type='discussions-op-text'>

<sup>Originally posted by **hayleigh-dot-dev** November 17, 2024</sup>
As the ecosystem grows it is becoming more difficult to keep track of when packages are updated with new breaking changes. Short of just making a list of every dependency of every package I maintain and regularly checking for new versions, there is no real way to know if any of my dependencies have updated until a user opens an issue or I notice by happenstance.

One thing I _can_ do reliably is run `gleam update` in my packages from time to time. I'd like to propose that command also reports when any major version bumps are available, something like:

```
 % gleam deps update
  Resolving versions
Downloading packages
 Downloaded 2 packages in 0.01s
 
Hint: the following dependencies have new major versions available...

- wibble@2
- wobble@3
```

I think this would have a general benefit to the community:

- package consumers get notified when things have moved on and can make an informed decision to upgrade or not.
- package authors have an easier time broadcasting new changes, which is particularly impactful for "foundation" packages like `mist` or `lustre` where the community at large benefits from being on the same version.
- package authors can more-easily make sure they're not fragmenting the community by sticking on old versions of dependencies unnecessarily.</div>

andho:
I'm giving this a try.

andho:
It would be better if the output indicates the current (exact) version as well:
```bash
 % gleam deps update
  Resolving versions
  Downloading packages
  Downloaded 2 packages in 0.01s
 
Hint: the following dependencies have new major versions available...

wibble    @1.8.3  ->  @2.2.0
wobbling  @1.3.2  ->  @3.0.0
```

lpil:
Yup it needs to show the latest versions.

Closing as duplicate of https://github.com/gleam-lang/gleam/issues/2583, thanks all

PR #3876, opened 2024-11-22
show latest major version available in `gleam deps update`

andho:
This PR aims to solve #3843.
The sample output from that issue:
```bash
 % gleam deps update
  Resolving versions
Downloading packages
 Downloaded 2 packages in 0.01s
 
Hint: the following dependencies have new major versions available...

- wibble@2
- wobble@3
```

# Todos
- [x] ~~Handle Async: fireoff the checks and show the results at the end.~~ Didn't make sense to do this.
- [ ] Telemetry? Not sure it's needed for this optional part.
- [x] ~~Merge version check into a single call~~ Hexpm API doesn't have this. Went with cache instead.
- [x] Get the direct dependencies only

### DIFF
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -2,7 +2,7 @@ use std::{rc::Rc, time::Instant};
 
 use gleam_core::{
     Result,
-    build::{Built, Codegen, Mode, NullTelemetry, Options, ProjectCompiler, Telemetry},
+    build::{Built, Codegen, NullTelemetry, Options, ProjectCompiler, Telemetry},
     manifest::Manifest,
     paths::ProjectPaths,
     warning::WarningEmitterIO,
@@ -15,7 +15,7 @@ use crate::{
 };
 
 pub fn download_dependencies(paths: &ProjectPaths, telemetry: impl Telemetry) -> Result<Manifest> {
-    crate::dependencies::download(
+    dependencies::download(
         paths,
         telemetry,
         None,
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -20,8 +20,10 @@ pub fn download_dependencies(paths: &ProjectPaths, telemetry: impl Telemetry) ->
         telemetry,
         None,
         Vec::new(),
-        dependencies::UseManifest::Yes,
-        dependencies::CheckMajorVersions::No,
+        dependencies::DependencyManagerConfig {
+            use_manifest: dependencies::UseManifest::Yes,
+            ..Default::default()
+        },
     )
 }
 
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -21,7 +21,6 @@ pub fn download_dependencies(paths: &ProjectPaths, telemetry: impl Telemetry) ->
         None,
         Vec::new(),
         dependencies::DependencyManagerConfig {
-            mode: Mode::Dev,
             use_manifest: dependencies::UseManifest::Yes,
             check_major_versions: dependencies::CheckMajorVersions::No,
         },
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -10,13 +10,19 @@ use gleam_core::{
 
 use crate::{
     build_lock::BuildLock,
-    cli,
-    dependencies::UseManifest,
+    cli, dependencies,
     fs::{self, ConsoleWarningEmitter},
 };
 
 pub fn download_dependencies(paths: &ProjectPaths, telemetry: impl Telemetry) -> Result<Manifest> {
-    crate::dependencies::download(paths, telemetry, None, Vec::new(), UseManifest::Yes)
+    crate::dependencies::download(
+        paths,
+        telemetry,
+        None,
+        Vec::new(),
+        dependencies::UseManifest::Yes,
+        dependencies::CheckMajorVersions::No,
+    )
 }
 
 pub fn main(paths: &ProjectPaths, options: Options, manifest: Manifest) -> Result<Built> {
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -2,7 +2,7 @@ use std::{rc::Rc, time::Instant};
 
 use gleam_core::{
     Result,
-    build::{Built, Codegen, NullTelemetry, Options, ProjectCompiler, Telemetry},
+    build::{Built, Codegen, Mode, NullTelemetry, Options, ProjectCompiler, Telemetry},
     manifest::Manifest,
     paths::ProjectPaths,
     warning::WarningEmitterIO,
@@ -21,8 +21,9 @@ pub fn download_dependencies(paths: &ProjectPaths, telemetry: impl Telemetry) ->
         None,
         Vec::new(),
         dependencies::DependencyManagerConfig {
+            mode: Mode::Dev,
             use_manifest: dependencies::UseManifest::Yes,
-            ..Default::default()
+            check_major_versions: dependencies::CheckMajorVersions::No,
         },
     )
 }
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -659,7 +659,7 @@ pub struct DependencyManagerConfig {
 }
 
 impl DependencyManagerConfig {
-    pub fn into_dependency_manager<Telem: Telemetry, P: dependency::PackageFetcher>(
+    fn into_dependency_manager<Telem: Telemetry, P: dependency::PackageFetcher>(
         self,
         runtime: tokio::runtime::Handle,
         package_fetcher: P,
@@ -678,7 +678,7 @@ impl DependencyManagerConfig {
     }
 }
 
-pub struct DependencyManager<Telem, P> {
+struct DependencyManager<Telem, P> {
     runtime: tokio::runtime::Handle,
     package_fetcher: P,
     mode: Mode,
@@ -692,7 +692,7 @@ where
     P: dependency::PackageFetcher,
     Telem: Telemetry,
 {
-    pub fn get_manifest(
+    fn get_manifest(
         &self,
         paths: &ProjectPaths,
         config: &PackageConfig,
@@ -738,7 +738,7 @@ where
         }
     }
 
-    pub fn download(
+    fn download(
         &self,
         paths: &ProjectPaths,
         new_package: Option<(Vec<(EcoString, Requirement)>, bool)>,
@@ -810,7 +810,7 @@ where
         Ok(manifest)
     }
 
-    pub fn resolve_versions(
+    fn resolve_versions(
         &self,
         project_paths: &ProjectPaths,
         config: &PackageConfig,
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -422,6 +422,12 @@ pub fn download<Telem: Telemetry>(
         .packages
         .iter()
         .map(|manifest_pkg| (manifest_pkg.name.to_string(), manifest_pkg.version.clone()))
+        .filter(|(name, _)| {
+            manifest
+                .requirements
+                .iter()
+                .any(|(required_pkg, _)| name == required_pkg)
+        })
         .collect();
     let major_versions_available = dependency::resolve_major_versions(
         PackageFetcher::boxed(runtime.handle().clone()),
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -422,15 +422,25 @@ pub fn download<Telem: Telemetry>(
     let major_versions_available =
         dependency::check_for_major_version_updates(&manifest, package_fetcher);
     if !major_versions_available.is_empty() {
-        // print to stderr instead of stdout because this is not part of the standard output of this
-        // command
-        eprintln!("\nHint: the following dependencies have new major versions available...\n");
-        for (name, (v1, v2)) in major_versions_available {
-            eprintln!("{}@{} -> {}@{}", name, v1, name, v2);
-        }
+        pretty_print_major_versions_available(major_versions_available);
     }
 
     Ok(manifest)
+}
+
+fn pretty_print_major_versions_available(versions: dependency::PackageVersionDiffs) {
+    let longest_package_name_length = versions.keys().map(|name| name.len()).max().unwrap_or(10);
+
+    // print to stderr instead of stdout because this is not part of the standard output of this
+    // command
+    eprintln!("\nHint: the following dependencies have new major versions available...\n");
+    for (name, (v1, v2)) in versions {
+        let padding = " ".repeat(longest_package_name_length - name.len());
+
+        // lazily assuming the longest version could be 8 characters. eg: 1.1234.2
+        // excluding qualifiers other than x.y.z
+        eprintln!("{name}:{padding} {v1:<8} -> {v2:<8}");
+    }
 }
 
 async fn add_missing_packages<Telem: Telemetry>(
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -81,9 +81,12 @@ pub fn tree(paths: &ProjectPaths, options: TreeOptions) -> Result<()> {
 fn get_manifest_details(paths: &ProjectPaths) -> Result<(PackageConfig, Manifest)> {
     let runtime = tokio::runtime::Runtime::new().expect("Unable to start Tokio async runtime");
     let config = crate::config::root_config(paths)?;
+    let package_fetcher: Box<dyn dependency::PackageFetcher> =
+        PackageFetcher::boxed(runtime.handle().clone());
     let (_, manifest) = get_manifest(
         paths,
         runtime.handle().clone(),
+        &*package_fetcher,
         Mode::Dev,
         &config,
         &cli::Reporter::new(),
@@ -385,11 +388,14 @@ pub fn download<Telem: Telemetry>(
 
     // Start event loop so we can run async functions to call the Hex API
     let runtime = tokio::runtime::Runtime::new().expect("Unable to start Tokio async runtime");
+    let package_fetcher: Box<dyn dependency::PackageFetcher> =
+        PackageFetcher::boxed(runtime.handle().clone());
 
     // Determine what versions we need
     let (manifest_updated, manifest) = get_manifest(
         paths,
         runtime.handle().clone(),
+        &*package_fetcher,
         mode,
         &config,
         &telemetry,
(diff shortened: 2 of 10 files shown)


andho:
> Hello! Are you still working on this one? Thanks

I was hoping to work on it this weekend but some things have come up. I can only get back to it next weekend.

andho:
Just the major version hist without any other output is weird though:
image

Helpful, but looks weird.

inoas:
Could a flag be provided that forces a non-zero exit-code if major deps are not bumped?

compiler-core/src/dependency.rs
```diff
@@ -832,4 +970,91 @@ mod tests {
...
+                (
+                    EcoString::from("depends_on_old_version_of_direct_pkg"),
+                    requirement::Requirement::Hex {
+                        version: Range::new("> 0.1.0 and <= 1.0.0".into()),
+                    },
+                ),
+            ]
+            .into_iter()
+            .collect(),
+            //requirements: HashMap::new(),
```
lpil, line 999: Unused code here

compiler-cli/src/dependencies.rs
```diff
@@ -981,6 +1014,15 @@ impl dependency::PackageFetcher for PackageFetcher {
         &self,
         package: &str,
     ) -> Result<hexpm::Package, Box<dyn std::error::Error>> {
+        {
+            let runtime_cache = self.runtime_cache.borrow();
+            let result = runtime_cache.get(package);
+
+            if let Some(result) = result {
+                return Ok(result.clone());
```
lpil, line 1022: Rather than clone the entire package lets record the highest version here.

compiler-core/src/dependency.rs
```diff
@@ -63,6 +64,65 @@ where
...
+        .iter()
+        .filter_map(|(package, version)| {
+            let Ok(hexpackage) = package_fetcher.get_dependencies(package) else {
+                return None;
+            };
+
+            let Some(latest) = &hexpackage
+                .releases
+                .iter()
+                .map(|release| release.version.clone())
```
lpil, line 86: Can take a reference here and avoid cloning

compiler-core/src/dependency.rs
```diff
@@ -63,6 +64,65 @@ where
...
+                .filter(|version| !version.is_pre())
+                .sorted_by(|a, b| b.cmp(a))
+                .next()
```
lpil, line 89: Using `.max` will be faster than sorting and taking the first

compiler-cli/src/dependencies.rs
```diff
@@ -285,9 +286,31 @@ pub fn download<Telem: Telemetry>(
...
+
+    // print to stderr instead of stdout because this is not part of the standard output of this
+    // command
+    eprintln!("\nHint: the following dependencies have new major versions available...\n");
+    for (name, (v1, v2)) in versions {
+        let padding = " ".repeat(longest_package_name_length - name.len());
+
+        // lazily assuming the longest version could be 8 characters. eg: 1.1234.2
+        // excluding qualifiers other than x.y.z
+        eprintln!("{name}:{padding} {v1:<8} -> {v2:<8}");
```
lpil, line 310: Let's render this to a string and then snapshot test the function. This function can then just print the string and not be responsible for formatting and printing

lpil (commented):
Very nice! I've left some small notes inline. Could you update the changelog too please 🙏

compiler-cli/src/dependencies.rs
andho: Oh! This function is originally used by the `DependencyProvider`. But as the major version need the same info, but to avoid multiple http calls to hexpm, I've added caching here.

I think the way to avoid cloning would be to have PackageFetcher own package (and return a ref) but that would have to touch the dependency resolution part so I didn't want to include in this PR (increased scope).

compiler-core/src/dependency.rs
andho: Haha, this is embarrassing.

compiler-cli/src/dependencies.rs
andho: Here's an implementation with PackageFetcher owning packages and RefCounting: https://github.com/andho/gleam/commit/851b9a6b309e9739f1d2b44e8178831f378f00eb

andho:
> Could a flag be provided that forces a non-zero exit-code if major deps updates are available?
> 
> Would that make sense?
> 
> The idea would be to have a CI check that goes red if deps are not up to date.

@inoas  I think the flag would make more sense in a `gleam deps outdated` command instead.

compiler-core/src/dependency.rs
andho: Fixed

compiler-core/src/dependency.rs
andho: Fixed.

compiler-cli/src/dependencies.rs
andho: Done.

compiler-cli/src/dependencies.rs
andho: Ah I just realized I never wired up the same `PackageFetcher` to be used in other parts of the dependency resolution so the cache is not even used. I'll see about getting that wired up soon.

andho: `PackageFetcher` is now passed by reference in all instances and instantiated in `list` and `download` only.

lpil (commented):
Great, thank you.

image

I gave it a try and while the information is there and useful I don't think it fits in at all with any of the other information printed. How can we show this better?

Should it be shown for all resolution or only for `gleam update`? If the latter it would be easier to make a good output format for it.

compiler-cli/src/dependencies.rs
lpil: It would be best to do that as to not clone everything. Do you think it would be complex or involved?

andho:
> Should it be shown for all resolution or only for `gleam update`?

It makes sense to show for `gleam update` and `gleam deps download`

Although it was helpful when I recently did `gleam build` and I saw that `decode` package was updated to 1.0.0. But due to the execution order it would take some rearranging to get this output to come last.

I'd say for an MVP show only in `gleam update` and `gleam deps download`. I'll only have to add a flag to `download` fn.

lpil:
Sounds great!

compiler-cli/src/dependencies.rs
andho: Changed to pass package as refs using Rc. DependencyProvider still needs to clone the package because it sorts the releases field before passing to pubgrub.

compiler-cli/src/dependencies.rs
```diff
@@ -219,73 +221,75 @@ pub fn download<Telem: Telemetry>(
     // manifest which will result in the latest versions of the dependency
     // packages being resolved (not the locked ones).
     use_manifest: UseManifest,
+    // If true we check for major version updates and print them to the console.
+    check_major_versions: bool,
```
lpil, line 225: Create an enum instead of using bools please. `true` and `false` mean nothing without context so they are hard to understand. See the other arguments for examples.

compiler-cli/src/dependencies.rs
```diff
@@ -219,73 +221,75 @@ pub fn download<Telem: Telemetry>(
...
-        &telemetry,
-    ))?;
-
-    if manifest_updated {
-        // Record new state of the packages directory
-        // TODO: test
-        tracing::debug!("writing_manifest_toml");
-        write_manifest_to_disc(paths, &manifest)?;
+    output_string
+        .push_str("\nHint: the following dependencies have new major versions available:\n\n");
```
lpil, line 272:
```suggestion
        .push_str("\nThe following dependencies have new major versions available:\n\n");
```

compiler-core/src/dependency.rs
```diff
@@ -63,6 +63,61 @@ where
     Ok(packages)
 }
 
+/**
+* Used to compare 2 versions of a package.
+*/
+pub type PackageVersionDiffs = HashMap<String, (Version, Version)>;
```
lpil, line 69: Let's either use this always or remove it

compiler-cli/src/dependencies.rs
```diff
@@ -543,6 +491,237 @@ fn same_requirements(
...
+
+    fn with_check_major_versions(mut self, check_major_versions: bool) -> Self {
+        self.check_major_versions = check_major_versions;
+        self
+    }
+
+    fn should_use_manifest(mut self) -> Self {
+        self.use_manifest = UseManifest::Yes;
+        self
+    }
```
lpil, line 536: This is cool!

Another option which can be convenient at times is to have a struct with public fields and have an `into_dependency_manager()` method or something like that

lpil (commented):
Looks great!! Thank you. I've left two tiny notes inline.

compiler-core/src/dependency.rs
andho: > Let's either use this always or remove it

I'm not sure what you mean. `PackageVersionDiffs` are used only for the major versions check currently. I don't see any other place making comparisons between 2 versions (except probably within pubgrub which makes comparisons betweens many versions probably).

Is you suggestion to remove the type and use `HashMap<String, (Version, Version)>` everywhere?

compiler-cli/src/dependencies.rs
andho: I looked into builder pattern and init pattern (misnamed but similar to what you're suggesting). But as the tokio runtime, Package Fetcher and Telemetry needs to be specified, I cannot implement Default. Which means I'll have to define all struct fields at callsite. Which means any addition or change to the struct will require changes at all callsites.

Alternatively I could go for a builder, but seems too small right now.

andho: Although...I could make it `DependencyManagerConfig.into_dependecy_manager(runtime: ..., package_fetcher: ..., telemetry: ...)`.

compiler-core/src/dependency.rs
lpil: Either always use it or always using the underlying type is fine. We don't want a mix.

lpil:
Are you still planning to work on this? Thank you 🙏

andho:
> Are you still planning to work on this? Thank you 🙏

Hey yeah, will be working on this 🙂. Was a bit busy over the last 2 weeks, will be resuming it this week 🤞

compiler-cli/src/build.rs
```diff
@@ -10,14 +10,19 @@ use gleam_core::{
...
 
 pub fn download_dependencies(telemetry: impl Telemetry) -> Result<Manifest> {
     let paths = crate::find_project_paths()?;
-    crate::dependencies::download(&paths, telemetry, None, Vec::new(), UseManifest::Yes)
+    dependencies::download(
+        &paths,
+        telemetry,
+        None,
+        Vec::new(),
+        dependencies::DependencyManagerConfig::default(),
```
lpil, line 24: This is much less clear than before- please explicitly state each item of config. The programmer should be able to tell what the desired behaviour is at each call-site. This will mean removing the default implementation.

compiler-cli/src/dependencies.rs
```diff
@@ -543,6 +490,242 @@ fn same_requirements(
     Ok(left == right)
 }
 
+pub struct DependencyManagerConfig {
+    pub mode: Mode,
+    pub use_manifest: UseManifest,
+    pub check_major_versions: CheckMajorVersions,
```
lpil, line 496: Document what each of these do please 🙏

compiler-cli/src/dependencies.rs
```diff
@@ -543,6 +490,242 @@ fn same_requirements(
...
+            telemetry,
+
+            mode: self.mode,
+            use_manifest: self.use_manifest,
+            check_major_versions: self.check_major_versions,
+        }
+    }
+}
+
+pub struct DependencyManager<Telem: Telemetry, P: dependency::PackageFetcher> {
```
lpil, line 528: Remove these constraints please 🙏 Constraints are only needed on the functions.

lpil (commented):
Thank you! A few small things

compiler-cli/src/build.rs
andho: Done

compiler-cli/src/dependencies.rs
andho: Done

compiler-cli/src/dependencies.rs
andho: Done

compiler-cli/src/dependencies.rs
```diff
@@ -678,6 +630,239 @@ fn same_requirements(
     Ok(left == right)
 }
 
+pub struct DependencyManagerConfig {
+    /// In `Prod` mode, dev dependencies are not considered during the executed operation.
+    /// Otherwise (`Dev` or `Lsp`), all dependencies are considered
+    pub mode: Mode,
```
lpil, line 636: This has moved this internal mode value into the public interface now. What was your motivation for this? From what I can see it is never set to anything but `Dev`.

andho: Yeah you're right. I didn't intend to move it into the public interface, but to bundle configuration related data into a single struct, to reduce number of arguments passed around.

For now I'll remove `mode` from `DependencyManagerConfig`.

andho: I have also made `DependencyManager` private as it's not directly used outside of `dependencies`.

compiler-cli/src/dependencies.rs
```diff
@@ -678,6 +633,237 @@ fn same_requirements(
...
+            Ok((false, manifest))
+        } else {
+            tracing::debug!("manifest_outdated");
+            let manifest =
+                self.resolve_versions(paths, config, Some(&manifest), packages_to_update)?;
+            Ok((true, manifest))
+        }
+    }
+
+    fn download(
```
lpil, line 726: It was tricky for me to determine that this was the public API of this class. Could you mark this as public and move the class to sub-module please 🙏 

I'm concerned about the API being understandable, particularly since the locking has been moved inside this class, so it is extremely important that the correct entrypoint is used.

lpil (commented):
Looking great! One more thing

compiler-cli/src/dependencies.rs
andho: Done.

lpil (commented):
Beautiful work, thank you so much!

I've been manually testing this and it's really nice and useful, but it prints about dependencies being out of date before they are reported as having been added.

```
$ gleam add lustre@2 gleam_json@1
  Resolving versions
Downloading packages
 Downloaded 5 packages in 0.48s

Hint: the following dependencies have new major versions available:

gleam_json 1.0.1 -> 2.3.0
lustre     2.0.1 -> 4.6.3

      Added lustre v2.0.1
      Added gleam_json v1.0.1
```

I think it would make more sense the other way around

```
$ gleam add lustre@2 gleam_json@1
  Resolving versions
Downloading packages
 Downloaded 5 packages in 0.48s
      Added lustre v2.0.1
      Added gleam_json v1.0.1

The following dependencies have new major versions available:

gleam_json 1.0.1 -> 2.3.0
lustre     2.0.1 -> 4.6.3
```

What do you think?

andho:
> Beautiful work, thank you so much!
> 
> I've been manually testing this and it's really nice and useful, but it prints about dependencies being out of date before they are reported as having been added.
> 
> ```
> $ gleam add lustre@2 gleam_json@1
>   Resolving versions
> Downloading packages
>  Downloaded 5 packages in 0.48s
> 
> Hint: the following dependencies have new major versions available:
> 
> gleam_json 1.0.1 -> 2.3.0
> lustre     2.0.1 -> 4.6.3
> 
>       Added lustre v2.0.1
>       Added gleam_json v1.0.1
> ```
> 
> I think it would make more sense the other way around
> 
> ```
> $ gleam add lustre@2 gleam_json@1
>   Resolving versions
> Downloading packages
>  Downloaded 5 packages in 0.48s
>       Added lustre v2.0.1
>       Added gleam_json v1.0.1
> 
> The following dependencies have new major versions available:
> 
> gleam_json 1.0.1 -> 2.3.0
> lustre     2.0.1 -> 4.6.3
> ```
> 
> What do you think?

Yeah the current output is confusing. But that's actually a bug. Major version availability should be shown only on `update` and `deps download` command as we discussed. I'll take a look.

PS. I'm traveling until 28th feb, so updates might be a bit slow.

andho:
Hi @lpil. I'm not able to reproduce the issue. I think you might have run an old version. "Hint:" had been removed from the output a few commits back.

lpil:
Sorry I'm not sure what issue you mean 😅

andho:
Haha, yeah it's been a while, sorry!

I'm referring to this:
> I've been manually testing this and it's really nice and useful, but it prints about dependencies being out of date before they are reported as having been added.
> 
> ```
> $ gleam add lustre@2 gleam_json@1
>   Resolving versions
> Downloading packages
>  Downloaded 5 packages in 0.48s
> 
> Hint: the following dependencies have new major versions available:
> 
> gleam_json 1.0.1 -> 2.3.0
> lustre     2.0.1 -> 4.6.3
> 
>       Added lustre v2.0.1
>       Added gleam_json v1.0.1
> ```
> 
> I think it would make more sense the other way around
> 
> ```
> $ gleam add lustre@2 gleam_json@1
>   Resolving versions
> Downloading packages
>  Downloaded 5 packages in 0.48s
>       Added lustre v2.0.1
>       Added gleam_json v1.0.1
> 
> The following dependencies have new major versions available:
> 
> gleam_json 1.0.1 -> 2.3.0
> lustre     2.0.1 -> 4.6.3
> ```
> 
> What do you think?

In the latest iteration, the major version check is only triggered for `update` and `deps download` commands. So it won't show up in `add` command.

lpil (commented):
Hello! Sorry for taking so long to get back to you.

I tried to rebase this on main but I got confused as I wasn't familiar enough with the code. Could you rebase this up to date for me please 🙏 Both squashing or preserving history is fine

andho:
@lpil I've rebased on main. Here's a summary of the changes:

1. Added `CheckMajorVersions` enum as a flag to indicate if gleam should check for major version updates.
2. Changed `PackageFetcher` in dependencies to not be boxed so that it's monomorphized instead of using dynamic dispatch.
3. Changed `PackageFetcher` to cache the package info retrieved from hex, so that duplicate calls to hexpm won't be made (during version resolution and then during major version check).
4. Addition of `CheckMajorVersions` caused `too many parameters` lint issue, so those **configuration/context** related values are now moved to `DependencyManager` struct, and the 3 functions that had `too many parameters` has been moved into an impl for `DependencyManager`. Those functions are `download`, `get_manifest` and `resolve_versions`.
   1. I have manually compared the current `main` branch and this PR branch, and there are no difference in those functions between the 2. Manual check is necessary as the methods were moved out to `compiler-cli/src/dependencies/dependency_manager.rs`.
5. `CheckMajorVersions` is enabled only for `gleam [deps] update` and `gleam download`.

lpil (approved):
Thank you so much!!!! Sorry for taking to long to merge this

PR #3886, opened 2024-11-24
Allow qualified records in clause guards

### LINKED ISSUE #3883: `case` guard using qualified constructor from other module does not compile
With `module.gleam` containing:
```gleam
pub type Wibble {
  Wibble
  Wobble
}
```

and 
```gleam
import module

fn wibble(i: Int, w: module.Wibble) {
  case i {
    0 if w == module.Wobble -> 0
    _ -> 42
  }
}
```

compilation fails with 
```
  │
5 │     0 if w == module.Wobble -> 0
  │                      ^^^^^^ I'm expecting a lowercase name here
```

If doing an unqualified import with
```gleam
import module.{Wobble}
```

using `w == Wobble` works as expected.

Gleam version 1.6.1

GearsDatapacks:
Fixes #3883
This PR adds a function which parses qualified records in the same way as unqualified records, by treating them as constant values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,9 @@
   file of same name without warning. It now produces an error.
   ([PgBiel](https://github.com/PgBiel))
 
+- Fixed a bug where qualified records were not allowed in clause guards.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.6.1 - 2024-11-19
 
 ### Bug fixed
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -129,6 +129,9 @@
   file of same name without warning. It now produces an error.
   ([PgBiel](https://github.com/PgBiel))
 
+- Fixed a bug where qualified records were not allowed in clause guards.
+  ([Surya Rose](https://github.com/GearsDatapacks))
```
lpil, line 133: This isn't a bug fix, it's a new feature.

lpil (approved):
Thank you!

PR #3895, opened 2024-11-25
fix string prefix assignments in nested patterns

### LINKED ISSUE #3894: javascript codegen bug with nested case patterns
The following code prints `undefined` on `gleam 1.6.2`

```gleam
import gleam/io

pub type Wibble {
  Wibble(wobble: String)
}

pub fn main() {
  let tmp = Wibble(wobble: "wibble")
  case tmp {
    Wibble(wobble: "w" as wibble <> _wobble) -> io.println(wibble)
    _ -> panic
  }
}
```

It generates the following JS, with `let wibble = "w".wobble;` being the culprit.

```js
import { CustomType as $CustomType, makeError } from "./gleam.mjs";

export class Wibble extends $CustomType {
  constructor(wobble) {
    super();
    this.wobble = wobble;
  }
}

export function main() {
  let tmp = new Wibble("wibble");
  if (tmp instanceof Wibble && tmp.wobble.startsWith("w")) {
    let wibble = "w".wobble; // <<<<<< THIS IS THE ISSUE >>>>>>>
    return $io.println(wibble);
  } else {
    throw makeError(
      "panic",
      "jsgen",
      11,
      "main",
      "`panic` expression evaluated.",
      {}
    )
  }
}
```

Thanks to @joshi-monster for spotting this one!

joshi-monster:
fix #3894

### DIFF
--- a/compiler-core/src/javascript/pattern.rs
+++ b/compiler-core/src/javascript/pattern.rs
@@ -409,7 +409,6 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
         subject: &Document<'a>,
         pattern: &'a TypedPattern,
     ) -> Result<(), Error> {
-        println!("traverse_pattern! {:?}", pattern);
         match pattern {
             Pattern::String { value, .. } => {
                 self.push_equality_check(subject.clone(), expression::string(value));
--- a/compiler-core/src/javascript/pattern.rs
+++ b/compiler-core/src/javascript/pattern.rs
@@ -409,6 +409,7 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
         subject: &Document<'a>,
         pattern: &'a TypedPattern,
     ) -> Result<(), Error> {
+        println!("traverse_pattern! {:?}", pattern);
         match pattern {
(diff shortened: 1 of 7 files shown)


compiler-core/src/javascript/pattern.rs
```diff
@@ -409,6 +409,7 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
         subject: &Document<'a>,
         pattern: &'a TypedPattern,
     ) -> Result<(), Error> {
+        println!("traverse_pattern! {:?}", pattern);
```
GearsDatapacks, line 412: Debug print here :)
(If you use `dbg!`, clippy will warn you!)

joshi-monster: oh no, good spot, thank you :purple_heart: 

you see the thing is I also have to disable clippy otherwise rust-analyser would be literally unusable and its lockfile also blocks recompilation/running the tests

joshi-monster: I'm struggling here ok

GearsDatapacks: I just run it once using `cargo clippy` before pushing, after tests pass

PR #3873, opened 2024-11-21
Mark removed modules as stale

sbergen:
Resolves #3862 - Removed or moved files are not tracked by the compiler

PR #3901, opened 2024-11-26
Use iterator instead of collect() for listing files in `io` module

sbergen:
When iterating Gleam source and cache files, avoids an unnecessary `collect()` to a `Vec`.

As discussed in #3873


lpil (approved):
Nice! Thanks!

Issue #3835, opened 2024-11-15
Result pattern matching hint is confusing when it is a literal result


lpil:
```
error: Type mismatch
    ┌─ /Users/louis/src/gleam/decode/test/decode/zero_test.gleam:885:19
    │
885 │   |> should.equal(Ok(100))
    │                   ^^^^^^^

Expected type:

    Int

Found type:

    Result(Int, a)

Hint: If you want to get a `Int` out of a `Result(Int, a)` you can pattern match
on it:

    case result {
      Ok(value) -> todo
      Error(error) -> todo
    }
```

In this case I think the hint should not be included. It could perhaps be suggested to remove the `Ok` or `Error`.

SoTeKie:
Hi! I'd like to take a stab at this if it's fine :)

lpil:
Thank you!

PR #3906, opened 2024-11-28
Remove hint about unwrapping Result

SoTeKie:
[Related GH issue](https://github.com/gleam-lang/gleam/issues/3835)

There's not an easy way currently to make this error hint more context specific without a decent amount of additional work, so the goal of this PR is to remove it for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Compiler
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you! Could you update the changelog please? 🙏

lpil (approved):
Thank you!!!

PR #3909, opened 2024-11-29
Fix clippy warnings for 1.83

GearsDatapacks:
This PR fixes newly occurring clippy warnings as of Rust v1.83

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -141,7 +141,7 @@ pub struct ModuleAnalyzerConstructor<'a, A> {
     pub package_config: &'a PackageConfig,
 }
 
-impl<'a, A> ModuleAnalyzerConstructor<'a, A> {
+impl<A> ModuleAnalyzerConstructor<'_, A> {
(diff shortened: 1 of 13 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1762,10 +1762,10 @@ fn get_compatible_record_fields<A: std::fmt::Debug>(
             };
 
             // The labels must be the same
-            if !argument
+            if argument
                 .label
                 .as_ref()
-                .is_some_and(|(_, arg_label)| arg_label == first_label)
+                .is_none_or(|(_, arg_label)| arg_label != first_label)
```
lpil, line 1768: this is so weird

lpil (approved):
Thank you!!

compiler-core/src/analyse.rs
GearsDatapacks: I thought so too. We could disable this lint if we wanted

lpil: If it comes up a few times let's give it a think

PR #3910, opened 2024-11-29
Improve performance of bit array slices on JavaScript

richard-viney:
This makes code such as the following O(N) on JavaScript instead of O(N²):

```gleam
pub fn sum_bytes(bytes: BitArray, sum: Int) -> Int {
  case bytes {
    <<x, rest:bytes>> -> sum_bytes(rest, sum + x)
    _ -> sum
  }
}
```

`string.slice()` already avoids copying data, but that optimisation hadn't made it into the prelude.


richard-viney:
Tests failing due to Rust 1.83.

GearsDatapacks:
There's a PR to update to 1.83, so you can just wait for that

lpil (approved):
Beautiful!!!! Thank you so much!

PR #3912, opened 2024-11-29
fix: typo in indexing tuple error

dinkelspiel:
Fixed indexing tuple when size is unknown error saying `to know it size` instead of `to know its size`


lpil (approved):
Thank you! You'll need to update the test snapshots as the format has changed

PR #3913, opened 2024-11-29
Suggest valid project names on InvalidProjectName

diemogebhardt:
This PR is regarding issue https://github.com/gleam-lang/gleam/issues/1728.

Depending on the `InvalidProjectNameReason` we might suggest a valid project name.

In case of the newly added `InvalidProjectNameReason::FormatNotLowercase` it looks like so:

```sh
We were not able to create your project as `Project_Name` does not have the
correct format. Project names may only contain lowercase letters.

Would you like to name your project 'project_name' instead? [y/n]: 
```

Lets take this as a basis to discuss what valid project names we might want to suggest for which `InvalidProjectNameReason`.

### DIFF
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -1,7 +1,7 @@
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::ValueEnum;
 use gleam_core::{
-    erlang,
+    erlang, error,
     error::{Error, FileIoAction, FileKind, InvalidProjectNameReason},
     parse, Result,
 };
@@ -203,16 +203,7 @@ jobs:
 
 impl Creator {
     fn new(options: NewOptions, gleam_version: &'static str) -> Result<Self, Error> {
-        let project_name = if let Some(name) = options.name.clone() {
-            name
-        } else {
-            get_foldername(&options.project_root)?
-        }
-        .trim()
-        .to_string();
-
-        validate_name(&project_name)?;
-
+        let project_name = get_valid_project_name(options.name.clone(), &options.project_root)?;
         let root = get_current_directory()?.join(&options.project_root);
         let src = root.join("src");
         let test = root.join("test");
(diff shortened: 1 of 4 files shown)


diemogebhardt:
Is there any guideline when to `use` at the top of the file or when to use the fully qualified names? I couldn't recognize a pattern.

GearsDatapacks:
I `use` pretty much everything. I think it's up to you really

lpil (commented):

compiler-cli/src/new.rs
```diff
@@ -349,16 +340,103 @@ fn validate_name(name: &str) -> Result<(), Error> {
...
+                        })
+                    }
+                }
+                None => Err(Error::InvalidProjectName {
+                    name: initial_name,
+                    reason,
+                }),
+            }
+        }
+        Err(error) => Err(error),
```
lpil, line 439: This function is quite hard to read! Could you rewrite it to use early returns rather than nested flow control please 🙏

compiler-cli/src/new.rs
```diff
@@ -349,16 +340,103 @@ fn validate_name(name: &str) -> Result<(), Error> {
...
+                }
+            }
+            _ => None,
+        },
+        InvalidProjectNameReason::ErlangReservedWord => Some(format!("{}_app", invalid_name)),
+        InvalidProjectNameReason::ErlangStandardLibraryModule => {
+            Some(format!("{}_app", invalid_name))
+        }
+        InvalidProjectNameReason::GleamReservedWord => Some(format!("{}_app", invalid_name)),
+        InvalidProjectNameReason::GleamReservedModule => None,
```
lpil, line 381: How come this one doesn't use this `_app` suffix?

compiler-cli/src/new.rs
```diff
@@ -349,16 +340,99 @@ fn validate_name(name: &str) -> Result<(), Error> {
...
+            }
+            _ => None,
+        },
+        InvalidProjectNameReason::ErlangReservedWord => Some(format!("{}_app", invalid_name)),
+        InvalidProjectNameReason::ErlangStandardLibraryModule => {
+            Some(format!("{}_app", invalid_name))
+        }
+        InvalidProjectNameReason::GleamReservedWord => Some(format!("{}_app", invalid_name)),
+        InvalidProjectNameReason::GleamReservedModule => Some(format!("{}_app", invalid_name)),
+        InvalidProjectNameReason::FormatNotLowercase => Some(invalid_name.to_lowercase()),
```
lpil, line 382: This one could be invalid still!

lpil: This needs to be addressed, in case you missed my comment!

lpil:
Looking good but! There's some comments of mine above that still need to be addressed!

diemogebhardt:
> Looking good but! There's some comments of mine above that still need to be addressed!

I give up as you either don't seem to see or simply don't read my replies. They have all been addressed, @lpil:

image

Feel free to close this unless you plan to read and reply to my replies to your comments from above. How frustrating.

diemogebhardt:
> Any comment with the "pending" thingy on it is a draft that you have not published yet, I cannot see them!

Ok, this explains everything... 🤦 I wasn't aware of this and got quite frustrated that you seemed to ignore all my comments across all of my PRs. Sorry, @lpil!

diemogebhardt:
> Looking good but! There's some comments of mine above that still need to be addressed!

Can you have a look at this https://github.com/gleam-lang/gleam/pull/3913#discussion_r1872416136 please? This should explain why the handling case of `InvalidProjectNameReason::FormatNotLowercase` cannot be invalid.

compiler-cli/src/new.rs
lpil: Thank you for the detailed explanation, I did not get it at first, my bad

lpil (approved):
Thank you!

PR #3923, opened 2024-12-01
Fix unsafe ints on JS not warning if an @external fn is called prior

richard-viney:
The unsafe integer warning wasn't appearing if an @external function had already been referenced.

There was a test for this but it needed to have the external function call _prior_ to the unsafe integer, not after it.


lpil (approved):
Thank you

PR #3925, opened 2024-12-01
Update Cap'n'proto dependency

GearsDatapacks:
Related: #3920
The API has changed for text, so this PR addresses that. The generator library, `capnpc`, was also updated.


lpil (approved):
Thank you!!

Issue #3928, opened 2024-12-01
feat(wasm): supporting cached dependencies


Beanow:
Currently the WASM compiler only supports a root package without any dependencies.

The [playground](https://playground.gleam.run/) works around this by acting like the contents of `gleam_stdlib` are *submodules* of the package being compiled, rather than dependencies. This approach no longer works when you're trying to use more dependencies than one.

**Suggested changes**

- Add support for loading the `/lib` path in a similar way to the `gleam-cli` compiler.
- Add *optional* support for loading a `gleam.toml` config.

This would be a backwards-compatible incremental feature. And because this would reduce the "drift" between `gleam-cli` and `gleam-wasm` should not introduce any controversial quirks specific to the WASM version.

**Alternative**

One could also attempt to "build project" with the WASM compiler.
So rather than loading modules from build cache, compiling dependencies from a package source.

This is a larger effort though and seems more in scope for a breaking change to the WASM compiler.
Besides, you'd still want to support build caching and need to support loading the config file in that case.

lpil:
Are you having performance problems running the compiler without caching? How long is it taking to run for you?

PR #3929, opened 2024-12-01
feat(wasm): support optional `/gleam.toml`

Beanow:
Issue: https://github.com/gleam-lang/gleam/issues/3928

Related PR: https://github.com/gleam-lang/gleam/pull/3924

Loads config if it's written to `/gleam.toml`, using the previous hard-coded config when it doesn't exist.


lpil:
Thank you for your contribution, but we do not accept unplanned changes.

Issue #2850, opened 2024-03-26
Replace textwrap library with a custom wrap function with a fixed maximum


lpil:
We use the textwrap library for wrapping error messages to a standard terminal width. The library also does other things such as inspect the terminal that we're running in, but we don't need this stuff, and it makes upgrading it harder. Let's replace it with once simple function that wraps text.

We could copy it from the library perhaps.

aznashwan:
I volunteer myself as tribute. ✋

lpil:
Thank you!!

ross-byrne:
Hello! I'm looking at picking up my first issue and this one looks pretty good. @aznashwan it's been a few months, do you still want to work on this? 

I see the functionality we want here and it looks like `textwrap` just wraps the `terminal_size` crate to get the terminal size or set the fallback. I'm guessing we're happy to just use that crate to re-implement this functionality ourselves, right?

lpil:
The work here is to stop taking terminal size into account. We want wrapping but not variable width.

aznashwan:
> Hello! I'm looking at picking up my first issue and this one looks pretty good. @aznashwan it's been a few months, do you still want to work on this?

Goes without saying by now that it had completely gone under my radar back then (I had transitioned to fixing the [`filepath` module](https://github.com/lpil/filepath/pulls)).

Please feel free to go for it, thank you!

ross-byrne:
@lpil Ah OK, even better so! I thought it was just about `textwrap` specifically.

I'll do some testing to see how `textwrap` currently works and try match that behavior (e.g. when it breaks words etc).
I'll try take as much "inspiration" from how they implemented it as I can.

PR #3933, opened 2024-12-02
Replaced textwrap package with custom implementation

ross-byrne:
Resolves: https://github.com/gleam-lang/gleam/issues/2850 

Removes `textwrap` package in favour of custom implementation.

Personally, I'm not convinced this is better than just using `textwrap` and removing the terminal width check:
```rust 
textwrap::fill(text, 75)
```

### Issues
There are some slight differences in output between this solution and `textwrap`'s more sophisticated solution. This has caused some minor differences with some of the snapshots. I've reviewed and updated them but I want to draw attention to it.

**Question:** are we OK with these differences?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -901,7 +901,6 @@ dependencies = [
  "strum",
  "tar",
  "termcolor",
- "textwrap",
  "thiserror",
  "toml",
  "tracing",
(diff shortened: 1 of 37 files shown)


lpil:
> Personally, I'm not convinced this is better than just using textwrap and removing the terminal width check:

The motivation was that it had previously caused build failures, so we want to remove it and simplify the dependency tree. Code is cheaper than dependencies.

ross-byrne:
> The motivation was that it had previously caused build failures

Totally fair, that's good context to have thanks.

lpil (approved):
Thank you!!!

PR #3939, opened 2024-12-02
feat: add `GLEAM_CACERTS_PATH` env variable

### LINKED ISSUE #3246: Add support to custom CA certificate stores

winstxnhdw:
Revival of #3459

Tagging @lpil as he was the previous reviewer.

Closes #3246

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
   making Gleam packages discoverable through global search of HexDocs.
   ([Diemo Gebhardt](https://github.com/diemogebhardt))
 
+- Allow users to set the `GLEAM_CACERTS_PATH` environment variable to specify a
+  path to a directory containing CA certificates to install Hex packages.
+  ([winstxnhdw](https://github.com/winstxnhdw))
+
 ### Language server
 
 ### Formatter
--- a/compiler-cli/src/http.rs
+++ b/compiler-cli/src/http.rs
@@ -10,6 +10,8 @@ use gleam_core::{
 use http::{Request, Response};
 use reqwest::{Certificate, Client};
 
+use crate::fs;
+
 static REQWEST_CLIENT: OnceLock<Client> = OnceLock::new();
 
 #[derive(Debug)]
@@ -50,20 +52,18 @@ fn init_client() -> Result<&'static Client, Error> {
         return Ok(client);
     }
 
-    let certificate_path = std::env::var("GLEAM_CACERTS_PATH").map_err(|error| Error::FileIo {
-        kind: FileKind::Directory,
-        action: FileIoAction::Read,
-        path: Utf8PathBuf::new(),
-        err: Some(error.to_string()),
-    })?;
+    let certificate_path = match std::env::var("GLEAM_CACERTS_PATH") {
+        Ok(path) => path,
+        Err(_) => {
+            return Ok(REQWEST_CLIENT.get_or_init(|| {
+                Client::builder()
+                    .build()
+                    .expect("Failed to create reqwest client")
+            }));
+        }
+    };
 
-    let certificate_bytes = std::fs::read(&certificate_path).map_err(|error| Error::FileIo {
-        kind: FileKind::File,
-        action: FileIoAction::Parse,
-        path: Utf8PathBuf::from(&certificate_path),
-        err: Some(error.to_string()),
-    })?;
-
+    let certificate_bytes = fs::read_bytes(&certificate_path)?;
     let certificate = Certificate::from_pem(&certificate_bytes).map_err(|error| Error::FileIo {
         kind: FileKind::File,
         action: FileIoAction::Parse,
--- a/compiler-cli/src/http.rs
+++ b/compiler-cli/src/http.rs
@@ -6,6 +6,7 @@ use gleam_core::{Error, Result};
 use http::{Request, Response};
 
 static REQWEST_CLIENT: OnceLock<reqwest::Client> = OnceLock::new();
+static CERTS_ENV_VAR: &str = "GLEAM_CACERTS_PATH";
 
 #[derive(Debug)]
 pub struct HttpClient;
@@ -27,7 +28,7 @@ impl gleam_core::io::HttpClient for HttpClient {
             .try_into()
             .expect("Unable to convert HTTP request for use by reqwest library");
         let mut response = REQWEST_CLIENT
-            .get_or_init(reqwest::Client::new)
+            .get_or_init(init_client)
             .execute(request)
             .await
             .map_err(Error::http)?;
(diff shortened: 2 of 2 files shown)


compiler-cli/src/http.rs
```diff
@@ -42,3 +43,21 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+    }
+}
+
+fn get_certificate() -> Result<Certificate, Box<dyn std::error::Error>> {
+    let certificate_path = std::env::var("GLEAM_CACERTS_PATH")?;
+    let certificate_bytes = std::fs::read(&certificate_path)?;
+    let certificate = Certificate::from_pem(&certificate_bytes)?;
+
+    Ok(certificate)
+}
```
lpil, line 63: In the event that the certs paths has been set we want to error if the certs cannot be read or parsed, rather than silently discarded as they are here.

lpil (commented):
Thank you!

It looks like these would be used for all HTTP requests rather than just for the Hex API. Is that what we want or would that be too restrictive if we wanted to make requests to other domains too?

How might we test this change?

compiler-cli/src/http.rs
winstxnhdw: I don't think this is usual behaviour.. It's quite common for runtimes like Python and Node to ignore missing certs as it is common behaviour to add the environment variable to your shell and not populate the certificate path until necessary.

winstxnhdw:
> It looks like these would be used for all HTTP requests rather than just for the Hex API. Is that what we want or would that be too restrictive if we wanted to make requests to other domains too?

Again, it seems standard to have the certificates apply to all HTTP requests. In fact, a partial application would be surprising behaviour.

winstxnhdw:
Hey, any updates?

lpil:
I just tested this and node does not silently discard the error, it emits an ugly looking warning.

Make it return an error if an error occurs, thank you.

winstxnhdw:
> I just tested this and node does not silently discard the error, it emits an ugly looking warning.
> 
> Make it return an error if an error occurs, thank you.

I am not sure what you mean. I am returning an error. Can you describe how you reproduced that warning? I am not getting an issue on my end.

lpil:
I set the variable and then ran npm install.

winstxnhdw:
Any ideas how I can propagate the error out of `get_or_init`?

lpil (commented):
You would need to not use get_or_init

compiler-cli/src/http.rs
```diff
@@ -42,3 +40,29 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+        return Ok(client);
+    }
+
+    let certificate_path = std::env::var("GLEAM_CACERTS_PATH")
+        .map_err(|_| Error::CannotReadCertificate { path: "".into() })?;
+
+    let certificate_bytes =
+        std::fs::read(&certificate_path).map_err(|_| Error::CannotReadCertificate {
+            path: certificate_path.clone(),
+        })?;
```
lpil, line 55: Use `cli::fs::read` please 🙏

compiler-cli/src/http.rs
```diff
@@ -42,3 +40,29 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+
+    let certificate =
+        Certificate::from_pem(&certificate_bytes).map_err(|_| Error::CannotReadCertificate {
+            path: certificate_path.clone(),
+        })?;
```
lpil, line 60: Use the existing file IO error variant please, with `FileIoAction::Parse`

compiler-cli/src/http.rs
```diff
@@ -42,3 +40,29 @@ impl gleam_core::io::HttpClient for HttpClient {
...
+
+    match Client::builder().add_root_certificate(certificate).build() {
+        Ok(client) => Ok(REQWEST_CLIENT.get_or_init(|| client)),
+        Err(_) => Err(Error::CannotReadCertificate {
```
lpil, line 64: This error says the problem is with reading the certificate, but the error is actually for building the HTTP client.

If you think this error should never happen (I see the cert is checked above) then `.expect` can be used

compiler-core/src/error.rs
```diff
@@ -1398,6 +1401,17 @@ https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-fo
...
+                let text = wrap_format!("An error occurred while trying to read the certificate file at: {path}");
+
+                vec![Diagnostic {
+                    title: "Failed to read certificate".into(),
+                    text,
+                    hint: None,
+                    level: Level::Error,
+                    location: None,
+                }]
+            }
```
lpil, line 1414: Can remove these two new error variants and use the existing File IO error

lpil (commented):
Very nice! Thank you very much. I've left a couple small notes (there's existing error variants for this) and could you update the changelog please

compiler-cli/src/http.rs
winstxnhdw: There seems to be no `cli::fs::read`?

winstxnhdw: Nevermind, I am guessing you meant `fs::read`.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
   making Gleam packages discoverable through global search of HexDocs.
   ([Diemo Gebhardt](https://github.com/diemogebhardt))
 
+- Respect the `GLEAM_CACERTS_PATH` environment variable
```
lpil, line 16: Could you explain what this new feature is please 🙏 It's not clear what respect means, and it also implies there was something it was supposed to do previously that it did not

compiler-cli/src/http.rs
lpil: It's inside the CLI package, aye

compiler-cli/src/http.rs
```diff
@@ -42,3 +46,31 @@ impl gleam_core::io::HttpClient for HttpClient {
             .map_err(Error::http)
     }
 }
+
+fn init_client() -> Result<&'static Client, Error> {
+    if let Some(client) = REQWEST_CLIENT.get() {
+        return Ok(client);
+    }
+
+    let certificate_path = std::env::var("GLEAM_CACERTS_PATH").map_err(|error| Error::FileIo {
```
lpil, line 55: This is reading an environment variable, not reading a directory.

I think this is also going to error and refuse to compile code unless the `GLEAM_CACERTS_PATH` variable is set?

lpil (commented):
Thank you

winstxnhdw:
@lpil Done!

lpil (approved):
Champion! Thank you very much!!

PR #3941, opened 2024-12-02
Remove dependency on `cod:del_paths`

joshi-monster:
I still kept it in conditionally because the alternative `code:del_path` version sends a message to the `code` gen_server for every path. This restores compatibility with at least Erlang 25 again.

I want to apologize again for introducing that. ~ :purple_heart:

### DIFF
--- a/compiler-cli/templates/gleam@@compile.erl
+++ b/compiler-cli/templates/gleam@@compile.erl
@@ -150,7 +150,11 @@ add_lib_to_erlang_path(Lib) ->
     code:add_paths(filelib:wildcard([Lib, "/*/ebin"])).
 
 del_lib_from_erlang_path(Lib) ->
-    code:del_paths(filelib:wildcard([Lib, "/*/ebin"])).
+    Paths = filelib:wildcard([Lib, "/*/ebin"]),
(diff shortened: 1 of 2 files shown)


compiler-cli/templates/gleam@@compile.erl
```diff
@@ -150,7 +150,11 @@ add_lib_to_erlang_path(Lib) ->
     code:add_paths(filelib:wildcard([Lib, "/*/ebin"])).
 
 del_lib_from_erlang_path(Lib) ->
-    code:del_paths(filelib:wildcard([Lib, "/*/ebin"])).
+    Paths = filelib:wildcard([Lib, "/*/ebin"]),
+    case erlang:function_exported(code, del_paths, 1) of
```
lpil, line 154: Let's do this using conditional compilation, to avoid doing the lookup multiple times.

```erl
-if(?OTP_RELEASE > 25).
    code:del_paths(Paths).
-else.
    lists:foreach(fun code:del_path/1, Paths).
-endif.
```

lpil (commented):
Oh wonderful! Thank you!

lpil (approved):
Nice, thank you!

Issue #3914, opened 2024-11-30
I can't run gleam with the beam


OdedYanovich:
I have followed the Lustre's Quickstart guide and every time i run it (gleam run -m lustre/dev start) it returns:

  Compiling argv
escript: exception error: no case clause matching {error,
                                          {no_translation,unicode,latin1}}
  in function  gleam@@compile_erlescript1732886578465401__1155:compile_package_loop/0 (C:\project_n\build\dev\erlang\argv_gleam_artefacts\gleam@@compile.erl, line 10)
  in call from escript:run/2 (escript.erl, line 904)
  in call from escript:start/1 (escript.erl, line 418)
  in call from init:start_it/1 (init.erl, line 1546)
  in call from init:start_em/1 (init.erl, line 1521)
  in call from init:do_boot/3 (init.erl, line 1210)
error: Shell command failure

There was a problem when running the shell command escript.

I tried to run a default project and the same thing happened, i can run js projects.

My erlang version is:
erl -v
Erlang/OTP 27 [erts-15.1.2] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1] [jit:ns]

Eshell V15.1.2 (press Ctrl+G to abort, type help(). for help) 

I use windows 11

joshi-monster:
Hi! Could you please provide some more information on where and how you're running things?

- Which terminal emulator?
- Which shell?
- Inside WSL or not?
- What's the currently active code page?
- Do you have this experimental flag in Windows active where they use 65001 as the default codepage globally?

Thank you!! :purple_heart:

joshi-monster:
wait. did you copy-paste the entire error? `\project_n\build\dev\erlang\argv_gleam_artefacts\gleam@@compile.erl` there should be a backslash between `argv` and `_gleam_artefacts`

OdedYanovich:
I am running it in vscode's powershell, i don't recall ever changing the settings on windows active

joshi-monster:
What about my other questions? Do you run vscode in windows or in WSL?

I tried it in host vscode powershell on a default installation of windows without issues.

OdedYanovich:
I also installed vscode in windows in the usual way.
Just to be sure, i run the project with regular powershell and cmd, to the same results.

OdedYanovich:
The way i downloaded Rebar3 was weird, could it be the problem?

joshi-monster:
[...]

PR #3943, opened 2024-12-02
Fix filepath handling in compile script for non-ascii characters.

joshi-monster:
Fixes 2 bugs:

1. `file:read_line` does not return the correct thing when reading non-ascii characters from rust. This caused the escript to create mojibake directories.
2.  `erlang:display` does not use unicode, which makes the Rust side unable to parse its output and return an error.

~~I will try this again tomorrow on my windows machine once I'm at work. I'm also too tired I think to trust myself to not have done something stupid here.~~

update:
`file:read_line` isn't a problem on windows, since the shell uses an entirely different way to for io
`erlang:display` is not a problem in production, only if `GLEAM_LOG=trace` is enabled on all platforms.


joshi-monster:
I left it on draft because I wanted to add that test tonight :D

lpil:
Thank you!

joshi-monster:
I managed to replicate the issue in #3914, and this PR also seems to fix that one. The trace output is still kinda broken (printing the erlang charlist as a list instead of as a string), but at least it compiles successfully with and without having trace active, and doesn't create mojibake anymore.

lpil (approved):
Lovely work! Thank you

PR #3945, opened 2024-12-03
Fix nested tuple access

### LINKED ISSUE #3870: Incorrect parsing of nested tuple access
As discussed in discord, there is likely a parser bug that is causing nested tuple access to result in odd compilation errors, as well as the formatter putting the access (`0.0` in the below example) on a new line.

With the following[ gleam code](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIAZkgAT4CuADjAPpzEAUAlJQLQB8lAxK0EBJYu0wDWo/O07AAOlUmD04wQCZZigL6LFdGgCNKFSgFsAhiw69K0ygqW0GzNuwB0ABi869KbUA=): 
```gleam
fn tups_int() -> #(#(Int), #(Int)) {
  #(#(1),#(2))
}
pub fn main() -> Int {
  tups_int().0.0
}
```
The compiler is reporting this error:

```
error: Unknown record field
   ┌─ /src/main.gleam:13:13
   │
13 │   tups_int().0.0
   │             ^ This field does not exist

    #(#(Int), #(Int))

It does not have any fields.

error: Type mismatch
   ┌─ /src/main.gleam:13:14
   │
13 │   tups_int().0.0
   │              ^^^

annotation of this function.

Expected type:

    Int

Found type:

    Float
 ```

GearsDatapacks:
Fixes #3870 
This PR moves the call to `maybe_lex_dot_access` to happen after any dot token, not just after a name token.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -214,6 +214,10 @@
   not emit a warning if an @external function had been referenced.
   ([Richard Viney](https://github.com/richard-viney))
 
+- Fixed a bug where nested tuple access would not be parsed correctly when
+  the left-hand side was a function call.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.6.1 - 2024-11-19
 
 ### Bug fixed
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!

PR #3946, opened 2024-12-03
Unaligned bit arrays on the JavaScript target

richard-viney:
### Summary 📘

This PR adds support for unaligned bit arrays on the JavaScript target.

In expressions:
- Arbitrary sized integer segments: 
  ```gleam
  <<1:4>>
  <<12:9-little>>
  <<12:29-big>>
  <<1234:100-little>>
  ```
- Arbitrary sized `bits` segments:
  ```gleam
  <<<<0xABCD:15>>:bits-10>>
  ```
In patterns:
- Arbitrary sized `int` segments:
  ```gleam
  let assert <<_:7, i:19-little-signed>> = <<0xABCDEF12:26>>
  ```
- Sized and unsized `bits` segments:
  ```gleam
  let assert <<_:7, a:bits-3, b:size(14)-bits, c:bits>> = <<0xABCDEF:24, 0x1234:16>>
  ```
There is a compiler warning if the above features are used when `gleam.toml` specifies a version < v1.9.0.

---

### Implementation Details 🛠️

- The `BitArray` class in the prelude now has `bitSize`, `byteSize`, `bitOffset`, and `rawBuffer` properties.
  - The value of any unused high bits in the buffer's first byte, and any unused low bits in the buffer's final byte, are undefined.
  - Public API for use by FFI code is now: `bitSize`, `byteSize`, `bitOffset`, `rawBuffer`, `byteAt(index)`.
  - Deprecated APIs still used by existing FFI code: `get buffer()`, `get length()`. Using these emits a deprecation warning at runtime, and throws an exception if the bit array is unaligned (i.e. `bitOffset` isn't zero or `bitSize` isn't a multiple of 8).
- JSDoc annotations have been added to allow type-checking by adding `// @ts-check` to the top of the file.

---

### Implications for `@external` JavaScript code 🌍

- All existing JavaScript FFI code that operates on bit arrays needs to be updated to support unaligned bit arrays.
- Code that isn't updated will:
  - Continue to work, provided it is only used with aligned bit arrays.
  - Emit deprecation warnings at runtime on use of `BitArray.length` and `BitArray.buffer`.
  - Throw a runtime exception if these deprecated APIs are called on a unaligned bit array.
- No existing code breaks because unaligned bit arrays on JavaScript weren't previously possible.

---

### Implications for `gleam/stdlib` 🤝

- I have the updates for `gleam/stdlib` ready, mostly affecting `gleam/bit_array`. It can only be merged once this PR goes in as its tests won't run on stable Gleam. It may be necessary to run the new stdlib tests on nightly for a short period, with them segregated into their own file so they can be included/excluded depending on the active Gleam version. I'll sort that out once this PR makes it through review.
- Future stdlib versions that support unaligned bit arrays on JavaScript work fine on earlier Gleam versions, there are no compatibility concerns there.

---

### Testing 🧪

There's certainly some complexity and tricky bitwise operations here, mostly in the JavaScript prelude. The following has been done to ensure correctness:

- Many new tests added to `language_tests.gleam`, and `test/javascript_prelude`.
- Every path and branch through the code that performs slicing, concatenation, and conversion to ints/floats is covered by at least one test.
- Extensive fuzzing has been performed on these operations which validated millions of combinations of bit array contents, segment sizes, offsets, endianness, signedness, etc. on JavaScript against the result on the Erlang target.
- Issues found during fuzz testing were fixed and added to the language tests and prelude tests.


compiler-core/src/javascript/expression.rs
```diff
@@ -295,7 +296,24 @@ impl<'module> Generator<'module> {
...
+                    [Opt::Bits { .. }, Opt::Size { value: size, .. }]
+                    | [Opt::Size { value: size, .. }, Opt::Bits { .. }] => match &**size {
+                        TypedExpr::Int { value: size, .. } => {
+                            Ok(docvec![value, ".slice(0, ", size, ")"])
+                        }
+
+                        TypedExpr::Var { name, .. } => Ok(docvec![value, ".slice(0, ", name, ")"]),
+
+                        _ => Err(Error::Unsupported {
+                            feature: "This bit array segment option UNREACHABLE".into(),
```
lpil, line 313: If this is unreachable how come it is here?

lpil (commented):
Wow, what a fantastic bit of work! Thank you!

I've not digested these changes properly yet, but I have some initial questions from my first review.

There's a lot of public functions in the API of the bit array class now, but generally the aim is to have none. What is the motivation for having these?

Similar for the deprecated functions, we can remove them.

There's quite a lot of code in the class. Could we move them to free-standing functions and have the generated code only use them if absolutely necessary? That would help with JavaScript bundlers performing dead code elimination.

Some existing tests have been removed or changed, why is this? New features shouldn't alter existing tests, it makes it harder to review and to prevent regressions.

Thanks again!

compiler-core/src/javascript/expression.rs
richard-viney: Sorry, this was leftover from testing and has been reverted as it it is reachable.

richard-viney:
Thanks for taking a look!

> There's a lot of public functions in the API of the bit array class now, but generally the aim is to have none. What is the motivation for having these?

These have been reduced down to the following: `get rawBuffer()`, `get bitSize()`, `get byteSize()`, `get isWholeBytes()`, and `byteAt()`.

`byteAt()` could potentially be moved to a free function too, but was pre-existing and I've seen it use in JS FFI code, so it's still there for now and I haven't deprecated it.

> Similar for the deprecated functions, we can remove them.

I've removed all except the `BitArray.length` `and `BitArray.buffer` accessors. Removing these would break most/all JS FFI code that operates on BitArrays, so they currently emit a deprecation warning at runtime if they're used.

> There's quite a lot of code in the class. Could we move them to free-standing functions and have the generated code only use them if absolutely necessary? That would help with JavaScript bundlers performing dead code elimination.

Done.

> Some existing tests have been removed or changed, why is this? New features shouldn't alter existing tests, it makes it harder to review and to prevent regressions.

The diff on the tests looked a bit more complex in places than it actually was so I've moved some things around to make it easier to digest. Some existing tests do need to be tweaked or removed, e.g. those that were testing for compilation errors if unaligned bit arrays were used on the JS target.

richard-viney:
This PR has been updated to make slicing bit arrays an O(1) operation under all circumstances. The `BitArray.bitOffset` member was added to enable this.

compiler-core/templates/prelude.mjs
```diff
@@ -91,52 +91,253 @@ export class NonEmpty extends List {
...
-      this.buffer.buffer,
-      this.buffer.byteOffset + start,
-      end - start
+  /** @internal */
+  sliceToFloat(start, end, isBigEndian) {
+    return bitArraySliceToFloat(this, start, end, isBigEndian);
+  }
+
+  /** @internal */
+  sliceToInt(start, end, isBigEndian, isSigned) {
```
lpil, line 275: These methods mean that the large functions will never be removed by a JS bundler, isn't that right? Given their large size we want it to be that when unused there's no references to these functions so they get removed during bundling and minification so Gleam users don't need to pay a price for a feature they are not using.

lpil (commented):
Incredible work, thank you so much! Sorry for taking so long to get back to you here, I'm still playing catch-up after some family matters.

This all looks great but I have left one discussion point below RE JavaScript dead code elimination.

compiler-core/templates/prelude.mjs
richard-viney: 255e24e7b4f01be8059fd2a020e704be240f081c restructures things such that the three slicing functions are removable by esbuild. It's an unrelated change to the rest of this PR as the previous slicing functions also weren't able to be removed, but yes they've gotten larger so it matters a bit more now.

There's still an increase in JS code size due `toBitArray()` now being more complicated, the need for a `BitArray.equals()` function, the new deprecation warnings on access of `BitArray.buffer` and `BitArray.length`, etc. This code:

```gleam
pub fn main() {
  <<>>
}
```

Now bundles and minifies to 2965 bytes (1095 bytes gzip'ed) whereas previously it was 1560 bytes (676 bytes gzip'ed).

lpil: You're the best 💜

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__discard_sized.snap
```diff
@@ -11,10 +11,10 @@ fn go(x) {
 
 
 ----- COMPILED JAVASCRIPT
-import { makeError } from "../gleam.mjs";
+import { makeError, bitArraySliceToInt } from "../gleam.mjs";
```
lpil, line 14: How come this gets imported? I can't see it used here

lpil (commented):
Beautiful! Thank you so much!!! I've one question inline RE unused JS functions

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__discard_sized.snap
richard-viney: Fixed. It was happening because of the discarded integer segment.

richard-viney:
No worries, happy to help.

FYI the companion PR for stdlib is ready for a look: https://github.com/gleam-lang/stdlib/pull/761

PR #3953, opened 2024-12-04
Fix caching issue with variant inference

### LINKED ISSUE #3947: Compiler error in incremental builds even though code is valid, clean build succeeds
In my project, I get compiler errors in one specific file when building incrementally. The error messages don't make sense to me and if I `gleam clean` and then rebuild the errors are gone even though I didn't modify the code.

I admit that the code that results in this bug is kind of weird and not exactly beautiful. I refactored my project and now the errors no longer appear, still I think this probably shouldn't happen in the first place.

## Steps to reproduce

- Download and unpack the reproduction project: [compiler-bug-reporduction-project.zip](https://github.com/user-attachments/files/17997176/compiler-bug-reporduction-project.zip)
- `gleam build --target javascript`, this should work as expected
- Now add some insignificant change to the file `src/wizard.gleam`, for example add `let foo = 0` somewhere, it's not important what change it is exactly
- Now, `gleam build --target javascript` fails for me
- After `gleam clean` it works again

## Versions

- **gleam:** 1.6.2 (issue was already present back in 1.5.x)
- **erlang:** Erlang/OTP 27 [erts-15.1.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]
- **MacOS:** 14.6.1
- **rebar3:** 3.24.0 on Erlang/OTP 25 Erts 13.2.2.1
- I manage gleam, erlang and rebar3 through Nix

## Logs

These are the logs for the failing command and for the successful build after `gleam clean`.

- [log_error.txt](https://github.com/user-attachments/files/17998308/log_error.txt)
- [log_success.txt](https://github.com/user-attachments/files/17998310/log_success.txt)

### LINKED ISSUE #3949: Language server reports false errors related to variant inference
Occasionally, the language server reports an error saying that it cannot infer the variant of a value, when the compiler (`gleam check`) reports no issues.

Steps to reproduce
I haven't found an easy way to reproduce this. It has occurred multiple times, but I can't reliably reproduce it very easily.
- Clone [this repo](https://github.com/GearsDatapacks/gleam-lsp-bug): `git clone git@github.com:GearsDatapacks/gleam-lsp-bug.git`. The `build` directory is intentionally included; running `gleam clean` seems to temporarily fix the bug, and I can't figure out another way to bring it back without using this very specific `build` directory.
- Open the repo: `zed gleam-lsp-bug` or `code gleam-lsp-bug` (Not tested in other editors such as nvim)
- Navigate to the `src/squirrel/internal/database/postgres.gleam` file, and save it (Ctrl + S)

Current behaviour:
Errors are reported on lines `1313` and `1317`

Expected behaviour:
No errors are reported

**If anyone can reliably reproduce this without a specific `build` directly, please let me know!**

GearsDatapacks:
Fixes #3949 and fixes #3947 
Looks like when I implemented variant inference, I forgot to serialise the `inferred_variant` in the cache files :sweat_smile: 

This PR fixes that, but I'm not sure how best to test it. Maybe an integration test?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,9 +218,9 @@
   pipelines that would be safe to remove.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
- - Fixed a bug where assigning the prefix of a string pattern to a variable
-   nested inside another pattern would produce invalid code on Javascript.
-   ([yoshi](https://github.com/joshi-monster))
+- Fixed a bug where assigning the prefix of a string pattern to a variable
+  nested inside another pattern would produce invalid code on Javascript.
+  ([yoshi](https://github.com/joshi-monster))
 
 - Fixed a bug where expressions which use an unsafe integer on JavaScript would
   not emit a warning if an @external function had been referenced.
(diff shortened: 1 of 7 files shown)


lpil (commented):
Thank you! Could you update the changelog also please?

It's a tricky one to test. I guess we'd need to have your example from the issue, compile once, edit the file in some meaningless way, and then compile it again. I think given the complexity the unit test you've added is enough.

Issue #3955, opened 2024-12-04
Spelling mistake in mod package_name


jknight-developer:
Got an error while trying to edit a package name and spotted a spelling mistake in the error output string.
The spelling mistake is also repeated in the tests for this file.

```
// compiler-core/src/config.rs
let error =
    "Package names may only container lowercase letters, numbers, and underscores";
Err(serde::de::Error::custom(error))
```

GearsDatapacks:
This issue can be closed now as the relevant PR has been merged

lpil:
Thank you

PR #3956, opened 2024-12-04
Fixed spelling mistake in mod package_name

jknight-developer:
From issue #3955

This commit changes the phrase "package names may only container" to "package names may only contain" in all three locations.


lpil (approved):
Thank you!!

PR #3963, opened 2024-12-06
LSP: Include braces in type import completions 

Frank-III:
close: #3498

continuation of #3516, to support tests with invalid syntax. 

I am not sure if I should make it a different function. Please review it 🙏

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -256,6 +256,10 @@
   unused variables in certain cases.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where the completer would not include braces in type import completions
+  when it should have.
+  ([Jiangda Wang](https://github.com/Frank-III))
+
 ## v1.6.1 - 2024-11-19
 
 ### Bug fixed
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -72,6 +72,8 @@ fn sort_text(kind: CompletionKind, label: &str) -> String {
 enum TypeCompletionForm {
     // The type completion is for an unqualified import.
     UnqualifiedImport,
+    // The type completion is for an unqualified import within braces.
+    UnqualifiedImportWithinBraces,
     Default,
 }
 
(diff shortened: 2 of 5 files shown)


compiler-core/src/language_server/tests.rs
```diff
@@ -601,15 +607,15 @@ impl<'a> TestProject<'a> {
     pub fn at<T>(
         &self,
         position: Position,
+        new_src: Option<&str>,
```
lpil, line 610: Remove this please. The other tests should not change, we don't want to have a parameter that every test needs to provide but only this one uses.

compiler-core/src/language_server/tests.rs
```diff
@@ -559,6 +559,7 @@ impl<'a> TestProject<'a> {
     pub fn positioned_with_io(
         &self,
         position: Position,
+        new_src: Option<&str>,
```
lpil, line 562: Remove this please. The other tests should not change, we don't want to have a parameter that every test needs to provide but only this one uses.

lpil (commented):
Looks really good! I've left one note about the API change in the tests. It'll need to not change the API the others use.

I think it would be best to write the code directly for this one case rather than have all the other tests and the macro adapt for this one unique case.

Frank-III:
> Looks really good! I've left one note about the API change in the tests. It'll need to not change the API the others use.
> 
> I think it would be best to write the code directly for this one case rather than have all the other tests and the macro adapt for this one unique case.

I did that in my first attempt, but I thought we might have more tests with src being invalid syntax in the future?  If you think no, I would just write the code for this test case 


> Remove this please. The other tests should not change, we don't want to have a parameter that every test needs to provide but only this one uses.

maybe we could offer two more functions `positioned_with_io_with_new_src` and `at_with_new_src` and have `positioned_with_io` and `at` calling these two respectively so we won't need to pass this parameter in every test?

lpil:
> I did that in my first attempt, but I thought we might have more tests with src being invalid syntax in the future?

They would do the same. It must not change the other tests

> maybe we could offer two more functions positioned_with_io_with_new_src and at_with_new_src and have positioned_with_io and at calling these two respectively so we won't need to pass this parameter in every test?

I think making helper functions for a single test case is premature, we can wait and see how many of these come up.

lpil (approved):
Lovely!! Thank you!!

PR #3964, opened 2024-12-06
Improve clarity of error messages for unknown record fields.

tubedude:
Hey Louis and team.

I'm made some small changes to the error messages on Custom Types to make them clearer and more consistent.

Most importantly, inserting this one:
> The field you are trying to access is not defined consistently across all variants of this custom type. To fix this, ensure that all variants include the field with the same name, position, and type.

Instead of this one:
> The field you are trying to access exists, but not on the variant which this value always is. A field that is not present in all variants can only be accessed when the value is inferred to be one variant.

The goal was to reduce ambiguity and provide more actionable guidance. Let me know if you have any feedback!

Thanks,

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1699,7 +1699,8 @@ Hint: Add some type annotations and try again.")
                         if variants == &UnknownField::NoFields {
                             text.push_str("\nIt does not have any fields.");
                         } else {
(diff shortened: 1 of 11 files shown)


GearsDatapacks:
Ah brilliant! I'm not very good at writing error messages so I'm glad someone stepped in to help :)

lpil (approved):
Lovely! Thank you!

PR #3970, opened 2024-12-08
suggest unlocking locked pkgs that cause dep resolution failures

### LINKED ISSUE #3622: Offer to attempt again with conflicting deps unlocked when version resolution fails
If you add a new dep it can fail due to conflicts with already locked versions of other deps.

I don't know how effective this might be, but I think we could do this:

- When we have a version conflict
- and some of the packages in the conflict chain are locked
- offer to unlock those packages and try again

raquentin:
Resolves #3622

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -723,26 +723,22 @@ fn resolve_versions<Telem: Telemetry>(
     ) {
         Ok(it) => it,
         Err(
-            ref e @ Error::DependencyResolutionFailedWithLocked {
+            ref err @ Error::DependencyResolutionFailed {
                 error: _,
                 ref locked_conflicts,
             },
         ) => {
             if !is_ci_env() {
-                return Err(e.clone());
+                return Err(err.clone());
             }
 
-            let should_try_unlock = cli::confirm(
+            if cli::confirm(
                 "\nSome of these dependencies are locked to specific versions. It may
 be possible to find a solution if they are unlocked, would you like
 to unlock and try again?",
-            )?;
-
-            if should_try_unlock {
-                // unlock pkgs
+            )? {
                 unlock_packages(&mut locked, locked_conflicts, manifest)?;
 
-                // try again
                 dependency::resolve_versions(
                     PackageFetcher::boxed(runtime.clone()),
                     provided_hex_packages,
@@ -751,7 +747,7 @@ to unlock and try again?",
                     &locked,
                 )?
             } else {
-                return Err(e.clone());
+                return Err(err.clone());
             }
         }
 
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -728,7 +728,8 @@ fn resolve_versions<Telem: Telemetry>(
                 ref locked_conflicts,
             },
         ) => {
-            if !is_ci_env() {
+            // Do not ask the user to unlock conflicts in CI or if they don't exist
+            if is_ci_env() || locked_conflicts.is_empty() {
                 return Err(err.clone());
             }
 
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -720,7 +720,7 @@ to unlock and try again?",
 
             if should_try_unlock {
                 // unlock pkgs
-                unlock_packages(&mut locked, &locked_conflicts, manifest)?;
+                unlock_packages(&mut locked, locked_conflicts, manifest)?;
 
                 // try again
                 dependency::resolve_versions(
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -715,12 +715,14 @@ fn resolve_versions<Telem: Telemetry>(
             let should_try_unlock = cli::confirm(
                 "\nSome of these dependencies are locked to specific versions. It may
 be possible to find a solution if they are unlocked, would you like
-to unlock and try again? [y/n]",
+to unlock and try again?",
             )?;
 
             if should_try_unlock {
+                // unlock pkgs
                 unlock_packages(&mut locked, &locked_conflicts, manifest)?;
 
+                // try again
                 dependency::resolve_versions(
                     PackageFetcher::boxed(runtime.clone()),
                     provided_hex_packages,
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -649,6 +649,23 @@ impl PartialEq for ProvidedPackageSource {
     }
 }
 
+// Estimates whether the CLI is ran in a CI environment for use in silencing
+// certain CLI dialogues.
+fn is_ci_env() -> bool {
+    let ci_vars = [
+        "CI",
+        "TRAVIS",
+        "CIRCLECI",
+        "GITHUB_ACTIONS",
+        "GITLAB_CI",
+        "JENKINS_URL",
+        "TF_BUILD",
+        "BITBUCKET_COMMIT",
+    ];
+
+    ci_vars.iter().any(|var| std::env::var_os(*var).is_some())
+}
+
 fn resolve_versions<Telem: Telemetry>(
     runtime: tokio::runtime::Handle,
     mode: Mode,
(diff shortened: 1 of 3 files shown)


compiler-cli/src/dependencies.rs
```diff
@@ -691,18 +708,55 @@ fn resolve_versions<Telem: Telemetry>(
...
 
     // Convert provided packages into hex packages for pub-grub resolve
-    let provided_hex_packages = provided_packages
+    let provided_hex_packages: HashMap<EcoString, hexpm::Package> = provided_packages
         .iter()
         .map(|(name, package)| (name.clone(), package.to_hex_package(name)))
         .collect();
 
-    let resolved = dependency::resolve_versions(
+    let root_requirements_clone = root_requirements.clone();
```
raquentin, line 716: Looking for suggestions of better designs for this section. Cloning for use in calling the function recursively on the `should_try_unlock` page feels a bit gross

compiler-core/src/error.rs
```diff
@@ -3570,7 +3624,29 @@ The error from the version resolver library was:
...
+
+            {}
+
+            ",
+                    wrap(error)
+                );
+                vec![Diagnostic {
+                    title: "Dependency resolution with a locked package".into(),
+                    text,
+                    hint: Some("Try removing locked version(s) in your manifest.toml and re-run the command.".into()),
```
raquentin, line 3643: A bit redundant, could remove, lmk

compiler-core/src/error.rs
```diff
@@ -406,54 +416,98 @@ impl Error {
...
+                } else {
+                    Error::DependencyResolutionFailed(
+                        format!(
+                            "Unable to find compatible versions for the version constraints in your gleam.toml.\n\
+                             The conflicting packages are:\n{}",
+                            conflicting_packages.into_iter().map(|s| format!("- {s}")).join("\n")
+                        )
+                    )
+                }
+            } // end [`ResolutionError::NoSolution`] arm
```
lpil, line 455: No leftover comments please

compiler-core/src/error.rs
```diff
@@ -406,54 +416,98 @@ impl Error {
...
+                Error::DependencyResolutionFailed(msg)
             }
 
-            ResolutionError::Failure(err) => format!(
-                "An unrecoverable error happened while solving dependencies: {err}"
-            ),
-        })
+            ResolutionError::Failure(err) => {
+                let default_msg = format!("Dependency resolution was cancelled. {err}");
+                if err.contains(", but it is locked to") {
```
lpil, line 496: I don't think this logic is correct. That error message is created when a locked version falls outside the requested version constraint, but that's not the situation in which we want to offer to unlock packages. We want to unlock if there's a conflict and any of the packages in the conflict are locked to specific versions.

I don't think such logic should be in the error module as it is only concerned with the definition, construction, and displaying of errors. It doesn't know anything about the wider context of the program or why errors would be emitted.

lpil (commented):
Looking good! Thank you!

I think the logic for deciding when to offer isn't quite right at the moment, I've left comment inline.

Could you also remove the new error variants that have been added please- we don't want to change anything about the errors shown in this case.

raquentin:
> Looking good! Thank you!
>
> I think the logic for deciding when to offer isn't quite right at the moment, I've left comment inline.
> 
> Could you also remove the new error variants that have been added please- we don't want to change anything about the errors shown in this case.

For sure, ty for review. And yeah I'll revisit the error logic and remove the `WithLocked` variant

raquentin:
On

> Could you also remove the new error variants that have been added please- we don't want to change anything about the errors shown in this case.

In this case how would you provide `locked_conflicts` to `dependencies::resolve_versions()`? Ty

raquentin:
Need anything else from me on this?

compiler-core/src/dependency.rs
```diff
@@ -789,14 +790,72 @@ mod tests {
...
+            error,
             "An unrecoverable error happened while solving dependencies: gleam_stdlib is specified with the requirement `~> 0.1.0`, but it is locked to 0.2.0, which is incompatible."
         ),
         _ => panic!("wrong error: {err}"),
         }
     }
 
+    #[test]
+    fn resolution_locked_version_doesnt_satisfy_requirements_locked() {
+        // we're creating a dependency logging v1.4.0 that requires gleam_stdlib v0.40.0
```
lpil, line 803:
```suggestion
        // We're creating a dependency logging v1.4.0 that requires gleam_stdlib v0.40.0
```

compiler-core/src/dependency.rs
```diff
@@ -789,14 +790,72 @@ mod tests {
         .unwrap_err();
 
         match err {
-        Error::DependencyResolutionFailed(msg) => assert_eq!(
-            msg,
+        Error::DependencyResolutionFailed{error, locked_conflicts: _} => assert_eq!(
```
lpil, line 793: Looks like you need to run the formatter here!

compiler-core/src/dependency.rs
```diff
@@ -789,14 +790,72 @@ mod tests {
...
+                    version: Version::new(1, 4, 0),
+                    requirements: requirements,
+                    retirement_status: None,
+                    outer_checksum: vec![0],
+                    meta: (),
+                }],
+            },
+        );
+
+        // now try and resolve versions with gleam_stdlib v0.20.0 in lock.
```
lpil, line 830:
```suggestion
        // Now try and resolve versions with gleam_stdlib v0.20.0 in lock.
```

compiler-core/src/dependency.rs
```diff
@@ -789,14 +790,72 @@ mod tests {
...
+            provided_packages,
+            "root_name".into(),
+            vec![("logging".into(), Range::new(">= 1.3.0 and < 2.0.0".into()))].into_iter(),
+            &vec![("gleam_stdlib".into(), Version::new(0, 20, 0))]
+                .into_iter()
+                .collect(),
+        )
+        .unwrap_err();
+
+        // expect failure
```
lpil, line 842:
```suggestion
        // Expect failure
```

compiler-core/src/error.rs
```diff
@@ -406,54 +412,98 @@ impl Error {
...
 
                 let mut conflicting_packages = HashSet::new();
                 collect_conflicting_packages(&derivation_tree, &mut conflicting_packages);
 
-                wrap_format!("Unable to find compatible versions for \
-the version constraints in your gleam.toml. \
-The conflicting packages are:
+                let conflict_names: Vec<EcoString> = conflicting_packages
+                    .iter()
+                    .map(|pkg| (*pkg).to_string().into())
```
lpil, line 424: str -> ecostring rather than str -> string -> ecostring please

compiler-core/src/error.rs
```diff
@@ -406,54 +412,98 @@ impl Error {
...
+                                conflicting_packages.into_iter().map(|s| format!("- {s}")).join("\n")
+                            ),
+                            locked_conflicts,
+                    }
+                } else {
+                    Error::DependencyResolutionFailed {
+                            error: format!(
+                                "Unable to find compatible versions for the version constraints in your gleam.toml.\n\
+                                 The conflicting packages are:\n{}",
+                                locked_conflicts.iter().map(|s| format!("- {s}")).join("\n")
```
lpil, line 447: Why are only some of the conflicting packages are listed in this case? Wouldn't we want to show them all? If not we need to update the message to say this is only a subset of them.

lpil (commented):
Thank you! I've left a few small comments inline, otherwise nearly ready to go. Would you mind updating the changelog also please 🙏

raquentin:
Thanks for review, been a bit busy but I believe I've resolved them all. Let me know if there's something else you'd like

lpil:
Heya! Would you mind rebasing on main? 🙏  Alternatively I can squash into one commit and rebase. Thank you

lpil:
Going to close this as there has been no recent activity, but please feel free to open a new one! Thank you.

lpil:
Sorry, I shouldn't have closed this. We will try and get this rebased and resolve the conflicts.

raquentin:
All good, my bad for being afk for a few months. I can get on some night this week and try and get this mergable. Thanks for keeping up with it

lpil:
Closing due to inactivity. Feel free to reopen!

PR #3974, opened 2024-12-10
Fix typo in error message for 'use'

diemogebhardt:
Stumbled across this error message and `the the` seems to be a typo?

```rust
                        text.push_str("All the arguments have already been supplied, \
so it cannot take the the `use` callback function as a final argument.\n")
```

Feel free to just close this PR, in case this isn't a typo.


lpil (approved):
Thank you!

lpil (approved):
Thank you

PR #3981, opened 2024-12-13
Add better help texts for Error::ShellProgramNotFound

enkerewpo:
Hi,
   I added some installation helper texts for `bun`, `deno`, `elixir` and split the matching logic into different OS.
Best regards


enkerewpo:
removed tailing spaces that caused ci rustfmt check failed

PR #4081, opened 2024-12-14
Improve unused variable warnings

### LINKED ISSUE #3425: Make unused variable warning more precise
Right now when a punned variable is unused you always get the same hint on how to ignore it (which is prepend a `_` to the var name). This is not really precise for punned variables:
```gleam
let Wibble(arg:, arg1:) = todo
//               ^^^^^ Warning: unused, you can ignore it with an underscore `_arg1`
```
If one were to blindly apply the compiler's suggestion here they would get a compilation error:
```gleam
let Wibble(arg:, _arg1) = todo
//               ^^^^^ Error: unlabelled arg after labelled one!

// Or even:
let Wibble(arg:, _arg1:) = todo
//                    ^ Error: I was not expecting this!
```
Which could be confusing for a newcomer. Once we improve unused detection using a call graph <https://github.com/gleam-lang/gleam/issues/2634> we could be more precise and change the suggestion for variables brought into scope through punning and provide a valid suggestion like this:
```gleam
let Wibble(arg:, arg1:) = todo
//               ^^^^^ Warning: unused, you can ignore it with an underscore `arg1: _`
```

GearsDatapacks:
Fixes #3425
This PR adds extra context to unused variable warnings, allowing them to correctly give hints about how to ignore those variables.


compiler-core/src/type_/error.rs
```diff
@@ -646,6 +646,34 @@ impl Named {
...
+    pub fn how_to_ignore(&self) -> String {
+        match self {
+            VariableOrigin::Variable(name) => {
+                format!("You can ignore it with an underscore: `_{name}`.")
+            }
+            VariableOrigin::LabelShorthand(label) => {
+                format!("You can ignore it with an underscore: `{label}: _`.")
+            }
+            VariableOrigin::AssignmentPattern => "You can safely remove it.".to_string(),
+            VariableOrigin::Generated => String::new(),
```
lpil, line 672: Never use an empty string to mean `None`, this should instead return an option

lpil (commented):
Thank you! Looks great. I've left one tiny note inline

lpil (approved):
Lovely work, thank you!!

Issue #4084, opened 2024-12-16
Format lists the same in constants as in expressions


lpil:
### Discussed in https://github.com/gleam-lang/gleam/discussions/3976

<div type='discussions-op-text'>

<sup>Originally posted by **FrostyX** December 11, 2024</sup>
I have this `let` variable:

```gleam
let map = [
  #("A", [Dot, Comma]),
  #("B", [Comma, Dot, Dot, Dot]),
  #("C", [Comma, Dot, Comma, Dot]),
  #("D", [Comma, Dot, Dot]),
  #("E", [Dot]),
  #("F", [Dot, Dot, Comma, Dot]),
  #("S", [Dot, Dot, Dot]),
]
```

The formatting is very readable. However, when I turn it into a constant, `gleam format` does this:

```gleam
const map = [
  #("A", [Dot, Comma]), #("B", [Comma, Dot, Dot, Dot]),
  #("C", [Comma, Dot, Comma, Dot]), #("D", [Comma, Dot, Dot]), #("E", [Dot]),
  #("F", [Dot, Dot, Comma, Dot]), #("S", [Dot, Dot, Dot]),
]
```

Can we change `gleam format` to do the same exact formatting for `const` as it does for `let`?</div>

PR #3081, opened 2024-05-02
Fix constant tuple list formatting

isaacharrisholt:
Fixes #3070

PR #4086, opened 2024-12-16
Formatter:  format list the same for const as it does for let

Frank-III:
close: #4084

it seems that we want to fit a list of tuples in `const` when it is possible before: #3081


lpil (approved):
Thank you

PR #4094, opened 2024-12-18
Improve dependency conflict error messages

dinkelspiel:
Improved the error message for dependency resolution failures to more clearly 
explain why the conflict occurs. Previously, the error message only listed the 
names of the conflicting packages:

```
- mist
- server
- cors_builder
```

Now, the message provides more detailed descriptions:

```
- Package `app` requires `cors_builder` 1.0.0 <= v < 2.0.0
- Conflict with the `mist` package:
    - `cors_builder` requires `mist` version 1.0.0 <= v < 2.0.0
    - `app` requires `mist` version 3.0.0 <= v < 4.0.0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,27 @@
   with a local password, reducing risk of your Hex password being compromised.
   ([Louis Pilfold](https://github.com/lpil))
 
+- Improved the error message for dependency resolution failures to more clearly 
+  explain why the conflict occurs. Previously, the error message only listed the 
+  names of the conflicting packages:
+
+  ```
+  - mist
+  - server
+  - cors_builder
+  ```
+
+  Now, the message provides more detailed descriptions:
+
+  ```
+  - Package `app` requires `cors_builder` 1.0.0 <= v < 2.0.0
+  - Conflict with the `mist` package:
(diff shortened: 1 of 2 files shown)


giacomocavalieri:
Hello! This is being worked in https://github.com/gleam-lang/gleam/pull/3640
Unfortunately it's blocked by an external issue with a dependency that doesn't allow us to display the version constraint nicely: https://github.com/pubgrub-rs/pubgrub/issues/258#issuecomment-2504675738

The default `to_string` implementation was deemed confusing and that's why it was removed from the error message.

dinkelspiel:
Of course i'm biased, but i'd still argue that it's good to merge this in the meantime. I had this issue today and it tells you next to nothing, I think it's better to merge this so it's there and gives information rather than having to wait for an external dependency to merge something so we can mitigate some poor soul having to search through each dependencies gleam.toml to find the issue.

giacomocavalieri:
I agree with you, but I've already had this talk with Louis and he won't merge this; otherwise https://github.com/gleam-lang/gleam/pull/3640 would have been merged already. There's nothing much we can do besides waiting 😅

lpil:
Please open or participate in an issue before implementing any changes. It is unlikely that any unplanned work will be accepted as the odds of it having the intended design is very low and it's not fun to have your work rejected after you spent the time on it. Thank you

PR #4100, opened 2024-12-21
Fix bug with extract variable top level statements

giacomocavalieri:
Experimenting with the nightly compiler I noticed a couple of bugs with the "extract variable" code action I've recently added:
- It would allow extracting top level expressions, not all that useful:
  ```gleam
  pub fn main() {
    1
  //^ [Extract variable]
  }

  // turned into:

  pub fn main() {
    let value = 1
    value
  }
  ```
- It would extract module selects on their own instead of considering those as part of an outer function call, resulting in an error:
  ```gleam
  pub fn main() {
    list.map(list.filter([1, 2, 3], todo), todo)
  //          ^ [Extract variable]
  }

  // turned into:

  pub fn main() {
    let value = list.
    list.map(valuefilter([1, 2, 3], todo), todo)
  }
  ```
  
This PR fixes both

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -2605,10 +2605,16 @@ pub struct ExtractVariable<'a> {
     module: &'a Module,
     params: &'a CodeActionParams,
     edits: TextEdits<'a>,
-    inside_capture_body: bool,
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you!!!

PR #4102, opened 2024-12-21
Expand function capture code action

### LINKED ISSUE #3938: Expand function capture code action
An action that converts a function capture syntax sugar into the anonymous function syntax
```gleam
pub fn main() {
  list.map(_, run)
}
```

```gleam
pub fn main() {
  fn(value) { list.map(value, run) }
}
```

giacomocavalieri:
This PR closes #3938


lpil (approved):
Lovely!!!! Thank you

PR #4103, opened 2024-12-22
Tweak code actions

GearsDatapacks:
This PR tweaks a couple of the code actions to use the `CodeActionKind` more representative of what the code action actually does. I've also ensured that all the quickfix code actions, which fix errors and warnings are marked as preferred, and any of the optional rewriting actions marked a not preferred.

This PR is a little subjective, so please close it if you don't want me to make this change

(Also should I update the changelog for a tiny change like this?)


lpil (approved):
Thanks!!

PR #4106, opened 2024-12-23
LSP Code Action: Generate dynamic decoder

### LINKED ISSUE #2955: LSP code action: generate dynamic decoder
We need to decide what the recommendation is for constructors with too many fields for the decodeN functions. Perhaps we can not support them for this first version.

We also need to decide what to do for contained types we don't know the decoder for. Perhaps we can add a `todo as "..."` for now.

GearsDatapacks:
Closes #2955 
This PR implements a code action to generate a dynamic decoder for a custom type.

### DIFF
--- a/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__generate_dynamic_decoder_tuple.snap
+++ b/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__generate_dynamic_decoder_tuple.snap
@@ -1,0 +1,35 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\npub type Wibble {\n  Wibble(tuple: #(Int, Float, #(String, Bool)))\n}\n"
+---
+----- BEFORE ACTION
+
+pub type Wibble {
+    ↑            
+  Wibble(tuple: #(Int, Float, #(String, Bool)))
+}
+
+
+----- AFTER ACTION
+import gleam/dynamic/decode
+
+pub type Wibble {
+  Wibble(tuple: #(Int, Float, #(String, Bool)))
+}
+
+fn wibble_decoder() -> decode.Decoder(Wibble) {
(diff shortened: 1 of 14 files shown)


GearsDatapacks:
Oh, do we want to add support for decoding tuples as well? If so, what would that decoder look like (I've only done basic decoding thus far)

lpil:
Yes please! Something like this:

```gleam
use hostname <- decode.field("hostname", decode.string)
use ip <- decode.field("ip", {
  use a <- decode.field(0, decode.int)
  use b <- decode.field(1, decode.int)
  use c <- decode.field(2, decode.int)
  use d <- decode.field(3, decode.int)
  decode.success(#(a, b, c, d))  
})
decode.success(Server(hostname:, ip:))
```

GearsDatapacks:
Ah I see, you can use `field` with more than just strings. Thanks!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__generate_dynamic_decoder_tuple.snap
```diff
@@ -0,0 +1,35 @@
...
+    use c <- decode.field(2, {
+      use a <- decode.field(0, decode.string)
+      use b <- decode.field(1, decode.bool)
+
+      decode.success(#(a, b))
+    })
+
+    decode.success(#(a, b, c))
+  })
+
```
lpil, line 33: Could we remove these empty lines please 🙏

lpil (approved):
Super cool! Neat how little code it is too. Very clean! Thank you!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__generate_dynamic_decoder_tuple.snap
lpil: I'll do it 👍

PR #4110, opened 2024-12-26
set REBAR_SKIP_PROJECT_PLUGINS when running rebar3

tsloughter:
This env var was added for tools like `gleam` to tell rebar3 they are using it just to build a dep and want to skip any project_plugins that might be in the configuration. Those were added specifically to not be pulled in when building a project as a dep.

Depends on https://github.com/erlang/rebar3/pull/2932 to actually have the benefit of faster builds.


lpil (approved):
Thank you!

PR #4114, opened 2024-12-27
Update default Elixir version in test.yml

### LINKED ISSUE #4113: Incompatible Erlang OTP and Elixir versions in the generated test workflow
Hello! Running `gleam new` and then uncommenting the Elixir line in test.yml produces a github workflow that will not run because Erlang OTP v27 is not supported by Elixir v1.15. Is there any reason we should not update the Elixir version for the test workflow to the most recent version, v1.18.1? I can confirm that Elixir 1.18.1 works with the github test runner.

jrstrunk:
Closes #4113, the prior Elixir version was not compatible with the Erlang OTP version provided by default, this PR fixes that. I tested the workflow with an Elixir version of `1` and it succeeded; hopefully this will prevent conflicts in the future as the Erlang OTP version is updated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -291,6 +291,10 @@
   when it should have.
   ([Jiangda Wang](https://github.com/Frank-III))
 
+- Fixed a bug where `gleam new` would generate the github `test` workflow
+  configuration with incompatible Erlang OTP and Elixir versions.
+  ([John Strunk](https://github.com/jrstrunk))
+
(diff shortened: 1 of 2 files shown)


lpil (commented):
Thank you. Could you update the changelog also please

jrstrunk:
Right! I didn't realize syncing with upstream would create a merge commit instead of rebasing, darn. I'll try to fix that.

lpil (approved):
Lovely, thank you

PR #4117, opened 2024-12-30
Fix trailing :bytes pattern on JS giving incorrect result

richard-viney:
Found this regression when testing v1.7.0-rc1. It's a one line fix and new tests have been added.


lpil (approved):
Thank you!

PR #4118, opened 2024-12-30
floating point values outside Erlang's floating point range should cause error

### LINKED ISSUE #4116: Illegal float literals crash when targeting Erlang
The following program crashes with a non-Gleam error when targeting Erlang.

```gleam
pub fn main() {
  let _ = 1.8e308
}
```

```
$ gleam run
  Compiling float_stuff
/home/ryan/projects/gleam/float_stuff/src/float_stuff.gleam:4:9: illegal float
/home/ryan/projects/gleam/float_stuff/src/float_stuff.gleam:4:16: syntax error before: '.'
/home/ryan/projects/gleam/float_stuff/build/dev/erlang/float_stuff/_gleam_artefacts/float_stuff.erl:4:2: function main/0 undefined
%    4| -export([main/0]).
%     |  ^

/home/ryan/projects/gleam/float_stuff/src/float_stuff.gleam:2:2: spec for undefined function main/0
%    2|   let _ = 1.8e308
%     |  ^

error: Shell command failure

```

Here is the result of running `GLEAM_LOG=trace gleam run`:
[TRACE.txt](https://github.com/user-attachments/files/18270388/TRACE.txt)

(gleam 1.6.3)

massivefermion:
closes #4116

### DIFF
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -582,6 +582,10 @@ pub enum Error {
     DeprecatedVariantOnDeprecatedType {
         location: SrcSpan,
     },
+
+    ErlangFloatUnsafe {
+        location: SrcSpan,
+    },
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -1040,7 +1044,8 @@ impl Error {
             | Error::UseFnIncorrectArity { location, .. }
             | Error::BadName { location, .. }
             | Error::AllVariantsDeprecated { location }
-            | Error::DeprecatedVariantOnDeprecatedType { location } => location.start,
+            | Error::DeprecatedVariantOnDeprecatedType { location }
+            | Error::ErlangFloatUnsafe { location } => location.start,
             Error::UnknownLabels { unknown, .. } => {
                 unknown.iter().map(|(_, s)| s.start).min().unwrap_or(0)
             }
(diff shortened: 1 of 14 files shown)


compiler-core/src/type_/error.rs
```diff
@@ -1681,3 +1686,25 @@ pub fn check_javascript_int_safety(int_value: &BigInt, location: SrcSpan, proble
         problems.warning(Warning::JavaScriptIntUnsafe { location });
     }
 }
+
+/// When targeting Erlang, adds an error if the given Float value is outside the range
+/// -1.7976931348623157e308 to 1.7976931348623157e308 which is the allowed range for
+/// Erlang's floating point numbers
```
lpil, line 1692: Could we link to there these numbers were taken from in the Erlang docs please 🙏

compiler-core/src/type_/tests/errors.rs
```diff
@@ -2655,3 +2655,18 @@ todo
...
+
+#[test]
+fn out_of_range_erlang_float_in_pattern() {
+    assert_error!(r#"let assert [1.8e308, b] = [x, y]"#);
+}
+
+#[test]
+fn out_of_range_erlang_float_in_const() {
+    assert_module_error!(r#"const x = 1.8e308"#);
+}
```
lpil, line 2672: Can we have tests for the negative case too please 🙏

lpil (commented):
Awesome work! Thank you. I've left some small notes inline

compiler-core/src/type_/error.rs
massivefermion: I actually can't find them anywhere in the Erlang docs. I got them from [here](https://github.com/elixir-lang/elixir/blob/v1.18.1/lib/elixir/lib/float.ex).

lpil (commented):
Could you rebase on main please? The changelog has a conflict 🙏

lpil (approved):
Thank you!!

PR #4124, opened 2024-12-31
Format function captures like regular function calls

### LINKED ISSUE #4121: Formatter: function captures in `list.fold` mess up with formatter
Hi!
I'm not 100% sure it's an issue, but I think it's good to discuss about it.
Currently, a `list.fold` without function capture is properly formatted, like this:

```gleam
// Formatting with no function capture
pub fn example() {
  list.fold(my_list, t, fn(a) {
    io.print("Not meh")
    io.print("Not meh")
  })
}
```

Adding a function capture (`_`) seems to mess up with the formatter:

```gleam
// Formatting with function capture
pub fn example() {
  list.fold(
    my_list,
    _,
    fn(a) {
      io.print("Meh")
      io.print("Meh")
    },
  )
}
```

I'm not sure it's a desired behaviour, because using a function capture in `list.map` does not changes the formatting:

```gleam
// Formatting with function capture
pub fn example() {
  list.map(_, fn(a) {
    io.println("example")
    io.println("exmaple")
  })
}
```

I think for consistency, everything should be formatted on one line while it's possible.

giacomocavalieri:
This PR fixes #4121, getting back to the formatter after a long while was fun! 😆

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## Unreleased
+
+### Formatter
+
+- Function captures are now formatted like regular function calls.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ## 1.7.0-rc2 - 2024-12-30
 
 ### Compiler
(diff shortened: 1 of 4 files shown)


lpil:
image
😆

lpil (approved):
Thank you

PR #4132, opened 2025-01-01
Require newer stdlib version

giacomocavalieri:
I noticed the projects created with the RC require stdlib to be `>= 0.44`, but since we're now at `0.51` I though it would be better if we encouraged that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,7 +115,7 @@
   consumers.
   ([Aleksei Gurianov](https://github.com/guria))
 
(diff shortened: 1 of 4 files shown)


lpil:
The build tool picks the newest compatible version so changing this would be for if we want to prevent specific older versions being used. I'm not totally sure if there a need for this right now- did you have a particular motivation?

giacomocavalieri:
Ah you're right I didn't put much thought into this, sorry for the noise!

PR #4133, opened 2025-01-01
Format record updates like function calls

### LINKED ISSUE #4120: Formatter: inconsistent formatting between record update and record creation
Hi! 
Currently, the formatter formats a record update like this:

```gleam
pub fn example() {
  Record(
    ..record,
    field: {
      use _ <- list.map(record.field)
      io.print("Example")
    },
  )
}
```

While, to follow the same pattern as record creation or even function call, I think we should expect formatting like this:

```gleam
pub fn example() {
  Record(..record, field: {
    use _ <- list.map(record.field)
    io.print("Example")
  })
}
```

Indeed, record creation get formatted as the second example.

giacomocavalieri:
This PR fixes #4120


lpil:
Oh bums I thought I merged this.

PR #4137, opened 2025-01-02
Allow to fill in labels in incomplete function calls

### LINKED ISSUE #3885: Support "Fill labels" code action if there are already arguments provided
The language server supports a "add labels" code action, but only if there are no explicit arguments provided:

before:

```gleam
fn wibble(a a: Int, b b: Float, c c: String) {
  a + b + c
}

fn wobble() {
  wibble()
  //    ^
}
```

after:

```gleam
fn wobble() {
  wibble(a: todo, b: todo, c: todo)
}
```

It would be great if this code action could also support cases where I already partially supply arguments to a function or record constructor. This would be really useful if I added a new argument(s) or field(s), and I want to now update all the call-sides with the new arguments. This code action could add the missing labels for me, so I would have to just fill out the values. ~ :purple_heart:

giacomocavalieri:
This PR fixes #3885

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,29 @@
 
 ### Language server
 
+- The language server can now fill in the labels of any function call, even when
+  only some of the arguments are provided. For example:
+
+  ```gleam
+  import gleam/string
+
+  pub fn main() {
+    string.replace("wibble")
+  }
+  ```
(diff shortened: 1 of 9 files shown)


lpil (approved):
Lovely, thank you!!

Issue #3968, opened 2024-12-08
Don't use self-invoking anonymous functions when generating blocks on JavaScript


GearsDatapacks:
Currently, this gleam code:
```gleam
pub fn main() {
  let a = {
    let b= 1
    b + 1
  }
}
```
Compiles to this javascript:
```js
export function main() {
  let a = (() => {
    let b = 1;
    return b + 1;
  })();
  return a;
}
```

This isn't ideal as it creates an anonymous function which is immediately invoked, which makes the code harder to read and adds the overhead of allocating a function, calling it, then deallocating it. It would be nice to not have to do this.

Perhaps Louis has a particular alternative in mind, but we could generate something like:
```js
let _block;
{
  let b = 1;
  _block = b + 1;
}
a = _block;
return a;
```
(Or in this case assign directly to `a` instead of `_block`, but I'm not sure the compiler would be able to do that)

lpil:
Yes!!

PR #4142, opened 2025-01-04
Stop using iifes for blocks in tail position

giacomocavalieri:
Currently the compiler wraps all blocks in iifes, but some could be dropped, for example if a block is in tail position there's no need to wrap that in an iife and we could just return its last expression.

This is only part of #3968 and doesn't addresses it completely, I've tried implementing this kind of optimisation for assignments as well but it's quite tricky to get right and couldn't get it to work properly. I'll try that again, but for the moment this is a little improvement!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Compiler
 
+- Improved code generation for blocks in tail position on the Javascript target.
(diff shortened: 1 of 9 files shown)


GearsDatapacks:
Yeah, I also looked into getting this to work with assignments, and I couldn't find a straightforward way to do it either. It might require a bit of a rework in code generation

lpil (approved):
Thank you

PR #4153, opened 2025-01-07
Add "Pattern match on argument" code action

### LINKED ISSUE #4139: Code action to destructure a function argument
It would be really nice to have an action to destructure a function argument if it is a tuple/type with a single constructor:

```gleam
pub fn wibble(a|rg: #(Int, a)) {
  //           ^ Cursor here [destructure argument]
}

// Adds a line at the top of the function like this one:
pub fn wibble(arg: #(Int, a)) {
  let #(value_1, value_2) = arg
}
```
Another example:
```gleam
pub type Wibble {
  Wibble(Int, label: String)
}

pub fn wibble(ar|g: Wibble) {
  //            ^ Cursor here [destructure argument]
}

// Adds a line art the top of the function like this one:
pub fn wibble(arg: Wibble) {
  let Wibble(value_1, label:) = arg
}
```

A couple of notes:
- For tuples/unlabelled arguments it could use simple names like `value_1`, `value_2`, ...
- For labelled arguments it would use the label shorthand syntax

giacomocavalieri:
This PR closes #4139.

The tests cover loads of different cases but here's a small example of what it looks like in practice:


compiler-core/src/parse.rs
```diff
@@ -1882,23 +1882,28 @@ where
...
                     None => vec1![Statement::Expression(UntypedExpr::Todo {
-                        kind: TodoKind::EmptyFunction,
-                        location: SrcSpan { start, end },
+                        kind: TodoKind::EmptyFunction {
+                            function_location: SrcSpan { start, end }
+                        },
+                        location: SrcSpan {
+                            start: left_brace_start + 1,
+                            end: right_brace_end
+                        },
```
giacomocavalieri, line 1900: Previously when the compiler inserted an implicit `todo` in an empty function body it marked its position as starting at the start of the function declaration. This would mess up code actions that rely on a function's body's statement to always come inside the curly braces.

I found this out implementing the "Destructure argument" code action and changed it here so that an implicit todo now has the following span:
```gleam
   pub fn main() { }
//                ^^ This span inside the function's body
// ^^^^^^^^^^^^^ Instead of this span outside the function's body
```
It still carries around the entire function span to produce nice error messages for incomplete code in the `TodoKind::EmptyFunction` variant

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__destructure_argument_empty_tuple.snap
```diff
@@ -0,0 +1,18 @@
...
+pub fn main(tuple: #()) {
+            ↑            
+  todo
+}
+
+
+----- AFTER ACTION
+
+pub fn main(tuple: #()) {
+  let #() = tuple
```
lpil, line 16: I think we shouldn't offer the action when there's no values to match on

compiler-core/src/language_server/tests/action.rs
```diff
@@ -71,6 +71,7 @@ const CONVERT_TO_USE: &str = "Convert to `use`";
 const EXTRACT_VARIABLE: &str = "Extract variable";
 const EXPAND_FUNCTION_CAPTURE: &str = "Expand function capture";
 const GENERATE_DYNAMIC_DECODER: &str = "Generate dynamic decoder";
+const DESTRUCTURE_ARGUMENT: &str = "Destructure argument";
```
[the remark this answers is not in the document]
giacomocavalieri: "Pattern match on argument"? Though one might think it could work for types with any number of variants. To be honest we could also extend this in the future to insert a case expression when there's more than a single variant

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__destructure_argument_empty_tuple.snap
giacomocavalieri: Ok, fixed!

compiler-core/src/language_server/tests/action.rs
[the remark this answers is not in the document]
lpil: I was just about to suggest the very same thing!

giacomocavalieri:
I've updated this to:
- use the term "pattern match" instead of "destructure"
- work on types with more than a single constructor as well, an example below:

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__pattern_match_on_argument_works_on_fn_arguments.snap
```diff
@@ -0,0 +1,25 @@
...
+fn map(list: List(a), fun: fn(a) -> b) { todo }
+
+
+----- AFTER ACTION
+
+pub fn main() {
+  [#(1, 2)]
+  |> map(fn(tuple) {
+  let #(value_0, value_1) = tuple
+})
```
giacomocavalieri, line 22: In this case of anonymous function with an empty body inside a multiline pipe I couldn't figure out a way to print it nicely with the correct indentation that is not too messy. Do you feel like this is ok, given the formatter will take care of it anyway?

compiler-core/src/language_server/code_action.rs
```diff
@@ -3247,3 +3252,487 @@ impl<'a> DecoderPrinter<'a> {
...
+/// }
+///
+/// // Adds this to the top of the function:
+///
+/// pub fn wibble(arg: Wibble) {
+///   let Wibble(label1:, label2:) = arg
+/// }
+/// ```
+///
+pub struct PatternMatchOnArgument<'a, A> {
```
lpil, line 3270: I think the name and documentation for this code action may be out of date now. It works for both function arguments and assignments?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__pattern_match_on_argument_works_on_fn_arguments.snap
lpil: Yeah it's fine

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Oops! Fixed it

lpil: The name is still out of date and the documentation is now missing!

giacomocavalieri: I really messed it didn't I 😆 It should be ok for real this time

lpil (approved):
Thank you!!

PR #4155, opened 2025-01-08
Fix invalid erlang for record updates inside blocks

### LINKED ISSUE #4154: Incorrect Erlang code with block pattern during record update
Hi !

I found a funny bug when experimenting with Gleam 1.7.0.
The following code:

```gleam
pub type A {
  A(b: B)
}

pub type B {
  B(c: C)
}

pub type C {
  C(val: Int)
}

pub fn example(a: A) {
  A(..a, b: { 
    B(..a.b, c: { 
      C(..a.b.c, val: 0) 
    }) 
  })
}
```

Compiles incorrectly to:

```erlang
example(A) ->
    _record = A,
    {a, (_record@1 = erlang:element(2, A),
        {b, (_record@2 = erlang:element(2, erlang:element(2, A)),
            {c, 0})})}.
```

Which is invalid, the `(_record, {` pattern is not working, but when removing the block pattern, like this:

```gleam
pub fn example(a: A) {
  A(..a, b: 
    B(..a.b, c:  
      C(..a.b.c, val: 0) 
    ) 
  )
}
```

Compiles correctly to:

```erlang
example(A) ->
    _record = A,
    {a,
        begin
            _record@1 = erlang:element(2, A),
            {b,
                begin
                    _record@2 = erlang:element(2, erlang:element(2, A)),
                    {c, 0}
                end}
        end}.
```

It triggers a hard to understand build error looking like (because of a `,` error)

```
project/src/module.gleam: spec for undefined function example/1
%   32|   A(..a, b: B(..a.b, c: { C(..a.b.c, val: 0) }))
%     |  ^
```

---

Edit: I made sure it's working with JS: everything gets compiled as IIFE, so there's no problem.

joshi-monster:
fix #4154

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,3 +50,8 @@
 - Fixed a bug where build tool could fail to add new dependencies when
   dependencies with optional dependencies are present in the manifest.
   ([Louis Pilfold](https://github.com/lpil))
+
+- Fixed a bug where a block expression containing a singular record update would
+  produce invalid erlang.
+  ([yoshi](https://github.com/joshi-monster))
+  
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -2377,13 +2377,6 @@ impl TypedStatement {
         }
     }
 
-    pub fn is_non_pipe_expression(&self) -> bool {
-        match self {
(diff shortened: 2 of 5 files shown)


lpil (approved):
Thank you

PR #4162, opened 2025-01-09
consistently call docvec! with `[]` instead of `()`

giacomocavalieri:
I noticed that the codebase used both `docvec!()` and `docvec![]`. Since `docvec!()` was the minority I replaced those with `[]` to have a consistent style across the codebase.
I also replaced all calls to `docvec![]` with `nil()` since that seems to be the most widespread convention across the codebase!


lpil (approved):
Thank you!

PR #4165, opened 2025-01-09
Fix clippy warnings for Rust 1.84

GearsDatapacks:


GearsDatapacks:
Hmm that's odd. Wasm tests pass on my machine.

GearsDatapacks:
Ok, increasing node version from 18 to 20 seemed to do the trick

lpil (approved):
Thank you!

Issue #4141, opened 2025-01-02
hex packages should not leak absolute path in .erl -file "directives"


inoas:
Currently erlang files export the full path as it was available on the developer machine that ran gleam publish.

## Current implementation

```erlang
-file("/Users/USERNAME/SOME/PATH/some_lib_or_app/src/some_lib_or_app.gleam", 26).
-spec my_fun(boolean(), fun(() -> FVC), fun(() -> FVC)) -> FVC.
my_fun(Var1, Var2, Var3) ->
% ...
```

## Expected implementation

```erlang
-file("some_lib_or_app/src/some_lib_or_app.gleam", 26).
-spec my_fun(boolean(), fun(() -> FVC), fun(() -> FVC)) -> FVC.
my_fun(Var1, Var2, Var3) ->
% ...
```

lpil:
I agree, thank you.

PR #4166, opened 2025-01-09
Use relative paths in -file attributes 

oneness:
Hi @lpil,
Please have a look. It seems the entire pipeline (load, compile and gen) does uses absolute path, which IMHO is a great thing for reproducibility and tooling. This PR  propagates the project root so it is stripped off from the generated module file directive.  I might be completely off and feel free to comment as you see fit.


lpil (commented):
Thank you!

I see build has failed in a strange way, I expect this is due to a new Rust version being released or such.

Could you update the tests such that they demonstrate the fix working please 🙏

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__bit_arrays__bit_array.snap
```diff
@@ -20,7 +20,7 @@ pub fn main() {
 
 -export([main/0]).
 
--file("/root/project/test/my/mod.gleam", 1).
+-file("my/mod/project/test/my/mod.gleam", 1).
```
lpil, line 23: Why do these start with `my/mod` now? Shouldn't they be `project/...`

lpil:
Closes https://github.com/gleam-lang/gleam/issues/4141

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__bit_arrays__bit_array.snap
lpil: Sorry, I'm confused. Why do we use the module name as a prefix? That results in an invalid path to a file that doesn't exist?

oneness: @lpil No worries. Here is how I understood this issue:
If you look at the #4141, here is actual `-file` directory being generated before the fix:
-file("/Users/USERNAME/SOME/PATH/some_lib_or_app/src/some_lib_or_app.gleam", 26).
And the expectation is:
-file("some_lib_or_app/src/some_lib_or_app.gleam", 26), which the fix is about. 
In our pipeline, I understood that the project getting parsed as `/Users/USERNAME/SOME/PATH`, the module name as `some_lib_or_app`.  Thus the fix.  Let me know if you think there is better way to fix this issue.

lpil: I'm not sure I understand, so I'll give an example of what the behaviour should be.

Current working directory: `/root/project`
Source file location: `/root/project/src/wibble.gleam`
`-file` value: `src/wibble.gleam`

The tests don't seem to match this from what I can see, why is that?

oneness: Ah, if that is what the expected behaviour, then the fix would be much simpler and I could update the code changes and tests to reflect that.  So to confirm:
Currently (main branch) generates fqn for the generated erlang -file directive like this:
`/root/project/src/wibble.gleam` given the cwd of `/root/project`, which is the bug.
The correct behaviour is:
the file directive should be `src/wibble.gleam`. 

Please confirm and will take care of this soon.

oneness:
Closing this PR in favor of https://github.com/gleam-lang/gleam/pull/4193.

PR #4169, opened 2025-01-12
Make case expressions fault tolerant

### LINKED ISSUE #3978: Add fault tolerant parsing for `case` expressions
Currently, `case something` is a syntax error (obviously). The problem is, the language server cannot provide autocomplete when typing the subject of the case expression. We could make this fault tolerant in the same way as `record.`, to allow for a better LSP experience.

GearsDatapacks:
Closes #3978
This PR makes parsing of case expressions fault tolerant, allowing analysis to continue when a case expressions is missing a body. This mostly allows for LSP completions for the case subjects.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,11 @@
   shell starting from OTP27.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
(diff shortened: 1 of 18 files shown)


lpil (approved):
Thank you!

PR #4183, opened 2025-01-17
Pick better names when generating arguments or variables names

### LINKED ISSUE #4177: Improve naming for LS code generation actions
I think we could improve the way the LS comes up with names when it's generating functions/patterns and such. Right now when it generates a function with multiple arguments what it does looks like this:
```gleam
// When it generates a function:
pub fn generated(value_1: Wibble, value_2: String) { todo }

// Or when it generates a let destructuring when there's no labels:
pub fn main() {
  let Thing(value_1, value_2) = todo  
}
```
It would be really nice if it could instead use the types to come up with names, like this:
```gleam
// When it generates a function:
pub fn generated(wibble: Wibble, string: String) { todo }

// Or when it generates a let destructuring when there's no labels:
pub fn main() {
  let Thing(wibble, string) = todo  
}
```

There's a couple of things to keep in mind:
- When extracting a variable we could end up shadowing an existing name, so it would be nice (even as a future addition) to have it take the current scope into account to not shadow anything
- When two arguments have the same type we would need a way to generate different names, in that case I'd just use a number suffix `int_1`, `int_2`, ...
- When generating pattern matching patterns we always want to use labels when they're defined, in that case we need to be careful to generate valid code. This piece of code could be a good test to see if we're doing things right:
  ```gleam
  pub type Wibble {
    Wibble(String, string: String)
  }

  pub fn main(wibble: Wibble) {
    let Wibble(string_1, string:)
  //    ^^^^^^ Auto-generated by the ls when matching on a function argument
  }
  ```
- If a type is a generic we might still want to default to some generic name like `value` instead of using single letters like `a`, `b`... though I'm not sure what would be best here. For example if I'm generating the pattern matching for a `Result(a, String)` what should it look like?
  ```gleam
  case result {
    Ok(a) -> todo
    // or Ok(value) -> todo
    Error(string) -> todo
  }
  ```

giacomocavalieri:
This PR closes #4177


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__extract_variable.snap
```diff
@@ -11,6 +11,6 @@ pub fn main() {
 
 ----- AFTER ACTION
 pub fn main() {
-  let value = [1, 2, 3]
-  list.map(value, int.add(1, _))
+  let list = [1, 2, 3]
```
lpil, line 14: I wonder if `ints` would be a good name for lists of ints, `floats` for floats, etc. What do you think?

lpil (approved):
Beautiful! I've left one question inline

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__extract_variable.snap
giacomocavalieri: That would be really noce, maybe we could limit it to the core types we know will produce a good plural form and still fallback to `list` for user defined types?

lpil: Could do, why would you want it to work that way?

giacomocavalieri: I think it wouldn't look nice if we just added an s at the end of the type name maybe generating gibberish
```
// from gleam's codebase, this would
// become variable_namess
type VariableNames
type Fish
type Dice
```

lpil: Not adding if it ends in `s` would be fine I think.

giacomocavalieri: Ok!

lpil (approved):
Thank you!!!

PR #4184, opened 2025-01-17
Make sure a pipeline step cannot be extracted as a variable

### LINKED ISSUE #4134: extract variable does not insert `_` for first arg when extracting from a pipe
```gleam
dict.get(d, key)
|> result.unwrap(0)
```
on the `unwrap` yields
```gleam
  let value = result.unwrap(0)
  dict.get(d, key)
  |> value
```
which does not compile. Should be `value = result.unwrap(_, 0)`.

giacomocavalieri:
This PR closes #4134


lpil (approved):
Beautiful, thank you!!

PR #4185, opened 2025-01-18
Implement renaming for local variables

### LINKED ISSUE #2956: LSP: rename local variable

GearsDatapacks:
Closes #2956
This PR implements renaming of local variables in the Language Server including:
- Setting up the new route and server capability
- Setting up a test environment for testing renaming
- Tree walking to find references to the variable
- Rejecting renames if an invalid name is provided (e.g. renaming a variable to `SomeVar`)

A few things I did while implementing this:
- I used definition location to decide whether two variables were the same. That's mainly because since variable shadowing exists in Gleam, two variables with the same name don't necessarily refer to the same variable.
- To allow renaming to also rename references to variables in clause guards, I added a `definition_location` field to the `ClauseGuard::Var` variant. Since `ClauseGuard` is just one type for both typed and untyped versions, this field has to be present in the untyped AST, where it has no meaning (we don't know where the variable is defined yet). That was the most straightforward solution I could find for the time being.
- Similarly, I added clause guards to the AST visitor. I only really implemented the bare-bones for what this PR needs; it seemed a bit wasteful to implement it all. I'll probably open a separate PR to implement the full visitor for it

### DIFF
--- a/compiler-core/src/language_server/tests/rename.rs
+++ b/compiler-core/src/language_server/tests/rename.rs
@@ -183,6 +183,36 @@ pub fn main() {
 }
 
 #[test]
+fn rename_local_variable_assignment_pattern() {
+    assert_rename!(
+        "
+pub fn main() {
+  let assert Error(12 as something) = Error(12)
+  something
+}
+",
+        "the_error",
+        find_position_of("something")
+            .nth_occurrence(2)
+            .to_selection()
+    );
+}
+
+#[test]
+fn rename_local_variable_from_definition_assignment_pattern() {
+    assert_rename!(
+        "
+pub fn main() {
+  let assert Error(12 as something) = Error(12)
+  something
+}
+",
+        "the_error",
+        find_position_of("something)").to_selection()
+    );
+}
+
+#[test]
 fn rename_local_variable_argument() {
     assert_rename!(
         "
--- a/compiler-core/src/language_server/tests/rename.rs
+++ b/compiler-core/src/language_server/tests/rename.rs
@@ -163,3 +163,17 @@ pub fn main() {}
         find_position_of("fn").to_selection(),
     );
 }
+
+#[test]
+fn no_rename_invalid_name() {
+    assert_no_rename!(
+        "
+pub fn main() {
+  let wibble = 10
+  wibble
+}
+",
+        "Not_AValid_Name",
+        find_position_of("wibble").nth_occurrence(2).to_selection()
+    );
+}
--- a/compiler-core/src/language_server/tests/rename.rs
+++ b/compiler-core/src/language_server/tests/rename.rs
@@ -239,6 +239,42 @@ pub fn wibble(wibble: Float) {
 }
 
 #[test]
+fn rename_local_variable_label_shorthand() {
+    assert_rename!(
+        "
+type Wibble {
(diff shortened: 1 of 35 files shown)


GearsDatapacks:
Not sure why checks are failing... I haven't changed anything to do with deps or cargo deny

compiler-core/src/language_server/tests/rename.rs
```diff
@@ -0,0 +1,208 @@
...
+        "
+pub fn main() {
+  let wibble = 10
+  wibble
+}
+",
+        "Not_AValid_Name",
+        find_position_of("wibble").nth_occurrence(2).to_selection()
+    );
+}
```
lpil, line 208: Does it work for renaming function arguments and variables that are created using the label shorthand  syntax?

lpil (commented):
Ahhh this is so exciting!! Amazing! I've left one question inline

compiler-core/src/language_server/tests/rename.rs
GearsDatapacks: Ah no probably not. Good catch!
I'll look into that when I can

GearsDatapacks:
I have now made it work properly with:
- Assignment patterns (e.g. `1 as one`)
- Label shorthand
- Arguments

One thing I wasn't totally sure about is the following:
```gleam
let Wibble(wibble:) = todo
//         ^ If you put you cursor here and trigger a rename
```
What should happen? Should it treat is as renaming the local variable: `wibble: wobble`, or should it (once we have this feature implemented) rename the label itself? The latter is probably something we want eventually, however currently this PR implements the former. Let me know what you think

GearsDatapacks:
Just realised that the `prepare_rename` function wasn't up-to-date with the regular `rename` function. I've now fixed that and checked `prepare_rename` in the tests, so that shouldn't happen in the future. However, it is quite a lot code duplication. All the pattern matching has to be performed in two different places. I wonder if there's a way around this

lpil (approved):
Fantastic!!! I love this!

I found a tiny bug, it's not renaming variables used in bit array literals

```gleam
pub fn starts_with(bits: BitArray, prefix: BitArray) -> Bool {
  let prefix_size = bit_size(prefix)

  case bits {
    <<pref:bits-size(prefix_size), _:bits>> if pref == prefix -> True
    _ -> False
  }
}
```

Try renaming `prefix_size` here

GearsDatapacks:
Yay, another edge-case. I'll look into it

lpil (approved):
Exciting!!!!

stefanwatt:
Awesome! thanks for implementing this!!!
Doesn't it mean that the code needed for "find all references" is already implemented as well?
I would assume that you would need to find references before you can rename them.

lpil:
It's not, no.

lpil:
It does not, but find references for local variables could be implemented without difficulty.

Do you find it useful for local variables?

GearsDatapacks:
Yes! That should be pretty easy to implement. We would  just need to add the code to connect what we already have.

stefanwatt:
> Do you find it useful for local variables?

Would obviously be nicer to have it work globally and on functions etc., but certainly better than nothing 🤷

PR #4186, opened 2025-01-18
Implement AST visiting for clause guards

GearsDatapacks:
I found that I needed clause guard visiting when implementing #4185, but in that PR I only implemented the minimum for what I needed. This PR mostly implements AST visiting for clause guards.
I didn't implement a separate function for every single possible arithmetic operation as I wager that we won't need that any time soon. If we do, we can implement it then.


lpil (approved):
Thank you

PR #4198, opened 2025-01-25
Fix bug with scoping in let assert

### LINKED ISSUE #4195: Unbound variable getting passed the compiler
```gleam
import gleam/list
import gleam/string

pub fn main() {
  let assert Ok(thing) =
    ["", ""] |> list.find(fn(item) { string.length(item) == 1 })
    as { "you should see this " <> thing }
}
```
the compiler should tell that `thing` is unbound in the `as` block, but it doesn't

GearsDatapacks:
Fixes #4195 
This PR fixes the scoping issue by analysing the panic message before bringing the variable binding into scope.


lpil (approved):
Thank you!!

PR #4203, opened 2025-01-26
Fix renaming for Zed

### LINKED ISSUE #4201: LS rename doesn't work in Zed
Image

Here's what I get in Zed when I use it (by pressing F2).

I don't know what this means. It works in Neovim.

GearsDatapacks:
Fixes #4201
This PR fixes LS renaming for Zed by using the `Range` response for the `prepareRename` request, as opposed to the `DefaultBehaviour` response, which Zed does not seem to support. I don't think there's any way to unit test this, but I have tested an confirmed that this now works in Zed. It could be worth testing in other editors (Neovim, Helix) as I only have the Language Server set up for Zed and VSCode.


lpil (approved):
Thank you!!!

PR #4213, opened 2025-02-01
Warn if the same operand is on either side of the equality operator

### LINKED ISSUE #4205: Warn on useless comparisons
Hi there,

I had a typo in some code and I was comparing a variable with itself, like `group_id != group_id`. One of those was supposed to be `group.id`, and I think it's a common typo.

The idea is that the compiler is able to detect such comparisons and warn about them, since they will always evaluate to the same and are completely useless.

Thanks 🙇🏽

Ramkarthik:
Closes #4205 

Based on the conversation in the comments thread of the mentioned issue, I have implemented this for variables, field access, and literals. I'm not sure about module selects. @giacomocavalieri Could you please share an example for module select? (I'm still fairly new to Gleam)

Currently, it only checks for the same operand. In the case of literals, it doesn't warn if the check is `1 == 2` (which is always `false`). Do we want to handle those cases? If yes, I will likely have to rename the `Warning::EqualityOnSameOperands` variant and change the warning title. Any equality check for literals on both sides of the operator is always going to be the same result, so maybe we want to warn?

This PR also doesn't handle `gt`, `gte`, `lt`, and `lte` for Int and Float literals (based on the discussion in the issue thread).

If there's a better way to phrase the warning (title, hint, etc), please let me know.

Currently, the warning message template is:

```
warning: Same operand on either side of the equality operator
   ┌─ /code/gleam/gleam/test/equality/src/equality.gleam:10:8
   │
10 │   case name == name {
   │        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ This will always be true
```

Not related: It looks like the person who added this particular changelog entry missed adding their name/link - https://github.com/gleam-lang/gleam/commit/65d7bb554609d8ff3b0f2ba7c6ad86ffc48ed6aa

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,33 @@
   The custom type variant constructor `X` is not in scope here.
   ```
 
+- The compiler now warns when using the same operand on either side of the equality operator:
+
+  ```gleam
+  type Wibble {
+    Wibble(id: Int)
+  }
+
+  pub fn main() {
+    let wibble = Wibble(1)
+    case wibble.id == wibble.id {
+      True -> Nil
+      False -> Nil
+    }
+  }
+  ```
+
+  Results in the following warning:
+
+  ```
+    warning: Same operand on either side of the equality operator
+    ┌─ /app.gleam:7:8
+    │
+  7 │   case wibble.id == wibble.id {
+    │        ^^^^^^^^^^^^^^^^^^^^^^ This will always be true
+  ```
+  ([Ramkarthik Krishnamurthy](https://github.com/ramkarthik))
+
 ### Build tool
 
 - `gleam new` now has refined project name validation - rather than failing on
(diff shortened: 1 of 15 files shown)


Ramkarthik:
My bad, I forgot to run `rust fmt` which is causing the `ci/rustfmt` check to fail. Since there will likely be suggested changes, I will combine this with the changes in the next commit.

lpil:
Thank you! Please see my comment here https://github.com/gleam-lang/gleam/issues/4205#issuecomment-2628939280

lpil:
Hi @Ramkarthik ! Are you still working on this? 🙏

Ramkarthik:
> Hi @Ramkarthik ! Are you still working on this? 🙏

Hi @lpil ! Apologies, I haven't had much time to get into this recently. I'll likely have some time by the end of this coming week. But in case you or someone plans to take it, I'm good with that.

lpil:
Going to close this as there has been no recent activity, but please feel free to open a new one! Thank you.

Issue #3936, opened 2024-12-02
Extract const code action


lpil:
An action that extracts a highlighted value and replaces it with a const
```gleam
pub fn main() {
  list.map([1, 2, 3], run)
  #        ^^^^^^^^^
}
```

```gleam
const value = [1, 2, 3]

pub fn main() {
  list.map(value, run)
}
```

matiascr:
I'm taking a look at this. I'm guessing this should also perform these kinds of conversions?

Before:
```gleam
pub fn main() {
  let l = ["A", "B", "C"]
  list.map(l, run)
}
```

After extracting whole composite type:
```gleam
const l = ["A", "B", "C"]

pub fn main() {
  list.map(l, run)
}
```

Or after extracting by individual literals:
```gleam
const string = "A"

pub fn main() {
  let l = [string, "B", "C"]
  list.map(l, run)
}
```

lpil:
Yeah!!

matiascr:
WIP @ https://github.com/gleam-lang/gleam/pull/4228

lpil:
If you'd like to PR just a subset of expressions to start that would be good by me!

GearsDatapacks:
I believe this can be closed now, as the relevant PR has been merged

PR #4228, opened 2025-02-10
Added extract constant code action

matiascr:
Addresses https://github.com/gleam-lang/gleam/issues/3936


GearsDatapacks:
FYI before this can be merged, you'll need to add tests. Those can be found in the `language_server/tests/action.rs` file. (It also helps us review changes as the snapshots show the output format of the action!)
Also, you will need to update CHANGELOG.md to include this change.

lpil:
Thank you both!

compiler-core/src/language_server/code_action.rs
```diff
@@ -2851,6 +2851,137 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+                .iter()
+                .filter_map(|definition| match definition {
+                    ast::Definition::ModuleConstant(module_constant) => {
+                        Some(module_constant.name.clone())
+                    }
+                    _ => None,
+                })
+                .collect(),
+        };
+        name_generator.reserve_variable_names(module_constant_names);
```
lpil, line 2919: Functions would also want to be taken into account here

compiler-core/src/language_server/code_action.rs
```diff
@@ -2851,6 +2851,137 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+            self.edits
+                .replace(expression_span, String::from(variable_name));
+        }
+
+        let mut action = Vec::with_capacity(1);
+        CodeActionBuilder::new("Extract constant")
+            .kind(CodeActionKind::REFACTOR_EXTRACT)
+            .changes(self.params.text_document.uri.clone(), self.edits.edits)
+            .preferred(false)
+            .push_to(&mut action);
```
lpil, line 2953: When is it possible for this to ever be `None`? It looks like in that case it would add a refactoring that doesn't contain any edits?

compiler-core/src/language_server/code_action.rs
```diff
@@ -2851,6 +2851,137 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+                }
+                // Extract literals directly
+                TypedExpr::Int { .. } | TypedExpr::Float { .. } | TypedExpr::String { .. } => {
+                    self.selected_expression = Some((expr_location, expr.type_()));
+                }
+                _ => (),
+            }
+        }
+
+        ast::visit::visit_typed_expr(self, expr);
```
lpil, line 2981: Don't we want to stop walking the tree if the expression has already been found?

compiler-core/src/language_server/code_action.rs
```diff
@@ -2851,6 +2851,137 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+
+        let container_function_start = self
+            .module
+            .ast
+            .definitions
+            .iter()
+            .filter(|definition| definition.is_function())
+            .map(|function| function.location().start)
+            .filter(|function_start| function_start < &expression_span.start)
+            .max();
```
lpil, line 2936: Could we do this during the tree visiting?

lpil (commented):
Thank you!! Looking very nice.

Could you add tests for these please:

- Expressions with function calls
- Expressions using local variables 
- Expressions using records
- Expressions use constants
- Expression with bool operators
- Expressions with logical operators

compiler-core/src/language_server/code_action.rs
matiascr: Not quite, it's required to continue in case we want to extract a literal that's inside, say, a function call or record constructor. Maybe it won't be needed after applying the rest of the changes, right now it's necessary.

lpil:
If you are ready for a review please un-draft the PR 🙏

compiler-core/src/language_server/code_action.rs
```diff
@@ -2852,6 +2852,277 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+
+        if !within(self.params.range, expr_range) {
+            ast::visit::visit_typed_expr(self, expr);
+            return;
+        }
+
+        if matches!(
+            self.type_of_extractable,
+            None | Some(ExtractableToConstant::Collection)
+        ) && is_literal_or_made_of_literals(expr)
```
lpil, line 3101: What is this bit doing? Would be useful to have documentation comments explaining the intent

compiler-core/src/language_server/code_action.rs
```diff
@@ -2852,6 +2852,277 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+                    .code
+                    .get((expr_location.start as usize)..(expr_location.end as usize))
+                    .expect("selected expression"),
+            );
+
+            self.type_of_extractable = Some(match expr {
+                TypedExpr::Int { .. } | TypedExpr::Float { .. } | TypedExpr::String { .. } => {
+                    ExtractableToConstant::Value
+                }
+                _ => ExtractableToConstant::Collection,
```
lpil, line 3114: This mapping of types to `ExtractableToConstant` isn't quite right. They're all values, and there are things that are not containers being marked as containers here.

compiler-core/src/language_server/code_action.rs
```diff
@@ -2852,6 +2852,277 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+/// ```
+pub struct ExtractConstant<'a> {
+    module: &'a Module,
+    params: &'a CodeActionParams,
+    edits: TextEdits<'a>,
+    selected_expression: Option<(SrcSpan, Arc<Type>)>,
+    container_function_start: Option<u32>,
+    type_of_extractable: Option<ExtractableToConstant>,
+    name_to_use: Option<EcoString>,
+    value_to_use: Option<EcoString>,
```
lpil, line 2885: Would be useful to have these documented as it's a little unclear what they are.

What's the different between "value to use" and "selected expression"?

"type of extractable" seems to not be a type.

lpil (commented):
Fab work! Thank you!

I've left some comments inline.

Could you add these tests also:

- Extracting results
- Extracting Nil
- Extracting bit arrays
- Extracting a custom type that has fields (a record)
- Extracting expressions which reference constant variables
- Extracting expressions from assignments that use complex patterns (the assignment can stay and just the value is replaced)
- Refusing to extract expressions which reference non-constant variables
- Refusing to extract expressions which call functions

And anything else you can think of. I may have missed something 😁 

Thank you!

compiler-core/src/language_server/code_action.rs
```diff
@@ -2852,6 +2852,277 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+}
+
+enum ExtractableToConstant {
+    Collection,
+    Value,
+    Assignment,
+}
+
+fn is_literal_or_made_of_literals(expr: &TypedExpr) -> bool {
```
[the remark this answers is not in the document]
matiascr: Good point, I'll update it and rename it to `can_be_constant` for clarity

matiascr:
Support for records is still missing and proving to be a bit tricky

lpil:
Hello! Did you see my comments above?

compiler-core/src/language_server/code_action.rs
```diff
@@ -2852,6 +2852,277 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
+    params: &'a CodeActionParams,
+    edits: TextEdits<'a>,
+    selected_expression: Option<(SrcSpan, Arc<Type>)>,
+    container_function_start: Option<u32>,
+    type_of_extractable: Option<ExtractableToConstant>,
+    name_to_use: Option<EcoString>,
+    value_to_use: Option<EcoString>,
+}
+
+enum ExtractableToConstant {
```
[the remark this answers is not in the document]
matiascr: Done, let me know if it's clear

compiler-core/src/language_server/code_action.rs
matiascr: Added explanation to the comments, let me know if that clears it up

compiler-core/src/language_server/code_action.rs
matiascr: Moved it to `visit_typed_function`, much cleaner now.

compiler-core/src/language_server/code_action.rs
matiascr: This code is now only reachable if the edits are actually going to take place

matiascr:
Hi @lpil , 
What would tests look like for the following?
- Extracting expressions which reference constant variables
- Extracting expressions from assignments that use complex patterns (the assignment can stay and just the value is replaced)

Could you give an example? Do you mean bracketed expressions or just patterns for pattern matching?

GearsDatapacks:
I think like this is what he meant:
```gleam
const x = 1
...
let y = [x, 2, 3]
```
And
```gleam
let #(a, b, c) = #(1, 2, 3)
```

matiascr:
Both those cases are already taken into account then. I've updated a test case to cover when the left-hand side is a pattern.
See:
- https://github.com/gleam-lang/gleam/pull/4228/files#diff-b5289de1adf4826f9c88ac877d14ace957c0c5483e37cb080020958b51f8e255
- https://github.com/gleam-lang/gleam/pull/4228/files#diff-1d42888f29b71c9631f6e337c469d2bf04f4ddabe5c7ad5710b3d93b43cfd1e1


Selecting the patttern (`let x = ...`) will extract the whole assignment, if possible. Selecting the value will keep the variable assignment and extract only the value.

lpil (approved):
Beautiful work! Very impressive! Thank you so much!

PR #4232, opened 2025-02-11
Make constructor arguments fit better on mobile

Nicd:
This is a minor change to the styling of constructor arguments in generated documentation. Currently they are indented quite a lot and waste space especially on small screens. With this change, the indentation is lessened and the argument name and description can overlap, allowing for better wrapping even with long argument names. Before and after:

Screenshot 2025-02-11 at 19 28 36 Screenshot 2025-02-11 at 19 41 21

## Technical details

This uses the [description list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) (dl) element which is both stylistically and semantically appropriate. Argument names are description terms (dt) and their descriptions are... descriptions (dd).

I kept the argument name as italic to match the previous style.

An alternative would be to use a regular list (as before) but remove the flexbox alignment and align it similar to this with CSS.

## Accessibility

Screen reader support seems to be adequate: https://adrianroselli.com/2025/01/updated-brief-note-on-description-list-support.html -- At the least the situation is mostly similar to how it is now.

## Browser support

Tested on macOS using Vivaldi (Chrome based), Safari, and Firefox. Looks the same on each one. The element is very old so support should be good.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
   making Gleam packages discoverable through global search of HexDocs.
   ([Diemo Gebhardt](https://github.com/diemogebhardt))
 
+- Improved the styling of constructor argument descriptions in the generated
+  documentation. ([Nicd](https://git.ahlcode.fi/nicd))
+
 - Allow users to set the `GLEAM_CACERTS_PATH` environment variable to specify a
   path to a directory containing CA certificates to install Hex packages.
   ([winstxnhdw](https://github.com/winstxnhdw))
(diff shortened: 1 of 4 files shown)


lpil (approved):
Soooo much nicer! Thank you. Could you update the changelog please 🙏

Nicd:
Whoops I pressed the wrong button, lemme remove that merge from there and add the changelog changes.

lpil (approved):
Thank you!!!

PR #4233, opened 2025-02-11
Convert to pipe

giacomocavalieri:
This PR implements a code action to turn a function call into a pipeline.

| Header | Header |
|--------|--------|
| When piping the first argument will rewrite the pipe to avoid using a capture hole |  |
| But it can work with arguments in any position by automatically inserting a capture hole |  |
| It respects the labels written down by the user never removing them (even when piping the first argument) |  |


giacomocavalieri:
Oh it looks like the diff looks particularly rubbish after rebasing:
- In the changelog I moved gears' entry above mine so that the "turn into pipe" and "turn into function" ones would be close together
- In the tests I moved the "turn into function" tests to be close together, it looks like they ended up mixed with the "generate decoder" tests during a rebase

lpil:
lmk when you need a review!

giacomocavalieri:
This is ready for review!

lpil (approved):
Beautiful work!!!! Thank you

PR #4234, opened 2025-02-11
Rename "Convert to/from use" action builders

giacomocavalieri:
I've renamed the `ConvertTo/FromUse` action builder to be consistent with the name of the action they refer to


lpil (approved):
Thank you

PR #4236, opened 2025-02-12
LSP Code Action: Inline Variable

### LINKED ISSUE #3935: Inline variable code action
An action that inlines a variable that is used once
```gleam
pub fn main() {
  let value = [1, 2, 3]
  list.map(value, run)
  #        ^^^^^
}
```

```gleam
pub fn main() {
  list.map([1, 2, 3], run)
}
```

GearsDatapacks:
Closes #3935

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,30 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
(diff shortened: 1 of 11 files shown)


lpil (approved):
Beautiful work! Thank you!!

PR #4237, opened 2025-02-12
Fix convert to call

giacomocavalieri:
I found a couple of bugs in the "convert to call" code action testing it out on some real code:
- It wouldn't work with module selects, moving only the bit after the `.`
- It wouldn't work with module selects that cannot be inferred, moving only the bit after `.`
- It would remove labels when piping into a capture with explicit labels written down

This PR fixes all of those issues

### DIFF
--- a/compiler-core/src/ast/tests.rs
+++ b/compiler-core/src/ast/tests.rs
@@ -366,12 +366,12 @@ fn find_node_tuple_index() {
 #[test]
 fn find_node_module_select() {
     let expr = TypedExpr::ModuleSelect {
-        location: SrcSpan { start: 1, end: 3 },
+        location: SrcSpan { start: 1, end: 4 },
+        field_start: 2,
         type_: type_::int(),
         label: "label".into(),
         module_name: "name".into(),
         module_alias: "alias".into(),
-        module_location: SrcSpan { start: 0, end: 0 },
         constructor: ModuleValueConstructor::Fn {
             module: "module".into(),
             name: "function".into(),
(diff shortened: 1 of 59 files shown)


compiler-core/src/ast/typed.rs
```diff
@@ -97,11 +97,27 @@ pub enum TypedExpr {
...
         module_name: EcoString,
         module_alias: EcoString,
+        /// The location of the module before the `.`:
+        ///
+        /// ```gleam
+        ///    wibble.wobble
+        /// // ^^^^^^ module_location
+        /// ```
+        ///
+        module_location: SrcSpan,
```
lpil, line 120: The second int is redundant, it is the same as the first int in location.

Perhaps we want to have location be the whole thing and then have a `field_start` or something which is the location of the first character after the `.`? Would be nice to have the warning and errors from from after the dot rather than on it too so you don't have to press right after jumping to the error to fix it.

compiler-core/src/ast/tests.rs
```diff
@@ -371,6 +371,7 @@ fn find_node_module_select() {
         label: "label".into(),
         module_name: "name".into(),
         module_alias: "alias".into(),
+        module_location: SrcSpan { start: 0, end: 0 },
```
lpil, line 374:
```suggestion
        module_location: SrcSpan { start: 0, end: 1 },
```

lpil (commented):
Thank you! Couple notes inline

compiler-core/src/ast/typed.rs
giacomocavalieri: I would love for the `location` to cover the entire field, I've always found it a bit odd that the module selection location doesn't cover the whole thing, and it means we always have to special case those in code actions that need to work on their entire span. I'll try and fix it!

lpil: Thanks Jak!

giacomocavalieri: Done! <https://github.com/gleam-lang/gleam/pull/4237/commits/fafcfb8e2c309835ebce96ad5809e6e63110ba76>

lpil (approved):
Thank you!!

PR #4245, opened 2025-02-14
"Generate function" no longer ignores labels

### LINKED ISSUE #4226: "Generate function" code action doesn't take labels into account
At the moment if I have a function that doesn't exist and is also using labels like this: `wibble(wobble: 1)`, the "Generate function" code action will generate a function that has no labels and so will be invalid. It would be nice if it took labels into account as well. One think I'm not sure how to deal with is what to do when the labels are not right:
- If there's duplicate labels should we generate an invalid function with duplicate labels? Or should the LS not prompt any action at all?
- If there's an unlabelled argument after labelled ones should the language server generate an invalid function with labelled arguments in the wrong place, a valid function but without the wrong-positioned labels, or should it refuse to generate anything at all?

giacomocavalieri:
This pr fixes #4226

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,9 @@
 - Fixed a bug where division and remainder operators would not work correctly
   in guards on the JavaScript target.
   ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you!!

PR #4247, opened 2025-02-14
Implement Git Dependencies

### LINKED ISSUE #1338: Support git deps with build tool

GearsDatapacks:
This PR closes #1338 
This is an implementation of Git Dependencies for the Gleam build tool. It works, although it is a bit slow at the moment (not sure how much we can do about that).
I haven't implemented anything like this before, so I have no idea if I've gone about this the right way at all. There are no tests yet because first I wanted to get feedback on the actual implementation and get everything working smoothly before writing tests.


compiler-cli/src/dependencies.rs
```diff
@@ -881,18 +905,97 @@ fn provide_local_package(
...
+    let commit = download_git_package(&package_name, repo, commit)?;
+
+    let package_source = ProvidedPackageSource::Git {
+        repo: repo.into(),
+        commit,
     };
-    Err(Error::GitDependencyUnsupported)
+
+    let package_path =
+        fs::canonicalise(&parent_path.join(format!("build/packages/{package_name}")))?;
```
lpil, line 990: Use ProjectPaths please

compiler-cli/src/dependencies.rs
```diff
@@ -434,23 +435,40 @@ async fn add_missing_packages<Telem: Telemetry>(
...
+        .copied()
         .filter(|package| package.is_hex())
         .inspect(|_| {
             num_to_download += 1;
         })
         .peekable();
 
     // If we need to download at-least one package
-    if missing_hex_packages.peek().is_some() {
+    if missing_hex_packages.peek().is_some() || !missing_git_packages.is_empty() {
```
lpil, line 458:
```suggestion
    if !missing_hex_packages.is_empty() || !missing_git_packages.is_empty() {
```

Let's make 'em match

lpil (commented):
Awesome! I've left some tiny notes inline, and the changelog will need updating.

Me or someone will have to give this some manual testing, but I'm a bit busy this weekend.

PgBiel (commented):
Nice job!!!

If it helps, for comparison, here's my previous progress on this feature: https://github.com/gleam-lang/gleam/compare/main...PgBiel:gleam:git-pkg

My main concern, as outlined in https://github.com/gleam-lang/gleam/issues/1338#issuecomment-2054564207, is that we should think about reducing the amount of work if the git repository was already cloned and is already in the correct commit. While it's not essential for the feature to work and could be skipped in an initial version, it'd be nice to ensure we don't need to re-fetch the repository if not needed to ensure optimal speed in updating dependencies. Perhaps this could be done by a simple comparison between HEAD and the commit hash (and, if it's not a commit hash, such as `main`, always re-fetch)?

GearsDatapacks:
That works, except the commit hash will often be partial. I guess we just check if it starts with that string? Are there any potential issues with that?

PgBiel:
If the user specified commit hash `abc` and the already checked out repository has commit `abcdefg`, I think it'd be fine to not re-fetch since otherwise it would be ambiguous.

With that said, the user could also be specifying a branch name, in which case we re-fetching is probably a good idea regardless.

So, to be sure, we could begin with re-fetching only if it's unlocked or if `gleam deps update [name]` is called.

GearsDatapacks:
Yep. I think that makes sense, although I'm still coming to terms with this part of the codebase so I'll have to try and figure out how exactly that works

compiler-cli/src/dependencies.rs
GearsDatapacks: Thing is, `missing_git_packages` is a `Vec`, while `missing_hex_packages` is an `Iterator`. This is because the `Downloader::download_hex_packages` function takes an iterator and not a vector. To make them the same, we would need to collect the hex packages then iterate over them again, which seems slightly wasteful for a smal code change such as this.

lpil:
I would like to not add that optimisation for this first pass, instead we focus on getting it working and we can optimise it later. The other BEAM build tools don't have the optimisation, so it seems not impactful enough to cause problems.

compiler-cli/src/dependencies.rs
lpil: OH! I see. Thank you

compiler-cli/src/publish.rs
```diff
@@ -24,7 +25,8 @@ pub fn command(replace: bool, i_am_sure: bool) -> Result<()> {
 
     let should_publish = check_for_gleam_prefix(&config)?
         && check_for_version_zero(&config)?
-        && check_repo_url(&config, i_am_sure)?;
+        && check_repo_url(&config, i_am_sure)?
+        && check_for_git_dependencies(&config)?;
```
giacomocavalieri, line 29: I'd move this check where we also check for path dependencies, otherwise this would happen before compiling, while the other one happens after compilation.

compiler-cli/src/publish.rs
```diff
@@ -254,6 +256,26 @@ core team.\n",
...
+        println!(
+            "You are trying to publish a package which uses git dependencies.
+Git dependencies cannot be used in published packages."
+        );
+        Ok(false)
+    } else {
+        Ok(true)
+    }
+}
+
```
giacomocavalieri, line 278: I'd change this so that it produces the same nice error used by the path dependencies check:
Screenshot 2025-02-16 alle 13 31 56

CHANGELOG.md
```diff
@@ -17,6 +17,12 @@
   path to a directory containing CA certificates to install Hex packages.
   ([winstxnhdw](https://github.com/winstxnhdw))
 
+### Build tool
+
+- The build tool now supports Git dependencies, and prevents packages being published
+  if they depend on any git packages.
+  ([Surya Rose](https://github.com/GearsDatapacks))
```
giacomocavalieri, line 24: I'd include an example showing how one can specify a git dependency and what's the different things one can put in the `ref` field. Also this information should be included somewhere that's not just the changelog. Maybe [here](https://gleam.run/writing-gleam/gleam-toml/)?

compiler-cli/src/dependencies.rs
```diff
@@ -881,18 +904,100 @@ fn provide_local_package(
     )
 }
 
+fn download_git_package(
+    package_name: &str,
+    repo: &str,
+    commit: &str,
+    project_paths: &ProjectPaths,
+) -> Result<EcoString> {
```
giacomocavalieri, line 912: I'd add a doc comment explaining what commit should be, is it just a commit hash? Or can it be a branch's name like `main` as well?

compiler-cli/src/publish.rs
GearsDatapacks: Ah yes, I couldn't find where that check happens. I'll try to find it now

CHANGELOG.md
GearsDatapacks: Yeah that's probably a good idea

compiler-cli/src/dependencies.rs
GearsDatapacks: Yeah it can be a branch name. The current name of the argument is not great, but the annoying thing is that `ref` is already a rust keyword. I guess I can change it to `ref_`?

giacomocavalieri: Yeah I'd go for `ref_` the same way we use `type_`

PgBiel:
> I would like to not add that optimisation for this first pass, instead we focus on getting it working and we can optimise it later. The other BEAM build tools don't have the optimisation, so it seems not impactful enough to cause problems.

I haven't studied all of the code yet, but in principle at least only fetching if the dependency is unlocked - usually because the chosen `url` or `ref` changed - sounds reasonable, right? Otherwise we'd keep fetching on each run, or each time anything changes.

In addition, I have the requirement of being able to supply the repository beforehand (i.e. have Gleam not fetch it, but use a pre-fetched repository). While this isn't equivalent to stopping fetches, could we at least make it so a failed fetch doesn't block a run / manifest update if the repository was already fetched and the given `ref` is present? (That is, only block the update if `git checkout` fails.)

lpil:
We only get packages if we don't already have them, we don't do it every time.

We will always halt in the event of an error.

PgBiel:
Oh okay. So are you saying that it won't try to fetch if the Git dependency is already in the `build/packages`?

(Just to clarify: My requirements are because I run Gleam builds in isolated build environments, without network access, so I need to vendor deps before the build.)

compiler-cli/src/dependencies.rs
```diff
@@ -881,18 +904,101 @@ fn provide_local_package(
...
+            program: "git".into(),
+            err: Some(error.kind()),
+        })?;
+
+    let commit = String::from_utf8(output.stdout)
+        .expect("Output should be UTF-8")
+        .trim()
+        .into();
+
+    Ok(commit)
```
PgBiel, line 972: If I understood correctly, this indicates that what is saved to the manifest is the commit, and not the original `ref` (e.g. if the user writes `ref = "main"`, in the manifest `[packages]`, we get `"bdc02d0as020..."` instead of `"main"`), which would be the correct behavior, I believe. Is this right?

GearsDatapacks: Yep. The manifest stores the full commit hash, regardless of the constraint in `gleam.toml`

lpil:
@GearsDatapacks shall we add an integration test to the github workflow?

GearsDatapacks:
Yep. Should I do something similar to the `path_deps` test?

lpil (approved):
Thank you!!

GearsDatapacks:
Ok, I'm not sure what exactly is causing the issue. It might be a problem with calling the git program from CI, or maybe it's an authentication thing.

GearsDatapacks:
Ok, the tests are passing now. I accidentally used the ssh url instead of the https one. In the process of debugging that, I improved the error reporting if the command fails, so hopefully mistakes such as that should be easier to catch in future.

lpil (approved):
Amazing!! People are going to be so excited for this one!! Thank you

PR #4248, opened 2025-02-15
Fix `gleam new` README.md formatting

diemogebhardt:
Not sure if this is something you care about:

The current `README.md` generated by `gleam new` violates rule [MD031 - Fenced code blocks should be surrounded by blank lines](https://github.com/markdownlint/markdownlint/blob/main/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines) of `markdownlint`.


lpil:
We do not accept unplanned changes. Please open a discussion if you have a proposal, thank you.

PR #4252, opened 2025-02-17
Fix generation of package interface when modules are cached

### LINKED ISSUE #2898: `gleam export package-interface` produces empty modules object

GearsDatapacks:
This PR follows on from #3669 and fixes #2898.
It uses partly the same approach of using the `ModuleInterface` (the part which is cached) to generate the package interface, rather than the AST. However, this is implemented in such a way that the cached modules do not need to be cloned when doing so.

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -202,7 +202,7 @@ where
                         cached_module_names,
                     },
                     error,
-                )
+                );
             }
             Outcome::TotalFailure(error) => return Outcome::TotalFailure(error),
         };
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -30,6 +30,11 @@ use camino::{Utf8Path, Utf8PathBuf};
 
 use super::{ErlangAppCodegenConfiguration, TargetCodegenConfiguration, Telemetry};
 
+pub struct Compiled {
+    pub modules: Vec<Module>,
+    pub module_names: Vec<EcoString>,
+}
+
 #[derive(Debug)]
 pub struct PackageCompiler<'a, IO> {
     pub io: IO,
@@ -104,7 +109,7 @@ where
         stale_modules: &mut StaleTracker,
         incomplete_modules: &mut HashSet<EcoString>,
         telemetry: &dyn Telemetry,
-    ) -> Outcome<Vec<Module>, Error> {
+    ) -> Outcome<Compiled, Error> {
         let span = tracing::info_span!("compile", package = %self.config.name.as_str());
         let _enter = span.enter();
 
@@ -145,6 +150,8 @@ where
             Loaded::empty()
         };
 
+        let mut module_names = Vec::new();
+
         // Load the cached modules that have previously been compiled
         for module in loaded.cached.into_iter() {
             // Emit any cached warnings.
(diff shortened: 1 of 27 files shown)


compiler-core/src/build/package_compiler.rs
```diff
@@ -30,6 +30,11 @@ use camino::{Utf8Path, Utf8PathBuf};
 
 use super::{ErlangAppCodegenConfiguration, TargetCodegenConfiguration, Telemetry};
 
+pub struct Compiled {
```
lpil, line 33: Why is this done? The `Module` struct already contains the name.

compiler-core/src/metadata/tests.rs
```diff
@@ -1557,6 +1630,8 @@ fn type_with_inferred_variant() {
         line_numbers: LineNumbers::new(""),
         src_path: "some_path".into(),
         minimum_required_version: Version::new(0, 1, 0),
+        type_aliases: HashMap::new(),
+        documentation: Vec::new(),
```
lpil, line 1634: Please add some new tests when new functionality is added to the metadata encoding and decoding. This is the area of the codebases that causes the most bugs

compiler-core/src/type_.rs
```diff
@@ -1208,6 +1212,7 @@ impl TypeVar {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct TypeConstructor {
     pub publicity: Publicity,
+    pub opaque: bool,
```
lpil, line 1215: This type design means that it is now possible for a type to be private and opaque at the same time, but that shouldn't be possible. Using a bool is also error prone and unclear, it'd need to be a new type. Perhaps opacity could be part of the existing publicity type? That said I'm not immediately into information about the value constructors being recorded here as this is information about the type constructor...

Do you have any thoughts on this? I'm not sure what's best

compiler-core/src/type_/fields.rs
```diff
@@ -201,6 +201,13 @@ impl FieldMap {
             .map(|(label, _position)| label.clone())
             .collect_vec()
     }
+
+    pub fn indices_to_labels(&self) -> HashMap<u32, EcoString> {
```
lpil, line 205: Looks like this clone isn't needed from where it is used

lpil (commented):
Wonderful!! Thank you! I've left a few questions inline

compiler-core/src/type_/fields.rs
GearsDatapacks: Good point, I'll remove it

compiler-core/src/type_.rs
GearsDatapacks: I don't have any immediate thoughts. I'll look at where it's used and try to figure something out

GearsDatapacks:
All your comments/questions should now be resolved. I've moved the `opaque` field to the `TypesVariantConstructors` type as it seemed more appropriate there. I've also used a custom type instead of a bool now.

lpil (approved):
Thank you!!!

PR #4256, opened 2025-02-18
Bundle nested FFI on publish

### LINKED ISSUE #4238: Nested external FFI files are not bundled when publishing
Hi!

When publishing a package with JS external FFI files not at root, they are not bundled in the output tarball.
The bug can easily be reproduced with. the following steps:

```sh
git clone git@github.com:ghivert/grille-pain.git
cd grille_pain
git checkout e1fa1168d5b6306cd134005721c78cac2bfad7e4
gleam publish
```

You should have the following display:

```
Source files:
  - LICENCE
  - README.md
  - gleam.toml
  - src/grille_pain/error.gleam
  - src/grille_pain/internals/css.gleam
  - src/grille_pain/internals/data/model.gleam
  - src/grille_pain/internals/data/msg.gleam
  - src/grille_pain/internals/data/toast.gleam
  - src/grille_pain/internals/effect_manager.gleam
  - src/grille_pain/internals/element.gleam
  - src/grille_pain/internals/global.gleam
  - src/grille_pain/internals/lustre/schedule.gleam
  - src/grille_pain/internals/setup.gleam
  - src/grille_pain/internals/shadow.gleam
  - src/grille_pain/internals/unsafe.gleam
  - src/grille_pain/internals/view/progress_bar.gleam
  - src/grille_pain/internals/view/theme.gleam
  - src/grille_pain/internals/view.gleam
  - src/grille_pain/lustre/toast.gleam
  - src/grille_pain/options.gleam
  - src/grille_pain/toast/level.gleam
  - src/grille_pain/toast.gleam
  - src/grille_pain.gleam

Name: grille_pain
Version: 1.1.1

Do you wish to publish this package? [y/n]:
```

Where you can see that every JS files are missing.

PgBiel:
Fixes #4238

Not sure if `erlang_files` should be modified as well, let me know if it'd be better to remove it from this PR.


lpil:
Are you ready for a review? :)

PgBiel:
Not yet, I'll be experimenting a bit further with `include/` still, that's the only missing part. I'll let you know once it's ready!

PgBiel:
Okay, I figured it out: the exported `include/` directory is auto-generated on erlang codegen and is always flat (at least, the files produced by Gleam always appear at the top-level - not sure if `erl` can produce anything there, but I couldn't get it to at first), so 1. no need to export nested files and 2. no need for .gitignore checks (since the auto-generated dir is inside `build/`, so it would always be ignored...).

compiler-cli/src/publish.rs
```diff
@@ -510,7 +508,12 @@ fn generated_erlang_files(
 
     // Erlang headers
     if include.is_dir() {
-        for file in fs::erlang_files(&include)? {
+        // Note that the include subdirectory of 'build/prod/.../package'  is
+        // auto-generated on Erlang codegen based on exposed record types, and
+        // is always flat, so we can check only for top-level Erlang files.
+        // Precisely because it is auto-generated, we also don't check for
+        // '.gitignore' since it would typically always ignore 'build/' anyway.
+        for file in fs::toplevel_erlang_files(&include)? {
```
lpil, line 516: Use the recursive version please. Include is not limited to not have subdirectories.

compiler-cli/src/publish.rs
```diff
@@ -754,3 +757,107 @@ fn prevent_publish_git_dependency() {
...
+        "src/exported_ffi.js",
+        "src/exported_ffi.mjs",
+        "src/exported_ffi.ts",
+        "src/nested/exported.gleam",
+        "src/nested/exported_ffi.erl",
+        "src/nested/exported_ffi.ex",
+        "src/nested/exported_ffi.hrl",
+        "src/nested/exported_ffi.js",
+        "src/nested/exported_ffi.mjs",
+        "src/nested/exported_ffi.ts",
```
lpil, line 795: This is missing the gitignored files!

lpil (commented):
Thank you! I've left some notes inline. Could you update the changelog too please 🙏

compiler-cli/src/publish.rs
PgBiel: yes, they are in a separate list so that I don't have to specify them twice while building the `.gitignore`  file for the test. They are still unexported.

PgBiel: I've made the test clearer with your feedback.

compiler-cli/src/publish.rs
```diff
@@ -754,3 +752,113 @@ fn prevent_publish_git_dependency() {
...
+        ".gitignore",
+        "build/",
+        "ignored.txt",
+        "priv/ignored",
+        "src/.hidden/hidden.gleam",
+        "src/.hidden/hidden_ffi.erl",
+        "src/.hidden/hidden_ffi.mjs",
+        "src/.hidden.gleam",
+        "src/.hidden_ffi.erl",
+        "src/.hidden_ffi.mjs",
```
PgBiel, line 805: It appears that any files and folders prefixed with a dot `.` are ignored regardless of `.gitignore`. This is a setting that we can toggle on `ignore::WalkDir`. Should we include those as well?

lpil: I don't think we should be using `ignore` for this as we don't want to ignore anything.

compiler-cli/src/publish.rs
lpil: Oh! I'm understanding the confusion now! There's a bug in the old code where it is excluding gitignored files from the publish for some reason. This shouldn't be the case, all files should be included.

compiler-cli/src/publish.rs
PgBiel: Pardon? I thought we would be ignoring .gitignored files.

What I'm asking is if we should ignore files and directories starting with a dot as well (currently they are ignored, regardless of .gitignore). I'd say that perhaps we shouldn't.

lpil: See my comment here: https://github.com/gleam-lang/gleam/pull/4256#discussion_r1976389845

PgBiel: Oh okay! I hadn't seen it, sorry.

We can still use `ignore::WalkDir` but disabling all filters, that should work (I'm using this for `erlang_files`).

Alternatively, we can try to use our existing `DirWalker` implementation, but it requires a reference to a `FilesystemReader`, so we can't just create one in the function as it'd still be borrowed, requiring some way to pass one as a parameter everywhere, which was why I avoided it for now... Let me know if we should find a way to do this instead.

compiler-cli/src/publish.rs
PgBiel: Done!

compiler-cli/src/publish.rs
PgBiel: Okay, I've disabled all filters for `ignore::WalkDir` which seems to do the job.

Note that this also affects the formatter, as it uses `fs::gleam_files`. It will now format `.gitignored` files as well.

compiler-cli/src/format.rs
```diff
@@ -79,7 +79,7 @@ pub fn unformatted_files(files: Vec<String>) -> Result<Vec<Unformatted>> {
         })?;
 
         if path.is_dir() {
-            for path in crate::fs::gleam_files_excluding_gitignore(&path) {
+            for path in crate::fs::gleam_files(&path) {
```
PgBiel, line 82: Just to make you aware: the formatter will now format `.gitignored` files as well (although not the build directory). Let me know if this should be changed.

compiler-cli/src/publish.rs
lpil: Ah! Thank you!

lpil (approved):
Wonderful, thank you!!! Would you mind updating the changelog?

lpil:
I will do the changelog!

PgBiel:
> I will do the changelog!

Thanks! Sorry for the small delay. I've rebased it now to erase conflicts.

PR #4258, opened 2025-02-18
Use `ProjectPaths` instead of the current directory

giacomocavalieri:
In this PR I've changed all gleam commands that need to access project's files to use `ProjectPaths` passed in as an argument. This means that the `ProjectPaths` only gets built once per command and reused everywhere, instead of building it from scratch in multiple places by looking at the current working directory.
This also means it will be easier to test each command, being able to feed it the path where a test project is located, instead of relying on the current working directory.

### DIFF
--- a/compiler-cli/src/add.rs
+++ b/compiler-cli/src/add.rs
@@ -2,6 +2,7 @@ use camino::{Utf8Path, Utf8PathBuf};
 
 use gleam_core::{
     error::{FileIoAction, FileKind},
+    paths::ProjectPaths,
     Error, Result,
 };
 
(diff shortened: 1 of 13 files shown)


compiler-cli/src/add.rs
```diff
@@ -22,16 +21,16 @@ pub fn command(packages_to_add: Vec<String>, dev: bool) -> Result<()> {
...
         Some((new_package_requirements.clone(), dev)),
         Vec::new(),
         UseManifest::Yes,
     )?;
 
     // Read gleam.toml and manifest.toml so we can insert new deps into it
-    let mut gleam_toml = read_toml_edit("gleam.toml")?;
-    let mut manifest_toml = read_toml_edit("manifest.toml")?;
+    let mut gleam_toml = read_toml_edit(paths.root_config().as_str())?;
+    let mut manifest_toml = read_toml_edit(paths.manifest().as_str())?;
```
lpil, line 33: Keep them as paths please

compiler-cli/src/add.rs
```diff
@@ -74,8 +73,14 @@ pub fn command(packages_to_add: Vec<String>, dev: bool) -> Result<()> {
...
 
     // Write the updated config
-    fs::write(Utf8Path::new("gleam.toml"), &gleam_toml.to_string())?;
-    fs::write(Utf8Path::new("manifest.toml"), &manifest_toml.to_string())?;
+    fs::write(
+        Utf8Path::new(paths.root_config().as_str()),
+        &gleam_toml.to_string(),
+    )?;
+    fs::write(
+        Utf8Path::new(paths.manifest().as_str()),
```
lpil, line 81: These convert from a path to a string and then back to a path!

lpil (approved):
Thank you!!

PR #4259, opened 2025-02-18
introduce `Command` struct and split `run` to make it easier to test

giacomocavalieri:
To make it easier to test the `run` command I've split it into a `setup` function separate from the command running: so we will be able to reuse that command in tests with a different project runner (I'll need this in the echo PR to write integration tests to capture a project's output once it's run)


lpil (approved):
💖

PR #4261, opened 2025-02-18
Add missing comma

GearsDatapacks:
I noticed when using the "Generate JSON Encoder" code action that I had missed a comma and the syntax was invalid. This PR fixes that small error


lpil (approved):
Thank you

PR #4262, opened 2025-02-18
Fix pattern match on arg

giacomocavalieri:
This PR fixes a bug with the "pattern match on argument/value" code actions that prevented them from pattern matching on private types in the same module they're defined in.


lpil (approved):
Thank you!

PR #4265, opened 2025-02-19
Friendly error when using multiple spreads

carlbordum:
This was unexpected to me, so figured other new gleamers could trip over it as well

(and an excuse to play with the compiler :))

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -157,6 +157,14 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                     "See: https://tour.gleam.run/basics/lists/".into(),
                 ],
             ),
+            ParseErrorType::ListSpreadWithAnotherSpread => (
+                "I wasn't expecting a spread here",
+                vec![
+                    "Lists are immutable and singly-linked, so to join two or more lists".into(),
+                    "all the elements of the lists would need to be copied into a new list.".into(),
+                    "This would be slow, so there is no built-in syntax for it.".into(),
+                ],
+            ),
             ParseErrorType::ListSpreadFollowedByElements => (
                 "I wasn't expecting elements after this",
                 vec![
@@ -375,6 +383,7 @@ pub enum ParseErrorType {
     UnknownTarget, // an unknown target was used
     ListSpreadWithoutElements, // Pointless spread: `[..xs]`
     ListSpreadFollowedByElements, // trying to append something after the spread: `[..xs, x]`
+    ListSpreadWithAnotherSpread, // trying to use multiple spreads: `[..xs, ..ys]`
     LowcaseBooleanPattern, // most likely user meant True or False in patterns
     UnexpectedLabel, // argument labels were provided, but are not supported in this context
     UnexpectedEof,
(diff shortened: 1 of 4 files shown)


lpil (commented):
Could you add a snapshot test for the error please so we can see what it looks like 🙏

compiler-core/src/parse/error.rs
```diff
@@ -157,6 +157,13 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                     "See: https://tour.gleam.run/basics/lists/".into(),
                 ],
             ),
+            ParseErrorType::ListSpreadWithAnotherSpread => (
+                "I wasn't expecting a spread here",
+                vec![
+                    "Hint: use list.append(list1, list2) to join two lists.".into(),
```
lpil, line 163: Not sure we want to recommend this, it's normally the wrong thing to do

carlbordum: Yea, I guess it is likely that the problem is modeled poorly. Perhaps we can add that as a comment, or suggest a common "pattern" to model such problems better.

On the other hand, you could argue that it is still helpful for newcomers. You have to type `[..xs, ..ys]` to get the recommendation, so I think it is pretty clearly what you are trying to do.

What do you think?

compiler-core/src/parse/snapshots/gleam_core__parse__tests__list_spread_followed_by_another_spread.snap
```diff
@@ -0,0 +1,22 @@
...
+}
+
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:5:13
+  │
+5 │   [..xs, 4, ..ys]
+  │             ^^ I wasn't expecting a spread here
+
```
giacomocavalieri, line 20: This is a bit confusing to me, the spread there is technically correct, the problem is there's a prior spread that shouldn't have been there. Maybe it could be pointing to the one that is not last and highlight how there can only be one at the end of the list?

carlbordum: I've moved it so it points the the first spread

compiler-core/src/parse/error.rs
carlbordum: What do you think of the new message?

lpil:
Is this ready for review? Press the button if so!

compiler-core/src/parse/snapshots/gleam_core__parse__tests__list_spread_followed_by_another_spread.snap
```diff
@@ -0,0 +1,23 @@
...
+  [..xs, 4, ..ys]
+}
+
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:5:4
+  │
+5 │   [..xs, 4, ..ys]
+  │    ^^ I wasn't expecting a spread here
```
lpil, line 19: In this case the new error shouldn't be returned, instead it should be the existing error for there being elements after a spread. This error should only happen when there's 2 spreads but no elements between them.

lpil (commented):
Thank you. I've left a note inline

lpil:
Closing this due to inactivity, but please feel free to reopen.

PR #4266, opened 2025-02-19
Increase version to 1.9.0-pre so nightlies support `gleam = ">= 1.9"`

richard-viney:
Not sure if this change aligns with the version bumping approach, but it allows https://github.com/gleam-lang/stdlib/pull/761 to specify `gleam = ">= 1.9"` when run against nightly.


lpil:
Thank you, but let's wait for the release to be out instead.

PR #4267, opened 2025-02-20
Make sure to pick a good name for ignored function args in generated doc

### LINKED ISSUE #4250: Strange docs for `_` function arguments
When a function has an argument that is ignored like this: `fn absurd(_: Void) -> a` the generated documentation will look like this:

Image

With no name before the type annotation. I would change the doc generation so that there's always something  before the type.

giacomocavalieri:
This PR closes #4250


lpil (approved):
Thank you!!

PR #4270, opened 2025-02-20
Update to Rust 1.85 and Rust Edition 2024

GearsDatapacks:
This PR is quite big as it's a lot of automated style and formatting changes


GearsDatapacks:
Weird. Seems like a couple of the checks are running on rust 1.84.1 and not 1.85

GearsDatapacks:
Aha! Looks like checks are finally passing. This is ready for review now

lpil (approved):
Thank you!!!

That removal of `else if let` is a bit unfortunately for the amount of nesting in the parser. Maybe we'll revisit that at some point.

Issue #3253, opened 2024-06-08
Pattern matching problem involving improper scoping


jtenner:
## To Reproduce

```rs
let assert #([d, ..], <<_:size(d)>>) = #([2, 3], <<2:2>>)
```
## System Specs

Runtime: `erlang`
OS: `ubuntu` using `wsl` on `Windows 10`
Gleam version output:
```
$ gleam --version
gleam 1.2.1
```

## The Problem

Identifiers are placed in the scope of each pattern, before they are initialized in the match arm. 

Is there any way I can help?

lpil:
Thank you

lpil:
Wow, OK! So I've experimented with this and it seems that you can only used variables declared inside the same pattern _if they are declared in the bit string itself_. I'm very surprised by this!

GearsDatapacks:
To my understanding, this is a limitation of erlang. So should this emit a compiler error when trying to compile?

lpil:
Yes, I think so. We will have to make the variables in the pattern not register immediately, and have a special case so that variables defined within a bit array can be referenced later in the bit array

GearsDatapacks:
Doing some testing of my own, I can't seem to get it to use variables from the pattern, even when defined in the bit array itself. Am I doing something wrong maybe? I don't have much experience with bit arrays or erlang.
Here is the code I am testing:
```gleam
let assert <<a, b:size(a)>> = <<1, 2:size(1)>>
```

lpil:
What problem are you having? This compiles

```erlang
-module(a).
-export([main/0]).

main() ->
    <<A, B:A>> = <<1, 2:1>>,
    erlang:display(A),
    erlang:display(B).
```

GearsDatapacks:
That's strange. On my machine I get:
```
<file path>: variable 'A' is unbound
%   10|         <<_, _:A>> -> _assert_subject;
%     |                ^
```

lpil:
That's not the same code as I've got there. Could you share all of it please?

GearsDatapacks:
That's all the code I have, if wrapped in a main function:
```gleam
pub fn main() {
  let assert <<a, b:size(a)>> = <<1, 2:size(1)>>
}
```
That's my whole project

lpil:
To understand what this Gleam code does at runtime you will need to view the compiled Erlang.

GearsDatapacks:
It compiles to this for me:
```erlang
-module(testing_gleam).
[...]

PR #4273, opened 2025-02-21
Implement variable sized bit array pattern matching on JavaScript

GearsDatapacks:
This PR implements the ability to pattern match on dynamically sized bit array segments on JavaScript.

### DIFF
--- a/compiler-core/src/javascript/pattern.rs
+++ b/compiler-core/src/javascript/pattern.rs
@@ -99,7 +99,7 @@ impl<'a> OffsetBits<'a> {
         self
     }
 
-    fn to_doc(self) -> Document<'a> {
+    fn into_doc(self) -> Document<'a> {
         let doc = if self.variables.is_empty() {
             self.constant.to_doc()
         } else if self.constant == 0 {
@@ -258,7 +258,7 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
                 Index::Int(i) => acc.append(eco_format!("[{i}]").to_doc()),
                 // TODO: escape string if needed
                 Index::String(s) => acc.append(docvec![".", maybe_escape_property_doc(s)]),
-                Index::ByteAt(i) => acc.append(docvec![".byteAt(", i.clone(), ")"]),
+                Index::ByteAt(i) => acc.append(docvec![".byteAt(", i.clone().into_doc(), ")"]),
                 Index::BitArraySliceToInt {
                     start,
                     end,
@@ -268,9 +268,9 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
                     "bitArraySliceToInt(",
                     acc,
                     ", ",
-                    start.clone(),
+                    start.clone().into_doc(),
                     ", ",
-                    end.clone(),
+                    end.clone().into_doc(),
                     ", ",
                     bool(endianness.is_big()),
                     ", ",
@@ -285,9 +285,9 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
                     "bitArraySliceToFloat(",
                     acc,
                     ", ",
-                    start.clone(),
+                    start.clone().into_doc(),
                     ", ",
-                    end.clone(),
+                    end.clone().into_doc(),
                     ", ",
                     bool(endianness.is_big()),
                     ")"
@@ -298,13 +298,13 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
                             "bitArraySlice(",
                             acc,
                             ", ",
-                            start.clone(),
+                            start.clone().into_doc(),
                             ", ",
-                            end.clone(),
+                            end.clone().into_doc(),
                             ")"
                         ]
                     }
-                    None => docvec!["bitArraySlice(", acc, ", ", start.clone(), ")"],
+                    None => docvec!["bitArraySlice(", acc, ", ", start.clone().into_doc(), ")"],
                 },
                 Index::StringPrefixSlice(i) => docvec!(acc, ".slice(", i, ")"),
             })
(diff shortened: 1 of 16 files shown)


compiler-core/src/javascript/pattern.rs
```diff
@@ -50,32 +50,104 @@ pub enum BitArrayTailSpreadType {
...
 }
 
-struct Offset {
-    bits: usize,
+#[derive(Debug, Clone)]
+pub struct OffsetBits<'a> {
+    constant: usize,
+    variables: Vec<Document<'a>>,
+    extra: Option<EcoString>,
+}
```
lpil, line 58: Could you document this struct? I don't understand it and its properties.

How come `variables` is a document and `extra` is a string? It seems like `extra` is under-specified and `variables` would better be `&str` or `EcoString`.

compiler-core/src/javascript/pattern.rs
```diff
@@ -50,32 +50,104 @@ pub enum BitArrayTailSpreadType {
...
+        }
+    }
+
+    fn is_whole_number_of_bytes(&self) -> bool {
+        self.variables.is_empty() && self.extra.is_none() && self.constant % 8 == 0
+    }
+
+    fn divide(mut self, by: usize) -> Self {
+        if self.variables.is_empty() {
+            self.constant /= by;
```
lpil, line 74: Don't use `=/` please, it's uncommon and easy to misread.

compiler-core/src/javascript/pattern.rs
```diff
@@ -50,32 +50,104 @@ pub enum BitArrayTailSpreadType {
...
+        } else {
+            self.extra = Some(eco_format!(" / {by}"));
+        }
+
+        self
+    }
+}
+
+impl<'a> Documentable<'a> for OffsetBits<'a> {
```
lpil, line 83: Never use traits 🙏  Use regular functions please, it is much clearer and compiles faster. It would also mean that the cloning of this struct could be removed I think.

compiler-core/src/javascript/pattern.rs
```diff
@@ -617,43 +704,53 @@ impl<'module_ctx, 'expression_gen, 'a> Generator<'module_ctx, 'expression_gen, '
...
+                                    .size
+                                    .is_literal_and(|size| size <= SAFE_INT_SEGMENT_MAX_SIZE)
+                                    && details.size.is_literal_and(|size| size % 8 == 0)
+                                    && offset.bits.is_whole_number_of_bytes() =>
                             {
+                                let size = match details.size {
+                                    BitArraySize::Literal(size) => size,
+                                    BitArraySize::Variable(_) => {
+                                        unreachable!("We already checked that it's a literal")
+                                    }
```
lpil, line 721: We try to avoid avoid this non-functional pattern of checking that something is true and then asserting it afterwards. In this case it is being check 3 times, really it should only be done once.

compiler-core/src/javascript/pattern.rs
```diff
@@ -50,32 +50,104 @@ pub enum BitArrayTailSpreadType {
...
+
+    fn is_whole_number_of_bytes(&self) -> bool {
+        self.variables.is_empty() && self.extra.is_none() && self.constant % 8 == 0
+    }
+
+    fn divide(mut self, by: usize) -> Self {
+        if self.variables.is_empty() {
+            self.constant /= by;
+        } else {
+            self.extra = Some(eco_format!(" / {by}"));
```
lpil, line 76: This seems to be the only time that `extra` is used, so it seems underspecified for it to be a string. This method also overwrites the previous division in one state, but it divides again in the other state. The behaviour should always be the same.

compiler-core/src/javascript/pattern.rs
```diff
@@ -50,32 +50,104 @@ pub enum BitArrayTailSpreadType {
...
+                    " + ".to_doc()
+                ),
+                " + ",
+                self.constant,
+            ]
+            .group()
+        };
+
+        match self.extra {
+            Some(extra) => doc.append(extra).group(),
```
lpil, line 106: Doesn't this do the wrong then when there are multiple variables and also an extra?

`x + y / z` being `x + (y / z)` rather than `(x + y) / z`

lpil (commented):
Wonderful, thank you!!!

I've left a bunch of little nit-pick notes inline, and it would be fab to remove all the clones.

I couldn't see any tests that involve extra, and there's no integration tests to check that that the runtime behaviour is correct. Here's an example bit array PR's integration tests: https://github.com/gleam-lang/gleam/pull/3946/files#diff-42aefabdf30f6176dcac76f43f95e6af038909579ce30683824c7fc2d98c3db9

Thanks again!

compiler-core/src/javascript/pattern.rs
GearsDatapacks: I believe it produces `((x + y) / z)`, because `doc` has already been `group`ed

compiler-core/src/javascript/pattern.rs
GearsDatapacks: Struct has now been documented an `extra` has been reworked.
`variables` is a `Vec<Document>` because we use the `ExpressionGenerator::local_var` function to obtain the correct name to use to reference the variable, and that returns a `Document`. I'm not sure what the best way to approach this is

compiler-core/src/javascript/pattern.rs
GearsDatapacks: The linter complains if I don't do it this way. Should we disable that lint?

compiler-core/src/javascript/pattern.rs
GearsDatapacks: From what I can tell, if we don't clone it we run into nasty lifetime issues which are tricky to resolve

lpil (approved):
Wonderful work, as always! Thank you!!

GearsDatapacks:
What should I do about the linter errors? Should I use the shorthand `*=` and `/=` or should I disable that lint?

GearsDatapacks:
Ah, looks like tests are failing due to #3253. I guess I just have to work around that for now?

lpil:
Reworking the test to avoid that bug could be good

GearsDatapacks:
Oops, forgot 928 doesn't fit within 8 bits

GearsDatapacks:
Ok sorry this has been a bit of a mess. Should be ready to auto-merge now

GearsDatapacks:
I've decided to follow your comment and replace the usage of `Document` in `variables` with `EcoString`. It makes it a bit easier to work with and removes most of the cloning. (The only cloning now is of `EcoString`s, which is cheap)

lpil (approved):
Beautiful!! Thank you!

PR #4274, opened 2025-02-21
Add code action to do string interpolation

### LINKED ISSUE #4260: Suggestion: Code action for string interpolation
Since we don't have string interpolation and must concatenate strings with <>. What if we added a code action that can be triggered while inside of a string and would put " <> | <> " at the cursor location, with | being the cursor ready to type the variable?

The code action could be called something like `insert variable` or `concat variable`.

(cc @giacomocavalieri, from discord dms :))

giacomocavalieri:
This PR fixes #4260 
And this is what it looks like in practice:


compiler-core/src/language_server/code_action.rs
```diff
@@ -4906,3 +4906,126 @@ impl<'ast> ast::visit::Visit<'ast> for ConvertToPipe<'ast> {
...
+            StringInterpolation::InterpolateValue { value_location } => {
+                let value_range = value_location.start as usize..value_location.end as usize;
+                let name = self
+                    .module
+                    .code
+                    .get(value_range)
+                    .expect("invalid value range");
+
+                if !is_valid_lowercase_name(name) {
+                    return vec![];
```
lpil, line 4978: It should still work in this case, just without the name.

lpil (commented):
Thank you! One small note inline

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Ok, fixed it!

lpil (commented):
Please call it interpolate in both cases! String means what `string.split` does, interpolate means to put something in the middle of text, which is what this does.

giacomocavalieri:
Oh ok! I'm fixing that (and also fixing a small bug I just noticed, might take a couple of minutes 😁)

PR #4275, opened 2025-02-21
Simplify CI Release Process

maennchen:
* Preparation for #4276
* Unify wasm / native release generation
* Unify nightly / version release generation
* The result should be identical to the current CI
* Example CI / Releases
  * Tagged CI (push) https://github.com/maennchen/gleam/actions/runs/13813751330
  * Tagged CI (release) https://github.com/maennchen/gleam/actions/runs/13813919061
  * Tagged Release https://github.com/maennchen/gleam/releases/tag/vSimplify
  * Nightly CI: https://github.com/maennchen/gleam/actions/runs/13813756622
  * Nightly Release: https://github.com/maennchen/gleam/releases/tag/nightly

### DIFF
--- a/.github/actions/build-release/action.yml
+++ b/.github/actions/build-release/action.yml
@@ -1,0 +1,143 @@
+name: "Build Gleam"
+description: "Build Gleam Release"
+inputs:
+  version:
+    description: "Build Version"
+    required: true
+  toolchain:
+    description: "Cargo Toolchain"
+    required: true
+  target:
+    description: "Cargo Installation Target"
+    required: true
+  cargo-tool:
+    description: "Cargo Tool used for Build (for example, `cross`)"
+    required: true
+  expected-binary-architecture:
+    description: "Expected Binary Architecture"
+    required: false
+    default: ""
+
+outputs:
+  archive:
+    description: "Path to build asset"
+    value: "${{ steps.build.outputs.archive }}"
+  files:
+    description: "Path to all files"
+    value: |
+      ${{ steps.build.outputs.archive }}
+      ${{ steps.build.outputs.archive }}.sha256
+      ${{ steps.build.outputs.archive }}.sha512
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install Rust toolchain
+      uses: actions-rust-lang/setup-rust-toolchain@v1
+      with:
+        toolchain: ${{ inputs.toolchain }}
+        target: ${{ inputs.target }}
+        cache-key: v1-${{ inputs.target }}
+
+    - name: Build WASM release binary
+      if: ${{ inputs.target != 'wasm32-unknown-unknown' }}
+      uses: clechasseur/rs-cargo@v3
+      with:
+        command: build
+        args: --release --target ${{ inputs.target }}
+        tool: ${{ inputs.cargo-tool }}
+
+    - name: Install wasm-pack
+      if: ${{ inputs.target == 'wasm32-unknown-unknown' }}
+      shell: bash
+      run: curl -sSL https://rustwasm.github.io/wasm-pack/installer/init.sh | sh
+
+    - name: Build WASM release binary
+      if: ${{ inputs.target == 'wasm32-unknown-unknown' }}
+      shell: bash
+      run: wasm-pack build --release --target web compiler-wasm
+
+    - name: Verify binary architecture
+      if: ${{ inputs.expected-binary-architecture }}
+      shell: bash
+      run: |
+        BINARY_PATH="target/${{ inputs.target }}/release/gleam"
+        if [[ "${{ inputs.target }}" == *"windows"* ]]; then
+          BINARY_PATH="${BINARY_PATH}.exe"
+        fi
+
+        if ! file -b "$BINARY_PATH" | grep -q "${{ inputs.expected-binary-architecture }}"; then
+          echo "error: Architecture mismatch"
+          echo "Expected architecture: '${{ inputs.expected-binary-architecture }}'"
+          echo "Found binary type: '$(file -b "$BINARY_PATH")'"
+          exit 1
+        fi
+        echo "ok: Architecture match"
+
+    - name: Build archive
+      id: build
+      shell: bash
+      run: |
+        case "$TARGET" in
+          *windows*)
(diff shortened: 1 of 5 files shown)


.github/actions/build-release/action.yml
```diff
@@ -0,0 +1,133 @@
...
+    description: "Build Version"
+    required: true
+  toolchain:
+    description: "Cargo Toolchain"
+    required: true
+  target:
+    description: "Cargo Installation Target"
+    required: true
+  cargo-tool:
+    description: "Cargo Tool" # TODO: What is this? describe...
```
maennchen, line 14: TODO: Put a better description

.github/actions/build-release/action.yml
```diff
@@ -0,0 +1,133 @@
...
+    required: true
+  binary:
+    description: "Binary" # TODO: What is this? describe...
```
maennchen, line 17: TODO: Put a better description

maennchen:
@lpil There’s one TODO open, where I could use some input / insights.

Besides that, yes.

.github/actions/build-release/action.yml
lpil: This is the expected binary architecture, used to verify the binary has been correctly produced. `expected-binary-architecture` or such would be a clearer name

.github/workflows/release.yaml
```diff
@@ -3,12 +3,52 @@ on:
...
+    - cron: "45 0 * * *"
 
 env:
   CARGO_TERM_COLOR: always
   RUSTFLAGS: "-D warnings"
+  VERSION: "${{ github.event_name == 'push' && github.ref_name || 'nightly' }}"
+
+permissions:
+  contents: write
+  id-token: write
```
lpil, line 17: Why are these two permissions here? I don't feel very comfortable giving a workflow commit or OIDC permissions.

.github/workflows/release.yaml
```diff
@@ -3,12 +3,52 @@ on:
...
+  packages: write
 
 jobs:
+  prepare-nightly:
```
lpil, line 21: Could you add a comment explaining what this job is for please 🙏

.github/workflows/release.yaml
```diff
@@ -124,50 +113,103 @@ jobs:
...
+          - scratch
+          - erlang
+          - erlang-slim
+          - erlang-alpine
+          - elixir
+          - elixir-slim
+          - elixir-alpine
+          - node
+          - node-slim
+          - node-alpine
```
lpil, line 142: Currently containers are not created until after the release is published, but if I'm reading this right the containers would be published _before_ the release is, and would still be released even if the release was rejected.

.github/workflows/release.yaml
```diff
@@ -124,50 +113,103 @@ jobs:
...
+      - name: Download Gleam release from previous job
+        uses: actions/download-artifact@v4
         with:
-          toolchain: stable
-          target: wasm32-unknown-unknown
-          profile: minimal
-          override: true
+          pattern: release-{x86_64-unknown-linux-musl,aarch64-unknown-linux-musl}
+
+      - name: "Unpack and rearrange release"
```
lpil, line 155: What does this do?

lpil (commented):
Very nice! Thank you so much for doing this. I've left some questions inline.

Is there a repo where I can see it successfully running and test the output?

.github/workflows/release.yaml
maennchen: `id-token` was mistakenly added from the SBOM PR. I'll remove it.
`contents: write` is necessary to change release assets. It was already given before, just not explicit.

.github/workflows/release.yaml
maennchen: Added.

.github/workflows/release.yaml
maennchen: Added an explanation. The approach might look a bit complicated, but makes more sense when additional files are used as with the SBoM PR.

.github/workflows/release.yaml
lpil: That's still not the behaviour from before unfortunately. When the release workflow finishes it leaves a GH release in draft mode. I can then check this release, the assets, and anything else before publishing it, and publication then results in the container images being built and published. With this new workflow the container images would be published even if I reject the release and do not publish it.

maennchen:
@lpil 

> Is there a repo where I can see it successfully running and test the output?

Yes:

* Example Release (`vjm20`): https://github.com/maennchen/gleam/releases/tag/untagged-b42e320abb87e6969c11
* Example CI: https://github.com/maennchen/gleam/actions/runs/13496871664

.github/workflows/release.yaml
maennchen: Ah, I misunderstood how that worked before. Yes, it now works differently.
I'll take the PR back into draft state and fix this.

lpil:
Is this ready for a review? Undraft if so! 🙏

maennchen:
@lpil I haven’t gotten around to it yet. Probably later this week or start of the next.

maennchen:
@lpil Ready for review again. I linked the CI runs & Releases in the PR body.

.github/workflows/release-containers.yaml
```diff
@@ -4,10 +4,15 @@ on:
     types:
       - "published"
 
+permissions:
+  packages: write
+
 jobs:
-  publish-container-images:
-    name: publish-container-images
+  build-container-images:
```
lpil, line 11: Could we revert to the previous name of publish-container-images please, and use that for the action too? The important outcome is that it publishes images, so I want that to be immediately obvious. And as a pedantic nit it builds images, not containers.

lpil (commented):
Wonderful, thank you! This looks great to me. I've left a tiny naming related note, and then we can merge I think

.github/workflows/release-containers.yaml
maennchen: That was left over form my wrong approach before. Changed back to the original name :)

maennchen:
@lpil No worries, I have lots of other things going on as well :)

lpil:
Hi @maennchen ! The nightly build failed: <https://github.com/gleam-lang/gleam/actions/runs/13982471191/job/39150515439>

PR #4282, opened 2025-02-26
Implement renaming for Functions, Constants and Type Variants

### LINKED ISSUE #3259: LSP: rename custom type variant
I'd like to rename one of the types and be able to refactor the whole app.

I currently am receiving a `method textDocument/rename is not supported` on Neovim.

GearsDatapacks:
This PR closes #3259 
This is pretty much ready, except for two things:
- It's missing tests
- There's an annoying bug to do with module shadowing which I still have to figure out how to resolve

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -900,20 +900,21 @@ impl TypedDefinition {
 
             Definition::CustomType(custom) => {
                 // Check if location is within the type of one of the arguments of a constructor.
-                if let Some(annotation) = custom
+                if let Some(constructor) = custom
                     .constructors
                     .iter()
                     .find(|constructor| constructor.location.contains(byte_index))
-                    .and_then(|constructor| {
-                        constructor
-                            .arguments
-                            .iter()
-                            .find(|arg| arg.location.contains(byte_index))
-                    })
-                    .filter(|arg| arg.location.contains(byte_index))
-                    .and_then(|arg| arg.ast.find_node(byte_index, arg.type_.clone()))
                 {
-                    return Some(annotation);
+                    if let Some(annotation) = constructor
+                        .arguments
+                        .iter()
+                        .find(|arg| arg.location.contains(byte_index))
+                        .and_then(|arg| arg.ast.find_node(byte_index, arg.type_.clone()))
+                    {
+                        return Some(annotation);
+                    }
+
+                    return Some(Located::RecordConstructor(constructor));
                 }
 
                 // Note that the custom type `.location` covers the function
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -20,6 +20,7 @@ use crate::type_::printer::Names;
 use crate::type_::{
     self, Deprecation, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
 };
+use std::collections::HashMap;
 use std::sync::Arc;
 
 use ecow::EcoString;
@@ -50,6 +51,7 @@ pub struct Module<Info, Statements> {
     pub type_info: Info,
     pub definitions: Vec<Statements>,
     pub names: Names,
+    pub references: HashMap<EcoString, Vec<SrcSpan>>,
 }
 
 impl TypedModule {
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1830,6 +1830,7 @@ pub enum Pattern<Type> {
     /// The constructor for a custom type. Starts with an uppercase letter.
     Constructor {
         location: SrcSpan,
+        name_location: SrcSpan,
         name: EcoString,
         arguments: Vec<CallArg<Self>>,
         module: Option<(EcoString, SrcSpan)>,
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -13,6 +13,7 @@ pub use self::constant::{Constant, TypedConstant, UntypedConstant};
 
 use crate::analyse::Inferred;
 use crate::build::{Located, Target};
+use crate::call_graph::ReferenceInformation;
 use crate::parse::SpannedString;
 use crate::type_::error::VariableOrigin;
 use crate::type_::expression::Implementations;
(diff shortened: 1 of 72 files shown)


GearsDatapacks:
Ok, this is ready for review now. I'm not super happy with the way I fixed the module access shadowing bug, but I couldn't think of another way to do it.

compiler-core/src/ast.rs
```diff
@@ -1064,6 +1073,8 @@ impl<A, B, C, E> Definition<A, B, C, E> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct UnqualifiedImport {
     pub location: SrcSpan,
+    /// The location excluding the potential `as ...` clause, or the `type` keyword
+    pub imported_name_location: SrcSpan,
```
lpil, line 1077: I think the start here is redundant as it's the same as `location`, it could be a single position int instead.

compiler-core/src/ast.rs
```diff
@@ -879,20 +887,21 @@ impl TypedDefinition {
...
+                    if let Some(annotation) = constructor
+                        .arguments
+                        .iter()
+                        .find(|arg| arg.location.contains(byte_index))
+                        .and_then(|arg| arg.ast.find_node(byte_index, arg.type_.clone()))
+                    {
+                        return Some(annotation);
+                    }
+
+                    return Some(Located::RecordConstructor(constructor));
```
lpil, line 904: It's a variant constructor, it may not always be a record as it could have no fields.

Could we also make it clear in the name that this is the definition rather than a usage 🙏

compiler-core/src/language_server/rename.rs
```diff
@@ -65,6 +65,102 @@ pub fn rename_local_variable(
...
 
+pub fn rename_module_value(
+    params: &RenameParams,
+    module_name: &EcoString,
+    name: &EcoString,
+    modules: &HashMap<EcoString, Module>,
+    name_kind: Named,
+) -> Option<WorkspaceEdit> {
+    if name::check_name_case(
+        Default::default(),
```
lpil, line 76: Could you use the default impl on the type that is being constructed rather than the method please? It makes it easier to understand what is being made when it is explicit.

compiler-core/src/language_server/rename.rs
```diff
@@ -65,6 +65,102 @@ pub fn rename_local_variable(
...
+        return None;
+    }
+
+    let mut workspace_edit = WorkspaceEdit {
+        changes: Some(HashMap::new()),
+        document_changes: None,
+        change_annotations: None,
+    };
+
+    for module in modules.values() {
```
lpil, line 91: Perhaps we should have an import graph so we can get these modules directly rather than traversing everything? Probably doesn't matter. Something for later perhaps.

That aside, this is a hashmap of modules that have been compiled, it does not include any that were loaded from the cache. This means there will be no renaming in any module that have not been recompiled in that editor session!

compiler-core/src/ast.rs
```diff
@@ -66,6 +68,12 @@ impl TypedModule {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub struct References {
+    pub imported_modules: HashSet<EcoString>,
+    pub value_references: HashMap<(EcoString, EcoString), Vec<SrcSpan>>,
```
lpil, line 74: It seems like it may be challenging to identify unused code with this data structure, and replacing the current broken usage tracking system is one of the main goals of the call graph work. Did you have thoughts on how that'd be done?

lpil (commented):
Super nice!! I'm really excited about this one.

I've not finished reviewing totally but I've gotta go so I wanted to drop my thoughts so far. See inline notes.

compiler-core/src/language_server/rename.rs
GearsDatapacks: Yeah that makes sense. I guess that means we need to store reference information in the module interface and cache it then. 
As for an import graph, I'm happy to use one if you want; not exactly sure what that data structure would look like, or how we would build it and where it would be stored.

compiler-core/src/ast.rs
GearsDatapacks: That's true, I wasn't thinking too much about the current usage tracking. We already generate an in-module call-graph for dependency ordering, so would it be possible to use just that for usage tracking? We can't use that for renaming because it doesn't track cross-module usage.

compiler-core/src/ast.rs
GearsDatapacks: Actually in the following case: `import some_module.{type SomeType as SomeAlias}`, the full location contains `type SomeType as SomeAlias`, whereas the `imported_name_location` only contains `SomeType`. Therefore, it is completely separate information to the full location.

GearsDatapacks:
Ok, I have now added the reference information to the cache so we can rename values in cached modules.
It feels a bit wasteful, it's quite a lot of information which is only needed for the LSP, but I don't think there's any other way around it.
I have responded to a couple of your comments, let me know your thoughts.

compiler-core/src/ast.rs
lpil: That one only tracks calls to module functions, which isn't enough for usage detection.

Usage detection, rename, and show-references all require the same thing: knowing all the references for each entity in the code. Usage detection uses this information differently from the other two though, it traverses the graph from all the public entities to find any unreachable nodes.

GearsDatapacks: Ok that makes sense, thanks.
I'll look into it

GearsDatapacks:
This should be ready for review again.
As discussed on Discord, I couldn't find a way to easily represent the information needed for renaming in a call-graph structure. Instead, I've added a `ReferenceTracker` struct, which keeps track of all the reference locations, but also builds a call graph which can later be used (in a future PR) to do usage tracking.
I have also changed the functionality so that renaming a value that is imported in an unqualified fashion now aliases it instead of globally renaming, and prevented users from renaming values from external packages.

lpil (approved):
This is amazing! Thank you!!!

PR #4284, opened 2025-02-27
Remove hole from function captures with a single hole

### LINKED ISSUE #4279: Reformat single argument captures into function references
I don't think there are any ambiguities here, but let me know if I'm wrong:
```gleam
list.map(x, io.debug(_))
// Becomes:
list.map(x, io.debug)
```

### LINKED ISSUE #4297: Show hint to import unqualified types/values
For example:
```gleam
import gleam/dict

type Grid(a) {
  Grid(Dict(#(Int, Int), a))
}
```

giacomocavalieri:
This PR fixes #4279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,6 +230,9 @@
 
 ### Formatter
 
+- Changed the formatting of function captures with a single hole to remove it.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Bug fixes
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -230,6 +230,9 @@
 
 ### Formatter
 
+- Changed the formatting of function captures with a single hole to remove it.
```
lpil, line 233: That's all function captures!

lpil (approved):
Thank you!

PR #4293, opened 2025-03-01
Implement LSP Hover for module names

### LINKED ISSUE #3451: LSP: Show module docs when hovering a module
I use the "View on hexdocs" link all the time but it doesn't work on modules 🙃 

```gleam
// show docs on hover here would be nice
import  mymodule



// show docs on hover my module here would be nice
mymodule.do_something()
```

There is bigger fish to fry, but still :)

GearsDatapacks:
Closes #3451 
This PR implements the ability to hover over a module name and see the documentation for it.
I've had to edit a couple of existing tests to account for the change in behaviour when hovering the module part of a module select.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -228,6 +228,10 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The Language Server now shows module documentation when hovering over a module
+  name.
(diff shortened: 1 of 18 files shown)


lpil (approved):
Wonderful!! Thank you very much!

Issue #4294, opened 2025-03-01
String pattern matching which doesn't work when target="javascript"


anolivetree:
Hi,
Following code works on erlang but not when target is "javascript"
Tested on v1.8.1

```
pub fn main() {
  let assert Ok(<<"Hello!\n":utf8>>) = Ok(<<"Hello!\n":utf8>>)
}
```

Compiled .mjs is below. It seems "\n" is not handled as escape sequence.

```
export function main() {
  let $ = new Ok(toBitArray([stringBits("Hello!\n")]));
  if (
    !$.isOk() ||
    $[0].byteAt(0) !== 0x48 ||
    $[0].byteAt(1) !== 0x65 ||
    $[0].byteAt(2) !== 0x6C ||
    $[0].byteAt(3) !== 0x6C ||
    $[0].byteAt(4) !== 0x6F ||
    $[0].byteAt(5) !== 0x21 ||
    $[0].byteAt(6) !== 0x5C ||
    $[0].byteAt(7) !== 0x6E ||
    !($[0].length == 8)
  ) {
    throw makeError(
      "let_assert",
      "strcmp",
      2,
      "main",
      "Pattern match failed, no pattern matched the value.",
      { value: $ }
    )
  }
  return $;
}
```

richard-viney:
This bug occurs when matching against any  string that contains control characters, e.g. `"\n"`, `"\t"`, `"\""`, `"\u{1f600}"`, etc.

I'll look into a fix shortly.

PR #4295, opened 2025-03-01
Fix pattern matching of UTF-8 strings containing escape chars

richard-viney:
Fixes https://github.com/gleam-lang/gleam/issues/4294.


lpil (approved):
Wonderful! Thank you

PR #4298, opened 2025-03-04
Add export package-information command

### LINKED ISSUE #4240: Add command to export deps/config
While implementing [Gleam support in Elixir Mix](https://github.com/elixir-lang/elixir/pull/14262), they asked me to avoid Regexes to parse the TOML config.

So hereby I request a new comand to export package deps in either JSON or erlang terms format (ie. `gleam deps export`).

I am not 100% sure if we'll need anything else from the config file, so exporting the whole thing (and not just the deps) might be desirable.

Papipo:
Closes #4240 

- Adds `Serialize` trait to `PackageConfig` and all types it depends on.
- Adds serde options to skip flags that have default values.
- Adds `gleam export package-information --out [out]` command
- Fixes `gleam export package-interface` format for `gleam_version_constraint`, which was being exported like: `"gleam-version-constraint":"1.8.0 <= v"` because it relied on pubgrub. Now it uses hexpm format and emits `>= 1.8.0`.


compiler-core/src/snapshots/gleam_core__config__package_config_to_json.snap
```diff
@@ -0,0 +1,5 @@
+---
+source: compiler-core/src/config.rs
+expression: json
+---
+{"name":"my_project","version":"1.0.0","gleam":">= 0.30.0","licences":["Apache-2.0","MIT"],"description":"Pretty complex config","documentation":{"pages":[{"title":"My Page","path":"my-page.html","source":"./path/to/my-page.md"}]},"dependencies":{"my_other_project":{"path":"../my_other_project"},"gleam_stdlib":{"version":">= 0.18.0 and < 2.0.0"}},"dev-dependencies":{"gleeunit":{"version":">= 1.0.0 and < 2.0.0"}},"repository":{"type":"github","user":"example","repo":"my_dep"},"links":[{"title":"Home page","href":"https://example.com/"}],"erlang":{"application_start_module":"my_app/application","extra_applications":["inets","ssl"]},"javascript":{"typescript_declarations":true,"runtime":"nodejs","deno":{"allow_env":["DATABASE_URL"],"allow_net":["example.com:443"],"allow_ffi":true,"allow_read":["./database.sqlite"]}},"internal_modules":["my_app/internal"]}
```
lpil, line 5: Please include the toml in the snapshot and pretty print the JSON 🙏  

See the other snapshot for examples of how we format them with both the input and the output

compiler-core/src/config.rs
```diff
@@ -653,66 +700,104 @@ where
...
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "is_default")]
     pub unstable: bool,
-    #[serde(default, deserialize_with = "uri_serde::deserialize_option")]
+    #[serde(
+        default,
+        serialize_with = "uri_serde::serialize_option",
+        deserialize_with = "uri_serde::deserialize_option"
+    )]
+    #[serde(skip_serializing_if = "Option::is_none")]
```
lpil, line 750: Can we merge this with the existing serde attribute?

compiler-core/src/config.rs
```diff
@@ -62,63 +59,94 @@ impl<'de> Deserialize<'de> for SpdxLicense {
...
         default,
-        rename = "gleam",
-        serialize_with = "serialise_range",
-        deserialize_with = "deserialise_range"
+        skip_serializing_if = "Option::is_none",
+        serialize_with = "serialise_hex_range",
+        deserialize_with = "deserialise_hex_range"
     )]
+    // gleam version field in Hex format (for [de]serialization purposes)
+    pub gleam: Option<version::Range>,
```
lpil, line 90: This should be part of the `gleam_version` field rather than a second one. They are tightly coupled as being two representations of the same thing, so it is invalid to have one without the other.

compiler-core/src/build.rs
```diff
@@ -58,6 +59,7 @@ use vec1::Vec1;
 pub enum Target {
     #[strum(serialize = "erlang", serialize = "erl")]
     #[serde(rename = "erlang", alias = "erl")]
+    #[default]
```
lpil, line 62: Remove this please. There is no universal default runtime.

compiler-cli/src/lib.rs
```diff
@@ -314,6 +314,12 @@ pub enum ExportTarget {
         /// The path to write the JSON file to
         output: Utf8PathBuf,
     },
+    /// Package information (gleam.toml) in JSON format
+    PackageInfo {
```
lpil, line 318: Let's use `package-information`, as we avoid abbreviations.

Papipo:
> Let's make it always render all the fields, even if they are the defaults. This will make it easier to understand what all the fields are.

I made that change so the resulting JSON would be more similar to the TOML. I think that reverting https://github.com/gleam-lang/gleam/pull/4298/commits/b6ad8e83887eb38f9bf96b1526123bb2684b3166 would just be what you are asking for, is that it?

Before that change the json was full of empty arrays, nulls, etc.

Papipo:
> It would also be good to have some round-trip tests, to ensure that a config file serialised and then deserialise doesn't have a bug that results in data-loss.

Isn't that covered already by deserialising into `PackageConfig` and then serialising to a JSON snapshot? Or do you mean something else?

Papipo:
@lpil I think all comments have been handled.

 Let me know if I should squash some commits or something.

compiler-core/src/config.rs
lpil: This hasn't been done

compiler-core/src/config.rs
```diff
@@ -62,25 +63,74 @@ impl<'de> Deserialize<'de> for SpdxLicense {
...
+
+        Ok(Self { pubgrub, hex })
+    }
+
+    pub fn lowest_version(&self) -> Option<Version> {
+        self.pubgrub.lowest_version()
+    }
+}
+
+impl fmt::Display for GleamVersion {
```
lpil, line 114: Remove this please, we don't custom implement traits.

lpil (commented):
Thank you, but not all the comments have been resolved. Please remove all the serialisation skipping, the trait use, and merge the duplicate serde attribute.

.github/workflows/ci.yaml
```diff
@@ -215,6 +215,11 @@ jobs:
         working-directory: ./test/project_erlang_windows
         if: ${{ runner.os == 'Windows' && matrix.run-integration-tests }}
 
+      - name: test/project_erlang export package-information
+        run: |
+          gleam export package-information --out="gleam.json"
```
lpil, line 220: Need to `cd` into the package directory here!

Papipo:
> Thank you, but not all the comments have been resolved. Please remove all the serialisation skipping, the trait use, and merge the duplicate serde attribute.

Dammit, I think I had some rebasing issue 😮‍💨

lpil:
Ah I hate it when that happens!

.github/workflows/ci.yaml
Papipo: Fixed in fixup commit

Papipo:
@lpil The duplicate serde attribute was removed rather than merged due to the removal of all skips.

compiler-core/src/config.rs
```diff
@@ -62,25 +63,72 @@ impl<'de> Deserialize<'de> for SpdxLicense {
...
+
+        Ok(Self { pubgrub, hex })
+    }
+
+    pub fn lowest_version(&self) -> Option<Version> {
+        self.pubgrub.lowest_version()
+    }
+
+    pub fn hex(&self) -> version::Range {
+        self.hex.clone()
```
lpil, line 114:
```suggestion
    pub fn hex(&self) -> &version::Range {
        &self.hex
```

lpil (commented):
Oh! The test failed due to ordering, but you've got `preserve_order` as a serde feature. Is that not enough to make the order deterministic or is there more to do there?

Papipo:
> Oh! The test failed due to ordering, but you've got `preserve_order` as a serde feature. Is that not enough to make the order deterministic or is there more to do there?

I thought it was enough as you can expect, sigh. Gonna investigate as it!

compiler-core/src/config.rs
Papipo: I get this:

```
cannot move out of `self.hex` which is behind a shared reference
move occurs because `self.hex` has type `hexpm::version::Range`, which does not implement the `Copy` trait
```

Papipo:
Should be fixed now, it was my fault (obsolete snapshot).

lpil:
Tests are failing!

Cargo.toml
```diff
@@ -23,7 +23,7 @@ regex = "1"
 termcolor = "1"
 # Data (de)serialisation
 serde = { version = "1", features = ["derive", "rc"] }
-serde_json = "1"
+serde_json = { version = "1", features = ["preserve_order"] }
```
lpil, line 26: I don't think this is being used any more?

Papipo: Probably not but I'll remove it and run the tests just in case.

Papipo: @lpil removed, cargo tests still pass.

lpil (commented):
Thank you! Could you update the changelog please

Papipo:
> Thank you! Could you update the changelog please

Done!

Papipo:
> Thank you! Could you update the changelog please

Done and rebased due to conflicts.

lpil (commented):
Oh! I've just noticed that the top level `{"gleam.toml":` is missing!

Papipo:
> Oh! I've just noticed that the top level `{"gleam.toml":` is missing!

Oh, nice catch. That was definitely lost when I trashed the first iteration. Working on it 🚀

Papipo:
In fact there is no test for this nesting thing, as the snapshots are for the `PackageConfig` serialisation and the nesting is happening at the export command level. Should I add a snapshop test for this?

lpil (commented):
We want the tests to test this output, not just a subset of that output 🙏

lpil (approved):
Thank you!!

PR #4301, opened 2025-03-05
Make JavaScript BitArray deprecation warnings more compact

richard-viney:
Each warning is now one line instead of three.


lpil (approved):
Oh yes, much better. Thank you

PR #4302, opened 2025-03-05
Change my name in the changelog

Nicd:
Sorry for the churn, I figured maybe it would be better to have my real name so someone can hire me and pay me the big bucks as I've made these awesome changes.


lpil (approved):
No trouble at all! Thank you for your help!

PR #4303, opened 2025-03-05
Refactor exhaustiveness checking

giacomocavalieri:
This is quite big change refactoring the pattern matching compilation code.
- I've changed a lot of names here and there to be a bit more descriptive and hopefully help understand what is going on in the code
- I've also added quite a few of documentation comments to try and explain all the terminology used and hopefully make things clearer to skim through
- After removing unused code, a couple of files were small enough to be deleted and just be made part of the `exhaustiveness.rs` one. Most notably, `Pattern` was moved there
- While the decision tree has changed in its shape, the `Match` data structure hasn't changed much. The code consuming the result of the compilation hasn't been touched by this refactoring. However, I've tried to make the public facing api easier to use; as a result, I think that now the code in `type_/expression.rs` is a lot nicer

After this refactoring gets merged I will move on to deal with compilation of string patterns (and maybe bit arrays as well!), and finally we could start using the tree for code generation on the js target


compiler-core/src/exhaustiveness.rs
```diff
@@ -29,208 +75,629 @@
...
-    /// The tuples are in the form `(name, source)` (i.e `bla = source`).
+    /// The tuples are in the form `(name, value)`, so `(wibble, var)`
+    /// corresponds to `let wibble = var`.
+    ///
     bindings: Vec<(EcoString, Variable)>,
 
     /// The index of the clause in the case expression that should be run.
-    clause_index: u16,
+    ///
+    branch_index: usize,
```
lpil, line 263: Why is this called the branch index? It seems to still be the index of the clause, not a branch.

compiler-core/src/exhaustiveness.rs
```diff
@@ -29,208 +75,629 @@
...
+/// Let's have a look at an example. Imagine we have a pattern like this one:
+/// `a is Wibble(1, _, [])`; after performing a runtime check to make sure `a`
+/// is indeed a `Wibble`, we'll need to perform additional checks on it's
+/// arguments: that pattern will be replaced by three new ones `a0 is 1`,
+/// `a1 is _` and `a2 is []`. Those new variables are the `args`.
+///
+#[derive(Clone, Debug)]
+pub struct RuntimeCheck {
+    kind: RuntimeCheckKind,
+    args: Vec<Variable>,
```
lpil, line 441: Why are the arguments not within the enum? There's a implicit relationship between the number of arguments and the value of the other field, but due to the type design not reflecting that the compiler has to assert the length when using them.

compiler-core/src/exhaustiveness.rs
```diff
@@ -29,208 +75,629 @@
...
-    /// 1. The "condition" to evaluate. We just use a dummy value, but in a real
-    ///    compiler this would likely be an AST node of sorts.
-    /// 2. The body to evaluate if the guard matches.
-    /// 3. The sub tree to evaluate when the guard fails.
-    Guard(usize, Body, Box<Decision>),
+    // todo)) since the tree is not used for code generation, the `guard` and
+    // `if_true` fields are unused.
+    // But they will be useful once we also use this for code gen purposes and not
+    // just for exhaustiveness checking
+    #[allow(dead_code)]
```
lpil, line 611: Why is the linter disabled here? This variant is used, as far as I can see.

lpil (approved):
Super lovely work!!! I've left a couple tiny questions inline

compiler-core/src/exhaustiveness.rs
giacomocavalieri: Oops my bad!

compiler-core/src/exhaustiveness.rs
giacomocavalieri: My bad should have been only for the field, not the entire variant!

lpil (approved):
Thank you!!

PR #4304, opened 2025-03-05
Make compiler hint to import unqualified types/values

### LINKED ISSUE #4297: Show hint to import unqualified types/values

mine-tech-oficial:
This PR closes #4297

- [x] Show hint to import unqualified types
- [x] Show hint to import unqualified values
- [x] Consider the type's arity
- [x] Consider the value's arity

Currently the way it decides which module to hint is by preferring imported modules over importable ones, and sorting by name length.

### DIFF
--- a/compiler-core/src/analyse/imports.rs
+++ b/compiler-core/src/analyse/imports.rs
@@ -53,7 +53,9 @@ impl<'context, 'problems> Importer<'context, 'problems> {
             self.problems.error(Error::UnknownModule {
                 location,
                 name: name.clone(),
-                suggestions: self.environment.suggest_modules(&name, Imported::Module),
+                suggestions: self
+                    .environment
+                    .suggest_qualified_modules(&name, Imported::Module),
             });
             return;
         };
--- a/compiler-core/src/analyse/imports.rs
+++ b/compiler-core/src/analyse/imports.rs
@@ -53,9 +53,7 @@ impl<'context, 'problems> Importer<'context, 'problems> {
             self.problems.error(Error::UnknownModule {
                 location,
                 name: name.clone(),
-                suggestions: self
-                    .environment
-                    .suggest_qualified_modules(&name, Imported::Module),
+                suggestions: self.environment.suggest_modules(&name, Imported::Module),
             });
             return;
         };
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2294,7 +2294,7 @@ but no type in scope with that name."
                         title: "Unknown type".into(),
                         text,
                         hint: match label_text {
-                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name)),
+                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name, true)),
                             Some(_) => None
                         },
                         level: Level::Error,
@@ -2315,6 +2315,7 @@ but no type in scope with that name."
                     variables,
                     name,
                     type_with_name_in_scope,
+                    suggestions
                 } => {
                     let text = if *type_with_name_in_scope {
                         wrap_format!("`{name}` is a type, it cannot be used as a value.")
@@ -2330,7 +2331,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: None,
+                        hint: suggestions.first().map(|suggestion| suggestion.suggestion(name, false)),
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2295,7 +2295,7 @@ but no type in scope with that name."
                         title: "Unknown type".into(),
                         text,
                         hint: match label_text {
-                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name, Layer::Type)),
+                            None => suggestions.first().map(|suggestion| suggestion.unqualified_suggestion(name, Layer::Type)),
                             Some(_) => None
                         },
                         level: Level::Error,
@@ -2332,7 +2332,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: suggestions.first().map(|suggestion| suggestion.suggestion(name, Layer::Value)),
+                        hint: suggestions.first().map(|suggestion| suggestion.unqualified_suggestion(name, Layer::Value)),
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
@@ -2387,7 +2387,7 @@ Private types can only be used within the module that defines them.",
                 } => Diagnostic {
                     title: "Unknown module".into(),
                     text: format!("No module has been found with the name `{name}`."),
-                    hint: suggestions.first().map(|suggestion| suggestion.suggestion(name)),
+                    hint: suggestions.first().map(|suggestion| suggestion.qualified_suggestion(name)),
                     level: Level::Error,
                     location: Some(Location {
                         label: Label {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2295,7 +2295,7 @@ but no type in scope with that name."
                         title: "Unknown type".into(),
                         text,
                         hint: match label_text {
-                            None => suggestions.first().map(|suggestion| suggestion.unqualified_suggestion(name, Layer::Type)),
+                            None => suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name)),
                             Some(_) => None
                         },
                         level: Level::Error,
@@ -2332,7 +2332,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: suggestions.first().map(|suggestion| suggestion.unqualified_suggestion(name, Layer::Value)),
+                        hint: suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name)),
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
@@ -2387,7 +2387,7 @@ Private types can only be used within the module that defines them.",
                 } => Diagnostic {
                     title: "Unknown module".into(),
                     text: format!("No module has been found with the name `{name}`."),
-                    hint: suggestions.first().map(|suggestion| suggestion.qualified_suggestion(name)),
+                    hint: suggestions.first().map(|suggestion| suggestion.suggest_import(name)),
                     level: Level::Error,
                     location: Some(Location {
                         label: Label {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::unwrap_used, clippy::expect_used)]
+use crate::ast::Layer;
 use crate::build::{Outcome, Runtime, Target};
 use crate::diagnostic::{Diagnostic, ExtraLabel, Label, Location};
 use crate::type_::collapse_links;
@@ -2294,7 +2295,7 @@ but no type in scope with that name."
                         title: "Unknown type".into(),
                         text,
                         hint: match label_text {
-                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name, true)),
+                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name, Layer::Type)),
                             Some(_) => None
                         },
                         level: Level::Error,
@@ -2331,7 +2332,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: suggestions.first().map(|suggestion| suggestion.suggestion(name, false)),
+                        hint: suggestions.first().map(|suggestion| suggestion.suggestion(name, Layer::Value)),
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2268,6 +2268,7 @@ Note: If the same type variable is used for multiple fields, all those fields ne
                     location,
                     name,
                     hint,
+                    suggestions
                 } => {
                     let label_text = match hint {
                         UnknownTypeHint::AlternativeTypes(types) => did_you_mean(name, types),
@@ -2292,7 +2293,10 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown type".into(),
                         text,
-                        hint: None,
+                        hint: match label_text {
+                            None => suggestions.first().map(|suggestion| suggestion.suggestion(name)),
+                            Some(_) => None
+                        },
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -2295,7 +2295,7 @@ but no type in scope with that name."
                         title: "Unknown type".into(),
                         text,
                         hint: match label_text {
-                            None => suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name)),
+                            None => suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name, Layer::Type)),
                             Some(_) => None
                         },
                         level: Level::Error,
@@ -2332,7 +2332,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name)),
+                        hint: suggestions.first().map(|suggestion| suggestion.suggest_unqualified_import(name, Layer::Value)),
                         level: Level::Error,
                         location: Some(Location {
                             label: Label {
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -762,6 +762,7 @@ impl Environment<'_> {
         let mut suggestions = self
             .importable_modules
             .iter()
+<<<<<<< HEAD
             .filter_map(|(importable, module_info)| match layer {
                 Layer::Type => match module_info.get_public_type(name) {
                     Some(type_) => match arity {
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -384,6 +384,7 @@ impl Environment<'_> {
         &mut self,
         module_alias: &Option<(EcoString, SrcSpan)>,
         name: &EcoString,
+        arity: Option<usize>,
     ) -> Result<&TypeConstructor, UnknownTypeConstructorError> {
         match module_alias {
             None => self
@@ -393,7 +394,7 @@ impl Environment<'_> {
                     name: name.clone(),
                     hint: self.unknown_type_hint(name),
                     suggestions: self
-                        .suggest_modules_for_type_or_value(Imported::Type(name.clone())),
+                        .suggest_modules_for_type_or_value(Imported::Type(name.clone()), arity),
                 }),
 
             Some((module_name, _)) => {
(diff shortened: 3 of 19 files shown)


mine-tech-oficial:
The hints should be done now, although I don't if its possible to find the arity (suppose someone wrote `Dict`, mentioning `dict.Dict`, but didn't put the types, which the code would recognize as "nullary"). Also, the current code for suggesting modules to import also doesn't take the arity into account.

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +747,50 @@ impl Environment<'_> {
...
+    ) -> Vec<TypeOrVariableSuggestion> {
+        let mut suggestions = self
+            .importable_modules
+            .iter()
+            .filter_map(|(importable, module_info)| {
+                match &imported {
+                    // Don't suggest importing modules if they are already imported
+                    _ if self
+                        .imported_modules
+                        .contains_key(importable.split('/').last().unwrap_or(importable)) =>
```
lpil, line 763: There's no fixed relationship between the name used once imported and the name of a module, and this code will produce the same name for multiple modules, so this doesn't successfully skip over already imported modules.

compiler-core/src/type_/error.rs
```diff
@@ -108,6 +108,38 @@ pub enum UnknownField {
...
+}
+
+impl TypeOrVariableSuggestion {
+    pub fn suggestion(&self, name: &str, is_type: bool) -> String {
+        let module = self.name();
+
+        if is_type {
+            format!("Did you mean to import `{module}.{{type {name}}}`?")
+        } else {
+            format!("Did you mean to import `{module}.{{{name}}}`")
```
lpil, line 124: I don't want to encourage unqualified importing of values. Perhaps we should only make these suggestions for types.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__imported_constructor_instead_of_type.snap
```diff
@@ -23,3 +23,4 @@ error: Unknown type
 The type `Wibble` is not defined or imported in this module.
 There is a value in scope with the name `Wibble`, but no type in scope with
 that name.
+Hint: Did you mean to import `module.{type Wibble}`?
```
lpil, line 26: I think the \` should be before the import keyword?

compiler-core/src/error.rs
```diff
@@ -2326,7 +2331,7 @@ but no type in scope with that name."
                     Diagnostic {
                         title: "Unknown variable".into(),
                         text,
-                        hint: None,
+                        hint: suggestions.first().map(|suggestion| suggestion.suggestion(name, false)),
```
lpil, line 2334: There's a vector of suggestions, but only the first one is used. Seems like only one should be given in that case.

compiler-core/src/type_/error.rs
```diff
@@ -108,6 +108,38 @@ pub enum UnknownField {
...
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TypeOrVariableSuggestion {
+    Imported(EcoString),
+    Importable(EcoString),
+}
+
+impl TypeOrVariableSuggestion {
+    pub fn suggestion(&self, name: &str, is_type: bool) -> String {
```
lpil, line 118: Never use bools as arguments please 🙏  This could be a `Layer`.

lpil (commented):
Thank you! I've added some comments inline, and snapshot tests will need to be added to test all the new paths through the code.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__imported_constructor_instead_of_type.snap
mine-tech-oficial: I think if we make the "import" part of the "code" section (between backticks), it should probably be written "Did you mean to write `code here`". What do you think?

mine-tech-oficial: Actually, the current code for module suggestions (in the qualified case) says "Did you mean to import `module`", so how it is right now should be more consistent

compiler-core/src/type_/error.rs
mine-tech-oficial: Okay! Tying with my point below, we could write the "generic" value case as "Did you mean to reference `module.{value}`" and, in case it isn't already imported, "Did you mean to import `module` and reference `module.{value}`". What do you think?

compiler-core/src/type_/environment.rs
mine-tech-oficial: To be fair I had just copied that line from the `suggest_modules` function in the `environment` module, think it'd be correct. I think I should remove the later code where it adds the imported modules to the suggestions vector, although I don't know how I'd detect if a module is imported or not

compiler-core/src/type_/error.rs
mine-tech-oficial: To be frank, I did think that bool wasn't a good fit, although I didn't know what to name the Enum as. Thanks!

mine-tech-oficial:
The code should now consider arity, but it is an `Option` as I don't think it is possible to find the arity in all cases.

mine-tech-oficial:
Hey @lpil, I've answered some of your questions, waiting on clarifications!

mine-tech-oficial:
I have implemented the code actions locally. Should I push them to this PR or wait for a future one?

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +751,74 @@ impl Environment<'_> {
...
+                        && (arity.is_none()
+                            || module_info.get_public_value(name).unwrap().type_.fn_arity()
+                                == arity) =>
+                {
+                    Some(TypeOrVariableSuggestion::Imported(module.clone()))
+                }
+                _ => None,
+            },
+        ));
+        // Filter and sort options based on if its already imported and on alphabetical order.
```
lpil, line 818: What's the filtering documented here?

compiler-core/src/type_/environment.rs
lpil: Here's some examples:

```gleam 
import my_module as list
```
```gleam
import myapp/list
```
In these two examples `gleam/list` would not be suggested, even though it should be.

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +751,74 @@ impl Environment<'_> {
...
+                                || module_info
+                                    .get_public_value(name)
+                                    .unwrap()
+                                    .type_
+                                    .fn_arity()
+                                    == arity) =>
+                    {
+                        Some(TypeOrVariableSuggestion::Importable(importable.clone()))
+                    }
+                    _ => None,
```
lpil, line 792: No catch-all patterns please. Move the guards into the body of each match clause to fix this, and it'll also permit you to make the code much easier to read by using multiple `if` and comments.

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +751,74 @@ impl Environment<'_> {
...
+                }
+                Imported::Value(name)
+                    if module_info.get_public_value(name).is_some()
+                        && (arity.is_none()
+                            || module_info.get_public_value(name).unwrap().type_.fn_arity()
+                                == arity) =>
+                {
+                    Some(TypeOrVariableSuggestion::Imported(module.clone()))
+                }
+                _ => None,
```
lpil, line 815: As above, thank you

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +751,74 @@ impl Environment<'_> {
...
+                        .imported_modules
+                        .contains_key(importable.split('/').last().unwrap_or(importable)) =>
+                    {
+                        None
+                    }
+                    Imported::Type(name)
+                        if module_info.get_public_type(name).is_some()
+                            && (arity.is_none()
+                                || module_info.get_public_type(name).unwrap().parameters.len()
+                                    == arity.unwrap()) =>
```
lpil, line 776: Use of unwrap is never permitted, and this is wastefully performing the same work multiple times!

compiler-core/src/type_/error.rs
```diff
@@ -108,6 +108,37 @@ pub enum UnknownField {
...
+    }
+
+    pub fn name(&self) -> &EcoString {
+        match self {
+            TypeOrVariableSuggestion::Imported(name)
+            | TypeOrVariableSuggestion::Importable(name) => name,
+        }
+    }
+
+    pub fn last_name_component(&self) -> &str {
```
lpil, line 134: This function seems unused.

compiler-core/src/type_/error.rs
```diff
@@ -108,6 +108,37 @@ pub enum UnknownField {
...
+    pub fn suggestion(&self, name: &str, layer: Layer) -> String {
+        let module = self.name();
+
+        match layer {
+            Layer::Type => format!("Did you mean to import `{module}.{{type {name}}}`?"),
+            Layer::Value => format!("Did you mean to import `{module}.{{{name}}}`"),
+        }
+    }
+
+    pub fn name(&self) -> &EcoString {
```
lpil, line 127: I don't think this needs to be public.

compiler-core/src/type_/error.rs
```diff
@@ -108,6 +108,37 @@ pub enum UnknownField {
     NoFields,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub enum TypeOrVariableSuggestion {
```
lpil, line 112: This seems to be exactly the same data structure as the existing ModuleSuggestion one, so let's reuse that instead.

That name makes more sense too, it is a module being suggested.

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +751,74 @@ impl Environment<'_> {
             .collect()
     }
 
+    /// Suggest modules to import or use, for an unqualified type
+    pub fn suggest_modules_for_type_or_value(
+        &self,
+        imported: Imported,
```
lpil, line 757: I don't think this is the right type to use here as there is an `Imported` variant which is unused, but really this code wants to know whether it is a type or a value being used. The type for that is `Layer`, so let's use that.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__imported_constructor_instead_of_type.snap
lpil: I think this is confusing as it looks like Gleam but when you copy it there would be a syntax error. Let's use either valid Gleam or make it look more different.

In this specific case I think it needs to say to update the existing import

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__same_imports_multiple_times.snap
```diff
@@ -35,3 +35,4 @@ error: Unknown variable
   │                                  ^^^
 
 The name `zoo` is not in scope here.
+Hint: Did you mean to import `wibble.{zoo}`
```
lpil, line 38: We do not want to recommend importing functions or constants in an unqualified way.

lpil (commented):
Thank you! Looking good. I've left some comments inline, and the tests still need to be added for each possible path through the code.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__imports__imported_constructor_instead_of_type.snap
mine-tech-oficial: Do you mean I should change the hint when the module is already imported?

lpil: Yes please. The errors should tell you a thing you can do to fix the problem.

mine-tech-oficial: Ok. Thanks!

compiler-core/src/error.rs
mine-tech-oficial: Actually, having a vector of suggestions allows the LS to give all possible quick fixes when a code action to do so is done (I have one basically ready, will do a PR after this one is pushed)

compiler-core/src/type_/environment.rs
mine-tech-oficial: Actually there is no filtering. It is a leftover from what I copied. Will remove!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Okay, will change!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Will resolve!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Will resolve!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Fixed!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Should be fixed now!

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__same_imports_multiple_times.snap
mine-tech-oficial: Makes sense. Fixed now!

compiler-core/src/type_/error.rs
mine-tech-oficial: Yeah, it is. I've removed it.

compiler-core/src/type_/error.rs
mine-tech-oficial: It isn't even used actually, so I've removed it.

mine-tech-oficial:
@lpil, I thin I have addressed all of your points! Right now I just need to comment and write tests, and I'm waiting on some clarifications.

compiler-core/src/type_/error.rs
lpil: Sounds good, and can rename the existing function in that class to make clear the context for it

compiler-core/src/type_/environment.rs
```diff
@@ -741,6 +748,104 @@ impl Environment<'_> {
...
+                            None => Some(TypeOrVariableSuggestion::Importable(importable.clone())),
+                        },
+                        None => None,
+                    },
+                }
+            },
+        ));
+
+        // Sort options based on if its already imported and on lexicographical order.
+        suggestions.into_iter().sorted().collect()
```
lpil, line 846: This function has 7 layers of identation and is quite hard to read. It would be fab if it could be cleaned up a bit 🙏 

It looks like if arity is none it always returns none? If so it could be make easier to read and also avoid iterating by returning early when it is none, perhaps using let-else.

lpil (commented):
Thank you

mine-tech-oficial:
Should be all done now!

compiler-core/src/type_/environment.rs
mine-tech-oficial: Refactored!

compiler-core/src/analyse/imports.rs
```diff
@@ -53,7 +53,9 @@ impl<'context, 'problems> Importer<'context, 'problems> {
             self.problems.error(Error::UnknownModule {
                 location,
                 name: name.clone(),
-                suggestions: self.environment.suggest_modules(&name, Imported::Module),
+                suggestions: self
+                    .environment
+                    .suggest_qualified_modules(&name, Imported::Module),
```
lpil, line 58: I don't think this makes sense here any more? It's suggesting modules, not qualified modules (which are not a thing in this context)

compiler-core/src/type_/error.rs
```diff
@@ -119,7 +119,7 @@ pub enum ModuleSuggestion {
 }
 
 impl ModuleSuggestion {
-    pub fn suggestion(&self, module: &str) -> String {
+    pub fn qualified_suggestion(&self, module: &str) -> String {
```
lpil, line 122: I think `suggest_import` may be more clear. It's not a suggestion that is qualified.

compiler-core/src/type_/error.rs
```diff
@@ -134,6 +134,18 @@ impl ModuleSuggestion {
         }
     }
 
+    pub fn unqualified_suggestion(&self, name: &str, layer: Layer) -> String {
```
lpil, line 137: I think this is `suggest_unqualified_import` perhaps?

compiler-core/src/type_/error.rs
```diff
@@ -134,6 +134,18 @@ impl ModuleSuggestion {
...
+        match self {
+            ModuleSuggestion::Importable(module) => match layer {
+                Layer::Type => format!("Did you mean to import `{module}.{{type {name}}}`?"),
+                Layer::Value => format!("Did you mean to import `{module}.{{{name}}}`?"),
```
lpil, line 141: This still has the confusing syntax that is easy to misread. Perhaps it can avoid using Gleam syntax altogether and copy the format used in the existing function.

compiler-core/src/type_/tests/errors.rs
```diff
@@ -2971,3 +2971,88 @@ pub fn main() {
...
+            "gleam/wibble",
+            "
+            pub fn wobble() { 1 }
+            "
+        ),
+        "
+        pub fn go() { wobble() }
+        "
+    );
+}
```
lpil, line 3058: Could you remove the indentation from the code in the tests please? It makes the snapshots a little harder to read

lpil (commented):
Thank you! Nearly there! Few small things around the wording

compiler-core/src/type_/error.rs
mine-tech-oficial: Yeah, after I had commited it I saw that we had agreed on a better name

compiler-core/src/type_/tests/errors.rs
mine-tech-oficial: Ok!

compiler-core/src/type_/error.rs
mine-tech-oficial: What about "Did you mean to import `module.name`"?

compiler-core/src/analyse/imports.rs
mine-tech-oficial: Ok, I will revert the change.

mine-tech-oficial:
Now it should be ready! Also, the way I've refactored, it allows for the normal module suggestion to consider arity in the future

compiler-core/src/type_/error.rs
lpil: That's not Gleam syntax either. It should either be Gleam syntax, or it should not look like code.

Let's use a sentence without any code-like bits, similar to the existing module error suggestion

mine-tech-oficial: What do you think about "Did you mean to import `{module}` and add `{name}` as a unqualified import" then?

GearsDatapacks: I was thinking something like "did you mean to import the `{value}` value from the `{module}` module?"

mine-tech-oficial: Oh, this is great! Thanks for the suggestion!

compiler-core/src/type_/error.rs
mine-tech-oficial: Ok!

mine-tech-oficial:
@lpil, now the hint is better (thanks Gears for the suggestion) and I have removed the indents in tests, so I think there should be nothing left to do!

mine-tech-oficial:
@lpil, I'm thinking on redoing this on the latest Gleam version, to facilitate merging, although it might take a bit. What do you think?

lpil:
Sounds good to me!

mine-tech-oficial:
Closed, continued in #4602

PR #4305, opened 2025-03-06
Implement the `unit` option for bit arrays on the Javascript target

GearsDatapacks:
Leaving this as a draft until 1.9 is released.


lpil (approved):
Very nice! Thank you!!

PR #4315, opened 2025-03-09
Improve the error message when echo tests fail on missing target executable

sbergen:
If you don't have an executable required by the echo tests installed, you'd get a somewhat cryptic error along the lines of:
```
thread 'tests::echo::echo_dict' panicked at test-output/src/tests/echo.rs:74:10:
spawn run process: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```

It will now be this instead (if you don't have bun installed):
```
thread 'tests::echo::echo_dict' panicked at test-output/src/tests/echo.rs:74:29:
Failed to spawn process 'bun': No such file or directory (os error 2)
```


lpil (approved):
Thank you

PR #4316, opened 2025-03-09
Update local development dependencies documentation

sbergen:
The `echo` tests now have more dependencies during runtime. (See also #4315)


lpil (approved):
Thank you

PR #4323, opened 2025-03-09
Fix echo code generation on the Erlang target

### LINKED ISSUE #4322: `echo` throws error when used at start of pipeline
When you have an `echo` as the first term of a pipeline, the compiler returns a rather cryptic error

Gleam: 1.9.0
OS: Ubuntu 22.04.5 LTS
Runtime: Erlang

Logs (filtered for the error part):
```
{compiling,"/home/pedrohcf/dev/chess_bot/build/dev/erlang/chess_bot/_gleam_artefacts/tests.erl"}
src/tests.gleam:4:5: function echo/4 undefined
%    4| }
%     |  ^^^^

src/tests.gleam:5:14: variable '_pipe' is unbound
src/tests.gleam:4:10: Warning: variable '_pipe' is bound multiple times in this pattern.
If you mean to ignore this value, use '_' or
a different underscore-prefixed name
%    4| }
%     |  ^^^^^^^^^

{failed,"/home/pedrohcf/dev/chess_bot/build/dev/erlang/chess_bot/_gleam_artefacts/tests.erl"}
{compiled,chess_bot,
          "/home/pedrohcf/dev/chess_bot/build/dev/erlang/chess_bot/ebin/chess_bot.beam"}
ERROR Failed error=ShellCommand { program: "escript", reason: Unknown }
error: Shell command failure

```

giacomocavalieri:
This PR fixes #4322 
- I fixed a bug in code generation that meant using echo to print an entire pipeline or record update would end up generating invalid erlang code, sorry for missing this! 🙇
- Since I was there I made a small change to the formatter to make sure it's easier to see echo is printing an entire pipeline if used at the beginning of it

### DIFF
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -6523,8 +6523,7 @@ fn echo_printing_multiline_pipeline() {
 }
 "#,
         r#"fn main() {
-  echo
-    first
+  echo first
     |> wibble
     |> wobble
 }
--- a/compiler-core/src/format/tests.rs
+++ b/compiler-core/src/format/tests.rs
@@ -6512,3 +6512,32 @@ fn echo_with_following_value_in_a_pipeline() {
 "#
     );
 }
+
+#[test]
+fn echo_printing_multiline_pipeline() {
+    assert_format_rewrite!(
+        r#"fn main() {
+  echo first
+  |> wibble
+  |> wobble
+}
+"#,
+        r#"fn main() {
+  echo
+    first
+    |> wibble
+    |> wobble
+}
+"#
+    );
+}
+
+#[test]
+fn echo_printing_one_line_pipeline() {
+    assert_format!(
+        r#"fn main() {
+  echo first |> wibble |> wobble
+}
+"#
+    );
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,4 +10,11 @@
 
 ### Formatter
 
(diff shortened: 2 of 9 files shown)


compiler-core/src/format/tests.rs
```diff
@@ -6512,3 +6512,32 @@ fn echo_with_following_value_in_a_pipeline() {
 "#
     );
 }
+
+#[test]
+fn echo_printing_multiline_pipeline() {
+    assert_format_rewrite!(
```
lpil, line 6518: This doesn't match how any of the other keywords are formatted today. The others use this:

```gleam
echo first
  |> wibble
  |> wobble
```

giacomocavalieri: Ok I can go for that as well! Do you think it would be clear that it's printing the whole thing and not just `first` though?
I was afraid people would parse it as `{ echo first } |> wibble |> wobble` instead of `echo { first |> wibble |> wobble }`

lpil: I think the formatting shifting the indentation by one step will be clear enough 👍

giacomocavalieri: Ok!

giacomocavalieri:
Should be all fixed now!

lpil (commented):
Wonderful, thank you. Would you mind updating the changelog? 🙏

lpil (approved):
Thank you!

inoas:
I really like the indent on echo in the formatter. <3

PR #4325, opened 2025-03-09
Fix and clarify cache file handling

### LINKED ISSUE #4320: Temporarily removed file does not always get recompiled
## Issue

Under the following scenario, a file will not get recompiled when it should
1. Add `a.gleam` that compiles ok, depends on `b.gleam` and uses function `f` in it.
2. Run `gleam build`
3. Remove `a.gleam` (or move it out of the `src` directory)
4. Modify the signature of function `f`.
5. Run `gleam build`
6. Restore `a.gleam` unchanged.
7. Run `gleam build`

**Expected:** There should be a compiler error, as the call to `f` is no longer valid
**Actual:** The compiler does not raise any errors, and running `gleam run` (on Erlang), can lead to the following error:
```
runtime error: Erlang error

A function was called but it did not exist.

stacktrace:
  b.f unknown source
``` 

## Context

This might seem like a very rare corner case, but is very often encountered with my [markdown code snippet validation tool](https://github.com/sbergen/checkmark), which temporarily adds the snippets as source files and runs the compiler to validate them.

## Cause

Based on my investigation of the issue, it seems like there are two issues at play here:
1. Modules who's source files have gone missing do not have their cache files deleted.
2. We do look at file mtimes, but we only take them into account when running as an LSP. I assume this is to optimize the use of undoing or reverting changes (i.e. not recompile if you undid your changes before recompiling).

So the compiler does this:
1. A valid version of `a`'s cache files are generated
2. When `a` is removed and `b` is changed, we don't do anything with `a`'s cache files. New cache files for `b` are generated.
3. When `a` is restored, the content is the same and we ignore the mtime. Since `b` didn't change from last time, the compiler concludes that nothing needs to be done.

## Suggested solution

I was considering deleting the cache files when implementing #3873, but it was easier at the time to just mark them as stale. Given this issue, I think it would be better to just delete the cache files when detecting that a source file was moved/removed, fixing both issues. I'd be happy to implement a fix, if this seems like a good solution.

The other clear alternative I see would be to never ignore the mtime.

sbergen:
Functional changes:
* Delete cache files when we detect that the source file for a module has been deleted. Resolves #4320
* Delete *all* cache files when loading a stale module
* Handles the detection of deleted *nested* modules properly. I messed this up in #3873 so I decided to do some related refactoring to clarify the related code.

Refactoring:
* Remove very repetitive use of `replace("/", "@")`. Add two versions of it: One for erlang module names, and one for cache file names. My understanding is that they just happen to do the same thing, so I duplicated it. If there's some reason the cache file names must match the erlang module names, then we should probably change it further.
* For the use of `replace("/", "@")` related to cache files, group these operations together as `CacheFiles`, as there was further duplication of the same logic in different parts of the code base.
* Group logic related to discovery, validation, and module name derivation from gleam source files under `GleamFile`

I was originally going to refactor first separately, but it was hard to keep doing the wrong thing, and since the actual fixes are quite small, I thought it might be better if I just outlined them in the PR.

### DIFF
--- a/compiler-core/src/build/module_loader.rs
+++ b/compiler-core/src/build/module_loader.rs
@@ -11,7 +11,7 @@ use serde::{Deserialize, Serialize};
 use super::{
     Mode, Origin, SourceFingerprint, Target,
     package_compiler::{CacheMetadata, CachedModule, Input, UncompiledModule, module_name},
-    package_loader::CodegenRequired,
+    package_loader::{CodegenRequired, GleamFile},
 };
 use crate::{
     Error, Result,
@@ -28,7 +28,6 @@ pub(crate) struct ModuleLoader<'a, IO> {
     pub target: Target,
     pub codegen: CodegenRequired,
     pub package_name: &'a EcoString,
-    pub source_directory: &'a Utf8Path,
     pub artefact_directory: &'a Utf8Path,
     pub origin: Origin,
     /// The set of modules that have had partial compilation done since the last
@@ -48,12 +47,12 @@ where
     /// Whether the module has changed or not is determined by comparing the
     /// modification time of the source file with the value recorded in the
     /// `.timestamp` file in the artefact directory.
-    pub fn load(&self, path: Utf8PathBuf) -> Result<Input> {
-        let name = module_name(self.source_directory, &path);
+    pub fn load(&self, file: GleamFile) -> Result<Input> {
+        let name = file.module_name.clone();
         let artefact = name.replace("/", "@");
-        let source_mtime = self.io.modification_time(&path)?;
+        let source_mtime = self.io.modification_time(&file.path)?;
 
-        let read_source = |name| self.read_source(path, name, source_mtime);
+        let read_source = |name| self.read_source(file.path.clone(), name, source_mtime);
 
         let meta = match self.read_cache_metadata(&artefact)? {
             Some(meta) => meta,
@@ -84,7 +83,7 @@ where
             }
         }
 
-        Ok(Input::Cached(self.cached(name, meta)))
+        Ok(Input::Cached(self.cached(file, meta)))
     }
 
     /// Read the timestamp file from the artefact directory for the given
@@ -129,12 +128,12 @@ where
         )
     }
 
-    fn cached(&self, name: EcoString, meta: CacheMetadata) -> CachedModule {
+    fn cached(&self, file: GleamFile, meta: CacheMetadata) -> CachedModule {
         CachedModule {
             dependencies: meta.dependencies,
-            source_path: self.source_directory.join(format!("{}.gleam", name)),
+            source_path: file.path,
             origin: self.origin,
-            name,
+            name: file.module_name,
             line_numbers: meta.line_numbers,
         }
     }
--- a/compiler-core/src/build/module_loader.rs
+++ b/compiler-core/src/build/module_loader.rs
@@ -49,12 +49,11 @@ where
     /// `.timestamp` file in the artefact directory.
     pub fn load(&self, file: GleamFile) -> Result<Input> {
         let name = file.module_name.clone();
-        let artefact = name.replace("/", "@");
         let source_mtime = self.io.modification_time(&file.path)?;
 
         let read_source = |name| self.read_source(file.path.clone(), name, source_mtime);
 
-        let meta = match self.read_cache_metadata(&artefact)? {
+        let meta = match self.read_cache_metadata(&file)? {
             Some(meta) => meta,
             None => return read_source(name).map(Input::New),
         };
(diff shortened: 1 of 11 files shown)


compiler-core/src/build/module_loader.rs
```diff
@@ -48,14 +47,13 @@ where
...
-    pub fn load(&self, path: Utf8PathBuf) -> Result<Input> {
-        let name = module_name(self.source_directory, &path);
-        let artefact = name.replace("/", "@");
-        let source_mtime = self.io.modification_time(&path)?;
+    pub fn load(&self, file: GleamFile) -> Result<Input> {
+        let name = file.module_name.clone();
+        let source_mtime = self.io.modification_time(&file.path)?;
 
-        let read_source = |name| self.read_source(path, name, source_mtime);
+        let read_source = |name| self.read_source(file.path.clone(), name, source_mtime);
```
sbergen, line 54: I didn't spend too much time trying to get rid of these two clones, but there might be something we could still optimize. I thought I'd get overall feedback first before going too detailed.

compiler-core/src/build/package_loader.rs
```diff
@@ -112,10 +110,13 @@ where
...
-        // Check for any removed modules, by looking at cache files that don't exist in inputs
-        for cache_file in gleam_cache_files(&self.io, &self.artefact_directory) {
-            let module = module_name(&self.artefact_directory, &cache_file);
+        // Check for any removed modules, by looking at cache files that don't exist in inputs.
+        // Delete the cache files for removed modules and mark them as stale
+        // to trigger refreshing dependent modules.
+        for module in CacheFiles::modules_with_meta_files(&self.io, &self.artefact_directory) {
             if (!inputs.contains_key(&module)) {
+                tracing::debug!(%module, "module_removed");
+                CacheFiles::new(&self.artefact_directory, &module).delete(&self.io);
```
sbergen, line 119: This line is the fix for the temporarily deleted files not being always recompiled.

compiler-core/src/build/package_loader.rs
```diff
@@ -293,18 +266,12 @@ where
...
+        // that this module needs to be recompiled until it successfully compiles at least once.
         // This can happen if the stale dependency includes breaking changes.
-        let artefact = cached.name.replace("/", "@");
-        let meta_path = self
-            .artefact_directory
-            .join(artefact.as_str())
-            .with_extension("cache_meta");
-
-        let _ = self.io.delete_file(&meta_path);
+        CacheFiles::new(&self.artefact_directory, &cached.name).delete(&self.io);
```
sbergen, line 274: Here we delete more files than before.

compiler-core/src/build/package_loader.rs
```diff
@@ -1715,3 +1682,143 @@ impl<'a> Inputs<'a> {
...
+
+        // my/module
+        let _ = module_path.set_extension("");
+
+        // Stringify
+        let name = module_path.to_string();
+
+        // normalise windows paths
+        name.replace("\\", "/").into()
+    }
```
sbergen, line 1740: This was moved, and made private so that it wouldn't be used with cache files (which caused some extra recompilation because we thought the modules were removed).

compiler-core/src/build/package_loader.rs
```diff
@@ -1715,3 +1682,143 @@ impl<'a> Inputs<'a> {
...
+                slash = "(/|\\\\)",
+            ))
+            .expect("is_gleam_path() RE regex")
+        })
+        .is_match(
+            path.strip_prefix(dir)
+                .expect("is_gleam_path(): strip_prefix")
+                .as_str(),
+        )
+    }
```
sbergen, line 1760: This was just moved, as it wasn't used anywhere else.

compiler-core/src/build/package_loader.rs
```diff
@@ -1715,3 +1682,143 @@ impl<'a> Inputs<'a> {
...
+        let mut module_path = path
+            .strip_prefix(dir)
+            .expect("Stripping package prefix from module path")
+            .to_path_buf();
+
+        // my@module
+        let _ = module_path.set_extension("");
+
+        // my/module
+        module_path.to_string().replace("@", "/").into()
```
sbergen, line 1822: This is the corrected and modified version of getting the module name from any cache file.

compiler-core/src/language_server/tests/compilation.rs
```diff
@@ -119,7 +119,7 @@ fn compile_recompile() {
     // This time it does not compile the module again, instead using the
     // cache from the previous run.
     let response = engine.compile_please();
-    assert!(response.result.is_ok());
+    assert_eq!(response.result, Ok(()));
```
sbergen, line 122: I got this test failing as a result of a typo, and the output wasn't too helpful. This makes it a lot better, so I thought I'd commit it.

compiler-core/src/build/package_loader.rs
```diff
@@ -1715,3 +1682,143 @@ impl<'a> Inputs<'a> {
         Ok(())
     }
 }
+
+/// A Gleam source file (`.gleam`) and the module name deduced from it
```
sbergen, line 1686: For everything downwards from here, I'm not sure if this is the best module, or the best location in the source file to put these.

lpil (commented):
Fantastic!! Thank you!!

I've left a note inline, and could you update the changelog also please 🙏

lpil (approved):
Beautiful, thank you!!!

PR #4326, opened 2025-03-10
Add type annotation to main in new project template

### LINKED ISSUE #4313: Add type annotation to `main` in projects generated by `gleam new`
Currently, the `gleam new` command generates a project with a `main` function without a type annotation. It is idiomatic in gleam to include type annotations, and therefore I'd suggest we add an annotation to `main` of the type `-> Nil`.

drewolson:
This PR adds an annotation of `-> Nil` to the `main` function created by the
`gleam new` command. It also adds the same annotation to the `main` function in
the generated `README.md` file.

I've added new snapshot tests for all `*.gleam` files created in both the erlang
and javascript templates.

Fixes #4313


compiler-cli/src/new/tests.rs
```diff
@@ -81,7 +92,18 @@ fn new_with_javascript_template() {
...
+            insta::assert_snapshot!(
+                crate::fs::read(
+                    Utf8PathBuf::from_path_buf(file_path.to_path_buf()).expect("Non Utf8 Path"),
+                )
+                .unwrap()
+            );
+        }
+    });
+
+    insta::glob!(&path, "my_project/**/*.gleam", |file_path| {
```
lpil, line 106: Sorry, what has this file been changed for?

lpil (commented):
Thank you. Could you update the changelog also please 🙏

drewolson:
> Thank you. Could you update the changelog also please 🙏

Done.

compiler-cli/src/new/tests.rs
drewolson: The existing snapshot tests only test the files at the top-level of the generated project. That means neither `src/<my_project>.gleam` nor `test/<my_project>.gleam` had a test around their content. Given that I was updating one of those files, I figured having a snapshot test would be valuable. Let me know if you'd prefer I remove this change.

drewolson: Done. Note that I'm excluding hidden directories so that we don't look at contents within `.git`.

compiler-cli/src/new/snapshots/gleam_cli__new__tests__new_with_javascript_template@test__my_project_test.gleam.snap
```diff
@@ -0,0 +1,16 @@
+---
+source: compiler-cli/src/new/tests.rs
+expression: "crate::fs::read(Utf8PathBuf::from_path_buf(file_path.to_path_buf()).expect(\"Non Utf8 Path\"),).unwrap()"
+---
+import gleeunit
+import gleeunit/should
+
+pub fn main() {
+  gleeunit.main()
```
lpil, line 9: Return annotation on the test main function too please 🙏

lpil (approved):
Thank you!!

PR #4328, opened 2025-03-11
Add labels when showing a hovered function type

### LINKED ISSUE #3507: LSP: function hover info should include argument labels
Signature help, which shows when starting to type arguments in a function call, includes the labels of said arguments, but the Hover info for functions does not (shows only the function type, no labels).

It would make the hover information more useful if it also included argument labels.

giacomocavalieri:
This PR closes #3507 by adding labels to function types that are displayed when hovering on something. This works for constructors, imported functions, functions and function heads!

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -18,7 +18,7 @@ use crate::type_::error::VariableOrigin;
 use crate::type_::expression::Implementations;
 use crate::type_::printer::Names;
 use crate::type_::{
-    self, Deprecation, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
+    self, Deprecation, FieldMap, ModuleValueConstructor, PatternConstructor, Type, ValueConstructor,
 };
 use std::sync::Arc;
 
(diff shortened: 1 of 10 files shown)


lpil (commented):
To me this introduces a new syntax which is a mix of the expression syntax and the type syntax, and it implies that labels are part of the type. Maybe it'd look more like the existing syntax if it included the function name?

Another thing is that for some expressions when you hover it shows the type at the top, but for these it does not. I'm not a big fan of this inconsistency. Is there a more definition-like format we could use for the others too?

giacomocavalieri:
> Maybe it'd look more like the existing syntax if it included the function name?

So what would it show in this example?
```gleam
fn wibble(a: Int, label b: String) {}

pub fn main() {
  let var = wibble
//     ^ Cursor here:
//         var(Int, label: String)
}
```

> Another thing is that for some expressions when you hover it shows the type at the top, but for these it does not.

Yeah I didn't think too much about it, not really sure. If the goal is to always keep a type you can copy-paste into Gleam code then I don't see a meaningful way to show labels for functions

lpil:
It doesn't have to be copy-paste-able, but we should be careful about making a new syntax, and how it could be confused.

PR #4332, opened 2025-03-11
Disallow unlabelled arguments after labelled ones

### LINKED ISSUE #4331: Unlabelled arguemnt after labelled argument error not being emitted for custom type variant constructor
```gleam
pub type Bad {
  Bad(labelled: Int, Float)
}
```

This code is invalid and should be rejected. Using this record will result in invalid code being generated also.

GearsDatapacks:
This PR fixes #4331.
I found a JS codegen test which relied on this bug, so I have removed it.
Edit: There was also a language test which uses this invalid syntax, which has also been removed.


lpil (approved):
Thank you!

PR #4333, opened 2025-03-11
Refresh snapshots

GearsDatapacks:
I noticed that there were a quite a few snapshots left over from removed/renamed tests which haven't been deleted. This PR just gets rid of any unused files.


lpil (approved):
Thank you

PR #4337, opened 2025-03-13
Don't use iifes when compiling blocks on JavaScript

### LINKED ISSUE #3968: Don't use self-invoking anonymous functions when generating blocks on JavaScript

GearsDatapacks:
This PR closes #3968 
I haven't added any new tests, because there are plenty of existing tests covering this. I have simply updated existing snapshots.


GearsDatapacks:
This should be ready for review now. My first iteration caused various codegen bugs due to changes in ordering. I have no fixed those, so this will currently only lift blocks into statement-level if there is no chance of changing execution order. This is documented in the `Order` type.
In future we could maybe be a bit more smart about this, for example, this case will not be optimised currently:
```gleam
add(1, {
  let x = 2
  io.debug(2)
})
```
However, since the expression `1` can never cause side effects, it would be safe to lift this block.

compiler-core/src/javascript/expression.rs
```diff
@@ -50,9 +98,10 @@ pub(crate) struct Generator<'module> {
     // at the top level of the function to use in place of pushing new stack
     // frames.
     pub tail_recursion_used: bool,
+    statement_level: Vec<Document<'ast>>,
```
lpil, line 101: Could you document what this is and how it's used please 🙏

lpil (approved):
Thank you!!!

PR #4338, opened 2025-03-13
Add code action to expand spread in patterns

giacomocavalieri:
This PR adds a new code action to expand a spread (`..`) used in a pattern with all the ignored fields.

This is what it looks like!


compiler-core/src/language_server/engine.rs
```diff
@@ -1016,38 +1019,36 @@ where
...
-                    let positional = positional.join("\n");
-                    let labelled = labelled.join("\n");
                     let content = match (positional.is_empty(), labelled.is_empty()) {
-                        (true, false) => format!("Unused labelled fields:\n{labelled}"),
-                        (false, true) => format!("Unused positional fields:\n{positional}"),
+                        (true, false) => format!("Ignored labelled fields:\n{labelled}"),
+                        (false, true) => format!("Ignored positional fields:\n{positional}"),
                         (_, _) => format!(
-                            "Unused positional fields:
+                            "Ignored positional fields:
```
CloserToTheCenter, line 1048: What's the reason for this rename? does it feel better after using this code action?

giacomocavalieri: I think it is clearer: we tend to use the "unused" wording for variables/import/private stuff that is defined but not used. What the `..` is doing is ignoring some fields

lpil: Undo this change please 🙏 I think it is more accurate to say they are unused rather than ignored, and ignored has a negative connotation.

giacomocavalieri: So should the action say "replace .. with unused fields"?

compiler-core/src/language_server/tests/action.rs
```diff
@@ -79,6 +79,7 @@ const CONVERT_TO_FUNCTION_CALL: &str = "Convert to function call";
 const INLINE_VARIABLE: &str = "Inline variable";
 const CONVERT_TO_PIPE: &str = "Convert to pipe";
 const INTERPOLATE_STRING: &str = "Interpolate string";
+const REPLACE_SPREAD_WITH_IGNORED_FIELDS: &str = "Replace `..` with ignored fields";
```
lpil, line 82: Not totally sure about this name. What else might we call it?

"Fill fields" or "fill unused fields" to match "fill labels"?

giacomocavalieri: Fill unused fields sounds great! I'd not use labels because there might be unlabelled ones as well

giacomocavalieri:
Should be all good, I've renamed the action to "fill in unused fields" and reverted the "unused -> ignored" change, thanks!!

PR #4339, opened 2025-03-13
Improve error message when giving the wrong password for a local token

### LINKED ISSUE #4317: Unclear error message when giving the wrong password for a local token
```
error: Failed to decrypt data

A problem was encountered decrypting data.

    Decryption failed
```

This isn't very friendly

scristobal:
Adds a new error variant `Error::FailedToDecryptApiKey` specifically to handle the case when the decryption of the Hex API token fails. 

Hopefully, the changes will make the error message a bit more friendly and possibly more useful. 

resolves #4317

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,9 @@
   project's `gleam.toml` as a JSON file.
   ([Rodrigo Álvarez](https://github.com/Papipo))
 
+- Improve error message when giving the wrong password for a local token.
+  ([Samuel Cristobal](https://github.com/scristobal))
+
 ### Language server
 
 - The language server now allows renaming of functions, constants,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@
   project's `gleam.toml` as a JSON file.
   ([Rodrigo Álvarez](https://github.com/Papipo))
 
-- Improve error message when giving the wrong password for a local token.
+- Improve error messages when encrypting and decrypting local Hex AIP key.
   ([Samuel Cristobal](https://github.com/scristobal))
 
 ### Language server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,7 +66,7 @@
   project's `gleam.toml` as a JSON file.
   ([Rodrigo Álvarez](https://github.com/Papipo))
 
-- Improve error messages when encrypting and decrypting local Hex AIP key.
+- Improve error messages when encrypting and decrypting local Hex API key.
   ([Samuel Cristobal](https://github.com/scristobal))
 
 ### Language server
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -111,7 +111,10 @@ encrypt your Hex API key.
             return Ok(None);
         };
         let password = self.ask_local_password()?;
-        let unencrypted = encryption::decrypt_with_passphrase(encrypted.as_bytes(), &password)?;
+        let unencrypted = encryption::decrypt_with_passphrase(encrypted.as_bytes(), &password)
+            .map_err(|e| Error::FailedToDecryptApiKey {
+                detail: e.to_string(),
+            })?;
         Ok(Some(UnencryptedApiKey { unencrypted }))
     }
 
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -62,7 +62,10 @@ encrypt your Hex API key.
             );
         }
         let password = self.ask_local_password()?;
-        let encrypted = encryption::encrypt_with_passphrase(api_key.as_bytes(), &password)?;
+        let encrypted = encryption::encrypt_with_passphrase(api_key.as_bytes(), &password)
+            .map_err(|e| Error::FailedToEncryptLocalHexApiKey {
+                detail: e.to_string(),
+            })?;
 
         crate::fs::write(&path, &format!("{name}\n{encrypted}"))?;
         println!("Encrypted Hex API key written to {path}");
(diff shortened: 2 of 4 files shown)


compiler-cli/src/hex/auth.rs
```diff
@@ -111,7 +111,10 @@ encrypt your Hex API key.
             return Ok(None);
         };
         let password = self.ask_local_password()?;
-        let unencrypted = encryption::decrypt_with_passphrase(encrypted.as_bytes(), &password)?;
+        let unencrypted = encryption::decrypt_with_passphrase(encrypted.as_bytes(), &password)
+            .map_err(|e| Error::FailedToDecryptApiKey {
+                detail: e.to_string(),
```
lpil, line 116: Do not call `to_string` on our error type please.

compiler-core/src/error.rs
```diff
@@ -321,6 +321,9 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
 
     #[error("Failed to decrypt data")]
     FailedToDecrypt { detail: String },
+
+    #[error("Failed to decrypt API key")]
+    FailedToDecryptApiKey { detail: String },
```
lpil, line 326: Could be any API key! FailedToDecryptLocalHexApiKey or such? The error attribute needs to say this info also.

compiler-core/src/error.rs
```diff
@@ -1487,6 +1490,21 @@ The error from the encryption library was:
                 }]
             }
 
+            Error::FailedToDecryptApiKey { detail } => {
+                let text = wrap_format!("Unable to decrypt the local Hex token with the given password.
```
lpil, line 1494: We call it an API key, not a token 🙏

compiler-core/src/error.rs
```diff
@@ -1487,6 +1490,21 @@ The error from the encryption library was:
...
+The error from the encryption library was:
+
+    {detail}"
+);
+                vec![Diagnostic {
+                    title: "Failed to decrypt Hex token".into(),
```
lpil, line 1500: Here too

lpil (commented):
Thank you!

It looks like the `FailedToDecrypt` is no longer used. Let's remove that error variant because of this.

compiler-cli/src/hex/auth.rs
scristobal: I modified `decrypt_with_passphrase` to return a more generic `DecryptError` 

https://github.com/gleam-lang/gleam/blob/578f87815cabfc5fdf3392495affb0afe770ade1/compiler-core/src/encryption.rs#L29

Unfortunately some extra boilerplate is required as `decrypt_with_passphrase` can fail in two different ways https://github.com/gleam-lang/gleam/blob/578f87815cabfc5fdf3392495affb0afe770ade1/compiler-core/src/encryption.rs#L15-L24

I would love to use thiserror's `#[from]` macro to automatically convert a `DecryptError` into a `FailedToDecryptApiKey` but it is not possible because of `Eq` and `PartialEq` derive macros [here](https://github.com/gleam-lang/gleam/blob/578f87815cabfc5fdf3392495affb0afe770ade1/compiler-core/src/error.rs#L61) and those not being implemented by `DecryptError` variants...

Hence the manual `map_err` and the `.to_string` call. 


---
An alternative will be to return directly a `FailedToDecryptApiKey` from `decrypt_with_passphrase` instead of a `DecryptError`. 

That approach would be more direct and simple, but also would make `decrypt_with_passphrase` a bit less flexible.

I would be happy to refactor if you think simplicity is a better trade-off.

scristobal:
> Thank you!
> 
> It looks like the `FailedToDecrypt` is no longer used. Let's remove that error variant because of this.

I removed that error variant and in exchange introduced `DecryptError` local to `encryption:: decrypt_with_passphrase` see my comment [here](https://github.com/gleam-lang/gleam/pull/4339#discussion_r2007855667)

In a similar fashion, I also went out on a limb and replaced the generic encryption "data" error 

https://github.com/gleam-lang/gleam/blob/fc5daa980ec10fbd200bba30277b182a63107871/compiler-core/src/error.rs#L319-L320

with a narrower version, similar to `FailedToDecryptLocalHexApiKey`

https://github.com/gleam-lang/gleam/blob/578f87815cabfc5fdf3392495affb0afe770ade1/compiler-core/src/error.rs#L319-L320
 

Let me know what do you think!

Thank you for your patience 🙏🏻

lpil:
Is this ready for review? Press the button if so! :)

CHANGELOG.md
```diff
@@ -66,6 +66,9 @@
   project's `gleam.toml` as a JSON file.
   ([Rodrigo Álvarez](https://github.com/Papipo))
 
+- Improve error messages when encrypting and decrypting local Hex AIP key.
```
lpil, line 69:
```suggestion
- Improve error messages when encrypting and decrypting local Hex API key.
```

compiler-core/src/error.rs
```diff
@@ -316,11 +316,11 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
         wrongfully_allowed_version: SmallVersion,
     },
 
-    #[error("Failed to encrypt data")]
-    FailedToEncrypt { detail: String },
+    #[error("Failed to encrypt local Hex APIT key")]
```
lpil, line 319:
```suggestion
    #[error("Failed to encrypt local Hex API key")]
```

compiler-cli/src/hex/auth.rs
lpil: This looks great!! Thank you!!

lpil (commented):
Wonderful! Thank you very much!

scristobal:
@lpil unfortunately I accepted [your suggestion](https://github.com/gleam-lang/gleam/pull/4339#discussion_r2008856424) before seeing your commit, that invalidated auto-merge. I tried to remove the commit by force pushing, but that didn't help either. 

Anyway, hopefully everything is in order now.

lpil (approved):
Thanks again!

Issue #4340, opened 2025-03-13
invalid code produced by convert to pipe code action


matiascr:
Running the "convert to pipe" code action on the following code

```
import gleam/int

fn main() {
  int.add(1, _)
  ^^^^^^^^
}
```

will produce the following (invalid) code:

```
1 |> int.add(_)
```

This happens also on records with more than one field and a capture

```
type Error {
  Error(String)
}

fn main() {
  Error(_)
}
```

turns into this
```
_ |> Error
```

Feel free to close this issue if it's expected that generated code may need tweaking.

lpil:
Oh dear! We should not be offering this code action for function captures

PR #4341, opened 2025-03-13
Changed convert to pipe to not extract functions and records that already contain captures

matiascr:
addresses https://github.com/gleam-lang/gleam/issues/4340


lpil (approved):
Thank you!!

PR #4345, opened 2025-03-14
Fix renaming of aliased and nested modules

### LINKED ISSUE #4344: Rename doesn't properly work for nested modules
I found a bug with the new module value renaming. If you have this code:
```gleam
// sub/mod.gleam
pub fn wibble() { todo }

// main.gleam
import sub/mod

pub fn main() {
  mod.wibble()
  //  ^ trigger rename
}
```
It becomes this:
```gleam
// sub/mod.gleam
pub fn wobble() { todo }

// main.gleam
import sub/mod

pub fn main() {
  mod.wibble()
}
```
Which is incorrect, it should rename the reference too.

GearsDatapacks:
Fixes #4344 
Turns out `imported_modules` stores the locally used names of the modules, not the fully qualified ones.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -342,7 +342,11 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 documentation,
                 contains_echo: echo_found,
                 references: References {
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you

PR #4346, opened 2025-03-14
Check if string patterns are unreachable

giacomocavalieri:
This PR expands the pattern matching decision tree to properly deal with string patterns and check if they are reachable or not. It also lays the foundation to properly do efficient code generation for string patterns, after this I'll start working on the code get part!

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -504,6 +504,16 @@ impl Variable {
         Self { id, type_ }
     }
 
+    /// Builds a `PatternCheck` that checks this variable matches the given pattern.
+    /// So we can build pattern checks the same way we informally describe them:
+    /// ```text
+    /// var is pattern
+    /// ```
+    /// With this builder method would become:
+    /// ```rs
+    /// var.is(pattern)
+    /// ```
+    ///
     fn is(&self, pattern: Id<Pattern>) -> PatternCheck {
         PatternCheck {
             var: self.clone(),
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -88,6 +88,7 @@ use crate::{
 use ecow::EcoString;
 use id_arena::{Arena, Id};
 use itertools::Itertools;
+use radix_trie::{Trie, TrieCommon};
 use std::{
     cell::RefCell,
     collections::{HashMap, HashSet, VecDeque},
@@ -302,7 +303,7 @@ pub enum Pattern {
     },
     StringPrefix {
         prefix: EcoString,
-        rest: Id<Pattern>,
+        rest_pattern: Id<Pattern>,
     },
     Assign {
         name: EcoString,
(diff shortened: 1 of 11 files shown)


compiler-core/src/exhaustiveness.rs
```diff
@@ -312,15 +313,15 @@ pub enum Pattern {
...
-        arguments: Vec<Id<Pattern>>,
+    Variant {
+        index: usize,
+        args_patterns: Vec<Id<Pattern>>,
     },
     List {
-        first: Id<Pattern>,
-        rest: Id<Pattern>,
+        first_pattern: Id<Pattern>,
+        rest_pattern: Id<Pattern>,
```
lpil, line 324: Undo these name changes please, do don't do hungarian notation.

compiler-core/src/exhaustiveness.rs
```diff
@@ -441,7 +442,7 @@ pub enum RuntimeCheck {
         args: Vec<Variable>,
     },
     EmptyList,
-    NonEmptyList {
+    List {
```
lpil, line 445: Undo this change, it's under-specified now. Both the empty and non-empty variants are lists, and this name fails to specify what is distinctive about this variant.

A list pattern would check that it is any list, not that it is a non-empty list.

compiler-core/src/exhaustiveness.rs
```diff
@@ -519,6 +505,13 @@ impl Variable {
     fn new(id: usize, type_: Arc<Type>) -> Self {
         Self { id, type_ }
     }
+
+    fn is(&self, pattern: Id<Pattern>) -> PatternCheck {
```
lpil, line 509: Could you document what this is please 🙏

lpil (commented):
Amazing work!! I've not done a proper review yet, but I've left some very small comments with the few mins I've had to look at this

giacomocavalieri:
Thank you! Should have addressed all the comments

compiler-core/src/exhaustiveness.rs
```diff
@@ -312,13 +313,13 @@ pub enum Pattern {
         name: EcoString,
     },
     Tuple {
-        elements: Vec<Id<Pattern>>,
+        elems: Vec<Id<Pattern>>,
```
lpil, line 316: No abbreviations please, `elements` and `arguments`. Though perhaps `arguments` should be `values` or `fields`?

compiler-core/src/exhaustiveness.rs
```diff
@@ -1095,6 +1010,132 @@ impl<'a> Compiler<'a> {
         }
     }
 
+    /// Comes up with new pattern cecks that have to match in case a given
```
lpil, line 1013:
```suggestion
    /// Comes up with new pattern checks that have to match in case a given
```

lpil (commented):
Code looks great! Really impressive work!

I've left some tiny nitpicks RE names

compiler-core/src/exhaustiveness.rs
giacomocavalieri: Ok, fixed! I've renamed `args` to `fields`

lpil (approved):
Amazing!!! Thank you!!

PR #4347, opened 2025-03-14
Remove echo code action

giacomocavalieri:
This PR implements a code action to remove `echo` debug printing from the hovered expression

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -154,7 +154,8 @@
   a module and across multiple modules.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
-- The language server now offers a code action to remove `echo`. For example:
+- The language server now offers a code action to remove all `echo`s in a
+  module. For example:
 
   ```gleam
   pub fn main() {
@@ -162,10 +163,11 @@
     |> echo
     // ^^^^ If you put your cursor over here
     |> list.filter(int.is_even)
+    |> echo
   }
   ```
 
-  Triggering the code action would remove the `echo` pipeline step:
+  Triggering the code action would remove all the `echo` pipeline steps:
 
   ```gleam
   pub fn main() {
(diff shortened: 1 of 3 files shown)


inoas:
could there be a way to remove it from an entire module or whole src code? gleam fix?

giacomocavalieri:
It seems I'm running into an issue where this is slowing down the LS abnormally, I'll have to look into it before it can be merged!

giacomocavalieri:
> could there be a way to remove it from an entire module or whole src code? gleam fix?

Could be a nice addition, I haven't thought of it that much, for now it just deletes the one under the cursor

lpil:
This needs to remove all of them to be useful I think. It's more key presses to run the code action than to delete the echo yourself if it only works for a single one.

giacomocavalieri:
All echos under the selection or in the entire module?

lpil:
The entire module please. Maybe the entire package?

giacomocavalieri:
Ok! When should the action trigger? Anywhere in the file? Sounds a bit annoying having that pop up every time if you have an echo in a module

lpil:
When on an echo sounds good

giacomocavalieri:
Yeah sounds good to me, just to make sure we're on the same page: once on an echo we should offer an action "remove all `echo`s from module" which will remove all echos, not just the hovered one?

lpil:
Yes please

giacomocavalieri:
This should be ready now!

inoas:
`gleam fix echo` could remove remove all echos from the project?

CHANGELOG.md
```diff
@@ -154,6 +154,45 @@
   a module and across multiple modules.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The language server now offers a code action to remove `echo`. For example:
```
lpil, line 157: The changelog makes it sounds a bit like it removes only a single echo

lpil (commented):
Super nice! Couple tiny notes inline

lpil (approved):
Thank you!!

PR #4348, opened 2025-03-14
LSP: Show references to values

### LINKED ISSUE #1541: LSP: Show references to values

GearsDatapacks:
Closes #1541
This PR uses the infrastructure introduced for value renaming to allow finding value references.
I've extracted the shared `match` expression in `prepare_rename` and `rename` into a new function, so the same long pattern match wouldn't need to be repeated a third time for this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,10 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The language server now supports finding references to values, both within
+  a module and across multiple modules.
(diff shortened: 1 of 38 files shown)


lpil (commented):
Brilliant! Thank you!

I've found a bug: when a value is imported in an unqualified fashion and then aliased those uses under that aliased name do not get listed as a reference. These would not want to be renamed, but they are references, so they should be included.

lpil (approved):
Thank you!!!

PR #4349, opened 2025-03-15
16-bit floats in bit arrays on JavaScript

richard-viney:
The JavaScript code for this was taken from my https://github.com/richard-viney/ieee_float Gleam package.

JS compilation error assertions were removed because they're no longer used.


lpil (approved):
Thank you!!

PR #4351, opened 2025-03-15
Remove unused `Lexer` struct field `location`

scristobal:
The `location` field in the `Lexer` struct is initialized to `0` but never really used/updated.

After removing it, all tests still pass.

I know it is a tiny change, but I hope can help other newcomers to the language :D


lpil (approved):
Thank you!

PR #4354, opened 2025-03-16
Tweak generated code for JSON encoders

### LINKED ISSUE #4353: Tweak generated code for "Generate JSON encoder" action
Currently, one thing I like about the "Generate dynamic decoder" action is the fact that if you change the type that is being decoded, the decoder function will contain a compile error. For example, let's say you have this type:
```gleam
type Person {
  Person(name: String, age: Int)
}
```
The generated decoder will look like this:
```gleam
fn person_decoder() -> decode.Decoder(Person) {
  use name <- decode.field("name", decode.string)
  use age <- decode.field("age", decode.int)
  decode.success(Person(name:, age:))
}
```
Now let's say you want to add a new field, `job: String`. That decoder function will now error because you are trying to construct a `Person` without providing all the fields. This signals to the programmer that they need to delete the decoder and rerun the code action.
However, this is what the generared JSON encoder looks like for the above type currently.
```gleam
fn encode_person(person: Person) -> json.Json {
  json.object([
    #("name", json.string(person.name)),
    #("age", json.int(person.age)),
  ])
}
```
Even with a new field, this compiles fine. Nothing here is ensuring all fields are covered, so it's possible that someone will forget to update the encoder, and end up introducing a bug in the code.
To fix this, I propose a slight change in the generated code. Instead, it would look like this:
```gleam
fn encode_person(person: Person) -> json.Json {
  let Person(name:, age:) = person
  json.object([
    #("name", json.string(name)),
    #("age", json.int(age)),
  ])
}
```
This way, adding a new field will cause the function to fail to compile, and help people remember to update their encoders as well as decoders when the type changes.

GearsDatapacks:
Closes #4353


compiler-core/src/ast.rs
```diff
@@ -247,6 +247,18 @@ impl<A> RecordConstructor<A> {
     pub fn put_doc(&mut self, new_doc: (u32, EcoString)) {
         self.documentation = Some(new_doc);
     }
+
+    pub fn argument_labels(&self) -> Vec<&str> {
```
lpil, line 251: This method doesn't seem generally useful as it discards non-labelled arguments, so you can only safely use it when you know for a fact that all the arguments are labelled. Let's inline this at the place it is used.

lpil (commented):
Thank you! One small note inline.

PR #4357, opened 2025-03-16
Fix rename range for type variant definitions

### LINKED ISSUE #4356: Rename range is wrong for definitions of type variants
When renaming a type variant from its definition, e.g.:
```gleam
type Person {
  Person(name: String, age: Int)
//^ trigger rename
}
```
It will use `Person(name: String, age: Int)` as placeholder text instead of just `Person`.
I have a fix for this locally, but I'm going to wait for #4348 to be merged because it changes the code surrounding this and it makes it easier to fix this.

GearsDatapacks:
Fixes #4356 
I've fixed that bug, and also added the rename location to the test snapshots to hopefully make these bugs easier to catch in future.


lpil (approved):
Thank you!!!!

PR #4363, opened 2025-03-16
Remove many abbreviations

giacomocavalieri:
This PR removes the use of various abbreviations:
- `r#gen` was renamed to `this` to follow an already existing convention in the codebase
- `elems` was renamed everywhere to `elements`
- when iterating over `elements` the loop variable is now consistently called `element` instead of `elem` or `e`
- there was one stray occurrence of `typ` that I renamed to `type_` to follow the already existing convention
- `retrn` was renamed to `return_` to be consistent with the choice of adding an underscore after names that would otherwise be Rust keywords
- `t` has been renamed to `type_` when it references a type
- I've renamed a couple other single-letter lambda arguments to use more descriptive names
- I did quite a bit of renaming for the arguments of the functions is `ast_folder`, previously a lot of those were single letter names

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -397,12 +397,17 @@ pub fn gleam_files(dir: &Utf8Path) -> impl Iterator<Item = Utf8PathBuf> + '_ {
     ignore::WalkBuilder::new(dir)
         .follow_links(true)
         .standard_filters(false)
-        .filter_entry(|e| !is_gleam_build_dir(e))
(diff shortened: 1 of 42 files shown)


GearsDatapacks:
I think there is at least one place where we use `retrn` as an abbreviation for `return`. Maybe we should rename that to `return_`?

giacomocavalieri:
Oh yeah great idea, I'll rename that as well

lpil (approved):
Thank you!

PR #4367, opened 2025-03-17
Make sure a variant is inferred when doing `let assert`

### LINKED ISSUE #4350: Variant inference does not work with `let assert`
When using `let assert` the compiler will not use the information that a variable should be narrowed when performing record field access or update:

Give this type:

```gleam
type Example {
  Wibble
  Wobble(x: Int, y: Int)
}
```

The following two snippets fail to type check:

```gleam
fn access_reproduction(example: Example) {
  let assert Wobble(x:, ..) = example

  example.x
}
```

```gleam
fn update_reproduction(example: Example) {
  let assert Wobble(x:, ..) = example

  Wobble(..example, x: x + 1)
}
```

With these errors:

```
error: Unknown record field
    ┌─ /...
    │
xyz │   example.x
    │           ^ This field does not exist

    Example

It does not have fields that are common across all variants.

```

```
error: Unsafe record update
    ┌─ /...
    │
xyz │   Wobble(..example, x: x + 1)
    │            ^^^^^^^ I'm not sure this is always a `Wobble`


This value cannot be used to build an updated `Wobble` as it could be some
other variant.

Consider pattern matching on it with a case expression and then
constructing a new record with its values.
```

With the variant inference feature I would expect both examples to work

giacomocavalieri:
This PR closes #4350

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -219,6 +219,9 @@
   if used inside a record update.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where variant inference wouldn't work on `let assert` assignments.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ## v1.9.1 - 2025-03-10
 
 ### Formatter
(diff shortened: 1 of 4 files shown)


lpil (approved):
That was quick! Thank you!

PR #4368, opened 2025-03-17
Do not loose information about definition groups

giacomocavalieri:
This will be needed to implement mutual recursion TCO on the JS target. At the moment we throw away the different definitions groupings after analysis by flattening all the groups when storing those in the typed module. In this PR I change the internal module representation to make sure we store each group separately; this way code generation will be able to work with groups of related functions!

As a side note I've also done quite some renaming here: the codebase was calling a module's definitions `statements` despite those being definitions (and their type is indeed `Un/TypedDefinition`). And the `Statement` type and `statement` name is already widely used to refer to the statements that are part of a function's body, so now we use distinct names to refer to module definitions and statements

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -135,13 +135,15 @@ fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
         return Ok(());
     }
 
-    let definitions = &module.ast.definitions;
-
-    if definitions.len() > 2 {
+    if module.ast.total_definitions() > 2 {
         return Ok(());
     }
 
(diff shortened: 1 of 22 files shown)


lpil:
Heya! How come this PR doesn't have the feature itself in it?

giacomocavalieri:
I was afraid it would turn up to be huge so I was thinking of splitting this into its own PR

lpil:
It's no trouble, this PR isn't large, and one can read the commit history.

lpil:
Still want this PR @giacomocavalieri ? :)

giacomocavalieri:
Nope! I've started from scratch, I figured taking care of the conflicts would be harder than a fresh start 😁

PR #4374, opened 2025-03-20
Fix renaming of types in variant constructor definitions

### LINKED ISSUE #4372: Rename Types incorrectly replace in named variants of types in other modules
if I have:

main.gleam
```gleam

pub type MyType {
    MyType
}
fn main() {
  let my_type = MyType
}
```
other_module.gleam
```gleam

pub type OtherType {
    OtherType(t: main.MyType)
}
```

When I try to rename MyType in main.gleam to MyTypeRenamed, other_module duplicates the diff, so it would be `OtherType(main.MyTypeRenamedRenamed)`.

If I rename it back to `MyType`, it removes the parenthesis from other module: `OtherType(main.MyType`

GearsDatapacks:
Fixes #4372 
The issue occurred because in the `custom_type_accessors` function, we traverse the `TypeAst` twice after we've already done so in `register_values_from_custom_type`. This caused 3 references to be registered for any types in the arguments, rather than just 1.
To fix this, I've changed it so that we use the type information we already have instead of rebuilding it continuously from the AST.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -993,25 +993,6 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             .type_
             .clone();
 
-        let Accessors {
-            shared_accessors,
-            variant_specific_accessors,
-        } = custom_type_accessors(constructors, &mut hydrator, environment, &mut self.problems)?;
-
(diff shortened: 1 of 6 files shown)


lpil (approved):
Thank you

PR #4378, opened 2025-03-21
Rework bit arrays

giacomocavalieri:
This started as a (failed) attempt at trying to reduce duplication in the code dealing with bit arrays: basically everything is copy-pasted for both expression and constant bit arrays. I couldn't figure out how to deal with that but I've still refactored the code a bit to remove _some_ duplication and make code a little less nested in a couple of places. I think the code improvements are still worth merging this


lpil (approved):
Thank you

PR #4380, opened 2025-03-22
Use a call graph for usage detection

### LINKED ISSUE #2634: Replace use detection with a call graph
I noticed that a private recursive function is not marked as unused.
For example this one raises no warnings:
```gleam
fn wobble() {
  wobble()
}
```

### LINKED ISSUE #3552: Unused import warning on constructor if type alias with the same name exists

GearsDatapacks:
This PR closes #2634 and fixes #3552 
Usage detection of local variables still uses the old system, because I couldn't find a way that made sense to integrate them into the call graph. Because everything else uses the call graph now though, this shouldn't cause any bugs.


compiler-core/src/type_/tests.rs
```diff
@@ -213,6 +214,7 @@ macro_rules! assert_warning {
             crate::build::Target::Erlang,
             None
         );
+        assert!(!warning.is_empty());
```
lpil, line 217: Nice!!

compiler-core/src/reference.rs
```diff
@@ -24,13 +25,48 @@ pub struct Reference {
...
+    ImportedValue,
+    ImportedModule,
+    ModuleAlias,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+enum EntityLayer {
+    Type,
+    Value,
+    Shadowed,
```
lpil, line 51: What is this variant for? Could you document these please 🙏

compiler-core/src/reference.rs
```diff
@@ -24,13 +25,48 @@ pub struct Reference {
...
 pub type ReferenceMap = HashMap<(EcoString, EcoString), Vec<Reference>>;
 
+#[derive(Debug, Clone, Copy)]
+pub struct EntityInformation {
+    pub origin: SrcSpan,
+    pub kind: EntityKind,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum EntityKind {
```
lpil, line 35: Could you document what this type is used for please 🙏

compiler-core/src/reference.rs
```diff
@@ -44,51 +80,243 @@ impl ReferenceTracker {
...
+            }
+        }
+
         self.type_references
             .entry((module, name))
             .or_default()
             .push(Reference { location, kind });
     }
+
+    pub fn register_type_reference_in_call_graph(&mut self, name: EcoString) {
```
lpil, line 283: What's the difference between this and register_type_reference? It's not clear to me why one is used some times and the other other times.

lpil (commented):
Amazing!! I love this!

I've left some small notes and questions inline 🙏

GearsDatapacks:
I have now addressed your comments. I've also renamed a few things I forgot to rename before, such as `unused_values` -> `unused`. That's due to the old way that this code worked, so the names should be a bit more clear now.

lpil (approved):
I'm so excited for this one!! Thank you!!

PR #4382, opened 2025-03-22
Use decision trees for code generation on the js target

### LINKED ISSUE #3969: Use decision trees for case expression code generation on the JS target
Code generation for js doesn't take advantage of the decision tree generative if-else branches that check the same conditions more than once. It'd be cool to have this optimisation

### LINKED ISSUE #4221: Small bug in javascript code generation
# Steps to reproduce
Write something like the following:
```gleam
      let add_line_point = case row_pos, kv_pair.data_type {
        decl_types.RowLeft, Some(data_type.NodeRef(node_idx))
          if node_idx == model.center_node
        -> True
        decl_types.RowCenter, Some(data_type.NodeRef(_))
        | decl_types.RowCenter, Some(data_type.EnumRef(_))
        -> True
        _else, _any -> False
      }
```

It will produce the following code in Javascript:
```javascript
        let add_line_point = (() => {
          let $ = kv_pair.data_type;
          if (row_pos instanceof RowLeft && $ instanceof Some && $[0] instanceof NodeRef && $[0][0] === model.center_node) {
            let node_idx = $[0][0];
            return true;
          } else if (row_pos instanceof RowCenter && $ instanceof Some && $[0] instanceof NodeRef) {
            return true;
          } else if (row_pos instanceof RowCenter && $ instanceof Some && $[0] instanceof EnumRef) {
            return true;
          } else {
            return false;
          }
        })();
```

Inside the first if statement there is an empty `let node_idx=...` statement.  I don't know if this is just done for clarity but maybe it can be changed to a comment `// node_idx == $[0][0]

I just saw it when debugging some javascript code, so decided to make an issue, but if it works as designed you can close.  

# Version
gleam 1.8.0

### LINKED ISSUE #4387: Crash for bit array pattern on JS target
This code:
```gleam
pub fn main() {
  case <<12:size(1)>> {
    <<1.2:float, _>> -> 1
    _ -> 2
  }
}
```
Throws an exception on the JavaScript target: `Error: Invalid bit array slice: start = 0, end = 64, bit size = 1 `. It looks like the pattern matching code doesn't check the length of the bit array is sufficient to call the `bitArraySliceToFloat` function

giacomocavalieri:
This PR closes #3969, closes #4221 and closes #4387

This is still a WIP as I haven't taken care of bit array patterns yet, but it's ready for everything else if anyone fancies giving this a look in the meantime.

What's missing:
- [X] Use the decision tree for cases
- [X] Reuse the decision tree for let restructuring and let assert 
- [x] Support bit arrays in the decision tree


compiler-core/src/javascript/expression.rs
```diff
@@ -1411,6 +1302,490 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+            None => {
+                let size_value: usize = if segment.type_.is_int() { 8 } else { 64 };
+                (Some(BigInt::from(size_value)), docvec![size_value])
+            }
+        };
+
+        Ok(SizedBitArraySegmentDetails { size, size_value })
+    }
+
+    pub(crate) fn guard(&mut self, guard: &'a TypedClauseGuard) -> Output<'a> {
```
giacomocavalieri, line 1550: All the code here up to line ~1900 is basically moved as-is from the pattern module where it was previously

lpil:
Please give PRs titles, just a number means folks have to look them up to know what this is

giacomocavalieri:
Oops my mistake!

giacomocavalieri:
This is ready for review now!

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__strings__string_prefix_assignment_with_multiple_subjects.snap
```diff
@@ -17,9 +17,6 @@ export function go(x) {
   if (x.startsWith("1")) {
     let prefix = "1";
     return prefix;
-  } else if (x.startsWith("11")) {
-    let prefix = "11";
-    return prefix;
```
lpil, line 22: Wow

giacomocavalieri: Mind blowing, innit?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__case_on_error.snap
```diff
@@ -0,0 +1,30 @@
...
+----- COMPILED JAVASCRIPT
+import { Error } from "../gleam.mjs";
+
+function a_result() {
+  return new Error(1);
+}
+
+export function main() {
+  let $ = a_result();
+  if (!$.isOk()) {
```
lpil, line 25: I think we can remove the special case for result and use instanceof like the other variants?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__keyword_var.snap
```diff
@@ -34,18 +34,26 @@ export function main() {
...
-    return false;
-  } else if (10 === 5) {
-    return true;
-  } else if (var$ > 5) {
-    let while$1 = var$;
-    return false;
   } else {
-    let class$1 = var$;
-    return false;
+    if (isEqual(toList([class$]), toList([5]))) {
```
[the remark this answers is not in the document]
giacomocavalieri: Not in general, sometimes there could be variable definitions between that `else` and the next `if`, making it impossible to turn it into an `else if`. It would be quite a bit of work and I don't think it would really be worth the effort

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__case_on_error.snap
giacomocavalieri: We would still need to special case results as we'd have to make sure that we import the constructors from the prelude to perform the `instanceOf` check. But if you want I can switch to instance of checks for these as well

lpil: Yes please

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__keyword_var.snap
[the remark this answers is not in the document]
lpil: I meant for the case in which there are no assignments specifically. Is that challenging?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__match_literal_unaligned_float.snap
```diff
@@ -0,0 +1,33 @@
...
+
+
+----- COMPILED JAVASCRIPT
+import { makeError, bitArraySliceToFloat } from "../gleam.mjs";
+
+function go(x) {
+  let n = 1;
+  if (
+    x.bitSize < n ||
+    x.bitSize < 64 + n ||
```
lpil, line 20: Previously Gears made it not output these redundant checks, it would be fab to retain that!

lpil (commented):
Some comments as I'm half way through reviewing!

compiler-core/src/exhaustiveness.rs
```diff
@@ -588,6 +589,17 @@ impl RuntimeCheck {
...
     }
+
+    /// Returns all the bit array segments referenced in this check.
+    /// For each segment it returns its name and the read action used to access
+    /// such segment.
+    ///
+    pub(crate) fn referenced_segment_patterns(&self) -> Vec<(&EcoString, &ReadAction)> {
+        match self {
+            RuntimeCheck::BitArray { test } => test.referenced_segment_patterns(),
+            _ => vec![],
```
lpil, line 600: No catch alls please

compiler-core/src/javascript/expression.rs
```diff
@@ -120,6 +117,11 @@ pub(crate) struct Generator<'module, 'ast> {
     /// ```
     ///
     statement_level: Vec<Document<'ast>>,
+
+    /// This will be true if we've generated a statement that is guaranteed to
+    /// throw. This means we can stop code generation for all following statements
+    /// in the same block!
+    pub statement_always_throws: bool,
```
lpil, line 124: I think this should be set to true for `panic` and `todo`. I can't see that having been done, has it?

compiler-core/src/ast.rs
```diff
@@ -3028,6 +3093,10 @@ pub struct Assignment<TypeT, ExpressionT> {
     pub value: Box<ExpressionT>,
     pub pattern: Pattern<TypeT>,
     pub kind: AssignmentKind<ExpressionT>,
+    pub compiled_case: CompiledCase,
+    /// This will be true for assignments that are automatically generated by
+    /// the compiler.
+    pub is_generated: bool,
```
lpil, line 3099: How come this is a field now?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__keyword_var.snap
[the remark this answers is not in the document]
giacomocavalieri: I've opened an issue for this! https://github.com/gleam-lang/gleam/issues/4522

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__case_match_literal_float.snap
```diff
@@ -0,0 +1,33 @@
...
+}
+
+
+----- COMPILED JAVASCRIPT
+import { bitArraySliceToFloat } from "../gleam.mjs";
+
+function go(x) {
+  if (x.bitSize >= 64) {
+    if (bitArraySliceToFloat(x, 0, 64, true) === 1.4) {
+      if (x.bitSize === 72) {
```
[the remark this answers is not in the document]
giacomocavalieri: I've opened an issue for this one! https://github.com/gleam-lang/gleam/issues/4524

giacomocavalieri:
I've rebased on main adding support for aliased segments for the bit array code generation. I've also found a small bug in the analysis of bit array aliased segments where it would wrongly accept a variable sized utf8 string. I've fixed it making sure it's rejected!

Norlock:
Thanks a lot for all the work, how much will this roughly improve on js?

giacomocavalieri:
Since this is improving the pattern matching logic it will depend on how much your code relies on complex pattern matching expressions. We've tested this on Lustre's DOM diffing algorithm (which is now written entirely in Gleam) and measured an impressive ~30% speedup!

I've also been benchmarking this on the todomvc lustre app and could measure a ~20% speedup, to me it's amazing seeing this kind of speedup on a "real world" app and not just in benchmarks!

Norlock:
Ok that's insane! I have a big Lustre app, so knowing this is really nice. After Lustre v5 and updates here my app will feel really fast.

PR #4384, opened 2025-03-23
Fix bug with code generation of case clauses

### LINKED ISSUE #4383: Fatal bug in javascript code generation of case clauses
I've found a bug that I introduced in #4337.
If the branch of a case expression needs to be wrapped in a block, for example:
```gleam
case bool {
  True -> a |> b
  False -> c |> d
}
```
This will be compiled to something like:
```js
let _block;
let _pipe = a;
_block = b(_pipe);

let _block;
let _pipe = c;
_block = d(_pipe);

if (bool) {
  return _block;
} else {
  return _block;
}
```
This is of course incorrect, those two statements should appear within the body of the `if` statement.

I'm working on a fix at the moment, but thought I'd open an issue first.

GearsDatapacks:
Fixes #4383 
I just forgot to take `statement_level` if the case clause guard wasn't a block.

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -211,14 +211,18 @@ impl<'module, 'a> Generator<'module, 'a> {
             Statement::Assignment(assignment) => self.assignment(assignment),
             Statement::Use(_use) => self.expression(&_use.call),
         }?;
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!!

PR #4392, opened 2025-03-24
Warn when asserting on variants which have been inferred

### LINKED ISSUE #4389: Show warning for `let assert` that is guaranteed to fail
It would be nice if the compiler warned us when a `let assert` is guaranteed to fail because of variant inference:
```gleam
let assert Wibble(n) = Wobble(1) 
```

GearsDatapacks:
Closes #4389


compiler-core/src/warning.rs
```diff
@@ -940,6 +940,24 @@ hidden from the package's documentation.",
                     }),
                 },
 
+                type_::Warning::AssertAssignmentOnInferredVariant { location } => Diagnostic {
+                    title: "Assertion on inferred variant".into(),
+                    text: wrap(
+                        "This assertion is will always crash since it matches on a variant which is never present.",
```
richard-viney, line 946: Small typo here: "is will always" should be "will always"

compiler-core/src/type_/expression.rs
```diff
@@ -1543,9 +1544,22 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            (AssignmentKind::Assert { .. }, Err(_)) => {
+                // There is only one pattern to match, so it is index 0
+                match output.is_reachable(0) {
+                    Reachability::Unreachable(UnreachableCaseClauseReason::ImpossibleVariant) => {
+                        self.problems
+                            .warning(Warning::AssertAssignmentOnInferredVariant {
+                                location: pattern.location(),
+                            })
+                    }
+                    // A duplicate pattern warning should not happen, since there is only on pattern.
```
richard-viney, line 1558: Typo: "on" => "one"

lpil (approved):
Thank you!

PR #4397, opened 2025-03-25
Fix various JavaScript code generation bugs

### LINKED ISSUE #4393: JavaScript codegen bug regarding `let assert`
I noticed a bug in JavaScript code generation: `let assert` doesn't perform a runtime check if it's the only expression in a block:
```gleam
fn assert_ok(result: Result(a, b)) -> Result(a, b) {
  {
    let assert Ok(_) = result
  }
}
```
This gets compiled to the same as:
```gleam
fn assert_ok(result: Result(a, b)) -> Result(a, b) {
  result
}
```
Which is incorrect.

### LINKED ISSUE #4394: JavaScript codegen bug when a `let` statement is the last in a block
For example:
```gleam
fn wibble() {
  let a = {
    let b = 1
    let c = b + 1
  }
  a
}
```
Doesn't properly return the value of `c` from the block.

### LINKED ISSUE #4395: JavaScript codegen bug with custom `let assert` message
The custom panic message for `let assert` is being lifted into statement-level when it shouldn't be.
For example:
```gleam
fn side_effects() {
  echo "some message"
}

pub fn main() {
  let assert Error(Nil) = echo Ok(10)
    as {
    let message = side_effects()
    message
  }
}
```
When run, this prints `some message` before `Ok(10)`

GearsDatapacks:
This PR fixes #4393, fixes #4394 and fixes #4395
I thought I'd put these all in one PR, as they are relatively small, relatively simple changes to the same part of the codebase. Let me know if I should split this into 3 PRs.
Only the first one is reported in the changelog as the latter two were introduced in this version when I modified code generation of blocks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -292,6 +292,10 @@
   when it could use the cache from previous compilations.
   ([Louis Pilfold](https://github.com/lpil))
 
+- Fixed a bug where `let assert` would not assert that the given value matched
+  the pattern if it was the only expression inside a block.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.9.1 - 2025-03-10
 
 ### Formatter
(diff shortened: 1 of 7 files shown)


lpil (approved):
Beautiful, thank you

Issue #3551, opened 2024-08-21
Improve `import a.b` suggestion


sobolevn:
Right now if a user (me 😞) tries to write `import gleam.io` (like how I am used to from Python), it would be a syntax error:

```
error: Syntax error
  ┌─ /src/main.gleam:1:14
  │
1 │ import gleam.io
  │              ^^ I was not expecting this

Found a name, expected one of: 
- `{`
```

Following this suggestion would navigate a person towards to: `import gleam.{io}`, which would raise:

```
error: Unknown module field
  ┌─ /src/main.gleam:1:15
  │
1 │ import gleam.{io}
  │               ^^

The module `gleam` does not have a `io` value.
```

Basically, this is not what we want. 

There are two major cases:
1. `import a/b` where `b` is submodule of `a` package
2. `import a.{b}` where `b` is a value in module `a`

I think that we should add a hint about both of these cases, so my initial error message would look like:

```
error: Syntax error
  ┌─ /src/main.gleam:1:14
  │
1 │ import gleam.io
  │              ^^ I was not expecting this

This syntax for import is not correct. Probably you meant:
- `import gleam/io` to import `io` module from `gleam` namespace
- `import gleam.{io}` to import `io` value from `gleam` module
```

Better wording is always welcome! I am not the best with picking the right words in a foreign language :)

What do others think? Is it helpful?

giacomocavalieri:
I think that's a great idea, we could also try and be more specific with the error message, for example if one writes `a.b` and module `a` actually has a value called `b` we could show the suggestion `a.{b}`

lpil:
Wow, what a fantastic idea! I agree with both of you.

vkobinski:
> I think that's a great idea, we could also try and be more specific with the error message, for example if one writes `a.b` and module `a` actually has a value called `b` we could show the suggestion `a.{b}`

For this to work, in this part of [part of the code](https://github.com/gleam-lang/gleam/blob/5406c800454a608e61d447ccd5799f10ae0e8149/compiler-core/src/parse.rs#L2414-L2488) I think the Parser would need to have some context or information about the modules.

vkobinski:
This change needs to be made in this [file](https://github.com/gleam-lang/gleam/blob/5406c800454a608e61d447ccd5799f10ae0e8149/compiler-core/src/parse/error.rs#L198). 
I think we’ll need to change the structure of this function to handle just this case. Maybe this case can be handled elsewhere.

[...]

PR #4401, opened 2025-03-28
Feature/error pythonic import

Zij-IT:
This is a minimal improvement to address #3551. I don't have the knowledge in the code-base in order to implement the last suggestion, however I figured that a perfect solution shouldn't block a good solution.

Thus, here is hopefully the first of many contributions :)

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -2706,7 +2706,7 @@ where
     fn parse_import(&mut self, import_start: u32) -> Result<Option<UntypedDefinition>, ParseError> {
         let mut start = 0;
         let mut end;
-        let mut module = String::new();
+        let mut module = EcoString::new();
         // Gather module names
         loop {
             let (s, name, e) = self.expect_name()?;
@@ -2742,11 +2742,11 @@ where
         let mut unqualified_types = vec![];
 
         if let Some((dot_start, dot_end)) = self.maybe_one(&Token::Dot) {
-            let _ = self.expect_one(&Token::LeftBrace).map_err(|e| {
+            if let Err(e) = self.expect_one(&Token::LeftBrace) {
                 // If the module does contain a '/', then it's unlikely that the user
                 // intended for the import to be pythonic, so skip this.
                 if module.contains('/') {
-                    return e;
+                    return Err(e);
                 }
 
                 // Catch `import gleam.io` and provide a more helpful error...
@@ -2755,17 +2755,17 @@ where
                     ..
                 } = &e.error
                 else {
-                    return e;
+                    return Err(e);
                 };
 
-                ParseError {
-                    error: ParseErrorType::PythonicImport {
-                        module: module.as_str().into(),
+                return Err(ParseError {
+                    error: ParseErrorType::IncorrectImportModuleSeparator {
+                        module,
                         item: name.clone(),
                     },
                     location: SrcSpan::new(dot_start, dot_end),
-                }
-            })?;
+                });
+            };
 
             let parsed = self.parse_unqualified_imports()?;
             unqualified_types = parsed.types;
@@ -2797,7 +2797,7 @@ where
             },
             unqualified_values,
             unqualified_types,
-            module: module.into(),
+            module,
             as_name,
             package: (),
         })))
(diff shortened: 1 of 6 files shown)


changelog/v1.9.md
```diff
@@ -80,6 +80,10 @@
   size.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Gleam will now issue a help message when attempting to import using python's
```
lpil, line 83:
```suggestion
- Gleam will now issue a help message when attempting to import using Python's
```

compiler-core/src/parse.rs
```diff
@@ -2686,8 +2686,26 @@ where
         let mut unqualified_values = vec![];
         let mut unqualified_types = vec![];
 
-        if self.maybe_one(&Token::Dot).is_some() {
-            let _ = self.expect_one(&Token::LeftBrace)?;
+        if let Some((dot_start, dot_end)) = self.maybe_one(&Token::Dot) {
+            let _ = self.expect_one(&Token::LeftBrace).map_err(|e| {
```
lpil, line 2690: This looks like it'll show the error when there's _any_ unexpected token, not specifically when the Python syntax is used.

compiler-core/src/parse/error.rs
```diff
@@ -411,6 +421,10 @@ pub enum ParseErrorType {
     TypeConstructorNoArguments,           // let a : Int()
     TypeDefinitionNoArguments,            // pub type Wibble() { ... }
     UnknownAttributeRecordVariant, // an attribute was used that is not know for a custom type variant
+    PythonicImport {
+        mod_or_namespace: EcoString,
+        importee: EcoString,
```
lpil, line 426: What's an importee?

Document this error please 🙏

compiler-core/src/parse/error.rs
```diff
@@ -411,6 +421,10 @@ pub enum ParseErrorType {
     TypeConstructorNoArguments,           // let a : Int()
     TypeDefinitionNoArguments,            // pub type Wibble() { ... }
     UnknownAttributeRecordVariant, // an attribute was used that is not know for a custom type variant
+    PythonicImport {
+        mod_or_namespace: EcoString,
```
lpil, line 425:
```suggestion
        module: EcoString,
```

lpil (commented):
Great, thank you! I've left some notes inline. Some tests will need to be added too. Snapshot test will work well here

compiler-core/src/parse/error.rs
Zij-IT: This is me probably abusing the English language since `ee` usually indicates the subject of the verb-stem. So in this case, it's the item being imported.

Given the recommendation to rename `mod_or_namespace` to `module`, I think it's likely clearer to rename `importee` to `item`

compiler-core/src/parse.rs
lpil: Sorry, I mispoke there. What I mean is that `import one/two.three` is not a Python import

Zij-IT:
Alright :) As far as I can tell the points that you made are now resolved. Please let me know if there's a stylistic issue with any of the commits.

compiler-core/src/parse/error.rs
```diff
@@ -338,6 +338,16 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
...
                 vec!["Hint: Did you mean `@deprecated`?".into()],
             ),
+            ParseErrorType::PythonicImport { module, item } => (
+                "I was expecting either `/` or `.{` here.",
+                vec![
+                    "This syntax for an import is incorrect. Perhaps you meant:".into(),
+                    format!(" - `import {module}/{item}` to import the `{item}` \
+module from the `{module}` namespace"),
+                    format!(" - `import {module}.{{{item}}}` to import the \
+`{item}` module from the `{module}` module"),
```
andho, line 348: In `import {module}.{{{item}}}` the `{{{item}}}` is not a "module" but a value.
Example `import gleam.{io}` will not import the io module but give an error that "The module `gleam` does not have a `io` value".

```suggestion
`{item}` value from the `{module}` module"),
```

Zij-IT: Done :)

PR #4404, opened 2025-03-29
LS: Trigger full rebuild if build directory has gone missing

### LINKED ISSUE #3954: LSP fails to read file
**Steps to reproduce**:
- Create a new gleam project, and open it with a code editor (I have tested in vscode and zed)
- Open the `app_name.gleam` file. The LSP works normally
- Run `gleam clean` in that project
- Save the file in the editor. 

The LSP then reports an error:
```txt

    project/build/packages/gleam_stdlib/gleam.toml

```

This seems to be because it doesn't properly fetch the source files for dependency packages when saving, even if they don't exist. Running `gleam check|build|run`, or restarting the language server fixes this.

sbergen:
Resolves #3954

I found writing tests for this challenging. If anyone has any pointers, I'd be glad to take another go at it 😅 

You can still mess things up by deleting individual files from within the build directory. However, that seemed very difficult to solve, and also affects the compiler (in addition to LS). Since the full directory can be deleted with `gleam clean`, this seems like it's worth covering as a special case.

I also stumbled upon an invalid modification time optimization while doing it.

### DIFF
--- a/compiler-core/src/language_server/tests/router.rs
+++ b/compiler-core/src/language_server/tests/router.rs
@@ -1,0 +1,121 @@
+use std::time::SystemTime;
+
+use crate::{
+    Error,
+    io::FileSystemWriter,
+    language_server::{files::FileSystemProxy, tests::Action},
+    paths::ProjectPaths,
+};
+
+use super::LanguageServerTestIO;
+
+type Router = crate::language_server::router::Router<LanguageServerTestIO, LanguageServerTestIO>;
+
+#[test]
+fn recompile_after_no_changes_does_not_redownload_dependencies() {
+    let paths = ProjectPaths::new("/app".into());
+    let (io, mut router) = set_up_minimal_router(&paths);
+
+    assert_eq!(
+        compile(&mut router, &paths),
+        Ok(()),
+        "Pre-condition: Initial compile should succeed"
+    );
+
+    {
+        let mut actions = io.actions.lock().unwrap();
+        assert!(
+            actions.contains(&Action::DownloadDependencies),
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/tests/router.rs
```diff
@@ -0,0 +1,121 @@
...
+        Ok(()),
+        "Recompile should succeed"
+    );
+
+    {
+        let actions = io.actions.lock().unwrap();
+        assert!(
+            !actions.contains(&Action::DownloadDependencies),
+            "Recompile should not re-download dependencies"
+        );
```
sbergen, line 45: Since I couldn't find any reliable way to check if a completely fresh project is set up, I'm now riding on this assumption to test things.

I'm not sure if it's the best thing to check, so I'm wondering if adding a function to `ProgressReporter` for a recreated project would make more sense? Or adding something (an id?) to `Project` to be able to check if they are newly created?

sbergen: FWIW, I tried comparing raw `Project` pointers, but at least on my machine, the new project ended up being created on the exact same address, so that's not reliable.

lpil:
I agree with your thoughts RE deletions in the build directory. We can consider this mis-use and no support it at all.

compiler-core/src/language_server/tests/router.rs
lpil: I think this is good! Can always improve it later

lpil (approved):
Super nice work, thank you! I really appreciate how you've been making big improvements in this area of the codebase.

PR #4405, opened 2025-03-29
Simplify LS compiler result type

sbergen:
While I was working on #4404, I noticed that the `Outcome` returned by `LspProjectCompiler::compile` could be replaced with `Result`, making the code quite a bit simpler. However, since the actual fix didn't end up touching that part of the code, I separated this into another PR.


sbergen:
At a closer look, I didn't read the `map` implementation properly, and this was actually not correct...

PR #4411, opened 2025-03-31
Compile bit array patterns into a decision tree

giacomocavalieri:
With this PR bit array patterns are compiled into a decision tree along with all the other patterns. I've also slightly refactored the structure of the decision tree to have a single `Switch` variant (this simplifies missing pattern generation and will greatly simplify code generation!)

As a result the compiler is now a lot smarter in finding unreachable branches containing bit array patterns! For example it can now tell that the second branch here is unreachable:
```gleam
case todo {
  <<len, payload:size(len), _:size(len)>> -> 1
  <<len, _:size(len)-unit(2)>> -> 2
}
```
Beware that it can't find all the unreachable branches! Bit array patterns are quite dynamic in nature and as soon as a variable size is in the mix, it's basically impossible to reliably tell what all the unreachable branches are.
There's a further optimization we could implement to make the compiler even smarter, they call it "read action interference", it doesn't change the final result but can help make the decision tree smaller and find more unreachable branches. It's very complex so I've not implemented it yet here as it's not strictly needed to compile bit array patterns

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -1,7 +1,11 @@
 //! An implementation of the algorithm described in:
-//! - <https://julesjacobs.com/notes/patternmatching/patternmatching.pdf>
-//! - <https://user.it.uu.se/~kostis/Papers/JFP_06.pdf> for binary pattern
-//!   matching
+//!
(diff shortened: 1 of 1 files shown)


lpil (approved):
Wow Jak, this is fantastic! I'm impressed with how approachable you have made this complex functionality.

PR #4426, opened 2025-04-02
Trigger "rewrite from `use`" code action only when hovering over the `use` line

giacomocavalieri:
Before the action would be suggested if the cursor is anywhere in the use block. However, it would end up being annoying most of the times suggesting this code action when triggered on seemingly unrelated pieces of code. With this change the action will only trigger if the cursor is placed in the first line of a use expression: the one containing the `use`


lpil (approved):
Much nicer, thank you

PR #4429, opened 2025-04-02
Implement Gleam `dev`

### LINKED ISSUE #4386: Implement Gleam `dev`
To clarify its meaning, the `test` directory will be deprecated and renamed to `dev`.
See https://github.com/gleam-lang/gleam/blob/main/docs/ideas/gleam-dev.md for details.

GearsDatapacks:
This PR closes #4386 by renaming the `test` directory to `dev`, and adding the `gleam dev` command.
Should we set up a test for the deprecation warning of the `test` directory? I'm not sure the best way to do that.

There are also quite a few places in the codebase which reference the `test` directory. I've tried my best to update them all in a way that makes sense, but have a look through and tell me whether those are sensible changes


GearsDatapacks:
Closing for now as we've had a discussion about how this feature will be implemented and it doesn't particularly match my changes in this PR, so it's probably just easier for me to start from scratch on this

PR #4432, opened 2025-04-02
Wrap binops in curly braces in the "Rewrite to pipe code action"

### LINKED ISSUE #4428: BUG: LSP convert-to-pipe action produces invalid code when it pulls out an expression
## Generating Invalid code

Given the following code: 
```gleam
import gleam/bool

pub fn bug() {
  bool.guard(1 != 0, Nil, fn() { Nil })
}
```

when running the "convert to pipe" lsp action, the following is generated:
```gleam
pub fn bug() {
  1 != 0 |> bool.guard(Nil, fn() { Nil })
}
```

which results in a compiler error, the `1!=0` should be wrapped in a block for it to be valid as follows: 

```gleam
pub fn bug() {
  {1 != 0} |> bool.guard(Nil, fn() { Nil })
}
```

## Creating incorrect Behaviour

This also occurs for non-boolean expressions, for example string concatenation where the output compiles but results in incorrect behaviour
```gleam
string.repeat("a" <> "b", 2) // outputs "abab"
```
is incorrectly transformed to
```gleam
"a" <> "b" |> string.repeat(2) // outputs "abb"
```

## Edge case that succeeds because of operator precedence rules

This bug does not manifest for arithmetic expressions because arithmetic operators have a higher precedence than pipe, where
```gleam
int.add(2*2,1)
```
and 
```gleam
2 * 2 |> int.add(1)
```
both correctly result in an output of 5

giacomocavalieri:
This PR closes #4428 

As I was writing the tests I realised that they were failing since binary operations are not fault tolerant! So this PR also updates the typer to be fault tolerant when typing binary operations.
There a test that started failing as a result of this, I have highlighted it with a comment in this PR as it might need more attention and to be special cased if we want to keep that old behaviour, I'd love if you had a look at that


compiler-core/src/type_/tests/warnings.rs
```diff
@@ -781,19 +781,6 @@ pub const x = some_module.x
...
-pub fn main() {
-  let x = 1
-  1 + ""
-  x
-}"#;
-    let warnings = VectorWarningEmitterIO::default();
-    _ = compile_module("test_module", src, Some(Rc::new(warnings.clone())), vec![]).unwrap_err();
-    assert!(warnings.take().is_empty());
-}
-
```
giacomocavalieri, line 796: Since inference for binary operations is now fault tolerant this does also raise a warning complaining that `1 + ""` is unused. Can we accept this, or should it be special cased somehow to still not raise a warning?

TanklesXL:
@giacomocavalieri does this change fix this kind of thing as well?: 

```gleam
fn bug() {
  bool.guard(1 == 2 || 2 == 3, Nil, fn() { Nil })
}

// transformed  with convert-to-pipe as:

fn bug() {
  1 == 2
  || 2
  == 3
  |> bool.guard(Nil, fn() { Nil })
}
```

giacomocavalieri:
Yes it does! I've added a test just to make sure 😁

PR #4433, opened 2025-04-02
Use correct location when inlining field access variable

### LINKED ISSUE #4430: "Inline Variable" action does not work with leading field lookups
Bug with the "inline variable" code refactor action. When in-lining a variable, if the definition body of the variable **starts by referencing a field** then the action breaks the code, by omitting the record.

## main example
```
type Foo {
  Foo(items: List(Int))
}

pub fn main() {
  let myfoo = Foo([])
  let xyz = myfoo.items       // reference any `record.field`, or bigger expressions starting like this
  echo xyz
}
```
In-lining the variable xyz results in this:
```
pub fn main() {
  let myfoo = Foo([])
  echo items                  // expected "echo myfoo.items"
}
```
## what works okay
Wrapping in braces works okay:
```
  let myfoo = Foo([])
  let xyz = {myfoo.items} 
  echo xyz
```
Leading with not a field reference works okay:
```
  let myfoo = Foo([])
  let item = list.first(myfoo.items) // fine leading with a module reference
  echo item
```

## Versions

Gleam 1.9.1

VsCode
Version: 1.98.2 (user setup)
Commit: ddc367ed5c8936efe395cffeec279b04ffd7db78
Date: 2025-03-12T13:32:45.399Z
Electron: 34.2.0
ElectronBuildId: 11161602
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Windows_NT x64 10.0.26100

giacomocavalieri:
This PR closes #4430
The field access location now spans the entire thing and is consistent with the way we've updated the module select location a while ago.

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -90,6 +90,7 @@ pub enum TypedExpr {
 
     RecordAccess {
         location: SrcSpan,
+        field_start: u32,
         type_: Arc<Type>,
         label: EcoString,
         index: u64,
--- a/compiler-core/src/ast/visit.rs
+++ b/compiler-core/src/ast/visit.rs
@@ -213,12 +213,13 @@ pub trait Visit<'ast> {
     fn visit_typed_expr_record_access(
         &mut self,
         location: &'ast SrcSpan,
+        field_start: &'ast u32,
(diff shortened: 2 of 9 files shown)


lpil (approved):
Thank you

PR #4436, opened 2025-04-02
Fix running of `running_modules` tests on CI

GearsDatapacks:
I noticed while working on [this PR](https://github.com/gleam-lang/gleam/pull/4429) that the `test/running_modules` tests weren't actually running properly. (The `make` rule is `test-all`, not `test`)
I also noticed that this test was using old dependencies which no longer compile on the latest Gleam version, so I've updated them.


lpil (approved):
Ouch! Thank you

PR #4438, opened 2025-04-02
Improve argument names for generated functions

### LINKED ISSUE #4408: Improve generate function code action argument naming
Two things we can do to improve the names used for arguments to the generated function:

- If the argument was given with a label then use the label name
- If the argument is a variable use the variable name

GearsDatapacks:
Closes #4408


lpil (approved):
Thank you!! So fast!!

PR #4440, opened 2025-04-03
Fix 1.86 clippy warnings

GearsDatapacks:
There are several warnings from `clippy` as of Rust v1.86; this PR fixes them.


lpil (approved):
Thank you

PR #4441, opened 2025-04-03
Use local Hex key to revoke local Hex API key

### LINKED ISSUE #4319: `gleam hex authenticate` must be run twice to switch accounts.
If you hold multiple hex accounts, you might want to replace the local token gleam uses to interact with the hex api from time to time. This consistently fails when changing account details and must always be run a second time to succeed.

Steps to reproduce:

- run `gleam hex authenticate` for the first time to generate a local token for hex account A
- run `gleam hex authenticate` a second time, confirm that you would like to replace the local token, and provide different credentials for hex account B
- observe the following error
```
% gleam hex authenticate

You already have a local Hex API token. Would you like to replace it
with a new one? [y/n]: y
https://hex.pm username: B
https://hex.pm password (will not be printed as you type): 

Please enter a new unique password. This will be used to locally
encrypt your Hex API key.

Local password (will not be printed as you type):
Encrypted Hex API key written to .../gleam/hex/hexpm/credentials
Deleting previous key `...` from Hex
error: Hex API failure

    an unexpected response was sent by Hex: 404 Not Found: {"message":"Page not found","status":404}
```
- Run `gleam hex authenticate` a third time exactly the same as above
```
% gleam hex authenticate

You already have a local Hex API token. Would you like to replace it
with a new one? [y/n]: y
https://hex.pm username: B
https://hex.pm password (will not be printed as you type):

Please enter a new unique password. This will be used to locally
encrypt your Hex API key.

Local password (will not be printed as you type):
Encrypted Hex API key written to .../gleam/hex/hexpm/credentials
Deleting previous key `...` from Hex
```
- Observe that the token has successfully been replaced.

You can continue to confirm the issue by attempting to authenticate with hex account A again and observe you must go through the command twice more.

scristobal:
resolves #4319 

After a bit of though, I realized we can use the existing key to revoke itself. Provided the Hex API allows such a thing.

As it turns out, using the existing key to revoke itself works just fine, but requires asking the user for the local password. Hence the command flow needs to change a bit. Here is the result:

<a href="https://asciinema.org/a/f9P3EuBbMvwfZfK32yggKWNXn" target="_blank"></a>

In any case, if the key couldn't be revoked, the user is hinted to finish the process manually, instead of failing, like in #4319.

Please do let me know if you think this approach is a good fit.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -111,6 +111,9 @@
 - Deprecate `HEXPM_USER` and `HEXPM_PASS` in favour of `HEXPM_API_KEY`.
   ([Samuel Cristobal](https://github.com/scristobal))
 
+- Improved Hex API key removal CLI process.
+  ([Samuel Cristobal](https://github.com/scristobal))
+
 ### Language server
 
 - The language server now allows renaming of functions, constants,
--- a/compiler-cli/src/hex.rs
+++ b/compiler-cli/src/hex.rs
@@ -1,6 +1,7 @@
 mod auth;
 
 use crate::{cli, http::HttpClient};
+use auth::EncryptedApiKey;
 use gleam_core::{
     Error, Result,
     hex::{self, RetirementReason},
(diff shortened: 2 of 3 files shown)


lpil (commented):
Thank you!

Rather than decrypting it first we should only do it if the new key cannot be used to delete the old one. The most common reason to overwrite the existing key is if the programmer has forgotten their password, as it's very rare for someone to have multiple Hex accounts. There's not really any reason to make more than one because Hex supports organisations and multiple maintainers per package.

scristobal:
follow up in https://github.com/gleam-lang/gleam/pull/4446

PR #4443, opened 2025-04-03
Only show completions for prelude items when not in a module qualifier

GearsDatapacks:
I noticed that when the LSP provides autocomplete for types and values after a module select, e.g.
```gleam
pub fn main() {
  let x = option.|
}
```
It was including completions for prelude values, like `Ok`, which would not be valid after `option.`
This PR fixes that behaviour


lpil (approved):
Oh much better! Thank you!

PR #4445, opened 2025-04-04
Fix bug with checking of top-level modules

GearsDatapacks:
It has been pointed out that there is a bug in the code for checking multiple top-level modules when publishing.
This PR fixes the bug, so that only top-level modules are checked for, and not top-level directories.


lpil (approved):
Beautiful, thank you

PR #4448, opened 2025-04-05
Fix slightly wrong error message for missing main function in test module

### LINKED ISSUE #4420: Slightly wrong error message for missing `main` function in test module
When running `gleam test`, if the `project_test` module doesn't have a `main` function, this error is printed:
```txt
error: Module does not have a main function

`project_test` does not have a main function so the module can not be run.
Hint: Add a public `main` function to to `src/project_test.gleam`.
```
Which isn't quite correct, as the file is `test/project_test.gleam` not `src/project_test.gleam`

scristobal:
resolves #4420

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -235,3 +235,6 @@
 - Fixed a bug where updating the remote URL of a Git dependency would fail to
   update the remote in the local dependency, causing a caching issue.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fix slightly wrong error message for missing main function in test module.
+  ([Samuel Cristobal](https://github.com/scristobal))
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -35,7 +35,7 @@ use camino::Utf8PathBuf;
 use ecow::EcoString;
 use itertools::Itertools;
 use serde::{Deserialize, Serialize};
-use std::fmt::Debug;
+use std::fmt::{Debug, Display};
 use std::sync::Arc;
 use std::time::SystemTime;
 use std::{collections::HashMap, ffi::OsString, fs::DirEntry, iter::Peekable, process};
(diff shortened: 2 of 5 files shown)


compiler-core/src/build.rs
```diff
@@ -572,6 +572,18 @@ impl Origin {
     }
 }
 
+impl Display for Origin {
```
lpil, line 575: Don't ever implement traits please! Prefer regular methods are they are much clearer.

compiler-core/src/type_.rs
```diff
@@ -1569,6 +1570,7 @@ fn assert_suitable_main_function(
 ) -> Result<(), crate::Error> {
     let not_found = || crate::Error::ModuleDoesNotHaveMainFunction {
         module: module_name.clone(),
+        origin: Origin::Src,
```
lpil, line 1573: This is hard-coded to src, but it could be one of the others origins. It'll need to be passed in.

CHANGELOG.md
```diff
@@ -30,6 +30,9 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fix slightly wrong error message for missing main function in dev module.
+  ([Samuel Cristobal](https://github.com/scristobal))
```
lpil, line 34: In the bug fix section please 🙏

lpil (commented):
Thank you! Few small notes.

compiler-core/src/build.rs
scristobal: indeed! added a `.folder_name` method to `Origin`

thanks for the feedback

compiler-core/src/type_.rs
scristobal: uuupppps 🙃 added as parameter to the assertion, also adapted existing tests using it

CHANGELOG.md
scristobal: done!

lpil (approved):
Lovely! Thank you!!

Issue #4280, opened 2025-02-25
Feat: LS show const values on hover


metruzanca:
Say you have this setup:
```gleam
// httpcodes.gleam

pub const redirect = 303

// server_test.gleam
// ...
 response.status
  |> should.equal(httpcodes.redirect)
//                          ^ hovering here currently only shows the type, in this case "Int"
```

Coming from writing typescript all day, I was expecting to see the value for constants.
```ts
const a = 100
console.log(a);
//          ^ shopws "const a: 100"
```

It would be nice if, like typescript, it showed a message like `const redirect: Int = 303`. (using valid syntax in this popup to avoid confusion)

GearsDatapacks:
I believe in typescript, the type is `100`, as that is a valid TS type. But yeah we could probably do this.
Do we also want to fully reduce the constant? E.g:
```gleam
const a = 10
const b = a
//    ^ hover here
```
What would we show in this case?

metruzanca:
> What would we show in this case?

That's a good question. I think just showing the value is already a nice DX that avoids a jump to definition but if it would be cool to, as you say, reduce down to the actual value.

e.g.
```gleam
const base_url = "https://zanca.dev"
const page_url = base_url <> "/gear"
//    ^ const page_url = "https://zanca.dev/gear"
```

Records too?
```gleam
const name = "Metru"
const default_context = Context(name)
//    ^ const default_context = Context( username: "Metru" )
```

If users want to see what its composed from, they can jump to definition. I think its more useful in both these scenarios to be able to quickly see the result.

lpil:
Showing computed values when the definition is an expression would require compile time compilation, which we don't have. Let's defer this until a second iteration of this feature.

Constants can be arbitrarily large. Do we need to truncate or trim them in any way?

giacomocavalieri:
Hello! So any ideas on what the design of this would be? I'm not really sure about truncating, are there any other language servers that do this we could take inspiration from?

In Rust it shows the entire constant declaration if the constant is not too long, but hides the value of the constant after the `=` if it is too long

lpil:
Is it too long as some number of lines?

giacomocavalieri:
Ah I got it wrong, it never shows the actual value unless it's a numeric value. For anything like structs, tuples, strings, and slices it always adds the value and only shows the head like this:

```
pub const wibble: [u8; 10] =
```

Quite odd, I don't think there's any value in showing that since hover already shows the type and the name is right under the cursor

[...]

PR #4449, opened 2025-04-05
Display const values on hover action

unknown:
Closes https://github.com/gleam-lang/gleam/issues/4280. Displays the expressions of consts when they are hovered.

```gleam
// consts.gleam
pub const pi = 3.1415
```

```gleam
// main.gleam
import consts

pub fn main() {
  consts.pi
}
```

For example, hovering `pi` in `consts.gleam` and `consts.pi` in `main.gleam` renders `const pi: Float = 3.1415`.

I think we can make further improvements in the future to other hover actions to be more consistent with this new behavior such as making hovering `a` in `let a = 42` render `let a: Int`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -458,6 +458,9 @@
   first line of the `use` expression to rewrite.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Hovering const values now displays their values.
+  ([David Mo](https://github.com/unknown))
+
 ### Formatter
 
 ### Container images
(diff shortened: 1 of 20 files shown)


lpil (commented):
Thank you but as the issue says we have not yet determined what the desired behaviour is. That design work needs to be done prior to implementation.

unknown:
> Thank you but as the issue says we have not yet determined what the desired behaviour is. That design work needs to be done prior to implementation.

Wow, not sure how I missed that! My apologies.

lpil:
No worries!

unknown:
To be clear, is this feature blocked by the compile-time computation part or by working out the details of the behavior e.g. truncation? If the latter, I can do a survey of other language servers and report back in the parent issue.

lpil:
Yup, and other things such as how to handle opaque types. Needs a good bit of thinking.

PR #4452, opened 2025-04-06
Fix changelog

GearsDatapacks:
I noticed that one entry was in the wrong section, so I have moved it.
I've also reworded a couple of entries/slightly changed the formatting to make them more consistent with the rest of the changelog.


lpil (approved):
Lovely, thank you

PR #4453, opened 2025-04-06
Improve error message for wrong operator on Ints/Floats

giacomocavalieri:
The error message now explains what operator can be used instead!


lpil (commented):
It already says that above, why would we want to have that sentence twice?

giacomocavalieri:
That's a great question I should have asked myself before doing this 😆

PR #4455, opened 2025-04-07
Fix bug with unqualify code action

### LINKED ISSUE #4454: Unqualify code action readds import
When you have imported a module unqualified but you have it written with the module, the code action will readd the existing import.

Example:
```gleam
import gleam/dict.{type Dict}

pub fn a(d: dict.Dict(Int, Int)) {
  Nil
}
```
becomes
```gleam
import gleam/dict.{type Dict, type Dict}

pub fn a(d: Dict(Int, Int)) {
  Nil
}
```

Gleam: v1.9.1

GearsDatapacks:
Fixes #4454
There was just a small typo here, so it didn't properly check for already existing type imports only value imports.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Fixed a bug where the code action to unqualify types and values would add an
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you

PR #4456, opened 2025-04-07
Implement autocomplete, hover and goto-definition for constants

### LINKED ISSUE #3631: Language server completion within consts
Not sure if it's a bug, but this could be a feature request as well.

My experience with [Record Types](https://tour.gleam.run/data-types/records/) and [Generic Custom Types](https://tour.gleam.run/data-types/generic-custom-types/) have been pretty clunky.

Here's a specific instance where when I press "Ctrl + Space", intellisense doesn't work (for Record Types), when I expect it to.

It should ideally show me what the type or the name of the parameter my cursor is currently on.

https://github.com/user-attachments/assets/dfc5b76c-28c7-48c7-b8c6-f7a2231268ff

---

Idk if it's the LSP getting confused between **Generics** and **Instantiating a Record Type** because it uses almost the same same syntax e.g. `MyType(generic)` (with generics) and e.g. `MyType("propertyvalue")` (Instantiating record types).

GearsDatapacks:
Fixes #3631 
I thought that while I was at it I may as well also add hover and goto-definition functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -149,6 +149,11 @@
     }
   }
   ```
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
+- The language server now provides hover, autocomplete and goto definition
+  for constant definitions.
+  ([Surya Rose](https://github.com/GearsDatapacks))
 
 ### Formatter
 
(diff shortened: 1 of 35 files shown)


lpil (approved):
Thank you!!! 💜

PR #4457, opened 2025-04-07
Add warning for truncated int segments

### LINKED ISSUE #791: Check bit string sizes statically
Currently you can write
```
pub const x = <<100:size(1)>>
```
which to me seems nonsensical (it will just get truncated).

It gets compiled into Erlang's `<<100:1>>` which then gets interpreted as `<<0:1>>`.

**Perhaps it would make sense to raise a compiler error if the value bit length is larger than the advertised length after the colon?**

I'm not versed well enough in the Erlang ecosystem to say if there is some legitimate use for this truncation though. Might very well be, and then it would probably make sense to keep allowing it.

giacomocavalieri:
This PR closes #791 by adding a warning when the compiler can tell that a segment with a positive literal int is going to be truncated, suggesting what value to use instead.

And there's also a code action to automatically fix it!

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -2410,7 +2410,7 @@ where
                 (-(BigInt::one() << (segment_bits - 1)))
                     ..((BigInt::one() << (segment_bits - 1)) - 1)
             }
-            Sign::Plus => BigInt::ZERO..((BigInt::one() << segment_bits) - 1),
+            Sign::Plus => BigInt::ZERO..(BigInt::one() << segment_bits),
         };
 
         if !safe_range.contains(&literal_value) {
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -2398,17 +2398,25 @@ where
 {
     pub(crate) fn check_for_truncated_value(&self) -> Option<BitArraySegmentTruncation> {
         // Both the size and the value must be two compile-time known constants.
-        let segment_bits = self.bits_size()?;
+        let segment_bits = self.bits_size()?.to_i64()?;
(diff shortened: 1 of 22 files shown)


compiler-core/src/ast.rs
```diff
@@ -2330,6 +2333,73 @@ impl TypedExprBitArraySegment {
...
 
+impl<TypedValue> BitArraySegment<TypedValue, Arc<Type>>
+where
+    TypedValue: HasType + HasLocation + Clone + bit_array::GetLiteralValue,
+{
+    pub(crate) fn check_for_truncated_value(&self) -> Option<BitArraySegmentTruncation> {
+        // Both the size and the value must be two compile-time known constants.
+        let segment_bits = self.bits_size()?;
+        let literal_value = self.value.as_int_literal()?;
+        if literal_value.sign() != Sign::Plus || segment_bits.sign() != Sign::Plus {
```
richard-viney, line 2344: Could also warn when the value is negative and out of range.

If the value is negative then the valid range is `(-(1 << (segment_bits - 1)))..((1 << (segment_bits - 1)) - 1)`

So these would also trigger a warning:

```gleam
<<-129>>
<<-530:10>>
<<-80_000:16>>
```

Note that in some cases the `truncated_into` value will be positive, even though the literal value is negative.

giacomocavalieri: Great I'm adding this as well then!

giacomocavalieri:
This is ready for review now!

GearsDatapacks:
Looks like you forgot to accept a snapshot

giacomocavalieri:
Oops, thank you!

lpil (approved):
Thank you!!

Issue #4451, opened 2025-04-06
LS: hover info for custom types and their constructors


unknown:
The language server currently doesn't return hover info for custom types and their constructors. It would be useful for the language server to support this.

```gleam
//// wibble.gleam

type Wibble {
  // ^ wibble.Wibble

  Wibble(Int)
  // ^ fn(Int) -> Wibble

  Wobble(String)
  // ^ fn(String) -> Wibble
}
```
The proposed behavior above would be consistent with the hover info of type annotations and constructor expressions.

GearsDatapacks:
Ah, I think that may be a bug. I'm sure I've used that feature before. I'll investigate

unknown:
> Ah, I think that may be a bug. I'm sure I've used that feature before. I'll investigate

I've tested every tagged release dating back to v1.0.0, and none of the versions provided hover info on my machine. Feel free to investigate further, but I'd like to work on this in the meantime!

GearsDatapacks:
Go ahead! I haven't had a chance to look at it yet

unknown:
Ran into a small issue with implementing the proposed behavior because `RecordConstructor` doesn't store any data related to which custom type it's a member of. Not sure if there are other approaches to get the custom type, but alternatively, we can display something similar to `rust-analyzer`'s behavior when hovering enums like the following:

```gleam
//// wibble.gleam

type Wibble {
  // ^ Wibble

  Wibble(Int)
  // ^ Wibble(Int)

  Wobble(String)
  // ^ Wobble(String)
}
```

unknown:
While working on this, I've also realized that type aliases do not have any hover info. Is this something that we want to support? If so, should I open a separate issue or merge that into this one?

lpil:
If you have a good idea for what they would show, sounds good.

PR #4458, opened 2025-04-07
LS: hover info for custom types

unknown:
Adds hover info for custom types and closes https://github.com/gleam-lang/gleam/issues/4451. I marked this PR as a draft, because the behavior I implemented differs slightly from the original proposed behavior I proposed. See https://github.com/gleam-lang/gleam/issues/4451#issuecomment-2783984769 for more context.

This PR changes the language server display type info like the following:
```gleam
//// wibble.gleam

type Wibble {
  // ^ Wibble

  Wibble(Int)
  // ^ Wibble(Int)

  Wobble(String)
  // ^ Wobble(String)
}
```

### DIFF
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -4,6 +4,7 @@ use crate::{
     ast::{
         self, CustomType, Definition, DefinitionLocation, ModuleConstant, PatternUnusedArguments,
         SrcSpan, TypedArg, TypedExpr, TypedFunction, TypedModule, TypedPattern,
+        TypedRecordConstructor,
     },
     build::{Located, Module, UnqualifiedImport, type_constructor_from_modules},
     config::PackageConfig,
@@ -822,6 +823,9 @@ where
                 Located::ModuleStatement(Definition::Function(fun)) => {
                     Some(hover_for_function_head(fun, lines, module))
                 }
+                Located::ModuleStatement(Definition::CustomType(type_)) => {
+                    Some(hover_for_custom_type(type_, lines))
+                }
                 Located::ModuleStatement(Definition::ModuleConstant(constant)) => {
                     Some(hover_for_module_constant(constant, lines, module))
                 }
@@ -837,7 +841,9 @@ where
                     ))
                 }
                 Located::ModuleStatement(_) => None,
-                Located::VariantConstructorDefinition(_) => None,
+                Located::VariantConstructorDefinition(constructor) => {
+                    Some(hover_for_constructor(constructor, lines, module))
+                }
                 Located::UnqualifiedImport(UnqualifiedImport {
                     name,
                     module: module_name,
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/engine.rs
```diff
@@ -1242,6 +1248,53 @@ fn hover_for_module_constant(
...
+    }
+
+    let arguments = arguments.join(", ");
+    let empty_str = EcoString::from("");
+    let documentation = constructor
+        .documentation
+        .as_ref()
+        .map(|(_, doc)| doc)
+        .unwrap_or(&empty_str);
+    let contents = format!("```gleam\n{name}({arguments})\n```\n{documentation}");
```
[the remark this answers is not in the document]
unknown: Yeah I made a note of that here: https://github.com/gleam-lang/gleam/issues/4451#issuecomment-2783984769

It's been a while since I worked on this, but if I remember correctly, the reason it differs is because `TypedRecordConstructor` doesn't store any information about the custom type its a member of (aka the return type of the constructor). Should I add this to `TypedRecordConstructor`?

GearsDatapacks (approved):
This looks good to me now! Just need to fix the merge conflicts and get Louis' stamp of approval

compiler-core/src/ast.rs
```diff
@@ -255,6 +255,7 @@ pub struct RecordConstructor<T> {
     pub location: SrcSpan,
     pub name_location: SrcSpan,
     pub name: EcoString,
+    pub type_name: EcoString,
```
lpil, line 258: Remove this please 🙏 All constructors share the same type, and that type is stored at the level above, so this is duplicating the information already there. Instead pass is down into the function as needed.

compiler-core/src/language_server/engine.rs
```diff
@@ -1326,6 +1331,49 @@ fn hover_for_constant(
...
 }
 
+fn hover_for_custom_type(type_: &CustomType<Arc<Type>>, line_numbers: LineNumbers) -> Hover {
+    let name = &type_.name;
+    let empty_str = EcoString::from("");
+    let documentation = type_
+        .documentation
+        .as_ref()
+        .map(|(_, doc)| doc)
+        .unwrap_or(&empty_str);
```
lpil, line 1341: unwrap_or_default() perhaps?

compiler-core/src/language_server/engine.rs
```diff
@@ -1326,6 +1331,49 @@ fn hover_for_constant(
...
+    let mut arguments = Vec::with_capacity(constructor.arguments.len());
+    let mut printer = Printer::new(&module.ast.names);
+
+    for argument in constructor.arguments.iter() {
+        let type_ = printer.print_type(&argument.type_);
+        arguments.push(type_);
+    }
+
+    let arguments = arguments.join(", ");
+    let empty_str = EcoString::from("");
```
lpil, line 1364: Here too.

lpil (commented):
Thank you! I've left some tiny notes inline.

I'm a bit confused by the motivation for this addition with the hover information as it is today. It shows the name of the type, the arguments the variant takes, and the documentation for the variant, but that is the exact same information that is on the source code being hovered.

What situation would this hover information help with?

compiler-core/src/language_server/engine.rs
unknown: The reason for having `empty_str` is that `unwrap_or_default` returns an `EcoString`, but we need a `&EcoString`.    Doing so is in line with some of the other hover functions (e.g. `hover_for_function_head`).

lpil:
@matiascr has pointed out in https://github.com/gleam-lang/gleam/issues/4955 that one may want to preview the rendered documentation you have just written! That is a good motivation for this addition.

matiascr:
I've already created an issue (https://github.com/gleam-lang/gleam/issues/4957) related to this.
It's also related to the kind of hover info @unknown is implementing but specifically for assertions, where it could also be improved. Were you planning on taking care of the case?

compiler-core/src/language_server/engine.rs
```diff
@@ -1326,6 +1331,49 @@ fn hover_for_constant(
...
+    let documentation = constructor
+        .documentation
+        .as_ref()
+        .map(|(_, doc)| doc)
+        .unwrap_or(&empty_str);
+    let contents = format!("```gleam\nfn({arguments}) -> {type_name}\n```\n{documentation}");
```
matiascr, line 1370: Should we take the opportunity to update this type info? For example turn it into 
```
Variant(label: FieldType)
Documentation of variant
```
instead of
```
fn(TypeOfField) -> ParentOfVariant
Documentation of variant
```

lpil:
Hello! Gleam v1.13.0 is out, so compiler dev is resuming. Would you mind incorporating the PR feedback above? Would love to get this merged! 🙏  Thank you

compiler-core/src/language_server/engine.rs
lpil: Changes in the design for hovers would be another issue 🙏  So discussing that doesn't end up blocking this one

unknown:
> Hello! Gleam v1.13.0 is out, so compiler dev is resuming. Would you mind incorporating the PR feedback above? Would love to get this merged! 🙏 Thank you

Will get back to this when I get the chance, hopefully this weekend!

lpil:
Fab! Looking forward to it! 💜

PR #4466, opened 2025-04-12
Ignore .DS_Store in new Gleam Projects

daniellionel01:
Hi!

I noticed in a lot of my gleam projects, that .DS_Store files pop up sometimes and I have to manually add it to the .gitignore every time. As far as I know it is best practice to not include that file, so I thought I add it to the template for a `gleam new` project.

Happy to discuss, if this makes sense 🙂


daniellionel01:
ok, sure!

PR #4467, opened 2025-04-12
Autocomplete already-imported modules

### LINKED ISSUE #4465: LSP: autocomplete already-imported modules
See this use case for example:

Image

I'm about to type "lustre.element". However, the symbol I want is far down in the suggestion list. Because I have just typed "lust", I want the `lustre` module name appears at the top, so that I can tab to finish "lustre", then type "." to narrow the suggestions.

hongquan:
Resolve #4465 

image

image

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -626,6 +626,45 @@ where
                 continue;
             };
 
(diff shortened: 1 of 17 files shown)


lpil:
Please do not make unplanned PRs! All changes must be agreed to first.

PR #4473, opened 2025-04-16
Fix JavaScript codegen bugs

### LINKED ISSUE #4278: Incorrect missing patterns given for `case` expression
To reproduce, you need this code:
```gleam
pub fn wibble(b: Bool, i: Int) {
  case b, i {
    False, 1 -> todo
    True, 2 -> todo
  }
}
```
The gleam compiler/LS gives this error:
```
The missing patterns are:

    False, _
    True, _
    _, _
```
Which is not correct. It could either give `_, _` or `False, _`/`True, _`, but both are not needed.

### LINKED ISSUE #4472: Wrong code generated when lifting blocks in pipelines
I've found another small bug with block expression lifting on the JavaScript target.
If you have some code like this:
```gleam
1
|> echo
|> {
  echo 2
  fn(x) { echo x }
}
```

This should print `1` then `2` then `1`, but currently it prints `2` then `1` then `1`.

GearsDatapacks:
Fixes #4278 and fixes #4472

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,3 +65,11 @@
 ### Formatter
 
 ### Bug fixes
+
+- Fixed a bug where `case` expressions in custom panic messages would compile
+  to invalid syntax on the JavaScript target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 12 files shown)


lpil (approved):
Thank you!!

PR #4479, opened 2025-04-18
Warn for pure unused functions

### LINKED ISSUE #4477: Add a warning for pure functions whose return values are not used
Coming from other programming languages, immutability can be a source of confusion for new Gleam programmers.
A suggestion to help with this is to track the purity of functions and emit a warning when one is used without capturing the result somehow, the same way we warn for other unused pure values.

A function is pure if it:
- Doesn't call any external functions
- Cannot panic in any way

GearsDatapacks:
Closes #4477


compiler-core/src/type_/error.rs
```diff
@@ -748,6 +748,10 @@ pub enum Warning {
         location: SrcSpan,
     },
 
+    UnusedPureFunctionCall {
```
lpil, line 751: We're interested in pure expressions in general, not just function calls. Other expressions such as access and maths should also warn

compiler-core/src/type_/expression.rs
```diff
@@ -232,13 +248,21 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             uses_javascript_externals: definition.has_javascript_external,
         };
 
+        let purity =
+            if implementations.uses_erlang_externals || implementations.uses_javascript_externals {
```
lpil, line 252: This is interesting. Do we want to take the current target into account when computing purity? It could be pure on one but not the other.

compiler-core/src/warning.rs
```diff
@@ -894,6 +894,26 @@ can already tell which branch is going to match with this value.",
                     }),
                 },
 
+                type_::Warning::UnusedPureFunctionCall { location } => Diagnostic {
+                    title: "Unused pure function call".into(),
+                    text: wrap(
+                        "Gleam is an immutable language, \
+meaning functions cannot mutate state simply by being called. This function is \
+pure, so it must be assigned to a variable to have an effect on the program.",
```
lpil, line 902: This is true for much of the core language, but it's not true for `panic`, FFI, etc. How about something like this:

> This expression computes a value without any side effects, but then the value isn't used at all. You might way to assign it to a variable, or delete the expression entirely if it not needed.

lpil (commented):
Fantastic! Thank you. I've left some notes inline.

Purity hasn't been included in the package interface, which I think is a good idea for now.

It could be even more useful if the compiler knows that functions such as `list.reverse` and `dict.get` are pure. Do we want to have a hard-coded list of pure stdlib functions?

compiler-core/src/type_/error.rs
GearsDatapacks: We already have a warning for that! I decided to make this a separate warning in order to provide more context to the user

compiler-core/src/type_/expression.rs
GearsDatapacks: That's a good point, I can do that

GearsDatapacks:
I think having a list of pure stdlib functions is good, especially for things like `dict.insert` which in a non-functional language would not return anything and could definitely be a source of confusion.

compiler-core/src/type_/error.rs
lpil: We should only split them if we want to handle each of them differently, but in this case I think the message would be the same.

Is there some different information you think we'd want to communicate for each one?

GearsDatapacks:
I've now added the `is_pure_standard_library_function` function which matches against all the current standard library functions except:
- `gleam/io.*` - those are inherently impure
- The deprecated `gleam/dynamic` modules - they will soon be removed and be redundant

compiler-core/src/type_.rs
```diff
@@ -920,7 +920,12 @@ impl ModuleValueConstructor {
             ModuleValueConstructor::Record { .. } | ModuleValueConstructor::Constant { .. } => {
                 false
             }
-            ModuleValueConstructor::Fn { purity, .. } => purity.is_pure(),
+            ModuleValueConstructor::Fn {
+                purity,
+                module,
+                name,
+                ..
+            } => purity.is_pure() || is_pure_standard_library_function(module, name),
```
lpil, line 928: This returns the wrong value in the case that in the stdlib `panic` etc is added, and it means that the stdlib functions are marked as impure but are later overridden.

How about instead functions within the `gleam_stdlib` package within non `gleam/io` modules that use FFI are initially marked as a third option during inference: `Purity::TrustedPure`. This would mean that this override isn't needed every time a module function is used, and it also communicates its not-inferred "pure" nature.

It would also mean that we don't need to add the name of the package to this variant, to check that it is in the stdlib.

GearsDatapacks: Yeah, that works. If we do that, we also probably don't need the hardcoded list of all stdlib functions

lpil: Maybe instead of tracking `pure` entirely we could track whether a function panics. The fact it uses FFI is already tracked, so we can look at that information along with whether it panics to determine if it is pure, the FFI tracking isn't duplicated, and the information that it panics could be useful for other things in future.

compiler-core/src/type_/error.rs
GearsDatapacks: Ah that's true. In the original warning message I referenced funciton-call specific stuff, but the new one that you suggested doesn't. I agree, I'll merge them again

compiler-core/src/type_.rs
GearsDatapacks: There's also the question of whether it uses `echo`, but yeah that could work

CHANGELOG.md
```diff
@@ -73,6 +73,34 @@
...
+  ```
+  warning: Unused pure function call
+    ┌─ /home/gears/projects/playground/gleam/src/playground.gleam:4:3
+    │
+  4 │   add(1, 2)
+    │   ^^^^^^^^^ This pure function call is never used
+
+  Gleam is an immutable language, meaning functions cannot mutate state
+  simply by being called. This function is pure, so it must be assigned to a
+  variable to have an effect on the program.
```
giacomocavalieri, line 99: This isn't technically correct though, that can happen in Gleam! Also "purity" is a bit of lingo I'd like to avoid including in the error message. I'd rather keep the old message that just warned about the unused value and said the expression could be safely removed

GearsDatapacks: Louis has already suggested a better error message, I just haven't pushed my latest commit yet

GearsDatapacks:
Based on our discussion on Discord, I've reverted some of my changes to the purity tracking system to make it simple but effective enough for the current use case.
I've also added a comment to the `Purity` enum explaining the limitations of this system and how it can be improved in future.

compiler-core/src/ast/typed.rs
```diff
@@ -849,6 +855,34 @@ impl TypedExpr {
...
+            | TypedExpr::Tuple { .. }
+            | TypedExpr::TupleIndex { .. }
+            | TypedExpr::Todo { .. }
+            | TypedExpr::Panic { .. }
+            | TypedExpr::Echo { .. }
+            | TypedExpr::BitArray { .. }
+            | TypedExpr::RecordUpdate { .. }
+            | TypedExpr::NegateBool { .. }
+            | TypedExpr::NegateInt { .. }
+            | TypedExpr::Invalid { .. } => Purity::Unknown,
```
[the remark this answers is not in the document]
GearsDatapacks: You're correct; this function is named poorly. The main reason that most of these are unknown is because they can't be called as functions: the expression `10()` is not valid Gleam so we don't need to worry about its purity. This function should probably named something like `called_function_purity` to properly describe its intention.

And yes, some of it is to avoid traversing the AST. It's possible to track the purity of this:
```gleam
case x {
  True -> io.println
  False -> function.identity
}()
```
However it's quite rare that such code will appear in real applications, so I thought it not worth it to bother.

lpil (approved):
Really fantastic documentation there! I love it. Thank you!

PR #4480, opened 2025-04-19
Bump pubgrub.

Kaholaz:
This PR requires https://github.com/gleam-lang/hexpm-rust/pull/43. Create a symlink named hexpm in the root of the project linking to my `hexpm-rust` branch to look at my changes. `Cargo.toml` needs to be cleaned before a merge.

Most changes are one of these three types of changes:
1. Accommodating the `hexpm-rust` change that parses `hexpm::version::Range`.
2. Changing imports to coincide with the 0.3 structure of pubgurb.
3. Changes to the dependency resolution due to changes to the `pubgrub::DependencyProvider` trait.

Out of those, I believe nr. 3 warrants the most explanation (2. is explained in the `hexpm-rust` pull  request -- let me know if anything is unclear). The major changes include `choose_package_version` of `pubgrub::DependencyProvider` being split into the functions `choose_version` and `prioritize`, and the trait requiring the the associated type `Err: Error + 'static`. The changes to the `choose_package_version`, might be the hardest one to get correct, but all tests are passing so I _think_ we are good. The change to the associated `Err` type required a new error type and caused this to propagate a bit. I don't know if this is the best way to do it, but it does the job :D

Again: Let me know if anything is unclear or if you want me to write a more detailed changelog.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -15,7 +15,7 @@ use gleam_core::{
     Error, Result,
     build::{Mode, Target, Telemetry},
     config::PackageConfig,
-    dependency,
+    dependency::{self, PackageFetchError},
     error::{FileIoAction, FileKind, ShellCommandFailureReason, StandardIoAction},
     hex::{self, HEXPM_PUBLIC_KEY},
     io::{HttpClient as _, TarUnpacker, WrappedReader},
@@ -314,7 +314,10 @@ fn remove_extra_requirements(config: &PackageConfig, manifest: &mut Manifest) ->
 pub fn parse_gleam_add_specifier(package: &str) -> Result<(EcoString, Requirement)> {
     let Some((package, version)) = package.split_once('@') else {
         // Default to the latest version available.
-        return Ok((package.into(), Requirement::hex(">= 0.0.0")));
+        return Ok((
+            package.into(),
+            Requirement::hex(">= 0.0.0").expect("'>= 0.0.0' should be a valid pubgrub range"),
+        ));
     };
 
     // Parse the major and minor from the provided semantic version.
@@ -357,7 +360,7 @@ pub fn parse_gleam_add_specifier(package: &str) -> Result<(EcoString, Requiremen
                 ),
             });
         }
-    };
+    }?;
 
     Ok((package.into(), requirement))
 }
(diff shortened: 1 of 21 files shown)


lpil:
I have published the new package as v4!

Kaholaz:
> I have published the new package as v4!

@lpil nice :D I rebased and tested the changes.

compiler-cli/src/dependencies.rs
```diff
@@ -1075,8 +1078,8 @@ fn provide_package(
     match provided.get(&package_name) {
         Some(package) if package.source == package_source => {
             // This package has already been provided from this source, return the version
-            let version = hexpm::version::Range::new(format!("== {}", &package.version));
-            return Ok(version);
+            let version = GleamVersion::new(format!("== {}", &package.version))?;
+            return Ok(version.hex().clone());
```
lpil, line 1082: Could make a function here that consumes the version rather than taking a reference, cloning, and throwing away the original.

```suggestion
            return Ok(version.into_hex());
```

Can be used on line 1139 also

compiler-core/src/dependency.rs
```diff
@@ -137,19 +134,47 @@ but it is locked to {locked_version}, which is incompatible.",
...
 
-pub trait PackageFetcher {
-    fn get_dependencies(&self, package: &str) -> Result<hexpm::Package, Box<dyn StdError>>;
+#[derive(Debug)]
+pub enum PackageFetchError {
+    ApiError(hexpm::ApiError),
+    FetchError(String),
+    IncompatibleLockedVersion(String),
+}
+impl std::fmt::Display for PackageFetchError {
```
lpil, line 143: We avoid implementing traits in this codebase, especially `std::fmt::Display` as being explicit about what the format is at point of writing clearer and less error prone.

What was the motivation for adding this? I couldn't find where it is used.

lpil (commented):
Wonderful!! Thank you! I've left a couple notes inline

compiler-core/src/dependency.rs
Kaholaz: `PackageFetchError` implements `std::fmt::Display` to be able to implement `std::error::Error`. The reason for the addition of this error is that `DependencyResolver` needs a concrete associated error type for it's errors. Previously it just erred with `Box<dyn Error>`. Is there a better way for me to resolve this issue?

compiler-cli/src/dependencies.rs
Kaholaz: Fixed :D

compiler-core/src/dependency.rs
Kaholaz: I changed this to use `thiserror` instead. I am unsure of how I should format the errors, so i kept them unformatted.

lpil (commented):
Hello! Thank you for this. I just published that new hexpm package, is there a small upgrade to do in this PR now that can be used?

Kaholaz:
@lpil Thank you; I have used the hexpm update to avoid a call to clone :D

lpil:
Some linter issues to resolve!

lpil (approved):
Bravo!! Thank you very much!

PR #4483, opened 2025-04-19
Make analysis of lists, tuples, negation operators, `panic`, `echo` and `todo` fault tolerant

### LINKED ISSUE #4482: Analysis of lists stops at the first error
If I write `[1, "a", 1 + "a"]` the compiler only show an error for the first `"a"` but won't show an error for the second expression trying to sum an int and a string. I think we'd have to make analysis fault tolerant for lists in order to enable this.

giacomocavalieri:
This PR closes #4482 and also improves fault tolerance for tuples, negation operators (`-` and `!`), `panic`, `echo` and `todo`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -66,6 +66,13 @@
       Person(name:, age:)
   ```
 
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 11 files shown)


lpil (approved):
Wonderful, thank you!

PR #4485, opened 2025-04-19
Improve generate function code action

giacomocavalieri:
This PR improves the generate function code action in a couple of ways:
- There was a bug where it would generate invalid code if the call had a variable used twice (e.g. `wibble(one, one)`)
- Now if someone uses a label and a variable, those two distinct names are used instead of duplicating the label name. For example, if you have this piece of code:
  ```gleam
  let number = 1
  let list = [1, 2, 3, 1]
  remove(each: number, in: list)
  ```
  The generated `remove` function would now look like this:
  ```diff
  +fn remove(each number: Int, in list: List(Int)) -> List(Int) {
  -fn remove(each each: Int, in in: List(Int)) -> List(Int) {
    todo
  }
  ```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -156,11 +156,37 @@
     }
   }
   ```
+
   ([Surya Rose](https://github.com/GearsDatapacks))
 
 - The language server now provides hover, autocomplete and goto definition
   for constant definitions.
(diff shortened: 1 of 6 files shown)


lpil (approved):
Thank you!

PR #4486, opened 2025-04-19
Improve the error message for type constructors used with the wrong number of type parameters

### LINKED ISSUE #4464: Improve error message with type constructors with no parameters
```gleam
pub type ABC(a) {
  A(val: a)
}

pub type B(a) {
  B(ABC, ABC(a))
}
```
Produces this error:
```txt
error: Incorrect arity
  ┌─ /src/main.gleam:7:5
  │
6 │   B(ABC, ABC(a))
  │     ^^^ Expected 1 argument, got 0

Functions and constructors have to be called with their expected number of
arguments.
```
I think error message could be improved here:
```diff
 error: Incorrect arity
   ┌─ /src/main.gleam:7:5
   │
 6 │   B(ABC, ABC(a))
+  │     ^^^ Expected 1 type argument, got 0
-  │     ^^^ Expected 1 argument, got 0
 
+ `ABC` requires 1 type argument, but none were provided.
- Functions and constructors have to be called with their expected number of
- arguments.
```
Citing functions is a bit misleading here, I think internally in the compiler we call `ABC` a _type constructor_ but that term is never explained anywhere in the tour and we only ever use _constructor_ to refer to the variants of a type, not the type constructor itself; so that's a bit confusing as well. I think it would be simpler and less confusing if we just said that the type needs parameters that were not provided

For reference this is what Rust does:
```txt
error[E0107]: missing generics for struct `ABC`
  --> src/main.rs:13:21
   |
13 | struct B<A>(ABC<A>, ABC);
   |                     ^^^ expected 1 generic argument
   |
note: struct defined here, with 1 generic parameter: `A`
  --> src/main.rs:9:8
   |
9  | struct ABC<A> {
   |        ^^^ -
help: add missing generic argument
   |
13 | struct B<A>(ABC<A>, ABC<A>);
   |                        +++


Generic type 'ABC' requires 1 type argument(s).
```
I'm not sure about pointing to the type and suggesting to add one, but the wording of the error is nice, we could take inspiration from that!

giacomocavalieri:
This PR closes #4464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,31 @@
   errors as soon as it finds one.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The error message for types used with the wrong number of arguments has been
+  improved. For example, this piece of code:
+
+  ```gleam
+  type Wibble(a)
+
+  type Wobble {
+    Wobble(Wibble)
+  }
+  ```
+
+  Produces the following error:
+
+  ```txt
+    error: Incorrect arity
+    ┌─ /src/one/two.gleam:5:10
+    │
+  5 │   Wobble(Wibble)
+    │          ^^^^^^ Expected 1 type argument, got 0
+
+  `Wibble` requires 1 type argument but none where provided.
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you!

PR #4487, opened 2025-04-20
Improve deprecated environment variable warning message

GearsDatapacks:
I noticed that although the `HEXPM_USER` and `HEXPM_PASS` variables are now deprecated as of Gleam 1.10, no information is provided in the warning message about what to use instead of them. This PR improves that.

Before:
```
warning: Use of deprecated environment variable

The environment variable `HEXPM_PASS` is deprecated.
```
After:
```
warning: Use of deprecated environment variable

The environment variable `HEXPM_PASS` is deprecated.

Use the `HEXPM_API_KEY` environment variable instead.
```

### DIFF
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -1,7 +1,8 @@
 use crate::{cli, fs::ConsoleWarningEmitter, http::HttpClient};
 use gleam_core::{
-    Error, Result, Warning, encryption, hex, paths::global_hexpm_credentials_path,
-    warning::WarningEmitter,
+    Error, Result, Warning, encryption, hex,
+    paths::global_hexpm_credentials_path,
+    warning::{DeprecatedEnvironmentVariable, WarningEmitter},
 };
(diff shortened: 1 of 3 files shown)


compiler-cli/src/hex/auth.rs
```diff
@@ -173,6 +176,9 @@ fn ask_password(warnings: &mut Vec<Warning>) -> std::result::Result<String, Erro
         .inspect(|_| {
             warnings.push(Warning::DeprecatedEnvironmentVariable {
                 name: PASS_ENV_NAME.into(),
+                message: Some(format!(
+                    "Use the `{API_ENV_NAME}` environment variable instead."
+                )),
```
lpil, line 181: Error messages should only ever be constructed in the error module, error and warning values we control should never hold text information themselves. This could be an enum indicating which deprecated environment variable it is

lpil (commented):
Good idea, this is much better. I've left a small note inline on the data structure design

lpil (approved):
Thank you!

PR #4489, opened 2025-04-20
Fix bug when transitioning from hex to git deps

### LINKED ISSUE #4488: Build tool failure when transitioning from Hex to Git deps
Steps to reproduce:
- Create a new project, and add a Hex dependency (For example, the standard library)
- Edit gleam.toml to change the Hex dependency to a git dependency of the same name
- Run `gleam deps update`
Error produced:
```
error: Shell command failure

There was a problem when running the shell command `git`.

The error from the shell command was:

    error: The following untracked working tree files would be overwritten by checkout:
        LICENCE
        README.md
        gleam.toml
        src/dict.mjs
        src/gleam/bit_array.gleam
        ...
```
This is because we are trying to run `git init` in the `build/package/gleam_stdlib` directory, which already contains files, so we cannot `git checkout` to the desired `ref`.

We can fix this by removing the directory every time before resolving git dependencies.

## Temporary solution
If you run into this issue before the fix is released, just run `gleam clean` before `gleam deps update`

GearsDatapacks:
Fixes #4488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -207,3 +207,7 @@
 - Fixed a bug where the "generate function" code action could generate invalid
   code when the same variable was passed as an argument twice.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where replacing a Hex dependency with a Git dependency of the
+  same name would cause the build tool to fail.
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you

Issue #4476, opened 2025-04-18
Show retirement notice when adding a retired package


giacomocavalieri:
It would be nice if `gleam add` showed a warning (and the retirement notice) when adding a retired package

lpil:
Good idea!

vnikjr:
i looked into the code and it seems that there isn't a function for getting the retirement status of a package in the `hexpm` crate, which would be required to implement this feature

i have opened an [issue](https://github.com/gleam-lang/hexpm-rust/issues/44) in the hexpm github to add this feature.

ce-eboutet:
<img width="782" height="161" alt="Image" src="https://github.com/user-attachments/assets/379663b0-2472-4d84-bac8-926f2db02505" />

Since only one of the versions contains a retirement status, should we display the warning only on a particular version ?

So in this case `gleam add package@0.10` will not display a warning but either `gleam add package` or `gleam add package@0.1.1` will ?

GearsDatapacks:
Retired versions are different from retired packages. That's a separate issue: #3854

ankddev:
I can work on this, but could you give me example of retired package for testing, since I'm not able to find one?
UPD: I found [`gleam_retry`](https://hex.pm/packages/gleam_retry), I hope that it is a retired package. I have some research and found that a package considered retired if all of its versions are retired. Am I right? Do we need a fix of #3854 before?

lpil:
I don't think we need to implement that first.

I would like to omit a warning if the version that is resolved has been retired. If the version has not been retired then no warning, even if others have been retired.

PR #4491, opened 2025-04-20
Added a retirement notice when adding a retired package 

vnikjr:
implements #4476

### Notes:
- I'm not sure how to test this as i do not know any retired packages to test it on
- this is my first pr, sorry if anything is off or unprofessional :)

### DIFF
--- a/compiler-cli/src/add.rs
+++ b/compiler-cli/src/add.rs
@@ -3,6 +3,7 @@ use camino::{Utf8Path, Utf8PathBuf};
 use gleam_core::{
     Error, Result,
     error::{FileIoAction, FileKind},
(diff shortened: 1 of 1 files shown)


lpil (commented):
Thank you!

Rather than making a new HTTP request for this let's keep this information from when we talk to the Hex API earlier during version resolution. This will be quite fiddly, but we don't want to be talking to the API in multiple places.

vnikjr:
Could you clarify where i can get the RetirementStatus/ReleaseMeta of the package, i couldn't find where it were being requested during version resolution

thanks.

lpil:
Hi @EvelynTheSilly! Sorry I didn't see your comment before. Are you still intending to work on this?

PR #4493, opened 2025-04-21
Fix caching bug with git dependencies

### LINKED ISSUE #4492: Git dependencies don't change remote URLs when updating `gleam.toml`
Steps to reproduce:
- Create a new project and add a git dependency
- Run `gleam deps update`
- Change the `git` field of the git dependency to something else
- Run `gleam deps update`

The manifest is updated, but the remote of the git repository in `build/packages` is not. This is because we are using `git remote add`, which does not update the remote if it already exists, causing it to be stuck as what is originally set, until `gleam clean` is run.

We can fix this a few ways. The most straightforward is probably to run `git remote remove origin` before adding the remote, to ensure that it is always set properly.

GearsDatapacks:
Fixes #4492 
I've also added a comment explaining why the current method is used, and how we might want to optimise it in the future. This should hopefully make it more clear what is happening in this function.

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -920,6 +920,36 @@ fn execute_command(command: &mut Command) -> Result<std::process::Output> {
     }
 }
 
+/// Downloads a git package from a remote repository. The commands that are run
+/// looks like this:
+///
+/// ```sh
+/// git init
+/// git remote remove origin
+/// git remote add origin <repo>
+/// git fetch origin
+/// git checkout <ref>
+/// git rev-parse HEAD
+/// ```
+///
+/// This is somewhat inefficient as we have to fetch the entire git history before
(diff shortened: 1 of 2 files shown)


compiler-cli/src/dependencies.rs
```diff
@@ -940,18 +970,28 @@ fn download_git_package(
...
         .output();
 
+    let _ = execute_command(
+        Command::new("git")
+            .arg("remote")
+            .arg("add")
+            .arg("origin")
+            .arg(repo)
+            .current_dir(&package_path),
+    )?;
```
lpil, line 993: Does removing and adding the remote again cause it to re-fetch?

GearsDatapacks: If it's the same origin, doesn't seem like it

lpil (approved):
Brilliant, thank you!!

PR #4495, opened 2025-04-21
Some refactoring to expression inference

giacomocavalieri:
While working on the fault tolerance of lists and tuples I noticed a couple of functions that could not fail but returned a result anyway. I've changed those and made sure some more functions that can't fail will not return an error!
With this we're very close to making all the `infer_` functions fault tolerant, now there's only one or two left that are a bit more tricky


lpil (approved):
Wonderful! Thank you

PR #4496, opened 2025-04-22
Update bool assert version constraint

GearsDatapacks:
I noticed that I forgot to update the feature tracking code for bool assert - it will now correctly require v1.11.


lpil (approved):
Oops I missed that. Thank you

PR #4500, opened 2025-04-24
Warn for individual unreachable patterns

### LINKED ISSUE #2616: Repeated case clauses do not raise compiler warnings in alternative clauses
See this example (modified from the `Alternative patterns` tutorial):

```gleam
import gleam/io
import gleam/int

pub fn main() {
  let number = int.random(10)
  io.debug(number)

  let result = case number {
    1 -> "This is my favourite number"
    2 | 4 | 6 | 8 -> "This is an even number"
    1 | 3 | 5 | 7 | 2 -> "This is an odd number"
    4 -> "We checked this already!"
    _ -> "I'm not sure"
  }
  io.debug(result)
}
```

Note the repeated cases of `1` and `2`, which do not trigger a warning. However, repeating `4` by itself produces a warning. See the following output:
```
8
"This is an even number"

warning: Unreachable case clause
   ┌─ /src/main.gleam:12:5
   │
12 │     4 -> "We checked this already!"
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

This case clause cannot be reached as a previous clause matches
the same values.

```

GearsDatapacks:
Fixes #2616 
This PR changes the unreachable pattern system to warn for individual patterns, rather than the whole unreachable clause. This allows us to properly warn for unreachable patterns in an alternative pattern.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -289,3 +289,8 @@
 
 - Fix slightly wrong error message for missing main function in test module.
   ([Samuel Cristobal](https://github.com/scristobal))
+
+- Fixed a bug where the compiler would not properly warn for unreachable
+  patterns in a `case` expression when the clause matched on multiple
+  alternative patterns.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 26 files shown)


lpil (approved):
Thank you!!

PR #4502, opened 2025-04-25
Improve generated code for `let assert` on the Erlang target

### LINKED ISSUE #4145: Referencing result of earlier segment when pattern matching generates invalid Erlang
The following code references the result of the pattern's first segment in its second segment:

```gleam
pub fn main() {
  let assert <<length, _:size(length)-unit(8)>> = <<3, 1, 2, 3>>
}
```

Building for Erlang gives:

```
/bit_array_test/src/bit_array_test.gleam:6:16: variable 'Length' is unbound
error: Shell command failure
```

Erlang can handle this kind of self-referencing pattern so Gleam probably should too.

The following Erlang works as expected:

```erlang
test_pattern() ->
    _assert_subject = <<3, 1, 2, 3>>,
    case _assert_subject of
        <<Length, _:Length/unit:8>> -> _assert_subject;
        _assert_fail -> erlang:error()
    end.
```

Gleam version: 1.6.3 and 1.7.0-rc3.

### LINKED ISSUE #4497: Simplifiy generated Erlang code for `let assert`
Currently, when generating Erlang for code like this:
```gleam
let assert Ok([1, 2, 3]) = x
```
The following code is generated:
```erlang
_assert_subject = X,
{ok, [1, 2, 3]} = case _assert_subject of
  {ok, [1, 2, 3]} -> _assert_subject;
  _ -> erlang:error(...)
end. 
```
This is somewhat inefficient; we match the same pattern twice, which is duplicated effort for no real reason.
Since we don't bind any variables in this case, we could simply generate the following:
```erlang
_assert_subject = X,
case _assert_subject of
  {ok, [1, 2, 3]} -> nil;
  _ -> erlang:error(...)
end. 
```
If we wanted to bind a variable, like this:
```gleam
let assert Ok([1, y, 3]) = x
```
We could generate code like this:
```erlang
_assert_subject = X,
Y = case _assert_subject of
  {ok, [1, Y, 3]} -> Y;
  _ -> erlang:error(...)
end. 
```
Finally, if we wanted to bind multiple variables:
```gleam
let assert Ok([a, b, c]) = x
```
We could generate the following:
```erlang
_assert_subject = X,
{A, B, C} = case _assert_subject of
  {ok, [A, B, C]} -> {A, B, C};
  _ -> erlang:error(...)
end. 
```

GearsDatapacks:
Closes #4497 and fixes #4145 
This PR improves the generated code for `let assert` in a couple of ways:
- Instead of pattern matching on the value twice, we match once, extract the variables, then return a tuple from the `case` expression.
- If a pattern always matches (e.g. a plain variable or tuple pattern) we omit the check entirely and just assign like a simple `let` assignment.
This simplification fixes the other bug by default, as we no longer ignore variables in the first pattern match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -169,6 +169,10 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The compiler will now generate more efficient code for `let assert` on the
+  Erlang target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Build tool
 
 - The build tool now supports placing modules in a directory called `dev`,
@@ -343,3 +347,8 @@
 - Fixed a bug where the language server would generate invalid code for the
   "fill in missing labels" code action.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where referencing an earlier segment of the bit array in a bit
+  array pattern in a `let assert` assignment would generate invalid code on the
+  Erlang target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -2235,6 +2235,28 @@ impl TypedPattern {
             labelled,
         })
     }
+
+    /// Whether the pattern always matches. For example, a tuple or simple
+    /// variable assignment always match and can never fail.
+    #[must_use]
+    pub fn always_matches(&self) -> bool {
+        match self {
+            Pattern::Variable { .. } | Pattern::Discard { .. } => true,
+            Pattern::Assign { pattern, .. } => pattern.always_matches(),
+            Pattern::Tuple { elements, .. } => {
+                elements.iter().all(|element| element.always_matches())
+            }
+            Pattern::Int { .. }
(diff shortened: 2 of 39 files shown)


giacomocavalieri:
This entire portion of code is being worked on by the decision tree PR, which already gets rid of needless checks using the decision tree the compiler produces; and merging this would likely cause loads of conflicts with it. I'd love to avoid that! 😁

GearsDatapacks:
Ah I see. I thought that PR was only for the javascript target. My bad!

giacomocavalieri:
Ah Gears I'm so sorry I totally messed up! You're right, it is only for the js target 🙇‍♂️
This would not overlap all that much with the decision tree branch. If you want to, though you could still use the decision tree on the Erlang target as well to see if the let assert could be simplified, I'm not sure what would be easier but that's what I'm doing on the JS target if you want to have a look at the implementation

Sorry again for the confusion!

GearsDatapacks:
All good. I think this implementation is going to be fine for now. I imagine Erlang performs some optimisations using a decision tree also. Perhaps once the plumbing is in place for the decision tree we can improve upon it further

lpil (approved):
Thank you!!!

PR #4506, opened 2025-04-26
Fix purity detection bugs

### LINKED ISSUE #4504: Purity tracker reports warnings in stdlib tests
There is bug in the purity tracker: In the tests for the standard library, a set of helper functions are defined, using FFI, to test for certain properties (equality, inequality, `Ok` variants, etc.). The purity tracker marks them as pure and so a warning is emitted when they are used
The reason for this bug is that the purity tracker special cases FFI in stdlib function, assuming it to be pure
However, in this case the intention is not to be pure, the intention is to panic, so it should not warn here.

A potential solution to this is to hardcode the set of Gleam stdlib modules which *are* pure, which would not include the `gleam/should` test module.

### LINKED ISSUE #4505: Functions are failed to be marked as impure if they use an impure pipeline
For example:
```gleam
fn impure(_x) {
  panic
}

fn also_impure(x) {
  x |> impure
}

pub fn main() {
  also_impure(10)
  Nil
}
```
This reports a warning for `also_impure` being pure, which is isn't. Purity tracking currently doesn't take into account pipelines.

GearsDatapacks:
Fixes #4504 and fixes #4505

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -320,9 +320,10 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             Target::JavaScript => implementations.uses_javascript_externals,
         };
 
-        let is_non_io_standard_library_function = environment.current_package
-            == STDLIB_PACKAGE_NAME
-            && environment.current_module != "gleam/io";
+        let is_non_io_standard_library_function = is_non_io_standard_library_module(
+            &environment.current_package,
+            &environment.current_module,
+        );
 
         let purity = if is_non_io_standard_library_function {
             Purity::TrustedPure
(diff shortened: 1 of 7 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -320,9 +320,10 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             Target::JavaScript => implementations.uses_javascript_externals,
         };
 
-        let is_non_io_standard_library_function = environment.current_package
-            == STDLIB_PACKAGE_NAME
-            && environment.current_module != "gleam/io";
+        let is_non_io_standard_library_function = is_non_io_standard_library_module(
+            &environment.current_package,
+            &environment.current_module,
+        );
```
lpil, line 326: I know it's unlikely, but if we added a new stdlib module the list would be out of date.

Could we instead check the origin and any non `gleam/io` modules in `src` would be trusted pure?

lpil (commented):
Thank you! I've left a note line

lpil:
actually, I'll apply my own feedback!

compiler-core/src/type_/expression.rs
GearsDatapacks: Yeah, that also works

Issue #4253, opened 2025-02-17
Create benchmarks for use of lists and custom types on JS


lpil:
If we have a benchmark of all the operations one would do with lists and custom types we can use this to evaluate codegen changes to improve performance.

We want to be able to run this on all 3 major JS engines.

unknown:
I'd be happy to work on this if you could provide some guidance on what you have in mind for operations on custom types.

lpil:
Everything and anything at all!

PR #4511, opened 2025-04-27
Create list benchmarks

unknown:
Creates a benchmark suite for list operations, as outlined in https://github.com/gleam-lang/gleam/issues/4253.

I've marked this PR as a draft because it depends the external package `gleamy_bench`, and I'm not sure if we're comfortable doing this. I also want to hear feedback on if these test functions sufficiently cover the range of list operations:

1. `odd_nums_between` covers list construction and list reversing
2. `count_ones` covers list destructuring via `case`
3. `slice` covers list destructuring via `case` with conditionals, list construction, and list reversing


lpil (commented):
Thank you! Very useful! Could you make it run on Erlang too, and add type annotations for the top level functions please.

unknown:
> Thank you! Very useful! Could you make it run on Erlang too, and add type annotations for the top level functions please.

Yeah, will do that! And to be clear, what should I do about depending on `gleamy_bench`?

lpil:
Depending on that package sounds good to me 👍

unknown:
> Depending on that package sounds good to me 👍

Cool, should I make a separate project and PR for custom types? Or should they be combined into a single project and this PR?

lpil:
All in the same project sounds good to me. Do you have a preference for one PR or more?

unknown:
> All in the same project sounds good to me. Do you have a preference for one PR or more?

I’m not sure when I’ll have the time to work on the custom types portion. I don’t have a preference, so it depends on if you’re okay with having this PR open for a while longer.

lpil (approved):
I'll merge this! Thank you

PR #4514, opened 2025-04-27
Fix "convert to use" code action when used on function calls with labelled out-of-order arguments

### LINKED ISSUE #4498: "convert to use" action - error when parameters are out of order
Issue with "convert to use" refactoring action.

When using parameter tags to call a function with "out-of-order" parameters, the refactoring action produces invalid code.

```gleam
fn refactors_okay(items) {
  list.fold(items, from: 0, with: fn (total, item) { 
    total + item 
  })
}

fn buggy(items) {
  // now the starting point "0" comes first
  list.fold(0, over: items, with: fn (total, item) {
    total + item
  })
}

// the above is valid gleam code but refactors to this invalid code
fn buggy_refactor(items) {
  use total, item <- list.fold(0) // missing the "items" parameter
  total + item
}
```

I organically arrived at out of order parameters because I had

```gleam
my_object.long_involved_constructor()
|> list.fold(over: items, with: fn (a, b) { todo })

// then I used the "inline pipe" action
list.fold(my_object.long_involved_constructor(), over: items, with: fn (a, b) { todo })

// then I used the "convert to use" action
use a, b <- list.fold(my_object.long_involved_constructor())
todo

// but now the "items" variable has gone missing
```

Gleam version 1.9.1

giacomocavalieri:
If a function was called with arguments that ended up being out of order because of labels, the "convert to use" code action would end up removing more arguments than it actually should. This PR fixes this bug, closes #4498

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -369,3 +369,8 @@
 - Fixed a bug where a using the pipe operator in the `size` option of a bit array
   segment would generate invalid code on the Erlang target.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where the language server would generate invalid code for the
+  "convert to use" code action, when used on a function call with labelled
+  arguments.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!

PR #4516, opened 2025-04-27
Fix pipe in size option on erlang

### LINKED ISSUE #3944: Invalid Erlang generated for pipe expression in bit array size segment
This Gleam code that uses a pipe in a `size` bit array segment:

```gleam
pub fn main() {
  <<<<0xAE>>:bits-size(5 |> identity)>>
}

fn identity(x) {
  x
}
```

Produces this error output from `gleam build` on Gleam 1.6.2:

```
  Compiling bit_array_test
/Users/richardviney/Code/bit_array_test/src/bit_array_test.gleam:8:39: syntax error before: ','
/Users/richardviney/Code/bit_array_test/build/dev/erlang/bit_array_test/_gleam_artefacts/bit_array_test.erl:4:2: function main/0 undefined
%    4| -export([main/0]).
%     |  ^

/Users/richardviney/Code/bit_array_test/src/bit_array_test.gleam:6:2: spec for undefined function main/0
%    6|   <<<<0xAE>>:bits-size(5 |> identity)>>
%     |  ^

error: Shell command failure

```

The generated Erlang code is here: [bit_array_test.erl.txt](https://github.com/user-attachments/files/17985959/bit_array_test.erl.txt)

---

I'm uncertain whether pipes in bit array segments are meant to work, but blocks and case expressions work so pipes probably should too.

GearsDatapacks:
Fixes #3944

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -365,3 +365,7 @@
 - Fixed a bug where the compiler would crash when type-checking code containing
   an assignment pattern inside a bit-array pattern.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where a using the pipe operator in the `size` option of a bit array
+  segment would generate invalid code on the Erlang target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!!

PR #4517, opened 2025-04-28
Fix assignments in bit arrays

### LINKED ISSUE #3251: Gleam compiler crash when pattern matching bit arrays
I have triggered a compiler crash when pattern matching on bit arrays.

## Versions

Gleam version 1.2.1
OS: Ubuntu 22.04.4

## Steps to reproduce

```
$ gleam new bit_array_crash
$ cd bit_array_crash

$ printf "import gleam/io

pub fn yo(bit_array: BitArray) {
  case bit_array {
    <<0 as n:size(8), _:bytes>> -> {
      io.debug(n)
      Nil
    }
    _ -> Nil
  }
}
" > src/bit_array_crash.gleam

$ gleam run

Panic: compiler-core/src/erlang/pattern.rs:201
        Pattern segment match not recognised
Gleam version: 1.2.1
Operating system: linux

variable set.
```

The code is this:

```gleam
import gleam/io

pub fn yo(bit_array: BitArray) {
  case bit_array {
    <<0 as n:size(8), _:bytes>> -> {
      io.debug(n)
      Nil
    }
    _ -> Nil
  }
}
```

It is the `0 as n:size(8)` that seems to be the issue.  Also if you change to this `<<_ as n:size(8), _:bytes>>` or this `<<0:size(8), _ as n:bytes>>`, you will also get the compiler crash.

## Compiler trace

[trace.txt](https://github.com/user-attachments/files/15746479/trace.txt)

### LINKED ISSUE #3375: Fatal compiler bug when compiling BitArray pattern to Erlang
When pattern matching on BitArrays and using the `as` pattern, it causes a fatal compile error when compiling to Erlang.

```gleam
pub fn main() {
  let assert <<65 as a, _rest:bytes>> = <<65>>
}
```

Gleam: 1.3.0
OS: macOS 15.0

When running `gleam build`, the following error occurs

```
  Compiling erlang_pattern_repro

warning: Unused variable
  ┌─ /Users/danielle/Repositories/Playground/gleam/erlang_pattern_repro/src/erlang_pattern_repro.gleam:2:22
  │
2 │   let assert <<65 as a, _rest:bytes>> = <<65>>
  │                      ^ This variable is never used

Hint: You can ignore it with an underscore: `_a`.
error: Fatal compiler bug!

Panic: compiler-core/src/erlang/pattern.rs:206
        Pattern segment match not recognised
Gleam version: 1.3.0
Operating system: macos

variable set.
```

GearsDatapacks:
Fixes #3375 and fixes #3251 (they are the same bug)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -361,3 +361,7 @@
 - Fixed instances where the "Extract variable" code action would produce invalid
   code, most noticeable in code inside `case` clauses and `use` expressions.
   ([Matias Carlander](https://github.com/matiascr))
+
+- Fixed a bug where the compiler would crash when type-checking code containing
+  an assignment pattern inside a bit-array pattern.
+  ([Surya Rose](https://github.com/GearsDatapacks))
--- a/compiler-core/src/erlang/pattern.rs
+++ b/compiler-core/src/erlang/pattern.rs
@@ -190,6 +190,39 @@ fn pattern_segment<'a>(
         | Pattern::Float { .. } => {
             print(value, &mut vars.borrow_mut(), env, &mut guards.borrow_mut())
         }
+
+        Pattern::Assign { name, pattern, .. } => {
+            vars.borrow_mut().push(name);
+            let variable_name = env.next_local_var_name(name);
+            match pattern.as_ref() {
+                Pattern::Int { value, .. } => {
+                    guards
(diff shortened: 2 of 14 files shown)


lpil (approved):
Thank you!!

PR #4518, opened 2025-04-28
Delete gleam compile file after compilation

### LINKED ISSUE #4474: Don't put gleam@@compile.erl in the _gleam_artefacts dir

DanKhaitov:
This PR ensures that gleam@@compile.erl is deleted after compilation. It is not needed after compilation and causes problems with integrations that run erlc on gleam builds.

I added a unit test in `test/delete_gleam_compile_file` that runs `gleam build` on a project and verifies artefact deletion.

Fixes #4474


lpil (commented):
Thank you for this!

It will need to be implemented inside the beam compiler and not inside the project compiler as otherwise the file won't get cleaned up when not using the project compiler, which is what the issue was for.

It could delete the file immediately after the escript instance has been successfully spawned, or it could be deleted using the `Drop` trait.

lpil:

PR #4523, opened 2025-04-29
Bug Fix: store typescript_declarations config and rebuild if changed

### LINKED ISSUE #4520: Typescript declarations are not emitted with existing build directory
Given a gleam application with `target = "javascript"` will build the application just fine. However when I enable `typescript_declarations = true` in the `gleam.toml` config and build the project again, the `.d.ts` are not emitted until I `rm -rf` the `build` directory that was emitted without the declarations turned on.

---
gleam v1.10.0
runtime: javascript

daniellionel01:
Code to fix https://github.com/gleam-lang/gleam/issues/4520

It currently creates a file `gleam-toml.lock` where it stores the configuration for `typescript_declarations`. If that has changed since the last build, it `rm -rf`s the build directory to make sure that it emits typescript declaration files this time. Might be a bit overkill and only necessary to for example `rm -rf` the dev/javascript directory, I am not sure.

Has been tested locally in a dedicated gleam project. But I can add something to the test suite to make it part of CI/CD if that is a requirement.

So in summary, some aspects that should be discussed in more depth:
- do we need to `rm -rf` the whole build directory or only parts of it? for example `build/dev/javascript`?
- where to store the `gleam-toml.lock` file
- what to call it (`gleam-config.lock`, `gleam.toml`, `config.lock`, ...)
- the format to store it in (toml? json?)
- add test to CI/CD?

Cheers!


CHANGELOG.md
```diff
@@ -370,6 +370,9 @@
   segment would generate invalid code on the Erlang target.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where enabling `typescript_declarations = true` wouldn't generate TypeScript declarations unless the build directory was manually deleted. The compiler now automatically rebuilds the project when this configuration changes.
+  ([daniellionel01](https://github.com/daniellionel01))
```
lpil, line 374: Wrap long lines to match the existing entries please

test/javascript_typescript_declarations/gleam.toml
```diff
@@ -0,0 +1,14 @@
+name = "javascript_typescript_declarations"
```
lpil, line 1: This test project doesn't seem to be used anywhere? What is it for?

Could be a good foundation for an integration test in a CI workflow.

compiler-core/src/build/project_compiler.rs
```diff
@@ -297,9 +356,7 @@ where
         // Write the TypeScript prelude, if asked for
         if self.config.javascript.typescript_declarations {
             let path = build.join("prelude.d.mts");
-            if !self.io.is_file(&path) {
-                self.io.write(&path, crate::javascript::PRELUDE_TS_DEF)?;
-            }
+            self.io.write(&path, crate::javascript::PRELUDE_TS_DEF)?;
```
lpil, line 359: Why has this been changed?

compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +264,62 @@ where
             })
     }
 
+    /// Checks if the TypeScript declarations setting has changed since the last build.
+    /// If it has changed, we clear the build directory to make sure it generates TypeScript definitions.
+    pub fn check_javascript_config(&self) -> Result<(), Error> {
```
lpil, line 269: Let's give this a more descriptive name that says what it does.

compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +264,62 @@ where
...
+        if !self.target().is_javascript() {
+            return Ok(());
+        }
```
lpil, line 272: We also only care about this if configured to perform code generation, otherwise it does nothing.

compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +264,62 @@ where
...
+
+        let build_path = self
+            .paths
+            .build_directory_for_target(self.mode(), self.target());
+        let config_path = build_path.join("gleam-toml.lock");
+
+        let current_setting = format!(
```
lpil, line 279: Always favour the declarative serde derive API over ad-hoc serde usage.

That said, we could just write an empty file and its existence can indicate that TS was asked for. Then we don't need to do any reading or parsing.

compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +264,62 @@ where
...
+
+        self.io.mkdir(&build_path)?;
+        self.io
+            .write(&config_path, &current_setting)
+            .map_err(|e| Error::FileIo {
+                action: FileIoAction::WriteTo,
+                kind: FileKind::File,
+                path: config_path,
+                err: Some(e.to_string()),
+            })
```
lpil, line 320: There's a bug here! The build directory doesn't have all the contents it needs to have after being reset.

We can fix it by not deleting the entire directory. It would be more proper for this compiler instance to only delete the sub-directory it owns.

compiler-core/src/build/project_compiler.rs
```diff
@@ -188,6 +188,9 @@ where
         // verify that this version is appropriate.
         self.check_gleam_version()?;
 
+        // Check if JavaScript configuration has changed
```
lpil, line 191: More detail in the documentation please 🙏

lpil (commented):
Thank you! I've left some notes inline.

Please un-draft and tag me when ready for a review.

compiler-core/src/build/project_compiler.rs
daniellionel01: Has been reverted, not sure what I thought there

test/javascript_typescript_declarations/gleam.toml
daniellionel01: I'm not very familiar with testing in rust, so I am not sure how to setup an testing workflow here. I've removed the project though

CHANGELOG.md
daniellionel01: done 👍

compiler-core/src/build/project_compiler.rs
daniellionel01: done 👍

compiler-core/src/build/project_compiler.rs
daniellionel01: done 👍 the file is now called `typescript-declarations.lock`

compiler-core/src/build/project_compiler.rs
daniellionel01: it now deletes the build/dev/javascript directory, is that ok? or should it delete the build/dev/javascript/[project name] directory?

daniellionel01: I'm not 100% sure if I understand this correctly. So right now it deletes the `build/dev/javascript` directory, but that shouldn't happen? By "sub-directory it owns" are you referring to `build/dev/javascript/[project name]`? Because from what I am able to tell we also need to generate the .d.mts files inside of the adjacent directories f.e. gleam_stdlib.
We could also delete all of the *.mjs files inside of the `build/dev/javascript` directory, would that be better?

daniellionel01:
> Thank you! I've left some notes inline.
> 
> Please un-draft and tag me when ready for a review.

Thank you! I've made changes for all of your comments. 

> Thank you! I've left some notes inline.
> 
> Please un-draft and tag me when ready for a review.

@lpil addressed all comments, thank you! Will be awaiting your feedback 🤘

I couldn't get the test for CI working yet, but have tested it locally quite a bit

lpil:
Could you rebase to remove the merge commit please 🙏

daniellionel01:
> Could you rebase to remove the merge commit please 🙏

@lpil  I'll be honest, I've never rebased anything and am failing hard to do so 🥲 I've tried for half an hour with GPT and messed up the branch like 3 times already. Any help is appreciated.

daniellionel01:
Created a new PR with the identical changes, sorry about the confusion @lpil 

https://github.com/gleam-lang/gleam/pull/4562

lpil:
Thank you

PR #4526, opened 2025-04-29
Regenerate snapshots

giacomocavalieri:
I noticed there were a few stale snapshots, so I've regenerated all the snapshots in `compiler_core`, getting rid of the stale ones.


lpil (approved):
Thank you!

PR #4527, opened 2025-04-30
Improve html snapshot testing

giacomocavalieri:
HTML snapshots are a bit hard to review at the moment, mainly because we always output the entire html document that is produced. Hayleigh inspired me to change the test code to use selectors to focus the snapshot on what really matters.

In order to do that I've added [`scraper`](https://crates.io/crates/scraper) as a dependency: it seems well maintained, has loads of downloads and a nice API.
I've also implemented a small macro to make the most common test case pretty straightforward to test:
```rs
assert_docs!(
  vec_of_selectors,
  "pub fn main() {

}"
)
```
This will generate the doc page for the given gleam module and only snapshot the elements that match with the given selectors. If no vector of selectors is specified then this includes the entire HTML document, the same way we did before this change.

I think the resulting code and snapshots are pretty neat!


giacomocavalieri:
Louis said they don't want to add a new dependency for this

PR #4529, opened 2025-04-30
Improve documentation printer

### LINKED ISSUE #828: Render imported types with module qualifiers in documentation

### LINKED ISSUE #2629: Inconsistent type variable names in generated documentation between types and functions
Hello! I noticed something in the documentation that is bugging me. Take this piece of code:
```gleam
pub opaque type Graph(direction, value, label) {
  Graph // This is not relevant
}

pub fn map_values(
  in graph: Graph(direction, value, label),
  with fun: fn(value) -> new_value
) -> Graph(direction, new_value, label) {
  todo
}
```
The generated docs will look something like this:
```gleam
pub opaque type Graph(direction, value, label)
// ^ Here type variable names are preserved

pub fn map_values(
  in graph: Graph(a, b, c),
  with fun: fn(b) -> d
) -> Graph(a, d, c)
// ^ Here they are not!
```
Is there a reason for this inconsistency between types and functions? In this case I'd argue that using single letters for the function types makes it way less clear, and someone using the docs would have to frequently scroll back to the `Graph` type definition to be reminded which type variable means what - I found myself thinking that a lot when I started writing this code _"mmh so the middle one is the graph values, I wonder which one is for labels and which one for the direction"_. While it won't likely be a problem for people who are familiar with a package, that's not the best experience for someone who's just getting started.

### LINKED ISSUE #3461: (hex)docs: types in function signatures should be clickable links to their documentation

GearsDatapacks:
This PR closes #2629, closes #828 and closes #3461
This PR is a continuation of #3471, thanks @yoshi-monster for doing the initial work!
What I've done in this PR:
- Extracted the logic for printing the definitions of types and values from the formatter, and made a `docs::printer::Printer` struct.
- This struct now prints highlighting HTML tags, because as Yoshi discovered, highlight.js doesn't support links.
- To support that, I've added a `ZeroWidthString` variant to the `Document` enum in the pretty printing module
- While I was doing this, I thought it made sense to rework the type printing for these, so we now have proper type-variable printing and module qualifiers.
- Since the snapshots were very large (~400 lines each), I've made a new function to pretty-print the generated documentation, including a way to turn off all the HTML highlighting so we can properly see the generated code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,11 +204,12 @@
   deprecated environment variables.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
-- When generating documentation, the build tool now prints correct type variable
-  names for function signatures. For example, this code:
+- When generating documentation, the build tool now prints type variable using
+  the same names as were used in the code for function signatures. For example,
+  this code:
 
   ```gleam
-  pub fn to_list(list: List(#(key, value))) -> Dict(key, value) {
+  pub fn to_list(entries: List(#(key, value))) -> Dict(key, value) {
     ...
   }
   ```
@@ -216,13 +217,13 @@
   Previously would have been printed as:
 
   ```gleam
-  pub fn to_list(list: List(#(a, b))) -> Dict(a, b)
+  pub fn to_list(entries: List(#(a, b))) -> Dict(a, b)
   ```
 
   But now is rendered as the following:
 
   ```gleam
-  pub fn to_list(list: List(#(key, value))) -> Dict(key, value)
+  pub fn to_list(entries: List(#(key, value))) -> Dict(key, value)
   ```
 
   ([Surya Rose](https://github.com/GearsDatapacks))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,59 @@
   deprecated environment variables.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- When generating documentation, the build tool now prints correct type variable
+  names for function signatures. For example, this code:
+
+  ```gleam
+  pub fn to_list(list: List(#(key, value))) -> Dict(key, value) {
+    ...
+  }
+  ```
+
+  Previously would have been printed as:
+
+  ```gleam
+  pub fn to_list(list: List(#(a, b))) -> Dict(a, b)
+  ```
+
+  But now is rendered as the following:
+
+  ```gleam
+  pub fn to_list(list: List(#(key, value))) -> Dict(key, value)
+  ```
+
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
+- When generating documentation, types from other modules are now rendered with
+  their module qualifiers. Hovering over them shows the full path to their module.
+  For example, this code:
+
+  ```gleam
+  import gleam/dynamic/decode
+
+  pub fn something_decoder() -> decode.Decoder(Something) {
+    ...
+  }
+  ```
+
+  Will now generate the following documentation:
+
+  ```gleam
+  pub fn something_decoder() -> decode.Decoder(Something)
+  ```
+
+  And hovering over the `decode.Decoder` text will show the following:
+
+  ```txt
+  gleam/dynamic/decode.{type Decoder}
+  ```
+
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
+- When generating documentation, types in rendered documentation code will now
+  link to their corresponding documentation pages.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Language server
 
 - The code action to add missing labels to function now also works in patterns:
--- a/compiler-core/src/docs/printer.rs
+++ b/compiler-core/src/docs/printer.rs
@@ -229,14 +229,13 @@ impl Printer<'_> {
             )
         };
 
-        let keywords = if opaque { "opaque type " } else { "type " };
+        let keywords = if opaque {
+            "pub opaque type "
+        } else {
+            "pub type "
+        };
 
-        let type_head = docvec![
-            self.keyword("pub "),
-            self.keyword(keywords),
-            self.title(name),
-            arguments
-        ];
+        let type_head = docvec![self.keyword(keywords), self.title(name), arguments];
 
         if constructors.is_empty() || opaque {
             return type_head;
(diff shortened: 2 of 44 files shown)


CHANGELOG.md
```diff
@@ -204,6 +204,59 @@
   deprecated environment variables.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- When generating documentation, the build tool now prints correct type variable
+  names for function signatures. For example, this code:
```
lpil, line 208: They were correct before! The change is that it uses the same type variable names that were used in the code.

CHANGELOG.md
```diff
@@ -204,6 +204,59 @@
...
+
+  ```gleam
+  pub fn to_list(list: List(#(key, value))) -> Dict(key, value) {
```
lpil, line 211:
```suggestion
  pub fn from_list(entries: List(#(key, value))) -> Dict(key, value) {
```

compiler-core/src/docs/snapshots/gleam_core__docs__tests__print_type_variables_in_function_signatures.snap
```diff
@@ -0,0 +1,17 @@
...
+<pre><code>pub type Dict(key, value)</code></pre>
+
+---- VALUES
+
+--- insert
+<pre><code>pub fn insert(
+  dict: Dict(key, value),
+  key: key,
+  value: value,
+) -> Dict(key, value)</code></pre>
```
lpil, line 17: Include the input code so we can tell what the snapshot is testing please 🙏

compiler-core/templates/documentation_module.html
```diff
@@ -55,7 +55,7 @@ <h2 id="{{ typ.name }}">
     {% endif %}
     <div class="custom-type-constructors">
       <div class="rendered-markdown">{{ typ.documentation|safe }}</div>
-      <pre><code class="hljs gleam">{{ typ.definition }}</code></pre>
+      <pre><code class="hljs gleam hljs-ignore">{{ typ.definition|safe }}</code></pre>
```
lpil, line 58: What're the `hljs gleam` classes used for now?

compiler-core/src/docs/printer.rs
```diff
@@ -0,0 +1,578 @@
...
+                constructors: vec![],
+                source_url: source_links.url(*location),
+                deprecation_message: match deprecation {
+                    Deprecation::NotDeprecated => "".to_string(),
+                    Deprecation::Deprecated { message } => message.to_string(),
+                },
+                opaque: false,
+            }),
+
+            _ => None,
```
lpil, line 154: No catch all patterns please 🙏

compiler-core/src/docs/printer.rs
```diff
@@ -0,0 +1,578 @@
...
+                documentation: markdown_documentation(documentation),
+                text_documentation: text_documentation(documentation),
+                source_url: source_links.url(*location),
+                deprecation_message: match deprecation {
+                    Deprecation::NotDeprecated => "".to_string(),
+                    Deprecation::Deprecated { message } => message.to_string(),
+                },
+            }),
+
+            _ => None,
```
lpil, line 211: No catch all patterns please 🙏

compiler-core/src/docs/printer.rs
```diff
@@ -0,0 +1,578 @@
...
+    fn next_letter(&mut self) -> EcoString {
+        let alphabet_length = 26;
+        let char_offset = b'a';
+        let mut chars = vec![];
+        let mut n;
+        let mut rest = self.next_type_variable_id;
+
+        loop {
+            n = rest % alphabet_length;
+            rest /= alphabet_length;
```
lpil, line 439: Don't use these "mutate this variable with this operation" operators please. Prefer the regular assignment syntax as it is clearer.

lpil (commented):
Fantastic!! Thank you!!

Some small notes inline

compiler-core/templates/documentation_module.html
GearsDatapacks: `gleam` seems redundant now, but `hljs` is referenced in the stylesheet and ensures that it's formatted the same as all the other code blocks. Perhaps we should slightly change that selector though

compiler-core/src/docs/printer.rs
```diff
@@ -0,0 +1,584 @@
...
+            let module_name = module.split('/').next_back().unwrap_or(module);
+            let qualified_name = docvec![
+                self.variable(EcoString::from(module_name)),
+                ".",
+                self.title(name)
+            ];
+            let title = eco_format!("{module}.{{type {name}}}");
+
+            self.link(
+                eco_format!("https://hexdocs.pm/{package}/{module}.html#{name}"),
```
lpil, line 473: I think this link will be wrong in these situations:

- If the module is from the root package and viewing the docs locally
- For paths or git deps, so not on Hex
- For deps that are not the most latest version of that package on Hex

For path and git deps I think we'll have to offer no link, for the root we'll need to use relative links, and the Hex one we'll need to use the current version.

lpil (commented):
Fantastic! Thank you so much! I've added one small comment inline 🙏

GearsDatapacks:
The links should all be correct now. These changes require a bit more plumbing as we now need information from the manifest about what specific version of a package is being used.

lpil (approved):
Thank you!!

Issue #3628, opened 2024-09-17
LSP ranges are incorrect when line contains multi-codepoint grapheme clusters


yoshi-monster:
The string is `"क्षि"`, or `"\u{0915}\u{094D}\u{0937}\u{093F}"`.

![image](https://github.com/user-attachments/assets/6ee7eee7-dd81-45ca-94c9-37c0dfc1999a)

From the looks of it it actually just uses the byte/code unit count.

Tested with Gleam v1.4.1, in VSCode, Zed and Helix :purple_heart:

lpil:
Thank you. I expect this'll be due to LSP using utf-16 and we're using utf-8.

aaron-loa:
Hello. I looked a bit into this. 
Supposedly by setting the position_encoding type to UTF-8 the problem should be solved.
After specifying the encoding it works correctly with editors that support UTF-8 encoding, like neovim. 
But it doesn't work with vscode.
This is the error I get:
`Error: Unsupported position encoding (utf-8) received from server Gleam Language Server`

Not sure about zed, I can't figure out how to change the path of the gleam language server to use my local binary.

Hovering is also affected by this problem, probably everything that interacts with positions.
I will try to look more into what can be done this week. But from a quick skimming through, a lot of things need to be changed :smile:

lpil:
Thank you @aaron-loa !!

lydell:
> probably everything that interacts with positions

Yep, that seems to be the case. Here’s an example with the “Add missing patterns“ code action:

```gleam
// ä
fn wibble() {
  case True {
    
  }
}
```

Becomes:

```gleam
// ä
fn wibble() {
  case True 
   False -> todo
   True -> todo
 }
}
```

aaron-loa:
I did some investigation, but completely forgot to write an update.

In compiler-core/src/line_numbers.rs the LineNumber structure doesn't do the translation between utf-8 -> to utf-16.
Ideally this should would work similary to this:
<https://github.com/artempyanykh/rust-lsp-document/blob/main/src/lib.rs>
This is a small library that does the translation between utf-16 and utf-8 indices, for language servers.

We could do this translation in the LineNumber struct, the only problem is we always need a reference to the source text. But adding the reference to the struct adds too many lifetime issues, i just gave up on this approach, but this would be the most ideal approach imo.

After this I tried a different approach, where you have to put the source text as an argument to the byte_index method, but I don't think this is a 'clean' solution, that's why I didnt' make a pr.
[...]

PR #4531, opened 2025-04-30
Fix bug with indentation detection and multibyte characters

GearsDatapacks:
This PR is part of #3628, fixing the second point mentioned in this comment: https://github.com/gleam-lang/gleam/issues/3628#issuecomment-2842807944
Basically, the code to detect indentation would skip over characters, when really it should be skipping over bytes. That should now be fixed in all places that used this method.
Since I noticed it is now used in quite a few places in the code actions module, I decided to extract it into its own function to make the code a bit simpler in various places.


lpil (approved):
Amazing! Thank you!

PR #4535, opened 2025-04-30
Remove unnecessary clones

GearsDatapacks:
I noticed a few unnecessary clones in the `code_action` module, some of which could be quite expensive, so I have removed them.


lpil (approved):
Thank you

PR #4542, opened 2025-05-01
Only print the user-friendly LSP message when stdin is a terminal

arielherself:
This PR adds a check using `std::io::IsTerminal` to avoid emitting redundant logs on proper LSP starts, so that it matches the common practice of this kind of tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -407,3 +407,7 @@
 - Fixed a bug where LSP ranges would be incorrect when a file contained characters
   that were represented wit more than one byte in UTF-8.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Only print the user-friendly LSP message when stdin is a terminal to avoid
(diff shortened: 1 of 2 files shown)


lpil (commented):
Good idea, thank you.

Put back the section about ignoring it, and update the changelog also please 🙏

arielherself:
Of course!

lpil (approved):
Lovely, thank you!!

lpil:
Hmm I'm not sure why the tests are not running. Could you rebase on main please? Maybe that'll fix it

lpil:
I think something may have gone wrong! There's some extra commits and a merge commit on this branch now instead of just your two commits

arielherself:
Oh I'm sorry, I forgot to update my fork :sob: I have just cherry-picked my commits to the latest version of main branch so the commit history is correct now.

Issue #614, opened 2020-05-28
Re-export types and values from other modules


lpil:
It would be nice to be able to re-export a type or function from another module.

Here's an idea for a syntax:

```rust
// one.gleam
pub type MyType {
  MyConstructor
}

pub fn my_function() -> Nil {
  Nil
}
```

```rust
// two.gleam

import two.{pub MyType, pub MyConstructor, pub my_function}
```

## Questions

Should it be permitted to export only a subset of a type's constructors? This would mean it would not possible for users to pattern match on all records as only some have been exported.

CrowdHailer:
I like the syntax, but only rexporting some constructors would be very awkward. (i'm sure I was discussing this somewhere previously but canot find it)

I'm keen that import `option.{Option}` should import Some and None, but I understand that it would mean you couldn't just search in a file to see where the Some and None constructors where defined.

lpil:
Let's use the syntax above as there's no other suggestions or major complaints.

lpil:
> Should it be permitted to export only a subset of a type's constructors? This would mean it would not possible for users to pattern match on all records as only some have been exported. The above syntax would allow that.

This makes me think of @hayleigh-dot-dev's suggestion to use a nested syntax for importing record constructors. I previously dismissed it but there's a second reason to have it here.

lpil:
Adding to v1 as this may require syntax changes

pvdrz:
I'd favor having the `pub` as a qualifier for `import` instead as it would be easier to read which imports are public and which ones are not:
```gleam
import two.{pub MyType, my_function, pub MyConstructor}
```
vs
```gleam
pub import two.{MyType, MyConstructor}
import two.my_function
```

inoas:
I'd also prefer to have separation between pub and non-pub imports as @pvdrz shows.

Reasoning, it may be a bit more verbose but:

- it far far better to read and 
- the formatter could group into two parts, alphabetically sorted pub imports and alphabetically sorted non-pub imports.

timjs:
Fully agree with @pvdrz and @inoas here for reasons they already stated!

inoas:
another question I'd have is: if you re-export a function (or type) but also define it in the same file, what happens?
imho compiler error or newly defined is being exported > reexported is being exported

lpil:
Considering `pub import a.{b}` and `import a.{pub b}` syntaxes.

It seems that `pub import a.{b}` implies you can import `a` from this module in some fashion.

[...]

PR #4545, opened 2025-05-01
Warn if a top-level definition shadows an imported one

### LINKED ISSUE #4519: Warn when unqualified import is shadowed by top-level definition.
Currently, when an unqualified import like `import wibble.{wobble}` is shadowed by a top-level definition in the module (`const wobble`, `fn wobble(...)`) the warning generated on the import says it is unused:

> Unused imported value
> This imported value is never used.

Technically this is true, but a clearer warning would notify the developer that the imported value can _never_ be used because it is shadowed by a top-level definition:

> Unusable imported value
> This imported value can never be used as it is shadowed by the `wobble` function.

This is a clearer message as the developer may look at their source code and be confused to see `wobble` is indeed used even though the compiler is warning them otherwise.

```
// The compiler will warn that `wobble` here is unused...
import wibble.{wobble}

fn something_else(...) {
  // ...but a developer might be confused to see that wobble
  // is used here.
  let x = wobble(...)
  //

  ...
}

// The actual wobble being used is this one
fn wobble(...) {
  ...
}
```

arielherself:
This PR hopefully resolves #4519. A new warning will be shown like this:

```
warning: Definition of "foo" shadows an imported value
  ┌─ /src/warning/wrn.gleam:3:4
  │
3 │ fn foo(x) {
  │    ^^^

The imported value could not be used in functions anyway.
```

However, I found a similar issue for constructors and type aliases. Especially if we import a constructor and define a type (not constructor) with the same name, the imported item will be used:

```gleam
import module1.{Bar}  // imported Bar is a constructor

type Bar {  // local Bar is a type
}

pub fn main() {
	Bar  // this resolves to module1.Bar
}
```

Is this an intended behavior? If not I think we should add an error for this.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -247,40 +247,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             self.register_type_alias(t, &mut env);
         }
 
-        for c in &statements.constants {
-            // don't warn if the definition is a re-export.
-            if let Constant::<(), ()>::Var {
-                location: _,
-                module: _,
-                name,
-                constructor: _,
-                type_: _,
-            } = &*c.value
-            {
-                if name == &c.name {
-                    continue;
-                }
-            }
-
-            if env.unqualified_imported_names.contains_key(&c.name) {
-                self.problems
-                    .warning(Warning::TopLevelDefinitionShadowsImport {
-                        location: c.location,
-                        name: c.name.clone(),
-                    })
-            }
-        }
-
         for f in &statements.functions {
-            let (_name_location, name) = f.name.clone().expect("A module's function must be named");
-            if env.unqualified_imported_names.contains_key(&name) {
-                self.problems
-                    .warning(Warning::TopLevelDefinitionShadowsImport {
-                        location: f.location,
-                        name,
-                    });
-            }
-
             if let Err(error) = self.register_value_from_function(f, &mut env) {
                 return self.all_errors(error);
             }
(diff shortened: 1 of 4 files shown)


GearsDatapacks:
The latter is intended behaviour. Types and values live in different namespaces

arielherself:
I also updated the test result of this existing test:

```rust
#[test]
fn used_shadowed_imported_value() {
    assert_warning!(
        (
            "thepackage",
            "wibble",
            "
pub const wibble = 1
"
        ),
        "
import wibble.{wibble}

pub const wibble = wibble
"
    );
}
```

Since redefining values is not actually legal, there should be a warning on shadowing of `wibble` here even if we used the imported `wibble` explicitly. Correct me if I'm wrong!

lpil:
That one should not warn 🙏 Until https://github.com/gleam-lang/gleam/issues/614 is resolved it has valid uses.

arielherself:
Thanks for your clarification! I've added logic to not complain about direct re-exports like `const c = c`.

compiler-core/src/analyse.rs
```diff
@@ -247,7 +247,40 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
+
         for f in &statements.functions {
+            let (_name_location, name) = f.name.clone().expect("A module's function must be named");
+            if env.unqualified_imported_names.contains_key(&name) {
+                self.problems
+                    .warning(Warning::TopLevelDefinitionShadowsImport {
+                        location: f.location,
+                        name,
+                    });
+            }
```
lpil, line 282: Please move the code for this check into the existing definition analysis rather than iterating over the functions and constants again 🙏

lpil (commented):
Thank you!

I've actually realised I was wrong with my previous guidance, the programmer should in that case use the qualified syntax. Sorry, I was quite tired when I wrote it! Could you remove that exception please 🙏

arielherself:
Of course! :purple_heart:

lpil:
Apologies, this was completed in this other PR https://github.com/gleam-lang/gleam/pull/4672. I am not able to remember the current state of all the ongoing work and didn't realise it was duplicating this one

PR #4547, opened 2025-05-01
Fix LSP ranges

### LINKED ISSUE #3628: LSP ranges are incorrect when line contains multi-codepoint grapheme clusters

### LINKED ISSUE #4543: Language server crashes when highlighting ellipsis in VSCode
* Open this file in VSCode: https://git.ahlcode.fi/nicd/hanuri/src/commit/61b15507603b72804724e8849b7cd932659e3955/frontend/src/hnr_front/views/data.gleam
* Highlight one of the "Loading…" texts where there is an ellipsis character:
  Image
* Gleam LS crashes

> error: Fatal compiler bug!
> 
> This is a bug in the Gleam compiler, sorry!
> 
> […]
> 
> Panic: compiler-core/src/language_server/code_action.rs:5613
> 	invalid value range
> Gleam version: 1.10.0
> Operating system: macos

VSCode version: 1.99.3

I don't know how to pass `GLEAM_LOG` to the LS in VSCode, it doesn't seem to inherit it from my shell.

GearsDatapacks:
Fixes #3628 and fixes #4543 
This PR changes the `LineNumbers` struct so that it takes into account multi-byte characters and converts between UTF-8 byte indices, which is what Rust uses, and UTF-16 line/character indices, which is what LSP uses.
I thought a lot about the best way to approach this, and eventually found the solution which is used in this PR. My reasoning is explained in the doc comment.
Let me know if you think I should use a different method to convert between them.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -403,3 +403,7 @@
 - Fixed a bug where some code actions would produce invalid code when a file
   contained characters that were represented with more than one byte in UTF-8.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where LSP ranges would be incorrect when a file contained characters
+  that were represented wit more than one byte in UTF-8.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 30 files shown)


lpil (approved):
Thank you! This is such a good fix!

PR #4549, opened 2025-05-02
Support UTF-16 and UTF-32 strings/codepoints on JavaScript

### LINKED ISSUE #4530: Compiler crash when using non-UTF-8 string patterns
For example, with the following code:
```gleam
pub fn main() {
  let assert <<"Hello":utf16, _rest:bits>> = <<"Hello, world!":utf16>>
}
```
If the command `gleam check` is run, the following error is produced:
```
error: Fatal compiler bug!

Panic: compiler-core/src/exhaustiveness.rs:2912
        non utf8 string in bit array pattern on js target
Gleam version: 1.10.0
Operating system: linux

variable set.
```

It seems the exhaustiveness checker panics when utf16 or utf32 strings are encountered, even if it is not the javascript target. Even if it were, it should emit a compiler error, not panic.

I'm working on supporting utf16 and utf32 segments on the JS target soon, so that should fix this. Just thought I'd report it now though so we don't forget.

GearsDatapacks:
Fixes #4530 
This PR has got a bit bigger than I originally wanted it to. Originally, this was meant to only affect the JavaScript target, however when implementing this I realised that there was a bug in the Erlang code generation too. I wasn't able to easily fix that bug before opening this PR, because it depended on some code that I wrote as part of this. I also couldn't wait and fix it after, because it would have caused several failing tests.
So I opted to fix it as part of this PR, hope that's OK.

This is one of the last main features missing from bit arrays on JS. Hopefully soon we can make JS bit arrays fully feature complete!


compiler-core/src/strings.rs
```diff
@@ -64,3 +66,57 @@ pub fn convert_string_escape_chars(str: &EcoString) -> EcoString {
...
+
+    for char in string.chars() {
+        length += char.len_utf16()
+    }
+
+    length
+}
+
+/// Gets the number of UTF-32 codepoints in a string (also known as the number of
+/// characters).
```
lpil, line 119: We deliberately avoid the word "characters" in Gleam because it's vague, but when we do use it we use it to mean graphemes as that's closer to how humans think.

compiler-core/src/javascript/decision.rs
```diff
@@ -1228,19 +1239,36 @@ impl<'generator, 'module, 'a> Variables<'generator, 'module, 'a> {
             " === "
         };
 
+        let escaped = convert_string_escape_chars(literal_string);
+        // We need to have this vector here so that we don't run into lifetime
+        // issues when calling `.as_slice` on the local vectors created when this
+        // isn't a UTF-8 string.
+        let mut _bytes_vec = Vec::new();
```
lpil, line 1246: Don't use an underscore prefix for used variables please 🙏

compiler-core/src/exhaustiveness.rs
```diff
@@ -2905,11 +2943,22 @@ fn segment_size(
...
+            }
+
+            ast::Pattern::String { value, .. } if segment.has_utf16_option() => {
+                ReadSize::ConstantBits(
+                    // Each utf16 code unit is 16 bits
+                    length_utf16(&convert_string_escape_chars(value)) * BigInt::from(16),
+                )
+            }
+            ast::Pattern::String { value, .. } if segment.has_utf32_option() => {
+                // Each utf32 codepoint is 32 bits
```
lpil, line 2958: How come we talk about code points here and code units in the previous branch?

compiler-core/src/exhaustiveness.rs
```diff
@@ -2875,7 +2908,12 @@ fn segment_matched_value(pattern: &TypedPattern) -> BitArrayMatchedValue {
 fn segment_size(
     segment: &TypedPatternBitArraySegment,
     pattern_variables: &HashMap<EcoString, ReadAction>,
+    // Override for the segment pattern, if we need to determine the size of an
+    // assignment segment.
```
lpil, line 2912: What's an override?

CHANGELOG.md
```diff
@@ -185,6 +185,10 @@
 - The compiler will not generate needless code for unused pattern variables.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- On the JavaScript target, bit arrays now support UTf-16 and UTF-32 string
+  segments.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
```
lpil, line 191: Could you add the Erlang bug fix to the changelog please 🙏

lpil (commented):
Fantastic, thank you!

What made you opt to use a guard for the Erlang ones? I'm worried that it might not optimise as well as using a pattern, but I've not tested it.

GearsDatapacks:
Ah you're right! I used a guard because that's what we do for all the other assignment patterns, but since we're reassigning anyway we can just use regular pattern matching.

compiler-core/src/exhaustiveness.rs
GearsDatapacks: I'm just a bit flakey with unicode terminology

CHANGELOG.md
GearsDatapacks: It doesn't affect Gleam code before 1.11 since assignment patterns will only been supported in 1.11

compiler-core/src/javascript/tests/case.rs
```diff
@@ -413,3 +413,59 @@ pub fn go(x) {
...
+    assert_js!(
+        r#"
+pub fn main() {
+  case <<"Hello":utf32-little>> {
+    <<"Hello" as m:utf32-little>> -> m
+    _ -> ""
+  }
+}
+"#
+    );
```
giacomocavalieri, line 470: Could we also have a test with an alternative pattern assigning a literal string with two different encodings?
```
<<"wibble" as m:utf16, _:bits>>
| <<"wobble" as m:utf32, 1, _:bits>> -> ...
```

CHANGELOG.md
lpil: oooh of course! Thank you

GearsDatapacks:
As we discussed on Discord, I have now made an error explaining which assignment patterns with `utf16` or `utf32` options are not currently supported.

PR #4550, opened 2025-05-02
avoid qualify code action on shadowed modules

### LINKED ISSUE #4548: "Qualify `x` as `wibble.x` code action does not respect shadowing.
Gleam encourages the reuse of meaningful variable names through shadowing, be that through `let` bindings, pattern matching, or lambda argument names. When triggering the code action to qualify a function with its module name, the transformation is applied to **any** variable that matches the unqualified name, even if it is an unrelated shadowing:

```gleam
import wibble.{x}

fn example() {
  echo x
       ^ place cursor here

  let x = 1

  echo x

  let cb = fn(x) {
    echo x
  }

  todo
}
```

Triggering the code action at the cursor position indicator above results in the following module:

```gleam
import wibble

fn example() {
  echo wibble.x

  let x = 1

  echo wibble.x

  let cb = fn(x) {
    echo wibble.x
  }

  todo
}
```

I would **expect** the module to look like this:

```gleam
import wibble.{x}

fn example() {
  echo wibble.x

  let x = 1

  echo x

  let cb = fn(x) {
    echo x
  }

  todo
}
```

It is important to note that placing my cursor on any of the shadowed `x` uses will **not let me trigger the code action** because the LSP is aware that these are not the same `x`. 

I ran into this refactoring some tests and ended up undoing the action and manually updating things as the amount of incorrect renames was much higher.

scristobal:
resolves #4548

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -3582,6 +3582,37 @@ pub fn main() {
             .select_until(find_position_of("(1)")),
     );
 }
+
+#[test]
+fn test_unqualified_to_qualified_import_variable_shadowing() {
+    let src = r#"
+
+import wibble.{wobble}
+
+pub fn example() {
+  echo wobble
+
+  let wobble = 1
+
+  echo wobble
+
+  let _ = fn(wobble) {
+    echo wobble
+  }
+
+  todo
+}
+"#;
+
+    assert_code_action!(
+        "Qualify wobble as wibble.wobble",
+        TestProject::for_source(src).add_hex_module("wibble", "pub fn wobble() { todo }"),
+        find_position_of("wob")
+            .nth_occurrence(2)
+            .select_until(find_position_of("ble").nth_occurrence(3))
+    );
+}
+
 /* TODO: implement qualified unused location
 #[test]
 fn test_remove_unused_qualified_action() {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -411,3 +411,8 @@
 - Only print the user-friendly LSP message when stdin is a terminal to avoid
   emitting redundant logs.
   ([Ariel Parker](https://github.com/arielherself))
+
+- Fixed a bug where the language server would wrongly override local variables
+  if they were shadowing an unqualified module function.
+  ([Samuel Cristobal](https://github.com/scristobal))
+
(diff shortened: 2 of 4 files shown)


lpil (commented):
Very clean, thank you! Could you update the changelog also please 🙏

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__unqualified_to_qualified_import_valiable_shadowing.snap
```diff
@@ -0,0 +1,41 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "\nimport wibble.{foo}\n\npub fn example() {\n  echo foo\n\n  let foo = 1\n\n  echo foo\n\n  let cb = fn(foo) {\n    echo foo\n  }\n\n  todo\n}\n"
+---
+----- BEFORE ACTION
+
+import wibble.{foo}
+   ↑
```
scristobal, line 8: btw. the arrow should not be pointing here, I guess, might also be a bug?

GearsDatapacks: It's because you're selecting the range from the `f` in `foo` to the `o` in `import`, which is a backwards range, so it's not quite sure what to do

compiler-core/src/language_server/tests/action.rs
```diff
@@ -3582,6 +3582,36 @@ pub fn main() {
             .select_until(find_position_of("(1)")),
     );
 }
+
+#[test]
+fn test_unqualified_to_qualified_import_valiable_shadowing() {
```
GearsDatapacks, line 3587: Small typo: `valiable`

compiler-core/src/language_server/tests/action.rs
```diff
@@ -3582,6 +3582,38 @@ pub fn main() {
...
 }
+
+#[test]
+fn test_unqualified_to_qualified_import_valiable_shadowing(){
+
+    let src = r#"
+import wibble.{foo}
+
+pub fn example() {
+  echo foo
```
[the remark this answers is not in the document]
scristobal: Right away! Thank you very much for that piece of knowledge. TIL, never more.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__unqualified_to_qualified_import_valiable_shadowing.snap
scristobal: oook, got it! I was actually confused about how selectors work. Now it is clear, thanks!

lpil (approved):
Thank you!!

PR #4555, opened 2025-05-04
Allow specifying a tag prefix for documentation

### LINKED ISSUE #3858: `path` option in `repository` can mess with git tags
When adding a path in repository in `gleam.toml`, when there's multiple packages in the repo, their versions can go out of sync.

Let's take an example:
- There's a package `foo`, located in `repo/foo`, in version 1.0.0.
- There's a package `bar`, located in `repo/bar` in version 1.1.0. 
- When you generate documentation, all generated paths expect to find a tag in the git repo, and the path refers the tag.  
  For `foo`, it will be `https://.../v1.0.0/foo/src/module.gleam`. For `bar`, it will be `https://.../v1.1.0/bar/src/module.gleam`.
- Tags in your git repo can be not synchronized, because sometimes you just cannot properly tag the repo for every version of every packages.

A solution would be to use the same version number for every packages. Every time a package version bumps, bump the other packages to the same version, and republish everything. That way, all tags will always be true. 
However, it's more work for the maintainer. 

One solution could be to point to a commit in the repo, instead of a tag.

---

[The problem can be seen in the Sketch documentation at the moment of writing.](https://hexdocs.pm/sketch)

sbergen:
Resolves #3858

I originally left the `v` in there, but looking at it, thought it would be nice to be able to override that also (see second commit). Now I'm not sure if we should make the tag prefix non-`Optional`, and use "v" as the default. Happy to tweak it some more if we want to change it.

### DIFF
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -152,7 +152,7 @@ pub struct PackageConfig {
     #[serde(default, rename = "dev-dependencies", serialize_with = "ordered_map")]
     pub dev_dependencies: Dependencies,
     #[serde(default)]
-    pub repository: Repository,
+    pub repository: Option<Repository>,
     #[serde(default)]
     pub links: Vec<Link>,
     #[serde(default)]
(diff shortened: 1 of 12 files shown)


compiler-core/src/config.rs
```diff
@@ -836,6 +842,24 @@ impl Repository {
...
+            | Repository::BitBucket { tag_prefix, .. }
+            | Repository::Codeberg { tag_prefix, .. }
+            | Repository::SourceHut { tag_prefix, .. }
+            | Repository::Gitea { tag_prefix, .. } => tag_prefix.as_ref(),
+
+            Repository::Custom { .. } | Repository::None => None,
+        };
+
+        match prefix {
+            Some(prefix) => format!("{prefix}{version}"),
```
lpil: Should we add a `-` between them?

lpil (commented):
Hey! Thank you for this. I've a design questions which I'm interested in your thoughts on:

It's possible with this config to specify a subdir but not specify a tag prefix, so you could configure it wrong and get collisions in a monorepo. It's also possible to have multiple packages using the same tag prefix. Could we improve this?

Could make the prefix mandatory if the path is set?

Or could use the package name as the prefix if the path is set? Then we would know it's unique.

compiler-core/src/config.rs
sbergen: It actually seems like a lot of people like to use `/` to put the tags in a directory. My thought was originally to give the freedom to the user to do what they like. However, if we'd use the package name automatically, then I think some delimiter would be good, and `/` would be a good candidate.

sbergen:
> Could make the prefix mandatory if the path is set?
> 
> Or could use the package name as the prefix if the path is set? Then we would know it's unique.

I like the idea of doing one of these, but I don't have a clear preference for one over the other.

PR #4557, opened 2025-05-04
Collapse `else if` in decision tree generation

### LINKED ISSUE #4522: Improve decision tree generated code for `if-else` branches
Generating code from the decision tree can result in this kind of structure on the JS target:
```js
if (cond) {
  // ...
} else {
  if (other_cond) {
    // ...
  } else {
    // ...
  }
}
```
It would be a nice touch if the generated code looked like this instead:
```js
if (cond) {
  // ...
} else if (other_cond) {
  // ...
} else {
  // ...
}
```

This is not as straightforward as it seems since sometimes we might have variable definitions in between an `else` and the nested `if`, making it impossible to collapse into an `else if`.

GearsDatapacks:
A first step towards #4522
Since decision tree generation has not yet been released, I didn't add this to the changelog


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__case_bit_array_assignment_discard.snap
```diff
@@ -18,16 +18,14 @@ export function go(x) {
...
-  } else {
-    if (x.bitSize >= 8) {
-      if ((x.bitSize - 8) % 8 === 0) {
-        let n = x.byteAt(0);
-        return n;
-      } else {
-        return 1;
-      }
+  } else if (x.bitSize >= 8) {
+    if ((x.bitSize - 8) % 8 === 0) {
```
lpil, line 22: With this approach is it possible to extend it to collapse this `else if` with the `if` on the line below?

Currently we've moved the final `else` down one indentation level rather than removing it entirely, so the difference to the code size isn't as high as we would like.

lpil (commented):
Thank you! I've left a question about future expansions to this approach

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__case_bit_array_assignment_discard.snap
GearsDatapacks: Yeah that should be possible!

lpil: Cool! I mostly am looking to make sure that we don't pick something that makes it hard to add in future.

I don't think this quite closes the issue because there's still more `else` than we would like. Shall I merge this and we leave it open or would you like to take that on in this branch also?

GearsDatapacks: I can implement it in this branch, yeah

GearsDatapacks:
I haven't had much time to work on this recently. As we're planning to do a release soon, maybe it's better to merge this now in its current state, then I'll have more time to do the rest of the work after the release. Does that sound OK to you?

lpil:
Sounds good

PR #4562, opened 2025-05-05
Bug Fix: store typescript_declarations config and rebuild if changed (Attempt 2)

daniellionel01:
New branch for bugfix related to https://github.com/gleam-lang/gleam/pull/4523

Had a merge commit in my history but I am too scared of messing up my branch with rebasing, because I have no experience with that action.

File changes are identical though

FYI @lpil


compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +265,48 @@ where
...
+
+            if ts_enabled {
+                self.io
+                    .write(&ts_indicator_path, "")
+                    .map_err(|e| Error::FileIo {
+                        action: FileIoAction::WriteTo,
+                        kind: FileKind::File,
+                        path: ts_indicator_path,
+                        err: Some(e.to_string()),
+                    })?;
```
lpil: `self.io.write` already returns the right error type so can return it as-is without wrapping it in another layer.

compiler-core/src/build/project_compiler.rs
```diff
@@ -261,6 +265,48 @@ where
...
+    pub fn rebuild_if_javascript_config_changed(&self) -> Result<(), Error> {
+        if !self.target().is_javascript() || !self.options.codegen.should_codegen(true) {
+            return Ok(());
+        }
+
+        let build_path = self
+            .paths
+            .build_directory_for_target(self.mode(), self.target());
+
+        let ts_indicator_path = build_path.join("typescript-declarations.lock");
```
lpil: This isn't a lock file so let's remove the .lock extension.

lpil (commented):
Super nice, thank you! Couple small notes inline.

compiler-core/src/build/project_compiler.rs
daniellionel01: Done!

compiler-core/src/build/project_compiler.rs
daniellionel01: Done!

daniellionel01:
@lpil Addressed both comments. Thank you!

lpil:
I'm gunna assume this is ready for review!

daniellionel01:
> I'm gunna assume this is ready for review!

yessir!

lpil (approved):
I've added tests and support the new sourcemaps config too, seeing as it had the same problem.

Thank you!!!

daniellionel01:
> I've added tests and support the new sourcemaps config too, seeing as it had the same problem.
> 
> Thank you!!!

oh yeah good catch! you're welcome 🌚

PR #4563, opened 2025-05-05
Bugfix: path dependencies module resolution bug with local paths

### LINKED ISSUE #2278: Path dependencies module resolution bug

daniellionel01:
Code to fix https://github.com/gleam-lang/gleam/issues/2278

I hope this is an acceptable solution, I've inspected closed PRs that didn't make it attempting to fix this issue (https://github.com/gleam-lang/gleam/pull/2334 and https://github.com/gleam-lang/gleam/pull/3398).

In summary this approach stores a hash of the manifest.toml of all dependencies that are a path dependencies in the build directory and rebuilds any packages if the manifest.toml should change.

Let me know if I've missed anything or if this approach does not scale well, thank you!

Also I did not want to clutter the codebase with it, but here is a script (thank you GPT) to test the scenario that produces the error mentioned in the linked issue. It assumes you're running it in the root of the gleam codebase, otherwise you can adjust the `GLEAM_BIN` path. Also this script probably only works on a mac or linux. Here is the script:

```bash
#!/bin/bash
set -e

# Use the locally built gleam executable
GLEAM_BIN="$PWD/target/debug/gleam"

# Ensure the local build exists
if [ ! -f "$GLEAM_BIN" ]; then
    echo "Error: Local Gleam build not found at $GLEAM_BIN"
    echo "Please run 'cargo build' first"
    exit 1
fi

echo "Using Gleam binary at: $GLEAM_BIN"

# Create a temporary directory for the test
TEST_DIR=$(mktemp -d)
echo "Creating test environment in $TEST_DIR"

# Create bar (dependency) project
echo "Creating bar project (the dependency)"
cd $TEST_DIR
mkdir -p bar
cd bar

# Initialize bar project
cat > gleam.toml << EOF
name = "bar"
version = "0.1.0"

[dependencies]
gleam_stdlib = "~> 0.29"
EOF

# Create bar module
mkdir -p src
cat > src/bar.gleam << EOF
import gleam/io

pub fn say_hello() {
  io.println("Hello from bar!")
}
EOF

# Create foo (root) project
echo "Creating foo project (the root project)"
cd $TEST_DIR
mkdir -p foo
cd foo

# Initialize foo project with bar as path dependency
cat > gleam.toml << EOF
name = "foo"
version = "0.1.0"

[dependencies]
bar = { path = "../bar" }
EOF

# Create foo module that imports bar
mkdir -p src
cat > src/foo.gleam << EOF
import bar

pub fn main() {
  bar.say_hello()
}
EOF

# First build of foo
echo "Building foo for the first time..."
cd $TEST_DIR/foo
$GLEAM_BIN build

# Add simplifile dependency to bar
echo "Adding simplifile dependency to bar..."
cd $TEST_DIR/bar
$GLEAM_BIN add simplifile

# Update bar module to use simplifile
cat > src/bar.gleam << EOF
import simplifile
import gleam/io

pub fn say_hello() {
  io.println("Hello from bar!")
}
EOF

# Build foo again - this should now succeed with our fix
echo "Building foo again after adding simplifile to bar..."
cd $TEST_DIR/foo
$GLEAM_BIN build

if [ $? -eq 0 ]; then
  echo -e "\n\033[0;32mSUCCESS: The fix works! Foo successfully built after bar added simplifile dependency.\033[0m"
else
  echo -e "\n\033[0;31mFAILURE: The fix didn't work. Foo failed to build after bar added simplifile dependency.\033[0m"
  exit 1
fi

# Clean up
echo -e "\nTest directory: $TEST_DIR"
echo "You can safely remove this directory when done with: rm -rf $TEST_DIR"
echo "Or inspect it to see the project structure."
```


compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
         }
     }
 
+    // For path dependencies, we need to check if their manifest.toml files have changed,
+    // which indicates that their dependencies have changed
+    for (key, requirement2) in requirements2 {
+        if let Requirement::Path { path } = requirement2 {
+            let dep_manifest_path = root_path.join(path).join("manifest.toml");
```
lpil: All path manipulation must be done in the ProjectPaths class, no other code is permitted to use `.join` etc.

compiler-cli/src/dependencies.rs
```diff
@@ -888,6 +948,11 @@ fn provide_local_package(
     } else {
         fs::canonicalise(&parent_path.join(package_path))?
     };
+
+    // We used to always force clean artifacts for path dependencies,
+    // but now we use the gleam.toml hash tracking in is_same_requirements
+    // to determine when path dependency dependencies have changed
```
lpil: What does this comment mean?

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
         }
     }
 
+    // For path dependencies, we need to check if their manifest.toml files have changed,
+    // which indicates that their dependencies have changed
+    for (key, requirement2) in requirements2 {
```
lpil: This equality checking logic should be in the `same_requirements` function as that's the one that checks equality, while this one iterates over the collection.

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
...
+
+            let manifest_content = match fs::read(&dep_manifest_path) {
+                Ok(content) => content,
+                Err(_) => {
+                    tracing::debug!("cannot_read_path_dependency_manifest_forcing_rebuild");
+                    return Ok(false);
+                }
+            };
+
+            let mut hasher = std::hash::DefaultHasher::new();
```
lpil: We use `xxhash_rust::xxh3::xxh3_64` for fingerprinting, not the default hasher.

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
...
+                return Ok(false);
+            }
+
+            let manifest_content = match fs::read(&dep_manifest_path) {
+                Ok(content) => content,
+                Err(_) => {
+                    tracing::debug!("cannot_read_path_dependency_manifest_forcing_rebuild");
+                    return Ok(false);
+                }
+            };
```
lpil: We check mtimes before we hash, to avoid extra work, so let's add that here.

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
...
+
+            let mut hasher = std::hash::DefaultHasher::new();
+            std::hash::Hash::hash(&manifest_content, &mut hasher);
+            let current_hash = std::hash::Hasher::finish(&hasher).to_string();
+
+            // If cached hash file doesn't exist, this is the first time we're checking this dependency
+            if !cached_hash_path.exists() {
+                // Save the current hash for future comparisons
+                if let Err(e) = fs::write(&cached_hash_path, &current_hash) {
```
lpil: The errors in this code are being silently discarded! They must always be passed back up.

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
...
+            // Check if we have a cached hash for this path dependency's manifest.toml
+            // Use the dependency name instead of the path for the manifest hash file
+            let cached_hash_path = root_path
+                .join("build")
+                .join("packages")
+                .join(format!("{}.manifest_hash", key));
+
+            if !dep_manifest_path.exists() {
+                tracing::debug!("path_dependency_manifest_not_found_forcing_rebuild");
+                return Ok(false);
```
lpil: This would result in rebuilding every single time when there's no manifest for the path dep!

compiler-cli/src/dependencies.rs
```diff
@@ -672,6 +672,66 @@ fn is_same_requirements(
         }
     }
 
+    // For path dependencies, we need to check if their manifest.toml files have changed,
```
lpil: The name for this function doesn't fit what it does any more- even if the requirements are the same the function can return false in the instance that the definition of one of the required deps has changed.

The functions have also been made effectful and to depend on file system state, so that would need to be communicated clearly in the name and arguments too.

These functions that check the requirements are the same could go unchanged (which would still fit the "is same requirements" name) and a new function could be created to check just the path deps have not changed their deps.

lpil (commented):
Sorry this took so long for me to review! I missed it somehow, and I've only just seen it now.

Thank you for this, I've left some comments inline. Tests will need to be added too.

compiler-cli/src/dependencies.rs
daniellionel01: caught my GPT generated snippet. Sorry, removing it.

compiler-cli/src/dependencies.rs
daniellionel01: yeah, I removed the comment

compiler-cli/src/dependencies.rs
daniellionel01: Done!

compiler-cli/src/dependencies.rs
daniellionel01: Done!

compiler-cli/src/dependencies.rs
daniellionel01: Done!

compiler-cli/src/dependencies.rs
daniellionel01: Sorry! `continue`-ing instead now

compiler-cli/src/dependencies.rs
daniellionel01: Done!

compiler-cli/src/dependencies.rs
daniellionel01: pushed an update!

daniellionel01:
@lpil Addressed all comments!

Also added tests, that you can checkout.

Thank you for the input!

giacomocavalieri:
Hello! If this is ready please remember to un-draft it, also would you mind rebasing on main? It seems there's some conflicts

daniellionel01:
> Hello! If this is ready please remember to un-draft it, also would you mind rebasing on main? It seems there's some conflicts

i'll get on rebasing!

daniellionel01:
@lpil @giacomocavalieri ready to merge from my end!

giacomocavalieri (approved):

lpil (approved):
Thank you!!!

PR #4564, opened 2025-05-05
Add Generate Variant action

giacomocavalieri:
This PR adds the "Generate variant" code action to automatically add a variant to a type starting from incorrect code. This works correctly with types defined in the same module, or other modules. The code action can be triggered on expressions:

```gleam
pub type Msg {
  ServerSentResponse(Json)
}

pub fn view() -> Element(Msg) {
  div([], [
    button([on_click(UserPressedButton)], [])
  //                 ^^^^^^^ [Generate variant]
  ])
}

// Generates:
pub type Msg {
  ServerSentResponse(Json)
  UserPressedButton
}
```

and on patterns:

```gleam
pub type Pet {
  Dog(name: String)
}

pub fn greet(pet: Pet) {
  case pet {
    Dog(name:) -> name <> " good boy"
    Cat(name:) -> "psps " <> name
//  ^^^^^ [Generate variant]
  }
}

// Will generate
pub type Pet {
  Dog(name: String)
  Cat(name: String)
}
```

In order to make this work for patterns I've improved on fault tolerance, meaning that now when a variant pattern is incorrect (it doesn't correspond to any constructor in scope), the compiler doesn't just give up but still tries to infer all the argument patterns

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -189,6 +189,11 @@
   continue to occur if a function references invalid type in its signature.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The compiler is now fault tolerant when analysing patterns for custom types,
+  meaning it won't stop at the first error it encounters (for example if a
+  constructor is wrong).
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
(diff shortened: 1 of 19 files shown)


lpil (approved):
Thank you!!!

PR #4567, opened 2025-05-07
Refuse to publish packages that import dev dependencies

### LINKED ISSUE #3143: Compiler should refuse to publish a package that uses a dev-dependency in its source
I mistakenly published a package including modules that import dev-dependencies, as I didn't know I had to put them under the `/test` directory.

During publishing I got these warnings

```sh

warning: Transitive dependency imported
  ┌─ /Users/jonas/Projects/chrobot/src/chrobot/internal/download_protocol.gleam:9:1
  │
9 │ import gleam/http/request
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^

The module `gleam/http/request` is being imported, but `gleam_http`, the
package it belongs to, is not a direct dependency of your package.
In a future version of Gleam this may become a compile error.

Run this command to add it to your dependencies:

    gleam add gleam_http


warning: Transitive dependency imported
   ┌─ /Users/jonas/Projects/chrobot/src/chrobot/internal/download_protocol.gleam:10:1
   │
10 │ import gleam/httpc
   │ ^^^^^^^^^^^^^^^^^^

The module `gleam/httpc` is being imported, but `gleam_httpc`, the package
it belongs to, is not a direct dependency of your package.
In a future version of Gleam this may become a compile error.

Run this command to add it to your dependencies:

    gleam add gleam_httpc
```

But the package was published regardless.

According to @hayleigh-dot-dev on discord:

> can you open an issue on the compiler repo. i think the compiler should refuse to publish a package that uses a dev-dependency in its source!

gideongrinberg:
This PR fixes #3143 by throwing an error when gleam publish is invoked and a dev dependency is imported in the code.

The check is implemented in `do_build_hex_tarball`. Another approach would be to implement the check in the analysis phase, but, as far as I can tell, there's no way to detect if the compilation has been invoked by `gleam publish` or something else.

I manually tested this and it works as expected, but I can add tests as well.


gideongrinberg:
CI passes for me: https://github.com/gideongrinberg/gleam/actions/runs/14873076625

compiler-core/src/error.rs
```diff
@@ -1083,6 +1089,17 @@ Please make sure internal types do not appear in public functions and try again.
                 location: None,
             }],
 
+            Error::CannotPublishLeakedDevDependency { module, package } => {
+                let text = &format!("The module `{module}` is being imported from `{package}`, but package is not a direct dependency of your project.\nRun this command to add it to your dependencies:\n\tgleam add {package}");
+                vec![Diagnostic {
+                    title: "Developer dependency imported".into(),
```
inoas, line 1095: `Development dependency` is "more" gender neutral (in non English contexts)

gideongrinberg: Fixed!

gideongrinberg:
@inoas are there additional changes you think I should make?

gideongrinberg:
All tests pass on my fork: https://github.com/gideongrinberg/gleam/actions/runs/14948892106

lpil:
Hello! Sorry for taking so long to review this. I've got a bit caught up in this current release. I'll get to this once v1.11 is out!

gideongrinberg:
Thanks, not a problem. I see that the merge commit (I accidentially synced my branch with main) is causing the CI failure. Should I revert the merge?

GearsDatapacks:
You'll need to rebase instead, and remove the merge commit before this is merged.

compiler-cli/src/publish.rs
```diff
@@ -338,8 +338,21 @@ fn do_build_hex_tarball(paths: &ProjectPaths, config: &mut PackageConfig) -> Res
     // refuse to publish as the package is not yet finished.
     let mut modules_containing_todo = vec![];
     let mut modules_containing_echo = vec![];
+    let dev_dependencies: Vec<&EcoString> = config.dev_dependencies.keys().collect();
 
     for module in built.root_package.modules.iter() {
+        for (imported_module, _) in &module.dependencies {
+            if let Some(import_info) = built.module_interfaces.get(imported_module) {
+                let package = &import_info.package;
+                if dev_dependencies.contains(&package) && module.origin == Origin::Src {
```
lpil, line 347: `dev_dependencies.contains(&package)` isn't enough to determine if it's a dev dep or not because it could be a transient dev dependency, which would not be in this list.

If we are going to use this approach we'll need to before this reject the package if it using transient deps.

compiler-cli/src/publish.rs
```diff
@@ -338,8 +338,21 @@ fn do_build_hex_tarball(paths: &ProjectPaths, config: &mut PackageConfig) -> Res
     // refuse to publish as the package is not yet finished.
     let mut modules_containing_todo = vec![];
     let mut modules_containing_echo = vec![];
+    let dev_dependencies: Vec<&EcoString> = config.dev_dependencies.keys().collect();
 
     for module in built.root_package.modules.iter() {
+        for (imported_module, _) in &module.dependencies {
```
lpil, line 344: This is cool but I think we could improve it!

`src/` code shouldn't ever really use dev dependencies, so we should make it emit a warning when this happens (as an error would be a breaking change).

In the checking for this during we code analysis can record on modules if they use dev deps or transient deps or not, and then in `gleam publish` and `gleam export erlang-shipment` we can return an error by checking this flag on the module.

gideongrinberg:
Thank you for the review. I am away for a few weeks but will finish this as soon as I can.

lpil:

lpil:

PR #4571, opened 2025-05-08
Remove some dependencies

GearsDatapacks:
Louis mentioned recently that if possible we would be removing dependencies from the compiler, so this PR does that:
- Removed `heck` dependency and reimplemented the two methods that were used. The reimplemented versions are simplified but all we need for the compiler.
- Removed `strsim` dependency as it seems to no longer be used anywhere
- Removed `bytes` dependency as it seemed to no longer be used anywhere
- Moved `walkdir` dependency from the root workspace to the `test-helpers-rs` project, as that is the only place it is used. This removes it from the production build and only requires it in a test build.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1183,7 +1183,6 @@ dependencies = [
  "serde_derive",
  "serde_json",
  "spdx",
- "strsim",
  "strum",
  "tar",
  "termcolor",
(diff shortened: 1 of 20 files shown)


lpil (approved):
Gears, you're the best.

Did have some clever way to find the unused ones?

GearsDatapacks:
`grep use <name>` is all I did

lpil:
Nice

PR #4573, opened 2025-05-08
Make function parameters fault tolerant

### LINKED ISSUE #4092: "Import" code action sometimes fails and deletes a range of code
### Description
Occasionally an "Import" action is suggested for the type annotation on function arguments that, when applied, _sometimes_ causes:
1. A range of code to be deleted near the cursor
2. No import to be added

### Reproduction Steps
I've experienced a bit of non-determinism with the issue that makes it difficult to give exact steps to reproduce. I am able to get the issue to occur somewhat frequently by repeatedly doing variations of the following:
1. Begin writing a top-level method with a type annotation that uses a missing import e.g. 
    ```gleam
    fn foo(set: set.Set(String)) {
      todo
    }
    ```
2. Return the cursor to the module name for the import and apply the suggested "Import action"
3. Notice that some code is deleted
### Environment:
- **Editor** VSCode
- **Gleam VSCode Extension:** 2.10.0
- **OS:** Ubuntu on WSL 2
- **Gleam:** 1.6.3
- **Gleam Runtime:** Erlang OTP 25

I am not sure where to look for logs that would be useful here, I only see `INFO compile=...` in the `Gleam Language Server` output in VSCode.

### Screenshots / Recording
| Suggested action| Applied action|
|--------|--------|
| image | image | 

Recording: https://github.com/user-attachments/assets/885af89e-9dd5-402f-830d-c399e51590ba

GearsDatapacks:
This PR seems to fix #4092.
This PR makes function parameters fault tolerant, meaning that if an error occurs (e.g. an invalid type is referenced), type inference can continue. As far as I can tell, this is what was causing the linked issue.
As detailed in the above issue, it's not easy to reproduce, especially using automated tests. I'm not sure if there's a good way to test it. For now I've added a test in `type_::tests` to test this fault tolerance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -185,6 +185,10 @@
 - The compiler will not generate needless code for unused pattern variables.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Function parameters are now fault tolerant, meaning that type-checking can
+  continue to occur if a function references invalid type in its signature.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Build tool
 
 - The build tool now supports placing modules in a directory called `dev`,
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -248,9 +248,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
         }
 
         for f in &statements.functions {
-            if let Err(error) = self.register_value_from_function(f, &mut env) {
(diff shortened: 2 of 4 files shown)


giacomocavalieri:
This is great!! Remember to update the Changelog

GearsDatapacks:
Oh yeah, thanks!

lpil (approved):
Soooooooo nice. Thank you!

PR #4575, opened 2025-05-09
Remove remaining unused deps

dvic:


compiler-core/Cargo.toml
```diff
@@ -73,8 +73,6 @@ tracing.workspace = true
 [build-dependencies]
 # Data (de)serialisation
 serde_derive = "1"
-# Cap'n Proto binary format codegen
-capnpc = "0"
```
GearsDatapacks, line 77: This is still used in `compiler-core/build.rs`, but the code is commented out. It is still a required dependency

compiler-wasm/Cargo.toml
```diff
@@ -12,9 +12,7 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 gleam-core = { path = "../compiler-core" }
 console_error_panic_hook = "0"
-serde-wasm-bindgen = "0"
 wasm-bindgen = { version = "=0.2.99", features = ["serde-serialize"] }
-tracing-wasm = "*"
```
GearsDatapacks, line 17: This is still used in the wasm version of the compiler.

lpil (commented):
I think these may all still be in use

compiler-wasm/Cargo.toml
dvic: Ah I did double check but I ran cargo test... 🤦‍♂️

compiler-core/Cargo.toml
dvic: Clear

dvic:
> I think these may all still be in use

Yes, I see now, I was a bit too quick with the checks (cargo test instead of `wasm-pack test --node compiler-wasm`)

Well at least this PR confirms there aren't any unused deps left ;)

PR #4576, opened 2025-05-09
Error for conflicting Gleam and Erlang files

### LINKED ISSUE #2738: Emit an error when external modules and Gleam modules would collide
To replicate:
- create a gleam program

- create a mjs file with the same name (excluding the .mjs/.gleam extension) as one of your gleam files
OR
- create an erlang file with the same name as one of your gleam files

- attempt to import an external gleam function from one of the external files with the same name

Problem:
Gleam will not recognize any exports from the .mjs or .erl and throw the following error:
SyntaxError: The requested module '(mjs/erlang file path)' does not provide an export named '(imported external function name)'
It would be much more clear if the error was specific to having a mjs or erlang file with the same name as a gleam file importing from it, instead of stating that the module does not have the specified export

GearsDatapacks:
Closes #2738
Errors for conflicting Gleam and JavaScript modules have already been implemented, so this does the same for Erlang and Gleam.

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -63,7 +63,7 @@ pub struct PackageCompiler<'a, IO> {
     pub compile_beam_bytecode: bool,
     pub subprocess_stdio: Stdio,
     pub target_support: TargetSupport,
-    pub cached_warnings: CachedWarnings,
+    pub package_kind: PackageKind,
 }
 
 impl<'a, IO> PackageCompiler<'a, IO>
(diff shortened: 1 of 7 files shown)


compiler-core/src/build/package_compiler.rs
```diff
@@ -63,7 +63,7 @@ pub struct PackageCompiler<'a, IO> {
     pub compile_beam_bytecode: bool,
     pub subprocess_stdio: Stdio,
     pub target_support: TargetSupport,
-    pub cached_warnings: CachedWarnings,
+    pub package_kind: PackageKind,
```
lpil, line 66: The package compiler is a lower level API, not coupled to the project compiler, and is used in other contexts (wasm compiler, `gleam compiler-package`), so it cannot know about a ProjectCompiler concept such as the project compiler package kind. The behaviour needs to be specified specifically through flags such as CachedWarnings.

lpil (commented):
Thank you! I've left a note about the configuration design as some project compiler level information has got into the package compiler here

PR #4577, opened 2025-05-09
Fix stack overflow crash on Windows

### LINKED ISSUE #4287: Stack overflow crash when chaining multiple decode.field calls
Gleam: 1.8.1
OS: Windows 11
Target: Both

### Issue description

When chaining approximately 50 `use x <- decode.field`, `gleam run` crashes with the following error : 

> thread 'main' has overflowed its stack

Additionally, I am using VSCode, and saving the file causes the Gleam Language Server to crash immediately.

### Steps to reproduce

- Use the attached [testing.gleam.txt](https://github.com/user-attachments/files/19025633/testing.gleam.txt) file, which reliably reproduces the crash (requires _simplifile_ and _gleam_json_ libs).
- The decoding succeeds if a certain number of `decode.field` calls are commented out.
- The JSON file being decoded : [characters.json](https://github.com/user-attachments/files/19025638/characters.json).

GearsDatapacks:
This PR fixes #4287.
I've added the `stacker` dependency as suggested in the above issue; if we don't want to go down that route, we may be able to fix this issue another way.


lpil:
Thank you.

I had a little look at this before but I couldn't see another good way to do this due to the analyser recursively descending down the tree. I couldn't immediately see a nice way to rewrite it.

Did you have some idea for a different approach? I'm happy with this but interested in your thoughts.

GearsDatapacks:
Nothing concrete really. The only thing I could think of would be a similar thing to how you make a DFS search tail recursive, which is to make our own stack. Not sure how practical that would be though.

test/stack_overflow/src/project.gleam
```diff
@@ -0,0 +1,60 @@
+import gleam/dynamic/decode
```
lpil, line 1: Couldn't this be an normal unit test?

Typically we only write integration tests when we want to test the command line interface itself, or some IO.

lpil (commented):
That was my thought too. Let's go ahead with this then. Code looks good! I've left 1 question inline.

test/stack_overflow/src/project.gleam
GearsDatapacks: Probably, yeah. I'll see

lpil (approved):
Thank you!!!

PR #4580, opened 2025-05-09
feat(Makefile): add fmt, fmt-check and lint targets

### LINKED ISSUE #4579: Add `lint` and `fmt` targets in the Makefile
Noticed that in [CONTRIBUTING](https://github.com/gleam-lang/gleam/blob/main/CONTRIBUTING.md#clippy-linter) mentioned use of clippy to prevent one of possible CI fails
However Makefile missing formatting targets like fmt and lint
These targets can be used in CI to avoid raw usage of `clippy` or `rustfmt`

vardbabayan:
- `make fmt`        → runs `cargo fmt` to auto-format all Rust code
- `make fmt-check`  → runs `cargo fmt -- --check` to verify formatting
- `make lint`       → runs `cargo clippy -- -D warnings` to treat all warnings as errors

These targets can be used in CI to avoid raw usage of clippy or rustfmt

Closes #4579

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,18 @@ test: ## Run the compiler unit tests
 	cd test/running_modules && make test
 	cd test/subdir_ffi && make
 
+.PHONY: fmt
+fmt: ## Run rustfmt to format all Rust code
+	cargo fmt
+
+.PHONY: fmt-check
+fmt-check: ## Check that all Rust code is properly formatted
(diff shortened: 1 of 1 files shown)


lpil:
Thank you but we do not accept unplanned changes, any change needs to be agreed upon in an issue before a PR can be accepted.

PR #4587, opened 2025-05-11
Fix pattern match scoping

### LINKED ISSUE #3253: Pattern matching problem involving improper scoping

GearsDatapacks:
Fixes #3253


lpil (approved):
Lovely, thank you!!

PR #4588, opened 2025-05-11
Fixing typo in error message

Varpie:
I just saw a typo when reading the changelogs, `You might way` -> `You might want`.

I had to modify the snapshot tests as well, which was done with grep and sed, hopefully this is the correct way to fix them.


lpil (approved):
Thank you!!

PR #4589, opened 2025-05-12
Use new printer for signature help

scristobal:


lpil:
Hello! Is this ready for a review?

lpil:
Hi @scristobal, are you still working on this?

scristobal:
> Hi @scristobal, are you still working on this?

Ups, sorry busy weeks at work lately. I think this should be there.

Thanks for your patience!

lpil:
Not at all, thank you!

compiler-core/src/type_/printer.rs
```diff
@@ -181,6 +181,12 @@ impl Names {
         _ = self.type_variables.insert(id, local_alias.clone());
     }
 
+    pub fn map_new_variable(&mut self, old_id: u64, new_id: u64) {
```
lpil: Could you document what this does please? 🙏 

I was a bit confused by the word "map" as in my mind that means a function like `list.map`. Is there a name that someone like me with less context would find extra clear?

compiler-core/src/type_/environment.rs
```diff
@@ -591,6 +591,8 @@ impl Environment<'_> {
                                 // Check this in the hydrator, i.e. is it a created type
                                 let v = self.new_unbound_var();
                                 let _ = ids.insert(*id, v.clone());
+                                let new_id = self.previous_uid();
+                                self.names.map_new_variable(*id, new_id);
```
lpil: This looks good. What was the situation in which it wouldn't work without this change? A comment explaining it would be helpful for future readers.

lpil (commented):
Lovely, thank you. I've left some little questions inline.

compiler-core/src/type_/environment.rs
scristobal: Absolutely, great feedback!

I added a comment explaining why this is necessary, also changed `new_id` name which was confusing, coming from a previous id... 🤔 

I hope it is more clear now!

compiler-core/src/type_/printer.rs
scristobal: With pleasure!

renamed the function to `reassign_type_variable_alias` which is much more explicit, also added a small comment on top describing what it does. Hope it helps!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__signature_help__help_for_use_function_call_uses_precise_types_when_missing_some_arguments.snap
```diff
@@ -12,7 +12,7 @@ pub fn main() {
 
 
 ----- Signature help -----
-guard(Bool, a, fn() -> a) -> Float
+guard(Bool, b, fn() -> b) -> Float
```
GearsDatapacks: Do you know why this has changed? The generics are named `a`, so this should print `a`, not `b` here.

GearsDatapacks (commented):
This looks good! One small change seems odd to me, asked inline

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__signature_help__help_for_use_function_call_uses_precise_types_when_missing_some_arguments.snap
lpil: I think it is that `a` is taken by the `guard` function, but this use of the printer doesn't do the type id resetting thing that you added the other day. Should we do that here too?

compiler-core/src/type_/environment.rs
```diff
@@ -596,6 +596,14 @@ impl Environment<'_> {
                                 // Check this in the hydrator, i.e. is it a created type
                                 let v = self.new_unbound_var();
                                 let _ = ids.insert(*id, v.clone());
+
+                                // Preserve any user-provided name from the original generic variable
+                                // for the new unbound variable. This ensures error messages and type
+                                // displays continue to use meaningful names (e.g., "something") rather
+                                // than auto-generated ones (e.g., "a", "b").
```
lpil: Doesn't this mean that the previous variable no longer has the name? Instead of both of them having that name due to their pointing to the same type.

Maybe this is the cause of the behaviour @GearsDatapacks pointed out above.

lpil:
Do you know why there's some AI review thing in this PR? It doesn't look like it's enabled for this repo so I don't understand why it's here.

lpil:
Hello! Did you see the questions above? 🙏 It seems there may be a bug

scristobal:
> Do you know why there's some AI review thing in this PR? It doesn't look like it's enabled for this repo so I don't understand why it's here.

I think this is my own config, somehow copilot decided to review all my PRs regardless of the repo!!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__signature_help__help_for_use_function_call_uses_precise_types_when_missing_some_arguments.snap
scristobal: maybe this is related? 
https://github.com/gleam-lang/gleam/blob/fd4b2666f86f4c34835b06e86cc13079f14bc16e/compiler-core/src/type_/printer.rs#L414-L431

I need to dig deeper!

lpil:
Sorry, I meant my comment above about the possible bug in the code.

giacomocavalieri:
Hello! Remember to un-draft this if it is ready for a review

compiler-core/src/type_/environment.rs
scristobal: yep, there was a subtle (or not so subtle) bug with the previous approach :S

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__signature_help__help_for_use_function_call_uses_precise_types_when_missing_some_arguments.snap
scristobal: there was a problem with the previous approach, with the new approach this test is no longer changing :D

scristobal:
> Hello! Remember to un-draft this if it is ready for a review

It is ready now! I made some changes to the approach and deferred the aliases instead. 
Thanks for your patience! Hope is all good now!

PR #4590, opened 2025-05-12
Improve formatting of `echo` and `assert`

giacomocavalieri:
With this PR I've improved formatting of binary expressions used on the right of an echo/assert. Previously they wouldn't be indented properly!
I've only updated the changelog for `echo` as `assert` hasn't been released yet


lpil (approved):
Lovely, thank you

PR #4593, opened 2025-05-13
Make sure build tool does not try and run module whose main function is private

### LINKED ISSUE #4592: Compiler tries running modules with private main function
If I have a module with a provate `main` function and run `gleam run -m <that_module>` the compiler will not complain about the missing `main` function and still compile and run it, resulting in a runtime crash.

giacomocavalieri:
This PR fixes #4592 

The build tool would not check that the main function defined by a module is not private so `gleam run` would result in a runtime crash, instead of a build error

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -945,7 +945,7 @@ to `{}/{module}.gleam`.", origin.folder_name()
             Error::MainFunctionIsPrivate { module } => vec![
                 Diagnostic {
                     title: "Module does not have a public main function".into(),
-                    text: format!(
+                    text: wrap_format!(
                         "`{module}` has a main function, but it is private, so it cannot be run."
                     ),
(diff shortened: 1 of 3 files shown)


compiler-core/src/error.rs
```diff
@@ -939,6 +942,18 @@ to `{}/{module}.gleam`.", origin.folder_name()
...
 
+            Error::MainFunctionIsPrivate { module } => vec![
+                Diagnostic {
+                    title: "Module does not have a public main function".into(),
+                    text: format!(
+                        "`{module}` has a main function, but it is private, so it cannot be run."
+                    ),
+                    level: Level::Error,
+                    location: None,
+                    hint: Some(format!("Make the `main` function in the `{module}` module public."))
```
lpil, line 953: Wrap lines!

PR #4596, opened 2025-05-13
Fix project generation test

GearsDatapacks:
Since the latest version of the standard library, `gleam_http` no longer works. This PR updates the integration test to use `gleam_http` v4.


lpil (approved):
Thank you!

PR #4600, opened 2025-05-15
Update to Rust 1.87

GearsDatapacks:
This PR updates the codebase to Rust 1.87 by fixing clippy warnings on the latest version.
Most of these warnings concerned the sizes of various types, and so I've done the most basic solution which is to `Box` the big stuff.
There might be a better solution to some of these cases, so please have a look.


lpil (approved):
Thank you!

PR #4602, opened 2025-05-16
Made compiler show a hint to import unqualified types/values

### LINKED ISSUE #4297: Show hint to import unqualified types/values

mine-tech-oficial:
Continues #4304 and closes #4297.

The hint also considers the type/value's arity, and decides which module to hint is by preferring imported modules over importable ones, and sorting by name length. Also, I'd say the current ideas can be adapted to the `environment::suggest_modules` function so that it, too, considers arity.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -937,7 +937,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
             )
             .collect();
         let typed_parameters = environment
-            .get_type_constructor(&None, &name)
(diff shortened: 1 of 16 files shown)


mine-tech-oficial:
I just need to know where in the changelog I put. Should I just make a unreleased section?

lpil:
Hello! I've left feedback in the other PR

lpil:

PR #4604, opened 2025-05-17
Fix duplicate field bug

### LINKED ISSUE #4603: Cannot use `x0` as record field label with positional field on JavaScript target
I just randomly tried this for fun after seeing the JavaScript generated code for positional fields:

```gleam
type Test {
  Test(Int, x0: Int)
}
```

The generated code has duplicate constructor arguments:

```javascript

class Test extends $CustomType {
  constructor(x0, x0) {
    super();
    this[0] = x0;
    this.x0 = x0;
  }
}
```

Running with `gleam run --target javascript` yields the following error:

```
<file path>:
  constructor(x0, x0) {
                  ^^

SyntaxError: Duplicate parameter name not allowed in this context
    at compileSourceTextModule (node:internal/modules/esm/utils:338:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:468:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:515:27)

Node.js v22.14.0
```

I'm using Gleam 1.10.0 on NixOS.

GearsDatapacks:
Fixes #4603

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   properly in the generated documentation.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where fields name `x0` in records could cause invalid code to be
+  generated on the JavaScript target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.11.0-rc1 - 2025-05-15
 
 ### Compiler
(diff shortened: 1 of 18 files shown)


lpil (approved):
Thank you!!

PR #4609, opened 2025-05-18
Remove entire echo statement when followed by a literal expression

giacomocavalieri:
When using the `remove echos from this module` code action, instead of just removing `echo`, the LS is a bit smarter and if it's safe to do so (that is echo is used as a standalone statement, it's not the last one in a function, and is printing a literal value) it's also going to remove the literal value it was printing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 ## Unreleased
 
+### Language server
+
+- When using the "remove `echo`" code action, the language server will also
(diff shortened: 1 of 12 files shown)


GearsDatapacks:
Will this work for echoing variables as well? That's a pretty common case too

giacomocavalieri:
Nice catch, I've updated it to remove vars as well!

lpil:
This would need to be in a new bit of the changelog!

giacomocavalieri:
Oops fixed it!

PR #4611, opened 2025-05-18
Add jsdoc comments to generated javascript code

### LINKED ISSUE #4597: Gleam Doc Comments into Javascript JsDoc
I would be nice to turn the `///` doc comments above functions and modules into `/** */` when exported to a Javascript target. That allows the documentation to be available when hovering the exported function.

### Gleam

```gleam
/// This is the Gleam version
pub fn add(a: Int, b: Int) -> Int {
  a + b
}
```

### Javascript

```js
/** This is the JS version */
export function add(a, b) {
  return a + b;
}
```

giacomocavalieri:
This PR fixes #4597 by adding jsdoc comments to generated functions, constants and constructors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## Unreleased
 
+### Build tool
+
+- Generated JavaScript functions, constants, and custom type constructors now
+  include any doc comment as a JSDoc comment, making it easier to use the
+  generated code and browse its documentation from JavaScript.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+## v1.11.0-rc1 - 2025-05-15
+
 ### Compiler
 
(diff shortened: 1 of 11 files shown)


PgBiel:
Chatted a bit on Discord, and it seems like this PR didn't handle `*/` in docstrings, which unfortunately generates a syntax error in the generated JSDoc.

For reference, here's how Glistix handles it on the Nix target: https://github.com/Glistix/glistix/pull/55/files#diff-7c475047d730fc741233c1939bb46932a43cf70105e9bb594d6c0093bd745b7a

In particular, we replace with `* /`. Don't have a better idea at the moment. Zero-width spaces were considered, but I'd rather make it clear that they are not the same character combination than to "hide" it and potentially make things hard to debug later on if someone copies the docs.

PR #4617, opened 2025-05-22
Consistent let assert and assert location information

lpil:


compiler-core/src/erlang.rs
```diff
@@ -2242,12 +2248,13 @@ fn assignment<'a>(
         AssignmentKind::Let | AssignmentKind::Generated => {
             let_(&assignment.value, &assignment.pattern, env)
         }
-        AssignmentKind::Assert { message, .. } => let_assert(
+        AssignmentKind::Assert { message, location } => let_assert(
```
GearsDatapacks, line 2251: Is this the correct location we want to pass in? Currently this will point to the `assert` keyword itself, within the `let assert` statement.

compiler-core/src/javascript/decision.rs
```diff
@@ -539,7 +543,17 @@ impl<'generator, 'module, 'a> LetPrinter<'generator, 'module, 'a> {
                 this.wrap_expression(message)
             })?,
         };
-        Ok(generator.throw_error("let_assert", &message, *location, [("value", subject)]))
+        Ok(generator.throw_error(
+            "let_assert",
+            &message,
+            *location,
```
GearsDatapacks, line 549: Same for here

GearsDatapacks (commented):
Good idea! I have one question inline

lpil:
Today I think so. If we find a use case for that information in future it can be added later without it being a breaking change.

GearsDatapacks (approved):
Sounds good to me!

compiler-core/src/ast.rs
```diff
@@ -2336,7 +2336,10 @@ pub enum AssignmentKind<Expression> {
     Generated,
     /// let assert x = ...
     Assert {
+        /// The src byte span of the `let assert`
```
giacomocavalieri, line 2339: It's not really clear if this spans just the two keywords or the entire expression, maybe we could add an example?
```
/// ```gleam
/// let assert Wibble = todo
/// ^^^^^^^^^^^^^^^^^^^^^^^^ This
/// ```
```
If it is indeed the entire expression

giacomocavalieri (approved):
Look good to me, just left a tiny comment

PR #4623, opened 2025-05-25
Remove a duplicated changelog entry

00-kat:
It's the same as the one directly below it.


GearsDatapacks:
Oops! Don't know how that happened, probably a mistake while rebasing

lpil (approved):
Thank you

PR #4632, opened 2025-05-28
Make `static_vcruntime` dependency windows-specific

GearsDatapacks:
The `static_vcruntime` package is only for windows, so this PR makes it a conditional dependency, meaning it is not even downloaded when building on non-windows targets.


lpil (approved):
Thank you!

PR #4638, opened 2025-05-30
Make sure bit array float segments will not match on NaN and Infinity

### LINKED ISSUE #4637: Incorrect 'Unreachable pattern' warning on v1.11.0-rc2
On Gleam v1.11.0-rc2 this code:

```gleam
pub fn main() {
  let positive_infinity = <<0x7FF0000000000000:64>>
  case positive_infinity {
    <<_:64-float>> -> "Float"
    <<0x7FF0000000000000:64>> -> "+Infinity"
    _ -> "Other"
  }
  |> echo
}
```

Gives the warning:

```
warning: Unreachable pattern
   ┌─ ./foo.gleam:10:5
   │
10 │     <<0x7FF0000000000000:64>> -> 2
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^

This pattern cannot be reached as a previous pattern matches the same
values.

```

But the pattern is very much reachable on the Erlang target. It's not reachable on the JavaScript target due to that target supporting infinity and NaN floating point values.

giacomocavalieri:
This PR closes #4637.
Now a float segment will not match on a sequence of bits representing a `NaN` or `Infinity`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,3 +11,7 @@
 ### Formatter
 
 ### Bug fixes
+
+- Fixed a bug where a bit array segment matching on a floating point number
+  would match with `NaN` or `Infinity` on the JavaScript target.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -859,7 +859,7 @@ pub enum BitArrayTest {
     /// pattern where the size is a variable with a negative value will never
     /// match. So we check this to make sure the test will fail.
     ///
-    VariableIsPositive {
+    VariableIsNotNegative {
         variable: VariableUsage,
     },
 


richard-viney:
Thanks for looking into this!

Are we sure this approach is the right one?

I can think of two options for float pattern matching behaviour here:

1. Match anything representable by `Float` on the current target.
2. Match only finite floats because that's what the Erlang target supports.

Going with the latter approach as this PR does seems less in line with the philosophy of the target defining the constraints of `Float` and `Int` types, and is also a breaking change on the JavaScript target as now `Infinity` and `NaN` bytes won't pattern match, whereas before they matched because that target supports those values in its `Float` type.

Thoughts?

giacomocavalieri:
Yes we want to go for 2 and make sure that pattern matching behaves exactly the same on both targets. The way the compiler behaves right now is not right and we don't consider fixing a bug a breaking change; but yeah code that relies on this bug of the compiler will stop working once we fix it

lpil:
Oh bother I forgot to merge this

PR #4645, opened 2025-05-31
Fix `assert` warnings

GearsDatapacks:
I noticed that the compiler was incorrectly warning for asserting the comparison of literal values, when the literal values contained variables. For example:
```gleam
assert <some_variable:utf8_codepoint> == <<"X">>
```
This PR fixes that.


lpil (approved):
Thank you!

PR #4656, opened 2025-06-01
Fix function generation for function captures

### LINKED ISSUE #4653: "Generate function" code action behaves weirdly with function captures
For example, if you trigger it on the following code:
```gleam
pub fn main() {
  something(_, 1)
}
```
It will generate:
```gleam
fn something(_capture: a, int: Int) -> b {
  todo
}
```
Here, it should probably generate a generic name rather than `_capture`

GearsDatapacks:
Fixes #4653


lpil (approved):
Thank you!

PR #4662, opened 2025-06-04
Add a document to track things we'd like for v2

giacomocavalieri:
There's probably a lot of stuff and pieces of deprecated syntax I might have forgotten but at least it's a start to track everything in a single place


lpil (approved):
Thank you

Issue #4667, opened 2025-06-04
Do not generate code for unused private types and functions


lpil:
They are unused, so they are not needed at all! Any definition that we emit an "unused" warning for should be skipped.

This should be implemented for both Erlang and JavaScript

PR #4669, opened 2025-06-04
Dead code elimination for definitions

lpil:
Closes https://github.com/gleam-lang/gleam/issues/4667


compiler-core/src/javascript.rs
```diff
@@ -308,6 +308,17 @@ impl<'a> Generator<'a> {
             // Handled in collect_definitions
             Definition::CustomType(CustomType { .. }) => None,
 
+            // If a custom type is unused then we don't need to generate code for it
```
GearsDatapacks, line 311: This comment references custom types but this code is about functions and constants

GearsDatapacks: This comment references custom types but this code is about functions and constants

GearsDatapacks (commented):
Looks good to me! Spotted one tiny mistake in a comment, but other than that code all looks fine

compiler-core/src/javascript.rs
lpil: oops

PR #4674, opened 2025-06-08
Add more control over list formatting

### LINKED ISSUE #4628: Allow more control when formatting lists
Currently the formatter attempts to make lists as compact as possible. It discards newlines and for simple expressions it fits them on one line where possible.

This is generally a good approach, but it falls apart in not-uncommon situations where the programmer has clear reason to format their data in a particular way. One such example of this would be when replicating the structure of data in some external definition or specification.

This has become sufficiently irritating to people that they have resorted to hoop-jumping, writing code in strange ways in order to get formatting they like. One example is through placing empty comments. Comments are for documentation and not for formatting, so we want to prevent this use.

## Proposal

I propose we have lists behave similar to pipelines, giving the programmer more control so they can pick between a series of acceptable formats. We should also preserve newlines within lists.

The formats would be:

```gleam
let items = [1, 2, 3, 4]
```

All on one line.

```gleam
let items = [
  "one", "two", "three", "four", "five", "six", "seven", "eight", "nine",
  "ten", "eleven", "twelve",
]
```

Split over multiple lines, with multiple elements per line, as many elements as will fit on each line.

```gleam
let items = [
  "one",
  "two",
  "three",
  "four",
  "five",
  "six",
  "seven",
  "eight",
  "nine",
  "ten",
  "eleven",
  "twelve",
]
```

Split over multiple lines, with one element per line.

If the programmer has their code in any of these styles (and it fits within the column limit) then it should stay as-is. If the code does not fit within the limit it should be re-written to one of the other styles.

We will need to decide on what the rule is for picking between which style to use. It would ideally be relatively easy to pick the style you want most of the time.

giacomocavalieri:
This PR closes #4628:
- by adding a trailing comma to a list the formatter will be forced to break it with one item per line
- by removing a trailing comma from a list the formatter will try to fit it on a single line, if it can't it will break it one item per line
- by putting multiple elements on the same line (if the list is only made up of simple constants), the formatter will try and fit as many elements as possible on a single line

So now the formatter will never pack a list of constants **unless explicitly told so**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,10 +49,9 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
-- The formatter now allows more control over how lists made up of simple
-  constants are formatted. If a list is split with multiple elements on the same
-  line, removing the trailing comma will make sure the formatter keeps each item
-  on its own line:
+- The formatter now allows more control over how lists are formatted.
+  If a list is split with multiple elements on the same line, removing the
+  trailing comma will make sure the formatter keeps each item on its own line:
 
   ```gleam
   pub fn my_favourite_pokemon() -> List(String) {
@@ -79,7 +78,7 @@
 
 - The formatter no longer removes empty lines between list items. In case an
   empty line is added between list items they will all be split on multiple
-  lines.
+  lines. For example:
 
   ```gleam
   pub fn main() {
@@ -103,6 +102,8 @@
     ]
   }
   ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
 ### Compiler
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,108 @@
...
+  ```gleam
+  pub fn my_favourite_pokemon() -> List(String) {
+    ["natu", "chimecho", "milotic"]
+  }
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- The formatter now allows more control over how lists made up of simple
+  constants are formatted. If a list is split with multiple elements on the same
```
lpil: I think we wanted to remove the whole concept of a simple constants, it being a band-aid for not having control over this previously

Issue #3167, opened 2024-05-20
Piped function call is not tail call optimised on the JS target


giacomocavalieri:
Using pipes and function captures makes it so the compiler won't TCO a function on the js target.
Consider:
```gleam
pub fn do_len(list, acc) {
  case list {
    [] -> acc
    [_, ..rest] -> do_len(rest, acc + 1)
  }
}
```
This code gets optimised to:
```js
+export function do_len(loop$list, loop$acc) {
  while (true) {
    let list = loop$list;
    let acc = loop$acc;
    if (list.hasLength(0)) {
      return acc;
    } else {
      let rest = list.tail;
      loop$list = rest;
      loop$acc = acc + 1;
    }
  }
}
```
However, if we rewrite the code using a pipe and a function capture it no longer applies the optimisation:
```js
export function do_len(list, acc) {
  if (list.hasLength(0)) {
    return acc;
  } else {
    let rest = list.tail;
    let _pipe = rest;
    return ((_capture) => { return do_len(_capture, acc + 1); })(_pipe);
  }
}
```
I would expect the generated code to look roughly the same for both examples. Maybe this is another point that would benefit from having pipes in the typed AST, then we could be smarter with the code generation

lpil:
Please include the code that doesn't optimise 🙏

lpil:
This is due to the function capture being used. If you run the format, in this case, it would be removed and the resulting code would be optimised. It would be nice to optimise for the captures too though.

giacomocavalieri:
I'd like to give this a try!

PR #4677, opened 2025-06-09
Implement basic function inlining

### LINKED ISSUE #3869: Implement basic function inlining
As discussed in #3722, and [this thread](https://github.com/gleam-lang/gleam/discussions/3722#discussioncomment-11335509).
As pointed out [here](https://github.com/gleam-lang/gleam/issues/3830#issuecomment-2489755358), some key functions such as `bool.guard` can cause a stack overflow with recursion in javascript, while giving a false appearance of being tail recursive.

Inlining `bool.guard` would solve this issue. As Louis suggested, we could initially implement a basic version of inlining (i.e. the inlining itself, without complex heuristics etc.), and later we can expand and improved upon this. @lpil could you elaborate a bit on how you were imagining we would decide which functions to inline?

I did a bit of research into how function inlining works when the about discussion was first created, so I'll write up a quick summary of how this will work soon, and then I'll look into implementing this.

GearsDatapacks:
Closes #3869 
This PR implements the basic process of function inlining, without any heuristics to determine which functions to inline, we just use a hardcoded list.
The code for this is pretty much done, so please take a look, but before this is merged there's a few more things we should do:
- Decide on the list of functions we want to use. For now I've chosen a few of the basic `result` and `bool` combinators which seem to be the highest value but we should decide if there are any others we want to inline as well.
- Benchmark. We should come up with some benchmarks to determine if the inlining process significantly impacts compiler performance, and if so think about ways to mitigate that.
- Test. I've written some basic codegen tests in the `javascript::tests::inlining` and `erlang::tests::inlining` modules, but please suggest any other tests to write also. As well, we should run this version of the compiler on lots of existing packages to make sure nothing is broken.


compiler-core/src/erlang/tests.rs
```diff
@@ -40,7 +41,12 @@ mod type_params;
 mod use_;
 mod variables;
 
-pub fn compile_test_project(src: &str, src_path: &str, dep: Option<(&str, &str, &str)>) -> String {
+pub fn compile_test_project(
+    src: &str,
+    src_path: &str,
+    dep: Option<(&str, &str, &str)>,
+    perform_inlining: bool,
```
lpil: How come this is configurable?

GearsDatapacks: There were a few tests which were sort of broken by inlining, so I decided to only enable it for the specific inlining tests. One example of that is the test for wrapping anonymous functions in brackets on the JavaScript target.
The test ensures that this Gleam code:
```gleam
fn(x) { x + 1 }(10)
```
Is compiled to this JavaScript code, ensuring that brackets are added for correct precedence:
```javascript
(x => x + 1)(10)
```
However, the inlining was causing it to compile to:
```javascript
{
  return 10 + 1
}
```
Which sort of defeats the point. I considered asking you what to do for these tests, but decided in the end to just follow the solution here.

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1570 @@
...
+//! Every time we encounter a function call, we decide whether or not we can
+//! inline it. For now, the criteria for inlining is very simple, although a
+//! more complex heuristic-based approach will likely be implemented in the
+//! future. For now though, a function can be inlined if:
+//! It is a standard library function within the hardcoded list - which can be
+//! found in the `inline_function` function - or, it is an anonymous function.
+//!
+//! Inlining anonymous functions allows us to:
+//! - Remove calls to parameters of higher-order functions once those higher-
+//!   order functions have been inlined
```
lpil, line 66: What does this mean?

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1570 @@
...
+//! - Remove calls to anonymous functions in pipelines
+//! - Remove calls to function captures in pipelines
```
lpil: I think I understand these, but it took me quite a bit of going back and forth to get an idea. Could you include examples for each of these please 🙏

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1570 @@
...
+            } => match &constructor.variant {
+                // If this variable can be inlined, replace it with its value.
+                // See the `inline_variables` documentation for an explanation.
+                ValueConstructorVariant::LocalVariable { .. } => {
+                    // We remove the variable as inlined variables can only be
+                    // inlined once.
+                    match self.inline_variables.remove(name) {
+                        Some(expression) => expression,
+                        None => expression,
+                    }
```
lpil, line 315: From the comment on `inline_variables` I got the impression that inlining can happen if a variable is used once, but it here looks like it happens for the first time but not any later times, and whether it is once or many times is not checked at all. Am I misunderstanding?

I couldn't see tests for this so it's a bit tricky to understand

compiler-core/src/erlang/tests.rs
lpil: The inlining solution is still a solution to the problem, and the non-inlining one isn't reachable any more (?) so I think we want to have the test do the inlining one that real code gets.

GearsDatapacks: Ok, sounds good

compiler-core/src/inline.rs
GearsDatapacks: We check which parameters are used exactly once when we convert a `Function` to an `InlinableFunction`, using [this field](https://github.com/gleam-lang/gleam/pull/4677/files#diff-23362c14c76a2ee7d19a8f4bffc60dd36e2186c906142507ffdbe3e08589ecb7R966). Maybe that needs to be made more clear?

CHANGELOG.md
```diff
@@ -12,6 +12,38 @@
...
+  ```gleam
+  pub fn count(from: Int, to: Int) -> Int {
+    use <- bool.guard(when: from >= to, return: from)
+    io.println(int.to_string())
+    count(from + 1, to)
+  }
+  ```
+
+  Would previously cause a stack overflow on the JavaScript target for large
+  value. Now it is rewritten to:
```
```suggestion
  values. Now it is rewritten to:
```

compiler-core/src/erlang/snapshots/gleam_core__erlang__tests__integration_test1_5.snap
```diff
@@ -23,6 +21,15 @@ add(X, Y) ->
...
+        add(_capture, 1)
+    end,
+    _pipe@2 = begin
+        _capture@1 = _pipe@1,
+        add(2, _capture@1)
+    end,
+    begin
+        _capture@2 = _pipe@2,
+        add(_capture@2, 3)
+    end.
```
giacomocavalieri: Maybe I'm misunderstanding inlining here, but this doesn't look like an optimisation

compiler-core/src/erlang/snapshots/gleam_core__erlang__tests__integration_test3.snap
```diff
@@ -20,7 +18,4 @@ pub fn y() { fn() { Point }()(4, 6) }
 -file("project/test/my/mod.gleam", 2).
 -spec y() -> point().
 y() ->
-    ((fun() -> fun(Field@0, Field@1) -> {point, Field@0, Field@1} end end)())(
-        4,
-        6
-    ).
+    {point, 4, 6}.
```
giacomocavalieri, line 21: Mind blown!

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1639 @@
...
+//!   let _pipe1 = "10"
+//!   let _pipe2 = int.parse(_pipe1)
+//!   let _pipe3 = result.unwrap(_pipe2, 0)
+//!   let _pipe4 = _pipe3 * _pipe3
+//!   something_else(_pipe4)
+//!   ```
+//!
+//! See documentation of individual functions to explain better how the process
+//! works.
+//!
```
giacomocavalieri, line 109: Love the documentation here!!

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1639 @@
...
+                type_,
+                record: Box::new(self.assignment(*record)),
+                constructor: self.boxed_expression(constructor),
+                args,
+            },
+        }
+    }
+
+    /// Where the magic happens. First, we check the left-hand side of the call
+    /// so see if it's something we can inline. If not, we continue to walk the
```
giacomocavalieri:
```suggestion
    /// to see if it's something we can inline. If not, we continue to walk the
```

giacomocavalieri (commented):
This is amazing, and the code was such a pleasure to read!! I've left just a couple of nits

compiler-core/src/erlang/snapshots/gleam_core__erlang__tests__integration_test1_5.snap
GearsDatapacks: Yeah, this is a tricky one. Basically, up until this point, we've had a special case in the code generators for function captures inside a pipeline. Now, this is hitting the inlining process first.

Since that was a special case for a very specific scenario, it was able to generate the most optimal code for that situation, whereas inlining is a bit more general and doesn't handle that case quite as well.

With a bit more code though, I could probably make this compile to what it did before

test-helpers-rs/src/lib.rs
```diff
@@ -35,6 +35,8 @@ impl TestCompileOutput {
             buffer.push('\n');
 
             let extension = path.extension();
+            dbg!(extension, &content);
```
lpil: dbg!

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__pipes__pipe_in_call.snap
```diff
@@ -39,5 +37,5 @@ main() ->
             _pipe@1 = 1,
             two(_pipe@1, 2)
         end,
-        _pipe
+        _pipe@1
```
DanielSherlock: This looks like it's now using the wrong pipe variable. I think it should output `_pipe` (as it did before) and not the `_pipe@1` variable from the inner pipe expression.

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__parameters_from_nested_functions_are_correctly_inlined.snap
```diff
@@ -0,0 +1,73 @@
...
+halve_all(A, B, C) ->
+    case divide(A, 2) of
+        {ok, Value} ->
+            X = Value,
+            case divide(B, 2) of
+                {ok, Value@1} ->
+                    Y = Value@1,
+                    case divide(C, 2) of
+                        {ok, Value@2} ->
+                            {ok, {X, Y, Value@2}};
```
DanielSherlock, line 61: Not an issue at all - I just found it interesting that `x` and `y` are left un-optimised after inlining while `z` gets the full treatment. (The same can be seen in the JS output.)

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1780 @@
...
+    if package != STDLIB_PACKAGE_NAME {
+        return false;
+    }
+
+    match (module, name) {
+        // These are the functions which we currently inline
+        ("gleam/bool", "guard") => true,
+        ("gleam/bool", "lazy_guard") => true,
+        ("gleam/result", "try") => true,
+        ("gleam/result", "then") => true,
```
DanielSherlock: Although I don't think you've officially decided yet on exactly which standard library functions to inline... I recently noticed that `result.then` has just been deprecated in [version 0.61.0](https://github.com/gleam-lang/stdlib/commit/de094326f150cb08ecdd17aa895292ad56b81aab) of the standard library.

compiler-core/src/erlang/tests/inlining.rs
```diff
@@ -0,0 +1,305 @@
...
+"
+        ),
+        "
+import gleam/bool
+
+pub fn main() {
+  bool.guard(True)
+}
+"
+    );
```
DanielSherlock: I read the snapshots first, so this test totally confused me.

It's probably worth adding a comment visible in the snapshot itself (perhaps right below `import gleam/bool`) to explain what's going on. Maybe something like:

> This test checks that a named function which uses it's parameter twice is inlined correctly. But as we currently only inline a specific handful of named functions (none of which use their parameters twice) the setup for this test has overwritten `bool.guard` to be a completely different function of the relevant form.

The same goes for the equivalent JS test.

compiler-core/src/inline.rs
GearsDatapacks: Ah yes, that's more recent than when this code was written

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__parameters_from_nested_functions_are_correctly_inlined.snap
GearsDatapacks: Yeah, this is a sort of a limitation of the current system. Since all three variables reference different `value` variables, inlining all of them would cause invalid code to be generated, so we avoid that situation for now. I believe this behaviour is documented somewhere in the code.
In the future we may be able to improve this

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__pipes__pipe_in_call.snap
GearsDatapacks: Good catch! I wonder what changed here

compiler-core/src/erlang/tests/inlining.rs
GearsDatapacks: Yeah that's possible. I'm not a big fan of this test really, I would like some other way to be able to test arbitrary function implementations. Maybe there's something I could change to make something configurable when testing

GearsDatapacks:
Not sure if this is the right way to go about it, but I've added a special `always_inline` function, which will always be inlined in test code, so we can easily test specific scenarios with custom functions.

lpil:
This is looking really good!

Have you done any measuring of the impact on this change on Gleam programs? Performance and code size are interesting, including with combinations of minification, gzip and/or brotoli for JavaScript. We want to make sure that we haven't in some places caused unexpected problems.

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_function_with_use_and_anonymous.snap
```diff
@@ -0,0 +1,38 @@
+---
+source: compiler-core/src/erlang/tests/inlining.rs
+expression: "\nimport gleam/bool\n\npub fn divide(a, b) {\n  use <- bool.lazy_guard(b == 0, fn() { panic as \"Cannot divide by 0\" })\n  a / b\n}\n"
+---
+----- SOURCE CODE
+
+import gleam/bool
+
+pub fn divide(a, b) {
+  use <- bool.lazy_guard(b == 0, fn() { panic as "Cannot divide by 0" })
```
giacomocavalieri, line 10: Do we have a test where this is not lazy to make sure we always throw the exception?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__inlining__inline_function_capture_in_pipe.snap
```diff
@@ -0,0 +1,22 @@
...
+source: compiler-core/src/javascript/tests/inlining.rs
+expression: "\npub fn main() {\n  1 |> add(4, _)\n}\n\nfn add(a, b) { a + b }\n"
+---
+----- SOURCE CODE
+
+pub fn main() {
+  1 |> add(4, _)
+}
+
+fn add(a, b) { a + b }
```
giacomocavalieri, line 11: Reading the generated code it seems like the inlining could also fix this issue here https://github.com/gleam-lang/gleam/issues/3167, could we add a test for it as well?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__inlining__inline_function_with_use_becomes_tail_recursive.snap
```diff
@@ -0,0 +1,45 @@
+---
+source: compiler-core/src/javascript/tests/inlining.rs
+expression: "\nimport gleam/bool\n\npub fn count(from: Int, to: Int) -> Int {\n  use <- bool.guard(when: from >= to, return: from)\n  echo from\n  count(from + 1, to)\n}\n"
+---
+----- SOURCE CODE
+
+import gleam/bool
+
+pub fn count(from: Int, to: Int) -> Int {
+  use <- bool.guard(when: from >= to, return: from)
```
giacomocavalieri, line 10: Could we add a test with some side effecting code as the `return:` argument like with the Erlang one? Just to make sure it's always evaluated in the correct order despite the inlining

compiler-core/src/inline.rs
giacomocavalieri: Yeah I had the same doubt reading this code, maybe it would be nice to add a comment in the code explaining that's not the case with a pointer to the relevant bit of code where you make the check.

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1786 @@
...
+fn expand_block(expression: TypedExpr) -> TypedExpr {
+    match expression {
+        TypedExpr::Block {
+            location,
+            statements,
+        } if statements.len() == 1 => {
+            let first = statements
+                .into_iter()
+                .next()
+                .expect("Vec1 always has a first element");
```
lpil: Prefer `statements.split_off_first()` rather than asserting here

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1786 @@
...
+                        // First, we do the actual inlining, by converting it to
+                        // an anonymous function.
+                        let TypedExpr::Fn {
+                            args: parameters,
+                            body,
+                            ..
+                        } = function.to_anonymous_function()
+                        else {
+                            unreachable!("to_anonymous_function always returns TypedExpr::fn");
+                        };
```
lpil: Don't panic when it's possible to write code where the invariants are enforced using the type system! Instead of returning a typed expr which _might_ be a ::fn, return the arguments and the body, as that's the only data that is actually needed.

compiler-core/src/inline.rs
```diff
@@ -0,0 +1,1786 @@
...
+                        .and_then(|module| module.inline_functions.get(name))
+                    {
+                        let TypedExpr::Fn {
+                            args: parameters,
+                            body,
+                            ..
+                        } = function.to_anonymous_function()
+                        else {
+                            unreachable!("to_anonymous_function always returns TypedExpr::fn");
+                        };
```
lpil: The above comment applies to this use of the method also

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__inlining__inline_function_with_use_becomes_tail_recursive.snap
GearsDatapacks: Which "Erlang one" are you referring to? All the tests are identical between the two targets

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_function_with_use_and_anonymous.snap
GearsDatapacks: There's [this test](https://github.com/gleam-lang/gleam/pull/4677/files#diff-64bb15a4afda7b3a89e58ff1d6fcf2ad72b31387143c3c9ff2075c82dbea393a), where a function is called which always panics. Does that cover what you want here?

PR #4681, opened 2025-06-12
Improve echo

lpil:
These are the improvements in the stdlib ported over.


compiler-core/templates/echo.mjs
```diff
@@ -93,26 +93,66 @@ function echo$inspect(v) {
...
+    if (v instanceof Set)
+      return `//js(Set(${[...v].map(echo$inspect).join(", ")}))`;
+    return echo$inspectObject(v);
+  } catch (e) {
+    if (e instanceof RangeError) return "//js(circular)";
+    throw e;
+  }
+}
+
+export function echo$inspectList(list) {
```
GearsDatapacks: Does it make sense to print charlists on the javascript target? I can't imagine they will crop up very much.

test-output/src/tests/echo.rs
```diff
@@ -134,6 +134,11 @@ fn echo_bool() {
     assert_echo!("echo_bool");
 }
 
+#[test]
+fn echo_charlist() {
+    assert_echo!("echo_charlist");
+}
+
```
GearsDatapacks, line 141: Can we get a test for echoing circular references?

compiler-core/templates/echo.mjs
lpil: I don't think it's useful but I think we should aim to have consistent printing where possible

test-output/src/tests/echo.rs
lpil: Oh yes, very silly of me to not have that

GearsDatapacks (approved):
Looks good now

lpil:
Thank you

PR #4685, opened 2025-06-15
Improve the look of error labels

giacomocavalieri:
This PR changes the look of secondary labels in error messages and warnings so they are no longer highlighted in red/yellow. This way additional context that is added to errors using extra labels looks a bit nicer.


lpil (approved):
Thank you!

PR #4686, opened 2025-06-16
Better error message for unknown variable named as an ignored variable

### LINKED ISSUE #4512: Show custom warning when using an ignored variable
[Playground link](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIADgK4BGABAGZI0C2AhnEgBQCUNwAOoxoAbCPhoB9OggQ0AvDT4gAFnEUCaGqTIA+APhoQcShAIC+AkKaA===). Currently, if you ignore a variable by prefixing it with `_`, then use it elsewhere in the code, you just get the normal "The name is not in scope" warning. Ideally, the compiler would realize you were trying to use a variable that's ignored, and say something like "Hint - maybe you meant to use this variable, but it's currently ignored".

giacomocavalieri:
When a variable has the same name as an ignored variable the compiler can now show more context in the "unknown variable" error, pointing to the ignored variable.
This PR fixes #4512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,33 @@
   shadows an imported name in the current module.
   ([Aayush Tripathi](https://github.com/aayush-tripathi))
 
+- The compiler can now tell when an unknown variable might be referring to an
+  ignored variable and provide an helpful error message highlighting it. For
+  example, this piece of code:
+
+  ```gleam
+  pub fn go() {
+    let _x = 1
+    x + 1
+  }
+  ```
+
+  Now results in the following error:
+
+  ```
+  error: Unknown variable
+    ┌─ /src/one/two.gleam:4:3
+    │
+  3 │   let _x = 1
+    │       -- Did you mean to use this ignored variable?
+  4 │   x + 1
+    │   ^
+
+  The name `x` is not in scope here.
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
 - `gleam update`, `gleam deps update`, and `gleam deps download` will now print
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,15 +2,6 @@
 
 ## Unreleased
 
-### Build tool
-
-- Generated JavaScript functions, constants, and custom type constructors now
-  include any doc comment as a JSDoc comment, making it easier to use the
-  generated code and browse its documentation from JavaScript.
-  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
-
-## v1.11.0-rc1 - 2025-05-15
-
 ### Compiler
 
 - Generated JavaScript functions, constants, and custom type constructors now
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -2,15 +2,6 @@
...
 
-### Build tool
-
-- Generated JavaScript functions, constants, and custom type constructors now
-  include any doc comment as a JSDoc comment, making it easier to use the
-  generated code and browse its documentation from JavaScript.
-  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
-
-## v1.11.0-rc1 - 2025-05-15
-
```
giacomocavalieri, line 13: This was duplicated and had a stray `v1.11.0-rc1` heading, my guess is it comes from a rebase gone wrong. I've deleted it!

compiler-core/src/type_/environment.rs
```diff
@@ -148,6 +153,7 @@ impl<'a> Environment<'a> {
 #[derive(Debug)]
 pub struct ScopeResetData {
     local_values: im::HashMap<EcoString, ValueConstructor>,
+    ignored_names: HashMap<EcoString, SrcSpan>,
```
lpil: Prefer immutable data structure when lots of copies have to be made
```suggestion
    ignored_names: im::HashMap<EcoString, SrcSpan>,
```

compiler-core/src/type_/error.rs
```diff
@@ -1297,6 +1298,7 @@ pub fn convert_get_value_constructor_error(
             location,
             name,
             variables,
+            ignored_variables: HashMap::new(),
```
lpil: This is a lot of never-used data to pass around so let's use `discarded_location: Option<SrcSpan>` or something like that instead

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__shadowed_fn_argument.snap
```diff
@@ -0,0 +1,23 @@
...
+
+
+----- ERROR
+error: Unknown variable
+  ┌─ /src/one/two.gleam:4:5
+  │
+3 │   fn(_y) {
+  │      -- Did you mean to use this ignored variable?
+4 │     y + x
+  │     ^
```
lpil: This seems a bit confusing as the actual error location is less prominent and comes after the suggestion for how to fix it.

Perhaps something like this?

```
error: Unknown variable
  ┌─ /src/one/two.gleam:4:5
  │
3 │   fn(_y) {
  │      -- This value is discarded
4 │     y + x
  │     ^ So this is not in scope

Hint: Change `_y` to `y` or reference another variable.
```

giacomocavalieri: this looks amazing

llakala:
Why doesn't Github notify me when an issue I make has a PR marking it as fixed?

Thanks so much for doing this!

giacomocavalieri:
Everything should be fixed and ready now!

PR #4687, opened 2025-06-16
Add warning for redundant comparisons

### LINKED ISSUE #4205: Warn on useless comparisons

giacomocavalieri:
This PR closes #4205 
I've also updated the assert warning when doing a redundant assertion:
- I've left a specialized warning when asserting on a known boolean
- I've removed the special casing for the other warnings since they are now covered by this new one that provides some better explanation

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -913,6 +913,7 @@ impl TypedExpr {
     }
 
     #[must_use]
+    // TODO)) is this the same as is_record_builder
     pub fn is_record_constructor(&self) -> bool {
         match self {
             TypedExpr::Var {
@@ -1048,15 +1049,6 @@ impl TypedExpr {
     pub(crate) fn is_invalid(&self) -> bool {
         match self {
             TypedExpr::Invalid { .. } => true,
-            _ => false,
-        }
-    }
-
-    pub(crate) fn is_literal_bool(&self) -> bool {
-        match self {
-            TypedExpr::Var {
-                constructor, name, ..
-            } if name == "True" || name == "False" => constructor.type_.is_bool(),
             _ => false,
         }
     }
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -694,14 +694,12 @@ impl TypedExpr {
         }
     }
 
-    pub fn is_known_value(&self) -> bool {
+    pub fn is_known_bool(&self) -> bool {
         match self {
-            TypedExpr::BinOp { left, right, .. } => left.is_known_value() && right.is_known_value(),
-
-            TypedExpr::NegateBool { value, .. } | TypedExpr::NegateInt { value, .. } => {
-                value.is_known_value()
-            }
-
+            TypedExpr::BinOp {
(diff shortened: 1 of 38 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -1741,6 +1743,155 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    _ => None,
+                }
+            }
+
+            // If two variables are the same then we know that the comparison
+            // will always succeed, but the opposite is not true: if the two
+            // variables are different we can't be sure the comparison will
+            // always fail.
+            (TypedExpr::Var { name: one, .. }, TypedExpr::Var { name: other, .. }) => {
+                if left.is_literal_bool() && right.is_literal_bool() {
```
lpil: We could do this for all record constructors that are the same rather than just bools?

giacomocavalieri: Yeah! I updated the comparison logic in my last commit! It's a lot smarter now, even for lists and tuples. What do you think?

compiler-core/src/ast/typed.rs
```diff
@@ -915,6 +913,7 @@ impl TypedExpr {
     }
 
     #[must_use]
+    // TODO)) is this the same as is_record_builder
```
lpil: What's this?

giacomocavalieri: Oops I noticed this function seemed to do the exact same thing as another one and made a comment to not forget about it... seems like I forgot 💀
I'm fixing this shortly!

compiler-core/src/warning.rs
```diff
@@ -1258,19 +1259,19 @@ doesn't fit in that many {unit}. It would be truncated by taking its {taken}, re
                     }
                 }
 
-                type_::Warning::AssertLiteralValue { location } => Diagnostic {
+                type_::Warning::AssertLiteralBool { location } => Diagnostic {
                     title: "Assertion of a literal value".into(),
                     text: wrap(
-                        "Asserting that a literal value is redundant since you \
-can already tell whether it will be true or false.",
+                        "Asserting on a literal Boolean is redundant since you \
```
lpil:
```suggestion
                        "Asserting on a literal bool is redundant since you \
```

compiler-core/src/warning.rs
```diff
@@ -1258,19 +1259,19 @@ doesn't fit in that many {unit}. It would be truncated by taking its {taken}, re
...
+can already tell whether it will be True or False.",
```
lpil:
```suggestion
can already tell whether it will be `True` or `False`.",
```

If you want to use code then add backticks

lpil (commented):
Thank you!

I've pushed a failing test for a bug I found

giacomocavalieri:
Nice catch! Should be fixed now

PR #4692, opened 2025-06-17
Friendly multi spread

giacomocavalieri:
This PR picks up #4265. I've kept the original commits and quoted @carlbordum in the CHANGELOG, thank you for the initial work!!


compiler-core/src/error.rs
```diff
@@ -3731,21 +3729,1020 @@ and explain your usecase for this pattern, and how you would expect it to behave
...
+                                        span: location,
+                                    },
+                                    path: path.clone(),
+                                    src: src.clone(),
+                                    extra_labels: vec![],
+                                }),
+                            },
+                    };
+
+                vec![diagnostic]
```
lpil: There's a lot of duplication here, making it verbose and hard to see the bits that change. Please extract just the variable bits please 🙏

compiler-core/src/parse.rs
```diff
@@ -622,11 +622,40 @@ where
...
+                                .push(DeprecatedSyntaxWarning::DeprecatedListPrepend {
+                                    location: SrcSpan { start, end },
+                                });
+                        }
+
+                        // Give a better error when there is two consecutive spreads
+                        // like `[..wibble, ..wabble, woo]`. However, if there's other
+                        // elements after the tail of the list
+                        println!("{:#?}", self.tok0);
+                        println!("{:#?}", elements_after_tail);
```
lpil: Debugging here!

compiler-core/src/error.rs
giacomocavalieri: It's the same as what we do for the other type errors, should I do the same for those? I don't think we've never had a problem for those. I went with this approach and removed the code abstracting the changing bits because for this new error I needed to add some extra labels so everything would have been turned into a triple with a final empty list except for this new error:
```diff
- (title, lines)
+ (title, lines, [])
```
For some reason it felt worse than just building the diagnostic directly, but I can revert it and go for triples instead!

compiler-core/src/parse.rs
giacomocavalieri: Oops!

compiler-core/src/error.rs
giacomocavalieri: Also to play nicely with the abstracting bit of code we built hints by hand as a string mixed with the other error message instead of using the proper field of the `Diagnostic` struct:
```rs
let (text_lines, label_text) = (
  vec!["Hint: wibble wobble", "Error!", "See: ..."],
  "label"
);

Diagnostic {
  text: text_lines.join("\n"),
  hint: None,
  ..
}
```
Instead of what would make the best use of the diagnostic fields:
```rs
let (text_lines, hint, label_text) = (
  vec!["Error!", "See: ..."],
  Some("wibble wobble"),
  "label",
);

Diagnostic {
  text: text_lines.join("\n"),
  hint,
  ..
}
```

So to abstract the common bits we would need some data structure like this:
```rs
SyntaxErrorData {
  lines: Vec<EcoString>,
  hint: Option<EcoString>,
  label_text: EcoString,
  extra_labels: Vec<ExtraLabel>,
}
```
So my though process was that we might as well build the `Diagnostic` directly like we're doing for type errors

lpil: For me the goal is to be able to clearly see what each match clause does, so having a tuple or that struct would be greatly helpful! Right now I can't identify what the important bits are from reading unless I take notes, because the majority of the code is the same in each clause.

giacomocavalieri:
Should be better now!

compiler-core/src/error.rs
```diff
@@ -3764,20 +3762,22 @@ and explain your usecase for this pattern, and how you would expect it to behave
                     error.location
                 };
 
-                vec![Diagnostic {
-                    title: "Syntax error".into(),
+                let title = String::from("Syntax error");
+                let ParseErrorDetails{ text, label_text, extra_labels, hint  } = error.error.details();
+                vec![Diagnostic{
```
lpil, line 3767: Formatting broken in this file again I guess :(

PR #4695, opened 2025-06-18
Format error module

giacomocavalieri:
I managed to hunt down the reason why the formatter would give up on the file and fix it.
It was a combination of an inline comment and long lines.

### DIFF
--- a/compiler-core/src/type_/snapshots/gleam_core__type___tests__record_update_variant_inference_fails_for_several_possible_variants.snap
+++ b/compiler-core/src/type_/snapshots/gleam_core__type___tests__record_update_variant_inference_fails_for_several_possible_variants.snap
@@ -24,7 +24,6 @@ error: Unsafe record update
 10 │       Vector2(..vector, y: y +. increase)
    │                 ^^^^^^ I'm not sure this is always a `Vector2`
 
-
 This value cannot be used to build an updated `Vector2` as it could be some
 other variant.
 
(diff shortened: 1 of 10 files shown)


compiler-core/src/type_/snapshots/gleam_core__type___tests__record_update_variant_inference_fails_for_several_possible_variants.snap
```diff
@@ -24,7 +24,6 @@ error: Unsafe record update
 10 │       Vector2(..vector, y: y +. increase)
    │                 ^^^^^^ I'm not sure this is always a `Vector2`
 
-
```
giacomocavalieri, line 27: There was a stray empty line in the error message

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__recursive_var.snap
```diff
@@ -14,5 +14,4 @@ error: Recursive type
 
 I don't know how to work out what type this value has. It seems to be
 defined in terms of itself.
-
```
giacomocavalieri, line 17: I switched to using the `Diagnostic`'s `hint` field which got rid of the empty line. Now it's consistent with all the other hints we show in error messages

lpil (approved):
Thank you!

PR #4696, opened 2025-06-18
Fix extracting `use` statements

### LINKED ISSUE #4675: Extract variable on `use` generates garbled code
Gleam: 1.11.0
BEAM: Erlang/OTP 27 [erts-15.2.7] [source] [64-bit] [smp:32:32] [ds:32:32:10] [async-threads:1] [jit:ns]
Editor: helix 25.01.1 (ae0dd313)
OS: NixOS 25.05

```gleam
pub fn a_test() {
  let assert Ok("foo") =
    Ok({
      use <- bool.guard(1 == 2, "foo")
      "bar"
    })
}

```

Going to the `use` and running the `Extract variable` LSP action results in this mess:

```gleam
pub fnlet string = {
      use <- bool.guard(1 == 2, "foo")
      "bar"
    }
       a_test() {
  let assert Ok("foo") =
    Ok(string)
}

```

GearsDatapacks:
This PR fixes #4675 
There was some duplicated logic that worked slightly differently between the two implementations, causing this bug.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,6 +145,10 @@
   with the wrong number of labels.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the language server would generate invalid code when the
+  "Extract variable" code action was used on a `use` expression.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!

PR #4706, opened 2025-06-20
Fix Forgejo repo/view-source links

### LINKED ISSUE #4688: Issues with View Source links in docs for forgejo repositories
Gleam version: 1.11.1

Currently the links to view source in the generated docs always include the line number of the source code as a range of lines, i.e. `#L11-11` even though it is only referencing a single line of the source code, which should be `#L11`.
Github seems to be able to handle ranges that only reference a single line, so it's not a problem for github repos, but for forgejo a link to a single line cannot be a range, and the line will not be highlighted on the page at all if it is linked as such. So I would recommend either making single line source code references not use ranges across the board, or have a special case for forgejo repos so that ranges aren't used for single line references for that repo type.

Another issue I have noticed is that there is an extra forward slash in the generated urls to forgejo repositories between the host and user, so it ends up linking to https://host//user/repo instead of https://host/user/repo. This applies to both the "Repository" link on the left sidebar, and the "View Source" links. This one doesn't actually break anything, so it's more of a visual issue.

For an example of these behaviours, you can take a look at [this package](https://hexdocs.pm/spacetraders_sdk) I published, which links to a forgejo repo. For example the first "View Source" link on [this docs page](https://hexdocs.pm/spacetraders_sdk/endpoints/agents.html) takes you to [this link](https://git.7cs.dev//lily/gleam-spacetraders-sdk/src/tag/v1.0.0/src/endpoints/agents.gleam#L80-80), which is incorrect and should be [this](https://git.7cs.dev/lily/gleam-spacetraders-sdk/src/tag/v1.0.0/src/endpoints/agents.gleam#L80) instead.

aayush-tripathi:
Closes #4688 

This PR cleans up how the docs generator builds repository and View-Source URLs for Gitea-style hosts like Forgejo.

- Trim trailing slashes so sidebar “Repository” and "View Source" links never contain an extra forward slash between the host and user.
- Single line references emit `#Lx` instead of `#Lx-Lx` when `start_line==end_line`.
- Updated existing tests to expect `#Lx` instead of `#Lx-Lx`.
- Added a new test to assert no double-slash in Gitea/Forgejo repo URLs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,8 @@
 
   ([Amjad Mohamed](https://github.com/andho))
 
+- Docs generator now strips trailing slashes from Gitea/Forgejo hosts so sidebar "Repository" and "View Source" links never include `//`, and single-line “View Source” anchors emit `#Lx` instead of `#Lx-x`. ([Aayush Tripathi](https://github.com/aayush-tripathi))
+
 ### Language server
 
 ### Formatter
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -827,7 +827,11 @@ impl Repository {
             }
             Repository::Gitea {
                 repo, user, host, ..
-            } => Some(format!("{host}/{user}/{repo}")),
+            } => {
+                let string_host = host.to_string();
+                let cleaned_host = string_host.trim_end_matches('/');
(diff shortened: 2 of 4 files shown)


lpil (approved):
Thank you very much!

PR #4713, opened 2025-06-23
Fix compiler crash for UTF-codepoint segments on the JavaScript target

### LINKED ISSUE #4700: Compiler bug when compiling a `utf8_codepoint` case to the Javascript target
# Error message
```
Panic: compiler-core/src/javascript/decision.rs:1276
        invalid slice type made it to code generation: UtfCodepoint
Gleam version: 1.11.1
Operating system: linux
```

# Repro
Set the `target` in `gleam.toml`:
```toml
target = "javascript"
```

Minimal code repro:
```gleam
pub fn main() {
  let bits = <<>>
  case bits {
    <<_:utf8_codepoint, _:bits>> -> todo
    _ -> todo
  }
}
```

GearsDatapacks:
Fixes #4700 
I've moved the checking for invalid target-specific segments to the `bit_array` file, so they are done as part of type-checking not code generation. This prevents the crash also.

### DIFF
--- a/compiler-wasm/src/tests.rs
+++ b/compiler-wasm/src/tests.rs
@@ -90,7 +90,7 @@ pub fn main() { wibble() }
     assert!(
         compile_package(0, "javascript")
             .unwrap_err()
-            .contains("Unsupported feature for compilation target")
+            .contains("The javascript target does not support")
     );
 }
 


lpil (approved):
Thank you

PR #4715, opened 2025-06-23
Only show relevant labels for "incorrect arity" error message

### LINKED ISSUE #3884: Confusing incorrect arity hint when labelled arguments are provided without labels
```gleam
fn wibble(a a: Int, b b: Float, c c: String) {
  todo
}

pub fn wobble() {
  wibble(1, 2.0)
}
```

produces

```txt
error: Incorrect arity
  ┌─ /home/arkan/Projects/oss/gleam/probe/src/probe.gleam:6:3
  │
6 │   wibble(1, 2)
  │   ^^^^^^^^^^^^ Expected 3 arguments, got 2

This call accepts these additional labelled arguments:

  - a
  - b
  - c
```

I think the hint (_"this call accepts these additional labelled arguments: ..."_) is confusing here, because I already provide values for `a` and `b`, but using positional arguments instead. While it is technically correct that I could provide any of the arguments as a labelled argument, changing the error from an _incorrect arity_ to a _type mismatch_ in the process, I don't think it's very useful to tell me that I could change the meaning of the arguments I already pass.

I chose different types to make it more obvious, but in my opinion all arguments that I already provide should not be included in the hint, even if the types were compatible. This is probably even more subjective. ~ :purple_heart:

giacomocavalieri:
This PR fixes #3884
There was a newer, more accurate function to figure out missing labels; it was as simple as using it instead of that older version 😁
And now the nice thing is it's consistent with the labels suggested by the "add missing labels" code action, win-win!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,36 @@
   ([Carl Bordum Hansen](https://github.com/carlbordum)) and
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The error message one gets when calling a function with the wrong number of
+  arguments has been improved and now only suggests the relevant missing labels.
+  For example, this piece of code:
+
+  ```gleam
+  pub type Pokemon {
+    Pokemon(id: Int, name: String, moves: List(String))
+  }
+
+  pub fn best_pokemon() {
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you

PR #4721, opened 2025-06-24
Trigger "add type signature" code action anywhere between the start of a function and the start of its body

### LINKED ISSUE #4717: Allow "add type signature" in a larger span
One little annoyance I keep running into is I wish the "add type annotation" were a bit more permissive with the span it can be triggered in.
At the moment one can trigger it only if the cursor is exactly over the function name or arguments list. But one thing I usually end up doing is this:
```gleam
pub fn function(a) {
//                  ^ Put my cursor here and try and trigger it
}
```
It won't show up so I have to tap back a few columns and then I'm able to trigger the code action, I think it would be nice if we triggered the code action even if the cursor is over the open `{`:

```gleam
pub fn function(a) {
//^^^^^^^^^^^^^^^^^^ Code action triggers over this entire area
//^^^^^^^^^^^^^^^^   Instead of just here!
```

giacomocavalieri:
This PR fixes #4717


compiler-core/src/ast.rs
```diff
@@ -683,6 +683,7 @@ impl Publicity {
 /// ```
 pub struct Function<T, Expr> {
     pub location: SrcSpan,
+    pub body_start: Option<u32>,
```
lpil, line 687: Isn't this redundant? The same value as `function.body.first().location().start`

giacomocavalieri: Nope they're two different things:
```gleam
pub fn wibble() {
//              ^ body.start
   

   1 + 1
// ^ function.body.first().location().start
}
```
If we were to use that, it'd mean the action would trigger in unexpected places inside the body of the function:
```gleam
pub fn wibble() {
    let a = 1
//^^ cursor here, [add type signature] would pop up
}
```

lpil: Ah, so it's head end rather than body start. Let's use the body start and not have this new field, to make it even more permissive.

giacomocavalieri: I'm not sure I really like the idea, it being too permissive could be confusing. If I were to put my cursor on the same line of this let binding I'd expect the code action to annotate the `let`, not the function:
```
pub fn wibble() {
      let a = 1
// ^^^^^^^^^^^^ On the same line of the let binding I would be confused if the
//              code action annotated the function instead of the let
}
```
I'm using the `{` because that way it triggers in the entire line(s) spanned by the function's head, which is more intuitive to me that triggering it elsewhere

giacomocavalieri:
I've fixed the conflict!

lpil (approved):
Thank you!!

PR #4723, opened 2025-06-24
Suggest unlocking locked deps

giacomocavalieri:


giacomocavalieri:
I just rebased the code but I don't understand this feature enough to update the changelog!

lpil (commented):
I've given this a test and the output isn't right. It doesn't print the error, but it does print a (malformatted) message that refers to the previous error.

```sh
louis ~/Desktop/thingy (main ?) $ gleam add gleam_http@2
# ...
louis ~/Desktop/thingy (main ?) $ gleam add gleam_httpc@4
  Resolving versions

Some of these dependencies are locked to specific versions. It may
            be possible to find a solution if they are unlocked, would you like
            to unlock and try again? [y/n]:
```

Closes https://github.com/gleam-lang/gleam/issues/3622
Closes https://github.com/gleam-lang/gleam/pull/3970

PR #4724, opened 2025-06-25
Sign Windows releases with Azure Trusted Signing

maennchen:
This PR adds Authenticode signing for `gleam.exe` in both tagged releases and nightly builds. A signed binary installs without SmartScreen warnings and gives users a clear way to verify integrity and origin.

## Rationale

* **User trust and installation flow** – removes “unknown publisher” dialogs and allows independent signature checks.
* **Supply-chain hardening** – helps detect tampering between CI and end users.
* **Parity with Elixir** – uses the same signing approach already in place for Elixir Windows artefacts.
* **No long-lived secrets** – the workflow relies on GitHub → Azure Workload Identity Federation, so there are no passwords or certs stored in the repo and no secret rotation.

The infrastructure is paid for and managed by the Erlang Ecosystem Foundation.

## Implementation details

1. `azure/login@v2` exchanges the GitHub OIDC token for an Azure token.
2. `azure/trusted-signing-action@v0.5.1` signs `gleam.exe` with the *Gleam* certificate profile.
3. The action adds an RFC 3161 timestamp, keeping the signature valid after certificate expiry.

## One-time project setup

1. **Settings → Environments → New environment**
2. Name it **`release`**.
3. Under *Deployment branches and tags* add `main` and `v*`.
4. Add environment **secrets** (values sent privately):

   * `AZURE_CLIENT_ID`
   * `AZURE_SUBSCRIPTION_ID`
   * `AZURE_TENANT_ID`
5. Add environment **variables**:

   * `AZURE_CERTIFICATE_PROFILE_NAME` – `Gleam`
   * `AZURE_TRUSTED_SIGNING_ACCOUNT_NAME` – `trusted-signing-gleam`

After this, every push to `main` or any `v*` tag will ship a signed Windows binary.

## Verification

Signature can be checked with [osslsigncode](https://github.com/mtrojnar/osslsigncode):

```bash
$ osslsigncode verify ./gleam.exe
```

## References
* A successful CI run that produced signed artefacts:
https://github.com/maennchen/gleam/actions/runs/15872878793
* Aligned Elixir Implementation: https://github.com/elixir-lang/elixir/pull/14604

### DIFF
--- a/.github/actions/build-release/action.yml
+++ b/.github/actions/build-release/action.yml
@@ -17,6 +17,21 @@ inputs:
     description: "Expected Binary Architecture"
     required: false
     default: ""
+  azure-tenant-id:
+    description: "Azure Tenant ID for Windows Code Signing"
+    required: false
+  azure-subscription-id:
+    description: "Azure Subscription ID for Windows Code Signing"
+    required: false
+  azure-client-id:
+    description: "Azure Client ID for Windows Code Signing"
+    required: false
+  azure-trusted-signing-account-name:
+    description: "Azure Trusted Signing Account Name for Windows Code Signing"
+    required: false
+  azure-certificate-profile-name:
+    description: "Azure Certificate Profile Name for Windows Code Signing"
+    required: false
 
 outputs:
   archive:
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!!

PR #4727, opened 2025-06-26
Implement `echo ... as ...`

### LINKED ISSUE #4714: Add custom messages to `echo as`
Hello! It would be really nice if we were able to add an additional message to `echo` the same way we can with `panic` and `todo` to make it easier to grep for specific debug messages and add additional context to make it easier to debug our code. It would look like this:
```gleam
echo wibble as "some context"

[1, 2, 3]
|> echo as "step 1"
|> list.map(wobble)
|> echo as "step 2"
```

But implementing this there's a couple of considerations, so I'd like to discuss it here first before diving into the implementation :)

1. What should this look like once it's printed? I'm not sure of the best look here, where should we put the additional message? Maybe next to the file/line number but not faded
2. What should the precedence be when it is mixed with other expressions that can be followed by `as`:
    1.`assert 1 == echo wobble as "message"` should this be the same as `{ assert 1 == echo wobble } as "message"` or `assert 1 == { echo wobble as "message" }`
    2. When both have a message it should be clear (albeit a bit ugly) `assert 1 == echo wobble as "one" as "other"` should be `{ assert 1 == echo wobble as "one" } as "other"`
    3. `echo panic as "wibble"` (or the same with `todo`), here I think it should always be `echo { panic as "wibble" }` since it's going to panic anyway and the message wouldn't be displayed otherwise.
    4. `let assert Ok(1) = echo wobble as "message"` should this be `{ let assert Ok(1) = echo wobble } as "message"` or `let assert Ok(1) == { echo wobble } as "message"`

If we have a look at what happens today with `assert` and `panic` it seems like the inner one always takes precedence, which sounds reasonable!
```gleam
let assert Ok(1) = panic as "wibble"
// Same as `let assert Ok(1) = { panic as "wibble" }`

assert 1 == panic as "wibble"
// Same as `assert 1 == { panic as "wibble" }`

assert 1 == panic as "wibble" as "wobble"
// Same as `assert { 1 == { panic as "wibble" } } as wobble`
```
Should we do the same for `echo`?

giacomocavalieri:
This PR closes #4714

I'd recommend going through the PR commit by commit, I tried keeping the git history as tidy and linear as possible!

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -499,6 +499,47 @@ fn remove_echo_with_message() {
 }
 
 #[test]
+fn remove_echo_with_message_and_comment() {
+    assert_code_action!(
+        REMOVE_ALL_ECHOS_FROM_THIS_MODULE,
+        r#"pub fn main() {
+  echo 1 + 2
+    // Hello!
+    as "message"
+}"#,
+        find_position_of("echo").to_selection()
+    );
+}
+
+#[test]
+fn remove_echo_with_message_and_comment_2() {
+    assert_code_action!(
+        REMOVE_ALL_ECHOS_FROM_THIS_MODULE,
+        r#"pub fn main() {
+  echo 1 + 2 as
+    // Hello!
+    "message"
+}"#,
+        find_position_of("echo").to_selection()
+    );
+}
+
+#[test]
+fn remove_echo_with_message_and_comment_3() {
+    assert_code_action!(
+        REMOVE_ALL_ECHOS_FROM_THIS_MODULE,
+        r#"pub fn main() {
+  echo 1 + 2 as
+    // Hello!
+    "message"
+
+  Nil
+}"#,
+        find_position_of("echo").to_selection()
+    );
+}
+
+#[test]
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -6817,6 +6818,17 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveEchos<'ast> {
...
 
+        // We also want to remove the echo message!
+        if message.is_some() {
+            let start = expression
+                .as_ref()
+                .map(|expression| expression.location().end)
+                .unwrap_or(location.start + 4);
+
+            self.echo_spans_to_delete
+                .push(SrcSpan::new(start, location.end));
```
lpil, line 6829: What if there's a comment?

```gleam
echo something
// Hello
as "message"
```

lpil (commented):
Looks fab! I left one question

compiler-core/src/language_server/code_action.rs
giacomocavalieri: It's always removed, I've added a couple of tests!

lpil (approved):
Thank you!

PR #4731, opened 2025-06-27
Enable inlining on the Erlang target

giacomocavalieri:
Now all generated modules include a `inline` compile option to enable the Erlang compiler to perform inlining. We're using the default option which should be pretty conservative when it comes to code size. Despite this, there's some really nice improvements, basically for free:
- With Wisp's benchmark:
  ```
  CORES  INLINING     REQ/S  MB/S
      1       yes  25835.33  2.44  ~15% faster
      1        no  21861.67  2.06 

      4       yes  79477.03  7.50  < 1% difference
      4        no  79117.93  7.47
  ```
- With Lustre's dom diffing benchmark:
  ```
  NO INLINING                              IPS        MIN     P99
  1000 rows                   vdom.diff()  3051.1898  0.2932  0.3984
  1000 rows (shuffled)        vdom.diff()  2997.1366  0.3053  0.4041
  1000 rows (keyed)           vdom.diff()  2690.2535  0.2192  0.4883
  1000 rows (keyed, shuffled) vdom.diff()  2063.8774  0.4704  0.5583
  
  WITH INLINING
  1000 rows                   vdom.diff()  6131.1367  0.1524  0.2293
  1000 rows (shuffled)        vdom.diff()  5977.3581  0.1562  0.2336
  1000 rows (keyed)           vdom.diff()  4669.7401  0.2060  0.2722
  1000 rows (keyed, shuffled) vdom.diff()  2155.2762  0.4480  0.5371
  ```
  Here the results are quite flashy, it's important to note that Lustre is not really representative of most real Gleam packages: here what's getting optimised are mainly all the calls to the `html` functions. Quite impressive nonetheless, server components are about to get way faster!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,6 +226,10 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- On the Erlang target each generated module enables inlining from the Erlang
+  compiler.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
 - `gleam update`, `gleam deps update`, and `gleam deps download` will now print
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -272,7 +272,7 @@ fn module_document<'a>(
 
     let module = docvec![
         header,
-        "-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch]).",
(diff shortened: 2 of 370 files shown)


lpil:
Broken tests!

Do you mean Mist benchmarks? Or are there Wisp ones I'm not aware of?

giacomocavalieri:
Oops yeah I meant Mist, I will fix the tests shortly!

giacomocavalieri:
Fixed the conflict!

PR #4734, opened 2025-06-28
Fix inconsistency in bit array size options on the JavaScript target

GearsDatapacks:
I noticed that a few very specific combinations of bit array options would not compile on the JavaScript target, while they would on the Erlang target. There was no specific reason for them not to compile, it was just an oversight in the codegen logic.
For example, the following code would compile on both targets:
```gleam
let x = 10
<<x:size(2)-unit(8)>>
```
However, this code would only compile on the Erlang target:
let x = <<10, 20>>
<<x:size(2)-unit(8)>>
```

This PR fixes that inconsistency. I also took this time to rework the logic for bit array generation, as I found matching on specific combinations to be clunky and error prone. Instead, we collect all the information first, then match on the relevant information to generate the code.


lpil (approved):
Thank you!! Fantastic work, as always!

PR #4735, opened 2025-06-28
Change "generate function" to be a quickfix

GearsDatapacks:
As suggested [on discord](https://discord.com/channels/768594524158427167/1388635403183194305/1388635403183194305), I have changed the "generate function" code action to be a quickfix, and also made it preferred by default, which makes it easier to apply to erroring code.
edit: I have also applied this change to the "generate variant" code action, which is in a similar position


lpil (approved):
Thank you!

Issue #4716, opened 2025-06-24
Allow to rename a module in a `ModuleSelect`


giacomocavalieri:
I was writing some lustre code:
```gleam
html.form(
  [
    attribute.method("POST"),
    attribute.action("..."),
    attribute.enctype("..."),
  ],
  [...],
)
```
And thought I would like to change the `attribute` module and alias it to some other name (like `attr`). It would be really nice if triggering the rename action allowed us to alias (or change an existing alias) an imported module and automatically update all relevant module selects:
```diff
- import lustre/attribute
+ import lustre/attribute as attr

html.form(
  [
-   attribute.method("POST"),
-   attribute.action("..."),
-   attribute.enctype("..."),
+   attr.method("POST"),
+   attr.action("..."),
+   attr.enctype("..."),
  ],
  [...],
)
```

What do you think?

GearsDatapacks:
That makes sense to me. We already do that for unqualified imports

lpil:
I dig it!

llpaull:
I'd like to give this a go!

vshakitskiy:
Can I pick up on this feature and give it a try? I saw that it's not being worked on for some time now. @lpil @llpaull

lpil:
Yes please!

PR #4750, opened 2025-07-02
Allow renaming module from `ModuleSelect`

llpaull:
This PR addresses #4716

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1925,7 +1925,7 @@ impl TypedClauseGuard {
                 module_alias,
                 ..
             } => Some(Located::ModuleName {
-                location: *location,
+                location: SrcSpan::new(location.start, location.start + (module_alias.len() as u32)),
                 module_name: module_name.clone(),
                 module_alias: module_alias.clone(),
                 layer: Layer::Value,
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -227,12 +227,13 @@ impl TypedExpr {
                     end: location.end,
                 };
 
-                // We subtract 1 so the location doesn't include the `.` character.
-                let module_span = SrcSpan::new(location.start, field_start - 1);
+                let module_span = SrcSpan::new(location.start, location.start + (module_alias.len() as u32));
 
                 if field_span.contains(byte_index) {
                     Some(self.into())
-                } else if module_span.contains(byte_index) {
+                }
+                // Allow leniency to have cursor left of the '.' instead of directly on top of module name
+                else if SrcSpan::new(location.start, field_start-1).contains(byte_index) {
                     Some(Located::ModuleName {
                         location: module_span,
                         module_name: module_name.clone(),
--- a/compiler-core/src/language_server/reference.rs
+++ b/compiler-core/src/language_server/reference.rs
@@ -286,7 +286,7 @@ pub fn reference_for_ast_node(
                 },
                 None => Referenced::ModuleName {
                     module_name: import.module.clone(),
-                    module_alias: import.module.clone(),
+                    module_alias: import.module.split('/').next_back().map(EcoString::from).unwrap(),
                     location: import.module_location,
                 }
             })
@@ -568,7 +568,7 @@ impl<'ast> Visit<'ast> for FindModuleNameReferences<'_> {
     ) {
         if module_alias == self.module_alias {
             self.references.push(ModuleNameReference {
-                location: *location,
+                location: SrcSpan::new(location.start, location.start + (module_alias.len() as u32)),
                 kind: ModuleNameReferenceKind::ModuleSelect,
             });
         }
@@ -596,8 +596,7 @@ impl<'ast> Visit<'ast> for FindModuleNameReferences<'_> {
     ) {
         if module_alias == self.module_alias {
             self.references.push(ModuleNameReference {
-                // Subtract 1 from field_start to exclude '.' from location
-                location: SrcSpan::new(location.start, field_start - 1),
+                location: SrcSpan::new(location.start, location.start + (module_alias.len() as u32)),
                 kind: ModuleNameReferenceKind::ModuleSelect,
             });
         }
--- a/compiler-core/src/language_server/tests/rename.rs
+++ b/compiler-core/src/language_server/tests/rename.rs
@@ -1393,7 +1393,7 @@ fn rename_module_from_import_with_alias() {
 #[test]
 fn rename_module_from_import_with_unqualified_member() {
     assert_rename!(
-        TestProject::for_source("import     mod .{   type Type   }").add_module("mod", ""),
+        TestProject::for_source("import     mod .  {   type Type   }").add_module("mod", ""),
         "module",
         find_position_of("mod"),
     );
@@ -1402,27 +1402,49 @@ fn rename_module_from_import_with_unqualified_member() {
 #[test]
 fn rename_module_from_import_with_unqualified_member_and_alias() {
     assert_rename!(
-        TestProject::for_source("import     mod.{ type Type  }   as      module").add_module("mod", ""),
+        TestProject::for_source("import     mod  . { type Type  }   as      module").add_module("mod", ""),
         "alias",
         find_position_of("module"),
     );
 }
 
 #[test]
-fn rename_module_from_import_nested() {
+fn rename_module_from_import_namespaced() {
+    let src = "
+import     testing     /   mod
+
+pub fn main() {
+  echo mod.str
+}
+";
     assert_rename!(
-        TestProject::for_source("import     testing/mod     ").add_module("testing/mod", ""),
+        TestProject::for_source(src).add_module("testing/mod", "pub const str = \"string\""),
         "module",
-        find_position_of("testing/mod")
+        find_position_of("testing     /   mod")
+    );
+}
+
+#[test]
+fn rename_module_from_import_namespaced_with_unqualified_member_and_alias() {
+    let src = "import     testing  / mod . {   str   ,  type    Type  }   as     module";
+    let mod_src =
+"
+pub const str = \"string\"
+pub type Type { Variant }
+";
+    assert_rename!(
+        TestProject::for_source(src).add_module("testing/mod", mod_src),
+        "alias",
+        find_position_of("module")
     );
 }
 
 #[test]
 fn rename_module_from_import_with_alias_to_orig() {
     assert_rename!(
-        TestProject::for_source("import mod as module").add_module("mod", ""),
+        TestProject::for_source("import     mod     as   module").add_module("mod", ""),
         "mod",
-        find_position_of("import mod as module")
+        find_position_of("import")
     );
 }
 
@@ -1432,13 +1454,13 @@ fn rename_module_from_variant_in_expression() {
 import mod
 
 pub fn main() {
-  echo mod.Variant
+  echo mod   .  Variant
 }
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant }"),
         "module",
-        find_position_of("mod.Variant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1448,13 +1470,13 @@ fn rename_module_from_constant_in_expression() {
 import mod
 
 pub fn main() {
-  echo mod.constant
+  echo mod  . constant
 }
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub const constant = 10"),
         "module",
-        find_position_of("mod.constant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1463,12 +1485,12 @@ fn rename_module_from_variant_in_const() {
     let src = "
 import mod
 
-const c = mod.Variant
+const c = mod       .   Variant
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant }"),
         "module",
-        find_position_of("mod.Variant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1477,12 +1499,12 @@ fn rename_module_from_constant_in_const() {
     let src = "
 import mod
 
-const c = mod.constant
+const c = mod   .   constant
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub const constant = 10"),
         "module",
-        find_position_of("mod.constant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1493,7 +1515,7 @@ import mod
 
 fn func(arg) {
   case arg {
-    mod.Variant1 -> todo
+    mod .   Variant1 -> todo
     mod.Variant2 -> todo
   }
 }
@@ -1501,7 +1523,7 @@ fn func(arg) {
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant1 Variant2 }"),
         "module",
-        find_position_of("mod.Variant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1510,9 +1532,9 @@ fn rename_module_from_variant_in_clause_guard() {
     let src = "
 import mod
 
-fn func(arg: List(mod.Type)) {
+fn func(arg) {
   case arg {
-    [x, ..rest] if x == mod.Variant1 -> 1 + func(rest)
+    [x, ..rest] if x == mod .   Variant1 -> 1 + func(rest)
     [_, ..rest] -> func(rest)
     [] -> 0
   }
@@ -1521,7 +1543,7 @@ fn func(arg: List(mod.Type)) {
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant1 Variant2 }"),
         "module",
-        find_position_of("mod.Variant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1532,7 +1554,7 @@ import mod
 
 fn func(arg: List(Int)) {
   case arg {
-    [x, ..rest] if x == mod.constant -> 1 + func(rest)
+    [x, ..rest] if x == mod .   constant -> 1 + func(rest)
     [_, ..rest] -> func(rest)
     [] -> 0
   }
@@ -1541,7 +1563,7 @@ fn func(arg: List(Int)) {
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub const constant = 10"),
         "module",
-        find_position_of("mod.constant")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1550,12 +1572,12 @@ fn rename_module_from_type_in_custom_type() {
     let src = "
 import mod
 
-type Custom { Var(mod.Type) }
+type Custom { Var(mod  . Type) }
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant }"),
         "module",
-        find_position_of("mod.Type")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1564,12 +1586,12 @@ fn rename_module_from_type_in_type_alias() {
     let src = "
 import mod
 
-type Alias = mod.Type
+type Alias = mod    .  Type
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant }"),
         "module",
-        find_position_of("mod.Type")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
@@ -1578,12 +1600,12 @@ fn rename_module_from_type_in_annotation() {
     let src = "
 import mod
 
-const c: mod.Type = mod.Variant
+const c: mod    . Type = mod.Variant
 ";
     assert_rename!(
         TestProject::for_source(src).add_module("mod", "pub type Type { Variant }"),
         "module",
-        find_position_of("mod.Type")
+        find_position_of("mod").nth_occurrence(2)
     );
 }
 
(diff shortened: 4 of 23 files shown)


compiler-core/src/language_server/reference.rs
```diff
@@ -267,6 +271,32 @@ pub fn reference_for_ast_node(
...
+                location,
+                module,
+                as_name,
+                ..
+            },
+        )) => match as_name {
+            Some(alias) => Some(Referenced::ModuleName {
+                name: module.clone(),
+                // alias location includes the "as " part
+                location: SrcSpan::new(alias.1.start + 3, alias.1.end),
```
lpil: There could be any number of spaces around the `as` so we cannot safely add 3 to it here

compiler-core/src/language_server/reference.rs
```diff
@@ -267,6 +271,32 @@ pub fn reference_for_ast_node(
...
+            }),
+            None => {
+                // location includes entire line, only want the final module name
+                let used = import.used_name()?;
```
lpil: This function will pattern match on the `as_name` again, repeating the work. It'd be nice not to do that. Perhaps extract the name splitting logic into a new function that can be used.

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +482,84 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+pub struct ModuleNameReference {
+    pub location: SrcSpan,
+    pub kind: ModuleNameReferenceKind,
+}
+
+pub enum ModuleNameReferenceKind {
+    Name,
+    Import,
+    Alias,
+}
```
lpil, line 525: Could you add a comment to document what each of these variants mean please 🙏

compiler-core/src/language_server/tests/rename.rs
```diff
@@ -1350,3 +1350,169 @@ pub fn main() {
...
 }
+
+#[test]
+fn rename_module_alias_from_import() {
+    assert_rename!(
+        (
+            "mod",
+            "
+pub fn function() {
+    function()
```
lpil: 2 space indentation for Gleam code please 🙏 

Could you also use a name other than `function`? GitHub's syntax highlighting gets a bit confused by it.

Might be worth extracting this repeated string to a `const` and making it a bit smaller, something like this:

```rust
const some_name = "pub fn fn1() { todo }
pub fn fn2() { todo }
pub const const1 = Nil
pub type Type { Variant }"
```

This would make it easier to read the tests and the snapshot output as there's less boilerplate to skim over.

compiler-core/src/language_server/tests/rename.rs
```diff
@@ -1350,3 +1350,169 @@ pub fn main() {
...
+}
+
+pub fn other_function() {
+    function()
+}
+"),
+        "Module",
+        find_position_of("mod")
+    );
+}
```
lpil, line 1658: Fab! Could you add tests for these:

- More space around `as` for a module alias on an import
- More space around `as` and `}` for a module alias on an import with unqualified members
- Modules in types
- Modules variants in patterns
- Module constants in clause guards
- Module variants in clause guards
- Module constants in constants
- Module variants in constants
- Module constants in expressions
- Module variants in expressions
- Record access with the same name as a module

And any other edge cases you can think of. We want to get full coverage!

CHANGELOG.md
```diff
@@ -252,6 +252,41 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- It is now possible to rename a module from a ModuleSelect. For example:
```
lpil: `ModuleSelect` is a name internal to the compiler only. Instead describe the change in terms that a Gleam programmer will understsand.

lpil (commented):
Brilliant, thank you. This will be really nice.

I've left some notes inline, thank you

compiler-core/src/language_server/reference.rs
llpaull: do you want the spaces around `as` to be kept when renaming, so only change the actual alias, or do you want (parts of) them to be removed?

lpil: Either seems good to me!

lpil:
Hello! Just figuring out what is going to go into what release. Are you still working on this? :)

llpaull:
Yes I am! As for a progress update, most tests are written but still need to write some for guard clause and record access with same name as module. From what I can tell and have tested, all references to the module (alias) can be found and are appropriately renamed but it can't yet be renamed from every reference.

compiler-core/src/ast.rs
```diff
@@ -723,7 +724,7 @@ pub type UntypedImport = Import<()>;
 pub struct Import<PackageName> {
     pub documentation: Option<EcoString>,
     pub location: SrcSpan,
-    pub module: EcoString,
+    pub module: (EcoString, SrcSpan),
```
lpil: I would prefer to have a `module` (or `module_name`) and a `module_location` field, so both are named, to make it a little clearer. 🙏

compiler-core/src/ast/constant.rs
```diff
@@ -97,16 +97,42 @@ impl TypedConstant {
...
+                    type_::ValueConstructorVariant::ModuleConstant { module, .. }
+                    | type_::ValueConstructorVariant::ModuleFn { module, .. }
+                    | type_::ValueConstructorVariant::Record { module, .. } => {
+                        Located::ModuleName {
+                            location: *location,
+                            name: module,
+                            layer: Layer::Value,
+                        }
+                    }
+                    // unreachable?
```
lpil: WIP comment left here

compiler-core/src/ast/typed.rs
```diff
@@ -334,9 +334,22 @@ impl TypedExpr {
...
+            Self::Fn {
+                body,
+                args,
+                return_annotation,
+                type_,
+                ..
+            } => args
                 .iter()
                 .find_map(|arg| arg.find_node(byte_index))
+                .or_else(|| return_annotation.as_ref().and_then(|ret| {
```
lpil: Avoid abbreviations please 🙏 
```suggestion
                .or_else(|| return_annotation.as_ref().and_then(|annotation| {
```

compiler-core/src/language_server/engine.rs
```diff
@@ -194,6 +195,7 @@ where
                 Some(location) => location,
                 None => return Ok(None),
             };
+            eprintln!("{node:?}");
```
lpil: Debug statement here!

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
     }
 }
+
+fn get_module_name_span<T>(import: &Import<T>) -> SrcSpan {
+    match &import.as_name {
+        Some((AssignName::Variable(name), location)) => {
+            SrcSpan::new(location.end - (name.len() as u32), location.end)
+        }
+
+        Some((AssignName::Discard(_), location)) => SrcSpan::new(location.end - 1, location.end),
```
lpil: I think this will fail when the discard is not a single character

```gleam
import thingy as _unused
```

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
         ast::visit::visit_typed_call_arg(self, arg);
     }
 }
+
+fn get_module_name_span<T>(import: &Import<T>) -> SrcSpan {
```
lpil: Document what this does and what it is used for please 🙏 

I'm a bit confused as to why you'd want the location of just the final bit of the name in the case there there is no `as_name`. Why is that?

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+        Some((AssignName::Discard(_), location)) => SrcSpan::new(location.end - 1, location.end),
+        None => {
+            let used_name = import.module.0.split('/').next_back();
+            let len = used_name.map(|str| str.len()).unwrap_or(0) as u32;
+            SrcSpan::new(import.module.1.end - len, import.module.1.end)
+        }
+    }
+}
+
+pub fn find_module_name_references(
```
lpil, line 482: Document what this does please

compiler-core/src/language_server/reference.rs
```diff
@@ -267,6 +273,16 @@ pub fn reference_for_ast_node(
             location: *name_location,
             target_kind: RenameTarget::Definition,
         }),
+        Located::ModuleName { location, name, .. } => Some(Referenced::ModuleName {
+            name: name.clone(),
+            location,
+        }),
+        Located::ModuleStatement(Definition::Import(import @ Import { .. })) => {
```
lpil:
```suggestion
        Located::ModuleStatement(Definition::Import(import)) => {
```
I think this is redundant?

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+    //
+    // import lustre/attribute.{type ...}
+    // __________________________________
+    Import,
+
+    // Import of module with alias
+    // location is everything starting from " as" until the alias
+    //
+    // import lustre/attribute   as attr
+    //                          ________
```
lpil: Please don't include whitespace in the location 🙏

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+    references: Vec<ModuleNameReference>,
+    module_name: EcoString,
+}
+
+impl<'ast> Visit<'ast> for FindModuleNameReferences {
+    fn visit_typed_import(&mut self, import: &'ast ast::TypedImport) {
+        if import.module.0 == self.module_name {
+            match &import.as_name {
+                Some(alias) => self.references.push(ModuleNameReference {
+                    location: SrcSpan::new(alias.1.start - 1, alias.1.end),
```
lpil:
```suggestion
                    location: alias.1,
```

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+        field_start: &'ast u32,
+        type_: &'ast std::sync::Arc<Type>,
+        label: &'ast EcoString,
+        module_name: &'ast EcoString,
+        module_alias: &'ast EcoString,
+        constructor: &'ast ModuleValueConstructor,
+    ) {
+        if *module_name == self.module_name {
+            self.references.push(ModuleNameReference {
+                location: SrcSpan::new(location.start, field_start - 1),
```
lpil: I think the -1 is to remove the `.` from `module.field`, is that right? Could you drop a comment explaining the reason please 🙏

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+    fn visit_typed_constant(&mut self, constant: &'ast TypedConstant) {
+        match constant {
+            Constant::Record {
+                module,
+                record_constructor: con,
+                ..
+            }
+            | Constant::Var {
+                module,
+                constructor: con,
```
lpil:
```suggestion
                constructor:,
```

Avoid abbreviations please

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                                    kind: ModuleNameReferenceKind::Name,
+                                });
+                            }
+                        }
+                        arguments
+                            .iter()
+                            .zip(args.iter())
+                            .for_each(|(arg, type_)| self.visit_type_ast(arg, type_));
+                    }
+                    // Lists store their arguments' type as a var link
```
lpil: What does this mean?

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                    Type::Var { type_: var_type } => {
+                        if let TypeVar::Link { type_ } = var_type.borrow().clone() {
```
lpil: Please handle all the other type variants please! Not just links. There's also no guarantee that the link depth will be just 1, it can commonly be many links deep.

There's a function called something like `collapse_type_links` which could be useful here.

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                            self.visit_type_ast(type_ast, &type_);
+                        }
+                    }
+                    _ => {}
```
lpil: Never use catch all pattern please! Using them commonly results in bugs sneaking in as any changes to the type silently go unhandled, and because it's easy to forget about other cases when writing the code the first time.

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+            TypeAst::Tuple(TypeAstTuple { elements, .. }) => {
+                if let Type::Tuple {
+                    elements: elements_type,
+                } = type_
+                {
+                    elements
+                        .iter()
+                        .zip(elements_type.iter())
+                        .for_each(|(e, type_)| self.visit_type_ast(e, type_));
+                }
```
lpil: If the tuple AST must always have a tuple type then panic in the case that it does not instead of silently skipping over the malformed code. We want to know when there is a bug.

Could it be linked here? Do the links need to be collapsed?

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +468,339 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                } = type_
+                {
+                    arguments
+                        .iter()
+                        .zip(args.iter())
+                        .for_each(|(arg, type_)| self.visit_type_ast(arg, type_));
+                    self.visit_type_ast(return_, ret_type);
+                }
+            }
+            _ => {}
```
lpil: Same as above- no catch alls, panic for impossible invariants, and what about links?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__rename__rename_module_from_local_var_annotation_fn_recursive-4.snap
```diff
@@ -0,0 +1,31 @@
...
+pub fn fn1() { Nil }
+pub const const1 = 5
+pub const const2 = <<0:8>>
+pub const const3 = "Hello"
+pub type Type { Variant1 Variant2(Int) }
+pub type GenericType(inner) { Node(inner, GenericType(inner)) Leaf }
+
+-- app.gleam
+import mod
+fn func() { let _: fn(mod.GenericType(mod.Type)) -> mod.GenericType(mod.Type) = fn(arg: mod.GenericType(mod.Type)) -> mod.GenericType(mod.Type) { arg } }
```
lpil: These snapshots with really long unformatted lines are hard to read and it's not immediately clear what bit is being focused. Please use more normal formatting.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__rename__rename_module_from_import_with_alias.snap
```diff
@@ -0,0 +1,173 @@
...
+const c4 = alias.fn1
+const c5: #(alias.Type, Int, fn() -> Nil) = #(alias.Variant1, alias.const1, alias.fn1)
+const c6: List(alias.Type) = [alias.Variant1]
+const c7 = [alias.const1]
+const c8 = [alias.fn1]
+const c9 = <<alias.const2:bits>>
+const c10 = alias.const3 <> alias.const3
+type Type1 { Var(alias.Type) }
+type Type2 = alias.GenericType(alias.Type)
+type Record { Record (bool: Bool) }
```
lpil: I can't tell what this test is testing, it's very large!

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__rename__rename_module_from_const_annotation_recursive-2.snap
```diff
@@ -0,0 +1,31 @@
...
+expression: "import mod\nconst c: mod.GenericType(mod.Type) = mod.Node(mod.Variant1, mod.Leaf)"
+---
+----- BEFORE RENAME
+-- mod.gleam
+pub fn fn1() { Nil }
+pub const const1 = 5
+pub const const2 = <<0:8>>
+pub const const3 = "Hello"
+pub type Type { Variant1 Variant2(Int) }
+pub type GenericType(inner) { Node(inner, GenericType(inner)) Leaf }
```
lpil: It would be nicer if this repeated code wasn't so large, each test uses very little of it. Ideally tests don't include things that they don't use, to make it clearer what they test and the snapshots easier to read. Currently it's quite slow to review these snapshots as it takes a moment for each one to spot the actual test code in the snapshot.

That said, a lot of code has been written using this boilerplate now, so would be annoying to change that.

compiler-core/src/language_server/reference.rs
llpaull: That might have accidentally been personal preference sneaking in. Would you rather have the entire nested name be selected on rename?

llpaull:
Tests have been slimmed down quite extensively, if you want any specific or more tests let me know

compiler-core/src/ast.rs
```diff
@@ -739,6 +740,17 @@ impl<T> Import<T> {
         }
     }
 
+    pub(crate) fn name_location(&self) -> SrcSpan {
```
lpil: This isn't the location of the module name, that would always be the value of the `module_location` field. Perhaps "used_name_location"? Please document what this function does also

compiler-core/src/ast.rs
```diff
@@ -739,6 +740,17 @@ impl<T> Import<T> {
...
+        match self.as_name.as_ref() {
+            // the location includes "as " so subtract name length from end to only get the alias
```
lpil: Comments start with a capital letter and long lines are wrapped please 🙏

compiler-core/src/language_server/engine.rs
```diff
@@ -341,7 +342,7 @@ where
                 // we should try to provide completions for unqualified values
                 Located::ModuleStatement(Definition::Import(import)) => this
                     .compiler
-                    .get_module_interface(import.module.as_str())
+                    .get_module_interface((import.module).as_str())
```
lpil: What are these parens for?

compiler-core/src/language_server/reference.rs
lpil: Comments start with a capital letter 🙏

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +463,319 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                            self.references.push(ModuleNameReference {
+                                location: *location,
+                                kind: ModuleNameReferenceKind::Name
+                            });
+                        }
+                    }
+
+                    for (arg, type_) in arguments.iter().zip(args_type) {
+                        self.visit_type_ast(arg, type_);
+                    }
```
lpil: Can't we call the original method to do the iteration?

compiler-core/src/language_server/rename.rs
```diff
@@ -227,3 +229,48 @@ fn alias_references_in_module(
...
+    let original = module_name.split('/').next_back()?;
+
+    for reference in references {
+        match reference.kind {
+            ModuleNameReferenceKind::Import => {
+                edits.insert(reference.location.end, format!(" as {}", &params.new_name))
+            }
+            // If new name is same as original, remove the alias
+            ModuleNameReferenceKind::AliasedImport if params.new_name == original => {
+                // subtract 1 from start to delete leading space added above
```
lpil: Capital letter please.

What leading space added above? What happens if there's multiple spaces?

compiler-core/src/parse.rs
```diff
@@ -2822,7 +2822,11 @@ where
             },
             unqualified_values,
             unqualified_types,
-            module,
+            module: module.clone(),
```
lpil: Do not clone the module name please 🙏

lpil: Do not clone the module name please 🙏

lpil (commented):
Hello! This is looking fab! I've left a few small notes inline, and I think we may be missing tests to cover these situations:

- Namespaced modules (e.g. `one/two`)
- Imports with spaces (e.g. `import   one  / two  .  { three , four }  as five`
- Uses of modules with spaces (e.g. `let x = thingy . Thingy`)
- Triggering rename of an aliased module from uses of it
- Triggering rename of a namespaced module from uses of it

compiler-core/src/language_server/reference.rs
llpaull: The reason I was not using the original method was because I was using the full namespaced module name, and the `TypeAstConstructor`'s `module` field contains the alias instead of the full name (as do other AST nodes). The `type_` field however does contain the full name which is why I decided to use this custom method.

I have now been trying to use the alias instead of the full module name which has cleaned this `Visit` Impl quite a lot but causes the hover action to fail since they both use `Located::ModuleName` and the alias is not used as the key inside the `ProjectCompiler`'s `importable_modules` hashmap.

So from what I can tell, either the AST nodes need to contain the full module name instead of/as well as the alias (see `TypedExpr::ModuleSelect`) or `Located::ModuleName` needs to contain both.

Which would you rather want or do you know of an alternative solution?

lpil: Sounds good to me.

compiler-core/src/language_server/reference.rs
```diff
@@ -452,3 +463,319 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+    }
+}
+
+impl FindModuleNameReferences {
+    fn visit_type_ast(&mut self, node: &TypeAst, type_: &std::sync::Arc<Type>) {
+        let type_ = collapse_links(type_.clone());
+        match node {
+            TypeAst::Constructor(TypeAstConstructor { module, arguments, .. }) => {
+                let _ = type_.named_type_information().inspect(|(module_name, _, args_type)| {
+                    if let Some((_, location)) = module {
```
lpil: Do the module check first as it's cheaper than the call to `named_type_information`.

Write these clauses using pattern matching instead of the `inspect` please, as it's a bit cryptic and the end result isn't used, making `inspect` not the right combinator for this action also.

compiler-core/src/language_server/rename.rs
llpaull: there are two scenarios imports can fall into:

1: Import does not contain unqualified member import -> delete all of (module_location.end+1, import.location.end), this will delete all spaces between module and alias as well as the alias itself but not remove any spacing after alias

2: Import does contain unqualified member import -> import AST node does not explicitly contain end location for unqualified imports so the only safe space to remove is the one required to separate tokens

By only deleting the required space, even if there are multiple, I tried to introduce consistency and allow the user to keep multiple import aliases aligned when using the LSP to rename.

Is that okay with you?

GearsDatapacks:
Hey @llpaull are you still working on this? Would love to get this feature merged

lpil:

PR #4752, opened 2025-07-03
Fix JavaScript code generation for `let assert`

### LINKED ISSUE #4712: Invalid code generated for `let assert` on the JavaScript target
If you compile code like this:
```gleam
pub fn main() {
  let assert <<a, b:size(a), c:size(b)>> = <<1, 2, 3, 4>>
  a + b + c
}
```
On the JavaScript target, this generates the following:
```javascript
if (
    $.bitSize < 8 ||
    $.bitSize < 8 + $.byteAt(0) ||
    $.bitSize !== 8 + $.byteAt(0) + bitArraySliceToInt($, 8, 8 + a, true, false)
  ) {
    <throw error>
  }
```
The last line of the condition references the variable `a`, which does not exist as we have never bound it anywhere.

This only occurs if there are two segments using a previous segment's value as the size. The pattern `<<a, b:size(a)>>` compiles and works fine.
This also does not affect code generation of `case` expressions. The same pattern used in a `case` expression also compiles to valid code.

GearsDatapacks:
Fixes #4712
This PR merges the codegen code for `let assert` and `case`, which fixes some edge-cases which were not properly handled by the `let assert` logic.
This does temporarily make some of the `let assert` code a bit bigger and less efficient, however we are planning to further improve this part of the codebase and iron out these inefficiencies in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -393,6 +393,10 @@
   not be allowed on the JavaScript target.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where invalid code would be generated for `let assert` in some
+  cases on the JavaScript target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ## v1.11.1 - 2025-06-05
 
 ### Compiler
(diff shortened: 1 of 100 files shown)


lpil (approved):
Thank you!

PR #4760, opened 2025-07-05
Fix conflict between prelude values and user-defined values

### LINKED ISSUE #4756: Importing a value called `Ok` generates invalid JavaScript code
If you have some code like this:
```gleam
import some_module.{Error, Ok}
import gleam
```

Where both the imported `Ok` as well as the qualified `gleam.Ok` values, the following JavaScript code is generated for these imports:
```javascript
import * as $gleam from "./gleam.mjs";
import { Ok } from "./gleam.mjs";
import { Error, Ok } from "./some_module.mjs";
```

Even though `gleam.Ok` is never used unqualified, it is still imported as such, and this causes a name clash between the two `Ok` values, which prevents the program from being run.

GearsDatapacks:
Fixes #4756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -401,6 +401,11 @@
   cases on the JavaScript target.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- Fixed a bug where using the prelude `Ok` and `Error` values in a qualified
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you

PR #4761, opened 2025-07-06
Allow endianness specifiers for multi-byte codepoint segments

### LINKED ISSUE #4601: Allow endianness specifiers for pattern matching on codepoints
I was decoding some UTF-16 data in Gleam, but it was little-endian encoded, which means I could not use the `utf16_codepoint` segment for bit array pattern matching, because it doesn't support the `little` option.
This should be straightforward to implement as Erlang supports this out of the box, and utf codepoint matching is not yet implemented on JavaScript.

GearsDatapacks:
Closes #4601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,6 +196,10 @@
 
   ([Louis Pilfold](https://github.com/lpil))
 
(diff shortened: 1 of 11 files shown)


lpil (approved):
Thank you!

PR #4763, opened 2025-07-06
Fix generated docs' formatting

giacomocavalieri:
This fixes a small bug where the generated documentation for functions would be split improperly:
Screenshot 2025-07-06 alle 21 42 37


lpil (approved):
wdym, looked great before

PR #4776, opened 2025-07-12
Fix record pattern formatting

giacomocavalieri:
I noticed a small bug in the formatter where it would needlessly break a record pattern if it contained a spread. Turns out I forgot to group it!


giacomocavalieri:
I force pushed to fix a conflict so it has disabled auto merging!

giacomocavalieri:
Rebased again!

lpil:
If you put the new entries not at the bottom of the list then you're less likely to get merge conflicts btw!

PR #4778, opened 2025-07-13
Improve code generated for division on the JS target

giacomocavalieri:
I noticed that when generating code on the Erlang target we make sure to not check if the divisor is 0 if it's a constant value we can tell is not 0. So I've implemented the same improvement for the JS target, which was missing it!


compiler-core/src/javascript/expression.rs
```diff
@@ -1428,30 +1428,58 @@ impl<'module, 'a> Generator<'module, 'a> {
...
-        self.tracker.int_division_used = true;
-        docvec!["divideInt", wrap_args([left, right])]
+
+        if right.non_zero_compile_time_number() {
+            let division = if let TypedExpr::BinOp { .. } = left {
+                docvec![left_doc.surround("(", ")"), " / ", right_doc]
+            } else {
+                docvec![left_doc, " / ", right_doc]
+            };
+            docvec!["Math.trunc", wrap_args([division])]
```
GearsDatapacks: I think you need to make this `globalThis.Math.trunc`, in case the user has a constructor called `Math` defined in the current module

GearsDatapacks (commented):
One small note

compiler-core/src/javascript/expression.rs
giacomocavalieri: Thank you!

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__division_by_zero_int.snap
```diff
@@ -0,0 +1,15 @@
+---
+source: compiler-core/src/javascript/tests/numbers.rs
+expression: "pub fn main() {\n  1 / 0\n}"
+---
+----- SOURCE CODE
+pub fn main() {
+  1 / 0
```
lpil, line 7: If we know it's 0 we could compile to just 0?

lpil (commented):
💜

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__division_by_zero_int.snap
giacomocavalieri: Totally! I think in the long run it should end up being part of a constant folding step rather than special casing it during the code generation

lpil: That's a good point, aye.

compiler-core/src/ast/typed.rs
```diff
@@ -582,11 +582,26 @@ impl TypedExpr {
...
-            Self::Int{ value, .. } | Self::Float { value, .. } if NON_ZERO.get_or_init(||
-                Regex::new(r"[1-9]").expect("NON_ZERO regex")).is_match(value)
-        )
+        match self {
+            Self::Int { int_value, .. } => int_value != &BigInt::ZERO,
+            Self::Float { value, .. } => NON_ZERO
+                .get_or_init(|| Regex::new(r"[1-9]").expect("NON_ZERO regex"))
+                .is_match(value),
+            _ => false,
+        }
```
lpil, line 586: How come this macro has been expanded?

compiler-core/src/ast/typed.rs
```diff
@@ -582,11 +582,26 @@ impl TypedExpr {
...
+        use regex::Regex;
+        static NON_ZERO: OnceLock<Regex> = OnceLock::new();
+
+        match self {
+            Self::Int { int_value, .. } => int_value == &BigInt::ZERO,
+            Self::Float { value, .. } => !NON_ZERO
+                .get_or_init(|| Regex::new(r"[1-9]").expect("NON_ZERO regex"))
+                .is_match(value),
+            _ => false,
+        }
```
lpil: Extract the regex bit duplicated here and in the other function into a helper please 🙏

compiler-core/src/ast/typed.rs
giacomocavalieri: I thought it made things easier to read and now it's checking on the int value of the int variant, it's no longer matching with the regex!

lpil (approved):
Thank you!

PR #4783, opened 2025-07-15
Rename args to arguments

giacomocavalieri:
As we've discussed earlier, we decided we shouldn't be using the "args" abbreviation, so I went and changed "args" with "arguments"!


lpil (approved):
Jak will do anything to get that contribution count up

giacomocavalieri:
Anything and everything 🫡

PR #4785, opened 2025-07-15
Optimise pattern matching on js target

### LINKED ISSUE #4773: Avoid making new object in `Ok(x) -> Ok(x)`
It's not uncommon in Gleam to have a case expression where the pattern and the body of a case clause are the same.

```gleam
case result {
  Ok(x) -> Ok(x)
  Error(e) -> Error(wrap_error(e))
}
```

Here the OK value is unchanged! In an ideal world the programmer might write `-> result` to avoid recreating the same value, but this is not possible if the case expression returns a values of different type to that matched on. e.g. matching on `Result(Int, Wibble)` and returning `Result(Int, Thingy)` changes the type, even though the ok value would be unchanged.

Seeing as re-using the variable is not always possible, and the programmer may decide not to do this for clarity or personal preference, it would be good to have an optimisation for this case. When we see a pattern that matches the body exactly we can reuse the existing value and avoid constructing a new one.

I believe optimisation is already performed by the Erlang compiler, though I'm unsure if it always happens. I think it would still be good to have this optimisation not be specific to the JS codegen.

giacomocavalieri:
This PR closes #4773

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -21,7 +21,7 @@ use crate::type_::expression::{Implementations, Purity};
 use crate::type_::printer::Names;
 use crate::type_::{
     self, Deprecation, HasType, ModuleValueConstructor, PatternConstructor, Type, TypedCallArg,
-    ValueConstructor, nil,
+    ValueConstructor, ValueConstructorVariant, nil,
 };
 use num_traits::Zero;
 use std::collections::HashSet;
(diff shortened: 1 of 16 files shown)


compiler-core/src/ast.rs
```diff
@@ -1623,6 +1623,176 @@ impl TypedClause {
...
+    ///
+    pub fn is_rebuilding_matched_value(&self) -> bool {
+        // The pattern must not have any alternative patterns.
+        if !self.alternative_patterns.is_empty() {
+            return false;
+        }
+
+        // And it must match on a single subject.
+        //
+        // TODO: this limitation could actually be lifted by checking if the
```
lpil: Is this unfinished?

giacomocavalieri: I think it could be implemented quite easily with no big effort, let me give it a try

giacomocavalieri: Turned out to be a piece of cake!

lpil: Yay!

compiler-core/src/ast.rs
```diff
@@ -1623,6 +1624,168 @@ impl TypedClause {
...
+                        .zip(arguments)
+                        .all(|(pattern, expression)| {
+                            pattern_and_expression_are_the_same(&pattern.value, &expression.value)
+                        })
+            }
+
+            _ => false,
+        },
+
+        (_, _) => false,
```
lpil: Remove the catch all please 🙏 Catch alls mean we don't know when something new is added.

giacomocavalieri: Fixed it! It also made me discover a couple of further optimisations I overlooked 😁

PR #4786, opened 2025-07-16
Remove wordy preamble

giacomocavalieri:
I've removed the preamble to version resolution errors so that we only display the actual error message.

I've also split a couple of error messages that were preventing the formatter to work properly 😁


lpil (approved):
Thank you!

PR #4789, opened 2025-07-17
Code actions to add and remove anonymous functions

### LINKED ISSUE #4614: Code action to anonymize/deanonymize a function
[Playground link for reference](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIcAtgA4JT4AEA5gDYQCG5A9E3DPgDpIK1WoxbsOvKHCQMBAygFcARnQBmSOk2QMA+qprkAFAEo6wAXU0R6SBeSXQYdALx0A2ljoAmTHQDMALoWdLb2jsEAPgB8mjz4AHTkrJSG6oahpsAhdnTRdJLSDPHSMJQQOPQAvsYClXJIiirq+QAWNPh6BiZmwSw2dg6wLu6ePv5BGtlhsJEx3LyJyYYFMsVIpeX4NUiVIJVAA==). Whenever I use something like `list.fold`, `list.map`, or use any function that takes another function as an argument, I default to writing an anonymous function. However, if I'm only applying a single transformation to each element, I don't need the anonymous function logic - I can just pass the function I want to apply directly.

It would be nice if the LSP provided two code actions to switch between these styles - just like the `convert to pipe` and `convert to function call` actions. If you had a "useless" anonymous function that just called some other function, it could "de-anonymize" the anonymous function with the shorter form. And, if you passed an existing function as an argument to some other function, it could turn it into the anonymous form, for adding more intermediate transformations!

treuherz:
I've had a go at implementing new code actions discussed in #4614, one to wrap a reference to a function in an anonymous function, and one to unwrap a trivial anonymous function into a bare reference.

These seemed pretty straightforward to implement so I'm a bit concerned I've missed a whole slew of edge cases. Any pointers as to more test cases I should add would be great

I know the issue was inconclusive about the names. These ones seemed straightforward but bikeshedding is welcome

Closes #4614

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashSet, iter, sync::Arc};
+use std::{any::Any, collections::HashSet, iter, sync::Arc};
 
 use crate::{
     Error, STDLIB_PACKAGE_NAME, analyse,
@@ -3650,6 +3650,15 @@ impl VariablesNames {
         for statement in statements {
             variables.visit_typed_statement(statement);
         }
+        variables
+    }
+    
+    fn from_expr(expr: &TypedExpr) -> Self {
+        let mut variables = Self {
+            names: HashSet::new(),
+        };
+
+        variables.visit_typed_expr(expr);
         variables
     }
 }
@@ -7497,5 +7506,88 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
         }
 
         ast::visit::visit_typed_expr_block(self, location, statements);
+    }
+}
+
+struct FunctionToWrap {
+    location: SrcSpan,
+    arguments: Vec<Arc<Type>>,
+    variables_names: VariablesNames,
+}
+
+pub struct WrapInAnonymousFunction<'a> {
+    module: &'a Module,
+    line_numbers: &'a LineNumbers,
+    params: &'a CodeActionParams,
+    targets: Vec<FunctionToWrap>,
+}
+
+impl<'a> WrapInAnonymousFunction<'a> {
+    pub fn new(
+        module: &'a Module,
+        line_numbers: &'a LineNumbers,
+        params: &'a CodeActionParams,
+    ) -> Self {
+        Self {
+            module,
+            line_numbers,
+            params,
+            targets: vec![],
+        }
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+
+        let mut actions = Vec::with_capacity(self.targets.len());
+        for target in self.targets {
+            let mut name_generator = NameGenerator::new();
+            name_generator.reserve_variable_names(target.variables_names);
+            let arguments = target
+                .arguments
+                .iter()
+                .map(|t| name_generator.generate_name_from_type(t))
+                .join(", ");
+
+            let mut edits = TextEdits::new(self.line_numbers);
+            edits.insert(target.location.start, format!("fn({arguments}) {{ "));
+            edits.insert(target.location.end, format!("({arguments}) }}"));
+
+            CodeActionBuilder::new("Wrap in anonymous function")
+                .kind(CodeActionKind::REFACTOR_REWRITE)
+                .changes(self.params.text_document.uri.clone(), edits.edits)
+                .push_to(&mut actions);
+        }
+        actions
+    }
+}
+
+impl<'ast> ast::visit::Visit<'ast> for WrapInAnonymousFunction<'ast> {
+    fn visit_typed_call_arg(&mut self, arg: &'ast TypedCallArg) {
+        if let Type::Fn { ref arguments, .. } = *arg.value.type_() {
+            let variables_names = VariablesNames::from_expr(&arg.value);
+
+            self.targets.push(FunctionToWrap {
+                location: arg.location,
+                arguments: arguments.clone(),
+                variables_names,
+            });
+        }
+    }
+
+    fn visit_typed_assignment(&mut self, assignment: &'ast TypedAssignment) {
+        if assignment.kind != AssignmentKind::Let {
+            return;
+        }
+
+        if let Type::Fn { ref arguments, .. } = *assignment.value.type_() {
+            let variables_names = VariablesNames::from_expr(&assignment.value);
+
+            self.targets.push(FunctionToWrap {
+                location: assignment.value.location(),
+                arguments: arguments.clone(),
+                variables_names,
+            });
+        }
     }
 }
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1,4 +1,4 @@
-use std::{any::Any, collections::HashSet, iter, sync::Arc};
+use std::{collections::HashSet, iter, sync::Arc};
 
 use crate::{
     Error, STDLIB_PACKAGE_NAME, analyse,
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -3652,7 +3652,7 @@ impl VariablesNames {
         }
         variables
     }
-    
+
     fn from_expr(expr: &TypedExpr) -> Self {
         let mut variables = Self {
             names: HashSet::new(),
@@ -7588,6 +7588,110 @@ impl<'ast> ast::visit::Visit<'ast> for WrapInAnonymousFunction<'ast> {
                 arguments: arguments.clone(),
                 variables_names,
             });
+        }
+    }
+}
+
+pub struct UnwrapAnonymousFunction<'a> {
+    module: &'a Module,
+    line_numbers: &'a LineNumbers,
+    params: &'a CodeActionParams,
+    targets: Vec<FnToUnwrap>,
+}
+
+struct FnToUnwrap {
+    fn_location: SrcSpan,
+    inner_function_location: SrcSpan,
+}
+
+impl<'a> UnwrapAnonymousFunction<'a> {
+    pub fn new(
+        module: &'a Module,
+        line_numbers: &'a LineNumbers,
+        params: &'a CodeActionParams,
+    ) -> Self {
+        Self {
+            module,
+            line_numbers,
+            params,
+            targets: vec![],
+        }
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+
+        let mut actions = Vec::with_capacity(self.targets.len());
+        for target in self.targets {
+            let mut edits = TextEdits::new(self.line_numbers);
+
+            edits.delete(SrcSpan {
+                start: target.fn_location.start,
+                end: target.inner_function_location.start,
+            });
+            edits.delete(SrcSpan {
+                start: target.inner_function_location.end,
+                end: target.fn_location.end,
+            });
+
+            CodeActionBuilder::new("Unwrap anonymous function")
+                .kind(CodeActionKind::REFACTOR_REWRITE)
+                .changes(self.params.text_document.uri.clone(), edits.edits)
+                .push_to(&mut actions);
+        }
+        actions
+    }
+}
+
+impl<'ast> ast::visit::Visit<'ast> for UnwrapAnonymousFunction<'ast> {
+    fn visit_typed_expr_fn(
+        &mut self,
+        location: &'ast SrcSpan,
+        _type_: &'ast Arc<Type>,
+        kind: &'ast FunctionLiteralKind,
+        arguments: &'ast [TypedArg],
+        body: &'ast Vec1<TypedStatement>,
+        _return_annotation: &'ast Option<ast::TypeAst>,
+    ) {
+        match kind {
+            FunctionLiteralKind::Anonymous { .. } => (),
+            _ => return,
+        }
+
+        // We need the existing argument list for the fn to be a 1:1 match for the args we pass
+        // to the called function. We figure out what the call-arg list needs to look like here,
+        // and bail out if any incoming args are discarded.
+        let mut expected_arguments = Vec::with_capacity(arguments.len());
+        for a in arguments {
+            match &a.names {
+                ArgNames::Named { name, .. } => expected_arguments.push(name),
+                ArgNames::NamedLabelled { name, .. } => expected_arguments.push(name),
+                ArgNames::Discard { .. } => return,
+                ArgNames::LabelledDiscard { .. } => return,
+            }
+        }
+
+        // match fn bodies with only a single function call
+        if let [stmt] = body.as_slice()
+            && let TypedStatement::Expression(expr) = stmt
+            && let TypedExpr::Call { fun, arguments, .. } = expr
+        {
+            let mut call_arguments = Vec::with_capacity(arguments.len());
+            for a in arguments {
+                match &a.value {
+                    TypedExpr::Var { name, .. } => call_arguments.push(name),
+                    _ => return,
+                }
+            }
+
+            if call_arguments != expected_arguments {
+                return;
+            }
+
+            self.targets.push(FnToUnwrap {
+                fn_location: *location,
+                inner_function_location: fun.location(),
+            })
         }
     }
 }
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -132,6 +132,7 @@ const WRAP_IN_BLOCK: &str = "Wrap in block";
 const GENERATE_VARIANT: &str = "Generate variant";
 const REMOVE_BLOCK: &str = "Remove block";
 const WRAP_IN_ANONYMOUS_FUNCTION: &str = "Wrap in anonymous function";
+const UNWRAP_ANONYMOUS_FUNCTION: &str = "Unwrap anonymous function";
 
 macro_rules! assert_code_action {
     ($title:expr, $code:literal, $range:expr $(,)?) => {
@@ -9165,5 +9166,61 @@ fn op_factory(a: Int, b: Int, c: Int) -> fn(Int) -> Int {
 }
 ",
         find_position_of("op_factory").to_selection()
+    );
+}
+
+#[test]
+fn unwrap_trivial_anonymous_function() {
+    assert_code_action!(
+        UNWRAP_ANONYMOUS_FUNCTION,
+        "import gleam/list
+
+pub fn main() {
+    list.map([1, 2, 3], fn(int) { op(int) })
+}
+
+fn op(i: Int) -> Int {
+    todo
+}
+",
+        find_position_of("fn(int)").to_selection()
+    );
+}
+
+#[test]
+fn unwrap_anonymous_function_unavailable_when_args_discarded() {
+    assert_no_code_actions!(
+        UNWRAP_ANONYMOUS_FUNCTION,
+        "import gleam/list
+
+pub fn main() {
+    list.index_map([1, 2, 3], fn(_, int) { op(int) })
+}
+
+fn op(i: Int) -> Int {
+    todo
+}
+",
+        find_position_of("fn(_, int)").to_selection()
+    );
+}
+
+#[test]
+fn unwrap_anonymous_function_unavailable_with_different_args() {
+    assert_no_code_actions!(
+        UNWRAP_ANONYMOUS_FUNCTION,
+        "import gleam/list
+
+const another_int = 7
+        
+pub fn main() {
+    list.map([1, 2, 3], fn(int) { op(another_int) })
+}
+
+fn op(i: Int) -> Int {
+    todo
+}
+",
+        find_position_of("fn(int)").to_selection()
     );
 }
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -131,6 +131,7 @@ const REMOVE_ALL_ECHOS_FROM_THIS_MODULE: &str = "Remove all `echo`s from this mo
 const WRAP_IN_BLOCK: &str = "Wrap in block";
 const GENERATE_VARIANT: &str = "Generate variant";
 const REMOVE_BLOCK: &str = "Remove block";
+const WRAP_IN_ANONYMOUS_FUNCTION: &str = "Wrap in anonymous function";
 
 macro_rules! assert_code_action {
     ($title:expr, $code:literal, $range:expr $(,)?) => {
@@ -9130,5 +9131,39 @@ fn remove_block_does_not_unwrap_a_block_with_multiple_statements() {
 }
 ",
         find_position_of("1").to_selection()
+    );
+}
+
+#[test]
+fn wrap_call_arg_in_anonymous_function() {
+    assert_code_action!(
+        WRAP_IN_ANONYMOUS_FUNCTION,
+        "import gleam/list
+
+pub fn main() {
+    list.map([1, 2, 3], op)
+}
+
+fn op(i: Int) -> Int {
+    todo
+}
+",
+        find_position_of("op").to_selection()
+    );
+}
+
+#[test]
+fn wrap_assignment_in_anonymous_function() {
+    assert_code_action!(
+        WRAP_IN_ANONYMOUS_FUNCTION,
+        "pub fn main() {
+    let op = op_factory(1, 2, 3)
+}
+
+fn op_factory(a: Int, b: Int, c: Int) -> fn(Int) -> Int {
+    todo
+}
+",
+        find_position_of("op_factory").to_selection()
     );
 }
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -43,9 +43,9 @@ use super::{
         FillInMissingLabelledArgs, FillUnusedFields, FixBinaryOperation,
         FixTruncatedBitArraySegment, GenerateDynamicDecoder, GenerateFunction, GenerateJsonEncoder,
         GenerateVariant, InlineVariable, InterpolateString, LetAssertToCase, PatternMatchOnValue,
-        RedundantTupleInCaseSubject, RemoveEchos, RemoveUnusedImports, UseLabelShorthandSyntax,
-        WrapInAnonymousFunction, WrapInBlock, code_action_add_missing_patterns,
-        code_action_convert_qualified_constructor_to_unqualified,
+        RedundantTupleInCaseSubject, RemoveEchos, RemoveUnusedImports, UnwrapAnonymousFunction,
+        UseLabelShorthandSyntax, WrapInAnonymousFunction, WrapInBlock,
+        code_action_add_missing_patterns, code_action_convert_qualified_constructor_to_unqualified,
         code_action_convert_unqualified_constructor_to_qualified, code_action_import_module,
         code_action_inexhaustive_let_to_case,
     },
@@ -438,6 +438,7 @@ where
             actions.extend(WrapInBlock::new(module, &lines, &params).code_actions());
             actions.extend(RemoveBlock::new(module, &lines, &params).code_actions());
             actions.extend(WrapInAnonymousFunction::new(module, &lines, &params).code_actions());
+            actions.extend(UnwrapAnonymousFunction::new(module, &lines, &params).code_actions());
             GenerateDynamicDecoder::new(module, &lines, &params, &mut actions).code_actions();
             GenerateJsonEncoder::new(
                 module,
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -44,7 +44,7 @@ use super::{
         FixTruncatedBitArraySegment, GenerateDynamicDecoder, GenerateFunction, GenerateJsonEncoder,
         GenerateVariant, InlineVariable, InterpolateString, LetAssertToCase, PatternMatchOnValue,
         RedundantTupleInCaseSubject, RemoveEchos, RemoveUnusedImports, UseLabelShorthandSyntax,
-        WrapInBlock, code_action_add_missing_patterns,
+        WrapInAnonymousFunction, WrapInBlock, code_action_add_missing_patterns,
         code_action_convert_qualified_constructor_to_unqualified,
         code_action_convert_unqualified_constructor_to_qualified, code_action_import_module,
         code_action_inexhaustive_let_to_case,
@@ -437,6 +437,7 @@ where
             actions.extend(InlineVariable::new(module, &lines, &params).code_actions());
             actions.extend(WrapInBlock::new(module, &lines, &params).code_actions());
             actions.extend(RemoveBlock::new(module, &lines, &params).code_actions());
+            actions.extend(WrapInAnonymousFunction::new(module, &lines, &params).code_actions());
             GenerateDynamicDecoder::new(module, &lines, &params, &mut actions).code_actions();
             GenerateJsonEncoder::new(
                 module,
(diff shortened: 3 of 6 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+                ArgNames::NamedLabelled { name, .. } => expected_arguments.push(name),
+                ArgNames::Discard { .. } => return,
+                ArgNames::LabelledDiscard { .. } => return,
+            }
+        }
+
+        // match fn bodies with only a single function call
+        if let [stmt] = body.as_slice()
+            && let TypedStatement::Expression(expr) = stmt
+            && let TypedExpr::Call { fun, arguments, .. } = expr
```
treuherz: I think this is the first use in the repo of Rust 1.88's new chained lets. I saw in another ticket that you don't try to support old Rusts, so I hope this is OK. It looks a lot more readable to me than having the whole nested pattern on one line but I don't have a lot of Rust experience so it could just be unfamiliarity

compiler-core/src/language_server/code_action.rs
```diff
@@ -3652,6 +3652,15 @@ impl VariablesNames {
         }
         variables
     }
+
+    fn from_expr(expr: &TypedExpr) -> Self {
```
lpil: No abbreviations please 🙏 expr -> expression

It's not very clear to me from the name what this does. A doc comment would help a lot, and maybe a more descriptive name

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
         ast::visit::visit_typed_expr_block(self, location, statements);
     }
 }
+
+struct FunctionToWrap {
```
lpil: Document this please

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    location: SrcSpan,
+    arguments: Vec<Arc<Type>>,
+    variables_names: VariablesNames,
+}
+
+pub struct WrapInAnonymousFunction<'a> {
```
lpil: Document this please

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+            CodeActionBuilder::new("Wrap in anonymous function")
+                .kind(CodeActionKind::REFACTOR_REWRITE)
+                .changes(self.params.text_document.uri.clone(), edits.edits)
+                .push_to(&mut actions);
+        }
+        actions
+    }
+}
+
+impl<'ast> ast::visit::Visit<'ast> for WrapInAnonymousFunction<'ast> {
```
lpil: I think this is computing the code action for all non-nested function references that are assigned to variables or are passed as arguments, but we want only the ones that are currently within the range specified by the code action parameters.

I think it would ideally be any non-called function rather than specifically these two positions in the AST, and think the non-nested restriction should be removed.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+            self.targets.push(FunctionToWrap {
+                location: assignment.value.location(),
+                arguments: arguments.clone(),
+                variables_names,
+            });
+        }
+    }
+}
+
+pub struct UnwrapAnonymousFunction<'a> {
```
lpil, line 11540: Document this please

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    module: &'a Module,
+    line_numbers: &'a LineNumbers,
+    params: &'a CodeActionParams,
+    targets: Vec<FnToUnwrap>,
+}
+
+struct FnToUnwrap {
```
lpil: Let's consistently use the term "function" rather than mixing fn and function.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    fn_location: SrcSpan,
+    inner_function_location: SrcSpan,
```
lpil: What are these? Document them please 🙏

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+        if assignment.kind != AssignmentKind::Let {
+            return;
+        }
+
+        if let Type::Fn { ref arguments, .. } = *assignment.value.type_() {
+            let variables_names = VariablesNames::from_expr(&assignment.value);
+
+            self.targets.push(FunctionToWrap {
+                location: assignment.value.location(),
+                arguments: arguments.clone(),
```
lpil, line 11488: Don't clone AST please 🙏

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    location: SrcSpan,
+    arguments: Vec<Arc<Type>>,
+    variables_names: VariablesNames,
+}
+
+pub struct WrapInAnonymousFunction<'a> {
+    module: &'a Module,
+    line_numbers: &'a LineNumbers,
+    params: &'a CodeActionParams,
+    targets: Vec<FunctionToWrap>,
```
lpil: A target is `JavaScript` or `Erlang`, so best rename this field.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+            CodeActionBuilder::new("Unwrap anonymous function")
+                .kind(CodeActionKind::REFACTOR_REWRITE)
+                .changes(self.params.text_document.uri.clone(), edits.edits)
+                .push_to(&mut actions);
+        }
+        actions
+    }
+}
+
+impl<'ast> ast::visit::Visit<'ast> for UnwrapAnonymousFunction<'ast> {
```
lpil, line 11690: Range check also needed here 🙏

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+            match &a.names {
+                ArgNames::Named { name, .. } => expected_arguments.push(name),
+                ArgNames::NamedLabelled { name, .. } => expected_arguments.push(name),
+                ArgNames::Discard { .. } => return,
+                ArgNames::LabelledDiscard { .. } => return,
+            }
+        }
+
+        // match fn bodies with only a single function call
+        if let [stmt] = body.as_slice()
```
lpil: stmt -> statement

compiler-core/src/language_server/code_action.rs
lpil: Could you replace it with `let else return` please? It'll be easier to document and a little easier to read without the indentation as you don't have to keep the context in your head until the end of the function.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+            for a in arguments {
+                match &a.value {
+                    TypedExpr::Var { name, .. } => call_arguments.push(name),
+                    _ => return,
+                }
+            }
+
+            if call_arguments != expected_arguments {
+                return;
+            }
```
lpil, line 11098: Document the logic in this section please

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7508,190 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+        for a in arguments {
+            match &a.names {
+                ArgNames::Named { name, .. } => expected_arguments.push(name),
+                ArgNames::NamedLabelled { name, .. } => expected_arguments.push(name),
+                ArgNames::Discard { .. } => return,
+                ArgNames::LabelledDiscard { .. } => return,
+            }
+        }
+
+        // match fn bodies with only a single function call
```
lpil: Capital letters at the start of comments please, and explain why something is done rather than what is done. Thank you

compiler-core/src/language_server/tests/action.rs
```diff
@@ -131,6 +131,8 @@ const REMOVE_ALL_ECHOS_FROM_THIS_MODULE: &str = "Remove all `echo`s from this mo
 const WRAP_IN_BLOCK: &str = "Wrap in block";
 const GENERATE_VARIANT: &str = "Generate variant";
 const REMOVE_BLOCK: &str = "Remove block";
+const WRAP_IN_ANONYMOUS_FUNCTION: &str = "Wrap in anonymous function";
+const UNWRAP_ANONYMOUS_FUNCTION: &str = "Unwrap anonymous function";
```
lpil: To me this sounds like the anonymous function remains but its wrapper will be removed. "Remove anonymous function wrapper" perhaps?

compiler-core/src/language_server/tests/action.rs
```diff
@@ -9132,3 +9134,93 @@ fn remove_block_does_not_unwrap_a_block_with_multiple_statements() {
...
 }
+
+#[test]
+fn wrap_call_arg_in_anonymous_function() {
+    assert_code_action!(
+        WRAP_IN_ANONYMOUS_FUNCTION,
+        "import gleam/list
+
+pub fn main() {
+    list.map([1, 2, 3], op)
```
lpil: 2 space indentation for Gleam please 🙏

language-server/src/tests/action.rs
```diff
@@ -9132,3 +9134,93 @@ fn remove_block_does_not_unwrap_a_block_with_multiple_statements() {
...
+    list.map([1, 2, 3], fn(int) { op(another_int) })
+}
+
+fn op(i: Int) -> Int {
+    todo
+}
+",
+        find_position_of("fn(int)").to_selection()
+    );
+}
```
lpil, line 13481: Let's add more tests to cover all the possible scenarios. We'll want to cover:

- Labelled arguments
- Pipes
- Qualified functions from other modules
- Records
- Comments

And anything else you can think of!

compiler-core/src/language_server/code_action.rs
treuherz: sorry, do you mean keep the chained lets but use `else return` to de-indent the body, or replace it with just

 ```rust
let [TypedStatement::Expression(TypedExpr::Call { fun, arguments, .. })] = body.as_slice()
else {
    return;
};
```

Comment (treuherz) [reply compiler-core/src/language_server/code_action.rs]: never mind, you can't do the other thing

File: compiler-core/src/language_server/code_action.rs (hunk 5 above)
Comment (treuherz) [reply compiler-core/src/language_server/code_action.rs]: the range check makes sense, I'd missed it before bc I was looking at `RemoveUnusedImports` which obviously doesn't need it 🤦 

Lots of the other actions store their `TextEdits` in the struct but only need it there for the range check, which can be done with just a `LineNumbers`. Storing `LineNumbers` seems more natural to me but I'm splitting hairs and will defer to consistency if you like

Comment (treuherz) [reply compiler-core/src/language_server/code_action.rs]: The non-nested restriction was inadvertent, if I'm reading it right I need to call back into the visitor after I'm done?

File: compiler-core/src/language_server/code_action.rs (hunk 9 above)
Comment (treuherz) [reply compiler-core/src/language_server/code_action.rs:11488]: this is a `Vec<Arc<Type>>` rather than an AST node, is that ok? Using it in-place to generate our new argument names in the visitor felt odd

**Comment by treuherz** [comment]
Ok, there was more to do here than I thought but the code is better than before 🎉 

I've copped out and not offered the action on functions with comments in. I think it's doable, but found an issue I wanted some advice on while I was on my way there. The `first_comment_between` helper has a binary search which will return _any_ comment in a given line-span, not necessarily the first. We could guarantee it finds the first by recursing back into the list with smaller boundaries, but I haven't written recursive algorithms in Rust before and I don't know how to tell if I'm going to blow up the stack.

**Comment by llakala** [comment]
Probably also want a changelog entry, right?

Thanks for doing this!

**Comment by treuherz** [comment]
Ok, I've fixed `first_comment_between` without recursion but actually unwrapping functions with comments is a bit of a minefield because unless I re-indent them afterwards it looks rubbish. How would you feel about keeping this as-is (i.e. not offering the action on functions with comments) and improving that in a follow-up?

**Comment by lpil** [comment]
Comments shouldn't change the behaviour of the language server, so we would want to provide code actions regardless of whether or not there are comments.

Having imperfect formatting afterwards when there's comments doesn't sound too bad, the programmer can run the formatter. Do you have some examples of what the output looks like?

I've less context than you, but it seems like comments inside the body of the function shouldn't be much of an issue as they can be left as-is:

```gleam
fn(x, y) {
  // comment
  thingy(x, y)
}
```
```gleam

  // comment
  thingy

```

For comments inside the parameters I think we can discard them happily.

```gleam
fn(
  // Hello
  x,
  y,
) {
  thingy(x, y)
}
```
```gleam




  thingy

```

compiler-core/src/language_server/code_action.rs
lpil: Using line numbers seems good to me!

treuherz:
It's not really anything that the autoformatter wouldn't tidy up. There are some snapshots [in this commit](https://github.com/treuherz/gleam/commit/fb6aad10f3af754ee4f1aa24af9fdc17fb24c36a), the worst it gets is probably things like [this one](https://github.com/treuherz/gleam/commit/fb6aad10f3af754ee4f1aa24af9fdc17fb24c36a#diff-1f0626636b72e2e29acf332cea450aafb7e957f9dd0032a78fac3d8e30ac4ef8R21) where the comment has moved lines but that's not too big a deal. If that's acceptable (or close enough to work on) I'll include it in this PR

compiler-core/src/parse/extra.rs
```diff
@@ -40,19 +40,34 @@ impl ModuleExtra {
...
+                .ok()
+        };
+
+        let mut best = None;
+        let mut search_list = &self.comments[..];
+        while let Some(index) = inner(search_list, start, end) {
+            best = self.comments.get(index).copied();
+            search_list = search_list.get(0..index).unwrap_or(&[]);
+        }
+        best
```
lpil: Why are we now traversing the comments and copying their locations multiple times?

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7511,311 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    module: &'a Module,
+    line_numbers: &'a LineNumbers,
+    params: &'a CodeActionParams,
+    functions: Vec<FunctionToWrap>,
+}
+
+/// Helper struct, a target for [WrapInAnonymousFunction].
+struct FunctionToWrap {
+    location: SrcSpan,
+    arguments: Vec<Arc<Type>>,
```
lpil: This could be borrowed I think
```suggestion
    arguments: &'a [Arc<Type>],
```

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_anonymous_function_in_anonymous_function.snap
```diff
@@ -0,0 +1,16 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "pub fn main() {\n  let f = fn(in) { ception(in) }\n}\n\n"
+---
+----- BEFORE ACTION
+pub fn main() {
+  let f = fn(in) { ception(in) }
```
lpil: I don't think we should be offering to wrap an anonymous function in another anonymous function

language-server/src/tests/snapshots/gleam_language_server__tests__action__wrap_pipeline_step_in_anonymous_function.snap
```diff
@@ -0,0 +1,32 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: "pub fn main() {\n  1 |> wibble |> wobble\n}\n\nfn wibble(i) {\n  todo\n}\n\nfn wobble(i) {\n  todo\n}\n\n"
+---
+----- BEFORE ACTION
+pub fn main() {
+  1 |> wibble |> wobble
```
lpil, line 7: Could we have a test for wibble taking 2 arguments please

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7511,311 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+
+        let mut actions = Vec::with_capacity(self.functions.len());
+        for function in &self.functions {
+            let mut edits = TextEdits::new(self.line_numbers);
+
+            // We need to delete the anonymous function's head but preserve
+            // comments between it and the inner function call.
+            edits.delete(self.span_until_comment(SrcSpan {
```
lpil: Rather than seek through the comments multiple times I think we could use the location stored in the `FunctionLiteralKind::Anonymous` variant.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7499,3 +7511,311 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveBlock<'ast> {
...
+                start: function.outer_function.start,
+                end: function.inner_function.start,
+            }));
+
+            // Now we need to delete the inner function call's arguments,
+            // preserving comments before the outer function tail.
+            edits.delete(self.span_until_comment(SrcSpan {
```
lpil: In the `TypedExpr::Call` variant we could store the byte index of the `(` to make it easier to get that position.

lpil (commented):
Awesome, really nice work!! I've left a few small notes inline. Thank you

compiler-core/src/language_server/code_action.rs
treuherz: I can't figure out how to get that past borrowck 😕.  The arguments are pulled from a field of an `Arc<Type>` [here](https://github.com/gleam-lang/gleam/pull/4789/files#diff-62c42fc2b6eac6a3755c10e81c8165cde697582c18ced980c797e59a7e93df45R7594). If we want to avoid cloning the argument list then I could clone the whole `Arc<Fn>` for the function instead and pull the arguments out later

compiler-core/src/language_server/code_action.rs
treuherz: see previous

compiler-core/src/language_server/code_action.rs
treuherz: I think by pulling more data through from the parser we can eliminate both comment-checks, but I'm not sure we can currently eliminate either. 

Here's what I think we get from the parser currently:

```
fn(arg_1: T1, arg_2: T2) -> T3  { [comments] inner_function(arg_1, arg_2) [comments] }
└─────────────head───────────┘               └────inner───┘
└─────────────────────────────────────────outer──────────────────────────────────────┘
```

So means if I pull the opening brace through the parser I can eliminate this first comment-seek, and if I pull the span of the inner argument-list through I can eliminate the second. Without either I think we risk mangling things if the input isn't well-formatted (e.g. too much space before the opening brace)

compiler-core/src/parse/extra.rs
treuherz: I've moved the copy out of the loop but I think the multiple-traversal is necessary to ensure we return the _first_ comment in the range rather than just _a_ comment in the range. I added some failing tests for the previous implementation in [this commit](https://github.com/gleam-lang/gleam/pull/4789/commits/0a1e6d6c030963fb798d255c01396fff93bfbf20), I can comment this new one more to explain the method properly

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_anonymous_function_in_anonymous_function.snap
treuherz: fixed

language-server/src/tests/snapshots/gleam_language_server__tests__action__wrap_pipeline_step_in_anonymous_function.snap
treuherz: added

lpil:
Do remember to undraft this PR and tag me when you're ready for a review!

compiler-core/src/language_server/code_action.rs
treuherz: I've pulled more locations through the parser, and have followed this approach so we now don't care about comments. I've left the bug-fix in though!

treuherz:
Sorry for the delay, I wanted some advice on which route to take to fix the problems, and then life got in the way. I think the new approach is better, and I've rebased on top of main. Ready for review @lpil

compiler-core/src/ast/untyped.rs
```diff
@@ -293,11 +294,18 @@ impl HasLocation for UntypedExpr {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum FunctionLiteralKind {
     Capture { hole: SrcSpan },
-    Anonymous { head: SrcSpan },
+    Anonymous { head: SrcSpan, body: SrcSpan },
```
lpil: Isn't the start of `body` always the end of `head`?

This data also is duplicating the data of the containing function data structure I think. It would be fab to not have that duplication there

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_anonymous_function_in_anonymous_function.snap
lpil: Is this snapshot unused now?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_capturing_pipeline_step_in_anonymous_function.snap
```diff
@@ -0,0 +1,32 @@
...
+
+fn wobble(i) {
+  todo
+}
+
+
+
+----- AFTER ACTION
+pub fn main() {
+  1 |> fn(int) { wibble(2, _)(int) } |> wobble
```
lpil:
```suggestion
  1 |> fn(int) { wibble(2, int) } |> wobble
```

I think in this case the code action text should be "Expand function capture"

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_multiargument_pipeline_step_in_anonymous_function.snap
```diff
@@ -0,0 +1,32 @@
...
+
+fn wobble(i) {
+  todo
+}
+
+
+
+----- AFTER ACTION
+pub fn main() {
+  1 |> fn(int, int_2) { wibble(int, int_2) }(2) |> wobble
```
lpil: Let's not offer the code action in this situation, as the code produces is undesirable

compiler-core/src/parse/extra.rs
```diff
@@ -40,19 +40,34 @@ impl ModuleExtra {
...
+                .ok()
+        };
+
+        let mut best = None;
+        let mut search_list = &self.comments[..];
+        while let Some(index) = inner(search_list, start, end) {
+            best = self.comments.get(index);
+            search_list = search_list.get(0..index).unwrap_or(&[]);
+        }
+        best.copied()
```
lpil: Why has this changed? It is much harder to understand now, and it seems to do a lot more work than the binary search of before.

compiler-core/src/ast/typed.rs
```diff
@@ -76,6 +76,7 @@ pub enum TypedExpr {
         type_: Arc<Type>,
         fun: Box<Self>,
         arguments: Vec<CallArg<Self>>,
+        arguments_start: Option<u32>,
```
lpil: Why is this optional?

language-server/src/tests/action.rs
```diff
@@ -9796,3 +9798,439 @@ fn remove_unreachable_branches_does_not_pop_up_if_all_branches_are_reachable() {
...
+
+",
+        find_position_of("wibble").to_selection()
+    );
+}
+
+#[test]
+fn wrap_capturing_pipeline_step_in_anonymous_function() {
+    assert_code_action!(
+        WRAP_IN_ANONYMOUS_FUNCTION,
```
lpil, line 13184: In this case the code action text should be "Expand function capture"

lpil (commented):
Thank you! I've left a bunch of notes inline.

Sorry for the delay, my inbox has got out of control!

compiler-core/src/parse/extra.rs
treuherz: A simple binary search doesn't guarantee it finds the first match only _a_ match. Without this change the tests I added below don't all pass correctly.

compiler-core/src/ast/untyped.rs
treuherz: Using `body.start` skips past the opening brace and any spaces between it and the first statement of the body. I can do that with `next_nonwhitespace` instead.

compiler-core/src/ast/typed.rs
treuherz: There are a bunch of places in inference code where this information isn't readily available. I'm not wild about what I've ended up with but not sure how else to go about this

language-server/src/tests/action.rs
treuherz: I'm going to exclude this case, the existing "Expand function capture" refactor has it covered.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_capturing_pipeline_step_in_anonymous_function.snap
treuherz: I'm going to exclude this case, the existing "Expand function capture" refactor has it covered.

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__wrap_multiargument_pipeline_step_in_anonymous_function.snap
treuherz: Excluded along with other captures.

compiler-core/src/ast/typed.rs
lpil: Position information comes from parsing, this data structure comes from analysis, which comes after parsing, so it must always be present.

compiler-core/src/parse/extra.rs
lpil: Wouldn't it be more efficient to bisection search and then seek backwards? It would be a lot easier to understand, I'm finding this code very hard to read.

compiler-core/src/ast/typed.rs
treuherz: The problem is cases like [infer_insert_pipe](https://github.com/treuherz/gleam/blob/main/compiler-core/src/type_/pipe.rs#L301) or [infer_apply_to_call_pipe](https://github.com/treuherz/gleam/blob/main/compiler-core/src/type_/pipe.rs#L262) where I don't think value is clear or well-defined. What if it was instead something like `argument_parentheses: Option<SrcSpan>`?

giacomocavalieri:
Hello! This has some conflicts with the main branch, would you mind rebasing if you think this is ready for review?

compiler-core/src/parse/extra.rs
treuherz: I don't know if that would always be more efficient, without foreknowledge of how long big a start/end span is being given. I've tried to make the function a bit clearer, wdyt?

treuherz:
Rebased on main

compiler-core/src/ast/typed.rs
lpil: Sounds good! Be sure to include detailed documentation for this field, as it's somewhat unclear.

compiler-core/src/parse/extra.rs
```diff
@@ -36,23 +36,51 @@ impl ModuleExtra {
...
-                    Ordering::Less
-                } else if comment.start > end {
-                    Ordering::Greater
-                } else {
-                    Ordering::Equal
-                }
-            })
-            .ok()
-            .and_then(|index| self.comments.get(index).copied())
+        // Helper function to find a comment that is between the given start
```
lpil: I think this could be written more clearly and more efficiently something like this:

```rust
self.comments
  .partition_point(|comment| comment.start >= start)
  .ok()
  .and_then(|index| self.comments.get(index).copied())
  .filter(|comment| comment.start < end)
```

treuherz: `partition_point` returns a `usize`, and doesn't define its return value if no elements match the predicate i.e. if no comment starts within the given span. It also doesn't account for the case where a comment ends in the span, which the existing code.

I'm really sorry this function has been such a source of difficulty. To be clear, in the current version of this PR, it isn't even used in the new code action. I've just kept the change here for the bugfix.

I'm going to push a version that takes the pre-existing implementation and then seeks backwards from that point, as you suggested before. I think the worst-case scenario could be worse but it's probably negligible, and you're right that it's much clearer.

compiler-core/src/ast/typed.rs
treuherz: done

treuherz:
Hope the comments are sufficiently addressed. I've tried to sane-ify the commits again too

compiler-core/src/parse/extra.rs
```diff
@@ -36,12 +36,17 @@ impl ModuleExtra {
             || self.module_comments.binary_search_by(cmp).is_ok()
     }
 
-    pub(crate) fn has_comment_between(&self, start: u32, end: u32) -> bool {
+    pub fn has_comment_between(&self, start: u32, end: u32) -> bool {
         self.first_comment_between(start, end).is_some()
     }
 
+    // TODO: this won't necessarily find the first comment, just any comment
```
lpil: No todos please 🙏 

Why is this saying something different to what the documentation and the code says?

compiler-core/src/ast/typed.rs
```diff
@@ -76,6 +76,13 @@ pub enum TypedExpr {
         type_: Arc<Type>,
         fun: Box<Self>,
         arguments: Vec<CallArg<Self>>,
+        /// Specifies the source location of the argument list, if present. Can
+        /// be None in cases where the function call has been inferred as part
+        /// of a pipeline.
+        ///
+        /// The endpoints of the span will sit before and after the list's
+        /// parentheses.
+        argument_parentheses: Option<SrcSpan>,
```
lpil: This should be a position rather than a span- the end is never used and it's also guaranteed to be the same as the location end.

compiler-core/src/parse/extra.rs
treuherz: Left behind from an earlier phase of this where I’d added a TODO and a failing test but not fixed it yet 😅

PR #4795, opened 2025-07-21
Make sure CHANGELOG doesn't trigger CI

giacomocavalieri:
So when someone (me) forgets to update the changelog and have to push it, it won't trigger CI a second time


lpil (approved):
Thanks!

PR #4800, opened 2025-07-22
Fix formatting of `as` messages

giacomocavalieri:
This PR fixes a bug with the formatter where the formatting of `as` messages following `echo`, `panic`, `todo`, `assert` and `let assert` would not be nested properly when preceded by a comment.

Before:
```gleam
todo as
// Comment
"wibble"
```
After:
```gleam
todo as
  // Comment!
  "wibble"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   after it.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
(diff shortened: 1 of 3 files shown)


JosephTLyons:
I wasn't sure if this was expected or not, so leaving this here after finding this PR via Google:

SCR-20250822-pbxk

PR #4801, opened 2025-07-23
Add negating literals in different bases to v2 list

giacomocavalieri:


lpil:
Wr need to agree on a design that we thing provides sufficient value for the amount of problems the migration would cause, but no discussion has been done on this one.

The change is potentially a lot more impactful than the others on this list, so we need to be extra thorough here too.

giacomocavalieri:
I'm not sure what you mean by design in the context of this change, shouldn't negating behave the same with all literals instead of this strange behaviour where these specific literals are not negated as one would expect?

> The change is potentially a lot more impactful than the others on this list, so we need to be extra thorough here too.

What other pieces of code could be impacted by this? I couldn't come up with other examples where the code's behaviour would change. And this piece of code seems extremely unlikely to ever happen in real life given it would already result in a warning/error and be formatted with the two expressions on different lines

lpil:
This is a list of things that we want to make a breaking change for in a future version 2 and serves as a checklist for the preparation for that.

We have made no such decision for negative non-decimal numbers, we don't even have a design proposal.

The process is that discussions come first, then issues, then code changes!

PR #4808, opened 2025-07-24
Improve "pattern match on value" code action to make it work on lists and more patterns

### LINKED ISSUE #4806: Pattern match on variable/argument code action doesn't work on lists
If I have this piece of code:
```gleam
pub fn wibble(list: List(a)) {
  //          ^^^^
  todo
}
```
I'd like to trigger the pattern match on argument code action on `list` and get something like this:
```gleam
pub fn wibble(list: List(a)) {
  case list {
    [first, ..rest] -> todo
    [] -> todo
  }
  todo
}
```

### LINKED ISSUE #4807: Pattern match on variable doesn't work on nested variables inside a pattern
If I have this piece of code:
```gleam
pub fn main() {
  let #(a_result, something_else) = todo
  //    ^^^^^^^^
}
```
I'd like to trigger the pattern match on variable code action on the `a_result` variable and get something like this:
```gleam
pub fn main() {
  let #(a_result, something_else) = todo
  case a_result {
    Ok(_) -> todo
    Error(_) -> todo
  }
}
```
But at the moment this doesn't work, and it doesn't work on other constructors' fields, not just tuples

giacomocavalieri:
With this PR the "pattern match on value/argument" code action:
- now works on lists, generating the patterns for the empty and non empty list
- now works on variables inside other destructuring patterns

Here's a little example of what is now possible!

https://github.com/user-attachments/assets/9337831a-fcc6-478c-8917-76b51ed57ee4

Fixes #4806 
Fixes #4807

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -4444,6 +4444,7 @@ pub struct PatternMatchOnValue<'a, A> {
     module: &'a Module,
     params: &'a CodeActionParams,
     compiler: &'a LspProjectCompiler<A>,
+    pattern_variable_under_cursor: Option<(&'a EcoString, Arc<Type>)>,
     selected_value: Option<PatternMatchedValue<'a>>,
     edits: TextEdits<'a>,
 }
@@ -4468,7 +4469,7 @@ pub enum PatternMatchedValue<'a> {
         function_range: Range,
     },
     LetVariable {
-        variable_name: EcoString,
+        variable_name: &'a EcoString,
         variable_type: Arc<Type>,
         /// The location of the entire let assignment the variable is part of,
         /// so that we can add the pattern matching _after_ it.
@@ -4476,7 +4477,7 @@ pub enum PatternMatchedValue<'a> {
         assignment_location: SrcSpan,
     },
     UseVariable {
-        variable_name: EcoString,
+        variable_name: &'a EcoString,
         variable_type: Arc<Type>,
         /// The location of the entire use expression the variable is part of,
         /// so that we can add the pattern matching _after_ it.
@@ -4500,6 +4501,7 @@ where
             params,
             compiler,
             selected_value: None,
+            pattern_variable_under_cursor: None,
             edits: TextEdits::new(line_numbers),
         }
     }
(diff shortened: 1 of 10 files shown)


lpil:
Drafting as there's failing tests atm!

P.S. may want to use mp4 rather than mov so everyone can play the video

giacomocavalieri:
Oops I forgot a test! It should be all good now

giacomocavalieri:
I fixed the conflicts and this I ready for review now!

compiler-core/src/language_server/code_action.rs
```diff
@@ -4467,16 +4468,16 @@ pub enum PatternMatchedValue<'a> {
...
         /// The location of the entire let assignment the variable is part of,
         /// so that we can add the pattern matching _after_ it.
         ///
         assignment_location: SrcSpan,
     },
     UseVariable {
-        variable_name: &'a EcoString,
-        variable_type: &'a Arc<Type>,
+        variable_name: EcoString,
+        variable_type: Arc<Type>,
```
lpil, line 4481: How come these values are no longer taken by reference?

compiler-core/src/language_server/code_action.rs
```diff
@@ -4855,24 +4952,15 @@ where
...
-                self.selected_value = Some(PatternMatchedValue::LetVariable {
-                    variable_name: name,
-                    variable_type: type_,
-                    assignment_location: assignment.location,
-                });
-                // If we've found the variable to pattern match on, there's no
-                // point in keeping traversing the AST.
-                return;
-            }
+        if let Some((name, type_)) = self.pattern_variable_under_cursor(&assignment.pattern) {
```
lpil: How come custom AST traversal logic has been added instead of using the visitor? Doesn't this mean that we traverse the type AST twice?

compiler-core/src/language_server/code_action.rs
giacomocavalieri: The type I can't really not clone it because later in the string prefix pattern I have to create a `type_::string()` and can't pass its reference around. But you're right, I can avoid the `variable_name` clone now!

compiler-core/src/language_server/code_action.rs
giacomocavalieri: The reason is I didn't notice that 😆
Fixed it!

PR #4809, opened 2025-07-25
Add "collapse nested case" code action

### LINKED ISSUE #4799: Code action to flatten a case expression's branch
Hello! One thing I wished I could do a couple of times is have a code action that would allow me to "flatten" a case expression branch when the only thing it's doing is pattern matching on a variable of the outer pattern... I think it's easier to explain with an example:
```gleam
case wibble {
  Ok(a) ->
//   ^ If I have my cursor here
    case a {
//  ^^^^^^^ And maybe even here
      Wibble -> 1
      Wobble -> 2
      Woo -> 3
    }

  Error(_) -> -1
}
```
It would turn the case expression into:
```gleam
case wibble {
  Ok(Wibble) -> 1
  Ok(Wobble) -> 2
  Ok(Woo) -> 3
  Error(_) -> -1
}
```

giacomocavalieri:
This PR closes #4799

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1530,7 +1530,18 @@ where
 {
     #[must_use]
     pub fn uses_label_shorthand(&self) -> bool {
-        self.label.is_some() && self.location == self.value.location()
+        self.label_shorthand_name().is_some()
+    }
+
+    /// If the call arg is defined using a label shorthand, this will return the
+    /// label name.
+    ///
+    fn label_shorthand_name(&self) -> Option<EcoString> {
+        if !self.is_implicit() && self.location == self.value.location() {
+            self.label.clone()
+        } else {
+            None
+        }
     }
 }
 
@@ -2533,6 +2544,32 @@ impl<A> Pattern<A> {
     }
 }
 
+/// A variable bound inside a pattern.
+#[derive(Debug, Clone)]
+pub enum BoundVariable {
+    /// A record's labelled field introduced with the shorthand syntax.
+    ShorthandLabel { name: EcoString, location: SrcSpan },
+    /// Any other variable.
+    Regular { name: EcoString, location: SrcSpan },
+}
+
+impl BoundVariable {
+    pub fn name(&self) -> EcoString {
+        match self {
+            BoundVariable::ShorthandLabel { name, .. } | BoundVariable::Regular { name, .. } => {
+                name.clone()
+            }
+        }
+    }
+
+    pub fn location(&self) -> SrcSpan {
+        match self {
+            BoundVariable::ShorthandLabel { location, .. }
+            | BoundVariable::Regular { location, .. } => *location,
+        }
+    }
+}
+
 impl TypedPattern {
     pub fn definition_location(&self) -> Option<DefinitionLocation> {
         match self {
@@ -2707,13 +2744,13 @@ impl TypedPattern {
         }
     }
 
-    pub fn bound_variables(&self) -> Vec<(EcoString, SrcSpan)> {
+    pub fn bound_variables(&self) -> Vec<BoundVariable> {
         let mut variables = Vec::new();
         self.collect_bound_variables(&mut variables);
         variables
     }
 
-    fn collect_bound_variables(&self, variables: &mut Vec<(EcoString, SrcSpan)>) {
+    fn collect_bound_variables(&self, variables: &mut Vec<BoundVariable>) {
         match self {
             Pattern::Int { .. }
             | Pattern::Float { .. }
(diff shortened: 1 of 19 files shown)


giacomocavalieri:
I've fixed the conflicts and this is ready for review now!

compiler-core/src/ast.rs
```diff
@@ -2724,7 +2760,14 @@ impl TypedPattern {
             }
             Pattern::Constructor { arguments, .. } => {
                 for argument in arguments {
-                    argument.value.collect_bound_variables(variables);
+                    if argument.uses_label_shorthand() {
+                        variables.push(BoundVariable::ShorthandLabel {
+                            name: argument.label.clone().expect("label for shorthand syntax"),
+                            location: argument.location,
+                        })
```
lpil: 🔴 Never use the check-then-assert pattern when it could be done in a type safe way

compiler-core/src/ast/typed.rs
```diff
@@ -1120,6 +1120,25 @@ impl TypedExpr {
             _ => false,
         }
     }
+
+    pub(crate) fn expression_flattening_block(&self) -> Option<&TypedExpr> {
```
lpil: Document what this does please. 🙏 

The name doesn't seem to match what it does, and it seems very specific to `CollapseNestedCase`, so I'm not convinced that this belongs as a method on the generic `TypedExpr` class.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7562,3 +7562,374 @@ impl<'ast> ast::visit::Visit<'ast> for RemovePrivateOpaque<'ast> {
...
+            //   Ok(a) -> case a {
+            // //^^^^^^^^^^^^^^^ Anywhere over here!
+            //   }
+            // }
+            // ```
+            //
+            let pattern_location = pattern
+                .first()
+                .map_or(*location, |first| first.location())
+                .merge(&pattern.last().map_or(*location, |end| end.location()));
```
lpil: Sorry, might be being slow here. When is it that there might not be any patterns?

compiler-core/src/language_server/code_action.rs
```diff
@@ -7562,3 +7562,374 @@ impl<'ast> ast::visit::Visit<'ast> for RemovePrivateOpaque<'ast> {
...
+                self.collapsed = Some(Collapsed {
+                    outer_clause_span: *location,
+                    outer_guard: guard,
+                    matched_variable: variable,
+                    matched_pattern_span: pattern_location,
+                    inner_clauses: clauses,
+                });
+                // We're done, there's no need to keep exploring as we know the cursor
+                // is over this pattern and it can't be over any other one!
+                return;
```
lpil: This means that nested further cases don't get checked, even though they could also be collapsed.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7562,3 +7562,374 @@ impl<'ast> ast::visit::Visit<'ast> for RemovePrivateOpaque<'ast> {
...
+        //
+        // For example:
+        //
+        // ```gleam
+        // Wibble(a, b, 1) -> case a { ... }
+        // Wibble(a, b, 1) -> case b { ... }
+        // ```
+
+        // We can't have any alternative patterns...
+        if alternative_patterns.is_empty()
```
lpil: Rather than complex conditionals and progressively more indentation could you extract this all to a helper function and use early returns please 🙏

lpil (commented):
Super nice! Left a few little notes inline

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Yeah it will flatten one expression at a time, so you get to decide what to do with the nested ones

compiler-core/src/ast/typed.rs
giacomocavalieri: You're right, I'm moving it to the code action module with a better name!

compiler-core/src/ast.rs
giacomocavalieri: Oops, fixed!

compiler-core/src/language_server/code_action.rs
giacomocavalieri: I was being cautious, since first return an option. But I could just `expect` it!

compiler-core/src/language_server/code_action.rs
lpil: ah! I misunderstood the code before. I feel vindicated on my request to make that helper function now 😁

giacomocavalieri:
Fixed the conflicts, seems like there's a spurious failure in CI, but tests should be good!

lpil (approved):
Amazing! Thank you!!

giacomocavalieri:
Done!

PR #4810, opened 2025-07-28
Add suggestions for variables with the same name in imported modules when unknown variable

### LINKED ISSUE #2697: Compiler Does Not Suggest Functions from Imported Packages for Unknown Variables/Functions

realraphrous:
closes #2697

As stated on this [Discord message](https://discord.com/channels/768594524158427167/1396106606047133809/1396106606047133809), this is my first contributions to OSS and this project. I would really appreciate any feedback and improvement on my code, this pull request or commits that I made !

## Summary

This PR add more suggestions when an unknown variable is encounter by listing all variables / identifier in imported modules that match the same name. Also, when the variable / identifier is called such as functions, pipeline steps or type constructors, the suggestions are based on the name and number of arguments provided to the call. (ref: https://github.com/gleam-lang/gleam/issues/2697#issuecomment-1997241885)

## Examples

For example, take this invalid program:
```gleam
import gleam/io

pub fn main() -> Nil {
  println("Hello, World!)
}
```

The error message reported by the compiler:
```text
error: Unknown variable
  ┌─ /home/raphrous/code/gleam/module_suggestion/src/mylib/mymodule.gleam:4:3
  │
4 │   println("Hello, World!")
  │   ^^^^^^^

The name `println` is not in scope here.
Consider using one of these variables:

    io.println
```

Also, it works with variables, type constructor, constant etc. Consider this custom module and this program:
```gleam
// mylib/mymodule
pub const my_const = 8

pub type MyType {
  Some(x: Int, y: Int)
}

// main file
import gleam/option
import mylib/mymodule

pub fn main() -> Nil {
  // option.Some need to be suggested but 
  // not mymodule.Some (cause of arity)
  Some(my_const)
}
```

With these changes we have new suggestions about variables / identifier in imported modules:
```text
error: Unknown variable
  ┌─ /path/to/project/src/project.gleam:7:3
  │
7 │   Some(my_const)
  │   ^^^^

The custom type variant constructor `Some` is not in scope here.
Consider using one of these variables:

    option.Some


error: Unknown variable
  ┌─ /path/to/project/src/project.gleam:7:8
  │
7 │   Some(my_const)
  │        ^^^^^^^^

The name `my_const` is not in scope here.
Consider using one of these variables:

    mymodule.my_const
```

## Implementations

The current implementation uses a new enumeration `VarUsage`. This is similar to the enum `FieldAccessUsage` and it helps the function `report_name_error` by giving a little more context about the usage of this variable: a function call or something else. This function can now report about imported modules containing a variable with the same name as the unknown variable and, in case of function call, the same arity.

I tried to experiment another implementation where the function `report_name_error` was not provided with more context. The function was keeped as simple as possible by reporting imported modules containing a variable with the same name, without checking if it was a call or something else. Instead, we were filtering at the call side, inside `infer_call` function for example, if the variable from the imported module was of type "function-like". 
Honestly it does not feel right. We needed to check if an `UnknownVariable` error was emitted, apply a filtering on imported modules that were reported and re-emit the same error with the filter applied. Also, we had to perform this manipulation at every location where a function call was possible, which likely implies code duplication, errors and omissions.

### Wording

First thing first, the wording. I am not sure about the error message wording:

- `Consider using one of these variables`

For many developers, `io.println` is a function and `uri.empty` is a constant, not really variables. I wanted to make the error message like:

- `Consider using on of these identifiers`

But the term `identifiers` is not used in the compiler, in error messages that I found or the language tour so it does not feel right. As I can see in the compiler, `Var` refer to what I call `identifier` so I thought it would be best to write `variables`.

### No type checking on arguments

Because we only check name and arity in the case of calls, the suggestions may be wrong because the type of the function parameters does not match the type of arguments provided with the call. If we want that, we need to give more context, such as arguments types, to the `report_name_error` function in order to find the right functions to suggest.

### Suggestions for pipeline calls like `1 |> add(2)`

Also, I had a problem with function call on pipeline. To better explain the situation, take this program:
```gleam
pub fn main() -> Nil {
  1 |> add(2)
}
```

It can be desugared in two ways:
```gleam
pub fn main() -> Nil {
  // First possibility
  add(1, 2)
  // Second possibility
  add(2)(1)
}
```

I didn't know how to implement this case. For sure, the compiler can report function with arity of 2 and a function with arity 1 that has a return type that is also a function with arity 1. I just didn't know what was the right choice ... 

### Tests

To conclude, theses commits does not add tests for now, it might be cool to have some !

## Ending

Thank you in advance for reading and helping me on my journey into OSS. Please feel free to give me feedback on all this work! 💛

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,35 @@
 
 ### Compiler
 
+- The compiler now suggest public values from imported modules when the variable
+  in unknown. These values are suggested based on name and arity.
+
+  Considering this program:
+
+  ```gleam
+  import gleam/io
+
+  pub fn main() -> Nil {
+    println("Hello, World!")
+  }
+  ```
+
+  The compiler will display this error message:
+  ```text
+    error: Unknown variable
+    ┌─ /path/to/project/src/project.gleam:4:3
+    │
+  4 │   println("Hello, World!")
+    │   ^^^^^^^
+
+  The name `println` is not in scope here.
+  Did you mean one of these:
+
+      - io.println
+  ```
+
+  ([raphrous](https://github.com/realraphrous))
+
 - The inference of record update expressions is now more fault tolerant: if
   there's an error in the record being updated, the compiler can still able to
   analyse the fields that are being provided.
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -3320,6 +3320,7 @@ but no type in scope with that name."
             discarded_location,
             name,
             type_with_name_in_scope,
+            possible_modules,
         } => {
             let title = String::from("Unknown variable");
 
@@ -3352,17 +3353,24 @@ but no type in scope with that name."
                 let text = if *type_with_name_in_scope {
                     wrap_format!("`{name}` is a type, it cannot be used as a value.")
                 } else {
-                    let is_first_char_uppercase =
-                        name.chars().next().is_some_and(char::is_uppercase);
-
-                    if is_first_char_uppercase {
+                    let mut text = if name.starts_with(char::is_uppercase) {
                         wrap_format!(
-                            "The custom type variant constructor \
-`{name}` is not in scope here."
+                            "The custom type variant constructor `{name}` is not in scope here."
                         )
                     } else {
                         wrap_format!("The name `{name}` is not in scope here.")
+                    };
+
+                    // If there are some suggestions about public values in imported
+                    // modules put a "did you mean" text after the main message
+                    if !possible_modules.is_empty() {
+                        text.push_str("\nDid you mean one of these:\n\n");
+                        for module_name in possible_modules {
+                            text.push_str(&format!("  - {module_name}.{name}\n"))
+                        }
                     }
+
+                    text
                 };
 
                 Diagnostic {
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -1105,7 +1105,42 @@ pub struct TypeValueConstructorField {
 }
 
 impl ModuleInterface {
+    /// Search for a public value that matches the given `name`
+    ///
     pub fn get_public_value(&self, name: &str) -> Option<&ValueConstructor> {
+        let value = self.values.get(name)?;
+        if value.publicity.is_public() {
+            Some(value)
+        } else {
+            None
+        }
+    }
+
+    /// Search for a public value that matches the given `name` and `arity`
+    ///
+    pub fn get_public_function(&self, name: &str, arity: usize) -> Option<&ValueConstructor> {
+        self.get_public_value(name).filter(|value_constructor| {
+            match value_constructor.type_.fn_types() {
+                Some((fn_arguments_type, _)) => fn_arguments_type.len() == arity,
+                None => false,
+            }
+        })
+    }
+
+    /// Search for an importable type that matches the given `name`
+    ///
+    pub fn get_public_type(&self, name: &str) -> Option<&TypeConstructor> {
+        let type_ = self.types.get(name)?;
+        if type_.publicity.is_public() {
+            Some(type_)
+        } else {
+            None
+        }
+    }
+
+    /// Search for an importable value that matches the given `name`
+    ///
+    pub fn get_importable_value(&self, name: &str) -> Option<&ValueConstructor> {
         let value = self.values.get(name)?;
         if value.publicity.is_importable() {
             Some(value)
@@ -1114,7 +1149,20 @@ impl ModuleInterface {
         }
     }
 
-    pub fn get_public_type(&self, name: &str) -> Option<&TypeConstructor> {
+    /// Search for an importable function that matches the given `name` and `arity`
+    ///
+    pub fn get_importable_function(&self, name: &str, arity: usize) -> Option<&ValueConstructor> {
+        self.get_importable_value(name).filter(|value_constructor| {
+            match value_constructor.type_.fn_types() {
+                Some((fn_arguments_type, _)) => fn_arguments_type.len() == arity,
+                None => false,
+            }
+        })
+    }
+
+    /// Search for an importable type that matches the given `name`
+    ///
+    pub fn get_importable_type(&self, name: &str) -> Option<&TypeConstructor> {
         let type_ = self.types.get(name)?;
         if type_.publicity.is_importable() {
             Some(type_)
@@ -1701,6 +1749,15 @@ pub enum FieldAccessUsage {
     ///
     RecordUpdate,
     /// Used as `thing.field`
+    Other,
+}
+
+/// This is used to know when a value is used as a call or not.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ValueUsage {
+    /// Used as `function(..)`, `Record(..)`, `left |> right` or `left |> right(..)`
+    Call { arity: usize },
+    /// Used as `variable`
     Other,
 }
 
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -468,7 +468,7 @@ impl Environment<'_> {
                 })?;
                 self.references
                     .register_module_reference(module_name.clone());
-                module.get_public_type(name).ok_or_else(|| {
+                module.get_importable_type(name).ok_or_else(|| {
                     UnknownTypeConstructorError::ModuleType {
                         name: name.clone(),
                         module_name: module.name.clone(),
@@ -534,14 +534,15 @@ impl Environment<'_> {
         name: &EcoString,
     ) -> Result<&ValueConstructor, UnknownValueConstructorError> {
         match module {
-            None => self.scope.get(name).ok_or_else(|| {
-                let type_with_name_in_scope = self.module_types.keys().any(|type_| type_ == name);
-                UnknownValueConstructorError::Variable {
+            None => self
+                .scope
+                .get(name)
+                .ok_or_else(|| UnknownValueConstructorError::Variable {
                     name: name.clone(),
                     variables: self.local_value_names(),
-                    type_with_name_in_scope,
-                }
-            }),
+                    type_with_name_in_scope: self.module_types.contains_key(name),
+                    possible_modules: self.get_possible_modules_with_value(name),
+                }),
 
             Some(module_name) => {
                 let (_, module) = self.imported_modules.get(module_name).ok_or_else(|| {
@@ -553,7 +554,7 @@ impl Environment<'_> {
                 })?;
                 self.references
                     .register_module_reference(module_name.clone());
-                module.get_public_value(name).ok_or_else(|| {
+                module.get_importable_value(name).ok_or_else(|| {
                     UnknownValueConstructorError::ModuleValue {
                         name: name.clone(),
                         module_name: module.name.clone(),
@@ -563,6 +564,46 @@ impl Environment<'_> {
                 })
             }
         }
+    }
+
+    /// Get the name of modules with a value that can be suggested and matches
+    /// the given `name`
+    ///
+    pub fn get_possible_modules_with_value(&self, name: &EcoString) -> Vec<EcoString> {
+        self.imported_modules
+            .iter()
+            .filter_map(|(module_name, (_, module))| {
+                if module.package == self.current_package {
+                    module.get_importable_value(name).map(|_| module_name)
+                } else {
+                    module.get_public_value(name).map(|_| module_name)
+                }
+            })
+            .cloned()
+            .collect_vec()
+    }
+
+    /// Get the name of modules with a function that can be suggested and
+    /// matches the given `name` and `arity`
+    ///
+    pub fn get_possible_modules_with_function(
+        &self,
+        name: &EcoString,
+        arity: usize,
+    ) -> Vec<EcoString> {
+        self.imported_modules
+            .iter()
+            .filter_map(|(module_name, (_, module))| {
+                if module.package == self.current_package {
+                    module
+                        .get_importable_function(name, arity)
+                        .map(|_| module_name)
+                } else {
+                    module.get_public_function(name, arity).map(|_| module_name)
+                }
+            })
+            .cloned()
+            .collect_vec()
     }
 
     pub fn get_type_variants_fields(
@@ -591,7 +632,7 @@ impl Environment<'_> {
             self.scope.get(&variant.name)
         } else {
             let (_, module) = self.imported_modules.get(module)?;
-            module.get_public_value(&variant.name)
+            module.get_importable_value(&variant.name)
         }
     }
 
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -166,6 +166,9 @@ pub enum Error {
         /// this will contain its location.
         discarded_location: Option<SrcSpan>,
         type_with_name_in_scope: bool,
+        /// Filled with the name of imported modules when the module has public value
+        /// with the same name as this variable
+        possible_modules: Vec<EcoString>,
     },
 
     UnknownType {
@@ -1456,6 +1459,7 @@ pub enum UnknownValueConstructorError {
         name: EcoString,
         variables: Vec<EcoString>,
         type_with_name_in_scope: bool,
+        possible_modules: Vec<EcoString>,
     },
 
     Module {
@@ -1481,12 +1485,14 @@ pub fn convert_get_value_constructor_error(
             name,
             variables,
             type_with_name_in_scope,
+            possible_modules,
         } => Error::UnknownVariable {
             location,
             name,
             variables,
             discarded_location: None,
             type_with_name_in_scope,
+            possible_modules,
         },
 
         UnknownValueConstructorError::Module { name, suggestions } => Error::UnknownModule {
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -427,9 +427,12 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 message,
             } => Ok(self.infer_echo(location, keyword_end, expression, message)),
 
-            UntypedExpr::Var { location, name, .. } => {
-                self.infer_var(name, location, ReferenceRegistration::Register)
-            }
+            UntypedExpr::Var { location, name, .. } => self.infer_var(
+                name,
+                location,
+                ValueUsage::Other,
+                ReferenceRegistration::Register,
+            ),
 
             UntypedExpr::Int {
                 location,
@@ -856,6 +859,14 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         // We use `stacker` to prevent overflowing the stack when many `use`
         // expressions are chained. See https://github.com/gleam-lang/gleam/issues/4287
         let infer_call = || {
+            // We need this in the case where `call.function` has a special call path depending
+            // on the type such as `UntypedExpr::Var`. In these cases, `infer_call` does not call
+            // `infer_or_error`. `infer_or_error` is responsible for registering warnings about
+            // unreachable code and thus, warnings about unreachable code are not registered.
+            if self.previous_panics {
+                self.warn_for_unreachable_code(call_location, PanicPosition::PreviousExpression);
+            }
+
             self.infer_call(
                 *call.function,
                 call.arguments,
@@ -1283,10 +1294,16 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         &mut self,
         name: EcoString,
         location: SrcSpan,
+        value_usage: ValueUsage,
         register_reference: ReferenceRegistration,
     ) -> Result<TypedExpr, Error> {
-        let constructor =
-            self.do_infer_value_constructor(&None, &name, &location, register_reference)?;
+        let constructor = self.do_infer_value_constructor(
+            &None,
+            &name,
+            &location,
+            value_usage,
+            register_reference,
+        )?;
         self.narrow_implementations(location, &constructor.variant)?;
         Ok(TypedExpr::Var {
             constructor,
@@ -1372,7 +1389,12 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             // If the left-hand-side of the record access is a variable, this might actually be
             // module access. In that case, we only want to register a reference to the variable
             // if we actually referencing it in the record access.
-            self.infer_var(name, location, ReferenceRegistration::DoNotRegister)
+            self.infer_var(
+                name,
+                location,
+                ValueUsage::Other,
+                ReferenceRegistration::DoNotRegister,
+            )
         } else {
             self.infer_or_error(container)
         };
@@ -2483,7 +2505,8 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
     fn infer_clause_guard(&mut self, guard: UntypedClauseGuard) -> Result<TypedClauseGuard, Error> {
         match guard {
             ClauseGuard::Var { location, name, .. } => {
-                let constructor = self.infer_value_constructor(&None, &name, &location)?;
+                let constructor =
+                    self.infer_value_constructor(&None, &name, &location, ValueUsage::Other)?;
 
                 // We cannot support all values in guard expressions as the BEAM does not
                 let (definition_location, origin) = match &constructor.variant {
@@ -2803,13 +2826,13 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
 
             let constructor =
                 module
-                    .get_public_value(&label)
+                    .get_importable_value(&label)
                     .ok_or_else(|| Error::UnknownModuleValue {
                         name: label.clone(),
                         location: select_location,
                         module_name: module.name.clone(),
                         value_constructors: module.public_value_names(),
-                        type_with_same_name: module.get_public_type(&label).is_some(),
+                        type_with_same_name: module.get_importable_type(&label).is_some(),
                         context: ModuleValueUsageContext::ModuleAccess,
                     })?;
 
@@ -3516,8 +3539,15 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         module: &Option<(EcoString, SrcSpan)>,
         name: &EcoString,
         location: &SrcSpan,
+        value_usage: ValueUsage,
     ) -> Result<ValueConstructor, Error> {
-        self.do_infer_value_constructor(module, name, location, ReferenceRegistration::Register)
+        self.do_infer_value_constructor(
+            module,
+            name,
+            location,
+            value_usage,
+            ReferenceRegistration::Register,
+        )
     }
 
     fn do_infer_value_constructor(
@@ -3525,6 +3555,7 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         module: &Option<(EcoString, SrcSpan)>,
         name: &EcoString,
         location: &SrcSpan,
+        value_usage: ValueUsage,
         register_reference: ReferenceRegistration,
     ) -> Result<ValueConstructor, Error> {
         let constructor = match module {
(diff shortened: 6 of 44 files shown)


lpil:
Hi @realraphrous! Thanks for making us your first open source PR 🎉 

Sorry for the delay reviewing this, I have been busy with the v1.12 release. Looking now

CHANGELOG.md
```diff
@@ -1,5 +1,38 @@
 # Changelog
 
+## Unreleased
+
+### Compiler
+
+- The compiler now suggest public variables and functions from imported modules
+  when the variable in unknown. These variables and functions are suggested
+  based on name and arity.
```
lpil: Nice, very clear explanation. Will be useful for me when I write the release post too!

Let's say "values" rather than "functions and variables" in this changelog.

compiler-core/src/type_/error.rs
```diff
@@ -171,6 +171,9 @@ pub enum Error {
         /// this will contain its location.
         discarded_location: Option<SrcSpan>,
         type_with_name_in_scope: bool,
+        /// Filled with the name of imported modules when the module has public value
+        /// with the same name as this variable
+        imported_modules_with_same_public_variable_name: Vec<EcoString>,
```
lpil: Descriptive, but a little long! Perhaps `module_candidates`? `possible_modules`? Something like that?

compiler-core/src/error.rs
```diff
@@ -2433,17 +2433,32 @@ but no type in scope with that name."
...
                                     wrap_format!("The name `{name}` is not in scope here.")
+                                };
+
+                                // If there are some suggestions about variables in imported modules
+                                // put a "consider" text after the main message
+                                if !imported_modules_with_same_public_variable_name.is_empty() {
+                                    let consider_text = imported_modules_with_same_public_variable_name
+                                        .iter()
+                                        .fold(
+                                            String::from("Consider using one of these variables:\n\n"),
```
lpil: They're not really variables as they're definitions in a module. How about this?

```txt
Did you mean one of these?

  - list.map
  - dict.map
```
I think we tend to use the `  - ` prefix with lists rendered in this module.

compiler-core/src/error.rs
```diff
@@ -2433,17 +2433,32 @@ but no type in scope with that name."
...
-                                } else {
+                                }
+                                else {
                                     wrap_format!("The name `{name}` is not in scope here.")
+                                };
+
+                                // If there are some suggestions about variables in imported modules
+                                // put a "consider" text after the main message
+                                if !imported_modules_with_same_public_variable_name.is_empty() {
+                                    let consider_text = imported_modules_with_same_public_variable_name
```
lpil: Little stylistic thing: Can we use a `for` loop here rather than `fold` here please 🙏 

The accumulator is being mutated, so prefer not to use the immutable accumulator passing pattern of `fold`.

compiler-core/src/error.rs
```diff
@@ -2433,17 +2433,32 @@ but no type in scope with that name."
...
+                                        .iter()
+                                        .fold(
+                                            String::from("Consider using one of these variables:\n\n"),
+                                            |mut acc, module_name| {
+                                                acc.push_str(&wrap_format!("    {module_name}.{name}\n"));
```
lpil: Do not wrap these items please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -724,7 +727,7 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
     }
 
     // Helper to create a new error expr.
-    fn error_expr(&mut self, location: SrcSpan) -> TypedExpr {
+    pub(crate) fn error_expr(&mut self, location: SrcSpan) -> TypedExpr {
```
lpil: If this is being made public could you add more information to the documentation please, to explain what it does and when it is used.

A more descriptive name would be useful too. `create_invalid_expression`?

compiler-core/src/type_/pipe.rs
```diff
@@ -139,17 +139,24 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
                     location,
                     ..
                 } => {
-                    let fun = match self.expr_typer.infer_or_error(*fun) {
+                    let fun_result = match *fun {
+                        UntypedExpr::Var { location, name } => self.expr_typer.infer_var(
+                            name,
+                            location,
+                            VarUsage::PipelineCall,
```
lpil: Rather than have a special case for piped functions, let's use `arguments.len() + 1` as the arity of the function, as that's by far the most common use of pipes.

compiler-core/src/type_.rs
```diff
@@ -1662,6 +1662,18 @@ pub enum FieldAccessUsage {
     Other,
 }
 
+/// This is used to know when a variable is used as a value or as a call:
+/// function call, a pipeline or a custom type variant constructor
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum VarUsage {
```
lpil: The name says it's specific to variables, but the definition is a lot more generic! Any expression could could produce a function which is called or not.

Perhaps this is a `ValueUsage`?

compiler-core/src/type_/expression.rs
```diff
@@ -3513,21 +3532,75 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                let imported_modules_with_same_public_variable_name = match var_usage {
+                    // This is a function call, we need to suggest a public
+                    // variable which is a function with the correct arity
+                    VarUsage::Call { arity } => self
+                        .environment
+                        .imported_modules
+                        .iter()
+                        .filter_map(|(module_name, (_, module))| {
+                            module
+                                .get_public_value(name)
```
lpil: Perhaps this is a `get_public_function(name, arity)`?

compiler-core/src/type_/expression.rs
```diff
@@ -1135,14 +1138,20 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         }
     }
 
-    fn infer_var(
+    pub(crate) fn infer_var(
```
lpil: I'm not the biggest fan of how this function has become part of the public API. The internal ReferenceRegistration being something that the rest of the code needs to know about is something I'd like to avoid- it's not very clear what it does or how to use it.

Maybe introduce a public `infer_called_var` function that takes the arity as an argument, leaving the existing functions with the same API as before.

compiler-core/src/type_/expression.rs
```diff
@@ -3943,6 +4025,30 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         kind: CallKind,
     ) -> (TypedExpr, Vec<TypedCallArg>, Arc<Type>) {
         let fun = match fun {
+            UntypedExpr::Var { location, name } => {
+                // Because we are not calling infer / infer_or_error directly,
+                // we do not warn if there was a previous panic. Check for that
+                // here. Not perfect, but works.
```
lpil: Why do we not need to do this for `infer_field_access` etc below?

What does the "not perfect bit" mean?

lpil (commented):
Looking great! Very nice work, especially for a first contribution.

I've left a bunch of small notes inline.

Could you add tests for all the different paths through this new feature please 🙏 

RE naming: You've been using the words identifier and variable here to refer to public constants and functions in other modules. We don't have a meaning for the word identifier, and a variable is a local variable defined with `let`, so it's something else. "values" fits generally.

Thanks again! Please un-draft this PR and ping me when you want another review, and do reply to any comments of mine that are unclear or need discussion.

compiler-core/src/type_/expression.rs
realraphrous: Without this kind of mitigation, the test `unreachable_use_after_panic` failed, no warning are emitted:
```gleam
pub fn wibble(_) { 1 }
pub fn main() {
    panic
    use <- wibble
    1
}
```

I checked, there is no test for a similar program and, when tested, there is no warn for unreachable code:
```gleam
// define: pub fn wibble(_) { 1 }
import mylibrary/mymodule

pub fn main() {
    panic
    use <- mymodule.wibble
    1
}
```

Basically, I wanted to just pass this test, but it seems that there is a problem with `use` expression.

Edit: It seems that `previous_panics` is set to false in `do_infer_call_with_known_fun` before the warning is registered. We have this call path :

1. `infer_use`
2. `infer_call`
3. `do_infer_call` (depending on expression)
  3.1. UntypedExpr::Var -> `infer_called_var`
  3.2. UntypedExpr::FieldAccess -> `infer_field_access`
  3.3. UntypedExpr::Fn -> `infer_fn_with_call_context`
  3.4. Others -> `infer` -> `infer_or_error`
4. `do_infer_call_with_known_fun`

But, warnings for unreachable code are mainly registered from `infer_or_error`. With these commits, I added the case for `UntypedExpr::Var` in `do_infer_call` and I am not going anymore in `infer_or_error`. So I broke one test by not giving warning for the first case as stated above. It also explains why I don't have warnings in the second example.

Honnestly, I don't know where to correct that, there is too much possibilities. Did you have advice for that ?

realraphrous: @lpil I don't know if I need to ping you for comment, really sorry if not. I does not know anything about Github ...

realraphrous:
Hi @lpil !
I have implemented all your suggestions in [this commit](https://github.com/gleam-lang/gleam/pull/4810/commits/29ec32d66cafc9f2c6f8644eb21153e4f5d367bf).
Also, I have added tests in [this commit](https://github.com/gleam-lang/gleam/pull/4810/commits/3e2fc4d1d3fd2c495af839f70d226af8a97d10fa).

compiler-core/src/type_/expression.rs
realraphrous: In my last iteration, I do not need to make that function public anymore !

compiler-core/src/type_/expression.rs
realraphrous: I have done that, `infer_var` stay private and I have introduced `infer_called_var` has a public function.

realraphrous:
Really sorry @lpil, did not saw that, with force-push, the review become tough ... I'm using Jujutsu and this is so common to just modify prior commit ...
Last changes were just a rebase on main

lpil:
Looks like the build is failing!

realraphrous:
This is failling cause of a Deno fetching error : `Failed fetching. Retrying in 4000ms...`
Is that possible to relaunch the pipeline without a new commit @lpil ?

realraphrous:
@lpil, I amended the last commit without any edit just to relaunch the pipeline. This was a Deno fetching issue for sure, now everything is OK !

CHANGELOG.md
```diff
@@ -4,6 +4,34 @@
...
+    error: Unknown variable
+    ┌─ /path/to/project/src/project.gleam:4:3
+    │
+  4 │   println("Hello, World!")
+    │   ^^^^^^^
+
+  The name `println` is not in scope here.
+  Consider using one of these variables:
+
+      io.println
```
GearsDatapacks: Looks like this entry needs to be updated with the latest wording of the error message

compiler-core/src/error.rs
```diff
@@ -2434,17 +2434,27 @@ but no type in scope with that name."
...
+
+                                // If there are some suggestions about public values in imported
+                                // modules put a "did you mean" text after the main message
+                                if !possible_modules.is_empty() {
+                                    let mut did_you_mean_text = String::from("Did you mean one of these:\n\n");
+                                    for module_name in possible_modules {
+                                        did_you_mean_text.push_str(&format!("  - {module_name}.{name}\n"));
+                                    }
+                                    text.push('\n');
+                                    text.push_str(&did_you_mean_text);
```
GearsDatapacks: Any reason you are building `did_you_mean_text` then pushing to `text` rather than just appending directly to `text` here?

compiler-core/src/type_/tests/errors.rs
```diff
@@ -3308,3 +3308,223 @@ pub fn main() {
...
+"
+        ),
+        "
+import module
+pub fn main() {
+    use <- wibble
+    1
+}
+"
+    );
```
GearsDatapacks, line 4205: Could we get tests for:
- Modules imported using an alias (`import some_module as some_alias`)
- Modules which have multiple segments (`import some/module`)

GearsDatapacks (commented):
This looks good! I've left a couple more notes inline, but this can probably be merged soon.

CHANGELOG.md
realraphrous: Thanks, had not spotted that !

compiler-core/src/error.rs
realraphrous: No reason at all. I will change that to push to the current `text` variable.

compiler-core/src/type_/tests/errors.rs
realraphrous: Ofc

GearsDatapacks (approved):
This looks great to me! Louis still needs to give it another look over before it's merged.

compiler-core/src/type_/expression.rs
```diff
@@ -4100,6 +4168,26 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         (fun, arguments, type_)
     }
 
+    pub fn infer_called_var(
```
lpil, line 4673: Could you add a detailed documentation comment explaining what this function doe and when it is to be used please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -3633,21 +3660,49 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    // a public value of any type
+                    ValueUsage::Other => self
+                        .environment
+                        .imported_modules
+                        .iter()
+                        .filter_map(|(module_name, (_, module))| {
+                            module.get_public_value(name).map(|_| module_name)
+                        })
+                        .cloned()
+                        .collect_vec(),
```
lpil: This block of code is repeated a few times. Could you extract it and the block above to a helper function or method please.

Looks like one of the places this block exists is in the environment class, so maybe that's a good place for it.

compiler-core/src/type_.rs
```diff
@@ -1680,6 +1689,15 @@ pub enum FieldAccessUsage {
     Other,
 }
 
+/// This is used to know when a value is used as a call or not.
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum ValueUsage {
+    /// Used as `call(..)`, `Type(..)`, `left |> right` or `left |> right(..)`
```
lpil:
```suggestion
    /// Used as `function(..)`, `Record(..)`, `left |> right` or `left |> right(..)`
```

It's not a type, it's value, specifically a record.

compiler-core/src/type_.rs
```diff
@@ -1103,6 +1103,15 @@ impl ModuleInterface {
         }
     }
 
+    pub fn get_public_function(&self, name: &str, arity: usize) -> Option<&ValueConstructor> {
```
lpil, line 1121: Document this function please 🙏 

Does this return internal values too?

lpil (commented):
Super nice! The approach in the code looks flawless to me!

Couple tiny notes inline RE documentation and extracting a helper, and could you add these final tests please:

- Internal values from the same package are suggested
- Internal values from a different package are not suggested

compiler-core/src/type_.rs
realraphrous: Yes, this function also return internal values. This is based on the function `get_public_value` that check for importable values, and importable values can be public and internal. I named this function like that to be consistent with other function of the same kind.

Also gonna document this function but `get_public_value`, on which this function is based, is not documented ...

lpil:

realraphrous:
Hej,
Sorry for this end of the year, i couldn't work on it for a while due to a major change in my life (a positive one !). Would it be possible to continue with it, but at the end of the year?

compiler-core/src/type_/pipe.rs
```diff
@@ -346,7 +373,33 @@ impl<'a, 'b, 'c> PipeTyper<'a, 'b, 'c> {
...
+            | UntypedExpr::Tuple { .. }
+            | UntypedExpr::TupleIndex { .. }
+            | UntypedExpr::Todo { .. }
+            | UntypedExpr::Panic { .. }
+            | UntypedExpr::Echo { .. }
+            | UntypedExpr::BitArray { .. }
+            | UntypedExpr::RecordUpdate { .. }
+            | UntypedExpr::NegateBool { .. }
+            | UntypedExpr::NegateInt { .. } => self.expr_typer.infer(function),
+        });
```
lpil: This block of code exists twice, extract it to a function please 🙏  Perhaps `infer_called`?

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_suggest_modules_from_the_same_package_with_internal_function.snap
```diff
@@ -0,0 +1,30 @@
...
+error: Unknown variable
+  ┌─ /src/one/two.gleam:4:5
+  │
+4 │     add(1, 1)
+  │     ^^^
+
+The name `add` is not in scope here.
+Did you mean one of these:
+
+  - module.add
```
lpil, line 33: Is this test showing that internal values from the same package can be suggested? If so, can you make it clear in this test that this is the case please 🙏

compiler-core/src/type_/tests/errors.rs
```diff
@@ -3399,3 +3399,321 @@ pub fn main() {
...
+    x + y
+}"
+        ),
+        "
+import module
+pub fn main() {
+    add(1, 1)
+}"
+    );
+}
```
lpil, line 4206: Could we get test for internal record constructors and internal types please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -846,6 +849,14 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         // We use `stacker` to prevent overflowing the stack when many `use`
         // expressions are chained. See https://github.com/gleam-lang/gleam/issues/4287
         let infer_call = || {
+            // We need this in the case where `call.function` has a special call path depending
+            // on the type such as `UntypedExpr::Var`. In these cases, `infer_call` does not call
+            // `infer_or_error`. `infer_or_error` is responsible for registering warnings about
+            // unreachable code and thus, warnings about unreachable code are not registered.
+            if self.previous_panics {
+                self.warn_for_unreachable_code(call_location, PanicPosition::PreviousExpression);
+            }
```
lpil, line 868: Won't think result in the warning being emitted twice when it's not a `Var`?

How come we don't do this in that path?

compiler-core/src/type_.rs
lpil: Let's call them `get_importable_*` if it permits all importable ones, not just public ones. Noting in the doc comment that it will return internal ones would be good too. 🙏

lpil (commented):
Lovely, thank you!! I've left some small notes inline

compiler-core/src/type_.rs
realraphrous: That was done in this commit, the diff is not very clear about it.
Now `get_public_function`  call `get_public_value` (also rewritten in this commit) that retrieve only public values.
The originals functions are all renamed to `get_importable_*` to better represent what they do (retrieving public and internal values) :

- `get_public_value` to `get_importable_value`
- `get_public_function` to `get_importable_function`
- `get_public_type` to `get_importable_type`

compiler-core/src/type_/expression.rs
realraphrous: With the current version of compiler (without this PR), warning in `use` expressions are not well done. As an example, this code emit a warning on `wibble` because there is an expression after a panic.

```gleam
fn wibble(_) { 1 }
pub fn wobble() {
  panic
  use <- wibble
  1
}
```

But the code below does not.

```gleam
import module
pub fn wobble() {
  panic
  use <- module.wibble
  1
}
```

And the code below emit a warning, but not the best one.

```gleam
pub fn wobble() {
  panic
  use <- fn(_) { 2 }
  1
}

warning: Unreachable code
  ┌─ /path/to/source.gleam:4:3
  │
4 │   1
  │   ^
```

This is because, when you follow the code (a -> b = a call b): `infer_use` -> `infer_call` -> `do_infer_call`.

In `do_infer_call`, there is 3 possibilities depending on which kind of `UntypedExpr` the expression is :

- `UntypedExpr::FieldAccess`: we are calling `infer_field_access`
- `UntypedExpr::Fn`: we are calling `infer_fn_with_call_context`
- Others: we are calling `infer` -> `infer_or_error`

In almost all cases `infer_or_error` is responsible for warning about an expression after a panic but in the first two cases, we never enter in `infer_or_error`.

With this PR, i added a new special case in `do_infer_call` for `UntypedExpr::Var` and i broke the first example which was emitting correct warnings. It seems to me that warnings in `use` expressions are very fragile. With these changes, we can be sure not to miss any warnings.

Also it can't be warned twice because `warn_for_unreachable_code` checks if a warning was already emitted in the current block.

```rust
if !self.already_warned_for_unreachable_code {
    self.already_warned_for_unreachable_code = true;
    ...
}
```

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_suggest_modules_from_the_same_package_with_internal_function.snap
realraphrous: Done, I have renamed all tests for better naming !

compiler-core/src/type_/tests/errors.rs
realraphrous: Done, more tests !

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_do_not_suggest_modules_with_public_function_with_incorrect_arity_2.snap
```diff
@@ -0,0 +1,26 @@
...
+}
+
+----- ERROR
+error: Unknown variable
+  ┌─ /src/one/two.gleam:4:10
+  │
+4 │     1 |> add(2)
+  │          ^^^
+
+The name `add` is not in scope here.
```
lpil, line 31: How come this one doesn't suggest a fix?

lpil (commented):
Hello! Some of the snapshot tests show that nothing is suggested, but I couldn't work out why. Why is this?

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unknown_variable_do_not_suggest_modules_with_public_function_with_incorrect_arity_2.snap
realraphrous: With these changes, we suggest a value like this:

- If the value is used as is, without being a function call or a type constructor, then we suggest values (variables and functions) that have the same name.
- If the value is used as a function call or type constructor, then we suggest functions or type constructor that have the same name and arity. (see https://github.com/gleam-lang/gleam/issues/2697#issuecomment-1997241885)

In addition to this, for pipelines, we only consider that the suggested function has arity `arguments.len() + 1` (see https://github.com/gleam-lang/gleam/pull/4810#discussion_r2259982804)

So with these changes, we do not offer suggestions based on the return type. In this case, `add` function has arity 1 with a function of arity 1 as its return type. To be suggested, the function `add` should be of arity 2.

Is that a correct behavior for you ? Do we need to change that and suggest even if number of arguments mismatch ?

realraphrous:
> Hello! Some of the snapshot tests show that nothing is suggested, but I couldn't work out why. Why is this?

@lpil, do you think it would be a great thing to add a little comment for each test explaining why it should pass or fail?

realraphrous:
I added comment for each test. If a test fail, a maintainer / contributor can just check the test + comment to know the compiler expected behavior !

lpil:
The comments look good, but because they're not in the snapshots the person would have to have 2 browser windows or 2 terminal windows open and for each failure manually search for the test to read the context. This is quite time consuming and easy to get lost with, so we want to have all the information in one place in the snapshot. That should make it much easier and quicker to work with.

Could you move the comments into the source code being snapshotted please? 🙏

realraphrous:
Ofc !

compiler-core/src/type_/environment.rs
```diff
@@ -864,10 +905,10 @@ impl Environment<'_> {
                     {
                         None
                     }
-                    Imported::Type(name) if module_info.get_public_type(name).is_some() => {
+                    Imported::Type(name) if module_info.get_importable_type(name).is_some() => {
                         Some(ModuleSuggestion::Importable(importable.clone()))
                     }
-                    Imported::Value(name) if module_info.get_public_value(name).is_some() => {
+                    Imported::Value(name) if module_info.get_importable_value(name).is_some() => {
```
lpil: 🔴 This is now suggesting _importable_ modules instead of public modules!

We only suggest public modules, internal modules should not be suggested as they are not part of the public API.

lpil (commented):
Left one little comment inline! Thank you.

compiler-core/src/type_/environment.rs
realraphrous: In the current codebase, that is, without these changes, this is already the case. These modified lines only reflect the changes I made to `get_public_*` functions. All `get_public_*` functions returned importable values (public and internal ones). These functions have been renamed to `get_importable_*` to better reflect what they were doing. Now, all functions `get_public_*` only return public values.

If you want, I can make the change on this, for sure !

realraphrous:
Hej @lpil !
This is done, is it fine for you ?

realraphrous:
@lpil, I rebased the branch on main

compiler-core/src/type_/expression.rs
lpil: Could you make issues for the existing bugs please 🙏  Thank you.

> With this PR, i added a new special case in do_infer_call for UntypedExpr::Var and i broke the first example which was emitting correct warnings. It seems to me that warnings in use expressions are very fragile. With these changes, we can be sure not to miss any warnings.

Is this broken currently? The tests are passing, so I'm not sure if you meant you've fixed it or if the test coverage is insufficient.

realraphrous: The test coverage was not sufficient unfortunately :c
I'm going to open an issue.

However, without that code, my new code causes a test to fail. What should I do in that case?

lpil (approved):
Thank you!!! Sorry this took me so long 🙏

realraphrous:
@lpil no problem at all! And thank you so much for taking the time to look into it and give me all that advice!

PR #4818, opened 2025-07-31
Add specific parse errors for angle-bracket generics

### LINKED ISSUE #4816: Specific parser error for angle-bracket generics
While getting started with Gleam, I tried to apply the syntax I know from other languages and write a generic function signature like this:

```gleam
fn something<T>(x: T)
```

I spent a while fiddling with variants of this syntax...

```gleam
fn something(T)(x: T)
fn<T> something(x: T)
fn something(x: T)
```

...but eventually went to look at the docs and found the correct answer, which I don't think I would have guessed:

```gleam
fn something(x: t)
```

----

Gleam has some other special parser errors to help catch patterns found in other languages, like Pythonic imports (`import gleam.io` instead of `import gleam/io`) and usage of JavaScript's `===`.

In a similar vein, do you think it is worthwhile providing a parser error like:

```text
error: Syntax error
  ┌─ test.gleam:3:13
  │
3 │ fn something<T>(x: T) {
  │             ^ I was expecting `(` here

If you were trying to define a generic function, Gleam does not use angle brackets.
Instead, you should write types in the parameter list with lower-case names.
Each unique lower-case name becomes a type variable:

    fn example(any_type: t) -> t

See: https://tour.gleam.run/functions/generic-functions/
```

----

If so, **I'd be keen to have a go contributing this**!
Just wanted to check through an issue, before I put the work to create a PR :) Thanks!

AaronC81:
This PR adds specific parse errors for these cases, where a developer new to Gleam (i.e. me!) tries to write generic functions or types using angle brackets: 

```gleam
fn something<T>(x: T) { ... }

let x: List<String> = []

type Maybe<String> { ... }
```

See the snapshots for examples of the concrete errors.

First PR, so any stylistic comments are welcome :)  
Thanks for the great contributing documentation, it was very easy to get started.

Closes #4816.


AaronC81:
Should I add a CHANGELOG entry for this change?  
I can see some PRs add one, but not all, and I couldn't spot any docs on which kinds of changes need an entry. 

If there should be one, let me know and I'm happy to update the PR

GearsDatapacks:
Yes, you'll need to add a changelog entry. Since we're in the release candidate phase for 1.12, this won't be merged until after 1.12 is out.
You'll need to add a new "Unreleased" section at the top of the file (h2) and then add a "Compiler" section below that (h3). Then you can look at existing changelog entries for an idea of what to write

AaronC81:
@GearsDatapacks - Great, thanks - added

compiler-core/src/parse/snapshots/gleam_core__parse__tests__type_angle_generics_usage_error.snap
```diff
@@ -0,0 +1,20 @@
...
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:15
+  │
+1 │ let list: List<Int> = []
+  │               ^ I was expecting `(` here.
+
+Instead of angle brackets, Gleam uses parentheses for generic types:
+
+    List(String)
+    Result(Int, Error)
```
lpil, line 18: Could we instead show the exact type they were going to write, but with the correct syntax? So here it would be `List(Int)`

compiler-core/src/parse/snapshots/gleam_core__parse__tests__function_definition_angle_generics_error.snap
```diff
@@ -0,0 +1,21 @@
...
+  ┌─ /src/parse/error.gleam:1:6
+  │
+1 │ fn id<T>(x: T) { x }
+  │      ^ I was expecting `(` here.
+
+If you were trying to define a generic function, Gleam does not use angle brackets.
+Instead, you should write types in the parameter list with lower-case names.
+Each unique lower-case name becomes a type variable:
+
+    fn example(arg: t) -> t
```
lpil: No abbreviations please 🙏

compiler-core/src/parse/error.rs
```diff
@@ -627,6 +629,37 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
...
+            ParseErrorType::TypeAngleGenerics => ParseErrorDetails {
+                text: [
+                    "Instead of angle brackets, Gleam uses parentheses for generic types:",
+                    "",
+                    "    List(String)",
+                    "    Result(Int, Error)",
+                    "",
+                    "See: https://tour.gleam.run/data-types/generic-custom-types/",
+                ]
+                .join("\n"),
```
lpil: Favour multi-line strings over joining arrays 🙏

compiler-core/src/parse/snapshots/gleam_core__parse__tests__type_angle_generics_definition_error.snap
```diff
@@ -0,0 +1,25 @@
...
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:11
+  │
+2 │ type Maybe<A> {
+  │           ^ I was expecting `(` here.
+
+Instead of angle brackets, Gleam uses parentheses for generic types:
+
+    List(String)
+    Result(Int, Error)
```
lpil: Same here, it would be great to show their actual type, so here we would show this:

```gleam
type Maybe(a) {
```

compiler-core/src/parse/snapshots/gleam_core__parse__tests__type_angle_generics_definition_error.snap
```diff
@@ -0,0 +1,25 @@
...
+
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:11
+  │
+2 │ type Maybe<A> {
+  │           ^ I was expecting `(` here.
+
+Instead of angle brackets, Gleam uses parentheses for generic types:
```
lpil:
```suggestion
Type parameters use lowercase names and surrounded by parentheses.
```

compiler-core/src/parse/snapshots/gleam_core__parse__tests__type_angle_generics_usage_error.snap
```diff
@@ -0,0 +1,20 @@
...
+let list: List<Int> = []
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:15
+  │
+1 │ let list: List<Int> = []
+  │               ^ I was expecting `(` here.
+
+Instead of angle brackets, Gleam uses parentheses for generic types:
```
lpil:
```suggestion
Type parameters use lowercase names and surrounded by parentheses.
```

compiler-core/src/parse/snapshots/gleam_core__parse__tests__function_definition_angle_generics_error.snap
```diff
@@ -0,0 +1,21 @@
...
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:6
+  │
+1 │ fn id<T>(x: T) { x }
+  │      ^ I was expecting `(` here.
+
+If you were trying to define a generic function, Gleam does not use angle brackets.
+Instead, you should write types in the parameter list with lower-case names.
+Each unique lower-case name becomes a type variable:
```
lpil:
```suggestion
Generic function type variables do not need to be predeclared like they
would be in some other languages, instead they are written with lowercase
names.
```

lpil (commented):
Great, thank you! I've added some small notes on the format of the error messages, mostly around showing what the programmer needs to write, and using wording that talks less from an outside point of view.

compiler-core/src/parse/snapshots/gleam_core__parse__tests__type_angle_generics_usage_error.snap
AaronC81: I've gone for

> Type parameters use lowercase names and _**are**_ surrounded by parentheses.

Hope that's alright, happy to tweak!

AaronC81:
Thank you very much for the review!

Addressed the wording changes.

The idea of giving customised messages is really good, and would make this much more useful, but is a little trickier.  
I'm keen to give that a go when I have some time soon. I'll reach out if I get stuck :)

AaronC81:
This now shows a customised suggestion for the type definition/usage cases!

This comes with some extra test cases - for example, to give a better suggestion, upnames are converted to lowercase names:

```gleam
// User entered:
type Either<A, B> {

// Error message says:
type Either(a, b) {
```

----

I had a look at extending this to function definitions too, but this looks a fair bit trickier.

You would need to deep-traverse the AST to spot all of the type arguments, and ideally cope with further syntax errors due to `<>` along the way, e.g.:

```gleam
// Source code:
fn is_in_set<A>(x: A, y: set.Set<A>)
// The suggestion you'd ideally want to give:
fn is_in_set(x: a, y: set.Set(a))

//             Easy case      Further syntax errors on these <>s
//                 v            v v
fn is_in_set<A>(x: A, y: set.Set<A>)
//                               ^
//                Harder case (nested type argument)
```

This particular case wasn't a comment in the original review, so I've left the function definition errors as they are. 

I think the overall updated error in this PR is still enough of an improvement that it's hopefully useful without this :)

PR #4845, opened 2025-08-12
Emit a warning when a `src` module imports a dev dependency

### LINKED ISSUE #4663: Add a warning when using a dev dependency in `/src`
Making that an error would be a breaking change, but at least we can make that a warning and remind folks that if they want to use a dependency in `/src` they should make that a regular dependency and not a dev one

GearsDatapacks:
Closes #4663 
Do we have existing code for testing preventing publishing? I couldn't find any

### DIFF
--- a/compiler-cli/src/export.rs
+++ b/compiler-cli/src/export.rs
@@ -3,7 +3,6 @@ use gleam_core::{
     Result,
     analyse::TargetSupport,
     build::{Codegen, Compile, Mode, Options, Target},
-    error::Error,
     paths::ProjectPaths,
 };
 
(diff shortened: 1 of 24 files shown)


compiler-cli/src/export.rs
```diff
@@ -52,6 +53,20 @@ pub(crate) fn erlang_shipment(paths: &ProjectPaths) -> Result<()> {
...
 
+    let mut modules_importing_dev_dependencies = Vec::new();
+
+    for module in built.root_package.modules.iter() {
+        if module.ast.type_info.imports_dev_dependency() {
+            modules_importing_dev_dependencies.push(module.name.clone());
+        }
+    }
+
+    if !modules_importing_dev_dependencies.is_empty() {
```
lpil: How come we have to do this but we don't have to do it for `test`?

lpil:
I wonder if this should be an error instead of a warning? It seems like a bug that they could be imported before.

compiler-cli/src/export.rs
GearsDatapacks: I'm not quite sure what you mean by do it for test

giacomocavalieri:
Yeah I'm all for saying this is a bug and making it an error, though it would be a breaking change and we've included it in the v2 document.

GearsDatapacks:
I don't mind either way, I was just following what was agreed in the linked issue

compiler-cli/src/export.rs
lpil: Why don't we have a similar bit of logic here that checks for `src/` modules that import `test/` modules?

GearsDatapacks: This is for importing a dev dependency, not a module in `dev/` or `test/` in the same project. Importing from `dev/` or `test/` already throws an error.

lpil: Oh of course. Sorry about that.

GearsDatapacks:
Ok, I have now changed the warning to an error in accordance to your suggestions.

PR #4848, opened 2025-08-13
Improve error message for private opaque types

giacomocavalieri:
I noticed that writing
```gleam
opaque type Wibble {
  Wobble
}
```
produced a confusing syntax error message that would stop analysis from happening in the entire module. I've improved the error message to provide an explanation and a hint, also the error now happens during the analysis step, so it won't stop other errors from being reported!

I've also included a LS quick fix to automatically remove it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,27 @@
 
 ### Language server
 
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you!

PR #4849, opened 2025-08-13
Omit label shorthand ':' in prepare_rename

### LINKED ISSUE #4843: Rename on label shorthand includes colon
Version: 1.12.0

When using the language server to rename local variable as a label shorthand
- from its usage `function_call(arg:)`
- or its definition `let Type(field:) = todo`
as a label shorthand, the `prepare_rename` response includes the colon (as well as any whitespace between the name and the colon), so your rename popup might look something like this 

Image

<disclaimer: I'm a first time contributor and mostly unfamiliar with github etiquette; don't be shy to point out if I missed anything basic>

# Root Cause

In my opinion, the cause traces all the way down to `parse.rs`, where in `parse_constructor_pattern_arg`, `parse_record_update_arg`, and `parse_fn_argument` (all marked with `// Argument supplied with a label shorthand`).
All 3 functions return some arg type with a contained variable. The `location` of the contained variable is the same as that of the arg. So unlike e.g. simple `let var = todo` assignments, where the `location` of `var` only spans `var`, in `function_call(arg:)` both the `CallArg` and the contained `UntypedExpr::Var` point to `arg:`.

I could be wrong here, but it seems to me that the main reason this is implemented this way is to make language server renames easier—the new name can simply be inserted after the location of the variable because it includes the colon.

https://github.com/gleam-lang/gleam/blob/fe9cb5dcf6daa3653f53638bd23071aaae46f34f/compiler-core/src/language_server/rename.rs#L63-L65

# Solutions

## Super Simple

As we have the label in `engine#prepare_rename`, we could simply trim the location to the label's length here:
https://github.com/gleam-lang/gleam/blob/fe9cb5dcf6daa3653f53638bd23071aaae46f34f/compiler-core/src/language_server/engine.rs#L644-L649

## Pulling at the Root

The ideal to me seems to decouple the variable's `location` from the argument's. As far as I can tell there are two places where that depends on this behavior

- `uses_label_shorthand` on `UntypedRecordUpdateArg` and `CallArg`
- `rename.rs#rename_local_variable`, where a rename simply appends at the end of the reference/definition: `edits.insert(reference.location.end, format!(" {}", params.new_name))`

`uses_label_shorthand` could just compare the _start_ of the location instead. However, the language server currently flattens the above types into `Referenced::LocalVariable` and would this require a larger effort to carry the distinction down into the server.

# Conclusion

As the simple solution is easy to do (and undo!) and fixes the issue at hand I would just go with that. Plus, I have a PR ready!

frunobulax-the-poodle:
As the location of a variable includes the colon, it is also included in the `prepare_rename` response, which is annoying

This PR trims the location to the length of the label. This is done for both declaration in destructuring and usage as argument (for which I've added a new test).

As the trimming only happens in prepare_rename, no other language server operations are affected!

Fixes #4843

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -105,6 +105,10 @@
   relevant variable.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Renaming a variable from a label shorthand (`name:`) no longer includes the
+  colon in the rename dialog (`name:` -> `name`)
+  ([fruno](https://github.com/frunobulax-the-poodle))
+
 ### Formatter
 
 - The formatter now removes needless multiple negations that are safe to remove.
--- a/compiler-core/src/language_server/engine.rs
+++ b/compiler-core/src/language_server/engine.rs
@@ -645,11 +645,16 @@ where
                 }) if location.contains(byte_index) => match origin.map(|origin| origin.syntax) {
                     Some(VariableSyntax::Generated) => None,
                     Some(
-                        VariableSyntax::Variable { .. }
-                        | VariableSyntax::AssignmentPattern
-                        | VariableSyntax::LabelShorthand(_),
-                    )
-                    | None => success_response(location),
+                        VariableSyntax::Variable(label) | VariableSyntax::LabelShorthand(label),
+                    ) => success_response(SrcSpan {
+                        start: location.start,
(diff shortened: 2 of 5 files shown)


lpil (approved):
Thank you!

PR #4851, opened 2025-08-14
Stop suggesting project names starting with the `gleam_` prefix

### LINKED ISSUE #4844: Invalid default project name suggested by build tool
When providing a new gleam project name that the build tool deems unacceptable, it suggests an alternate name of `gleam_app`. 

Modules prefixed with gleam_ are restricted to core libraries so we should likely be suggesting a different name by default, or better yet asking the user for input to correct the name if possible. 

Image

giacomocavalieri:
If someone picked `gleam new gleam`, instead of suggesting `gleam_app` it will now suggest `app_gleam` so it doesn't have the `gleam_` prefix!

Fixes #4844


lpil (approved):
Thank you!

PR #4860, opened 2025-08-17
Stop removing blocks from clause guards

### LINKED ISSUE #4790: Do not remove blocks from guards
I was writing a piece of code with loads of guard conditions, something like this:
```gleam
case wibble {
  <<a, b>>
  if { a != 1 && a != 2 && a != 3 }
  && { b != 1 && b != 2 && b != 3 } -> todo
  _ -> todo
}
```
Using blocks to help make the code more readable. However the formatter decided to strip all the blocks away and turn the condition into this:
```gleam
case wibble {
  <<a, b>>
  if a != 1
  && a != 2
  && a != 3
  && b != 1
  && b != 2
  && b != 3 -> todo
  _ -> todo
}
```
I think we should always keep blocks, same way we do that in regular expressions

giacomocavalieri:
This PR fixes #4790


lpil (approved):
Thanks!!

PR #4863, opened 2025-08-18
Fix renaming module selects in guards

### LINKED ISSUE #4862: Rename replaces invalid range for module access in guards
When a piece of code references a constant in a case clause guard, such as in the following code:
```gleam
// wibble.gleam
pub const something = 10

// main.gleam
import wibble

pub fn main() {
  case 10 {
    x if x == wibble.something -> todo
    _ -> todo
  }
}
```

If that constant is then renamed, the module in the qualified path is renamed, rather than the constant itself:
```gleam
// wibble.gleam
pub const new_name = 10

// main.gleam
import wibble

pub fn main() {
  case 10 {
    x if x == new_name.something -> todo
    _ -> todo
  }
}
```

GearsDatapacks:
Fixes #4862

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,3 +159,7 @@
   a dev dependency, while previously it would incorrectly let these
   dependencies to be imported.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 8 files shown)


lpil (approved):
Thank you!

PR #4865, opened 2025-08-20
Fix renaming variables in record updates

### LINKED ISSUE #4859: Invalid code after LSP rename action
I've been doing some refactoring when I noticed that renaming a function argument using a rename LSP action produces invalid code.

## Steps to reproduce

Initial code:

```gleam
type Groups {
  Groups(one: Int, two: Int, three: Int, four: Int, five: Int)
}

fn rebalance(grouping: Groups) -> Groups {
  case grouping {
    Groups(_, _, three, four, five) if five > three ->
      Groups(..grouping, three: 0, four: four + three * 2, five: five - three)
    Groups(_, _, three, four, five) ->
      Groups(..grouping, three: three - five, four: four + five * 2, five: 0)
  }
}
```

If I put my cursor on the `grouping` argument of the function `rebalance` and execute a `rename` LSP action via `vim.lsp.buf.rename` function with a new name set to `groups` I get the following:

```gleam
// type definition omitted
fn rebalance(groups: Groups) -> Groups {
  case groups {
    Groups(_, _, three, four, five) if five > three ->
      Groups(..groupsthree: 0, four: four + three * 2, five: five - three)
    Groups(_, _, three, four, five) ->
      Groups(..groupsthree: three - five, four: four + five * 2, five: 0)
  }
}
```

Notice how there's not comma between a spread with an updated name in a record update and the `three` field.

Even weirder thing happens if I use the initial code and rename `grouping` argument in the `rebalance` function to a longer new name, e.g. `not_grouping`:

```gleam
// type definition omitted
fn rebalance(not_grouping: Groups) -> Groups {
  case not_grouping {
    Groups(_, _, three, four, five) if five > three ->
      Groups(..not_groupingping, three: 0, four: four + three * 2, five: five - three)
    Groups(_, _, three, four, five) ->
      Groups(..not_groupingping, three: three - five, four: four + five * 2, five: 0)
  }
}
```

Notice that this time comma is there but the name of the spreaded variable in the record update is wrong.

## Versions

Gleam: 1.12.0
OS: macOS 15.5
Editor: Neovim v0.11.3 with neovim/nvim-lspconfig v2.3.0

GearsDatapacks:
Fixes #4859 
As I mentioned in the issue, this doesn't fix the underlying problem but it avoids this most common situation.

### DIFF
--- a/compiler-core/src/language_server/reference.rs
+++ b/compiler-core/src/language_server/reference.rs
@@ -400,10 +400,16 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
                 location: definition_location,
                 ..
             } if definition_location == self.definition_location => {
-                self.references.push(VariableReference {
-                    location: *location,
-                    kind: VariableReferenceKind::Variable,
-                })
+                if !self
+                    .references
+                    .iter()
+                    .any(|reference| reference.location == *location)
+                {
+                    self.references.push(VariableReference {
+                        location: *location,
+                        kind: VariableReferenceKind::Variable,
+                    })
+                }
             }
             _ => {}
         }
--- a/compiler-core/src/language_server/reference.rs
+++ b/compiler-core/src/language_server/reference.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use ecow::EcoString;
 use lsp_types::Location;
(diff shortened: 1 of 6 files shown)


compiler-core/src/language_server/reference.rs
```diff
@@ -400,10 +400,16 @@ impl<'ast> Visit<'ast> for FindVariableReferences {
...
+                if !self
+                    .references
+                    .iter()
+                    .any(|reference| reference.location == *location)
+                {
+                    self.references.push(VariableReference {
+                        location: *location,
+                        kind: VariableReferenceKind::Variable,
+                    })
+                }
```
lpil: It's not very clear why this is done, could we document this?

Why is it that this is done here but not in the other cases where we push a VariableReference?

If the definition locations are to be unique should this be a set instead?

GearsDatapacks: Yeah I should document this. I only put it here because this is the only place it is needed to fix this issue, but it's probably worth putting the check everywhere to be safe.
I didn't want to use a set because that loses ordering and I thought it best to keep the edits in lexical order, but maybe that doesn't matter.

Issue #3786, opened 2024-11-04
Error's type printing: Consider only annotated type variables for the current definition rather than the whole module


yoshi-monster:
I talked about this on Discord with Gears, and he thinks this is actually a bug worth reporting 😄 

When the compiler introduces auto-generated type variable names for reporting errors (and in the LSP), 1.5.1 always starts at `a`. The nightly release of the compiler instead seems to reserve explicitly annotated type variable names used elsewhere in the module. This means that the names generated will not usually start with `a`, but rather with the next free letter.

```gleam
import gleam/io

type Wibble(a, b, c, d, e, f, g) {
  Wibble(value: a)
}

fn wibble(value: h) {
  io.debug(value)
  Nil
}

fn wobble() {
  let x: List(i) = []
  wibble(x)
  wibble(fn() -> List(j) { [] })
}

pub fn main() {
  let x: Nil = []
}
```

produces:

```
error: Type mismatch
   ┌─ C:\Users\joshua\Documents\globe\src\globe.gleam:19:16
   │
18 │   let x: Nil = []
   │                ^^

Expected type:

    Nil

Found type:

    List(k)
```

type variables a..j are reserved because of the other usages above.

I ran `git bisect` and figured out this was introduced in https://github.com/gleam-lang/gleam/commit/8c5834bcdc0115abeea16c7e1ee9826d42462564

~ 💜

GearsDatapacks:
The correct behaviour here would be to only track type variable names in the current function, rather than the whole module. This would require a rework of the `Names` class. 
This is pretty low priority as it still produces correct annotations, even if somewhat redundant ones.

lpil:
Thank you

ankddev:
I'm trying to make it working properly for errors as well!

ankddev:
Okay, I don't know how to implement it since on code actions we have access to AST nodes while in errors we haven't. Because of this we can't just register new type variables by traversing AST. In #5046 I've implemented solution, but it doesn't consider type variables at all. One possible solution is to save list of type variables in error itself on the caller side (is it possible?; I've only brief look at code regarding this). Any ideas?

CC (maybe you have some ideas?): @lpil @GearsDatapacks @giacomocavalieri

PR #4870, opened 2025-08-23
Only consider local type variables for type printing in code actions

GearsDatapacks:
This PR partially addresses #3786 
This PR changes how various code actions use the `type_::Printer`, such that printed type variables are now local to functions.
This PR doesn't change error messages or hover type info. For those, it would require more plumbing to be able to access the AST in order to determine which type variables are in scope.

I figured code actions are the highest priority feature of this, since adding type annotations and ending up with type variables like `e` and `f` is a common annoyance I face, which I then have to edit the code for, however I have never really noticed that problem in error messages or hover.

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -9307,6 +9307,7 @@ fn collapse_nested_case_does_not_remove_labels() {
   }
 }
 
+pub type Wibble {
   Wibble(field: Int, field2: String)
   Wobble
 }
@@ -9327,6 +9328,7 @@ fn collapse_nested_case_does_not_remove_labels_with_shorthand_syntax() {
         2 -> 4
         _ -> -1
       }
+
     Wobble -> todo
   }
 }


compiler-core/src/type_/printer.rs
```diff
@@ -305,7 +305,7 @@ pub enum PrintMode {
 /// names that types and modules have been aliased with in the current module.
 #[derive(Debug)]
 pub struct Printer<'a> {
-    names: &'a Names,
+    pub names: &'a Names,
```
lpil: How come this is public now? I couldn't see the use

compiler-core/src/type_/printer.rs
```diff
@@ -333,6 +333,41 @@ impl<'a> Printer<'a> {
...
+            printed_type_variables: Default::default(),
+            printed_type_variable_names: Default::default(),
+        }
+    }
+
+    pub fn clear_type_variables(&mut self) {
+        self.printed_type_variable_names.clear();
+    }
+
+    pub fn register_type_variable(&mut self, name: EcoString) {
```
lpil, line 375: Could you document what these functions are for please 🙏

compiler-core/src/type_/printer.rs
GearsDatapacks: Ah that was probably just a mistake from while I was testing stuff

PR #4872, opened 2025-08-23
fix: message in `echo .. as ..` is omitted in browser

### LINKED ISSUE #4871: `echo ... as ...` message incorrectly omitted when running in the browser
I noticed today that the `echo ... as ...` message (introduced in version 1.12) is not printed when running in the browser. 

---

Consider the following code:

```gleam
pub fn main() {
  echo 11 as "lucky number"
}
```

Compare the output of this code [as shown in the release post](https://gleam.run/news/no-more-dependency-management-headaches/#Echo-messages) to that [when run in the playground today](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIADgK4BGABAGZI0C2AhnEgBQCUNwAOoxoQcACwQ106Gixg0+IADYUcAawCeNJBSZVo8gQF8BIA0A=).

---

This is a super easy fix, we just need to add `${string_message}` before the newline in the part of `echo.mjs` that targets the browser, specifically [line 20](https://github.com/gleam-lang/gleam/blob/5288b747918a006d584d2d110bdc1bd2ca0378af/compiler-core/templates/echo.mjs#L20).

This is exactly what was done in the branches of the `echo` function targeting Node and Deno when the feature was first added. It looks to me very much like the browser was just accidentally skipped.

ankddev:
Closes #4871.
# Description
Fix of small bug where message in `echo .. as ..` will be omitted in browser target.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -269,3 +269,6 @@
 - Fixed a bug where renaming a constant which is referenced in another module
   inside a guard would generate invalid code.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where `echo .. as ..` message will be omitted in browser target.
+  ([Andrey Kozhev](https://github.com/ankddev))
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you!!!

PR #4876, opened 2025-08-24
Show record field documentation on hover

### LINKED ISSUE #4869: LSP does not show hints for doc comments on fields records
Doc comments are supported on record fields. The following example:

```gleam
pub type MetaInfo {
  MetaInfo(
    /// The announce url of the tracker.
    /// According to the specification this is always set.
    /// In the real world most torrents ditch it in favor of announce list or trackless peers
    ///
    /// The url supports http tracking via get requests and udp tracking. 
    /// It is worth noting that many trackers will accept either protocols regardless of the one specified    
    announce: Option(BitArray),
    info: Info,
  )
}
```

Correctly gets rendered by the docs:

Image

However, the LSP doesn't seem to detect them and when hovering on the field in VSCode it only shows the type:

Image

The expected behavior would be for the doc comments to show like any other like in functions and types.

I'm running this on Gleam 1.12.0, Arch Linux, Code version 1.103.1 and the Gleam extension v2.12.1.

GearsDatapacks:
Closes #4869

### DIFF
--- a/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__hover__no_documentation_for_shared_record_field.snap
+++ b/compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__hover__no_documentation_for_shared_record_field.snap
@@ -1,0 +1,27 @@
+---
+source: compiler-core/src/language_server/tests/hover.rs
+expression: "\npub type Wibble {\n  Wibble(\n    /// This is some documentation about the wibble field.\n    wibble: Int\n  )\n  Wobble(\n    /// This won't show up because we don't know which wibble field it is\n    wibble: Int\n  )\n}\n\npub fn wibble(w: Wibble) {\n  w.wibble\n}\n"
+---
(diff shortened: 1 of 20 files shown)


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__hover__no_documentation_for_shared_record_field.snap
```diff
@@ -0,0 +1,27 @@
...
+pub fn wibble(w: Wibble) {
+  w.wibble
+  ▔▔▔▔▔▔↑▔
+}
+
+
+----- Hover content -----
+Scalar(
+    String(
+        "```gleam\nInt\n```\n",
```
lpil: Can we show the documentation for both in this instance? Seeing as extra docs is better than no docs

lpil (commented):
Thank you! 1 Q inline

lpil (approved):
Wonderful, thank you!!

PR #4880, opened 2025-08-26
Remove unnecessary assignments for blocks when the value is unused

GearsDatapacks:
I noticed that currently, for this code:
```gleam
pub fn main() {
  {
    let x = 10
    echo x
  }

  Nil
}
```

We generate the following JavaScript:
```javascript
export function main() {
  let _block;
  {
    let x = 10;
    _block = echo(x);
  }
  _block;

  return undefined;
} 
```

However, since we are not using the value returned by the first block, it doesn't really make sense to generate the `let _block` part; it just adds unnecessary code.

This PR removes it so it now just generates:
```javascript
export function main() {
  {
    let x = 10;
    echo(x);
  }

  return undefined;
} 
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,10 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The code generated for blocks on the JavaScript target has been improved and
+  is now smaller in certain cases.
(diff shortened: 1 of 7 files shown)


lpil (approved):
Thank you!

PR #4882, opened 2025-08-26
update the `toml` dependency

### LINKED ISSUE #4881: Update toml Rust dependency
The `toml` library has a new version, but our configuration deserialisation code fails if it is upgraded to.

Upgrade to the latest version and change the configuration deserialisation code to work with the new version.

giacomocavalieri:
This PR closes #4881 

The new version of `toml` also improves on the look of errors, I think they're a bit nicer now! Since the tests were failing and needed updating I turned those into snapshots to make it easier to see and test the new error messages


lpil (approved):
Wow what an annoying change to have to make. Thank you Jak!!

PR #4884, opened 2025-08-26
Properly infer variants when `echo` is present

### LINKED ISSUE #4883: `echo` interferes with variant narrowing in `case` expressions.
I have this call to `list.filter`:

```gleam
list.filter(all, fn(item) {
  case item {
    Orphan(..) -> False
    Item(..) ->
      string.starts_with(
        string.lowercase(item.label),
        string.lowercase(filter),
      )
      && !list.contains(selected, item)
  }
})
```

Inside you can see I am making use of variant narrowing to access `item.label` in the `Item(..)` case.

Unfortunately, if i add an echo such that the case is now:

```diff
  list.filter(all, fn(item) {
-   case item {
+   case echo item {
      Orphan(..) -> False
      Item(..) ->
        string.starts_with(
          string.lowercase(item.label),
          string.lowercase(filter),
        )
        && !list.contains(selected, item)
    }
  })
```

I receive a type error:

```
error: Unknown record field
    ┌─ /...
    │
503 │           string.lowercase(item.label),
    │                                 ^^^^^ Did you mean `value`?

    Item

It has these accessible fields:

    .value

```

--- 

`echo` is very useful for debugging and it's very unfortuante that it affects type checking when used, it would be great to see this fixed.

GearsDatapacks:
Fixes #4883

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -306,3 +306,7 @@
 - Fixed a bug where renaming a variable used in a record update would produce
   invalid code in certain situations.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where adding `echo` to the subject of a `case` expression would
+  prevent variant inference from working correctly.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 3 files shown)


lpil (approved):
Nice, thank you!

PR #4885, opened 2025-08-27
Allow the "pattern match on variable" code action in case expression

### LINKED ISSUE #4798: Code action to further pattern match on a pattern
Hello! There's on thing I wished many times I could do, that is having a code action that would allow me to further pattern match on some variable/discard in a case expression:
```gleam
case something {
  Ok(_) -> todo
//   ^ [pattern match on value]
  Error(_) -> todo
}
```
Would result in:
```gleam
case something {
  // Here we could reuse the code that picks a name based on the type
  // that we already use for many code actions...
  Ok(wibble) -> 
    case wibble {
      Wibble -> todo
      Wobble -> todo
      Woo -> todo
    }

  Error(_) -> todo
}
```
And if the type has a single variant it could go as far as using it directly in place of the `_`/variable:
```diff
case something {
- Ok(_) -> todo 
+ Ok(User(name:, id:)) -> todo
  Error(_) -> todo
}
```

giacomocavalieri:
This PR fixes #4798


CHANGELOG.md
```diff
@@ -201,6 +201,36 @@
...
+  ```
+
+  Triggering the action over the `user` variable would result in the following
+  code:
+
+  ```gleam
+  pub fn main() {
+    case find_user() {
+      Ok(user) ->
+        case user {
```
lpil: I thought we said we didn't want to produce a nested case here? Seeing as it's rarely what is wanted, and you can already use the "pattern match on value" action to create a nested case today by typing the variable name in the clause body to trigger it.

giacomocavalieri: You're right, I forgot about it 😆
It should be all fixed now, and the code ends up being quite a bit nicer!

giacomocavalieri:
This is ready for review!

lpil (approved):
I love this!!!!

Issue #4836, opened 2025-08-06
Return an error if asked to update a package that does not exist


daniellionel01:
When attempting to update a major version and violating the version constraints in `gleam.toml` the CLI does not throw an error, as it does when attempting to do the same with `gleam add`

Example:

```sh
$ gleam update gleam_httpc@5
  Resolving versions

The following dependencies have new major versions available:

gleam_httpc 4.1.1 -> 5.0.0
```

gleam v1.12.0

lpil:
This feature doesn't exist! `gleam update` doesn't take name-requirement pairs, it only takes names. This code is trying to update a package called `gleam_httpc@5`, which does not exist.

We need to make `gleam update` error when asked to update packages that do not exist.

EtienneBoutet:
I will try to fix this problem :)

lpil:
Thank you!!

PR #4887, opened 2025-08-27
gleam update should return error if package does not exists 

EtienneBoutet:
Fixes https://github.com/gleam-lang/gleam/issues/4836

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -225,6 +225,9 @@ pub fn update(paths: &ProjectPaths, packages: Vec<String>) -> Result<()> {
         UseManifest::Yes
     };
 
+    let config = crate::config::read(paths.root_config())?;
+    assert_packages_exist_locally(&config, packages.clone())?;
+
     // Update specific packages
     _ = download(
         paths,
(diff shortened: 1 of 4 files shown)


compiler-cli/src/dependencies.rs
```diff
@@ -225,6 +225,9 @@ pub fn update(paths: &ProjectPaths, packages: Vec<String>) -> Result<()> {
         UseManifest::Yes
     };
 
+    let config = crate::config::read(paths.root_config())?;
+    assert_packages_exist_locally(&config, packages.clone())?;
```
lpil: This checks that the packages are direct dependencies, but they could be transitive dependencies! You'll need to do the checking after loading the manifest, checking that they exist in there.

EtienneBoutet: Hi @lpil thanks for the review. I changed the code to use the manifest.

giacomocavalieri:
@lpil all the comments have been addressed, do you think this could be merged now?

compiler-cli/src/dependencies.rs
```diff
@@ -225,6 +225,9 @@ pub fn update(paths: &ProjectPaths, packages: Vec<String>) -> Result<()> {
         UseManifest::Yes
     };
 
+    let manifest = read_manifest_from_disc(paths)?;
+    assert_packages_exist_locally(&manifest, packages.clone())?;
```
lpil, line 229: I think this now fails when there is no manifest. Could we move this inside the update function? Then it would only read the manifest a single time too!

vshakitskiy:
Hey @EtienneBoutet! Sorry for the ping, I definitely don't want to step on your toes here 🙏.

I just noticed this hasn't had activity for some time, so I wanted to ask, would you mind if I picked this up to get it merged? If you're still planning to work on it, just let me know!

ce-eboutet:
Hi @vshakitskiy please go ahead.

lpil:
Thank you both! Going to close this to continue in the other.

PR #4888, opened 2025-08-28
Add create unknown module code action

tarkah:
Adds a new code action to create an unknown module. This allows you to easily define new modules without having to leave your editor.

https://github.com/user-attachments/assets/d1e6a46f-e308-4fab-91f2-772e4835214c

### DIFF
--- a/language-server/src/tests.rs
+++ b/language-server/src/tests.rs
@@ -25,6 +25,7 @@ use lsp_types::{Position, TextDocumentIdentifier, TextDocumentPositionParams, Ur
 
 use gleam_core::{
     Result,
+    build::Origin,
     config::PackageConfig,
     io::{
         BeamCompiler, Command, CommandExecutor, FileSystemReader, FileSystemWriter, ReadDir,
@@ -456,7 +457,7 @@ impl<'a> TestProject<'a> {
     pub fn src_from_module_url(&self, url: &Url) -> Option<&str> {
         let module_name: EcoString = self.module_name_from_url(url)?.into();
 
-        if module_name == "app" {
+        if module_name == LSP_TEST_ROOT_PACKAGE_NAME {
             return Some(self.src);
         }
 
@@ -473,6 +474,31 @@ impl<'a> TestProject<'a> {
             .or_else(|| find_module(&self.hex_modules))
             .or_else(|| find_module(&self.dev_hex_modules))
             .or_else(|| find_module(&self.indirect_hex_modules))
+    }
+
+    pub fn src_from_origin_and_module_name(
+        &self,
+        origin: Origin,
+        module_name: &str,
+    ) -> Option<&str> {
+        let find_module = |modules: &Vec<(&'a str, &'a str)>| {
+            modules
+                .iter()
+                .find(|(name, _)| *name == module_name)
+                .map(|(_, src)| *src)
+        };
+
+        match origin {
+            Origin::Src => {
+                if module_name == LSP_TEST_ROOT_PACKAGE_NAME {
+                    return Some(self.src);
+                }
+
+                find_module(&self.root_package_modules)
+            }
+            Origin::Test => find_module(&self.test_modules),
+            Origin::Dev => find_module(&self.dev_modules),
+        }
     }
 
     pub fn add_module(mut self, name: &'a str, src: &'a str) -> Self {
@@ -615,6 +641,18 @@ impl<'a> TestProject<'a> {
         TextDocumentPositionParams::new(TextDocumentIdentifier::new(url), position)
     }
 
+    pub fn build_src_path(&self, position: Position, module: &str) -> TextDocumentPositionParams {
+        let path = Utf8PathBuf::from(if cfg!(target_family = "windows") {
+            format!(r"\\?\C:\src\{module}.gleam")
+        } else {
+            format!("/src/{module}.gleam")
+        });
+
+        let url = Url::from_file_path(path).unwrap();
+
+        TextDocumentPositionParams::new(TextDocumentIdentifier::new(url), position)
+    }
+
     pub fn build_test_path(
         &self,
         position: Position,
@@ -658,11 +696,32 @@ impl<'a> TestProject<'a> {
         let mut engine = self.build_engine(&mut io);
 
         // Add the final module we're going to be positioning the cursor in.
-        _ = io.src_module("app", self.src);
+        _ = io.src_module(LSP_TEST_ROOT_PACKAGE_NAME, self.src);
 
         let _response = engine.compile_please();
 
         let param = self.build_path(position);
+
+        (engine, param)
+    }
+
+    pub fn positioned_with_io_in_src(
+        &self,
+        position: Position,
+        module: &str,
+    ) -> (
+        LanguageServerEngine<LanguageServerTestIO, LanguageServerTestIO>,
+        TextDocumentPositionParams,
+    ) {
+        let mut io = LanguageServerTestIO::new();
+        let mut engine = self.build_engine(&mut io);
+
+        // Add the final module we're going to be positioning the cursor in.
+        _ = io.src_module(LSP_TEST_ROOT_PACKAGE_NAME, self.src);
+
+        let _response = engine.compile_please();
+
+        let param = self.build_src_path(position, module);
 
         (engine, param)
     }
@@ -679,10 +738,9 @@ impl<'a> TestProject<'a> {
         let mut engine = self.build_engine(&mut io);
 
         // Add the final module we're going to be positioning the cursor in.
-        _ = io.src_module("app", self.src);
+        _ = io.src_module(LSP_TEST_ROOT_PACKAGE_NAME, self.src);
 
-        let response = engine.compile_please();
-        assert!(response.result.is_ok());
+        let _response = engine.compile_please();
 
         let param = self.build_test_path(position, test_name);
 
@@ -701,10 +759,9 @@ impl<'a> TestProject<'a> {
         let mut engine = self.build_engine(&mut io);
 
         // Add the final module we're going to be positioning the cursor in.
-        _ = io.src_module("app", self.src);
+        _ = io.src_module(LSP_TEST_ROOT_PACKAGE_NAME, self.src);
 
-        let response = engine.compile_please();
-        assert!(response.result.is_ok());
+        let _response = engine.compile_please();
 
         let param = self.build_dev_path(position, test_name);
 
@@ -723,6 +780,30 @@ impl<'a> TestProject<'a> {
         let (mut engine, params) = self.positioned_with_io(position);
 
         executor(&mut engine, params, self.src.into())
+    }
+
+    pub fn in_module_at<T>(
+        &self,
+        origin: Origin,
+        module: &str,
+        position: Position,
+        executor: impl FnOnce(
+            &mut LanguageServerEngine<LanguageServerTestIO, LanguageServerTestIO>,
+            TextDocumentPositionParams,
+            EcoString,
+        ) -> T,
+    ) -> T {
+        let (mut engine, params) = match origin {
+            Origin::Src => self.positioned_with_io_in_src(position, module),
+            Origin::Test => self.positioned_with_io_in_test(position, module),
+            Origin::Dev => self.positioned_with_io_in_dev(position, module),
+        };
+
+        let code = self
+            .src_from_origin_and_module_name(origin, module)
+            .expect("Module doesn't exist");
+
+        executor(&mut engine, params, code.into())
     }
 }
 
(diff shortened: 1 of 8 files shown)


GearsDatapacks:
Hi, thank you for this! This seems like something we want, but we generally don't accept unplanned PRs. Would you mind making an issue about this, so we can discuss how exactly we want this to work? I feel I would want this code action available for all modules, not just sub-modules of the current one, though others may have different opinions.

tarkah:
Absolutely! I've moved this to draft for now and created an issue here: https://github.com/gleam-lang/gleam/issues/4894

lpil:
Is this ready for review?

tarkah:
> Is this ready for review?

Yup! I've just moved it off draft.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
         expression => Some(expression),
     }
 }
+
+/// Code action to create an unknown module
+///
+/// ```gleam
+/// // foo.gleam
+/// // Diagnostic: Unknown module
+/// import foo/bar/baz
```
lpil: nit: Remove foobar references please, it has a not-very-cheery WW2 association so we avoid it in favour of more lighthearted examples.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+/// ```
+///
+/// Would create:
+///
+/// ```gleam
+/// // foo/bar/baz.gleam
+/// ```
```
lpil, line 11376: These two code examples are not very clear, so let's describe this code action using words instead.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+        let Some(Error::Type { errors, .. }) = self.error else {
+            return actions;
+        };
+
+        let origin_path = self
+            .module
+            .input_path
+            .as_str()
+            .strip_suffix(&format!("{}.gleam", self.module.name))
+            .expect("origin is ancestor of module path");
```
lpil: Path manipulation can only happen in the `ProjectPaths` class, so pass that into the code action class from the LS engine instance and use `module.origin` to determine which directory to use.

This also enables us to avoid panicking, which we always try to avoid.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+
+        for unknown_module in unknown_modules {
+            if !self.code_action_span.intersects(*unknown_module.location) {
+                continue;
+            }
+
+            let uri = Url::from_file_path(format!("{origin_path}/{}.gleam", unknown_module.name))
+                .expect("origin path is absolute");
+
+            CodeActionBuilder::new(format!("Create module {}.gleam", unknown_module.name))
```
lpil: This should either be `Create module wibble/wobble` or `Create src/wibble/wobble.gleam` as module names don't have a `.gleam` suffix, Gleam file paths do.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+        let unknown_modules = errors.iter().filter_map(|error| {
+            if let TypeError::UnknownModule { name, location, .. } = error {
+                return Some(UnknownModule { name, location });
+            }
+
+            None
+        });
+
+        for unknown_module in unknown_modules {
+            if !self.code_action_span.intersects(*unknown_module.location) {
```
lpil: Contains rather than intersects please. We don't want to trigger this if someone selects a larger block that happens to include this error.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7908,91 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+        error: &'a Option<Error>,
+    ) -> Self {
+        Self {
+            module,
+            error,
+            code_action_span: lsp_range_to_src_span(params.range, lines),
+        }
+    }
+
+    pub fn code_actions(self) -> Vec<CodeAction> {
```
lpil, line 11957: Add comments explaining intent to the code in this function please 🙏

lpil (commented):
Thank you! This is looking really nice. I've left a small comments inline, and the build is failing currently.

Would you mind updating the changelog file too please 🙏

tarkah:
Thanks for the feedback @lpil! I've addressed your comments, let me know if there is anything else that should be addressed.

compiler-core/src/language_server/code_action.rs
```diff
@@ -7895,3 +7909,106 @@ fn single_expression(expression: &TypedExpr) -> Option<&TypedExpr> {
...
+            if !error_contains_action {
+                continue;
+            }
+
+            let uri =
+                Url::from_file_path(format!("{origin_directory}/{}.gleam", unknown_module.name))
+                    .expect("origin directory is absolute");
+
+            CodeActionBuilder::new(format!(
+                "Create module {}/{}.gleam",
```
lpil:
```suggestion
                "Create {}/{}.gleam",
```

lpil (commented):
Fantastic! Thank you! Looks really good.

I missed in the last review that there are no tests, sorry about that. Could you add tests, including ones that make sure that `src`, `test`, and `dev` are appropriately picked for the new file.

tarkah:
@lpil Thanks! Tests are added. Let me know what you think of the refactoring needed to support defining an origin & module with `TestProject::at` in a backwards compatible way.

compiler-core/src/language_server/tests.rs
```diff
@@ -841,3 +898,14 @@ pub fn apply_code_edit(src: &str, mut change: Vec<lsp_types::TextEdit>) -> Strin
 
     result
 }
+
+pub enum Cursor<'a> {
+    Root(Position),
+    Module(Origin, &'a str, Position),
+}
+
+impl From<Position> for Cursor<'_> {
```
lpil: Never implement traits in this codebase please.

compiler-core/src/language_server/tests.rs
```diff
@@ -695,16 +727,41 @@ impl<'a> TestProject<'a> {
...
 
-        executor(&mut engine, params, self.src.into())
+        let find_module = |modules: &[(&'a str, &'a str)], module: &'a str| {
+            modules
+                .iter()
+                .find_map(|(name, code)| (*name == module).then_some(*code))
+                .expect("Module doesn't exist")
+        };
+
+        let ((mut engine, params), code) = match cursor {
```
lpil: Rather than overloading this function please create a new function for positioning within a different file, if needed.

compiler-core/src/language_server/tests/action.rs
```diff
@@ -9634,3 +9738,98 @@ pub fn main() {
         find_position_of("subtract").to_selection()
     );
 }
+
+#[test]
+fn create_unknown_module_under_src() {
+    assert_code_action_file_operations!(
```
lpil: Could we not teach the existing macro how to understand these file operation changes too? It has the nice output which we want to have here, and it would be good to to have to maintain twice as many macros.

language-server/src/tests.rs
```diff
@@ -663,8 +697,7 @@ impl<'a> TestProject<'a> {
         // Add the final module we're going to be positioning the cursor in.
         _ = io.src_module("app", self.src);
 
-        let response = engine.compile_please();
-        assert!(response.result.is_ok());
+        let _response = engine.compile_please();
```
lpil, line 743: 🔴 Compile errors no longer cause the tests to fail!

lpil (commented):
Thank you, very nice. The tests look great, but I've left some notes inline about the changes to the test API. Once that is sorted we can merge, thanks!

language-server/src/tests.rs
tarkah: This code action requires the type error to be present, `positioned_with_io` allows them so I can test when invoking code actions in `self.src` but if we assert on compiler errors under `positioned_with_io_in_src/dev/test` then I can't trigger this code action within specific modules.

How do you suggest we go about that?

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__create_unknown_module_under_dev.snap
```diff
@@ -0,0 +1,6 @@
+---
+source: compiler-core/src/language_server/tests/action.rs
+expression: output
+---
+// --- File operations
+- Create file:///dev/wobble/woo.gleam
```
tarkah, line 6: It seems snapshots including paths don't work since they have different output on windows, hence these tests are failing under windows. I'll change from doing snapshots to just normal asserts in code.

tarkah:
@lpil Appreciate the great feedback. I've addressed your comments, though see https://github.com/gleam-lang/gleam/pull/4888#discussion_r2323652837

language-server/src/tests.rs
lpil: We have existing code actions that only trigger when code fails to compile, how are those tested?

tarkah: They all only use `TestProject::at / positioned_with_io` which doesn't make this assertion

tarkah: Hey @lpil, friendly bump. How'd you like to move forward on this?

tarkah:
Hey @lpil checking in on this, can I get another review / finalize a decision on https://github.com/gleam-lang/gleam/pull/4888#discussion_r2328702587? 

I've rebased and this is otherwise good to go.

lpil (commented):
Sorry, I don't understand why so many any so large changes have been made to the test APIs, there doesn't seem to be anything particularly unusual about this new code action.

Why is it that the tests need to be made so much more complex?

tarkah:
> Why is it that the tests need to be made so much more complex?

All changes revolve around needing a `TestProject::in_module_at` function that allows me to trigger the code action from any origin & module. `TestProject::at` only allows executing code actions from the root `src` module.

This code action's behavior is dependent on which origin it is invoked in and I need to test that the new module is properly created within that origin. It also seems generally useful to be able to trigger a code action from anywhere, not just the root `src` module, so future tests should be able to benefit from this.

I'm happy to abandon these changes and only test the code action against `Origin::Src` but then we don't have great test coverage on it.

tarkah:
@lpil friendly bump. I've rebased this and am curious if you have opinions based on my response here: https://github.com/gleam-lang/gleam/pull/4888#issuecomment-3433340800

This is a very handy action and I'd love to see it land. I'm happy to make any required changes to the test code. Thanks!

lpil:
Apologies in taking so long to get back to you here, my inbox has got out of control.

Your explanation seems reasonable to me, thank you. Let's go with the approach you've suggested.

tarkah:
> Apologies in taking so long to get back to you here, my inbox has got out of control.
> 
> Your explanation seems reasonable to me, thank you. Let's go with the approach you've suggested.

No worries at all! Sounds great, I've gone ahead and rebased so it should be good to go :+1:

lpil (approved):
Perfect!!! Thank you very much!!

lpil:
Oops, looks like there's some test failures. Could you look at that please 🙏

giacomocavalieri:
Looks like paths are not normalised so it fails on windows. Shouldn't be too hard to fix from the looks of it!

tarkah:
Oops, I don't have a windows machine to test, but the revised change should succeed on windows.

lpil:
Oh no, it's still failing! Sorry this is such a pain.

I thought we had some code to do this already, but I can't find it now. Can anyone else spot it? 🤔

giacomocavalieri:
I ran into that function by chance yesterday!! It should be this one https://github.com/gleam-lang/gleam/blob/ce919f72b236694d9c3bcd7a7f225fd6b8e63fc6/test-helpers-rs/src/lib.rs#L108

tarkah:
> Oh no, it's still failing! Sorry this is such a pain.
> 
> I thought we had some code to do this already, but I can't find it now. Can anyone else spot it? 🤔

I still can't test directly (no windows machine), but I believe it should be fixed now. The issue is windows paths have both a Prefix and Root, but I was just skipping the Prefix. I now filter out both which should fix it.

Sorry for the long delay & back / forth on this one :sweat_smile: Thankfully rebasing has been painless. 

https://github.com/gleam-lang/gleam/pull/4888/changes#diff-f48b7f8255f6c984f3be0e1fd47691017300b670cf78654428914123ea535c0eR127

lpil (approved):
Thank you!!!!!

PR #4890, opened 2025-08-28
Stop suggesting to use value discarded in a different function

### LINKED ISSUE #4889: Wrong suggestion for discarded argument
When using a name that is the same as a discarded argument the compiler now gives a nice hint. But I noticed that the hint carries out to functions where that name is not discarded!
```gleam
pub fn main(_x) {
  todo
}

pub fn wibble() {
  x
}
```
Results in the following error - and wrong hint -
```
error: Unknown variable
  ┌─ /Users/giacomocavalieri/Desktop/prova/src/prova.gleam:8:3
  │
3 │ pub fn main(_x) {
  │             -- This value is discarded
  ·
8 │   x
  │   ^ So this is not in scope

Hint: Change `_x` to `x` or reference another variable
```

I'm working on a fix!

giacomocavalieri:
This PR fixes #4889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 
 - The compiler will now raise warning for inefficient use of `list.length()`
   when trying to check is list empty via `0 < list.length(list)` or
-`list.length(list) > 0` as well as in other cases. For example, the following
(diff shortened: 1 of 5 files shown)


lpil (approved):
Thank you

Issue #4838, opened 2025-08-07
Code action generates incorrect module names for function parameters


daniellionel01:
Hello!

Regarding the code action `generate function` I just ran into an issue where I generated functions for a mist application with a websocket. Here's how to reproduce it:

Initialize a fresh gleam project and install these dependencies:
`$ gleam add mist gleam_http gleam_erlang`

Then add this code to the main module:
```gleam
import gleam/bytes_tree
import gleam/erlang/process
import gleam/http/request.{type Request}
import gleam/http/response.{type Response}
import mist.{type Connection, type ResponseData}

pub fn main() -> Nil {
  let assert Ok(_) =
    fn(request: Request(Connection)) -> Response(ResponseData) {
      case request.path_segments(request) {
        ["ws"] -> serve(request)
        _ -> response.set_body(response.new(404), mist.Bytes(bytes_tree.new()))
      }
    }
    |> mist.new
    |> mist.bind("localhost")
    |> mist.port(1234)
    |> mist.start

  process.sleep_forever()
}

pub fn serve(request: Request(Connection)) {
  mist.websocket(
    request:,
    on_init: init_chat_socket,
    handler: loop_chat_socket,
    on_close: close_chat_socket,
  )
}
```

If you now hover over `loop_chat_socket` and execute the `Generate function` code action, it produces this function:
```gleam
fn loop_chat_socket(
  value: a,
  websocket_message: mist.WebsocketMessage(b),
  websocket_connection: websocket.WebsocketConnection,
) -> mist.Next(a, b) {
  todo
}
```

This is incorrect though, since there is no module named `websocket`. If you change `websocket` to `mist` it works fine. So for some reason it switches up `mist` and `websocket` for the `websocket_connection` parameter.

gleam v1.12.0
macOS Sequoia 15.5

GearsDatapacks:
My guess is it's a type alias to an internal module. That's still something we are trying to figure out how to do correctly, as type aliases are not part of the type system. 
See: #3898

daniellionel01:
> My guess is it's a type alias to an internal module. That's still something we are trying to figure out how to do correctly, as type aliases are not part of the type system. See: [#3898](https://github.com/gleam-lang/gleam/issues/3898)

ohh ok! yeah that makes sense

PR #4891, opened 2025-08-28
Favour public type aliases over internal names when printing types

### LINKED ISSUE #3898: Favour public alias over internal definition for types in the language server
Recently I figured out the LSP correctly resolves types with qualified names, but it follows every aliases. So every time you hover over an `Element(a)` in Lustre, the LSP actually resolves to `vdom.Element(a)`, even when you explicitly imports `Element` from `lustre/element`. 

This is an understandable behavior, because `Element` is defined as

```gleam
pub type Element(msg) = 
  vdom.Element(msg)
```

However, such a behaviour is a bit disappointing, because it does not resolves to what you're looking for, but rather to what it is under-the-hood. Typically, going to definition with LSP brings us to `lustre/internals/vdom.gleam`, which is part of the hidden API of Lustre. Ideally, the LSP should stop resolving types to the alias, and let the user continue diving in the code if needed, which is simple enough thanks to go to definition. 

In the same context, when using the LSP to add type annotations, it completes everything with `vdom.Element(a)`, which is rather disappointing too. 
A workaround can be used by using `import lustre/element.{type Element}`, so the LSP now completes types in an unqualified way — with `Element(a)`.

GearsDatapacks:
This PR closes #3898 and #4838
I've extended the `Names` struct to keep track of all public aliases which alias internal types, across all packages. Then, whenever we're printing an internal type, we check for public aliases which reexport that type, and if there is one, print that instead. I've also applied this process to the documentation printer so generated documentation for packages will use the public type aliases rather than printing `@internal SomeInternalType`.

This doesn't yet apply to public type aliases which do not fully reexport a type. For example, if you have:
```gleam
@internal
pub type SomeInternalType(a)

pub type PartialReexport = SomeInternalType(Int)
```

`PartialReexport` can only be used if referencing `SomeInternalType(Int)`, and not, for example, `SomeInternalType(Float)`. This case is not covered yet, so if generic parameters differ, the internal type name is always used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -152,6 +152,32 @@
   of inline tables if they are missing.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- When generating documentation, the build tool will now print the names of
+  public type aliases instead of internal type names when annotating functions
+  and types. For example, for the following code:
+
+  ```gleam
+  import my_package/internal
+
+  pub type ExternalAlias = internal.InternalRepresentation
+
+  pub fn do_thing() -> ExternalAlias { ... }
+  ```
+
+  This is what the build tool used to generate:
+
+  ```gleam
+  pub fn do_thing() -> @internal InternalRepresentation
+  ```
+
+  Whereas now it will not use the internal name, and instead produce:
+
+  ```gleam
+  pub fn do_thing() -> ExternalAlias
+  ```
+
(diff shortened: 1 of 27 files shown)


inoas:
Thankful noise:
This is super valuable especially as there is no auto reexport nor cyclic modules and I ran into the problem where it would use the internal type aliases with cake as well as lustre.

Thank you so much <3

PR #4892, opened 2025-08-28
Fix bug when inlining shadowed variables

### LINKED ISSUE #4877: Bug with inlining when variable names are duplicated
I've found a bug with inlining function parameters in certain cases. For example:
```gleam
let a = 10
let b = 20

fn(x) {
  let a = 7
  x + a
}(a + b)
```

This gets inlined to:
```gleam
let a = 10
let b = 20
let a = 7
a + b + a
```
Which is wrong. This is adding the new value of `a` twice, whereas it should instead add both `a` variables once each.

I'm not sure what code the inliner should generate here. It can either not inline the `x` parameter, for example:
```gleam
let a = 10
let b = 20
let x = a + b
let a = 7
x + a
```

Or, it can replace the name of one of the conflicting variables, something like:
```gleam
let a = 10
let b = 20
let _a_1 = 7
a + b + _a_1
```

GearsDatapacks:
This PR fixes #4877 
As noted in the comments of one of the functions, ideally we would merge this deduplication logic with that done in the code generators, however for now we need to ensure we don't have critical bugs such as this, so that larger item of work can wait until later.

### DIFF
--- a/compiler-core/src/inline.rs
+++ b/compiler-core/src/inline.rs
@@ -226,9 +226,9 @@ impl Inliner<'_> {
         // that means there is a potential conflict in names and we need to rename
         // the variable.
         if !unique_in_scope && self.position == Position::InlinedFunction {
-            // Prefixing the variable name with an underscore ensures it does
+            // Prefixing the variable name with `_inline_` ensures it does
             // not conflict with other defined variables.
-            let new_name = eco_format!("_{name}_{}", self.variable_number);
+            let new_name = eco_format!("_inline_{name}_{}", self.variable_number);
             self.variable_number += 1;
             _ = self.renamed_variables.insert(name, new_name.clone());
             new_name
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_shadowed_variable.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_shadowed_variable.snap
@@ -29,7 +29,7 @@ main() ->
     A = 10,
     B = 20,
     begin
-        _a_0 = 7,
-        (A + B) + _a_0
+        _inline_a_0 = 7,
+        (A + B) + _inline_a_0
     end,
     A.
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_shadowed_variable_nested.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__inlining__inline_shadowed_variable_nested.snap
@@ -29,11 +29,11 @@ pub fn sum(a, b) {
 -spec sum(integer(), integer()) -> integer().
 sum(A, B) ->
     begin
-        _a_0 = 7,
+        _inline_a_0 = 7,
         begin
-            _a_1 = 10,
-            ((A + B) + _a_0) - _a_1
+            _inline_a_1 = 10,
+            ((A + B) + _inline_a_0) - _inline_a_1
         end,
-        _a_0
+        _inline_a_0
     end,
     A.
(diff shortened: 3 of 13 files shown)


compiler-core/src/inline.rs
```diff
@@ -250,13 +314,162 @@ impl Inliner<'_> {
         Assignment {
             location,
             value: self.expression(value),
-            pattern,
+            pattern: self.register_pattern_variables(pattern),
```
lpil, line 321: Do we need to do anything like this for case expressions too?

GearsDatapacks: Yes we probably do. Good catch!

GearsDatapacks:
Ok, I've made it work for `case` expressions as well!

compiler-core/src/inline.rs
```diff
@@ -177,16 +181,67 @@ struct Inliner<'a> {
...
+    /// move all the deduplicating logic from the code generators to here where
+    /// we perform inlining, but that is a fairly large item of work.
+    fn define_variable(&mut self, name: EcoString) -> EcoString {
+        let unique_in_scope = self.in_scope.insert(name.clone());
+        // If the variable name is already defined, and we are inlining a function,
+        // that means there is a potential conflict in names and we need to rename
+        // the variable.
+        if !unique_in_scope && self.position == Position::InlinedFunction {
+            // Prefixing the variable name with an underscore ensures it does
+            // not conflict with other defined variables.
```
lpil, line 230: Could we retain a capital letter start and use some suffix?

What happens when there's multiple layers of shadowing?

GearsDatapacks: Sorry, not sure what you mean by capital letter start. Gleam variables all start with lowercase letters. We could use a suffix, but it would need to be something that's not valid as a Gleam variable so there's no way it clashes with a different variable.

Also not exactly sure what you mean by multiple layers of shadowing. Do you mean like the code in [this test](https://github.com/gleam-lang/gleam/pull/4892/files#diff-913a58a86e5b96bdb7ee632d1d2384e9e3afeea57aee8a95a26aa8cd0e792989)?

lpil: Sorry, I'm not explaining very well, I think it's because I have a weak understanding of the inlining code compared to the older bits of the codebase.

> Also not exactly sure what you mean by multiple layers of shadowing. Do you mean like the code in [this test](https://github.com/gleam-lang/gleam/pull/4892/files#diff-913a58a86e5b96bdb7ee632d1d2384e9e3afeea57aee8a95a26aa8cd0e792989)?

Yes I think that covers it! Thank you.

> Sorry, not sure what you mean by capital letter start. Gleam variables all start with lowercase letters. We could use a suffix, but it would need to be something that's not valid as a Gleam variable so there's no way it clashes with a different variable.

Sorry I was super unclear there. I meant in the generated Erlang!

That said, now I've looked more at what we do for other codegen I think my previous request of having it generate names with uppercase letters isn't as consist as I thought it was. In the `ast` module we define a bunch of variables that we insert during codegen, and they all use an underscore prefix.

Is it possible to collide with the other generated variables? So if I had an inlined variable called `pipe` and it was inlined as `_pipe`, which is the `|>` variable.

Could we have a prefix or suffix or something that includes the word "inline"? I think that would be a nice addition for clarity, similar to the other variables mentioned above.

P.S. thanks for waiting! I've been slow with this PR.

GearsDatapacks: > Is it possible to collide with the other generated variables? So if I had an inlined variable called `pipe` and it was inlined as `_pipe`, which is the `|>` variable.

I believe the naming schemes are different. Pipes would be `_pipe$x`, while this is `_pipe_x`, because this code is separate from codegen and we use different characters for the deduplication IDs per target, so I went with `_` here which works on both Erlang and JS. As mentioned it would be nice to merge these two pieces of code together so there aren't so many inconsistencies.

> Could we have a prefix or suffix or something that includes the word "inline"? I think that would be a nice addition for clarity, similar to the other variables mentioned above.

Yes that's a good idea! Maybe `_inline_varname_id`?

lpil: Sounds good to me!

lpil (approved):
Thank you!!!

PR #4893, opened 2025-08-28
Warn for unused arguments

giacomocavalieri:
This PR introduces a new warning for arguments that are never actually used in a recursive function but only passed along the recursive calls. For example:

```gleam
import gleam/io

pub fn greet(x, times) {
  case times {
    0 -> Nil
    _ -> {
      io.println("Hello, Joe!")
      greet(x, times - 1)
    }
  }
}
```

Would raise the following warning:

```txt
warning: Unused function argument
  ┌─ /Users/giacomocavalieri/Desktop/prova/src/prova.gleam:3:14
  │
3 │ pub fn greet(x, times) {
  │              ^ This argument is unused
  ·
8 │       greet(x, times - 1)
  │             -

This argument is passed to the function when recursing, but it's never used
for anything.

```

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -1033,6 +1033,13 @@ impl TypedExpr {
         }
     }
 
+    pub fn var_constructor(&self) -> Option<&ValueConstructor> {
+        match self {
+            TypedExpr::Var { constructor, .. } => Some(constructor),
+            _ => None,
+        }
+    }
+
     #[must_use]
     pub(crate) fn is_panic(&self) -> bool {
         match self {
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -1033,9 +1033,11 @@ impl TypedExpr {
         }
     }
 
-    pub fn var_constructor(&self) -> Option<&ValueConstructor> {
+    pub fn var_constructor(&self) -> Option<(&ValueConstructor, &EcoString)> {
         match self {
-            TypedExpr::Var { constructor, .. } => Some(constructor),
+            TypedExpr::Var {
+                constructor, name, ..
+            } => Some((constructor, name)),
             _ => None,
         }
     }
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -877,49 +877,6 @@ impl ValueConstructorVariant {
             ValueConstructorVariant::Record { field_map, .. } => field_map.as_ref(),
         }
     }
-
-    /// If this is the constructor of a module function, returns the function's
-    /// module and name.
-    ///
-    fn function_module_and_name(&self) -> Option<(EcoString, EcoString)> {
-        match self {
-            ValueConstructorVariant::ModuleFn { name, module, .. } => {
-                Some((module.clone(), name.clone()))
-            }
-
-            ValueConstructorVariant::ModuleConstant { .. }
-            | ValueConstructorVariant::LocalConstant { .. }
-            | ValueConstructorVariant::LocalVariable { .. }
-            | ValueConstructorVariant::Record { .. } => None,
-        }
-    }
-
-    /// If this is a constructor for an argument, returns the name of the
-    /// function where it is defined and its index in the function's parameter
-    /// list.
-    ///
-    fn argument_function_and_index(&self) -> Option<(EcoString, usize)> {
-        match self {
-            ValueConstructorVariant::LocalVariable {
-                location: _,
-                origin:
-                    VariableOrigin {
-                        syntax: _,
-                        declaration:
-                            VariableDeclaration::FunctionParameter {
-                                function_name,
-                                index,
-                            },
-                    },
-            } => Some((function_name.clone()?, *index)),
-
-            ValueConstructorVariant::LocalVariable { .. }
-            | ValueConstructorVariant::ModuleConstant { .. }
-            | ValueConstructorVariant::LocalConstant { .. }
-            | ValueConstructorVariant::ModuleFn { .. }
-            | ValueConstructorVariant::Record { .. } => None,
-        }
-    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -877,6 +877,49 @@ impl ValueConstructorVariant {
             ValueConstructorVariant::Record { field_map, .. } => field_map.as_ref(),
         }
     }
+
+    /// If this is the constructor of a module function, returns the function's
+    /// module and name.
+    ///
+    fn function_module_and_name(&self) -> Option<(EcoString, EcoString)> {
+        match self {
+            ValueConstructorVariant::ModuleFn { name, module, .. } => {
+                Some((module.clone(), name.clone()))
+            }
+
+            ValueConstructorVariant::ModuleConstant { .. }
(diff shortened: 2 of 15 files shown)


compiler-core/src/type_.rs
```diff
@@ -877,6 +877,49 @@ impl ValueConstructorVariant {
...
+            }
+
+            ValueConstructorVariant::ModuleConstant { .. }
+            | ValueConstructorVariant::LocalConstant { .. }
+            | ValueConstructorVariant::LocalVariable { .. }
+            | ValueConstructorVariant::Record { .. } => None,
+        }
+    }
+
+    /// If this is a constructor for an argument, returns the name of the
```
lpil: What's a constructor for an argument? Arguments don't have constructors, types and values have constructors.

compiler-core/src/ast/typed.rs
```diff
@@ -1029,6 +1029,13 @@ impl TypedExpr {
         }
     }
 
+    pub fn var_constructor(&self) -> Option<&ValueConstructor> {
```
lpil:
```suggestion
    pub fn variable_value_constructor(&self) -> Option<&ValueConstructor> {
```

compiler-core/src/type_/environment.rs
```diff
@@ -770,13 +780,25 @@ impl Environment<'_> {
...
+        unused: HashMap<EcoString, (VariableOrigin, SrcSpan, usize)>,
         problems: &mut Problems,
     ) {
-        for (origin, location, used) in unused.into_values() {
-            if !used {
+        for (origin, location, usages) in unused.into_values() {
+            if usages == 0 {
                 problems.warning(Warning::UnusedVariable { location, origin });
             }
+            // If the function parameter is actually used somewhere, but all the
```
lpil, line 851: It seems a bit rubbish that we now track usage in 2 different ways and then take one and subtract it from the other to get the actual number of usages. Could we instead consider recursing not a usage?

compiler-core/src/type_/error.rs
```diff
@@ -752,13 +752,21 @@ pub enum VariableSyntax {
...
-#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 /// The source of a variable, such as a `let` assignment, or function parameter.
 pub enum VariableDeclaration {
     LetPattern,
     UsePattern,
     ClausePattern,
-    FunctionParameter,
+    FunctionParameter {
+        /// The name of the function defining the parameter. This will be null
```
lpil: There is no null in Rust!

compiler-core/src/type_/expression.rs
```diff
@@ -4428,8 +4458,45 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            .and_then(|constructor| constructor.variant.function_module_and_name())?;
+
+        if self.environment.current_module == module
+            && function == function_defining_argument
+            && argument_index == index_in_argument_list
+        {
+            Some(argument.definition_location().span)
+        } else {
+            None
+        }
```
lpil, line 4551: It's a bit unsatisfying to me that this data is all being fetched on every call when it will have already been fetched just before during analysis.

lpil (commented):
Thanks! I'm not totally sold on the implementation, it's more complexity and more computation than I was hoping for for this feature.

compiler-core/src/type_/expression.rs
giacomocavalieri: Sorry I don't see where this data is fetched before. Here we're checking for the specific case of a function being passed an argument it defines and so they both need to have that specific shape that is not checked elsewhere when typing them.

compiler-core/src/type_/error.rs
giacomocavalieri: Oops

compiler-core/src/type_/environment.rs
giacomocavalieri: I though it would be tricky for two reasons:
- we'd need to pass down some context when typing a variable (we'd need to know if we're typing a call argument and the name and module of the function to tell if we can ignore the usage)
- we'd still need something to store the places where each argument is used to have an error message that is not confusing, so we would still need something like this

compiler-core/src/type_.rs
giacomocavalieri: Yeah not a nice explanation, I'll change it!

compiler-core/src/type_/expression.rs
lpil: What do you mean by have a specific shape?

compiler-core/src/type_/environment.rs
lpil: > we'd need to pass down some context when typing a variable (we'd need to know if we're typing a call argument and the name and module of the function to tell if we can ignore the usage)

Looks like we already do this in `infer_call_argument` for anonymous function arguments. Could do something similar there for variable arguments. This seems like it would be simpler, remove the new third usage tracking system, and avoid fetching the state a second time.

> we'd still need something to store the places where each argument is used to have an error message that is not confusing, so we would still need something like this

I don't think we need to highlight every place the function is called recursively in the error message, if anything I think that would be noise. A more verbose error message is worse unless it has additional information that helps the programmer understand.

giacomocavalieri: I think not highlighting any of the recursive calls would be confusing:
- I see an unused argument warning
- I think I can safely remove it, and try doing that
- Now I get an error in the recursive calls, my first thought is "this is a bug, the argument was actually used in all those places"

I think that piece of context is better to have than no additional context. What do you think? I can still remove it if you don't like it, but that was my reasoning

compiler-core/src/type_/expression.rs
giacomocavalieri: We only record the recursive usage if the call argument is a locally defined argument and the function a call to the same function where it's defined. That's the shape we're expecting. I don't see that being checked anywhere else, but you're saying it's being fetched earlier during analysis, I can't find where.

lpil: We don't store the current function calls' arguments, but that it very easy to add.

It is all the other information that is being re-fetched here.

compiler-core/src/type_/environment.rs
lpil: I'm not convinced. The message says it's used recursively, and if the message isn't noisy and verbose they are more likely to read it. I think showing lots of repetitive information is more likely to result in them being confused.

Either way I will not accept a PR that adds a third way to track usage.

giacomocavalieri:
Fixed conflicts, this is ready for review now!

lpil (approved):
Beautiful, thank you very much!!

compiler-core/src/type_/expression.rs
```diff
@@ -4437,8 +4483,13 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 location,
             ),
 
+            // If the argument is a regular var then we specialise.
```
lpil: This comment is a bit vague- specialise how and why?

giacomocavalieri: Oh my I think I have messed up with copy pasting 😆 Fixing it!

lpil (approved):
Amazing, thank you!!!!

inoas:
Thank you, so cool!

PR #4897, opened 2025-08-29
Support "Generate function" code action in other modules

### LINKED ISSUE #4407: Support generate function code action in other modules
```gleam
import wibble

pub fn main() {
  wibble.wobble()
      // ^^^^^^
}
```

Here triggering the code action would create the function in the `wibble` module if there is not already a function with that name.

GearsDatapacks:
Closes #4407 

I'm not super happy with how I've represented invalid `ModuleSelect`s here. It's more or less the same as what we do for invalid `RecordAccess`, but I wonder if there's a better way to do it. Suggestions welcome.

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -188,6 +188,10 @@ pub enum TypedExpr {
     Invalid {
         location: SrcSpan,
         type_: Arc<Type>,
+        /// Extra information about the invalid expression, useful for providing
+        /// addition help or information, such as code actions to fix invalid
+        /// states.
+        extra_information: Option<InvalidExpression>,
     },
 }
 
@@ -1161,4 +1165,12 @@ impl bit_array::GetLiteralValue for TypedExpr {
             None
         }
     }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum InvalidExpression {
+    ModuleSelect {
+        module_name: EcoString,
+        label: EcoString,
+    },
 }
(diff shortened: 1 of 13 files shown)


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__generating_function_in_other_module_uses_local_names.snap
```diff
@@ -0,0 +1,21 @@
...
+
+pub fn main() -> List(Nil) {
+  wibble.wibble(1, #(True, "Hello"))
+         ↑                          
+}
+
+
+----- AFTER ACTION
+// --- Edits applied to module 'wibble'
+import gleam.{type Int as Number, type Bool as Boolean, type String as Text, type Nil as Nothing}
```
lpil, line 17: This is illegal 😆

compiler-core/src/type_/expression.rs
```diff
@@ -1419,7 +1419,54 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            ) => {
+                self.problems.error(Error::UnknownModuleValue {
+                    name: name.clone(),
+                    module_name: module_name.clone(),
+                    location,
+                    value_constructors,
+                    type_with_same_name,
+                    context,
+                });
+                TypedExpr::ModuleSelect {
```
lpil: I share you uncertainty that this is a good way to represent this. Perhaps we could have a new node? `InvalidModuleSelect` or something like that?

GearsDatapacks: That could work, but it adds a whole lot of places where we have to handle that, including codegen where we just have to panic anytime we encounter it. Maybe we could add more information to the `Invalid` variant itself

lpil: Either or seems good.

GearsDatapacks:
I've changed the implementation to store extra data in `TypedExpr::Invalid`. I opted for that option because I thought it's conceivable we might want to store different kinds of information in this way for invalid expressions, and it's easier to just store the information in one expression variant, rather than creating a new one each time

compiler-core/src/ast/typed.rs
```diff
@@ -1162,3 +1166,11 @@ impl bit_array::GetLiteralValue for TypedExpr {
...
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum InvalidExpression {
+    ModuleSelect {
+        module_name: EcoString,
+        label: EcoString,
+    },
+}
```
lpil, line 1176: Very nice

lpil (approved):
Excellent!! I'm gunna use the heck out of this. Thank you!

PR #4898, opened 2025-08-29
fix: codepoint to erlang in pattern bug

bcpeinhardt:
Currently, unicode character literals in bit array string pattern segments are not correctly translated to \x in Erlang. This PR fixes that.

The previous output would have looked like:
```erl
Foo = <<"\x{00A9}foo"/utf8>>,
    _ = case Foo of
        <<"\u{00A9}"/utf8, Rest/bitstring>> ->
            Rest;

        _ ->
            Foo
    end,
```
Notice the mismatch of representation for the same value.


compiler-core/src/erlang/tests/bit_arrays.rs
```diff
@@ -314,3 +314,19 @@ pub fn main() {
 "
     );
 }
+
+#[test]
+fn unicode_character_encoding_in_bit_array_pattern_segment() {
+    assert_erl!(
+        "
+pub fn main() -> Nil {
+  let foo = <<\"\\u{00A9}foo\":utf8>>
```
giacomocavalieri: It's a convention in the compiler to use `wibble`, `wobble`, `woo` instead of `foo`, `bar`, `baz`. Could you update the names here?

compiler-core/src/erlang/tests/bit_arrays.rs
```diff
@@ -314,3 +314,19 @@ pub fn main() {
...
+  let _bits = case foo {
+    <<\"\\u{00A9}\":utf8, rest: bits>> -> rest
+    _ -> foo
+  }
+  Nil
+}
+"
```
giacomocavalieri: To avoid escaping the string quotes in the example you can wrap the string in `r##`, makes it easier to read the test:
```rust
r#"
  no need to escape these " \
"#
```

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@
 
 ### Compiler
 
+- Literals using \u{XXXX} syntax in bit array pattern segments are now translated to
+  Erlang's \x{XXXX} syntax correctly.
```
giacomocavalieri: I'd move this point to the "Bug fixes" section.

```suggestion
- Literals using `\u{XXXX}` syntax in bit array pattern segments are now translated to
  Erlang's `\x{XXXX}` syntax correctly.
```

giacomocavalieri (commented):
Nice catch! Looks great to me I've left just a couple of nits in the code

compiler-core/src/erlang/tests/bit_arrays.rs
bcpeinhardt: Can do! Sorry, old habits haha

CHANGELOG.md
bcpeinhardt: Ah I didn't scroll down far enough to see there was one! I just went "yeah it's in the compiler" haha

compiler-core/src/erlang/tests/bit_arrays.rs
bcpeinhardt: I always forget raw strings exist

bcpeinhardt:
> Nice catch! Looks great to me I've left just a couple of nits in the code

Addressed 😎 Thank you!

giacomocavalieri:
Fantastic! Looks great to me, it's only missing Louis' seal of approval :)

lpil (approved):
Perfect, thank you!!

PR #4901, opened 2025-08-30
Add `big if -> True` to syntax to remove in v2

giacomocavalieri:
I've also updated the v2 document to mark the "shadowing imported values" as complete

This is not fixing the issue, I will open an issue once this is merged and I think it would be a great first issue, so I'll leave it up for grabs :)


lpil (commented):
I love the title of this issue

Issue #4906, opened 2025-08-31
Add warning for use of deprecated `if -> ...` syntax


giacomocavalieri:
Right now the compiler allows writing:
```gleam
case wibble {
  big if -> True
//    ^^ this should be followed by a guard!!
}
```
We can't outright remove the syntax as that would be a breaking change, but we can have the compiler raise a warning that this is a bit of deprecated syntax and will be removed.

To fix this:
- One would need to update the `parse_case_clause_guard` in `compiler-core/parse.rs` so that if `parse_case_clause_guard_inner` returns a `None`, a warning is raised. Looking at how other warning are raised in the same file (using `self.warnings.push`) should help adding this new one!
- Add a test to make sure that the warning looks right, the tests should end up in `compiler-core/src/type_/tests/warnings.rs` and could look something like this:
  ```rs
  fn the_test() {
    assert_warning!("...the code that should raise the warning...")
  }
  ```
- Update the `docs/v2.md` document to mark this as complete

lpil:
Wow, very nicely written ticket with that initial code investigation! Thank you

Courtcircuits:
Can I work on that one ?

giacomocavalieri:
Sure! Thanks for the help!

Courtcircuits:
I just ran the tests after implementing the warning and got this error : 

<img width="1759" height="940" alt="Image" src="https://github.com/user-attachments/assets/16cdba34-7d27-493c-ab85-d64caaea1cf2" />

Is it because you have a guardrail that fails the test if a new behaviour is detected or is it just that my code throws an error ?

GearsDatapacks:
This issue can be closed now!

PR #4907, opened 2025-08-31
Added warning on `big if -> True` syntax

Courtcircuits:
These changes solve #4906 
Now when using the `big if -> True` syntax you should be warned with that lovely message : 

```
Snapshot: empty_guard_clause
Source: compiler-core/src/type_/tests/warnings.rs:4320
────────────────────────────────────────────────────────────────────────────────────────────
Expression:
pub fn main() {
    let wibble = 10
    case wibble {
        10 if -> True
        _ -> False
    }

}
────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────────────────
    9     9 │ }
   10    10 │
   11    11 │ ----- WARNING
   12    12 │ warning: Deprecated empty guard clause 
         13 │+  ┌─ test/path:5:12
   14    14 │
         15 │+5 │         10 if -> True
         16 │+  │            ^^ This should be followed by a guard or removed!!
         17 │+
   17    18 │
   18    19 │ Empty clause within a guard is deprecated.
   19    20 │ To keep its behaviour, either remove the guard or fill the clause with
   20    21 │ `True` instead.
────────────┴───────────────────────────────────────────────────────────────────────────────
```

> Snippet from `cargo insta`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -381,4 +381,4 @@
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
 - Fixed a bug where the compiler allowed to write a guard with an empty clause.
-  ([Tristan-Mihai Radulescu][https://github.com/Courtcircuits])
+  ([Tristan-Mihai Radulescu](https://github.com/Courtcircuits))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -379,3 +379,6 @@
 - Fixed a bug where the compiler would suggest to use a discarded value defined
   in a different function.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the compiler allowed to write a guard with an empty clause.
+  ([Tristan-Mihai Radulescu][https://github.com/Courtcircuits])
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1622,11 +1622,19 @@ where
     //   if a < b
     //   if a < b || b < c
     fn parse_case_clause_guard(&mut self) -> Result<Option<UntypedClauseGuard>, ParseError> {
-        if self.maybe_one(&Token::If).is_some() {
-            self.parse_clause_guard_inner()
-        } else {
-            Ok(None)
+        let Some((start, end)) = self.maybe_one(&Token::If) else {
+            return Ok(None);
+        };
+        let guard_clause_result = self.parse_clause_guard_inner();
+        // if inner clause is none, a warning should be thrown to prevent that this behaviour
+        // is deprecated
+        if let Ok(None) = guard_clause_result {
+            self.warnings
+                .push(DeprecatedSyntaxWarning::DeprecatedEmptyGuardClause {
+                    location: SrcSpan { start, end },
+                });
         }
+        guard_clause_result
     }
 
     fn parse_clause_guard_inner(&mut self) -> Result<Option<UntypedClauseGuard>, ParseError> {
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1625,16 +1625,16 @@ where
         let Some((start, end)) = self.maybe_one(&Token::If) else {
             return Ok(None);
         };
-        let guard_clause_result = self.parse_clause_guard_inner();
-        // if inner clause is none, a warning should be thrown to prevent that this behaviour
-        // is deprecated
-        if let Ok(None) = guard_clause_result {
+        let clause_guard_result = self.parse_clause_guard_inner();
+        // If inner clause is none, a warning should be shown to let the user
+        // know that empty clauses in guards are deprecated.
+        if let Ok(None) = clause_guard_result {
             self.warnings
-                .push(DeprecatedSyntaxWarning::DeprecatedEmptyGuardClause {
+                .push(DeprecatedSyntaxWarning::DeprecatedEmptyClauseGuard {
                     location: SrcSpan { start, end },
                 });
         }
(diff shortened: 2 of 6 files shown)


compiler-core/src/parse.rs
```diff
@@ -1622,11 +1622,19 @@ where
...
     //   if a < b || b < c
     fn parse_case_clause_guard(&mut self) -> Result<Option<UntypedClauseGuard>, ParseError> {
-        if self.maybe_one(&Token::If).is_some() {
-            self.parse_clause_guard_inner()
-        } else {
-            Ok(None)
+        let Some((start, end)) = self.maybe_one(&Token::If) else {
+            return Ok(None);
+        };
+        let guard_clause_result = self.parse_clause_guard_inner();
```
GearsDatapacks: This is called a `clause_guard`, not a `guard_clause`

compiler-core/src/parse.rs
```diff
@@ -1622,11 +1622,19 @@ where
...
+        // if inner clause is none, a warning should be thrown to prevent that this behaviour
+        // is deprecated
```
GearsDatapacks: Please use capital letters to start comments and end them with a full stop.
The wording also doesn't seem quite right to me. The warning doesn't prevent the behaviour being deprecated, it just tells the user about it.

compiler-core/src/warning.rs
```diff
@@ -242,6 +242,17 @@ pub enum DeprecatedSyntaxWarning {
...
     },
 
+    /// If a guard has an empty clause :
+    /// ```gleam
+    /// case wibble {
+    ///     big if -> True
+    ///         ^^ this should be followed by a guard or removed!!
+    /// }
+    /// ```
+    DeprecatedEmptyGuardClause {
```
GearsDatapacks: Same as above, `ClauseGuard` not `GuardClause`

compiler-core/src/warning.rs
```diff
@@ -355,7 +366,29 @@ To match on all possible lists, use the `_` catch-all pattern instead.",
                     extra_labels: vec![],
                 }),
             },
-
+            Warning::DeprecatedSyntax {
+                path,
+                src,
+                warning: DeprecatedSyntaxWarning::DeprecatedEmptyGuardClause { location },
+            } => Diagnostic {
+                title: "Deprecated empty guard clause syntax".into(),
```
GearsDatapacks: These are just called guards, not guard clauses.

```suggestion
                title: "Deprecated empty guard syntax".into(),
```

compiler-core/src/warning.rs
```diff
@@ -355,7 +366,29 @@ To match on all possible lists, use the `_` catch-all pattern instead.",
...
+                text: wrap(
+                    "Empty clause within a guard is deprecated.
+To keep its behaviour, either remove the guard or fill the clause with `True` instead.",
+                ),
+                hint: None,
+                level: diagnostic::Level::Warning,
+                location: Some(Location {
+                    label: diagnostic::Label {
+                        text: Some("This should be followed by a guard or removed!!
```
GearsDatapacks: Since the deprecated syntax warnings just suggest fixing the syntax in such a way that behaviour stays the same, I think it makes sense to just suggest removing the `if` here.

```suggestion
                        text: Some("This can be removed"),
```

compiler-core/src/warning.rs
```diff
@@ -355,7 +366,29 @@ To match on all possible lists, use the `_` catch-all pattern instead.",
...
             },
-
+            Warning::DeprecatedSyntax {
+                path,
+                src,
+                warning: DeprecatedSyntaxWarning::DeprecatedEmptyGuardClause { location },
+            } => Diagnostic {
+                title: "Deprecated empty guard clause syntax".into(),
+                text: wrap(
+                    "Empty clause within a guard is deprecated.
```
GearsDatapacks: This is just a guard, not a clause within a guard. Also make sure to escape newlines so the `wrap` function works properly.

```suggestion
                    "This syntax for an empty guard is deprecated. \
```

compiler-core/src/warning.rs
```diff
@@ -355,7 +366,29 @@ To match on all possible lists, use the `_` catch-all pattern instead.",
...
+To keep its behaviour, either remove the guard or fill the clause with `True` instead.",
```
GearsDatapacks: Let's not suggest adding `True` as that isn't really good practice.

```suggestion
To have a clause without a guard, remove this.
```

GearsDatapacks (commented):
Thanks for this! The implementation looks good, I've just left a few nits about wording and grammar inline.

Courtcircuits:
@GearsDatapacks thanks a lot for your review ! I just commited your suggestions.

GearsDatapacks (approved):
Looks good to me! Need Louis to review it before being merged though.

CHANGELOG.md
```diff
@@ -379,3 +379,6 @@
 - Fixed a bug where the compiler would suggest to use a discarded value defined
   in a different function.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the compiler allowed to write a guard with an empty clause.
+  ([Tristan-Mihai Radulescu][https://github.com/Courtcircuits])
```
okkdev: The link should be in regular parens `()` 😄

lpil (approved):
Fantastic work!! Thank you so much! 💜💜💜

Courtcircuits:
> Fantastic work!! Thank you so much! 💜💜💜

My pleasure !

PR #4909, opened 2025-08-31
feat(parser): add a hint for `#`-styled comments

### LINKED ISSUE #4908: Improve error message for python-styled comments starting with `#`
In several days I am giving a talk about Gleam and I wanted to refresh my skills :)
So, I wrote some code and added a comment:

```gleam
fn some() {
  let a = 1
  # let b = 2
}
```

And it failed :)

```
error: Syntax error
  ┌─ /src/main.gleam:3:5
  │
3 │   # let b = 2
  │     ^^^ I was not expecting this

Found the keyword `let`, expected one of: 
- `(`

```

Since, unlike Python, Gleam uses `//` for comments.
But, I had to scratch my head for a bit to understand what's wrong.
Because `#` is a valid syntax for tuples, it expects `(` after `#` :)

I propose changing this error message to:

```
error: Syntax error
  ┌─ /src/main.gleam:3:5
  │
3 │   # let b = 2
  │     ^^^ I was not expecting this

Found the keyword `let`, expected one of: 
- `(`
Hint: Maybe you meant to create a comment? Comments in Gleam start with `//`, not `#`
```

It is rather easy to do, so I created a PR for that :)

sobolevn:
Closes #4908


compiler-core/src/parse.rs
```diff
@@ -982,6 +984,16 @@ where
         Ok(Some(expr))
     }
 
+    fn add_comment_style_hint(&self, mut err: ParseError) -> ParseError {
+        if let ParseErrorType::UnexpectedToken { ref mut hint, .. } = err.error {
+            *hint = Some(
+                "Maybe you meant to create a comment? Comments in Gleam start with `//`, not `#`"
```
lpil: Wrap this before 79 columns or use the `wrap` function on this string please 🙏 Will need to update the changelog too

lpil (commented):
Nice and clean! Thank you. One tiny note inline

compiler-core/src/parse.rs
sobolevn: Done! Thanks for the review.

lpil: It's still too long! It also looks rather strange with just the # on the second line. Could you wrap it earlier please 🙏

sobolevn: Sure, done!

sobolevn:
Thanks for the quick review!

PR #4911, opened 2025-08-31
fix(typing): do not report futher errors, after contructor arity is reported

sobolevn:
This is a very cool and not-so-dirty hack that I decided to try :) 
Now, only 1 error is reported, incorrect typing error is gone.


sobolevn:
Ok, this PR affects LSP in a bad way :(
No idea how to solve it yet :)

giacomocavalieri:
I don't think the way to solve the problem is to stop reporting errors after we encounter it. We want analysis to continue, the variable should be assigned the right type

sobolevn:
@giacomocavalieri yes, I agree. I updated my PR :)

PR #4923, opened 2025-09-02
Prevent publishing with default main function

### LINKED ISSUE #4895: Prevent packages from being published with the default main function
Sometimes people accidentally publish packages with the default `main` function included. Let's make it so folks can't do that accidentally.

If they have this main function then `gleam publish` should fail.

```gleam
pub fn main() -> Int {
  io.println("Hello from $PACKAGE_NAME")
}
```

If there's a documentation comment on the function then let's take that as it being intentional and permit the publication.

jcha0713:
This PR prevents accidental publishing of packages with a main function that contains only the default println statement. If the main function has documentation, it will be considered intentional and allowed to publish.

Closes #4895.


jcha0713:
I'm using the existing `HexPackageSquatting` error for this validation, but not sure if this is appropriate. Should I create a new error specific type for this case or is reusing `HexPackageSquatting` acceptable?

GearsDatapacks:
I think you've slightly misunderstood this issue. What we want is to prevent someone writing a whole package, but forgetting to remove the `main` function, so there's accidentally a public `main` function as part of the package's API. The existing package squatting detection already covers the case where a user tries to publish a default package.

jcha0713:
Ah! now I understand it better. Thanks for the comment! @GearsDatapacks

CHANGELOG.md
```diff
@@ -156,6 +156,9 @@
   of inline tables if they are missing.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- `gleam publish` now blocks publishing packages that contain the default main function
+  to prevent accidental publishing of unmodified template code. ([Joohoon Cha](https://github.com/jcha0713))
```
GearsDatapacks: Wrap at 80 columns please and put your name on a new line

compiler-cli/src/publish.rs
```diff
@@ -155,6 +157,61 @@ fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
     Ok(())
 }
 
+// Checks if publishing packages contain default main functions.
+// Main functions with documentation are considered intentional and allowed.
```
GearsDatapacks: Change these to doc comments please (`///`)

compiler-cli/src/publish.rs
```diff
@@ -155,6 +157,61 @@ fn check_for_name_squatting(package: &Package) -> Result<(), Error> {
...
+
+    Ok(())
+}
+
+fn is_default_main(main: &TypedFunction, package_name: &EcoString) -> bool {
+    if main.body.len() != 1 {
+        return false;
+    }
+
+    let Statement::Expression(expr) = main.body.first() else {
```
GearsDatapacks:
```suggestion
    let Statement::Expression(expression) = main.body.first() else {
```
No shorthands please

compiler-core/src/error.rs
```diff
@@ -817,6 +820,27 @@ package deletion or account suspension.
...
 
+            Error::CannotPublishWithDefaultMain { package_name } => {
+                let text = wrap_format!(
+                    "Packages with the default main function cannot be published
+
+Remove or modify the main function that contains only:
+    `io.println(\"Hello from {package_name}!\")`
+
+Alternatively, add documentation to the main function \
+to indicate it is intentional.
```
GearsDatapacks: I'm not 100% sure about this error message. Maybe something like:
```
Seems like you may have forgotten to remove the default generated `main` function
for this project. If you meant to include it in the public API of your package, add a
documentation comment. Otherwise, remove it and try again.
```

GearsDatapacks (commented):
Thanks for this! Looking good. I've left a few comments about style and wording.

I'm not sure exactly what the error message for this should look like. I've left a suggestion inline, but maybe good to get others' opinions on it.

.github/workflows/ci.yaml
```diff
@@ -498,3 +498,11 @@ jobs:
...
         run: make test-all
         working-directory: ./test/assert
+
+      - name: Test publishing with default main
+        run: ./test.sh
+        working-directory: ./test/default_main
+
+      - name: Test publishing with default main and doc
+        run: ./test.sh
+        working-directory: ./test/default_main_with_doc
```
lpil: Very cool to see these tests. Just to make it a bit extra clear for someone skimming over the project can we put publish in the name please.

I don't think we need a test for the doc comment, it's not something we need to do thorough regression testing for. Could you remove that test please 🙏 

```suggestion
        working-directory: ./test/publishing_default_main
```

compiler-core/src/error.rs
lpil: Let's remove the doc comment bit, it's not something we want to push people towards doing.

jcha0713:
Thanks for the feedback! I've updated the code.

lpil (approved):
Thank you!!

PR #4926, opened 2025-09-03
Make it harder to publish `gleam_*` packages

### LINKED ISSUE #4793: Make it even harder to publish a `gleam_*` package
Someone typed "I am part of the Gleam core team" to publish a package with the `gleam_` prefix, so I guess the honesty system isn't enough.

Perhaps we should put a password in the Gleam binary itself?

GearsDatapacks:
Closes #4793 

I went with the proposed approach of putting a password into the source code, but not printing it so a bad actor would need to read the source code. I haven't made a changelog entry as I feel it's better to not draw too much attention to this change.

The password is just a random suggestion, change it if you want.

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -21,6 +21,8 @@ use std::{collections::HashMap, io::Write, path::PathBuf, time::Instant};
 
 use crate::{build, cli, docs, fs, http::HttpClient};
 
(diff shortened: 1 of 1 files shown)


giacomocavalieri:
Based password

lpil (approved):
Flawless, no notes 💜

PR #4928, opened 2025-09-03
Stop warning for modules that are used only by private functions

giacomocavalieri:
We were discussing how this should no longer be a warning, so I fixed it!

### DIFF
--- a/compiler-core/src/reference.rs
+++ b/compiler-core/src/reference.rs
@@ -1,4 +1,4 @@
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use crate::ast::{Publicity, SrcSpan};
 use bimap::{BiMap, Overwritten};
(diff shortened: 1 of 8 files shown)


compiler-core/src/reference.rs
```diff
@@ -440,4 +454,45 @@ impl ReferenceTracker {
...
+        index: NodeIndex,
+    ) {
+        // If the entity is a module there's no way it can reference other
+        // modules so we just ignore it.
+        // This also means that module aliases do not count as using a module!
+        if entity.layer == EntityLayer::Module {
+            return;
+        }
+
+        for node in self.graph.neighbors_directed(index, Direction::Outgoing) {
```
lpil, line 471: Oh neat, nice idea

lpil (approved):
Perfect, thank you!!

PR #4930, opened 2025-09-03
Fix formatting of panic/todo block messages

### LINKED ISSUE #4929: Formatting `panic as {}` is inconsistent
Hi !

With the latest version of the formatter, we got a bug with `panic as`.
Before the update, we had the following code:

```gleam
panic as {
  data |> to_string
  <> " is incorrect, please do some stuff"
}
```

With the latest formatter, we have

```gleam
panic as {
    data |> to_string
    <> " is incorrect, please do some stuff"
  }
```

It breaks the `gleam format --check` in the CI, and it breaks the line-guides.
Maybe it's on purpose, but it seems like a bug to me. 😄

giacomocavalieri:
This PR closes #4929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -409,5 +409,9 @@
   in a different function.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the formatter would format a panic message adding more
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!!!

Issue #4924, opened 2025-09-02
Invalid code generated for `let assert` redefining variable


GearsDatapacks:
I have some parsing code which looks something like this:
```gleam
import gleam/string

pub fn main() {
  let string = "Hello, world!"
  let assert Ok(#(prefix, string)) = string.split_once(string, "\n")
    as { "Failed to split: " <> string }
}
```

This gets compiled to the following Erlang code:
```erlang
main() ->
    String = <<"Hello, world!"/utf8>>,
    _assert_subject = gleam@string:split_once(String, <<"\n"/utf8>>),
    case _assert_subject of
        {ok, {Prefix, String@1}} -> _assert_subject;
        _assert_fail ->
            erlang:error(#{gleam_error => let_assert,
                        message => (<<"Failed to split: "/utf8,
                            String@1/binary>>),
                       ...)
    end.
```

In the message, the variable `String@1` is referenced. This is incorrect. That's the value we bind if the pattern succeeds, but if it fails it is unbound, leading to the Erlang compiler producing an error. The message should reference `String`, not `String@1`

lpil:
Ouch! Good find, thank you

PR #4933, opened 2025-09-03
Use correct env when generating assert message for erlang

DanielleMaywood:
Fixes https://github.com/gleam-lang/gleam/issues/4924

This moves the code generation for the message step a little earlier so that the passed `env` isn't polluted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -415,3 +415,7 @@
 - Fixed a bug where switching from a hex dependency to a git dependency would
   result in an error from the compiler.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the compiler would reference a redeclared variable in a let
+  assert message, instead of the original variable, on the Erlang target.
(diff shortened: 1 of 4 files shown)


GearsDatapacks:
Hello! I forgot to mention in the issue, but this bug also applies to the JavaScript target. Do you think you could fix that too? If not, no worries, I can fix it after this is merged.

DanielleMaywood:
@GearsDatapacks I've just taken a look at the JavaScript target code and I think it might take a while for me to even know what the heck is going on 😅

GearsDatapacks:
Fair enough. It should be a pretty similar change, although there might be some complexity with how we handle side effects in JavaScript codegen. But I can do that after this is merged.

Louis, Jak or I will review this when we get the chance!

lpil (approved):
Lovely! Thank you! Would you mind updating the changelog file also?

I'll merge this and we can let someone else pick up the Erlang side.

lpil (approved):
Thank you!

PR #4936, opened 2025-09-04
improve error message for invalid `gleam.toml` dependencies

### LINKED ISSUE #4925: Improve error message for invalid dependency
Right now if I try and use a git dependency and mess up one of the fields the error I get looks like this:
```txt
error: File IO failure

    /Users/giacomocavalieri/Documents/progetti/giacomocavalieri.me/gleam.toml

    TOML parse error at line 6, column 7
  |
6 | jot = { a = "58499d73c0a980f4d6cfb04b7cbfab9198238b76" }
  |       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
data did not match any variant of untagged enum Requirement
```
That big underline is not all that useful and is talking about compiler internals, I think it would be nice if it provided a better error message here. Not sure what it would look like!

giacomocavalieri:
This PR closes #4925

### DIFF
--- a/compiler-core/src/requirement.rs
+++ b/compiler-core/src/requirement.rs
@@ -11,23 +11,11 @@ use serde::Deserialize;
 use serde::de::{self, Deserializer, MapAccess, Visitor};
 use serde::ser::{Serialize, SerializeMap, Serializer};
 
-#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
-#[serde(untagged, remote = "Self", deny_unknown_fields)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 pub enum Requirement {
-    Hex {
-        #[serde(deserialize_with = "deserialise_range")]
-        version: Range,
-    },
-
-    Path {
-        path: Utf8PathBuf,
-    },
-
-    Git {
-        git: EcoString,
-        #[serde(rename = "ref")]
-        ref_: EcoString,
-    },
+    Hex { version: Range },
+    Path { path: Utf8PathBuf },
+    Git { git: EcoString, ref_: EcoString },
 }
 
 impl Requirement {
@@ -90,18 +78,6 @@ impl Serialize for Requirement {
 }
 
 // Deserialization
-
-fn deserialise_range<'de, D>(deserializer: D) -> Result<Range, D::Error>
-where
-    D: Deserializer<'de>,
-{
-    let version = String::deserialize(deserializer)?;
-    Range::new(version).map_err(de::Error::custom)
-}
-
-#[derive(Debug, Copy, Clone)]
-pub struct Void;
-
 impl FromStr for Requirement {
     type Err = Error;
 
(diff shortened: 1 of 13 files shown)


lpil (commented):
The error message is nicer, but now we've moved from serde to a hybrid of serde with ad-hoc replacement of some of its functionality.

Is the expectation that we convert more of our serde code here? For such a tiny improvement in an error message this is a lot of very complex code, and it's something we have to maintain going forward. We have very little time and money so we want to be as efficient as possible, doing high-impact work that has the lowest development cost, but this seems to be the inverse of that.

Is there no way we can improve these error messages without such a cost?

compiler-core/src/requirement.rs
```diff
@@ -129,11 +105,72 @@ impl<'de> Visitor<'de> for RequirementVisitor {
...
 
-    fn visit_map<M>(self, visitor: M) -> Result<Self::Value, M::Error>
+    fn visit_map<M>(self, mut visitor: M) -> Result<Self::Value, M::Error>
     where
         M: MapAccess<'de>,
     {
-        Requirement::deserialize(de::value::MapAccessDeserializer::new(visitor))
+        let Some((key, value)) = visitor.next_entry()? else {
+            return Err(de::Error::custom(
+                "no field found, expecting one of `version`, `path`, `git`, `ref` ",
```
lpil: Is there a way to know to update this string when new fields are added?

compiler-core/src/requirement.rs
```diff
@@ -146,6 +183,93 @@ impl<'de> Deserialize<'de> for Requirement {
...
+        E: de::Error,
+    {
+        match value {
+            "git" => Ok(DependencyField::Git),
+            "path" => Ok(DependencyField::Path),
+            "ref" => Ok(DependencyField::Ref),
+            "version" => Ok(DependencyField::Version),
+            _ => Err(de::Error::unknown_field(
+                value,
+                &["version", "path", "git", "ref"],
```
lpil: And this array

giacomocavalieri:
> For such a tiny improvement in an error message

I don't agree that stopping to expose compiler internals in an error message is a tiny improvement for developer experience.

> Is the expectation that we convert more of our serde code here? 

I don't think I've ran into other confusing error messages like the one this is fixing, so no

> Is there no way we can improve these error messages without such a cost?

I couldn't figure out any, maybe a serde expert can help here. But I'm not one 😅
This is also more boilerplat-y than it needs to be because of the `toml` crate's errors we would get if we parsed to a `&str` directly. Otherwise we could get away with most of the visitors.

lpil:
I don't know what you mean by exposing compiler internals.

This is a tiny and low impact improvement because it's a very uncommon error message for a feature that is rarely used, and the existing error does successfully communicate that there's a mistake in the `gleam.toml` syntax and where in the file that error is.

We have no evidence that anyone has ever got stuck or confused here. As always the approach is to leave things as they are until we have good evidence that there is a problem that we need to solve to meaningfully improve the Gleam experience.

If it was a trivial fix, sure, we could add that. This is the most complex serde code I've ever seen, and it clearly has some sharp edges for future maintenance.

compiler-core/src/requirement.rs
giacomocavalieri: Yeah there's a bit of a trick we could use to make sure adding a new possible field turns into a compile time error we need to address:
```rust
pub enum Wibble {
  A, B
}

pub fn all_wibbles() -> Vec<Wibble> {
    let mut wibbles = Vec::with_capacity(2);
    loop {
        match wibbles.last() {
            None => wibbles.push(Wibble::A),
            Some(Wibble::A) => wibbles.push(Wibble::B),
            Some(Wibble::B) => break
        };
    }
    wibbles
}
```
We could do this for the fields to make sure we never forget one

giacomocavalieri:
> We have no evidence that anyone has ever got stuck or confused here. 

I did! That's why I set up to fix it in the first place.
I find it annoying having to browse to <https://gleam.run/writing-gleam/gleam-toml/> every time I'm trying to use a git dep because I forget what the fields are.

lpil:
You didn't get stuck, you saw exactly where the problem was and opened the documentation to find out what the correct syntax was. Being annoyed that you have to open the documentation isn't a good motivation for adding this complexity and brittleness to the compiler codebase.

Efficiency and project maintainability are our #1 priorities.

compiler-core/src/requirement.rs
```diff
@@ -146,6 +144,31 @@ impl<'de> Deserialize<'de> for Requirement {
     }
 }
 
+fn replace_field<'de, M: MapAccess<'de>, A: Deserialize<'de>>(
```
lpil: What does this do? Can we put this in the other function if it's specific to just that function? It doesn't look like it could be used by anything else, and it's confusing what it is with, especially with such a generic name.

compiler-core/src/requirement.rs
```diff
@@ -119,21 +95,43 @@ impl<'de> Visitor<'de> for RequirementVisitor {
...
+        while let Some(field) = visitor.next_key::<DependencyField>()? {
+            match field {
+                DependencyField::Git => replace_field("git", &mut git, &mut visitor)?,
+                DependencyField::Ref => replace_field("ref", &mut ref_, &mut visitor)?,
+                DependencyField::Path => replace_field("path", &mut path, &mut visitor)?,
+                DependencyField::Version => replace_field("version", &mut version, &mut visitor)?,
+            }
+        }
+
+        match (git, ref_, version, path) {
```
lpil: This pattern matching is quite hard to read and it's going to get a lot worse as it gets bigger.

compiler-core/src/requirement.rs
```diff
@@ -119,21 +95,43 @@ impl<'de> Visitor<'de> for RequirementVisitor {
...
+            (Some(_git), None, None, None) => Err(de::Error::missing_field("ref")),
+
+            (None, None, Some(version), None) => {
+                Requirement::hex(&version).map_err(de::Error::custom)
+            }
+
+            (None, None, None, Some(path)) => Ok(Requirement::Path { path }),
+
+            _ => Err(de::Error::custom(
+                "expecting exactly one of `version`, `path`, or both `git` and `ref`",
```
lpil: Is there a strategy to updating this?

PR #4938, opened 2025-09-05
Implement new API for custom types on the JavaScript target

GearsDatapacks:
As discussed in #4263 

This PR implements the new JavaScript API for FFI code. A couple of notes:
- Most of the changes here are snapshots. The main change is in `javascript.rs`, and I've also modified the JavaScript prelude.
- For the prelude, some of the classes there don't exactly fit the mold of a Gleam custom type. I've given my best guess as to what is useful to expose there, but probably going through the process of updating existing packages to the new API will be a good indicator for this.

### DIFF
--- a/compiler-core/templates/prelude.mjs
+++ b/compiler-core/templates/prelude.mjs
@@ -10,6 +10,10 @@ export class CustomType {
   }
 }
 
+export function isCustomType(value) {
+  return value instanceof CustomType;
+}
+
 export class List {
   static fromArray(array, tail) {
     let t = tail || new Empty();
@@ -53,6 +57,14 @@ export class List {
   }
 }
 
+export function isList(value) {
+  return value instanceof List;
+}
+
+export function listToArray(list) {
+  return list.toArray();
+}
+
 // @internal
 export function prepend(element, tail) {
   return new NonEmpty(element, tail);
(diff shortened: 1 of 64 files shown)


lpil (commented):
Brilliant!!! Thank you!

Can we use the arrow function syntax please 🙏 

```js
export const Person$Teacher = (name, age, title) =>
  new Teacher(name, age, title);
export const Person$isTeacher = (value) =>
  value instanceof Teacher;
export const Person$Teacher$name = (value) => value.name;
export const Person$Teacher$age = (value) => value.age;
export const Person$Teacher$title = (value) => value.title;
```

compiler-core/templates/prelude.mjs
```diff
@@ -10,6 +10,10 @@ export class CustomType {
   }
 }
 
+export function isCustomType(value) {
```
lpil, line 13: Remove this function please 🙏 There's no use case for it

lpil: Remove this function please 🙏

compiler-core/templates/prelude.mjs
```diff
@@ -53,6 +57,14 @@ export class List {
   }
 }
 
+export function isList(value) {
```
lpil: Remove this please 🙏

compiler-core/templates/prelude.mjs
```diff
@@ -1442,6 +1510,10 @@ export class Result extends CustomType {
   }
 }
 
+export function isResult(value) {
+  return value instanceof Result;
+}
```
lpil: Remove this please 🙏

lpil (commented):
Few tiny little things!!

PR #4940, opened 2025-09-05
🐛 Traverse alternative patterns when finding node

### LINKED ISSUE #4935: References not working in case expressions when searching from a pattern that uses `|`
In the following example
```gleam
pub type CaseVariants {
  Variant1
  Variant2
  Variant3
}

fn f(cv: CaseVariants) {
  case cv {
    Variant1 -> todo
    Variant2 -> todo
    ^^^^^^^^
    Variant3 -> todo
  }
}

fn g(cv: CaseVariants) {
  case cv {
    Variant1 | Variant2 | Variant3 -> todo
               ^^^^^^^^   ^^^^^^^^
  }
}
```
In `f`, using the "go to reference" code action works for all the variants in the case patterns. 
However, in function `g`, when using the `|` operator the LS doesn't provide references for either of the underscored variants. It only provides references for the first one.

The other way around actually works, so, when searching for the references of `Variant2`, the one used in `g` is available. But from the non-first patterns when using `|`, "go to definition" and "go to reference" don't work.

fruno-bulax:
Closes #4935

`find_node_at` did not traverse into alternative patterns, so language server actions like go to definition could not be triggered on them.

Note that _clause guards_ are still not traversed! I tried poking at it a little but couldn't figure out how to map them onto the `Located` type. I will open a follow-up issue after this PR is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -328,6 +328,16 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- You can now go to definition, rename, etc. from alternative patterns!
+  ```gleam
+  case wibble {
+    Wibble | Wobble -> 0
+    //         ^- Previously you could not trigger actions from here
+  }
+
+  ```
+  ([fruno](https://github.com/fruno-bulax))
+
 ### Formatter
 
 - The formatter now removes needless multiple negations that are safe to remove.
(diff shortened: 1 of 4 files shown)


lpil (approved):
Ah! Thank you very much. Super clean

PR #4947, opened 2025-09-09
Add an `hex owner transfer` command

giacomocavalieri:
This PR adds a new `owner` subcommand, with a `transfer` command to transfer ownership of packages on hex. In future the `owner` command could gain new commands to add and remove owners as well, since that's supported by Hex

### DIFF
--- a/compiler-cli/src/owner.rs
+++ b/compiler-cli/src/owner.rs
@@ -3,12 +3,13 @@ use gleam_core::{Result, hex};
 use std::time::Instant;
 
 pub fn transfer(package: String, new_owner_username_or_email: String) -> Result<()> {
-    let should_transfer_ownership = cli::confirm(&format!(
-        "Transferring ownership of this package will remove all current owners
-and make {new_owner_username_or_email} its new owner.
+    println!(
+        "Transferring ownership of this package will remove all current owners and make
+{new_owner_username_or_email} its new owner.
 Do you wish to transfer ownership of `{package}` to {new_owner_username_or_email}?",
-    ))?;
+    );
 
+    let should_transfer_ownership = cli::confirm_with_text(&package)?;
     if !should_transfer_ownership {
         println!("Not transferring ownership.");
         return Ok(());


compiler-cli/src/owner.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{cli, http::HttpClient};
+use gleam_core::{Result, hex, paths::ProjectPaths};
+use std::time::Instant;
+
+pub fn transfer(paths: &ProjectPaths, new_owner_username_or_email: String) -> Result<()> {
+    let config = crate::config::root_config(paths)?;
+    let should_transfer_ownership = cli::confirm(&format!(
```
lpil: Can we make them type the name of the Hex package please 🙏 

Doing it to the wrong package by mistake would be devastating.

compiler-cli/src/lib.rs
```diff
@@ -190,6 +191,10 @@ enum Command {
     #[command(subcommand)]
     Hex(Hex),
 
+    /// Deal with package ownership
+    #[command(subcommand)]
+    Owner(Owner),
```
lpil: We already have a namespace for Hex commands, `Hex`! Using that would make it clearer that we're taking about the Hex package owner too.

`gleam hex transfer-ownership`? `gleam hex owner transfer`?

compiler-cli/src/owner.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{cli, http::HttpClient};
+use gleam_core::{Result, hex, paths::ProjectPaths};
+use std::time::Instant;
+
+pub fn transfer(paths: &ProjectPaths, new_owner_username_or_email: String) -> Result<()> {
+    let config = crate::config::root_config(paths)?;
```
lpil: The config isn't needed at all, so take the name of the package as a CLI argument rather than from the project itself. This will make it easier for folks as they don't need to clone (or fake) the repo in order to administer the project.

compiler-cli/src/lib.rs
giacomocavalieri: I like `gleam hex owner transfer` because in future we could also add `gleam hex owner add` and `gleam hex owner remove`. I'll go with that if that's ok for you!

compiler-cli/src/owner.rs
giacomocavalieri: Right! I hadn't thought of this

PR #4948, opened 2025-09-09
Display all changes to package versions after resolution

lpil:
image

### DIFF
--- a/compiler-core/src/manifest.rs
+++ b/compiler-core/src/manifest.rs
@@ -479,3 +479,149 @@ zzz = { version = "> 0.0.0" }
         }
     }
 }
+
+#[derive(Debug)]
+pub struct Resolved {
+    pub manifest: Manifest,
+    pub added: Vec<(EcoString, Version)>,
+    pub changed: Vec<(EcoString, Version, Version)>,
+    pub removed: Vec<EcoString>,
+}
+
+impl Resolved {
+    pub fn any_changes(&self) -> bool {
+        !self.added.is_empty() || !self.changed.is_empty() || !self.removed.is_empty()
+    }
+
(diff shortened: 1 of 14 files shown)


compiler-core/src/manifest.rs
```diff
@@ -479,3 +479,147 @@ zzz = { version = "> 0.0.0" }
         }
     }
 }
+
+#[derive(Debug)]
+pub struct Resolved {
+    pub manifest: Manifest,
+    pub added: Vec<(EcoString, Version)>,
+    pub changed: Vec<(EcoString, Version, Version)>,
```
GearsDatapacks: Maybe worth making a `Change` struct? Took me a second to understand what this 3-tuple meant

compiler-core/src/manifest.rs
```diff
@@ -479,3 +479,147 @@ zzz = { version = "> 0.0.0" }
...
+    pub removed: Vec<EcoString>,
+}
+
+impl Resolved {
+    pub fn any_changes(&self) -> bool {
+        !self.added.is_empty() || !self.changed.is_empty() || !self.removed.is_empty()
+    }
+
+    pub fn add_added(manifest: Manifest) -> Resolved {
```
GearsDatapacks: I don't find the name `add_added` to be very easily understandable, thought not really sure what better to call this. Maybe `all_added` or something? A doc comment could be pretty useful here too.

GearsDatapacks (commented):
Implementation looks good! A couple stylistic notes

compiler-core/src/manifest.rs
lpil: Ah yeah that's what I intended to write, ugly typo!

Issue #4903, opened 2025-08-30
Optimise comparison with singleton custom types on JavaScript


GearsDatapacks:
Currently, if you have some code like this:
```gleam
pub type Wibble {
  Wibble
  Wobble
}

pub fn is_wibble(w: Wibble) -> Bool {
  w == Wibble
}
```
On the JavaScript target, this code is generated for the `is_wibble` function:
```javascript
export function is_wibble(w) {
  return isEqual(w, new Wibble());
}
```

This works, but it's very slow. It needs to construct a new `Wibble`, and then call the `isEqual` function, which supports any shape of data, and so does a lot of extra logic which isn't necessary.

For the case when we're comparing to a "singleton" custom type, i.e. one with no fields, we could instead generate this:
```javascript
export function is_wibble(w) {
  return w instanceof Wibble;
}
```

This is what we generate for pattern matching. With some basic benchmarks, it seems this is over 10x faster:
```
Input               Function                       IPS           Min           P99
100k elements       equality                   40.0693       23.9345       27.1689
100k elements       instanceof                523.1122        1.5164        4.1579
```

giacomocavalieri:
Sounds like a good first issue!

lpil:
Great idea!

We've been talking about introducing singleton values for variants without any fields, so we don't have to construct a new instance for each usage. I wonder if doing `w === the_wibble_singleton || w instanceof Wibble` would improve performance for the singleton, and to what extent it would hurt performance for non-singleton code + how much it would increase compressed code size.

GearsDatapacks:
Here's a benchmark I put together for that, comparing current equality, just an `instanceof` check, just an equality check with a singleton, and the combined approach mentioned above. I've done 3 inputs: One is a list containing 50% `Wibble`s and 50% `Wobble`s. One is a list containing only `Wibble`, the last contains only `Wobble`.
```
Input               Function                       IPS           Min           P99
50/50               equality                   49.0317       18.7387       39.0100
50/50               instanceof                658.4750        1.4840        1.7985
50/50               singleton                 799.4373        1.2025        1.6991
50/50               both                      658.7560        1.4713        2.2350
all Wibble          equality                   41.3033       23.8008       26.1263
all Wibble          instanceof                712.1356        1.3708        1.5471
all Wibble          singleton                 835.5677        1.1676        1.3381
all Wibble          both                      783.8846        1.2418        1.5262
all Wobble          equality                   76.7850       12.7801       13.3744
[...]

PR #4951, opened 2025-09-10
Optimise comparison with singleton custom types on JavaScript #4903. …

re-masashi:
…also fixed some clippy errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,10 @@
   version upgrades has been improved.
   ([Louis Pilfold](https://github.com/lpil))
 
+- `gleam new` now creates the project directory using the confirmed project
+  name when a suggested rename is accepted.
+  ([Adi Salimgereyev](https://github.com/abs0luty))
+
 ### Language server
 
 - The "inline variable" code action can now trigger when used over the let
--- a/compiler-cli/src/new.rs
+++ b/compiler-cli/src/new.rs
@@ -204,7 +204,18 @@ jobs:
 
 impl Creator {
     fn new(options: NewOptions, gleam_version: &'static str) -> Result<Self, Error> {
-        let project_name = get_valid_project_name(options.name.clone(), &options.project_root)?;
+        Self::new_with_confirmation(options, gleam_version, crate::cli::confirm)
+    }
+
+    fn new_with_confirmation(
+        mut options: NewOptions,
+        gleam_version: &'static str,
+        confirm: impl Fn(&str) -> Result<bool, Error>,
+    ) -> Result<Self, Error> {
+        let name =
+            get_valid_project_name(options.name.as_deref(), &options.project_root, &confirm)?;
+        options.project_root = name.project_root(&options.project_root);
+
         let root = get_current_directory()?.join(&options.project_root);
         let src = root.join("src");
         let test = root.join("test");
@@ -218,7 +229,7 @@ impl Creator {
             workflows,
             gleam_version,
             options,
-            project_name,
+            project_name: name.decided().to_string(),
         };
 
         validate_root_folder(&me)?;
@@ -406,16 +417,28 @@ fn suggest_valid_name(invalid_name: &str, reason: &InvalidProjectNameReason) -> 
     }
 }
 
-fn get_valid_project_name(name: Option<String>, project_root: &str) -> Result<String, Error> {
-    let initial_name = match name {
-        Some(name) => name,
-        None => get_foldername(project_root)?,
-    }
-    .trim()
-    .to_string();
+fn get_valid_project_name(
+    provided_name: Option<&str>,
+    project_root: &str,
+    confirm: impl Fn(&str) -> Result<bool, Error>,
+) -> Result<ProjectName, Error> {
+    let initial_name = match provided_name {
+        Some(name) => name.trim().to_string(),
+        None => get_foldername(project_root)?.trim().to_string(),
+    };
 
     let invalid_reason = match validate_name(&initial_name) {
-        Ok(_) => return Ok(initial_name),
+        Ok(_) => {
+            return Ok(match provided_name {
+                Some(_) => ProjectName::Provided {
+                    decided: initial_name,
+                },
+                None => ProjectName::Derived {
+                    folder: initial_name.clone(),
+                    decided: initial_name,
+                },
+            });
+        }
         Err(Error::InvalidProjectName { reason, .. }) => reason,
         Err(error) => return Err(error),
     };
@@ -429,18 +452,29 @@ fn get_valid_project_name(name: Option<String>, project_root: &str) -> Result<St
             });
         }
     };
+
     let prompt_for_suggested_name = error::format_invalid_project_name_error(
         &initial_name,
         &invalid_reason,
         &Some(suggested_name.clone()),
     );
-    match crate::cli::confirm(&prompt_for_suggested_name)? {
-        true => Ok(suggested_name),
-        false => Err(Error::InvalidProjectName {
-            name: initial_name,
-            reason: invalid_reason,
-        }),
+
+    if confirm(&prompt_for_suggested_name)? {
+        return Ok(match provided_name {
+            Some(_) => ProjectName::Provided {
+                decided: suggested_name,
+            },
+            None => ProjectName::Derived {
+                folder: initial_name,
+                decided: suggested_name,
+            },
+        });
     }
+
+    Err(Error::InvalidProjectName {
+        name: initial_name,
+        reason: invalid_reason,
+    })
 }
 
 fn get_foldername(path: &str) -> Result<String, Error> {
@@ -459,5 +493,39 @@ fn get_foldername(path: &str) -> Result<String, Error> {
             .ok_or(Error::UnableToFindProjectRoot {
                 path: path.to_string(),
             }),
+    }
+}
+
+#[derive(Debug, Clone)]
+enum ProjectName {
+    Provided { decided: String },
+    Derived { folder: String, decided: String },
+}
+
+impl ProjectName {
+    fn decided(&self) -> &str {
+        match self {
+            Self::Provided { decided } | Self::Derived { decided, .. } => decided,
+        }
+    }
+
+    fn project_root(&self, current_root: &str) -> String {
+        match self {
+            Self::Provided { .. } => current_root.to_string(),
+            Self::Derived { folder, decided } => {
+                if current_root == "." || folder == decided {
+                    return current_root.to_string();
+                }
+
+                // If the name was invalid and generated suggestion was accepted,
+                // align the directory path with the new name.
+                let original_root = Utf8Path::new(current_root);
+                let new_root = match original_root.parent() {
+                    Some(parent) if !parent.as_str().is_empty() => parent.join(decided),
+                    Some(_) | None => Utf8PathBuf::from(decided),
+                };
+                new_root.to_string()
+            }
+        }
     }
 }
--- a/compiler-cli/src/new/tests.rs
+++ b/compiler-cli/src/new/tests.rs
@@ -1,6 +1,6 @@
 use std::path::PathBuf;
 
-use camino::Utf8PathBuf;
+use camino::{Utf8Path, Utf8PathBuf};
 use gleam_core::Error;
 
 #[test]
@@ -335,6 +335,38 @@ fn skip_existing_git_files_when_skip_git_is_true() {
 
     assert!(path.join("README.md").exists());
     assert!(path.join(".gitignore").exists());
+}
+
+#[test]
+fn suggested_project_name_updates_directory() {
+    let tmp = tempfile::tempdir().unwrap();
+    let base = Utf8PathBuf::from_path_buf(tmp.path().to_path_buf()).expect("Non Utf8 Path");
+    let original_root = base.join("gleam_testproject");
+
+    let creator = super::Creator::new_with_confirmation(
+        super::NewOptions {
+            project_root: original_root.to_string(),
+            template: super::Template::Erlang,
+            name: None,
+            skip_git: true,
+            skip_github: true,
+        },
+        "1.0.0-gleam",
+        |_| Ok::<bool, Error>(true),
+    )
+    .unwrap();
+
+    let expected_root = base.join("testproject");
+    assert_eq!(creator.project_name, "testproject");
+    assert_eq!(
+        Utf8Path::new(&creator.options.project_root),
+        expected_root.as_path()
+    );
+
+    creator.run().unwrap();
+
+    assert!(expected_root.exists());
+    assert!(!original_root.exists());
 }
 
 #[test]


compiler-core/src/javascript/expression.rs
```diff
@@ -1536,11 +1536,62 @@ impl<'module, 'a> Generator<'module, 'a> {
             return docvec![left_doc, operator, right_doc];
         }
 
+        // Optimise comparison with singleton custom types on JavaScript (https://gleam-lang/gleam/issues/4903)
```
lpil: Could you remove the link to the issue and write a detailed comment explaining what this code is doing please 🙏 

Here's an example a good example of such a comment:

https://github.com/gleam-lang/gleam/blob/1a5ce6265a0d652440ec17ea46e3e993cb62c967/compiler-core/src/javascript/expression.rs#L1045-L1071

compiler-core/src/javascript/expression.rs
```diff
@@ -1536,11 +1536,62 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+            let right_doc = self
+                .not_in_tail_position(Some(Ordering::Strict), |this| this.wrap_expression(right));
+            let constructor_name = name.to_doc();
+
+            if should_be_equal {
+                return docvec![right_doc, " instanceof ", constructor_name];
+            } else {
+                return docvec!["!(", right_doc, " instanceof ", constructor_name, ")"];
+            }
+        }
```
lpil: Could we extract the duplicated bits to helper methods please 🙏

compiler-core/src/javascript/tests/singleton_comparison_optimisation.rs
```diff
@@ -0,0 +1,120 @@
+use crate::assert_js;
```
lpil: Ah! I this this new module isn't part of the codebase, so these tests never run.

Could you move these tests to the javascript/test/custom_types.rs module please

compiler-core/src/javascript/expression.rs
re-masashi: alright!

compiler-core/src/javascript/expression.rs
re-masashi: surely

compiler-core/src/javascript/tests/singleton_comparison_optimisation.rs
re-masashi: OHHHHHHHHHHHH YES. MB I COMPLETELY MISSED THAT

lpil: This one gets me all the time! Such a Rust footgun

re-masashi:
please review

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__imported_aliased_ok.snap
```diff
@@ -19,14 +17,14 @@ pub fn func() {
...
-import { Ok as Y, CustomType as $CustomType, isEqual } from "../gleam.mjs";
+import { Ok as Y, CustomType as $CustomType } from "../gleam.mjs";
 
 export class Ok extends $CustomType {}
 
 export function func() {
   let $ = (var0) => { return new Y(var0); };
   let y = $;
-  if (isEqual(y, (var0) => { return new Y(var0); })) {
+  if (y instanceof Y) {
```
lpil: There's a bug here! This has gone from being an equality check against the constructor to being a check if `y` is any value produced by that constructor.

compiler-core/src/javascript/expression.rs
```diff
@@ -1536,14 +1536,80 @@ impl<'module, 'a> Generator<'module, 'a> {
...
         // Other types must be compared using structural equality
         let left =
             self.not_in_tail_position(Some(Ordering::Strict), |this| this.wrap_expression(left));
         let right =
             self.not_in_tail_position(Some(Ordering::Strict), |this| this.wrap_expression(right));
+
         self.prelude_equal_call(should_be_equal, left, right)
     }
 
+    fn singleton_equal_helper(
```
lpil: Give this a descriptive name please 🙏

compiler-core/src/javascript/expression.rs
lpil: This hasn't been done yet! The patterns are still duplicated. Could we write something like this?:

```rust
if let Some(document) = self.zero_arity_variant_equality(left, right, should_be_equal) {
    return document;
}
if let Some(document) = self.zero_arity_variant_equality(right, left, should_be_equal) {
    return document;
}

File: compiler-core/src/javascript/expression.rs [hunk 6]
```diff
@@ -1967,12 +2033,62 @@ impl<'module, 'a> Generator<'module, 'a> {
             }
 
             ClauseGuard::Equals { left, right, .. } => {
+                if let (
+                    _,
+                    ClauseGuard::Constant(Constant::Record {
+                        arguments, name, ..
+                    }),
+                    // Check if it's a singleton (no arguments)
```
lpil: No arguments doesn't mean a singleton, it could be use of the constructor as a value.

compiler-core/src/javascript/expression.rs
```diff
@@ -1967,12 +2033,62 @@ impl<'module, 'a> Generator<'module, 'a> {
             }
 
             ClauseGuard::Equals { left, right, .. } => {
+                if let (
+                    _,
```
lpil: This value isn't matched on, so no need to put it in a tuple with the other.

compiler-core/src/javascript/expression.rs
```diff
@@ -1536,14 +1536,80 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+        //  pub fn is_wibble(w: Wibble) -> Bool {
+        //    w == Wibble
+        //  }
+        // ```
+        // Instead of `isEqual(w, new Wibble())`, generate `w instanceof Wibble`
+        // because the first approach needs to construct a new Wibble, and then call the isEqual function,
+        // which supports any shape of data, and so does a lot of extra logic which isn't necessary.
+
+        if let (
+            _,
```
lpil: This value isn't matched on, so no need to put it in a tuple with the other.

compiler-core/src/javascript/tests/custom_types.rs
```diff
@@ -687,3 +687,122 @@ pub type Wibble {
...
+
+pub fn process(s: State) -> String {
+  case s {
+    state if state == Active -> "active"
+    _ -> "inactive"
+  }
+}
+"#,
+    );
+}
```
lpil: Could we get tests for these please:

- The variant is defined in another module and used in a qualified fashion (`module.Variant`)
- The variant is defined in another module and used in an unqualified fashion (`import module.{Variant}`)
- The variant is defined in another module and used in an unqualified fashion while aliased (`import module.{Variant as OtherName}`)

Both in expressions and in case clauses please.

lpil (commented):
Very nice! Thank you. I've left a few more comments inline.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__imported_aliased_ok.snap
re-masashi: oh wait. yea. ill fix it.

compiler-core/src/javascript/expression.rs
re-masashi: ohh yes.

re-masashi:
so, i havent been able to come up with a way to add the optimization in the guard clauses. im reverting to all guard clauses having `==` for now.

re-masashi:
```gleam
import other_module.{A as B}
pub fn func() {
  case B {
    x if x == B -> True
    _ -> False
  }
}
```

what should the generated code be?
if (isEqual(x, new B())) or if (x instanceof B)

GearsDatapacks:
We want this optimisation to work for all equality checks, so instanceof

re-masashi:
if something is aliased, for instance, Ok being aliased to Y,
then how will i know if Y is a singleton?

GearsDatapacks:
As far as I can tell, it doesn't change the method of detection at all. Y is still a record with no fields

re-masashi:
> As far as I can tell, it doesn't change the method of detection at all. Y is still a record with no fields

but since `Ok` has a field, shouldnt `Y` do too?

what should this compile to?
```gleam
import gleam.{Ok as Y}
pub type X {
  Ok
}
pub fn func() {
  case Y {
    y if y == Y -> True
    _ -> False
  }
}
```
this?

```js
  let $ = (var0) => { return new Y(var0); };
  let y = $;
  if (y instanceof Y) {
    return true;
  } else {
    return false;
  }
```

GearsDatapacks:
Nope, Y is not a singleton here.

re-masashi:
im really sorry for asking this but what makes something a singleton? having 0 arity? or what else?

re-masashi:
okay ive figured it out. i just have to clean the code now

GearsDatapacks:
Yes, a singleton as I referred to it in the original issue, though perhaps not the correct term, is a variant of a custom type with no fields, aka 0 arity.

re-masashi:
please review

compiler-core/src/javascript/expression.rs
GearsDatapacks: Looks like this still needs to be renamed

compiler-core/src/javascript/expression.rs
```diff
@@ -1966,16 +2046,40 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+            ClauseGuard::Equals { left, right, .. }
+            | ClauseGuard::NotEquals { left, right, .. } => {
+                let should_be_equal = matches!(guard, ClauseGuard::Equals { .. });
+                let both_ctor_values = ctor_tag(left).is_some() && ctor_tag(right).is_some();
+
+                if !both_ctor_values {
+                    if is_var(left)
+                        && !looks_like_constructor_alias(left)
+                        && let Some(tag) = ctor_tag(right)
+                    {
```
GearsDatapacks: I don't really understand why this uses different logic from the expression comparison. Could you change `singleton_equal_helper` to take a `Document` instead of a `TypedExpr`, and reuse that code here?

compiler-core/src/javascript/expression.rs
```diff
@@ -178,6 +178,26 @@ pub(crate) struct Generator<'module, 'ast> {
...
+fn is_var(g: &TypedClauseGuard) -> bool {
+    matches!(g, ClauseGuard::Var { .. })
+}
+
+fn looks_like_constructor_alias(var: &TypedClauseGuard) -> bool {
+    match var {
+        ClauseGuard::Var { type_, .. } => matches!(&**type_, Type::Fn { .. }), // lambda
+        _ => false,
+    }
+}
```
GearsDatapacks: It's not really clear what these functions do. The names don't seem to properly describe their behaviour. Although, if you follow the comment above, these functions might not be necessary.

GearsDatapacks (commented):
A couple notes

compiler-core/src/javascript/expression.rs
re-masashi: yeah. it seems like i wont need these functions if i use `Document` instead of `TypedExpr`.

re-masashi: ill do the needful

compiler-core/src/javascript/expression.rs
re-masashi: if it is a constructor alias, then we get a false positive.
the result causes code which returns true or false for all values, but in reality, it should not.

re-masashi: normal expression comparison does not need alias checking like for guards.
in guards, the variable can be an alias for the constructor value itself.

GearsDatapacks: What's a constructor alias? I haven't heard that term used before

re-masashi: not sure if that's the exact term.
but i'm referring to any variable that's an alias for a constructor.
for example

```gleam
let pluh = Ok
```
so pluh is a "constructor alias". makes ample sense imo.

compiler-core/src/javascript/expression.rs
lpil: I agree, I don't think these functions belong here. I think they would be best being inlined into the places they are used 🙏

compiler-core/src/javascript/expression.rs
lpil: This is still outstanding 🙏

compiler-core/src/javascript/expression.rs
lpil: Not sure what a constructor alias is here, but I agree the logic here is a bit hard to understand.

I'd like to see a similar pattern as I suggested here: https://github.com/gleam-lang/gleam/pull/4951/files#r2337440251

The logic I think should be:
1. The pattern is for a variant/variant constructor.
2. No arguments have been supplied.
3. The type of the pattern is a `Named` type. This could be checked with a new `type.is_named()` method, similar to the existing `type.is_result` method.

compiler-core/src/javascript/expression.rs
re-masashi: okay sure

compiler-core/src/javascript/expression.rs
re-masashi: okay. ill do it.

re-masashi:
ive made the necessary changes

compiler-core/src/javascript/expression.rs
```diff
@@ -1966,16 +2028,56 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+                    }
+                }
 
-            ClauseGuard::NotEquals { left, right, .. } => {
-                let left = self.guard(left);
-                let right = self.guard(right);
-                self.prelude_equal_call(false, left, right)
+                let l = self.guard(left);
+                let r = self.guard(right);
+                self.prelude_equal_call(want_equal, l, r)
```
GearsDatapacks: Please do not rename these variables. `l` and `r` are less descriptive names than `left` and `right` and there is no reason to change them.

compiler-core/src/javascript/expression.rs
GearsDatapacks: This still hasn't been done

compiler-core/src/javascript/expression.rs
GearsDatapacks: I still think this code is too complex. Here are my thoughts:
- I'm not sure why the left-hand-side has to be a variable for this check to take place. You haven't imposed this limitation for regular expression comparison, so it seems odd that it would be required in guards
- I also don't know why you have checked that both sides are not constructor values. Again, this was not done for expression comparison so I don't see why it would be different here
- You've explained what a constructor alias is, but not why it needs to be checked for here. Could you give a code example where this check is needed?
- It is also probably a good idea to do what is suggested for the expressions and make one function for checking whether a comparison is eligible for this optimisation, calling it twice with the different orderings

re-masashi:
> I'm not sure why the left-hand-side has to be a variable for this check to take place. You haven't imposed this limitation for regular expression comparison, so it seems odd that it would be required in guards

For normal expression context, either of the sides can be any expression, so the optimization triggers for both.

> I also don't know why you have checked that both sides are not constructor values. Again, this was not done for expression comparison so I don't see why it would be different here

The main intent was to avoid redundant optimizations in cases like `Active == Active`.
but uhh, if the eligibility function does the right type/arity checks, this explicit double check may be unnecessary, as is the case for expressions. ig removing it makes the logic consistent between guards and expressions.

> You've explained what a constructor alias is, but not why it needs to be checked for here. Could you give a code example where this check is needed? 

“Constructor alias” cases happen when a variable is pattern-bound to the constructor value itself.

for example, 
```gleam
import gleam.{Ok as Y}
pub type X {
  Ok
}
pub fn func() {
  case Y {
    y if y == Y -> True
    _ -> False
  }
}
```
so y == Y should not optimize to `instanceof`; otherwise, every instance would match. 

> It is also probably a good idea to do what is suggested for the expressions and make one function for checking whether a comparison is eligible for this optimisation, calling it twice with the different orderings

okay, we can refactor it i guess. ill do that

GearsDatapacks:
I don't see why `Wibble == Wibble` shouldn't be optimised. There's no point in the code anyway, and `new Wibble() instanceof Wibble` works fine. It would simplify the code greatly and there are no drawbacks

re-masashi:
i've made the necessary changes. please check

compiler-core/src/javascript/expression.rs
GearsDatapacks: Once again, this still needs to be done

compiler-core/src/javascript/expression.rs
```diff
@@ -1966,16 +2028,42 @@ impl<'module, 'a> Generator<'module, 'a> {
...
-                let left = self.guard(left);
-                let right = self.guard(right);
-                self.prelude_equal_call(true, left, right)
-            }
+            ClauseGuard::Equals { left, right, .. }
+            | ClauseGuard::NotEquals { left, right, .. } => {
+                fn eligible_singleton_cmp(
+                    lhs: &TypedClauseGuard,
+                    rhs: &TypedClauseGuard,
+                ) -> Option<EcoString> {
```
GearsDatapacks: You haven't really simplified this logic, just refactored it so the code looks slightly different. Could you change it to use the same logic as regular expression comparison please

lpil:
Thank you both!

I think if you focus on that early comment on mine that hasn't been done yet then it'll become clearer how to do the guard side too. https://github.com/gleam-lang/gleam/pull/4951/files#r2336546527

re-masashi:
> Thank you both!
> 
> I think if you focus on that early comment on mine that hasn't been done yet then it'll become clearer how to do the guard side too. https://github.com/gleam-lang/gleam/pull/4951/files#r2336546527

i extracted the redundant parts into helper methods. please lmk if i need to do anything else

lpil (commented):
Hello! Nice, nearly there!

Could you please use the design I requested in my previous comments 🙏 https://github.com/gleam-lang/gleam/pull/4951/files#r2337440251
So instead of having one call to a `handle_singleton_equality` that duplicates the code for left and right, instead of having one function and calling it twice in the `equal` method, and the same pattern used for guards too.

The tests with `Variant Other(...)` look great! Could you also add versions of these tests that check equality against `Other`, to ensure that the optimisation isn't applied when it shouldn't be. Thank you

You'll also need to rebase on main to fix the tests

re-masashi:
okay! ill do the needful. thanks a lot

re-masashi:
by the way, if i keep the logic the same for the clause guard and normal expressions, then the alias cases get a false positive.
```
   12    12 │ 
   13    13 │ 
   14    14 │ ----- COMPILED JAVASCRIPT
   15    15 │ import * as $gleam from "../gleam.mjs";
   16       │-import { Ok as Y, CustomType as $CustomType, isEqual } from "../gleam.mjs";
         16 │+import { Ok as Y, CustomType as $CustomType } from "../gleam.mjs";
   17    17 │ 
   18    18 │ export class Ok extends $CustomType {}
   19    19 │ 
   20    20 │ export function func() {
   21    21 │   let $ = (var0) => { return new Y(var0); };
   22    22 │   let y = $;
   23       │-  if (isEqual(y, (var0) => { return new Y(var0); })) {
         23 │+  if (y instanceof Y) {
   24    24 │     return true;
   25    25 │   } else {
   26    26 │     return false;
   27    27 │   }
   ```

**Comment by GearsDatapacks** [comment]
You must be doing the check differently, because `Y` in that case is not a constructor with 0 arity

**Comment by re-masashi** [comment]
> You must be doing the check differently, because `Y` in that case is not a constructor with 0 arity

yeah it's an alias. i need to check for that

**Comment by re-masashi** [comment]
im getting some new "helper" functions in the generated code. i did not change any part of the code which should trigger something like that. 
```
   12    12 │ ----- COMPILED JAVASCRIPT
   13    13 │ import { CustomType as $CustomType } from "../gleam.mjs";
   14    14 │ 
   15    15 │ export class Wibble extends $CustomType {}
         16 │+export const Wibble$Wibble = () => new Wibble();
         17 │+export const Wibble$isWibble = (value) => value instanceof Wibble;
   16    18 │ 
   17    19 │ export class Wobble extends $CustomType {}
         20 │+export const Wibble$Wobble = () => new Wobble();
         21 │+export const Wibble$isWobble = (value) => value instanceof Wobble;
   18    22 │ 
   19    23 │ export function is_not_wibble(w) {
   20    24 │   return !(w instanceof Wibble);
   21    25 │ }
```
```
   17    17 │     super();
   18    18 │     this.value = value;
   19    19 │   }
   20    20 │ }
         21 │+export const Result$Ok = (value) => new Ok(value);
         22 │+export const Result$isOk = (value) => value instanceof Ok;
         23 │+export const Result$Ok$value = (value) => value.value;
         24 │+export const Result$Ok$0 = (value) => value.value;
   21    25 │ 
   22    26 │ export class Error extends $CustomType {}
         27 │+export const Result$Error = () => new Error();
         28 │+export const Result$isError = (value) => value instanceof Error;
   23    29 │ 
   24    30 │ export function is_error(r) {
   25    31 │   return r instanceof Error;
   26    32 │ }
```

**Comment by re-masashi** [comment]
https://github.com/re-masashi/gleam/commit/377f9fa2839c150fcc9da4c06a1b48ab3ec62658

is it because of this commit?

**Comment by GearsDatapacks** [comment]
Yes, this is a change from `main` which your branch inherited after you rebased. These changes are correct

**Comment by re-masashi** [comment]
okay
ill commit the new changes then

**Comment by re-masashi** [comment]
i hope everything is okay this time. please check

**Comment by lpil** [comment]
> The tests with Variant Other(...) look great! Could you also add versions of these tests that check equality against Other, to ensure that the optimisation isn't applied when it shouldn't be. Thank you

These test are missing 🙏

**Comment by re-masashi** [comment]
> > The tests with Variant Other(...) look great! Could you also add versions of these tests that check equality against Other, to ensure that the optimisation isn't applied when it shouldn't be. Thank you
> 
> These test are missing 🙏

OH! il add them rn. 
also, i forgot that to format AFTER running clippy. ill fix it rn

**Comment by re-masashi** [comment]
can someone please review this :(

**Comment by re-masashi** [comment]
do I need to add or remove or change something?

File: compiler-core/src/javascript/expression.rs [hunk 14]
```diff
@@ -1536,14 +1536,75 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+        {
+            let left_doc = self
+                .not_in_tail_position(Some(Ordering::Strict), |this| this.wrap_expression(left));
+            Some(self.singleton_equal_helper(left_doc, name, should_be_equal))
+        } else {
+            None
+        }
+    }
+
+    fn singleton_equal_helper(
```
GearsDatapacks: Can you give this a more descriptive name please?

compiler-core/src/javascript/expression.rs
```diff
@@ -2078,6 +2149,26 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+        right: &'a TypedClauseGuard,
+        should_be_equal: bool,
+    ) -> Option<Document<'a>> {
+        if let ClauseGuard::Constant(Constant::Record {
+            arguments, name, ..
+        }) = right
+            && arguments.is_empty()
+            && right.type_().is_named()
+            && let ClauseGuard::Var { type_, .. } = left
+            && !matches!(&**type_, Type::Fn { .. })
```
GearsDatapacks: Can we use the same logic as the expression equality here, where we check the `ValueContructor` to see if it has zero arity? I think it would make this easier to read and understand

GearsDatapacks (commented):
Looking good! A couple small notes

re-masashi:
any updates?

GearsDatapacks:
There's still a piece of feedback that you haven't addressed yet

re-masashi:
> There's still a piece of feedback that you haven't addressed yet

oh yeah ill rename the function but im kindof confused about the "same logic" thing. i hope you dont mind elaborating. because a simple zero arity check doesnt work for the case of aliases. so it cannot the logic cannot be 1:1 but please suggest something.

GearsDatapacks:
So currently you use the following check for guards:
```rust
 if let ClauseGuard::Constant(Constant::Record {
            arguments, name, ..
        }) = right
            && arguments.is_empty()
            && right.type_().is_named()
            && let ClauseGuard::Var { type_, .. } = left
            && !matches!(&**type_, Type::Fn { .. }) {
            ...
        }
```

I'm suggesting instead to use:
```rust
 if let ClauseGuard::Constant(Constant::Record {
            record_constructor: Some(constructor)
        }) = right
            && let ValueConstructorVariant::Record { arity: 0, .. } = constructor.vaiant {
            ...
        }
```

Which is the same logic as you use for expressions, and makes the check easier to understand in my opinion

re-masashi:
ive done that. thanks a ton for the help.

do i need to handle rebasing etc now?

re-masashi:
fixed the merge conflicts.

re-masashi:
can we merge this if its okay :(

re-masashi:
> Thank you!! Excellent work!!
> 
> Could you update the changelog please 🙏

sure.

re-masashi:
okay! did it

lpil:
I'm not sure why github closed this branch, sorry about that. I must have made a mistake while I was trying to update your branch.

I have rebased your changes into main! Thank you very much!!!

PR #4953, opened 2025-09-10
Support Tangled as repository option in `gleam.toml`

naomieow:


compiler-core/src/docs/source_links.rs
```diff
@@ -63,6 +63,10 @@ impl SourceLinker {
                     format!("https://git.sr.ht/~{user}/{repo}/tree/{tag}/item/{path_in_repo}#L"),
                     "-".into(),
                 )),
+                Repository::Tangled { user, repo, .. } => Some((
+                    format!("https://tangled.sh/{user}/{repo}/tree/{tag}/{path_in_repo}#L"),
+                    "-".into(),
```
lpil, line 68: I'm not sure this is right- I think this would be a URL that this could output, but if I click on it nothing is highlighted: https://tangled.sh/@smokesignal.events/quickdid/blob/main/src/cache.rs#L15-20

naomieow: Tangled currently doesn't support multiline highlights, but I have a pending pull request upstream to add it with the syntax used 
https://tangled.sh/@tangled.sh/core/pulls/573

lpil: Oh brilliant. Shall we wait to see if they'll accept that then?

naomieow: Makes sense, will update when I hear news.

naomieow: @lpil Upstream PR has been merged and deployed, https://tangled.sh/@smokesignal.events/quickdid/blob/main/src/cache.rs#L15-20 should now behave as expected.

lpil: hero!

lpil (approved):
Thank you!!!

lpil:
Oh sorry, could you rebase on main 🙏

naomieow:
@lpil Rebased! (hopefully correctly haha)

lpil (approved):
Thank you!!

Issue #4958, opened 2025-09-11
Compiler won't mark certain bit array string patterns as unreachable


giacomocavalieri:
If I write:
```gleam
case <<>> {
  <<"\u{0000}a\u{0000}b":utf8>> -> todo
  <<"ab":utf16>> -> todo
  _ -> todo
}
```
The compiler won't warn that the second branch is unreachable since it's matching on exactly the same bytes

giacomocavalieri:
I'm working on a fix!

giacomocavalieri:
Oops this was already fixed by interference pruning I think

PR #4960, opened 2025-09-11
Mark impossible bit array segments as unreachable

### LINKED ISSUE #4959: Compiler won't warn for certain impossible int bit array patterns
If I write a bit array pattern like this one:
```gleam
case todo {
  <<-2:unsigned>> -> todo
  _ -> todo
}
```
The compiler should warn that that branch is unreachable since it's impossible that an unsigned number matches with the value `-2`. This happens in all cases where we're matching on a number that can't be represented with the given size and signedness.

This not being handled properly also leads to some code generation bugs on the javascript target.

giacomocavalieri:
This PR fixes #4959 and #4958

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -65,8 +65,8 @@ pub mod printer;
 
 use crate::{
     ast::{
-        self, AssignName, BitArraySize, Endianness, IntOperator, TypedBitArraySize, TypedClause,
-        TypedPattern, TypedPatternBitArraySegment,
+        self, AssignName, BitArraySize, Endianness, IntOperator, SrcSpan, TypedBitArraySize,
+        TypedClause, TypedPattern, TypedPatternBitArraySegment,
     },
     strings::{
         convert_string_escape_chars, length_utf16, length_utf32, string_to_utf16_bytes,
@@ -410,7 +410,7 @@ impl Body {
     fn assign_segment_constant_value(&mut self, name: EcoString, value: &BitArrayMatchedValue) {
         let value = match value {
             BitArrayMatchedValue::LiteralFloat(value) => BoundValue::LiteralFloat(value.clone()),
-            BitArrayMatchedValue::LiteralInt(value) => BoundValue::LiteralInt(value.clone()),
+            BitArrayMatchedValue::LiteralInt { value, .. } => BoundValue::LiteralInt(value.clone()),
             BitArrayMatchedValue::LiteralString { value, .. } => {
                 BoundValue::LiteralString(value.clone())
             }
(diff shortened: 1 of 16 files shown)


compiler-core/src/exhaustiveness.rs
```diff
@@ -1178,11 +1211,55 @@ pub enum BitArrayMatchedValue {
     },
 }
 
+impl PartialEq for BitArrayMatchedValue {
```
lpil: Does this need to be a trait implementation? We always try to avoid traits in favour of regular methods, to keep the codebase as explicit and as easy to understand as possible 🙏

giacomocavalieri: The `Eq` implementation is needed by other pieces of code that handle the decision tree, like the inlining code. I can try removing this but that means we'd have to stop using `==` in a lot of other places and manually implement the equality trait for the decision tree as well

lpil: Right, but in other places where we want to check equality in the way that we actually think equality as working this no longer works as expected. This means the programmer has to know that `==` works one for all types _except_ this one, which is an unreasonable barrier to being able to work easily in the compiler codebase and not in the spirit of how we write code.

Further looking at the implementation I don't think this is an equality check. It's checking that the values held are known to be equal, not that the data structures are equal.

One can easily find all the places where equality is used for this type by removing the trait implementation, so replacing those `==` calls with a call to a descriptively named function would be easy (if boring), and would result in easier to understand code.

giacomocavalieri: Sure, that makes sense! I've updated the code to use a custom function where we need to check for equivalence of bit array checks

compiler-core/src/exhaustiveness.rs
```diff
@@ -1314,6 +1270,59 @@ impl BitArrayMatchedValue {
...
+    /// for the exact same thing in a whole bit array pattern as they could be
+    /// matching at different positions!
+    ///
+    /// ```gleam
+    /// <<1, _>>
+    /// <<_, 1>>
+    /// // Both are matching on the same value but at different positions!
+    /// ```
+    ///
+    fn checking_same_value(&self, other: &Self) -> bool {
```
lpil, line 1285: Doesn't the exhaustiveness checker need this function too? You said it was used there also.

giacomocavalieri: There it used the Eq implementation but I needed to check for sameness and not strict equality only in the `fails_if_failing` (and the other) functions. In all the other places there's no need to change implementation and it's ok to still use Eq like we were doing before

lpil: Fab, thanks!

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__bit_arrays__case_match_unsigned_constant_pattern.snap
```diff
@@ -16,9 +14,5 @@ pub fn go(x) {
 
 ----- COMPILED JAVASCRIPT
 export function go(x) {
-  if (x.bitSize === 8 && x.byteAt(0) === 254) {
-    return 1;
-  } else {
-    return 2;
-  }
+  return 2;
```
lpil, line 17: This is so cool

lpil (approved):
Perfect!! Thank you!!

PR #4963, opened 2025-09-12
Offline package resolution

### LINKED ISSUE #2162: Allow `gleam build` and `gleam test` to share libraries across projects (improve offline support)
Thank you for your work on gleam.

This request is a more fleshed out version of #1069, but as I did not create that issue, I am unable to reopen it.

I wanted to try `gleam`, installed it, ran `gleam new` which succeeded, and got on a flight without wifi, thinking I'll be able to play with it while offline.  

But this was not to be, because neither `gleam build` nor `gleam test` work on first run in a project without an internet connection.

```
$ gleam test
  Resolving versions

An error occurred while determining what dependency packages and versions
should be downloaded.

Decision making failed
```

It would be great if `gleam new` did the steps necessary to have a minimally working version of the package it creates be buildable. 

I imagine that one way to do this would be to use a centralized cache for gleam packages that can be shared across repositories, so that once a user has successfully built one project from a quickstart `gleam new`,  building another project from a barebones `gleam new` would grab those same standard libraries, instead of re-downloading them for every first `gleam build` in any given project.

I do want to acknowledge that `gleam build` and `gleam test` do not require an internet connection on subsequent calls (provided no new libraries are added).

benev0:
requires gleam-lang/hexpm-rust#48
resolves #2162

Caches `/packages/{name}` protobuf as a gz and `/packages/{name}/releases/{version}` as json for use in offline package resolution.  Requests fail-over to use local caches.

TODO:
- [x] notify when cache is used
- [ ] prune dependency resolver to only allow locally available dependency solutions
- [ ] harden or remove json cache

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -13,7 +13,20 @@ use camino::{Utf8Path, Utf8PathBuf};
 use ecow::{EcoString, eco_format};
 use flate2::read::GzDecoder;
 use gleam_core::{
-    build::{Mode, Target, Telemetry}, config::PackageConfig, dependency::{self, PackageFetchError}, error::{FileIoAction, FileKind, ShellCommandFailureReason, StandardIoAction}, hex::{self, HEXPM_PUBLIC_KEY}, io::{FileSystemReader, FileSystemWriter, HttpClient as _, TarUnpacker, WrappedReader}, manifest::{Base16Checksum, Manifest, ManifestPackage, ManifestPackageSource, Resolved}, paths::{global_hexpm_package_release_response_cache, global_hexpm_packages_response_cache, ProjectPaths}, requirement::Requirement, warning::WarningEmitterIO, Error, Result, Warning
+    Error, Result, Warning,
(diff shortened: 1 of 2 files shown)


benev0:
I have opened this because I am ready for feedback on anything/everything. 

Before this is merged: JSON caching is an area of major concern as it is unsigned and unverifiable. I am currently looking into ways to either cache it safely, use another safe source, or validate with another safe source (match checksum from protobuf content).

compiler-cli/src/dependencies.rs
```diff
@@ -1093,8 +1095,44 @@ async fn lookup_package(
...
+                            path: cache_path.clone(),
+                            err: Some(err.to_string()),
+                        })?;
+
+                    serde_json::from_slice(&cached_result).map_err(|err| Error::FileIo {
+                        action: FileIoAction::Read,
+                        kind: FileKind::File,
+                        path: cache_path,
+                        err: Some(err.to_string()),
+                    })?
```
benev0: cache reading concerns here as json source is unverifiable in current state

PR #4964, opened 2025-09-12
Add labels when filling fields of variant starting with no fields

### LINKED ISSUE #4962: Fill label on variant without parenthesis doesn't work correctly
```gleam
let assert Ok(actor.Star|ted) = todo
```
If do a "fill label" code action on that position above get this
```gleam
let assert Ok(actor.Startepid:, data:d) = todo
```

When the parenthesis is added the code action works as expectec
```gleam
let assert Ok(actor.Star|ted()) = todo
```
Result:
```gleam
let assert Ok(actor.Started(pid:, data:)) = todo
```

I think a better behavior for the code action would be to add parentheses or maybe not be available at all if the parentheses is not there.

---
```
❯ gleam --version
gleam 1.12.0

❯ erl
Erlang/OTP 28 [erts-16.0.2] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit]
```

giacomocavalieri:
This PR closes #4962

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -647,3 +647,7 @@
 - Fixed a bug where the "Extract variable" code action would shadow existing
   variables, constants and function names.
   ([Matias Carlander](https://github.com/matiascr))
+
+- Fixed a bug where the language server would not fill in the missing labels of
+  a pattern correctly, generating invalid code.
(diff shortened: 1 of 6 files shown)


lpil (approved):
Amazing!!! Thank you!!

PR #4965, opened 2025-09-12
Offer unqualify action for type aliases

### LINKED ISSUE #4961: LSP offers incorrect unqualify code action for nested types.
```
pub fn add_event_listener(
  name: String,
  decoder: Decoder(event.Handler(msg)),
                         ^
) -> Effect(msg)
```

If I list the code actions available when my cursor is at the position above indicated by `^`, i see:

> Qualify `Decoder` as `decode.Decoder`

but i would expect to see:

> Unqualify `event.Handler`

this is confusing and wrong.

giacomocavalieri:
This PR closes #4961
The problem was due to the language server not dealing correctly with type aliases.
I've took the chance to add a few comments as the code was not super clear to me before.

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -13,7 +13,6 @@ use crate::{
     build::{Located, Module},
     config::PackageConfig,
     exhaustiveness::CompiledCase,
-    io::{BeamCompiler, CommandExecutor, FileSystemReader, FileSystemWriter},
     language_server::{edits, reference::FindVariableReferences},
     line_numbers::LineNumbers,
     parse::{extra::ModuleExtra, lexer::str_to_keyword},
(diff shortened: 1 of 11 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -1369,21 +1369,26 @@ impl QualifiedConstructor<'_> {
...
+pub struct QualifiedToUnqualifiedImportFirstPass<'a, IO> {
     module: &'a Module,
+    compiler: &'a LspProjectCompiler<FileSystemProxy<IO>>,
     params: &'a CodeActionParams,
     line_numbers: &'a LineNumbers,
     qualified_constructor: Option<QualifiedConstructor<'a>>,
 }
 
-impl<'a> QualifiedToUnqualifiedImportFirstPass<'a> {
+impl<'a, IO: FileSystemReader + FileSystemWriter + BeamCompiler + CommandExecutor + Clone>
```
lpil: It doesn't look like it compiles the project or reads files or anything, so these could be removed?

I don't think code action classes should ever get any IO capabilities

giacomocavalieri: Yeah seems like we've overlooked this, I've removed the IO capabilities from all actions and the completer as well!

Issue #4900, opened 2025-08-29
`Extract variable` shadows another variable


yoshi-monster:
If there is no other place to put the let binding a variable that is still used later may get shadowed. I noticed this when triggering the code action multiple times on a function taking the same type of argument to give them more descriptive names.

```gleam
fn wibble(a, b) {
  a + b
}

fn main() {
  wibble(1, 2)
  //     ^ Extract variable

  let int = 1
  wibble(int, 2)
  //          ^ Extract variable
  
  let int = 1
  let int = 2
  wibble(int, int)
}
```

The easiest thing would probably be to start counting and never shadow an existing variable.

~ 💜

lpil:
Thank you

fruno-bulax:
I think this should be closed since the PR has been merged, no?

lpil:
Thank you! You're right

PR #4966, opened 2025-09-12
Fix shadowing variables for extract variable

matiascr:
Addresses https://github.com/gleam-lang/gleam/issues/4900

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -2766,6 +2766,15 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
     }
 
     fn visit_typed_function(&mut self, fun: &'ast ast::TypedFunction) {
+        let fun_range = self.edits.src_span_to_lsp_range(SrcSpan {
+            start: fun.location.start,
+            end: fun.end_position,
+        });
+
+        if !within(self.params.range, fun_range) {
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -2762,6 +2765,42 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
         });
     }
 
+    fn visit_typed_function(&mut self, fun: &'ast ast::TypedFunction) {
+        // We reset the name generator to purge the variable names from other scopes.
```
giacomocavalieri, line 2778: This will repeat the initialisation for all functions, even ones that do not contain the cursor. I think it would be better to put this behind a `within` check like we do for other code actions

giacomocavalieri:
Very nice, and really thorough tests, I love it! I've just left a comment

compiler-core/src/language_server/code_action.rs
matiascr: You're right, I forgot to add them, I'll add the required checks

matiascr: Fixed in https://github.com/gleam-lang/gleam/pull/4966/commits/e8b9f686078ac324264bd9ddfd08aa3f748c11bf

giacomocavalieri:
Looks good to me, this just needs a final review from Louis

Issue #4950, opened 2025-09-09
LSP code action to ignore label in pattern match


francishamel:
Hi 👋 I'm not sure if I'm using the right term but I think it could be useful to add a code action that appends `_` to a label in a pattern match to ignore it. There's already a hint for this case but it would be neat to quickly fix the issue via a code action. I added a screenshot to illustrate what I mean.

<img width="787" height="199" alt="Image" src="https://github.com/user-attachments/assets/edc91c12-fdee-41ad-a92a-a0750449b96c" />

lpil:
Could be useful! So would it change `prop1:,` to `prop1: _`?

Is there a reason to limit it to labelled arguments? Seems like being able to ignore any unused variable assigning pattern would be useful.

francishamel:
Yeah, that's what I had in mind! 😄 

> Is there a reason to limit it to labelled arguments?

Not at all, I just didn't think further than this specific case tbh but it indeed would make sense to allow this code action on any unused variable assigning pattern.

lpil:
Makes sense! Let's do the more general one :)

clr-cera:
Hi! I would like to try to solve this issue, but it would be my first contribution and I'm a little lost.

Is there documentation explaining the language server structure?
Should this code action be preferred?

GearsDatapacks:
There isn't much documentation on the structure of the project unfortunately. I was working on some but it never got to a finished state. 

You're welcome to try this issue out, but do note that it's not particularly beginner friendly and you may want to have a go at some other more straightforward issues beforehand. Feel free to still give it a shot though!

In order to fix this issue, you will need to do something like the following:
- Create a new function in `code_action.rs` for this code action. You probably want to take inspiration from the "remove unused imports" and "fix name" code actions, as this one will need to derive its behaviour from warnings emitted by the compiler. 
- As to whether it should be preferred, I'm not exactly sure. I would lean on the side of yes; I generally like preferred code actions to be ones where there's really only a single solution to a certain error or warning. Go for preferred and we can always change it later. 
- Once you have your function, you will need to call it from the `code_actions` function in `engine.rs`.
- After that, add tests to `tests/action.rs` to verify that the code action is working as intended. 

Hopefully that helps, let us know if you have any more questions!

vshakitskiy:
I would like to finish this one, if people don't mind 🙂

lpil:
Go for it!

ankddev:
[...]

PR #4967, opened 2025-09-12
Discard unused variable code action

clr-cera:
This PR creates the code action to discarding an unused variable.

References Issue #4950 

As suggested the code action is currently set as preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,31 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The compiler now applies an optimisation known as "interference based pruning"
+  when compiling bit array pattern matching where matches are performed at the
+  start of bit arrays.
+  This optimisation drastically reduces compile times, memory usage and the
+  compiled code size, removing many redundant checks.
+  It is particularly important for network protocol applications where it is
+  typical to match on some fixed patterns at the start of the bitarray.
+  For example:
+
+  ```gleam
+  pub fn parser_headers(headers: BitArray, bytes: Int) -> Headers {
+    case headers {
+      <<"CONTENT_LENGTH" as header, 0, value:size(bytes), 0, rest:bytes>>
+      | <<"QUERY_STRING" as header, 0, value:size(bytes), 0, rest:bytes>>
+      | <<"REQUEST_URI" as header, 0, value:size(bytes), 0, rest:bytes>>
+      // ...
+      | <<"REDIRECT_STATUS" as header, 0, value:size(bytes), 0, rest:bytes>>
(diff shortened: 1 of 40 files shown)


clr-cera:
Hi, I am having two problems with this pr!

First, is the code action always applied after formatting? When the variable is assigned through a pattern, the code action must also remove the ' as ' before the variable name, and I haven't figured out how to find this span.

Also, the tests for the code action are failing, although they produce the right result, how can I fix that?

GearsDatapacks:
1. No, the user isn't guaranteed to have formatted their code before running a code action. What we usually do when a certain new piece of location information is required is to add it to the AST. However in this case, I should think it would be possible to remove the code from the end of the inner pattern to the end of the `as` pattern, both of which you should already have the locations for. 
2. That's because we use snapshot testing! The way it works is that first you run the tests to generate the output, then you manually review it to determine whether it is correct or not (use `cargo insta review`). After that, the correct value is stored and the test will only fail if it differs from the accepted snapshot. See `docs/compiler/README.md` for more information

clr-cera:
> 1. No, the user isn't guaranteed to have formatted their code before running a code action. What we usually do when a certain new piece of location information is required is to add it to the AST. However in this case, I should think it would be possible to remove the code from the end of the inner pattern to the end of the `as` pattern, both of which you should already have the locations for.
> 
>     2. That's because we use snapshot testing! The way it works is that first you run the tests to generate the output, then you manually review it to determine whether it is correct or not (use `cargo insta review`). After that, the correct value is stored and the test will only fail if it differs from the accepted snapshot. See `docs/compiler/README.md` for more information

Thank you! Now I understood the snapshot testing, and wow, it works so well!
I will try to find the locations in the AST

clr-cera:
I couldn't find the assign pattern related to the unused variable warning, I tried to visit all assign patterns locations and check if the variable location start is inside, but this returned no patterns, and I did not think of any other way. 

Is there a way to, given a location, find the location of a string before it?

I will try to fix it later, thanks for helping!

lpil:

clr-cera:
> Hello! Are you still working on this one? :)

Hi!
Oh, I got very busy with uni and personal life by the past few months, and completely forgot! I'm sorry for not writing updates.

I don't think I will be able to work on this issue until next year due to some of my deadlines, and I believe I would need help, as I got stuck finding a way to remove the 'as' in the assign pattern and got nowhere.

If someone wants to solve this issue on another PR, or wants to continue from here, feel free to do so, and I'm here to explain what I've done or discuss it!

lpil:
No worries @clr-cera ! Thank you again. I'll close this, and anyone looking at the linked PR in future can reference thisone.

PR #4968, opened 2025-09-12
Allow rename when hovering external variable

### LINKED ISSUE #4941: Renaming external functions does not work
If you try to rename a function which is marked as external, it fails. For example:
```gleam
pub fn main() {
  halt(42)
}

@external(erlang, "erlang", "halt")
fn halt(code: Int) -> a
// ^ Try to rename here
```
This should allow renaming the function the same as pure-Gleam functions, but for some reason the rename is not allowed here.

giacomocavalieri:
This PR fixes #4941
I went in thinking the fix would be pretty straightforward and it ended up removing one of the oldest TODO I have memory of in the codebase 😆

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -822,7 +822,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
 
     fn ensure_function_has_an_implementation(
         &mut self,
-        body: &Vec<UntypedStatement>,
+        body: &[UntypedStatement],
         external_erlang: &Option<(EcoString, EcoString, SrcSpan)>,
         external_javascript: &Option<(EcoString, EcoString, SrcSpan)>,
         location: SrcSpan,
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -913,7 +913,7 @@ impl<'comments> Formatter<'comments> {
         header.append(" ").append(wrap_block(body)).group()
     }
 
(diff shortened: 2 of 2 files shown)


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__external_fn__reference_to_imported_elixir_external_fn.snap
```diff
@@ -18,7 +18,7 @@ fn id(x) { x }
 -export([main/0]).
 
 -file("project/test/my/mod.gleam", 6).
--spec id(L) -> L.
+-spec id(K) -> K.
 id(X) ->
```
giacomocavalieri, line 22: @lpil I'm particularly interested in what you think about this one here. I can't figure out how this change can affect the generation of type variables here. Anything comes to mind?

giacomocavalieri:
@lpil as you suggested I switched to a `Vec<>` instead of `Option<Vec1<>>` and it was indeed much better! 😆

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__external_fn__reference_to_imported_elixir_external_fn.snap
lpil: I can't work it out either!

PR #4971, opened 2025-09-13
Add code action to insert omitted labels

### LINKED ISSUE #4969: Code action to add labels to a function/constructor
I have the following piece of code where I've built a record with a lot of labels, not specifying any of those:
```gleam
sql.GetDistributorProductsRow(
  "03",
  "s3",
  "b3",
  "sub3",
  Some("c3"),
  "u3",
  Some("d3"),
)
```
It would be nice to have a code action that, when hovering over the `sql.GetDistributorProductsRow`, allows to add all the labels that were omitted

giacomocavalieri:
This PR fixes #4969


compiler-core/src/ast/typed.rs
```diff
@@ -76,6 +76,12 @@ pub enum TypedExpr {
         type_: Arc<Type>,
         fun: Box<Self>,
         arguments: Vec<CallArg<Self>>,
+        /// Whether the call has the right number and type of arguments.
+        /// This is useful for the language server to know if a call did type
+        /// check to offer some specific actions. Otherwise it would have to
+        /// perform type checking again.
+        ///
+        is_well_typed: bool,
```
lpil: I'm not a huge fan of this. I think I would prefer it's a best-attempt that could be wrong when there's a type error, which is what we do for all the other code actions today.

Was there a reason you thought this code action would particularly want this extra check?

giacomocavalieri: I thought it would be annoying because if I trigger it on invalid code it will for sure generate code that needs more fixing: fix the type error _and_ fix the wrong labels that were added. 
I can't think of any code action in the LS that will surely introduce new errors when triggered on invalid code.

lpil: I think it won't be too confusing as the programmer would have to be looking directly at the type error on the call in order to trigger the code action.

Could we go for the simpler approach of not having the check, and then if it does prove to be enough of a drawback on the developer experience we can then look into how we could improve the situation 🙏

giacomocavalieri:
Done! I've removed the well typed check, should be all good now!

PR #4972, opened 2025-09-13
Make sure an empty type parameter list is no longer a syntax error

giacomocavalieri:
If I wrote `Result()` it would be reported as a syntax error instead of a wrong arity error, which would stop analysis from happening altogether. I've made it an analysis error so that it doesn't stop other errors from popping up.


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__qualified_to_unqualified_import_constructor_constructor_name_exists.snap.new
```diff
@@ -1,26 +0,0 @@
----
```
giacomocavalieri: This is a `new` unaccepted snapshot that somehow got merged, sorry for not spotting it earlier!

compiler-core/src/error.rs
```diff
@@ -2229,6 +2229,27 @@ but {given} where provided."
                         }
                     }
 
+                    TypeError::TypeExpectingNoArgumentsAndEmptyArgumentsList { location, name } => {
+                        let text = wrap_format!(
+                            "`{name}` has no type arguments, but here it's given an arguments list."
```
lpil: It has no type _parameters_, but it has been given arguments. Could also say it expects no arguments.

Day the number of arguments given as "arguments list" isn't a term used in the language

compiler-core/src/error.rs
```diff
@@ -2229,6 +2229,27 @@ but {given} where provided."
...
+                        );
+                        Diagnostic {
+                            title: "Incorrect type constructor".into(),
```
lpil: What's incorrect about it? Could this be more specific?

compiler-core/src/type_/error.rs
```diff
@@ -655,6 +655,14 @@ pub enum Error {
         package: EcoString,
         location: SrcSpan,
     },
+
+    /// This happens when a type has no type arguments (for example `Int`) but
```
lpil: Type parameters

compiler-core/src/type_/error.rs
```diff
@@ -655,6 +655,14 @@ pub enum Error {
...
+    /// it is written with an explicit empty argument list: `Int()`.
```
lpil: No such thing as an argument list.

I think it'd be more accurate to say that this error is for a type being used as a type constructor?

compiler-core/src/type_/hydrator.rs
```diff
@@ -183,7 +184,19 @@ impl Hydrator {
                 }
 
                 // Ensure that the correct number of arguments have been given to the constructor
-                if arguments.len() != parameters.len() {
+                //
+                // This is a special case for when the type expects no parameters
```
lpil: Either a type has parameters, or a type constructor expects arguments. A type does not expect arguments.

lpil (commented):
Fab, thank you! I've left some little nitpick-y comments around naming.

giacomocavalieri:
Does it sound better now? I was having a really hard time explaining it 😆

CHANGELOG.md
```diff
@@ -219,6 +223,8 @@
   Comments in Gleam start with `//`, not `#`
   ```
 
+  ([sobolevn](https://github.com/sobolevn))
```
lpil, line 227: oops, my bad. Thank you

PR #4990, opened 2025-09-18
Add code action to extract function

### LINKED ISSUE #4905: Code action: Extract function
Sometimes, I am writing a function and decide that a certain part of it has either become too big, or the same logic is needed elsewhere. In that case, it would be useful to have a code action to extract part of a function into another function. 

The user could select a portion of code, and trigger the code action. The language server would then detect all of the local variables in that portion of code, and convert them into parameters of the new function.

GearsDatapacks:
Closes #4905


lpil (approved):
Amazing!!! Thank you! Super excited for this one

PR #4995, opened 2025-09-19
Fix a bug with JS codegen of pattern matching expression using string literals

### LINKED ISSUE #4993: Wrong compilation of bit array pattern matching with literal string after a sized segment
When I have an unknown sized segment followed by a string literal (with multiple characters), the generated javascript code is invalid:
```gleam
pub fn go(x, n) {
  case x {
    <<_:size(n), "ab">> -> 1
    _ -> 2
  }
}
```
```js
export function go(x, n) {
  if (
    n >= 0 &&
    x.bitSize >= n &&
    x.bitSize === 16 + n &&
    bitArraySliceToInt(x, n, 8 + n, true, false) === 97 &&
      bitArraySliceToInt(x, n, 8 + n, true, false) === 98
  ) {
    return 1;
  } else {
    return 2;
  }
}
```
The bug is really hard to notice, this code produces the right output:
```gleam
pub fn go(x, n) {
  case x {
    <<_:size(n), "a", "b">> -> 1
    _ -> 2
  }
}
```
```js
export function go(x, n) {
  if (
    n >= 0 &&
    x.bitSize >= n &&
    x.bitSize >= 8 + n &&
    bitArraySliceToInt(x, n, 8 + n, true, false) === 97 &&
    x.bitSize === 16 + n &&
    bitArraySliceToInt(x, 8 + n, 16 + n, true, false) === 98
  ) {
    return 1;
  } else {
    return 2;
  }
}
```

Looking at the diff, it seems like the offset is not incremented properly if the string is written as a single chunk:
```diff
export function go(x, n) {
  if (
    n >= 0 &&
    x.bitSize >= n &&
    x.bitSize === 16 + n &&
    bitArraySliceToInt(x, n, 8 + n, true, false) === 97 &&
-   bitArraySliceToInt(x, n, 8 + n, true, false) === 98
+   bitArraySliceToInt(x, 8 + n, 16 + n, true, false) === 98
  ) {
    return 1;
  } else {
    return 2;
  }
}
```

I'm working on a fix!

giacomocavalieri:
This PR fixes #4993 
Luckily the bug happened in the code generation step and not in the exhaustiveness checking one! That made it an easy fix 😁

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -714,3 +714,7 @@
 - Fixed a bug where running `gleam update` would not properly update git
   dependencies unless `gleam clean` was run first.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where the compiler would produce wrong JavaScript code for binary
+  pattern matching expressions using literal strings and byte segments.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -1443,14 +1443,18 @@ impl<'generator, 'module, 'a> Variables<'generator, 'module, 'a> {
                 from_byte += 1;
             }
         } else {
+            let mut start = start.clone();
+
             // If the string doesn't start at a byte aligned offset then we'll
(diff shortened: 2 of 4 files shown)


lpil (approved):
Thank you!!

PR #5000, opened 2025-09-21
Add erlang.application_start_argument configuration

lpil:
The previous configuration allowed the programmer to set the module, but not the argument. This is now supported.

In future they may be a more sophisticated system for defining stateful OTP applications, but for today this gives the programmer full control over the application module OTP configuration.


lpil:
Thank you!

PR #5003, opened 2025-09-21
Perform inlining in anonymous functions

### LINKED ISSUE #5002: Function inlining does not take place inside anonymous functions
It seems I forgot to apply inlining to anonymous functions. For example if you have the following code:
```gleam
import gleam/result

fn some_function_that_is_not_inlined(f) {
  f()
}

pub fn main() {
  use <- some_function_that_is_not_inlined()
  use x <- result.map(Ok(1))
  x + 1
}
```

This is compiled to the following Erlang code:
```erlang
main() ->
    some_function_that_is_not_inlined(
        fun() -> gleam@result:map({ok, 1}, fun(X) -> X + 1 end) end
    ).
```

Whereas the compiler should apply inlining to the code inside the anonymous function too.
(This is example is Erlang, but the same applies to the JavaScript target)

GearsDatapacks:
Fixes #5002


GearsDatapacks:
Closing due to #5010. I will work on this locally in combination with the other inlining fixes

PR #5006, opened 2025-09-22
Format alias fmt

unsecretised:
#5005 

Adds an alias to be able to do `gleam fmt` as well as `gleam format`


lpil:
Thank you for the contribution! Unfortunately I can't accept this contribution as it's not something we have planned.

If you would like help setting up an alias in your shell let me know.

PR #5011, opened 2025-09-24
fix: formatter moves comments out of tuple return type

### LINKED ISSUE #5004: Formatter moves comments out of tuple return types
I found a couple odd cases with the formatter when adding comments inside of tuple return types.

Both of these inputs...

```gleam
fn parse_data(str: String) -> #(
  List(String), // qualifications
) {
  todo
}
```
```gleam
fn parse_data(str: String) -> #(
  // qualifications
  List(String),
) {
  todo
}
```

...yield this output:

```gleam
fn parse_data(
  str: String,
  // qualifications
) -> #(List(String)) {
  todo
}
```

I get the same results using either `gleam format` from the command line (Gleam 1.12.0) or "Format Document" in VSCode (VSCode 1.104.0; Gleam VSCode extension 2.12.1).

ankddev:
Closes #5004 
Fix formatting of tuple return types: now comments are not moved out.
### Showcase
```
$ cat src\test_gleam.gleam
import gleam/io

pub fn main() -> Nil {
  io.println("Hello from test_gleam!")
}

pub fn hello() -> #(
  // test 0
  String, // test 1
) {
  #("hello")
}

$ ..\..\RustroverProjects\gleam\target\debug\gleam format

$ cat src\test_gleam.gleam
import gleam/io

pub fn main() -> Nil {
  io.println("Hello from test_gleam!")
}

pub fn hello() -> #(
  String,
  // test 0
  // test 1
) {
  #("hello")
}
```

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -834,7 +834,18 @@ impl<'comments> Formatter<'comments> {
                     .expect("Function in a statement must be named")
                     .1,
             )
-            .append(self.wrap_arguments(arguments, function.location.end));
+            .append(
+                self.wrap_arguments(
+                    // We count end location of arguments here.
+                    // Before this change, wrap_arguments consumed all comments
+                    // leading in moving comments from return annotation to argument list
+                    arguments,
+                    function
+                        .return_annotation
+                        .as_ref()
+                        .map_or(function.location.end, |ann| ann.location().start),
+                ),
+            );
 
         // Add return annotation
         let signature = match &function.return_annotation {
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -836,10 +836,9 @@ impl<'comments> Formatter<'comments> {
             )
             .append(
                 self.wrap_arguments(
-                    // We count end location of arguments here.
-                    // Before this change, wrap_arguments consumed all comments
-                    // leading in moving comments from return annotation to argument list
                     arguments,
+                    // Calculate end location of arguments to not consume comments in
+                    // return annotation
                     function
                         .return_annotation
                         .as_ref()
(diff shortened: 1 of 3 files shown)


ankddev (commented):
@lpil I marked this PR as draft since I have some questions.

Firstly, is it acceptable that comments in tuple return type move to the end of tuple, right before closing `)`? I found that basic tuple format as following:
```gleam
// Before formatting
#(
  // hello 0
  "hello" // hello 1
  // hello 2
)
// After formatting
#(
  // hello 0
  "hello",
  // hello 1
  // hello 2
)
```
I have no idea at how to place them better.

Secondly, I don't know why test is failed. There are no snapshots for this test generated. I checked other PRs with formatter changes and it seems like that code in `assert_format!()` must be equal to the result of formatting? Am I right?

lpil:
Yup that's the correct formatting! Let me know if you're ready for a review 🙏

ankddev:
@lpil Ready for review!

compiler-core/src/format.rs
```diff
@@ -834,7 +834,18 @@ impl<'comments> Formatter<'comments> {
                     .expect("Function in a statement must be named")
                     .1,
             )
-            .append(self.wrap_arguments(arguments, function.location.end));
+            .append(
+                self.wrap_arguments(
+                    // We count end location of arguments here.
+                    // Before this change, wrap_arguments consumed all comments
+                    // leading in moving comments from return annotation to argument list
```
lpil: Document the current code, not the code of the past!

What does "we count end location of arguments here" mean?

lpil (commented):
Thank you! Code looks super clean. One tiny note below.

compiler-core/src/format.rs
ankddev: @lpil Done

lpil (approved):
Super nice!! Thank you

PR #5019, opened 2025-09-30
feat(LS): `Annotate all top level definitions` code action

### LINKED ISSUE #4751: Add a code action to add type annotations to all functions in a module / project
Currently, you have to either add all type annotations in one fell swoop (this adds them for `let` assignments too), or go through every function independently (boring).

Maybe it should work on constants too, I have no opinion on this though because it's trivial to see their type anyway.

ankddev:
Closes #4751 
Add code action to add type annotation to all top level definitions.
### Showcase
```gleam
pub const answer = 42
pub fn add(x, y) {
  x + y
}
pub fn add_one(thing) {
  //     ^ Triggering "Annotate all top level definitions" code action here
  let result = add(thing, 1)
  result
}
```
would become:
```gleam
pub const answer: Int = 42
pub fn add(x: Int, y: Int) -> Int {
  x + y
}
pub fn add_one(thing: Int) -> Int {
  let result = add(thing, 1)
  result
}
```

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1367,7 +1367,7 @@ pub struct AnnotateTopLevelDefinitions<'a> {
     module: &'a Module,
     params: &'a CodeActionParams,
     edits: TextEdits<'a>,
-    is_hovering_definition: bool,
+    is_hovering_definition_requiring_annotations: bool,
 }
 
 impl<'a> AnnotateTopLevelDefinitions<'a> {
@@ -1380,7 +1380,7 @@ impl<'a> AnnotateTopLevelDefinitions<'a> {
             module,
             params,
             edits: TextEdits::new(line_numbers),
-            is_hovering_definition: false,
+            is_hovering_definition_requiring_annotations: false,
         }
     }
 
@@ -1389,7 +1389,7 @@ impl<'a> AnnotateTopLevelDefinitions<'a> {
 
         // We only want to trigger the action if we're over one of the definition
         // which is lacking some annotations in the module
-        if !self.is_hovering_definition || self.edits.edits.is_empty() {
+        if !self.is_hovering_definition_requiring_annotations {
             return vec![];
         };
 
@@ -1414,7 +1414,7 @@ impl<'ast> ast::visit::Visit<'ast> for AnnotateTopLevelDefinitions<'_> {
 
         // We're hovering definition which needs some annotations
         if overlaps(code_action_range, self.params.range) {
-            self.is_hovering_definition = true;
+            self.is_hovering_definition_requiring_annotations = true;
         }
 
         self.edits.insert(
@@ -1441,15 +1441,14 @@ impl<'ast> ast::visit::Visit<'ast> for AnnotateTopLevelDefinitions<'_> {
             return;
         }
 
+        let code_action_range = self.edits.src_span_to_lsp_range(fun.location);
+        if overlaps(code_action_range, self.params.range) {
+            self.is_hovering_definition_requiring_annotations = true;
+        }
+
         // Create new printer to ignore type variables from other definitions
         let mut printer = Printer::new_without_type_variables(&self.module.ast.names);
         collect_type_variables(&mut printer, fun);
-
-        let code_action_range = self.edits.src_span_to_lsp_range(fun.location);
-
-        if overlaps(code_action_range, self.params.range) {
-            self.is_hovering_definition = true;
-        }
 
         // Annotate each argument separately
         for argument in arguments_to_annotate {
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -11318,3 +11318,16 @@ fn wobble(other) { todo }
         find_position_of("wobble").to_selection()
     );
 }
+
+#[test]
+fn annotate_all_top_level_definitions_not_suggested_if_annotations_present() {
+    assert_no_code_actions!(
+        ANNOTATE_TOP_LEVEL_DEFINITIONS,
+        r#"
+fn wibble(one: Int) -> Int { one }
+
+fn wobble(one) { wibble(one) }
+"#,
+        find_position_of("wibble").to_selection()
+    );
+}


CHANGELOG.md
```diff
@@ -1,5 +1,44 @@
 # Changelog
 
+## Unreleased
+
+### Language server
+
+- The language server nof offers code action to add type annotations to all
```
```suggestion
- The language server now offers code action to add type annotations to all
```

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,179 @@ impl<'a> AddAnnotations<'a> {
...
+        self.visit_typed_module(&self.module.ast);
+
+        // We only want to trigger the action if we're over one of the definition in
+        // the module
+        if !self.is_hovering_definition {
+            return vec![];
+        };
+
+        let mut action = Vec::with_capacity(1);
+        CodeActionBuilder::new("Annotate all top level type definitions")
```
giacomocavalieri: I don't think the name is correct the code action is not annotating _type_ definitions, I'd call it "Annotate all top level definitions"

compiler-core/src/language_server/tests/action.rs
```diff
@@ -10554,3 +10555,39 @@ pub fn main() {
...
+  thing + 2
+}
+
+pub fn add_one(thing) {
+  thing + 1
+}
+"#,
+        find_position_of("const").select_until(find_position_of("="))
+    );
+}
```
giacomocavalieri, line 11227: I would add a couple more tests:
- A test that makes sure no code action is suggested if no top level definition is missing any annotation, even when hovering a definition
- A test with some top level definition that's only partially annotated to make sure we're adding the missing ones
- A test with multiple functions, each one with generic variables, to make sure they get the right names:
  ```gleam
  fn wibble(one) { todo } // expect one: a -> b

  fn wobble(other) { todo } // expect other: a -> b
  ```
- A test that makes sure no code action is suggested if we're inside a function's body and not hovering its definition (at least I think that's the behaviour we want, @lpil what do you think?)

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,179 @@ impl<'a> AddAnnotations<'a> {
...
+        // Annotate the return type if it isn't already annotated
+        if fun.return_annotation.is_none() {
+            self.edits.insert(
+                fun.location.end,
+                format!(" -> {}", self.printer.print_type(&fun.return_type)),
+            );
+        }
+    }
+
+    fn visit_typed_expr_fn(
```
giacomocavalieri: Why is this going over `expr_fn`s? Those are not top level definitions that should be annotated by this action

giacomocavalieri (commented):
Thank you, that's a great start! I left a couple of comments and questions inline

compiler-core/src/language_server/tests/action.rs
ankddev: @lpil @giacomocavalieri I have no idea why with two generic function it leads to following result if code both for "Add annotation" and this new action is identical in this part:
```
Expression:
fn wibble(one) { todo }

fn wobble(other) { todo }

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
+new results
────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
          0 │+----- BEFORE ACTION
          1 │+
          2 │+fn wibble(one) { todo }
          3 │+
          4 │+fn wobble(other) { todo }
          5 │+   ↑
          6 │+
          7 │+
          8 │+----- AFTER ACTION
          9 │+
         10 │+fn wibble(one: a) -> b { todo }
         11 │+
         12 │+fn wobble(other: c) -> d { todo }
────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```
So, I commented it out for now and drafted PR.

giacomocavalieri: Ti emit sounds like the type printer is not properly reset in between functions. I'm not super knowledgeable on this but it might be enough to reset it after visiting each function, that might fix this! Let me know if it works, if it doesn't I can try and give this a proper more in depth look :)

ankddev: @giacomocavalieri I played with it and found that removing global printer inside struct and just creating new printer on each visiting fixes issue.

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,115 @@ impl<'a> AddAnnotations<'a> {
...
+            ),
+        );
+    }
+
+    fn visit_typed_function(&mut self, fun: &'ast ast::TypedFunction) {
+        // Create new printer to ignore type variables from other definitions
+        let mut printer = Printer::new_without_type_variables(&self.module.ast.names);
+        collect_type_variables(&mut printer, fun);
+
+        ast::visit::visit_typed_function(self, fun);
```
giacomocavalieri: Why is this visiting the typed function? This will traverse its entire AST but that's not needed!
```suggestion
```

compiler-core/src/language_server/tests/action.rs
```diff
@@ -10554,3 +10555,119 @@ pub fn main() {
...
+
+#[test]
+fn annotate_all_top_level_definitions_already_annotated() {
+    assert_no_code_actions!(
+        ANNOTATE_TOP_LEVEL_DEFINITIONS,
+        r#"
+pub const answer: Int = 42
+
+pub fn add_two(thing: Int) -> {
```
giacomocavalieri: This test has a syntax error, so the action will not appear no matter what!
```suggestion
pub fn add_two(thing: Int) -> Int {
```

compiler-core/src/language_server/code_action.rs
ankddev: Why then "Add annotations" code action visiting it? Should we change this too (in another PR)?

ankddev:
Hmm, interesting. Why it shows code action for annotated functions? It worked yesterday (I tested it in Zed and Neovim), but it doesn't now.
UPD: I found it

giacomocavalieri:
Great, this looks good to me!! It only needs Louis' approval now :)

ankddev:
Ah, CI was failed because of network issues

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,113 @@ impl<'a> AddAnnotations<'a> {
...
+            is_hovering_definition: false,
+        }
+    }
+
+    pub fn code_actions(mut self) -> Vec<CodeAction> {
+        self.visit_typed_module(&self.module.ast);
+
+        // We only want to trigger the action if we're over one of the definition in
+        // the module
+        if !self.is_hovering_definition || self.edits.edits.is_empty() {
```
lpil: Can we only trigger if the definition being hovered is lacking some annotations please 🙏

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,113 @@ impl<'a> AddAnnotations<'a> {
...
+                })
+                .unwrap_or(fun.location),
+        );
+
+        if overlaps(code_action_range, self.params.range) {
+            self.is_hovering_definition = true;
+        }
+
+        // Annotate each argument separately
+        for argument in fun.arguments.iter() {
```
lpil: Can we collect the missing argument and return annotations first please, so that we can check if a new printer is needed at all prior to constructing it

compiler-core/src/language_server/code_action.rs
```diff
@@ -1360,6 +1360,113 @@ impl<'a> AddAnnotations<'a> {
...
+        let mut printer = Printer::new_without_type_variables(&self.module.ast.names);
+        collect_type_variables(&mut printer, fun);
+
+        let code_action_range = self.edits.src_span_to_lsp_range(
+            fun.body_start
+                .map(|body_start| SrcSpan {
+                    start: fun.location.start,
+                    end: body_start,
+                })
+                .unwrap_or(fun.location),
```
lpil: Just check the head location, not the body please.

compiler-core/src/language_server/code_action.rs
ankddev: Due to refactor I've made it currently firstly checks if there are some missing annotation and returns if not and only after that checks for overlapping, so should be fixed.

ankddev:
@lpil Done!

PR #5028, opened 2025-10-01
Fix incorrect missing patterns

### LINKED ISSUE #4278: Incorrect missing patterns given for `case` expression

abs0luty:
Fixes [#4278](https://github.com/gleam-lang/gleam/issues/4278).

Running this code:

```gleam
pub fn wibble(b: Bool, i: Int) {
  case b, i {
    False, 1 -> todo
    True, 2 -> todo
  }
}
```

Produces this error:

```
The missing patterns are:

    False, _
    True, _
    _, _
```

Ideally, one would expect either `_, _` or `False, _` and `True, _`.

The root cause is in the pivot choice for building the decision tree. When we pivot on an infinite subject (like Int) before a finite one (like `Bool`), the "fallback" branch for the infinite split collapses to a textual `_`, which then overlaps with the more specific missing patterns discovered under the finite splits. 

The clean fix is to make the compiler prefer finite-domain subjects first when choosing the pivot. It shrinks the search space and avoids noisy fallbacks. 

So the solution is to teach `find_pivot_check` to prefer finite-branching subjects (tuples, lists, and named types with known constructors) over infinite ones (e.g. ints, floats, strings).

```gleam
pub fn wibble(b: Bool, i: Int) {
  case b, i {
    False, 1 -> todo
    True, 2 -> todo
  }
}
```

This makes `Bool` beat `Int` in the example, so the top split is on b, eliminating the top-level `_, _` catch-all that comes from an infinite fallback:

```
error: Inexhaustive patterns
  ┌─ /Users/adi.salimgereyev/hello_world/src/hello_world.gleam:6:3
  │  
6 │ ╭   case b, i {
7 │ │     False, 1 -> 1
8 │ │     True, 2 -> 2
9 │ │   }
  │ ╰───^

The missing patterns are:

    False, _
    True, _
```

More about creating effective decision trees can be found [here](https://www.cs.tufts.edu/~nr/cs257/archive/luc-maranget/jun08.pdf).

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -787,6 +787,20 @@ enum BranchMode {
 }
 
 impl BranchMode {
+    /// Returns a heuristic estimate of the branching factor.
+    ///
+    /// This value is used by the pivot-selection to prefer splits
+    /// with fewer branches, which tends to produce smaller and shallower
+    /// decision trees.
+    fn branching_factor(&self) -> usize {
+        match self {
+            BranchMode::Infinite => usize::MAX,
+            BranchMode::Tuple { elements } => elements.len(),
+            BranchMode::List { .. } => 2,
+            BranchMode::NamedType { constructors, .. } => constructors.len(),
+        }
+    }
+
     fn is_infinite(&self) -> bool {
         match self {
             BranchMode::Infinite => true,
@@ -2250,7 +2264,7 @@ impl<'a> Compiler<'a> {
         // the first branch, and use the variable it's pattern matching on to create
         // a new node in the tree. All the branches will be split into different
         // possible paths of this tree.
-        match find_pivot_check(first_branch, &branches) {
+        match find_pivot_check(first_branch, &branches, self.environment) {
             Some(PatternCheck { var, .. }) => self.split_and_compile_with_pivot_var(var, branches),
 
             // If the branch has no remaining checks, it means that we've moved all
@@ -2713,7 +2727,11 @@ impl<'a> Compiler<'a> {
 /// Returns a pattern check from `first_branch` to be used as a pivot to split all
 /// the `branches`.
 ///
-fn find_pivot_check(first_branch: &Branch, branches: &VecDeque<Branch>) -> Option<PatternCheck> {
+fn find_pivot_check(
+    first_branch: &Branch,
+    branches: &VecDeque<Branch>,
+    env: &Environment<'_>,
+) -> Option<PatternCheck> {
     // To try and minimise code duplication, we use the following heuristic: we
     // choose the check matching on the variable that is referenced the most
     // across all checks in all branches.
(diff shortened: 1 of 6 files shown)


compiler-core/src/exhaustiveness.rs
```diff
@@ -2713,24 +2713,46 @@ impl<'a> Compiler<'a> {
...
+    env: &Environment<'_>,
+) -> Option<PatternCheck> {
+    // Count how often each variable is referenced across branches (old heuristic).
     let mut var_references = HashMap::new();
     for branch in branches {
         for check in &branch.checks {
             let _ = var_references
                 .entry(check.var.id)
-                .and_modify(|references| *references += 1)
+                .and_modify(|r| *r += 1)
```
giacomocavalieri: We prefer using full names instead of shorthands in the compiler, could you revert this?

compiler-core/src/exhaustiveness.rs
```diff
@@ -2713,24 +2713,46 @@ impl<'a> Compiler<'a> {
...
         .iter()
-        .max_by_key(|check| var_references.get(&check.var.id).cloned().unwrap_or(0))
+        .max_by_key(|check| {
+            let mode = check.var.branch_mode(env);
+            let is_finite = !mode.is_infinite();
+            let size = domain_size_hint(&mode);
+            let refs = var_references.get(&check.var.id).cloned().unwrap_or(0);
+
+            // Note: use usize::MAX - size to prefer smaller finite domains.
+            (is_finite, usize::MAX - size, refs)
```
giacomocavalieri: Could you add a comment explaining why the tuple is structured this way? It's a bit hard to understand at first glance why the three items are ordered the way they are

compiler-core/src/exhaustiveness.rs
```diff
@@ -2713,24 +2713,46 @@ impl<'a> Compiler<'a> {
...
+
     first_branch
         .checks
         .iter()
-        .max_by_key(|check| var_references.get(&check.var.id).cloned().unwrap_or(0))
+        .max_by_key(|check| {
+            let mode = check.var.branch_mode(env);
+            let is_finite = !mode.is_infinite();
+            let size = domain_size_hint(&mode);
+            let refs = var_references.get(&check.var.id).cloned().unwrap_or(0);
```
giacomocavalieri: No shorthands here as well `references` instead of `refs`

compiler-core/src/exhaustiveness.rs
```diff
@@ -2713,24 +2713,46 @@ impl<'a> Compiler<'a> {
...
-                .and_modify(|references| *references += 1)
+                .and_modify(|r| *r += 1)
                 .or_insert(0);
         }
     }
 
+    // Score each candidate in the first branch with a tuple:
+    // (is_finite, -domain_size_hint, references)
+    // Higher is better. domain_size_hint is "how many choices" the split makes.
+    fn domain_size_hint(mode: &BranchMode) -> usize {
```
giacomocavalieri: Can we give this a name that's a bit easier to understand? Something like `number_of_choices` feels more descriptive of what this is doing.

And instead of defining it here I'd make it a method of `BranchMode`:
```rs
impl BranchMode {
  fn number_of_choices(&self) -> usize {
    // ...
  }
}
```

lpil:
Brilliant, thank you! This PR is fantastic and very impressive!

I agree with Jak's feedback 🙏 

Also v interested in the literature you mentioned. We try to have this information in the code comments for complex bits of the codebase, so more documentation would be fab!

compiler-core/src/exhaustiveness.rs
abs0luty: Sorry didn't notice that!

compiler-core/src/exhaustiveness.rs
abs0luty: Sure!

compiler-core/src/exhaustiveness.rs
abs0luty: Sorry, fixed that!

compiler-core/src/exhaustiveness.rs
abs0luty: I'll just call it `BranchMode::branching_factor()`.

compiler-core/src/exhaustiveness.rs
abs0luty: By the way, I think `is_finite, -size` can be reduced to just `-size`, because we can set `size = usize::MAX` for infinite subjects.

giacomocavalieri (approved):
This is amazing, thank you!

lpil:
I'll update the changelog for you 🙏

Issue #5020, opened 2025-09-30
Extract function code action loses code when there is a `use`  inside


kwando:
<img width="408" height="208" alt="Image" src="https://github.com/user-attachments/assets/3a250c79-3304-48a2-a1a0-fc95e72104eb" />
<img width="372" height="211" alt="Image" src="https://github.com/user-attachments/assets/1fb2b1ce-70f5-415e-9d27-0c7a885b5bad" />

And if you do not mark the block we get some extra code, see `Nil` in the extracted function. But maybe this is what you have to do if you are not selecting _all_ the lines in the block.

<img width="411" height="210" alt="Image" src="https://github.com/user-attachments/assets/5e672dfc-1591-46f2-9cce-9137d7070bfc" />
<img width="370" height="295" alt="Image" src="https://github.com/user-attachments/assets/83f1ccfd-5f6c-4477-aec7-c049721e2f90" />

giacomocavalieri:
Oops! I'm working on a fix

lpil:
Oh no! Thanks for the report

kwando:
This is what RCs are for :)

ankddev:
I think this issue can be closed

lpil:
Thank you!

PR #5029, opened 2025-10-01
Fix "Extract function" code action working on `use` expressions

giacomocavalieri:
This PR fixes two bugs with the extract function code action:
- #5020 where it would delete the body of the use statement being extracted
- and another one I found where it wouldn't properly type an extracted use statement


giacomocavalieri:
Yeah I was in a rush to get out and forgot to review the snapshots, will do that once I'm home 😁

compiler-core/src/language_server/code_action.rs
```diff
@@ -8623,10 +8623,16 @@ impl<'a> ExtractFunction<'a> {
...
-                );
+                let location = statements
+                    .first()
+                    .location()
+                    .merge(&match &statements.last() {
+                        ast::Statement::Expression(expression) => expression.location(),
+                        ast::Statement::Assignment(assignment) => assignment.location,
+                        ast::Statement::Assert(assert) => assert.location,
+                        ast::Statement::Use(use_) => use_.call.location(),
+                    });
```
lpil: This code is a bit cryptic, and it is in 3 places in this PR. Could we extract it to a new documented helper method please? 🙏

lpil (commented):
Thank you! One tiny thing!

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Yeah I was thinking about it but couldn't come up with a good name, what do you think it should be called? `entire_location`?

giacomocavalieri: Ended up moving that to the location function, I think that's what makes the most sense for use statements

lpil (approved):
Thank you!

PR #5030, opened 2025-10-01
✨Use parsed float value for useless comparison warnings

### LINKED ISSUE #5001: Incorrect warn on useless comparisons
hello!
I've started lerning gleam on https://tour.gleam.run/basics/number-formats/
And I faced with inconsistency of echo result and warning when I wrote:
```
pub fn main() {
  // Scientific notation Float literals
  echo  7.0e2 == 700.0
}
```
the output is:
```
src/main.gleam:3
True
warning: Redundant comparison
  ┌─ /src/main.gleam:3:9
  │
3 │   echo  7.0e2 == 700.0
  │         ^^^^^^^^^^^^^^ This is always `False`

This comparison is redundant since it always fails.
```
Why does warning say 'false', if the echo prints 'true'?

fruno-bulax:
Closes #5001

Parallel to the implementation for ints, we're now parsing the literal value and passing it along the AST.
This allows us to do smarter useless comparison warnings instead of relying on string comparison.

To maintain the `Eq` trait impl on many types, I created a `NotNan` type to wrap f64 that be `NaN` and can therefore properly implement `Eq` and `Ord`. I wasn't quite sure where to put the type, I'll gladly move it.

I noticed that we still do string-based checks in clauses for both ints and floats, I'll create a follow-up issue after this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,3 +87,7 @@
 - Fixed a bug where the "pattern match on variable" code action would generate
   invalid patterns by repeating a variable name already used in the same pattern.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where useless comparison warnings for floats compared literal
+  strings, claiming for example that `1.0 == 1.` was always false.
+  ([fruno](https://github.com/fruno-bulax/))


compiler-core/src/parse/snapshots/gleam_core__parse__tests__nested_block.snap
```diff
@@ -34,6 +34,9 @@ expression: "{ 1 { 1.0 2.0 } 3 }"
                                         end: 9,
                                     },
                                     value: "1.0",
+                                    float_value: NotNan(
+                                        1.0,
+                                    ),
```
giacomocavalieri: Isn't this the `ONE` constant you defined earlier?
```suggestion
                                    float_value: NotNan::ONE,
```

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__float_literals_redundant_comparison_different_repr.snap
```diff
@@ -0,0 +1,15 @@
...
+---
+----- SOURCE CODE
+pub fn main() { 1_0.0 == 10.0 }
+
+----- WARNING
+warning: Redundant comparison
+  ┌─ /src/warning/wrn.gleam:1:17
+  │
+1 │ pub fn main() { 1_0.0 == 10.0 }
+  │                 ^^^^^^^^^^^^^ This is always `True`
```
giacomocavalieri, line 13: Really nice!!

compiler-core/src/parse.rs
```diff
@@ -4696,3 +4699,44 @@ impl PatternPosition {
         }
     }
 }
+
+/// A thin f64 wrapper that does not permit NaN.
+/// This allows us to implement `Eq`, which require reflexivity.
+///
+/// Used for gleam float literals. May contain infinity.
```
giacomocavalieri: If this is used for Gleam float literals it shouldn't contain infinity as well, should it?

giacomocavalieri (approved):
This is great, well done!! I left just a tiny nit inline

compiler-core/src/parse/snapshots/gleam_core__parse__tests__nested_block.snap
fruno-bulax: This is the snapshot of the parser test, not rust code! 🕵️

giacomocavalieri: Oh my, sorry for that, I totally missed the context 😆

compiler-core/src/parse.rs
fruno-bulax: This is true for the Erlang target (and there is a compile time check for it), but on JS you can enter float literals that are too large (e.g. `1.0e500`) and they turn into (negative) infinity.

giacomocavalieri: Aha, now I see, thanks for explaining! Would you mind adding this explanation to the comment? That would be really helpful for future people wondering the same thing

fruno-bulax: Good point, done!

compiler-core/src/parse.rs
```diff
@@ -4696,3 +4699,46 @@ impl PatternPosition {
...
 }
+
+/// A thin f64 wrapper that does not permit NaN.
+/// This allows us to implement `Eq`, which require reflexivity.
+///
+/// Used for gleam float literals, which cannot be NaN.
+/// While there is no syntax for "infinity", float literals
+/// may overflow into (possibly negative) infinity on the JS target.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct NotNan(f64);
```
lpil: This is a rather confusing name to see in the codebase. Can we call it something like `FloatValue` or `LiteralFloatValue` instead 🙏

CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Patterns aliasing a string prefix have been optimised to generate faster code
   on the Erlang target.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+- Useless comparison warnings for floats now consider the parsed value.
```
lpil: Space between changelog entries please

This entry describes the compiler internals, but it doesn't tell the Gleam programmer what has changed from their point of view, which is what this document is for. Could we have that here please 🙏

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -4129,6 +4129,71 @@ fn int_literals_redundant_comparison_8() {
...
+
+#[test]
+fn float_literals_redundant_comparison_infinity() {
+    assert_warning!("pub fn main() { 1.0e500 == 1.0e600 }");
+}
+
+#[test]
+fn float_literals_redundant_comparison_signed_zero() {
+    assert_warning!("pub fn main() { 0.0 == -0.0 }");
+}
```
lpil, line 4195: Can we have test for floats like `10.` please

lpil (commented):
Fab work! This is really nice!

I've left some small notes inline 🙏

compiler-core/src/parse.rs
fruno-bulax: Naming, my eternal nemesis. I went for `LiteralFloatValue`, since that more narrowly describes what it is.

CHANGELOG.md
fruno-bulax: Reworded and moved to bug fixes where it actually belongs!

compiler-core/src/type_/tests/warnings.rs
fruno-bulax: Added!

lpil (approved):
Thank you!!! Really nice work

Issue #4931, opened 2025-09-03
Confusing error when a record field does not typecheck


yoshi-monster:
When a record field definition refers to an unknown type, the errors you get when using that record always claim that record field doesn't exist, including claiming the record has one less field:

```gleam
pub type Wibble {
  Wibble(x: Int, f: fn(Wobble) -> Int)
}

pub fn wibble() {
  Wibble(1, fn(_) { 2 })
}

pub fn wobble(wibble: Wibble) {
  wibble.f
}

pub fn woo(wibble: Wibble) {
  Wibble(..wibble, x: 1)
}
```

results in 

```
│
2 │   Wibble(x: Int, f: fn(Wobble) -> Int)
  │                        ^^^^^^ Did you mean `Wibble`?

The type `Wobble` is not defined or imported in this module.

error: Incorrect arity
  ┌─ /src/main.gleam:6:3
  │
6 │   Wibble(1, fn(_) { 2 })
  │   ^^^^^^^^^^^^^^^^^^^^^^ Expected 1 argument, got 2


error: Unknown record field
   ┌─ /src/main.gleam:10:10
   │
10 │   wibble.f
   │          ^ Did you mean `x`?

The value being accessed has this type:

    Wibble

It has these accessible fields:

    .x

warning: Redundant record update
   ┌─ /src/main.gleam:14:3
   │
14 │   Wibble(..wibble, x: 1)
   │   ^^^^^^^^^^^^^^^^^^^^^^ This record update specifies all fields

Hint: It is better style to use the record creation syntax.
```

I think all of those are confusing since the record does have that field.

~ 💜

lpil:
Agreed, this is confusing. Let's preserve the arity of the constructor.

PR #5033, opened 2025-10-02
Recover from errors in field definitions of custom type variants

abs0luty:
- Closes https://github.com/gleam-lang/gleam/issues/4931

Ideally, compiler should not ignore a record field where type checking failed. So a solution is to report the error, but put unbound type variable, instead of skipping the field. This practice is already used in other parts of analyzer. For instance:

https://github.com/gleam-lang/gleam/blob/03bfad9281813684132d00a3152f6c3e6bcc807d/compiler-core/src/analyse.rs#L1529-L1554


lpil (approved):
Lovely, thank you!!

PR #5034, opened 2025-10-02
Make pattern matching to favour first appearing variable

giacomocavalieri:
Looking at the recent change to pattern matching I wanted to try and change it so that in case of ties (both in branching factor and references) the first variable is favoured over following ones. In this case:
```gleam
case wibble, wobble {
  True, False -> todo
}
```
The missing patterns reported are:
```
False, _
True, True
```
Instead of
```
False, False
_, True
```
In general the first variable in the match will have all the patterns enumerated while the following ones will have an `_` instead of it being the other way around. I think that looks a little less foreign but no big deal if this isn't merged.

There's also a small refactoring to avoid doing `usize::MAX` tricks and use a custom type instead.

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -2748,27 +2748,29 @@ fn find_pivot_check(
         }
     }
 
-    // Maximize ( -branching_factor, references ).
+    // We want to picke the variable that has the smallest branching factor
+    // possible, this tends to yield smaller, shallower decision trees.
+    // So, for example, we will pick a list (branching factor of 2 `[]` and
+    // `[_, ..]`) over an integer (infinitely many choices).
     //
-    // - branching_factor: prefer the split that creates the FEWEST switch arms.
-    //   Fewer children tends to yield smaller, shallower decision trees.
-    //
-    // - references: break ties by choosing the subject that appears in the most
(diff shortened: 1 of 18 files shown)


abs0luty:
👏 Beautiful

compiler-core/src/exhaustiveness.rs
```diff
@@ -814,6 +816,29 @@ impl BranchMode {
...
+impl Ord for BranchingFactor {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self, other) {
+            (BranchingFactor::Infinite, BranchingFactor::Infinite) => std::cmp::Ordering::Equal,
+            (BranchingFactor::Infinite, BranchingFactor::Finite(_)) => std::cmp::Ordering::Greater,
+            (BranchingFactor::Finite(_), BranchingFactor::Infinite) => std::cmp::Ordering::Less,
+            (BranchingFactor::Finite(one), BranchingFactor::Finite(other)) => one.cmp(other),
+        }
+    }
+}
```
lpil: This is quite a bit harder for me to understand now, and it means we have to use traits. Please revert to using numbers. 🙏

giacomocavalieri: Done!

PR #5035, opened 2025-10-02
Make sure the "turn into pipe" code action is suggested in final step of a pipe

giacomocavalieri:
I noticed a small bug where the LS wouldn't suggest the "turn into pipe" code action when triggered on a value inside the final step of a pipeline:
```gleam
wibble
|> wobble(woo(1))
//        ^^^^^^ Wouldn't trigger on this!!
```


lpil (approved):
Thank you!!

PR #5037, opened 2025-10-02
Use correct line number information for generating functions in other modules

### LINKED ISSUE #5036: Generate function in different module drops the new function in a random place
This is on `v1.13.0-rc1`. I expect the function to be generated at the end of the file, but I've seen it breaking existing modules by generating it in the middle of existing functions.

GearsDatapacks:
Fixes #5036

I also fixed another bug I found along the way where it would generate the function one character too early if the module did not end in a newline.

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -5208,11 +5208,7 @@ impl<'a> GenerateFunction<'a> {
 
         if let Some(module) = module {
             if let Some(module) = self.modules.get(module) {
-                let insert_at = if module.code.is_empty() {
-                    0
-                } else {
-                    (module.code.len() - 1) as u32
-                };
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -5272,6 +5268,10 @@ impl<'a> GenerateFunction<'a> {
 
         let publicity = if publicity.is_public() { "pub " } else { "" };
 
+        // Make sure we use the line number information of the module we are
+        // editing, which might not be the module where the code action is
+        // triggered.
+        self.edits.line_numbers = &module.ast.type_info.line_numbers;
```
giacomocavalieri, line 5274: Sneaky! Nice catch!!

lpil (approved):
Thank you!

PR #5038, opened 2025-10-02
Improve help text for run, test, and dev commands

giacomocavalieri:
This PR improves the help text for the `run`, `dev`, and `test` command by saying which function they're going to run in the help text. This is something people usually ignore, especially for `gleam test`, thinking it's doing more than it actually does. So maybe this could help clear some confusion (given someone will read the help text 😆)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@
 
 ### Build tool
 
+- The help text displayed by `gleam dev --help`, `gleam test --help`, and
+  `gleam run --help` has been improved: now each one states which function it's
+  going to run.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Language server
 
 - The "inline variable" code action can now trigger when used over the let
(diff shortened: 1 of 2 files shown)


lpil (commented):
I like the idea, but I'm not a huge fan of the syntax used here, and the line slightly long.

What does `gleam help` look like with this?

giacomocavalieri:
`gleam help` hasn't changed: 
Screenshot 2025-10-03 alle 13 31 17
While `gleam test/dev/main --help` looks like this:
Screenshot 2025-10-03 alle 13 31 30

lpil:
Maybe we should specify the module name instead of the file, since it doesn't have to be at that path.

giacomocavalieri:
I had no idea this was the case, so `gleam dev` runs the `main` function from the`<PROJECT_NAME>_dev` module regardless of where the module is? Does that work even if the module is under `src/` or `test/`?

GearsDatapacks:
That's correct. No distinction is made there

giacomocavalieri:
I see, so should we only suggest the module name? @lpil how would you prefer that information be presented, is `<PACKAGE_NAME>_dev` ok?

lpil:
Sounds good to me!

PR #5043, opened 2025-10-04
Fix crash in "pattern match on variable code action"

### LINKED ISSUE #5042: Language server crash when triggering "Pattern match on vairable" code action
I noticed a bug where the language server sometimes triggers when trying to trigger a code action. The minimal example I can find is the following:
```gleam
pub type Wibble {
  Wibble(Wobble)
}

pub type Wobble {
  Wobble
  Wubble
}

pub fn main() {
  let Wibble(wobble) = todo
  //         ^ Trigger code action here

  case todo {
    _ -> todo
  }
}
```
(Here the `case` expression is important for some unknown reason, and without it the bug does not occur)

Version: Latest `main` 195c44abaf4faca7edf2ca2bccfbce90c4be115f

Error message:
```
error: Fatal compiler bug!

Panic: /home/gears/projects/gleam/gleam/compiler-core/src/language_server/code_action.rs:4691
	assertion failed: self.is_char_boundary(n)
Gleam version: 1.13.0-rc1
Operating system: linux

```

This is a bug with the "Pattern match on variable code action", as indicated by the line of the crash: https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/language_server/code_action.rs#L4691

giacomocavalieri:
This PR closes #5042

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -7539,6 +7539,33 @@ fn map(list: List(a), fun: fn(a) -> b) { todo }
 }
 
 #[test]
+// https://github.com/gleam-lang/gleam/issues/5042
+fn pattern_match_on_variable_crashes() {
+    assert_code_action!(
+        PATTERN_MATCH_ON_VARIABLE,
+        r#"
+pub type Wibble {
+    Wibble(Wobble)
+}
+
+pub type Wobble {
+    Wobble
+    Wubble
+}
+
+pub fn main() {
+    let Wibble(wobble) = todo
+
+    case todo {
+      _ -> todo
+    }
+}
+"#,
+        find_position_of("wobble").to_selection()
+    );
+}
+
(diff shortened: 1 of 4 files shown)


compiler-core/src/language_server/tests/action.rs
```diff
@@ -7538,6 +7538,34 @@ fn map(list: List(a), fun: fn(a) -> b) { todo }
...
+}
+
+pub type Wobble {
+    Wobble
+    Wubble
+}
+
+pub fn main() {
+    let Wibble(wobble) = todo
+    //         ^ Trigger code action here
```
GearsDatapacks: Probably don't need this comment as the snapshot formatter already tells you where the code action is being triggered

giacomocavalieri: Right, I'll remove that!

lpil (approved):
Thank you!

PR #5044, opened 2025-10-07
Optimise string pattern matching on the Erlang target

### LINKED ISSUE #5039: Pattern match on string prefix can be slow on erlang
When pattern matching on a string prefix and capturing the matched value:
```gleam
"a" as letter <> rest -> todo 
```
On the Erlang target, this is compiled to:
```erl
<<J:1/binary, Tail/binary>> when J =:= <<"a"/utf8>> -> todo
```
Meaning the Erlang compiler can't easily optimise the pattern matching yielding code that can be up to 24x times slower compared to a regular `"a" <> rest -> todo` (which instead is compiler to the fast `<<"a"/binary, Tail/binary>>`)

It would be great for the compiler to change the generated code so that capturing the constant prefix can be optimised as well.

giacomocavalieri:
This PR closes #5039

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Compiler
 
+- Patterns aliasing a string prefix have been optimised to generate faster code
+  on the Erlang target.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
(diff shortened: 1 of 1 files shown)


lpil (approved):
Amazing!!! Thank you. The code generated looks much nicer now

PR #5046, opened 2025-10-07
feat: only consider local type variables in error's type printing

### LINKED ISSUE #3786: Error's type printing: Consider only annotated type variables for the current definition rather than the whole module

ankddev:
This PR fixes #3786.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,57 @@
   the message is clearer.
   ([vyacheslavhere](https://github.com/vyacheslavhere))
 
+- Now build tool ignores type variables defined in other functions and types,
+  improving hints in errors. For example:
+
+  ```gleam
+  pub type Wibble(a, b, c, d, e, f, g) {
+    Wibble(value: a)
+  }
+
+  pub fn main() {
+    let x: Nil = []
+  }
+  ```
+
+  Previously this code would have emit following error:
+
+  ```text
+  error: Type mismatch
+    ┌─ C:\Users\user\projects\test_gleam\src\test_gleam.gleam:6:16
+    │
+  6 │   let x: Nil = []
+    │                ^^
+
+  Expected type:
+
+      Nil
+
+  Found type:
+
+      List(h)
+  ```
+
+  Now it will emit following error:
+
+  ```text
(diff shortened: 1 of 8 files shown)


GearsDatapacks (commented):
Hello! Could you add some tests to this please? I would like tests for:
- Ignoring type variables in other functions
- Correctly using type variables defined in the current function
- Correctly generating a new name to avoid clashing with type variables in the current function

ankddev:
@GearsDatapacks Okay, can you please give me some guidance about where they should be located? :)

GearsDatapacks:
They should go in the `type_::tests::errors` module

ankddev:
There are some problems with this, I'll address them later, so drafting PR for now!

lpil:

ankddev:
@lpil I would like to work on this, but I'm a bit stuck with it (see https://github.com/gleam-lang/gleam/issues/3786#issuecomment-3696188506).

PR #5049, opened 2025-10-10
Make gleam deps tree -p and -i mutually exclusive

silby:
`--invert` is currently silently ignored when `--package` is given also. I believe it's more predictable and more unixy for this to be a usage error instead.

Other tweaks to this are conceivable such as making `-i` a flag that doesn't take an arg instead of conflicting with `-p`. (This would raise the question of what you get when you do `gleam deps tree -i` by itself, which I guess would be what you get if you give the name of the current package to `-i` today, which is just your own package's name and version, since nothing in the project depends on it.) But as that is a more breaking change to the option usage I don't propose that here.

You can also decide that silently-ignoring an option is fine and close this; this took five minutes so no harm done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   on the Erlang target.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
-- Analysis it now fault tolerant in the presence of errors in field definitions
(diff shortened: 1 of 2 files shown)


silby:
repushed w/ changelog entry

PR #5053, opened 2025-10-18
Do not offer generate function code action for unsupported targets

### LINKED ISSUE #4904: "Generate function" code action is offered for function unsupported by current target
If you have some code which references a function that isn't defined for the current target, for example if the target were Erlang:
```gleam
pub fn main() {
  wibble()
}

@external(javascript, "./ffi.mjs", "wibble")
fn wibble() -> Nil
```

Hovering over the call to `wibble` offers the "Generate function" code action, which is incorrect. The function is not missing, it just doesn't support the current target.

GearsDatapacks:
Fixes #4904 

I've reused the `extra_information` field introduced recently as that seemed to make the most sense here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## Unreleased
 
+### Bug fixes
+
+- Fixed a bug where the "Generate function" code action would be incorrectly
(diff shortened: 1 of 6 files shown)


lpil (approved):
Very nice! Thank you very much

PR #5054, opened 2025-10-19
Move "Did you want a Bool" error to analysis step

### LINKED ISSUE #5052: Move the "Did you want a Bool" error to the analysis step
In Gleam a common beginner mistake was to use `true` and `false` instead of `True` and `False` in pattern matching:
```gleam
case is_admin(user) {
  // vv This should be `True`
  true -> admin_dashboard()
  // vv This should be `False`
  false -> page_404()
}
```
So using `true` and `false` in a case expression is a syntax error with a helpful error message:
```
error: Syntax error
  ┌─ /src/main.gleam:5:5
  │
5 │     true -> 1
  │     ^^^^ Did you want a Bool instead of a variable?

See: https://tour.gleam.run/basics/bools/
Hint: In Gleam boolean literals are `True` and `False`.
```
The problem with this being a syntax error is that it stops analysis from happening in the entire module. It would be best if this error was emitted during the analysis phase instead. This requires a couple of changes:
- Remove the `ParseErrorType::LowcaseBooleanPattern` variant and make it a variant of `Error` (in the `compiler_core/src/type_/error.rs` file)
- Once you've done that you should be able to follow the compiler errors to remove the piece of code that's emitting that error in the parsing step
- That same logic that emits the error should be moved to the piece of the compiler that's analysing variable patterns (that will most likely be the `unify` function in the `compiler_core/src/type_/pattern.rs` file, the branch where this should happen is probably the one matching on the `Pattern::Variable` variant)
- And don't forget to add tests in `compiler_core/src/type_/tests.rs` to make sure the error is actually emitted. It should be pretty straightforward to copy one of the other tests using `assert_module_error!` and adapt it with a code snippet producing this error instead

There's other changes that we could make, like being smarter about where this error happens. But that can be a separate issue to keep this nice and approachable for newcomers :)

mxtthias:
Fixes #5052

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,11 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The lowercase bool pattern error is no longer a syntax error, but instead a
+  part of the analysis step. This allows the entire module to be analyzed, rather
+  than stopping at the syntax error.
+  ([mxtthias](https://github.com/mxtthias))
+
 ### Build tool
 
 - The help text displayed by `gleam dev --help`, `gleam test --help`, and
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -4025,6 +4025,22 @@ with no constructors."
                             extra_labels: vec![],
                         }),
                     },
+
+                    TypeError::LowercaseBoolPattern { location } => Diagnostic {
+                        title: "Lowercase bool pattern".to_string(),
+                        text: "See: https://tour.gleam.run/basics/bools/".into(),
+                        hint: Some("In Gleam bool literals are `True` and `False`.".into()),
+                        level: Level::Error,
+                        location: Some(Location {
+                            label: Label {
+                                text: Some("This is not a bool".into()),
+                                span: *location,
+                            },
+                            path: path.clone(),
+                            src: src.clone(),
+                            extra_labels: vec![],
+                        }),
+                    },
                 })
                 .collect_vec(),
 
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -671,6 +671,10 @@ pub enum Error {
     ExternalTypeWithConstructors {
         location: SrcSpan,
     },
+
+    LowercaseBoolPattern {
+        location: SrcSpan,
+    },
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -1309,7 +1313,8 @@ impl Error {
             | Error::NonUtf8StringAssignmentInBitArray { location }
             | Error::PrivateOpaqueType { location }
             | Error::SrcImportingDevDependency { location, .. }
-            | Error::ExternalTypeWithConstructors { location, .. } => location.start,
+            | Error::ExternalTypeWithConstructors { location, .. }
+            | Error::LowercaseBoolPattern { location } => location.start,
             Error::UnknownLabels { unknown, .. } => {
                 unknown.iter().map(|(_, s)| s.start).min().unwrap_or(0)
             }
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1277,22 +1277,14 @@ where
                         }
                     }
                 } else {
-                    match name.as_str() {
-                        "true" | "false" => {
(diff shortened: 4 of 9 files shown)


mxtthias:
I'm really not sure if I've done this the right way, so any feedback is much appreciated!

compiler-core/src/error.rs
```diff
@@ -3985,6 +3985,22 @@ modules should not import them. Perhaps change `{package}` to a regular dependen
...
                     },
+
+                    TypeError::LowcaseBooleanPattern { location } => Diagnostic {
+                        title: "Lowercase boolean pattern".to_string(),
+                        text: "See: https://tour.gleam.run/basics/bools/".into(),
+                        hint: Some("In Gleam boolean literals are `True` and `False`.".into()),
+                        level: Level::Error,
+                        location: Some(Location {
+                            label: Label {
+                                text: Some("Did you want a Bool instead of a variabel?".into()),
```
giacomocavalieri: Small typo
```suggestion
                                text: Some("Did you want a Bool instead of a variable?".into()),
```

compiler-core/src/type_/error.rs
```diff
@@ -664,6 +664,10 @@ pub enum Error {
         location: SrcSpan,
         name: EcoString,
     },
+
+    LowcaseBooleanPattern {
```
giacomocavalieri: Since you're here could you rename the variant to `LowercaseBooleanPattern`? We are trying to get rid of shorthand names as we come across them, and this one too could be updated

mxtthias:
@giacomocavalieri Thank you for your review! Please let me know if you want me to squash the commits or so.

giacomocavalieri:
Thank you! No need to squash, the only thing we want to avoid is merge commits, anything else is totally fine 😊

giacomocavalieri:
Looks great to me, now it just needs Louis' review

compiler-core/src/type_/error.rs
```diff
@@ -664,6 +664,10 @@ pub enum Error {
         location: SrcSpan,
         name: EcoString,
     },
+
+    LowercaseBooleanPattern {
```
lpil:
```suggestion
    LowercaseBoolPattern {
```

compiler-core/src/error.rs
```diff
@@ -3984,6 +3984,22 @@ modules should not import them. Perhaps change `{package}` to a regular dependen
                             extra_labels: vec![],
                         }),
                     },
+
+                    TypeError::LowercaseBooleanPattern { location } => Diagnostic {
+                        title: "Lowercase boolean pattern".to_string(),
```
lpil:
```suggestion
                        title: "Lowercase bool pattern".to_string(),
```

compiler-core/src/error.rs
```diff
@@ -3984,6 +3984,22 @@ modules should not import them. Perhaps change `{package}` to a regular dependen
...
+                        text: "See: https://tour.gleam.run/basics/bools/".into(),
+                        hint: Some("In Gleam boolean literals are `True` and `False`.".into()),
```
lpil:
```suggestion
                        hint: Some("In Gleam bool literals are `True` and `False`.".into()),
```

compiler-core/src/error.rs
```diff
@@ -3984,6 +3984,22 @@ modules should not import them. Perhaps change `{package}` to a regular dependen
...
+                        level: Level::Error,
+                        location: Some(Location {
+                            label: Label {
+                                text: Some("Did you want a Bool instead of a variable?".into()),
```
lpil:
```suggestion
                                text: Some("This is not a bool".into()),
```

CHANGELOG.md
```diff
@@ -15,6 +15,11 @@
   on the Erlang target.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The "Did you want a Bool" error is no longer a syntax error, but instead a
```
lpil:
```suggestion
- The lowercase bool pattern error is no longer a syntax error, but instead a
```

mxtthias:
@lpil Thank you for the comments! Really appreciate help with the wording!

giacomocavalieri:
Hello! This has conflicts with the main branch, would you mind rebasing?

PR #5057, opened 2025-10-20
Fix code generation for alternative string prefix patterns with an alias

### LINKED ISSUE #5055: Issue with #5039 not acting with certain pattern matching code
There was an optimisation added to 1.13.0  #5039  to improve the erlang pattern matching.
I am building the below code:
```
pub fn pop2(input, _) {
  case input {
    "a" as j <> tail
    | "b" as j <> tail
    | "c" as j <> tail
    | "d" as j <> tail
    | "e" as j <> tail...

    | "Z" as j <> tail -> pop2(tail, j)
```
This appears to still be building as:
```
-file("src/strcmptest.gleam", 50).
-spec pop2(binary(), binary()) -> nil.
pop2(Input, _) ->
    case Input of
        <<J:1/binary, Tail/binary>> when J =:= <<"a"/utf8>> ->
            pop2(Tail, J);

        <<J:1/binary, Tail/binary>> when J =:= <<"b"/utf8>> ->
            pop2(Tail, J);

        <<J:1/binary, Tail/binary>> when J =:= <<"c"/utf8>> ->
            pop2(Tail, J);

        <<J:1/binary, Tail/binary>> when J =:= <<"d"/utf8>> ->
            pop2(Tail, J);

        <<J:1/binary, Tail/binary>> when J =:= <<"e"/utf8>> ->
            pop2(Tail, J);...

        <<J:1/binary, Tail/binary>> when J =:= <<"Z"/utf8>> ->
            pop2(Tail, J);
```

This is the same as when using gleam 1.12.0
Initially I thought this was due to the recursive call to pop2 but when I rebuilt my uri library (https://github.com/pendletong/uri) the various parsers within parser.gleam still have the above structure:
```
-file("src/gluri/internal/parser.gleam", 794).
?DOC(false).
-spec parse_alpha(binary()) -> {ok, {binary(), binary()}} | {error, nil}.
parse_alpha(Str) ->
    case Str of
        <<Char:1/binary, Tail/binary>> when Char =:= <<"a"/utf8>> ->
            {ok, {Char, Tail}};

        <<Char:1/binary, Tail/binary>> when Char =:= <<"b"/utf8>> ->
            {ok, {Char, Tail}};

        <<Char:1/binary, Tail/binary>> when Char =:= <<"c"/utf8>> ->
            {ok, {Char, Tail}};...



        <<Char:1/binary, Tail/binary>> when Char =:= <<"Z"/utf8>> ->
            {ok, {Char, Tail}};
```
is built from
```
fn parse_alpha(str: String) -> Result(#(String, String), Nil) {
  case str {
    "a" as char <> tail
    | "b" as char <> tail
    | "c" as char <> tail...

   | "Z" as char <> tail -> Ok(#(char, tail))

```

giacomocavalieri:
This PR closes #5055

There was a bug in a recently merged change where the generated code would be invalid for alternative patterns with string prefixes. Now that is fixed (I didn't update the changelog as this falls under that same change)

### DIFF
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__string_prefix_assignment_with_multiple_subjects.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__string_prefix_assignment_with_multiple_subjects.snap
@@ -27,7 +27,7 @@ go(X) ->
             Digit;
 
         <<"2"/utf8, _/binary>> ->
-            Digit = <<"1"/utf8>>,
+            Digit = <<"2"/utf8>>,
             Digit;
 
         _ ->
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -1551,7 +1551,7 @@ fn record_constructor_function(tag: &EcoString, arity: usize) -> Document<'_> {
 fn clause<'a>(clause: &'a TypedClause, environment: &mut Env<'a>) -> Document<'a> {
     let Clause {
         guard,
-        pattern: pat,
+        pattern,
         alternative_patterns,
         then,
         ..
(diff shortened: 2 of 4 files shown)


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__strings__string_prefix_assignment_with_multiple_subjects.snap
```diff
@@ -27,7 +27,7 @@ go(X) ->
             Digit;
 
         <<"2"/utf8, _/binary>> ->
-            Digit = <<"1"/utf8>>,
+            Digit = <<"2"/utf8>>,
```
lpil, line 30: oh no! I should have caught this here

PR #5059, opened 2025-10-20
Fix formatting of doubly negated literal ints

### LINKED ISSUE #5056: Formatting integers with multiple negation always results in negative values
```gleam
pub fn main() -> Nil {
  echo --1

  Nil
}
```

produces:

```sh
1
```

But formatting the Gleam file results in:

```gleam
pub fn main() -> Nil {
  echo -1

  Nil
}
```

which, when ran, produces:

```sh
-1
```

I would've assumed that, if the number of times negations was `n`, then the reduced version would contain `n % 2` negations.

If my assumption is wrong, then the [blog post](https://gleam.run/news/formalising-external-apis/) might be incorrect with its example of:

```
pub fn useless_negations() {
...
```
---

```
OS: macOS 26.0
Memory: 64 GiB
Architecture: aarch64
Gleam version: 1.13.0
```

giacomocavalieri:
This PR closes #5056
I forgot to actually remove the double negation when dealing with literal ints, that means the formatter will produce the wrong code! It should be fixed now, sorry about that 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,3 +28,7 @@
   offered when calling a function unsupported by the current target, leading to
   invalid code if the code action was accepted.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where the formatter would not remove the right number of double
+  negations from literal integers.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))


JosephTLyons:
Thanks for fixing this! 

Also, just checked and the changelog line appears to be a duplicate of the previous bug fix added.

giacomocavalieri:
Oops I really messed that up, fixed!

PR #5061, opened 2025-10-20
Add record update syntax for constant definitions

### LINKED ISSUE #3695: Allow spreading const records in type constructors for consts
Currently, `const` declarations do not support spreading records:
```rs
const a = Foo(1, 2)
const b = Foo(..a, 3)
```

I propose that if the spread record is also a `const`, it should be allowed.

abs0luty:
- Closes #3695

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1,7 +1,8 @@
 use std::{collections::HashSet, iter, sync::Arc};
 
 use crate::{
-    Error, STDLIB_PACKAGE_NAME, analyse,
+    Error, STDLIB_PACKAGE_NAME,
+    analyse::Inferred,
     ast::{
         self, ArgNames, AssignName, AssignmentKind, BitArraySegmentTruncation, BoundVariable,
         BoundVariableName, CallArg, CustomType, FunctionLiteralKind, ImplicitCallArgOrigin, Import,
@@ -902,7 +903,7 @@ impl<'ast> ast::visit::Visit<'ast> for FillInMissingLabelledArgs<'ast> {
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {
@@ -1667,14 +1668,14 @@ impl<'ast, IO> ast::visit::Visit<'ast> for QualifiedToUnqualifiedImportFirstPass
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {
         let range = src_span_to_lsp_range(*location, self.line_numbers);
         if overlaps(self.params.range, range)
             && let Some((module_alias, _)) = module
-            && let analyse::Inferred::Known(_) = constructor
+            && let Inferred::Known(_) = constructor
             && let Some(import) = self.get_module_import(module_alias, name, ast::Layer::Value)
         {
             self.qualified_constructor = Some(QualifiedConstructor {
@@ -1705,7 +1706,7 @@ impl<'ast, IO> ast::visit::Visit<'ast> for QualifiedToUnqualifiedImportFirstPass
         arguments: &'ast Vec<CallArg<ast::TypedConstant>>,
         tag: &'ast EcoString,
         type_: &'ast Arc<Type>,
-        field_map: &'ast Option<FieldMap>,
+        field_map: &'ast Inferred<FieldMap>,
         record_constructor: &'ast Option<Box<ValueConstructor>>,
     ) {
         let range = src_span_to_lsp_range(*location, self.line_numbers);
@@ -1909,12 +1910,12 @@ impl<'ast> ast::visit::Visit<'ast> for QualifiedToUnqualifiedImportSecondPass<'a
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {
         if let Some((module_alias, _)) = module
-            && let analyse::Inferred::Known(_) = constructor
+            && let Inferred::Known(_) = constructor
         {
             let QualifiedConstructor {
                 used_name,
@@ -1948,7 +1949,7 @@ impl<'ast> ast::visit::Visit<'ast> for QualifiedToUnqualifiedImportSecondPass<'a
         arguments: &'ast Vec<CallArg<ast::TypedConstant>>,
         tag: &'ast EcoString,
         type_: &'ast Arc<Type>,
-        field_map: &'ast Option<FieldMap>,
+        field_map: &'ast Inferred<FieldMap>,
         record_constructor: &'ast Option<Box<ValueConstructor>>,
     ) {
         if let Some((module_alias, _)) = module {
@@ -2158,7 +2159,7 @@ impl<'ast> ast::visit::Visit<'ast> for UnqualifiedToQualifiedImportFirstPass<'as
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {
@@ -2167,7 +2168,7 @@ impl<'ast> ast::visit::Visit<'ast> for UnqualifiedToQualifiedImportFirstPass<'as
                 self.params.range,
                 src_span_to_lsp_range(*location, self.line_numbers),
             )
-            && let analyse::Inferred::Known(constructor) = constructor
+            && let Inferred::Known(constructor) = constructor
         {
             self.get_module_import_from_value_constructor(&constructor.module, name);
         }
@@ -2193,7 +2194,7 @@ impl<'ast> ast::visit::Visit<'ast> for UnqualifiedToQualifiedImportFirstPass<'as
         arguments: &'ast Vec<CallArg<ast::TypedConstant>>,
         _tag: &'ast EcoString,
         _type_: &'ast Arc<Type>,
-        _field_map: &'ast Option<FieldMap>,
+        _field_map: &'ast Inferred<FieldMap>,
         record_constructor: &'ast Option<Box<ValueConstructor>>,
     ) {
         if module.is_none()
@@ -2401,7 +2402,7 @@ impl<'ast> ast::visit::Visit<'ast> for UnqualifiedToQualifiedImportSecondPass<'a
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {
@@ -2436,7 +2437,7 @@ impl<'ast> ast::visit::Visit<'ast> for UnqualifiedToQualifiedImportSecondPass<'a
         arguments: &'ast Vec<CallArg<ast::TypedConstant>>,
         tag: &'ast EcoString,
         type_: &'ast Arc<Type>,
-        field_map: &'ast Option<FieldMap>,
+        field_map: &'ast Inferred<FieldMap>,
         record_constructor: &'ast Option<Box<ValueConstructor>>,
     ) {
         if module.is_none() {
@@ -6222,7 +6223,7 @@ impl<'ast, IO> ast::visit::Visit<'ast> for GenerateVariant<'ast, IO> {
         name: &'ast EcoString,
         arguments: &'ast Vec<CallArg<TypedPattern>>,
         module: &'ast Option<(EcoString, SrcSpan)>,
-        constructor: &'ast analyse::Inferred<type_::PatternConstructor>,
+        constructor: &'ast Inferred<type_::PatternConstructor>,
         spread: &'ast Option<SrcSpan>,
         type_: &'ast Arc<Type>,
     ) {


compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            Constant::Record {
+                module,
+                location,
+                name,
+                arguments,
+                spread,
                 // field_map, is always None here because untyped not yet unified
                 ..
-            } if arguments.is_empty() => {
+            } if arguments.is_empty() && spread.is_none() => {
```
lpil: Can we swap this guard for `spread: None` please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
                 module,
                 location,
                 name,
                 arguments,
+                spread: Some(spread_const),
+                ..
+            } => {
+                let constructor = self.infer_value_constructor(&module, &name, &location)?;
+
+                let (tag, field_map, _variant_index) = match &constructor.variant {
```
lpil: `_variant_index` is unused here

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,115 @@ fn custom_type_module_constants() {
...
+
+#[test]
+fn const_record_spread_non_record() {
+    assert_module_error!(
+        "pub type Person { Person(name: String, age: Int) }
+        pub const not_a_record = 42
+        pub const bob = Person(..not_a_record, name: \"Bob\")"
+    );
+}
+
```
lpil, line 2417: Could you add tests for these please 🙏 : 

- Multi-variant custom types
- Variants without any arguments
- Unlabelled fields

compiler-core/src/ast/constant.rs
```diff
@@ -40,6 +40,7 @@ pub enum Constant<T, RecordTag> {
         module: Option<(EcoString, SrcSpan)>,
         name: EcoString,
         arguments: Vec<CallArg<Self>>,
+        spread: Option<Box<Self>>,
```
lpil: For non-constant expressions record updates have their own AST node, is there a reason not to do that here? It would mean that the overrides have to be labelled, which is nice to enforce.

compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                // Unify types - the spread should have the same type as what the constructor returns
+                let expected_type = match constructor.type_.as_ref() {
+                    Type::Fn { return_: ret, .. } => ret.clone(),
+                    _ => constructor.type_.clone(),
+                };
+                unify(expected_type.clone(), typed_spread.type_())
+                    .map_err(|e| convert_unify_error(e, typed_spread.location()))?;
+
+                // Resolve the spread if it's a constant variable
```
lpil: What does this mean?

compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                        ValueConstructorVariant::LocalConstant { literal }
+                        | ValueConstructorVariant::ModuleConstant { literal, .. } => {
+                            literal.clone()
+                        }
+                        _ => typed_spread,
+                    },
+                    _ => typed_spread,
+                };
+
+                // Extract spread arguments
```
lpil: What does this mean?

compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    let label = arg.label.clone();
+                    let typed_value = self.infer_const(&None, arg.value);
+
+                    // Find expected type for this field and validate field exists
+                    if let Some(label_name) = &label
+                        && let Some(field_map) = &field_map
+                    {
+                        match field_map.fields.get(label_name) {
+                            Some(&index) => {
+                                if let Type::Fn {
```
lpil: Sorry, I'm getting a bit confused. What's this case for fields that contain functions? It looks like we do some special type checking of just function arguments, but we want to type check all field types.

compiler-core/src/type_/expression.rs
```diff
@@ -3828,14 +3828,157 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    if arg.uses_label_shorthand() {
+                        self.track_feature_usage(FeatureKind::LabelShorthandSyntax, arg.location);
+                    }
+
+                    let label = arg.label.clone();
+                    let typed_value = self.infer_const(&None, arg.value);
+
+                    // Find expected type for this field and validate field exists
+                    if let Some(label_name) = &label
+                        && let Some(field_map) = &field_map
```
lpil: Why is field map optional at this point?

lpil (commented):
Thank you! This will be a relly nice addition. I've left a few questions in line.

Could you add a test that ensures that `pub const dog = Animal(..alice)` emits the "fieldless record update" warning please 🙏

compiler-core/src/ast/constant.rs
```diff
@@ -46,6 +46,17 @@ pub enum Constant<T, RecordTag> {
...
 
+    RecordUpdate {
+        location: SrcSpan,
+        module: Option<(EcoString, SrcSpan)>,
+        name: EcoString,
+        record: Box<Self>,
+        arguments: Vec<ConstantRecordUpdateArg<Self>>,
+        tag: RecordTag,
+        type_: T,
+        field_map: Option<FieldMap>,
```
lpil: If this is optional due to it only being present after type checking then please use the `Inferred` type instead, or make it a type parameter like the other information added during analysis. 🙏

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,187 @@ fn custom_type_module_constants() {
...
+#[test]
+fn const_record_spread_variant_without_args_warning() {
+    // Test that spreading 0-arity constructors produces a warning
+    assert_warning!(
+        "pub type Status { Active Inactive }
+        pub const status1 = Active
+        pub const status2 = Active(..status1)"
```
lpil, line 2452: This should be an error, `Active` is not a record constructor as it doesn't have any fields.

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,187 @@ fn custom_type_module_constants() {
...
+    );
+}
+
+#[test]
+fn const_record_spread_variant_without_args() {
+    // Test spreading with variants that have no arguments
+    assert_module_infer!(
+        "pub type Status { Active Inactive }
+        pub const status1 = Active
+        pub const status2 = Active(..status1)",
```
lpil: This should be an error, Active is not a record constructor as it doesn't have any fields.

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,187 @@ fn custom_type_module_constants() {
...
+    );
+}
+
+#[test]
+fn const_record_spread_unlabelled_fields() {
+    // Test spreading with tuple-like constructors (unlabelled fields)
+    assert_module_infer!(
+        "pub type Point { Point(Int, Int) }
+        pub const origin = Point(0, 0)
+        pub const point = Point(..origin)",
```
lpil: This should be an error, Active doesn't have any labelled fields.

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,187 @@ fn custom_type_module_constants() {
...
+            ("dog", "Animal"),
+        ]
+    );
+}
+
+#[test]
+fn const_record_spread_multi_variant() {
+    // Test spreading with multi-variant custom types
+    assert_module_infer!(
+        "pub type Pet { Dog(name: String, age: Int) Cat(name: String, breed: String) }
```
lpil: Split this onto multiple lines please, it's tricky to read with this minified syntax

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,187 @@ fn custom_type_module_constants() {
...
+        "pub type Point { Point(Int, Int) }
+        pub const origin = Point(0, 0)
+        pub const point = Point(..origin)",
+        vec![
+            ("Point", "fn(Int, Int) -> Point"),
+            ("origin", "Point"),
+            ("point", "Point"),
+        ]
+    );
+}
```
lpil, line 2464: Could you add tests for multi-variant custom type where the other variant is being used for the update please 🙏

lpil (commented):
Thank you! Looking fab. I've left some notes inline, mostly around behaviour that doesn't match record updates for expressions.

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__expression_constructor_update.snap
```diff
@@ -13,10 +13,10 @@ pub fn update_person(person: Person) {
 }
 
 ----- ERROR
-error: Invalid record constructor
+error: Invalid record update
   ┌─ /src/one/two.gleam:7:3
   │
 7 │   constructor(..person)
-  │   ^^^^^^^^^^^ This is not a record constructor
+  │   ^^^^^^^^^^^ This constructor has no labelled fields
```
lpil: This error is wrong now, it is not a record constructor at all. It is a normal function.

abs0luty: Oops, didn't notice this!

abs0luty:
I suppose I now need to accommodate for changes in c2fed7e!

abs0luty:
@lpil Dependency validation fails and this is not directly related to the PR 🤔

- Made an issue for this: #5186

compiler-core/src/javascript/expression.rs
```diff
@@ -2253,6 +2257,7 @@ impl<'module, 'a> Generator<'module, 'a> {
                     return record_constructor(type_.clone(), None, name, arity, self.tracker);
                 }
 
+                // Spreads are fully expanded during type checking, so we just handle arguments
```
lpil: No such thing as a spread in Gleam!

What does this comment mean?

compiler-core/src/ast/constant.rs
```diff
@@ -349,3 +418,26 @@ impl<A, B> bit_array::GetLiteralValue for Constant<A, B> {
         }
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ConstantRecordUpdateArg<Constant> {
```
lpil: Looks like this could replace `UntypedRecordUpdateArg`, the definitions are the same.

compiler-core/src/javascript/expression.rs
```diff
@@ -1839,6 +1839,7 @@ impl<'module, 'a> Generator<'module, 'a> {
...
```
lpil: No such thing as a spread in Gleam!

What does this comment mean?

compiler-core/src/type_/error.rs
```diff
@@ -340,7 +340,7 @@ pub enum Error {
         location: SrcSpan,
     },
 
-    RecordUpdateInvalidConstructor {
+    InvalidRecordConstructor {
```
lpil: Undo this rename please, it was more descriptive before.

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                };
+
+                let (tag, field_map) = match &constructor.variant {
+                    ValueConstructorVariant::Record {
+                        name, field_map, ..
+                    } => (
+                        name.clone(),
+                        match field_map {
+                            Some(fm) => Inferred::Known(fm.clone()),
+                            None => Inferred::Unknown,
```
lpil: This doesn't look right- the field map has been inferred by this point, but none exists. This is saying that inference hasn't run yet.

If you remove this not-quite-right use of `Inferred` then you can remove all the conditional logic with `unreachable` clauses below!

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    Type::Fn { return_: ret, .. } => ret.clone(),
+                    _ => constructor.type_.clone(),
+                };
+                if let Err(error) = unify(expected_type.clone(), typed_record_type.clone()) {
+                    self.problems
+                        .error(convert_unify_error(error, typed_record.location()));
+                    return self.new_invalid_constant(location);
+                }
+
+                // If the record being updated is a reference to a constant variable, resolve it to get the actual record value
```
lpil: Wrap long lines please

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    | ValueConstructorVariant::LocalConstant { literal } => {
+                        return literal.clone();
+                    }
+                };
+
+                // Type-check the record being updated
+                let typed_record = self.infer_const(&None, *record.clone());
+                let typed_record_type = typed_record.type_();
+
+                // Unify types - the record being updated should have the same type as what the constructor returns
```
lpil: Wrap long lines please

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                let expected_type = match constructor.type_.as_ref() {
+                    Type::Fn { return_: ret, .. } => ret.clone(),
```
lpil: No abbreviations please

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                        | ValueConstructorVariant::ModuleConstant { literal, .. } => {
+                            literal.clone()
+                        }
+                        _ => typed_record,
+                    },
+                    _ => typed_record,
+                };
+
+                // Get the field arguments from the record that we'll use as the base
+                let (base_args, base_tag) = match resolved_record {
```
lpil: No abbreviations please

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                            tag,
+                            field_map,
+                            record_constructor: Some(Box::new(constructor)),
+                        };
+                    }
+                };
+
+                // Check if the constructor has labelled fields
+                // Tuple-like constructors (unlabelled fields) cannot be spread
+                if matches!(field_map, Inferred::Unknown) {
```
lpil: Don't use `matches` please.

This isn't checking if the record constructor has unlabelled fields, it is checking if type inference has run yet. It has at this point, so it should be impossible for it to be `Inferred::Unknown`

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+                    self.problems.error(Error::InvalidRecordConstructor {
+                        location: constructor_location,
+                    });
+                    return self.new_invalid_constant(location);
+                }
+
+                let mut remaining_fields = match &field_map {
+                    Inferred::Known(field_map) => field_map.fields.clone(),
+                    Inferred::Unknown => unreachable!("record update without a field map return"),
+                };
```
lpil: Never use the check-then-assert pattern. Always use pattern matching.

If you avoid `matches` and other boolean predicates then it'll be easier not fall into this trap.

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                let explicit_args_len = arguments.len();
```
lpil: No abbreviations please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -3828,12 +3835,265 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                // Type-check explicit override arguments
+                let mut typed_overrides = Vec::new();
+                for arg in arguments {
```
lpil: No abbreviations please

compiler-core/src/type_/tests.rs
```diff
@@ -2305,6 +2305,164 @@ fn custom_type_module_constants() {
...
+#[test]
+fn const_record_update_unlabelled_fields() {
+    assert_module_error!(
+        "pub type Point { Point(Int, Int) }
+
+        pub const origin = Point(0, 0)
+        pub const point = Point(..origin)"
+    );
+}
+
```
lpil, line 2465: Could you add a test that shows it handles re-specialisation of generic types please? 🙏 

The const version of this code

```rust
pub type Box(a) {
  Box(name: String, value: a)
}

pub fn main() {
  let base = Box("score", 50)
  let updated = Box(..base, value: "Hello")
}
```

compiler-core/src/ast_folder.rs
```diff
@@ -1184,6 +1226,39 @@ pub trait UntypedConstantFolder {
...
+                record,
+                arguments,
+                tag,
+                type_,
+                field_map,
+            } => {
+                let record = Box::new(self.fold_constant(*record));
+                let arguments = arguments
+                    .into_iter()
+                    .map(|arg| ConstantRecordUpdateArg {
```
lpil: No abbreviations please

compiler-core/src/ast/constant.rs
abs0luty: https://github.com/gleam-lang/gleam/blob/ba3282809d8fcda3d2d7a7357a8077ab64b9fc9c/compiler-core/src/ast.rs#L1595-L1600

The value of `UntypedRecordUpdateArg` is not generic, it's an untyped expression! However, I can do something like this, which is a common pattern:

```diff
+ struct RecordUpdateArg<A> {
       pub label: EcoString,
       pub location: SrcSpan,
-      pub value: UntypedExpr,
+      pub value: A,
  }

+  pub type UntypedRecordUpdateArg = RecordUpdateArg<UntypedExpr>;
```

PR #5062, opened 2025-10-20
Emphasize mismatching part of complex type mismatches

### LINKED ISSUE #4578: Emphasize mismatching part of complex type mismatches
When a type mismatch is complex, and some types are unbound (is this the correct term?), it can be hard to see where the error actually is. Here is a concrete example I actually wasted quite a bit of time on:

```
Expected type:

    drift.Step(List(String), Input, Output, Nil)

Found type:

    drift.Step(List(string), Input, a, b)
```

It could be made clearer by emphasizing where the type mismatch was exactly:
```
Expected type:

    drift.Step(List(String), Input, Output, Nil)
                    ^^^^^^

Found type:

    drift.Step(List(string), Input, a, b)
                    ^^^^^^
```

abs0luty:
- Closes #4578

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -12,7 +12,7 @@ use crate::type_::error::{
     IncorrectArityContext, InvalidImportKind, MissingAnnotation, ModuleValueUsageContext, Named,
     UnknownField, UnknownTypeHint, UnsafeRecordUpdateReason,
 };
-use crate::type_::printer::{Names, Printer};
+use crate::type_::printer::{Names, Printer, TypeAnnotation, TypePathStep};
 use crate::type_::{FieldAccessUsage, error::PatternMatchKind};
 use crate::{ast::BinOp, parse::error::ParseErrorType, type_::Type};
 use crate::{bit_array, diagnostic::Level, type_::UnifyErrorSituation};
@@ -2143,6 +2143,9 @@ But function expects:
                         situation,
                     } => {
                         let mut printer = Printer::new(names);
+                        let expected = collapse_links(expected.clone());
+                        let given = collapse_links(given.clone());
+
                         let mut text = if let Some(description) =
                             situation.as_ref().and_then(|s| s.description())
                         {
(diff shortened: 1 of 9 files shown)


compiler-core/src/error.rs
```diff
@@ -4578,6 +4615,205 @@ fn hint_alternative_operator(op: &BinOp, given: &Type) -> Option<String> {
...
+    // Criterion 2: Deep nesting (depth >= 2)
+    // For example: Wrapper(List(Int)) has depth 2
+    if max_depth >= 2 {
+        return true;
+    }
+
+    // If we get here, we have exactly 1 mismatch at depth 1
+    // This is like List(Int) vs List(String) - simple enough to not need smart diff
+    false
+}
```
abs0luty, line 4815: @lpil what's your opinion on when to display errors this way?

giacomocavalieri:
I would also like to think how these will appear in different settings. As nice as this looks when running `gleam build` I think it could get confusing real fast when the error pops up in an IDE like Zed or VSCode where wrapping can happen:
- VSCode: Screenshot 2025-10-21 alle 11 36 41
- Zed: Screenshot 2025-10-21 alle 11 36 54

lpil:
That looks like a bug in Zed to me @giacomocavalieri. We tell it to render it as markdown.

giacomocavalieri:
Yeah, the wrapping can still make it hard to read though

abs0luty:
@giacomocavalieri Good point about wrapping. Thanks! I think we can solve it in many different ways:

We can break the mismatch into bullet points under the main message. Something like "Mismatch at argument 2: expected Float, found String" without trying to align under the original text, or something like this:

  ```
  Box(#(#(Int, a), b, List(c)))
  
  Expected a to be Float, found String
  Expected b to be Float, found Int
  Expected c to be Float, found Int
  ```

Or even separate error messages:

```
Box(#(#(Int, ...), Int, List(Int)))
  
Expected type: Float
Found type: String

Box(#(#(Int, String), ..., List(Int)))
  
Expected type: Float
Found type: Int

Box(#(#(Int, String), Int, List(...)))
  
Expected type: Float
Found type: Int
```

We can also control the block width manually (e.g., add zero-width spaces or break into smaller lines ourselves), although there's always a limit to this:
  ```
  Box(
    #(
      #(Int, String),
             ^^^^^^ expected Float
      Int,
      ^^^ expected Float
      List(Int)
           ^^^ expected Float
    )
  )
  ```


What do you think?

lpil:
The first two I find hard to understand, but that third one could be good to explore.

How would we ensure it renders correctly in all LS clients? I don't think the protocol specifies how the diagnostic text should be handled, and in some UIs all newlines will be removed.

abs0luty:
@lpil I think we can also do this: in the LS context we can try to separate these type mismatches. The internal data structure stays the same, but in compiler mode it appears as one large type mismatch error with many arrows. In the LS it shows as several highlights with distinct errors.

What do you think?

lpil:
Closing this as stale. Please feel free to reopen 🙏 Thank you

PR #5064, opened 2025-10-21
Fix a couple of bugs with the pattern match on value code action

giacomocavalieri:
This fixes two bugs I found with the "pattern match on value code action":
- It now generates valid syntax when expanding a list tail in a case branch
- It will no longer generate invalid patterns with duplicate variables
- It now uses the same rules to come up with names for tuple items as all other pieces of data

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1678,6 +1678,24 @@ impl TypedClause {
     }
 }
 
+/// Returns true if a pattern and an expression are the same: that is the expression
+/// would be building the exact matched value back.
+/// For example, if I had a branch like this:
+///
+/// ```gleam
+/// [a, b, c] -> [a, b, c]
+/// ```
+///
+/// The pattern and the expression would indeed be the same. However, if I had
+/// something like this:
+///
+/// ```gleam
+/// [first, ..rest] -> [first]
+/// ```
+///
+/// They wouldn't be the same! I'm not building back exactly the value the
+/// pattern can match on.
+///
 fn pattern_and_expression_are_the_same(pattern: &TypedPattern, expression: &TypedExpr) -> bool {
     match (pattern, expression) {
         // A pattern could be the same as a block if the block is wrapping just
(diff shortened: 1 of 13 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -4660,16 +4706,37 @@ impl<'a, IO> PatternMatchOnValue<'a, IO> {
     fn match_on_clause_variable(
         &mut self,
         variable_type: Arc<Type>,
-        variable_location: SrcSpan,
+        variable_location: PatternLocation,
         clause_location: SrcSpan,
+        bound_variables: Vec<BoundVariable>,
```
lpil: Couldn't this be `&[BoundVariable]`?

compiler-core/src/language_server/code_action.rs
```diff
@@ -5053,15 +5144,56 @@ impl<'ast, IO> ast::visit::Visit<'ast> for PatternMatchOnValue<'ast, IO> {
...
+                ..
+            } = tail.as_ref()
+        {
+            (name, location, type_)
+        } else {
+            ast::visit::visit_typed_pattern_list(self, location, elements, tail, type_);
+            return;
+        };
+
+        // TODO: this is a bit of an hack where we hardcode the start of the tail
```
lpil: `TODO` left in here! The whitespace handling is still absent

compiler-core/src/ast.rs
```diff
@@ -1667,6 +1667,16 @@ impl TypedClause {
...
     }
+
+    /// This returns the names of all the variables bound in this case clause.
+    /// For example if we had `#(a, b) | c` this will return "a", "b", and "c".
+    pub fn bound_variables(&self) -> Vec<BoundVariable> {
+        std::iter::once(&self.pattern)
+            .chain(&self.alternative_patterns)
+            .flatten()
+            .flat_map(|pattern| pattern.bound_variables())
+            .collect_vec()
```
lpil: Generally good to return an iterator rather than forcing the vec allocation, in case callers don't need a vec

compiler-core/src/language_server/code_action.rs
giacomocavalieri: Regarding this, to properly handle that we'd need to add a new field to a list patterns' tail to also store its start, would that be ok?

lpil: Yup! That's what we normally do!

compiler-core/src/ast.rs
```diff
@@ -2488,6 +2518,34 @@ impl Default for Inferred<()> {
...
+    /// //       ^^ no name
+    ///
+    /// [wibble, ..rest]
+    /// //       ^^^^^^ a variable name
+    ///
+    /// [wibble, .._rest]
+    /// //       ^^^^^^^ a discarded name
+    /// ```
+    ///
+    pub pattern: Box<Pattern<Type>>,
```
lpil: Box the tailpattern rather than boxing an element within the tail pattern 🙏 As there's nothing in this data structure that warrants boxing this field

giacomocavalieri: Oh right!

PR #5065, opened 2025-10-22
Add a missing BitArray constructor return type in the prelude

richard-viney:
The lack of this return type causes the following error:

```
error TS7013: Construct signature, which lacks return-type annotation, implicitly has an 'any' return type.
```


lpil (approved):
Thank you!

PR #5067, opened 2025-10-23
Fix invalid TS definition for long variant constructors with no args

richard-viney:
In v1.13.0 the following Gleam code

```gleam
pub type ValueRepresentation {
  UniversalResourceIdentifier
}
``` 

generates the following invalid TypeScript definition

```ts
export function ValueRepresentation$UniversalResourceIdentifier(
  ,
): ValueRepresentation$;
```

The comma isn't valid there and gives the error `Parameter declaration expected.` from `tsc`.

This issue occurs when the variant constructor function name is long, and it takes no arguments.

### DIFF
--- a/compiler-core/src/javascript/typescript.rs
+++ b/compiler-core/src/javascript/typescript.rs
@@ -670,12 +670,14 @@ impl<'a> TypeScriptGenerator<'a> {
         )
         .to_doc();
 
+        let has_arguments = !arguments.is_empty();
+
         docvec![
             "export function ",
             name_with_generics(function_name, type_parameters),
             "(",
             docvec![break_("", ""), join(arguments, break_(",", ", ")),].nest(INDENT),
-            break_(",", ""),
+            break_(if has_arguments { "," } else { "" }, ""),
             "): ",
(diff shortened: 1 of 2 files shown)


compiler-core/src/javascript/typescript.rs
```diff
@@ -655,12 +655,14 @@ impl<'a> TypeScriptGenerator<'a> {
...
 
+        let has_arguments = !arguments.is_empty();
+
         docvec![
             "export function ",
             name_with_generics(function_name, type_parameters),
             "(",
             docvec![break_("", ""), join(arguments, break_(",", ", ")),].nest(INDENT),
-            break_(",", ""),
+            break_(if has_arguments { "," } else { "" }, ""),
```
lpil: If there's no arguments then we want to replace this whole section with the literal `():` 🙏

richard-viney: Sure thing, updated to now keep the definition all on one line when there are no arguments.

Note that this does mean it can now exceed 80 columns.

richard-viney:
> The tests are missing!

Not any more!

PR #5071, opened 2025-10-24
Instantiate generics in mutually recursive functions before unification

### LINKED ISSUE #2533: Unbound type variable in top level definition after type inference
In this case, after type inference, the `main` function ends up having an inferred type with an unbound type variable instead of a generic one:
```gleam
pub type Foo(a) {
  Foo(value: Int)
}

pub fn main() {
  Foo(1)
}
```

Adding a type annotation turns the variable into a generic one though:
```gleam
// now the inferred type has a generic type variable instead
// of an unbound one
pub fn main() -> Foo(a) { ... }
```

### LINKED ISSUE #2550: Type inference issue expected Test(a) but found Test(a)
```gleam
type Test(a) {
  Test(a)
}
fn test(value: Test(a)) {
  test2(value)
}
fn test2(value: Test(a)) -> Test(a) {
  test(value)
}
```
Gives this error on the test2 call
```
Type mismatch

annotation of this function.

Expected type:

    Test(a)

Found type:

    Test(a)
```

Similar issue #1893

abs0luty:
- Closes #2533
- Closes #2550

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,25 @@
 
   ([Gavin Morrow](https://github.com/gavinmorrow))
 
+- Type inference now preserves generic type parameters when constructors or functions are used without explicit annotations, 
+  eliminating false errors in mutually recursive code:
+
+  ```gleam
+  type Test(a) {
+    Test(a)
+  }
+
+  fn it(value: Test(a)) {
+    it2(value)
+  }
+
+  fn it2(value: Test(a)) -> Test(a) {
+    it(value)
+  }
+  ```
+
+  Previously this could fail with an incorrect "type mismatch" error. ([Adi Salimgereyev](https://github.com/abs0luty))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,25 +55,6 @@
 
   ([Gavin Morrow](https://github.com/gavinmorrow))
 
-- Type inference now preserves generic type parameters when constructors or functions are used without explicit annotations, 
-  eliminating false errors in mutually recursive code:
-
-  ```gleam
-  type Test(a) {
-    Test(a)
-  }
-
-  fn it(value: Test(a)) {
-    it2(value)
-  }
-
-  fn it2(value: Test(a)) -> Test(a) {
-    it(value)
-  }
-  ```
-
-  Previously this could fail with an incorrect "type mismatch" error. ([Adi Salimgereyev](https://github.com/abs0luty))
-
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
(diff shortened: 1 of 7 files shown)


compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__type_variables_in_let_bindings_are_considered_when_adding_annotations.snap
```diff
@@ -15,7 +15,7 @@ fn wibble(a, b, c) {
 
 ----- AFTER ACTION
 
-fn wibble(a: e, b: f, c: g) -> fn(b, c) -> d {
+fn wibble(a: e, b: f, c: g) -> fn(b, c) -> h {
```
GearsDatapacks: This is not correct. The returned function is annotated as `fn(b, c) -> d`, so the annotations of this function should reflect that.

abs0luty: Got it!

giacomocavalieri:
Hello! Would you mind rebasing on main to fix the conflicts?

abs0luty:
@giacomocavalieri Only now noticed your message, sure!

giacomocavalieri:
Hello! It looks like tests are failing, would you mind having a look at those?

ankddev:
They all fail with this compile error:

```
error: unnecessary qualification
Error:     --> language-server/src/code_action.rs:1639:51
     |
1639 |     fn visit_typed_function(&mut self, fun: &'ast ast::TypedFunction) {
     |                                                   ^^^^^^^^^^^^^^^^^^
     |
     = note: `-D unused-qualifications` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(unused_qualifications)]`
help: remove the unnecessary path segments
     |
1639 -     fn visit_typed_function(&mut self, fun: &'ast ast::TypedFunction) {
1639 +     fn visit_typed_function(&mut self, fun: &'ast TypedFunction) {
     |
```

CHANGELOG.md
```diff
@@ -4,6 +4,25 @@
...
+  fn it(value: Test(a)) {
+    it2(value)
+  }
+
+  fn it2(value: Test(a)) -> Test(a) {
+    it(value)
+  }
+  ```
+
+  Previously this could fail with an incorrect "type mismatch" error. ([Adi Salimgereyev](https://github.com/abs0luty))
```
ankddev: You should place your name on new line

CHANGELOG.md
```diff
@@ -4,6 +4,25 @@
 
 ### Compiler
 
+- Type inference now preserves generic type parameters when constructors or functions are used without explicit annotations,
```
ankddev: This line is very long, I think

ankddev (commented):
Some notes about changelog entry

CHANGELOG.md
```diff
@@ -79,6 +79,24 @@
...
+
+  ```gleam
+  type Test(a) { Test(a) }
+
+  fn it(value: Test(a)) { it2(value) }
+  fn it2(value: Test(a)) -> Test(a) { it(value) }
+  ```
+
+  Previously, `it2` could be checked against the preregistered signature 
+  before that `a` had been instantiated to a fresh type variable.
```
giacomocavalieri: I don't think the changelog entry should be this technical about the implementation. I think this could be filed under "bug fixed" saying something along the lines of "The compiler no longer reports false type mismatch errors when typing mutually recursive function definitions".
Unification, preregistered types, fresh type variables, don't make much sense for anyone who doesn't know the inner workings of the compiler

giacomocavalieri (commented):
Hello! I left a little comment

PR #5076, opened 2025-10-25
fix: relative links in gleam toml

### LINKED ISSUE #4772: Permit adding relative links in `gleam.toml`
Currently links in `gleam.toml` have to be full URLs with a protocol. It is nice that we validate their syntax, but it means that you cannot use relative links, which @oderwat has a use-case for.

Should we disable the validation entirely and let the programmer be responsible for checking these issues?

I think we should prevent packages with relative links from being published to Hex, as those links would not be valid on HexDocs.

### Discussed in https://github.com/gleam-lang/gleam/discussions/4768

<sup>Originally posted by **oderwat** July 10, 2025</sup>
We need to build our own Gleam documentation server that works similarly to https://hexdocs.pm/, but builds documentation by accessing our private repositories containing private Gleam packages for our team's use.

The documentation updater clones the repositories and modifies the `gleam.toml` file to add back-links to the documentation server index. This is typically an HTTPS address for our documentation server, but when running locally, it's something like `http://localhost:8000`.

We attempted to add `links = [{ title = "Home", href = "/index.html" }]` to the `gleam.toml` file, but this doesn't work because the documentation builder validates that links have a valid scheme, throwing an error: `uri without scheme for key 'links'...`.

Is this validation really necessary? It also prevents adding relative links within the documentation (including fragment identifiers using "#").

We could work around this by using a fixed URI and performing string replacement in the generated documentation, but this approach feels clunky and inelegant.</div>

Courtcircuits:
# What has been done ? 

This PR aims to resolve #4772. I made the gleam.toml parsing less strict by allowing the usage of relative links within the documentation.

For example this manifest is now correct : 

```toml
name = "wisp_basic_auth_courtcircuits"
version = "1.0.298"

description = "Basic HTTP Authentication Scheme for Wisp - just testing gealm export"
[....]
links = [
                { title = "Website", href = "https://wisp.gleam.io" },
                { title = "Website internal", href = "./internal" },
]
[...]
```

The generated doc looks like this : 

```bash
gleam docs build
```

image

As you can see this doc contains a relative link to "./internal" which wasn't possible previously.

And the hex doc looks like this : 

```bash
gleam docs publish
```

image

You can find the [hexdoc page here](https://hex.pm/packages/wisp_basic_auth_courtcircuits).
The link called "Website internal" is not present since it would be strange for hex to contain relative links.

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -553,7 +553,12 @@ fn metadata_config<'a>(
         .iter()
         .map(|l| (l.title.as_str(), l.href.clone()))
         .filter(|(_, href)| !href.clone().is_internal())
-        .map(|(title, href)| (title, href.as_uri().expect("Internal link not marked as internal")))
+        .map(|(title, href)| {
+            (
+                title,
+                href.as_uri().expect("Internal link not marked as internal"),
+            )
+        })
         .chain(repo_url.into_iter().map(|u| ("Repository", u)))
         .collect();
 


compiler-cli/src/publish.rs
```diff
@@ -547,19 +548,28 @@ fn metadata_config<'a>(
             }),
         })
         .collect();
+    let links: Vec<(&str, Uri)> = config
+        .links
+        .iter()
+        .map(|l| (l.title.as_str(), l.href.clone()))
+        .filter(|(_, href)| !href.clone().is_internal())
```
lpil, line 555: Put the filter before the mapping and remove the `clone` please 🙏

compiler-core/src/config.rs
```diff
@@ -980,11 +980,44 @@ pub struct DocsPage {
...
+        }
+    }
+}
+
 #[derive(Deserialize, Serialize, Debug, PartialEq, Eq, Clone)]
 pub struct Link {
     pub title: String,
-    #[serde(with = "uri_serde")]
-    pub href: Uri,
+    pub href: Href,
```
lpil, line 1020: The Href type here looks like it'll permit data that's not valid URLs. How about instead we use the `URL` type from the `url` crate? It's already an indirect dep, so we can add it and use it without adding a new dep.

Courtcircuits: I read through `url` crate documentation and it seems like it needs a "base url" to validate a relative path, which is the goal of this PR.

[`url` documentation says](https://docs.rs/url/latest/url/#base-url) : 

Many contexts allow URL references that can be relative to a base URL:

```html
<link rel="stylesheet" href="../main.css">
```

Since parsed URLs are absolute, giving a base is required for parsing relative URLs:

```rust
use url::{Url, ParseError};

assert!(Url::parse("../main.css") == Err(ParseError::RelativeUrlWithoutBase))
```

Use the join method on an Url to use it as a base URL:

```rust
use url::Url;

let this_document = Url::parse("http://servo.github.io/rust-url/url/index.html")?;
let css_url = this_document.join("../main.css")?;
assert_eq!(css_url.as_str(), "http://servo.github.io/rust-url/main.css");
```

----

I think the main problem with my implementation is that recursive paths are valid, which should only be the case if the "base url" provides a path (e.g. http://myhexdoc.com/foo/bar). What I could do is to check if the user provided an "Home Page" within the link list, if so it will use that home page link as the base url. If not then I could fall back to a strict generic base url without path like `http://foo.com/`.

What do you think about that strategy ?

lpil: Sorry, I don't understand. What's a recursive path?

If the URL crate doesn't support relative ones then we'll need something else which does.

I don't think we can safely set a base URL as we don't know how many path segments the programmer would use? Wouldn't `https://example.com` as a base be invalid for `../something.html`?

PR #5078, opened 2025-10-25
Emit `optional_field` for optional fields in "generate dynamic decoder" action

### LINKED ISSUE #5077: using optional_field when autogenerating decoder
when generating decoder for a type, use optional_field by default for fields with option.Option(a) types
alternatively, add additional option for using optional_filed for that fields instead of regular field
example:
for that type
```
pub type Voice {
  Voice(
    file_id: String,
    file_unique_id: String,
    duration: Int,
    mime_type: Option(String),
    file_size: Option(Int),
  )
}
```
should be generated(or should be option to generate) this decoder
```
pub fn voice_decoder() -> decode.Decoder(Voice) {
  use file_id <- decode.field("file_id", decode.string)
  use file_unique_id <- decode.field("file_unique_id", decode.string)
  use duration <- decode.field("duration", decode.int)
  use mime_type <- decode.optional_field(
    "mime_type",
    option.None,
    decode.optional(decode.string),
  )
  use file_size <- decode.optional_field(
    "file_size",
    option.None,
    decode.optional(decode.int),
  )
  decode.success(Voice(
    file_id:,
    file_unique_id:,
    duration:,
    mime_type:,
    file_size:,
  ))
}
```

abs0luty:
- Closes #5077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,10 @@
   used on tuples.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The "generate dynamic decoder" code action now emits `decode.optional_field` for
+  `Option` fields.
+  ([Adi Salimgereyev](https://github.com/abs0luty))
+
 ### Formatter
 
 ### Bug fixes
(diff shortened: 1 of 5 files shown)


lpil:
Thank you, but we do not accept unplanned changes. Any changes to Gleam's design needs to be agreed to first 🙏

PR #5080, opened 2025-10-26
Typecheck let bindings using annotated type instead of inferred type

### LINKED ISSUE #4642: let bindings are typechecked using their inferred type not their (narrowed) annotated type
When a let binding produces a type error because of its annotation, the rest of the code is type-checked with the actual inferred type of the binding instead of the (possibly narrowed) annotated type. This potentially produces misleading follow-up type errors:

```gleam
pub fn main() {
  let x: String = 5  // type error: expected String, got Int
  let y: Int = x     // valid
  let z: String = x  // type error: expected String, got Int
}
```

I think the assignment to `y` should produce the type error here, not `z`. I would have expected one of the major benefits of adding explicit annotations to be to help the compiler produce better errors in exactly those situations. My assumption here of course is that if the user adds an explicit annotation, they want and expect that annotation to be assumed the correct type.

~ 💜

abs0luty:
- Closes #4642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,19 @@
 
   ([Gavin Morrow](https://github.com/gavinmorrow))
 
+- Let bindings now continue to assume their annotated type after mismatches,
+  preventing misleading follow-up errors:
+
+  ```gleam
+  pub fn main() {
+    let x: String = 5 // type error: expected String, got Int
+    let y: String = x // valid
+    let z: Int = x    // type error: expected Int, got String
+  }
+  ```
+
+  ([Adi Salimgereyev](https://github.com/abs0luty))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__module_could_not_unify9.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__module_could_not_unify9.snap
@@ -1,6 +1,8 @@
 ---
 source: compiler-core/src/type_/tests/errors.rs
+assertion_line: 771
 expression: "fn main() { let assert [1, 2, ..x]: List(String) = [1,2,3] x }"
+snapshot_kind: text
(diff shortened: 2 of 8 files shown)


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__module_could_not_unify9.snap
```diff
@@ -6,6 +6,34 @@ expression: "fn main() { let assert [1, 2, ..x]: List(String) = [1,2,3] x }"
...
+    String
+
+Found type:
+
+    Int
+
+error: Type mismatch
+  ┌─ /src/one/two.gleam:1:28
+  │
+1 │ fn main() { let assert [1, 2, ..x]: List(String) = [1,2,3] x }
```
lpil: This should not produce any new errors!

compiler-core/src/type_/tests/functions.rs
```diff
@@ -261,33 +261,6 @@ pub fn main() {
...
-        r#"
-pub fn main() {
-  // Since the value is invalid the type is the annotation
-  let a: Int = Junk
-  let b: String = 1
-  let c = a + 2
-}
-"#
-    );
-}
```
lpil, line 289: Do not delete tests please 🙏

abs0luty: Oops my bad, you're right. I somewhy decided it was testing against changes made here!

CHANGELOG.md
```diff
@@ -152,6 +148,18 @@
...
   ([Adi Salimgereyev](https://github.com/abs0luty))
 
+- Let bindings now continue to assume their annotated type after mismatches,
+  preventing misleading follow-up errors:
+  ```gleam
+  pub fn main() {
+    let x: String = 5 // type error: expected String, got Int
+    let y: Int = x    // valid
+    let z: String = x // type error: expected String, got Int
+  }
```
giacomocavalieri, line 66: This example seems to describe the old behaviour where we use the type of the actual value and not the one of the annotation. Could you change it?

Issue #5079, opened 2025-10-26
Confusing message when building with git dependencies without git


sbergen:
When building with git dependencies without git installed, the following error is reported:
```
There was a problem when running the shell command `git`.

The error from the shell command library was:
 
     Could not find the stdio stream
```

This happens e.g. when trying to build a Docker container according to [these instructions](https://gleam.run/deployment/linux-server/).

This looks similar to  #1367, so using `ProcessExecutor::exec` instead of `dependencies::execute_command` would probably fix the issue.

ankddev:
I will work on this!

devstopfix:
In case anyone else gets this error, I received it after forking a repo and adding a git dependency.

I was able to fix inside GitHub action with

```yaml
- run: gleam deps download
        shell: bash # Required for GitHub deps
```

and inside a docker container build with the middle line:

```
FROM erlang:28.1.0-alpine AS build
RUN apk update && apk add git
COPY --from=ghcr.io/gleam-lang/gleam:v1.13.0-erlang-alpine /bin/gleam /bin/gleam
```

PR #5081, opened 2025-10-27
feat: show better error when installing Git deps without Git

ankddev:
Closes https://github.com/gleam-lang/gleam/issues/5079

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -38,7 +38,7 @@ use crate::{
     TreeOptions,
     build_lock::{BuildLock, Guard},
     cli,
-    fs::{self, ProjectIO, get_os},
+    fs::{self, ProjectIO},
     http::HttpClient,
     text_layout::space_table,
 };
@@ -802,7 +802,7 @@ fn execute_command(command: &mut Command) -> Result<std::process::Output> {
         Err(error) => Err(match error.kind() {
             ErrorKind::NotFound => Error::ShellProgramNotFound {
                 program: "git".into(),
-                os: get_os(),
+                os: fs::get_os(),
             },
 
             other => Error::ShellCommand {


compiler-cli/src/dependencies.rs
```diff
@@ -794,21 +795,20 @@ fn provide_local_package(
...
+            ErrorKind::NotFound => Error::ShellProgramNotFound {
+                program: "git".into(),
+                os: get_os(),
+            },
+
+            other => Error::ShellCommand {
+                program: "git".into(),
+                reason: ShellCommandFailureReason::IoError(other),
+            },
+        }),
```
lpil, line 812: The `ShellCommandFailureReason::ShellCommandError(stderr)` bit has been lost! Please put that back 🙏

ankddev: Oh, yeah!

ankddev:
@lpil Fixed!

compiler-cli/src/dependencies.rs
```diff
@@ -37,7 +38,7 @@ use crate::{
     TreeOptions,
     build_lock::{BuildLock, Guard},
     cli,
-    fs::{self, ProjectIO},
+    fs::{self, ProjectIO, get_os},
```
lpil: Do not import functions in an unqualified fashion please 🙏

compiler-cli/src/dependencies.rs
lpil: This hasn't been done! It now considers stderr being used at all a failure, and stderr is discarded when there actually is a failure.

ankddev: @lpil I can't really get it... IMO, it preserves original behaviour. I only have idea to rewrite this as:
```rust
fn execute_command(command: &mut Command) -> Result<std::process::Output> {
    let output = command.output().map_err(|e| {
        if e.kind() == ErrorKind::NotFound {
            Error::ShellProgramNotFound {
                program: "git".into(),
                os: fs::get_os(),
            }
        } else {
            Error::ShellCommand {
                program: "git".into(),
                reason: ShellCommandFailureReason::IoError(e.kind()),
            }
        }
    })?;

    if !output.status.success() {
        let reason = match String::from_utf8(output.stderr) {
            Ok(stderr) => ShellCommandFailureReason::ShellCommandError(stderr),
            Err(_) => ShellCommandFailureReason::Unknown,
        };
        return Err(Error::ShellCommand {
            program: "git".into(),
            reason,
        });
    }

    Ok(output)
}
```
But it seems to be same, but in other form.

PR #5088, opened 2025-10-28
Improve "add omitted labels" code action

giacomocavalieri:
With this the "add omitted label" code action can now be triggered on more function calls, even ones that have some of the labels already written down. This makes it a lot more useful.


lpil (approved):
Wow, tiny fix! Very nice.

Changelog?

giacomocavalieri:
I've already added it to the changelog!

PR #5092, opened 2025-10-29
Make sure renaming a variable in an alternative pattern renames all occurrences

### LINKED ISSUE #5091: Code action "rename variable" does not rename alias of alternative patterns
In the following case:
```gleam
case acc, current {
  A(_), A(_) as repeat
  | B(_), B(_) as repeat
  | C(_), C(_) as repeat
  -> {
    do_something(repeat)
  }
  _ -> panic
}
```

When attempting to rename `repeat`, there are a few inconsistencies:
- The code action is only offered when the one in the first pattern is selected (in this case for `A(_), A(_) as repeat`).
- The code action, when used in the first pattern, is not renaming the other uses of the alias, producing invalid Gleam code.
  The produced code is:
  ```gleam
  case acc, current {
    A(_), A(_) as renamed_alias
    | B(_), B(_) as repeat
    | C(_), C(_) as repeat
    -> {
      do_something(renamed_alias)
    }
    _ -> panic
  }
  ```

giacomocavalieri:
This PR fixes #5091 
This was a bit tricky to get right, I hope the doc comments make it clearer!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,10 @@
 
 ### Bug fixes
 
+- Fixed a bug where renaming a variable from an alternative pattern would not
+  rename all its occurrences.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 - Fixed a typo in the error message emitted when trying to run a module that
   does not have a main function.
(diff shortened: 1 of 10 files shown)


lpil (approved):
Thank you!

PR #5093, opened 2025-10-29
Show an error when renaming with an invalid new name

### LINKED ISSUE #4866: Return more descriptive error message when renaming fails
Currently, when the user tries to rename something in a way that isn't supported, such as trying to rename a constructor to a snake_case name, the language server currently responds with a generic "Rename failed" error message, which doesn't tell the user anything about why renaming failed.
The LSP specification allows us to specify a more detailed error message, which I believe we should do.

giacomocavalieri:
This PR closes #4866 
And with this change it will hopefully be easier to return an error from other language server functions besides the renaming one.

### DIFF
--- a/compiler-core/src/language_server/tests/rename.rs
+++ b/compiler-core/src/language_server/tests/rename.rs
@@ -8,24 +8,28 @@ use crate::language_server::tests::{TestProject, find_position_of};
 
 use super::hover;
 
+/// Returns the rename range and edit to apply if the rename is valid and can be
+/// carried out.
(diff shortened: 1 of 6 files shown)


lpil (commented):
Is this something we can test at all? Seems like we should be able to assert on the message sent out

giacomocavalieri:
Yeah we can absolutely test that! I forgot to add tests though, I'll do that in a second 😁

giacomocavalieri:
I've added the tests, this is ready now!

Issue #5094, opened 2025-10-30
guard expression with variable shadowing brakes type guarantee


EliasDerHai:
There is a bug regarding guard expressions that seems to occur in combination with variable shadowing 

```gleam
import gleam/option.{type Option, Some}
import gleeunit

pub fn main() -> Nil {
  gleeunit.main()
}

pub type Container {
  Container(x: Option(Int), y: Int)
}

pub fn guard_clause_shadowing_bug_test() {
  let x: Option(Int) = Some(999)

  case Some(42) {
    Some(x) if x < 0 -> Container(Some(x), 1)
    _ -> {
      // x is type Option(Int) but at runtime (js) Option is lost 
      // as such `42` is set where a Option(42) should be according to the Container type
      echo x
      Container(x:, y: 0)
    }
  }
}
```

Note that without the first branch - so simply 
```gleam
let Container(x:, y: _) = case x {
    _ -> {
      Container(x:, y: 0)
    }
  }
``` 
 does **not** produce the bug!

- gleam 1.13.0
- target = "javascript"

gleam_log = trace output: 
```
[34mDEBUG[0m [1mdownload_deps[0m[2m:[0m locking_directory [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/build/packages/gleam.lock"
[35mTRACE[0m [1mdownload_deps[0m[2m:[0m reading_file [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/gleam.toml"
[34mDEBUG[0m [1mdownload_deps[0m[2m:[0m reading_manifest_toml
[35mTRACE[0m [1mdownload_deps[0m[2m:[0m reading_file [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/manifest.toml"
[35mTRACE[0m [1mdownload_deps[0m[2m:[0m reading_file [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/build/packages/packages.toml"
[34mDEBUG[0m [1mdownload_deps[0m[2m:[0m locking_directory [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/build/gleam-dev-erlang.lock"
[34mDEBUG[0m [1mdownload_deps[0m[2m:[0m locking_directory [3mpath[0m[2m=[0m"/.../code/backend/reproduction/guard_bug/build/gleam-dev-javascript.lock"
[...]

PR #5097, opened 2025-10-31
Fix guard variable scoping

EliasDerHai:
Fixes [guard expression with variable shadowing brakes type guarantee](https://github.com/gleam-lang/gleam/issues/5094)

This is my first contribution. Please let me know if I messed up. 

I updated 2 snapshots, but don't know if this is fine.

Regarding the change in the decision.rs:

```
            let check_bindings = this.variables.bindings_ref_doc(&check_bindings);
```
is done in the main scope of the function but through `bindings_ref_doc` -> `body_binding_doc` -> `next_local_var` the expression_generator's `current_scope_vars` is mutated - which is reused in the other branches. This leads to the issue that I observed. 

I tried to cover the fix with a minimal test.

### DIFF
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__list_with_guard.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__list_with_guard.snap
@@ -24,8 +24,8 @@ export function go(x) {
     if (first < 10) {
       return first * 2;
     } else {
-      let first$1 = x.head;
-      return first$1;
+      let first = x.head;
+      return first;
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,3 +117,8 @@
 - Fixed a bug where useless comparison warnings for floats compared literal
   strings, claiming for example that `1.0 == 1.` was always false.
   ([fruno](https://github.com/fruno-bulax/))
+
+- Fixed a bug where pattern variables in case clause guards would incorrectly
+  shadow outer scope variables in other branches when compiling to JavaScript.
+  ([Elias Haider](https://github.com/EliasDerHai))
+
--- a/compiler-core/src/javascript/tests/case_clause_guards.rs
+++ b/compiler-core/src/javascript/tests/case_clause_guards.rs
@@ -576,3 +576,30 @@ pub fn main() {
 "#,
     );
 }
+
+// https://github.com/gleam-lang/gleam/issues/5094
+#[test]
+fn guard_pattern_does_not_shadow_outer_scope() {
+    assert_js!(
+        r#"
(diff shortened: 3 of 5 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__list_with_guard.snap
```diff
@@ -24,8 +24,8 @@ export function go(x) {
     if (first < 10) {
       return first * 2;
     } else {
-      let first$1 = x.head;
-      return first$1;
+      let first = x.head;
+      return first;
```
lpil, line 28: Isn't this an error in JavaScript? There is already a `first` defined in this function

EliasDerHai: No it's not an error in js since `let` and `const` are [block-scoped](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block#block_scoping_rules_with_let_const_class_or_function_declaration_in_strict_mode). 

Shadowing doesn't pollute the outer scope - in contrary to `var` which is bad practice exactly for that reason.

I ran a quick test on how my change behaves for "double-guards" - eg. : 

```gleam
pub fn main() {
  let x = 42
  case x {
    x if x < 0 -> x
    x if x > 40 -> x * 2
    _ -> 0
  }
}
```

This previously compiled to: 
```js
export function main() {
  let x = 42;
  let x$1 = x;
  if (x$1 < 0) {
    return x;
  } else {
    let x$2 = x;
    if (x$2 > 40) {
      return x$2 * 2;
    } else {
      return 0;
    }
  }
}
```

with my change compiles to:
```js
export function main() {
  let x = 42;
  let x$1 = x;
  if (x$1 < 0) {
    return x;
  } else {
    let x$1 = x;
    if (x$1 > 40) {
      return x$1 * 2;
    } else {
      return 0;
    }
  }
}
```

Again we see that `x$1` is repeatedly defined (in different scopes). - Fine by js but maybe odd for humans to read. It could very well be though that what we declare as `x` in gleam are different types. Back-tracing declarations from upper scopes for the purpose of potential reuse doesn't make much sense and is probably quite error-prone though.

Each guard produces a declaration as far as I can see. Since expressions within if-conditions aren't possible we're forced to declare in upper scope. if we shadow vars or 'soft-shadow' with counter increment doesn't matter much or?

Do you suggest putting more effort to avoid variable shadowing in js?
Do you see a way how the change could cause a bug?

lpil: Oh yes! Thank you for your thorough explanation. Very silly of me there

PR #5110, opened 2025-11-04
Fix record constructor functions code generation

### LINKED ISSUE #5108: Language server incorrectly warns about comparison being `True`
The language server incorrectly reports that comparing constructor functions is always `True`, when in reality such comparisons always evaluate to `False`.
### Steps to reproduce:
```gleam
type Comparison {
  Wibble
  Wobble(String)
}

pub fn main() {
  echo Wibble == Wibble
  echo Wobble == Wobble
}
```

My IDE shows this warning message on both `echo` lines:

```sh
Redundant comparison

This is always `True`.

This comparison is redundant since it always succeeds.
```
### Running Gleam:
```sh
  Compiling comparison
warning: Redundant comparison
  ┌─ /home/wiskiy/dev/gleamy/comparison/src/comparison.gleam:7:8
  │
7 │   echo Wibble == Wibble
  │        ^^^^^^^^^^^^^^^^ This is always `True`

This comparison is redundant since it always succeeds.

warning: Redundant comparison
  ┌─ /home/wiskiy/dev/gleamy/comparison/src/comparison.gleam:8:8
  │
8 │   echo Wobble == Wobble
  │        ^^^^^^^^^^^^^^^^ This is always `True`

This comparison is redundant since it always succeeds.

   Compiled in 0.26s
    Running comparison.main
src/comparison.gleam:7
True
src/comparison.gleam:8
False
```
### Environment:
- Gleam: v1.13.0
- OS: Arch Linux
- Runtime: Erlang

abs0luty:
- Closes #5108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,24 @@
   containing scientific notation or trailing zeros (i.e. `100` and `1e2`).
   ([ptdewey](https://github.com/ptdewey))
 
+- Code generation for record constructor functions with arguments has been
+  changed. They are now generated as named functions instead of anonymous
+  functions, which allows them to compare equal when referenced. For example:
+
+  ```gleam
+  pub type Foo {
+    Wibble
+    Wobble(String)
+  }
+
+  pub fn main() {
+    echo Wibble == Wibble  // True
+    echo Wobble == Wobble  // False (previously) -> True
+  }
+  ```
+
+  ([Adi Salimgereyev](https://github.com/abs0luty))
+
 ### Build tool
 
 - The help text displayed by `gleam dev --help`, `gleam test --help`, and
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -335,8 +335,20 @@ fn register_imports_and_exports(
             typed_parameters,
             opaque,
             external_erlang,
+            publicity: _,
             ..
         }) => {
+            // Export constructor functions for constructors with arguments
+            // We need to export even for private types since they're used within the module
+            for constructor in constructors {
+                let arity = constructor.arguments.len();
+                if arity > 0 {
+                    let constructor_name = constructor_function_name(name, &constructor.name);
+
+                    exports.push(atom_string(constructor_name).append("/").append(arity));
+                }
+            }
+
             // Erlang doesn't allow phantom type variables in type definitions but gleam does
             // so we check the type declaratinon against its constroctors and generate a phantom
             // value that uses the unused type variables.
@@ -460,8 +472,11 @@ fn module_statement<'a>(
             module_function(function, module, is_internal_module, line_numbers, src_path)
         }
 
+        Definition::CustomType(CustomType {
+            name, constructors, ..
+        }) => module_constructors(name, constructors, module, line_numbers),
+
         Definition::TypeAlias(TypeAlias { .. })
-        | Definition::CustomType(CustomType { .. })
         | Definition::Import(Import { .. })
         | Definition::ModuleConstant(ModuleConstant { .. }) => None,
     }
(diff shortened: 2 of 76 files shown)


giacomocavalieri:
I think this is not the right way to solve this, we should change the code generation for js so that they are indeed the same. I think this is already the case on the erlang target

abs0luty:
We definitely should change the way we handle function comparisons:

```gleam
type A {
  A(String)
}

fn a() {}

pub fn main() {
  echo a == a // True
  echo A == A // False
}
```

I'm unsure what the right criteria for function equality should be. For example, should two identically defined anonymous functions count as equal, or only named functions/constructors? 

My assumption is one would expect memory references to be checked, and as two anonymous functions are distinct objects only functions with the same name would compare as equal.

abs0luty:
I considered manually changing how constructor functions are compared at runtime instead of making a change to constructor call API. However, tracking dynamic code flow at runtime would add unnecessary overhead during compilation. In my opinion, modifying API is the better approach:

```gleam
type Bar {
  Foo(String)
}

fn a() {}

pub fn main() {
  let b = a
  let c = a
  
  // because of modified codegen, d and f are now references 
  // to the same named function associated with Foo
  let d = Foo
  let f = Foo

  echo b == c // True
  echo d == f // False -> True
}
```

lpil:
> we should change the code generation for js so that they are indeed the same. I think this is already the case on the erlang target

For performance reasons we want to introduce singleton values for record constructors and for zero arity variants, which would solve this too.

This bit I'm unsure about is function capturing and anonymous functions. Do we have this issue with them today?

abs0luty:
@giacomocavalieri @lpil I've updated code generation so record constructors are emitted as singletons, giving them the same equality semantics as named functions. I want to know your opinion on the changes and naming strategy I used:

## Erlang

Constructor functions with arguments use `type_name$constructor_name` naming where both parts are snake_case. For example, `Comparison.Wobble` becomes `'comparison$wobble'/1`. These are generated as top-level Erlang functions that construct the appropriate tagged tuple. All constructor functions are unconditionally exported with their arity, even for private types, because they may be used within the module. 

When a constructor is referenced as a value, the compiler generates `fun 'comparison$wobble'/1` for same-module references or `fun module:'comparison$wobble'/1` for cross-module references. The dollar sign separator ensures no collision with user-defined functions since Gleam function names cannot contain dollar signs.

## JavaScript

Constructor functions use `TypeName$ConstructorName` naming where the type name is in PascalCase and the constructor name preserves its original casing. For example, `Comparison.Wobble` becomes `export const Comparison$Wobble = ($0) => new Wobble($0)`. These are exported as named arrow functions at module level. 

When referenced as a value, the compiler simply uses the identifier `Comparison$Wobble` directly instead of wrapping it in an arrow function. 

**The potential problem I see is Gleam users seeing compiler errors due to cached old codegen output when updating to next release. Not sure what to do with this.**

lpil:
Thank you for the detail! We don't want to make any changes to the API of generated Erlang modules, so we cannot add constructor functions. I think we should disable this warning for record constructor functions, the same way we do for anonymous functions.

abs0luty:
> ```gleam
> type A {
>   A(String)
> }
>
> fn a() {}
>
> pub fn main() {
>   echo a == a // True
>   echo A == A // False
> }
> ```

@lpil I think I overstated the risk of people hitting compiler errors due to cached codegen when upgrading.

The change in API of erlang and javascript modules is not breaking _in the sense that it doesn't modify any existing definitions_, it's just adding new functions for constructors? For instance, resolving issues like #5127 (where some unused types should not be emitted) would require such changes.

The PR with minor changes (like if there is any definition for constructor function cause recompiling) will not break any existing code, so I guess it's fine?

lpil:
It's not a breaking change, but is changing how we do code generation in an unplanned way. We do not intend to add new functions to the generated Erlang modules presently.

abs0luty:
> I think we should disable this warning for record constructor functions, the same way we do for anonymous functions.

- @lpil Made a separate PR for that #5163, not sure what to do with this one. I believe I should close it and forget about updated code generation for now?

lpil:
Yup sounds good, thank you!

PR #5116, opened 2025-11-06
Fix bug with aliased prefix patterns referenced in guards

### LINKED ISSUE #5115: Erlang codegen bug when using string patterns in guards
As of #5044 (not yet released), the Erlang code generation for string prefix patterns has changed. However, there is a bug when a string prefix is used in a guard. For example, in the following code:
```gleam
pub fn strip_prefix(string, prefix) {
  case string {
    "!" as p <> rest if p == prefix -> Ok(rest)
    _ -> Error(Nil)
  }
}
```

Now compiles to the following Erlang:
```erlang
strip_prefix(String, Prefix) ->
  case String of
    <<"!"/utf8, rest/binary>> when P =:= Prefix ->
      P = <<"!"/utf8>>,
      {ok, Rest};
    _ -> {error, nil}
  end.
```

Here, the variable `P` is bound in the body of the `case` expression, but it is referenced in the guard `P =:= Prefix` beforehand, which is invalid.

giacomocavalieri:
This PR fixes #5115
With this fix the referenced variable is always replaced with its literal value so the code is still as fast as it can be.

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -7,7 +7,7 @@ mod pattern;
 mod tests;
 
 use crate::build::{Target, module_erlang_name};
-use crate::erlang::pattern::PatternPrinter;
+use crate::erlang::pattern::{PatternPrinter, StringPatternAssignment};
 use crate::strings::{convert_string_escape_chars, to_snake_case};
 use crate::type_::is_prelude_module;
 use crate::{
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!

Do we not need to update the changelog?

giacomocavalieri:
I don't think so, the buggy code was not released yet

PR #5117, opened 2025-11-06
✨ Include constant int segments in interference analysis

### LINKED ISSUE #5013: Compiler could warn for more unreachable bit array patterns
Right now the compiler will not warn for some unreachable bit array patterns. Take this example:
```gleam
case <<>> {
  <<"a", _:bits>> -> 1
  <<97, _:bits>> -> 2
  _ -> 2
}
```
The second branch is actually unreachable since they both match on the same exact byte. However, the compiler is not warning. This would have to be implemented by extending interference analysis on integers (right now it's only concerned with strings)

### LINKED ISSUE #5126: Large case statement takes an unreasonable amount of time and memory to compile
The compiler takes ~26s and ~10GB of ram to compile a [single case statement](https://github.com/Lemorz56/glepack/blob/main/src/glepack/decode.gleam#L25). When I `todo` this function, the compile times drop to sub-second speeds.

## Steps to reproduce
```sh
git clone https://github.com/Lemorz56/glepack
cd glepack
gleam build
```
 
## Logs
[gleam.log](https://github.com/user-attachments/files/23443979/gleam.log)

## Environment
- Gleam: v1.13.0 & v1.11.1
- OS: NixOS 
- Runtime: Erlang

fruno-bulax:
Closes #5013
Closes #5126

This PR extends jak's interference analysis to integer segments. The tricky bit was getting the right bit-representation for integers, plugging it into the analysis seemed to have just worked ✨

One hitch is that, unlike with strings, you can have ludicrously large literal integer segments without typing them out. Take for example this lovely test that tried to allocate a petabyte of memory to represent the segment.
```rust
#[test]
fn javascript_unsafe_int_segment_size_in_pattern() {
    assert_js_warning!(
        r#"
pub fn go() {
  let assert <<0:9_007_199_254_740_992>> = <<>>
}
"#
    );
}
```

To avoid this, I included an arbitrary limit of roughly 8kb beyond which I don't produce the bit pattern.

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -3598,19 +3598,17 @@ fn int_to_bits(
     let mut bytes = int_to_bytes(value, endianness, signed);
     let bytes_size = bytes.len() * 8;
 
-    // There are 3 cases, which are easier to handle separately by endianness
-    // If the size of the bigint bytes equals the expected bits, we can return them as-is
-    // If there are more bits than we need, we need to trim some of the most significant bits.
-    // E.g. `6:3` yields one byte of which we need to trim the 5 most significant bits.
-    // Values like 999:3 are illegal and caught by the guard at the start of the function.
-    // If there are fewer bits than we need, we need to add some.
-    // E.g. `6:13` yields one byte which we need to pad with 5 bits
     let bits = match (endianness, bytes_size.cmp(&size)) {
         (_, Ordering::Equal) => BitVec::from_vec(bytes),
 
+        // Even though we return an error if `value` cannot be represented in `size` bits,
+        // we may need to slice off up a couple bits if the read size is not a multiple of 8.
+        // <3:4> yields the bytes [0b0000_0011]. We slice off 4 bits and get 0011
         (Endianness::Big, Ordering::Greater) => {
             BitVec::from_bitslice(&bytes.view_bits()[bytes_size - size..])
         }
+        // If the number needs fewer bits than the read size, we pad it.
+        // <3:9> yields the bytes [0b0000_0011]. We add one digit and get 0_0000_0011
         (Endianness::Big, Ordering::Less) => {
             let mut bits = BitVec::repeat(pad_digit, size - bytes_size);
             bits.extend_from_raw_slice(&bytes);
@@ -3782,29 +3780,36 @@ fn representable_with_bits(value: &BigInt, bits: u32, signed: bool) -> bool {
     if bits == 0 {
         return false;
     };
-    let is_negative = value.sign() == Sign::Minus;
-    if is_negative && !signed {
-        return false;
-    }
 
-    let bits_unsigned = value.bits();
-    let required_bits = if !signed {
-        bits_unsigned
-    } else if !is_negative {
-        // The minimal representation of an unsigned number always has 1 as its most
-        // significant bit and thus always needs an extra bit for the sign.
-        bits_unsigned + 1
-    } else if let Some(trailing_zeros) = value.trailing_zeros() {
-        // Negative powers of two don't need an extra sign bit. E.g. `-2 == 0b10`
-        let is_power_of_2 = trailing_zeros == bits_unsigned - 1;
-        bits_unsigned + (!is_power_of_2 as u64)
-    } else {
-        // The value is 0
-        1
+    let required_bits = match (value.sign(), signed) {
+        // Zero always needs one bit.
+        (Sign::NoSign, _) => 1,
+
+        // `BigInt::bits` does not consider the sign!
+        (Sign::Plus, false) => value.bits(),
+        // Therefore we need to add the sign bit here: `10` -> `010`
+        (Sign::Plus, true) => value.bits() + 1,
+
+        // A negative number must be signed
+        (Sign::Minus, false) => return false,
+        (Sign::Minus, true) => {
+            let bits_unsigned = value.bits();
+            let trailing_zeros = value
+                .trailing_zeros()
+                .expect("trailing_zeros to return a value for non-zero numbers");
+            let is_power_of_2 = trailing_zeros == bits_unsigned - 1;
+
+            // Negative powers of two don't need an extra sign bit. E.g. `-2 == 0b10`
+            if is_power_of_2 {
+                bits_unsigned
+            } else {
+                bits_unsigned + 1
+            }
+        }
     };
 
     match required_bits.to_u32() {
-        Some(r) => bits >= r,
+        Some(required_bits) => bits >= required_bits,
         None => false,
     }
 }
@@ -3844,6 +3849,18 @@ mod representable_with_bits_test {
         // But not in <= 4 bits as a signed number.
         assert!(!representable_with_bits(&BigInt::from(-9), 4, true));
         assert!(!representable_with_bits(&BigInt::from(-9), 3, true));
+    }
+
+    #[test]
+    fn zero_representable_with_bits_test() {
+        for i in 0..12 {
+            println!("{i}: {}", BigInt::from(i).bits());
+        }
+        assert!(!representable_with_bits(&BigInt::ZERO, 0, false));
+        assert!(!representable_with_bits(&BigInt::ZERO, 0, true));
+
+        assert!(representable_with_bits(&BigInt::ZERO, 1, false));
+        assert!(representable_with_bits(&BigInt::ZERO, 1, true));
     }
 
     #[test]


compiler-core/src/exhaustiveness.rs
```diff
@@ -3558,6 +3572,80 @@ fn segment_matched_value(
...
+        }
+        Endianness::Little => {
+            let mut bytes = value.to_signed_bytes_le();
+            let bytes_size = bytes.len() * 8;
+            match bytes_size.cmp(&size) {
+                Ordering::Greater => {
+                    // First truncate the trailing most significant *bytes*
+                    // E.g. truncate 0xffffff:10 to two bytes (0xffff)
+                    let required_bytes = size.div_ceil(8);
+                    bytes.truncate(required_bytes);
```
fruno-bulax: Actually, I think misunderstood this; truncation happens in bit array literals, but in patterns it will simply not match against anything. Would it be better to return `None` in these cases?

compiler-core/src/type_/tests/warnings.rs
```diff
@@ -4515,6 +4515,34 @@ pub fn wibble(bits) {
...
+pub fn wibble(bits) {
+  case bits {
+    <<0b1:1, _:1>> -> 1
+    <<0b11:2>> -> 2
+    _ -> 3
+  }
+}"#
+    );
+}
+
```
giacomocavalieri, line 4545: These tests are great, there's an additional one that I would like to add just to be extra safe: where the number in the segment is bigger than its size, followed by something that could have the same value:
```gleam
case x {
  <<97:3>> -> 1
  <<"a">> -> 2
  _ -> 3
}
```
In this case the second one should actually be reachable, because the first one will never match

compiler-core/src/exhaustiveness.rs
giacomocavalieri: Yes exactly! The code to perform this check actually already exists in the codebase, it's implemented in the `representable_with_bits` function, so you could add a call to that to the top of the function to take care of this special case

fruno-bulax: The exponentiation of BigInts in that function made some alarm bells go off in my head. I've verified it takes a looooong time (minutes!) for large `bits` (=exponents). I think I'll be able to optimize it though!
I've also realized I may not be handling signed-ness correctly so I'm putting this PR back to a draft until I've cleaned everything up.

compiler-core/src/type_/tests/warnings.rs
fruno-bulax: I added this and double-checked that macro _would_ show multiple warnings if there were any.

compiler-core/src/exhaustiveness.rs
fruno-bulax: I replaced the implementation with one that requires no bigint arithmetic!

I've also replaced the return type of `into_to_bits` with a proper result, so we don't even have to do the check twice.

compiler-core/src/exhaustiveness.rs
```diff
@@ -1212,6 +1213,9 @@ pub enum BitArrayMatchedValue {
         /// is deemed unreachable: it is the location this literal value comes
         /// from in the whole pattern.
         location: SrcSpan,
+        /// The bits representing the given literal integer, with the correct
```
lpil:
```suggestion
        /// The bits representing the given literal int, with the correct
```

CHANGELOG.md
```diff
@@ -45,6 +45,25 @@
   than stopping at the syntax error.
   ([mxtthias](https://github.com/mxtthias))
 
+- The [interference-based pruning](https://gleam.run/news/formalising-external-apis/#Improved-bit-array-exhaustiveness-checking)
+  from 1.13 has been extended to integer segments!
```
lpil:
```suggestion
  from 1.13 has been extended to int segments!
```

compiler-core/src/exhaustiveness.rs
```diff
@@ -3558,6 +3575,92 @@ fn segment_matched_value(
...
+    // E.g. `6:3` yields one byte of which we need to trim the 5 most significant bits.
+    // Values like 999:3 are illegal and caught by the guard at the start of the function.
+    // If there are fewer bits than we need, we need to add some.
+    // E.g. `6:13` yields one byte which we need to pad with 5 bits
+    let bits = match (endianness, bytes_size.cmp(&size)) {
+        (_, Ordering::Equal) => BitVec::from_vec(bytes),
+
+        (Endianness::Big, Ordering::Greater) => {
+            BitVec::from_bitslice(&bytes.view_bits()[bytes_size - size..])
+        }
```
giacomocavalieri, line 3609: Maybe I'm not understanding this properly but shouldn't it be impossible to have a `Greater` here given the previous checks?

compiler-core/src/exhaustiveness.rs
```diff
@@ -3675,18 +3778,35 @@ fn superset(
...
+        bits_unsigned + (!is_power_of_2 as u64)
     } else {
-        // Unsigned numbers range in [0, 2^bits[
-        BigInt::from(0) <= value && value < BigInt::from(2).pow(bits)
+        // The value is 0
+        1
+    };
+
+    match required_bits.to_u32() {
+        Some(r) => bits >= r,
```
giacomocavalieri: Could you use the full name here instead of the single letter?
```suggestion
        Some(required_bits) => bits >= required_bits,
```

compiler-core/src/exhaustiveness.rs
```diff
@@ -3675,18 +3778,35 @@ fn superset(
...
+        return false;
+    }
+
+    let bits_unsigned = value.bits();
+    let required_bits = if !signed {
+        bits_unsigned
+    } else if !is_negative {
+        // The minimal representation of an unsigned number always has 1 as its most
+        // significant bit and thus always needs an extra bit for the sign.
+        bits_unsigned + 1
```
giacomocavalieri: I'm not sure this comment is right, if a number is signed but positive wouldn't the most significant bit be 0 in that case? Isn't it 1 only when a signed number is negative? And the test is checking that the number is indeed positive.

```suggestion
        // The minimal representation of an unsigned number always needs 1 
        // extra bit for the sign.
        bits_unsigned + 1
```

compiler-core/src/exhaustiveness.rs
```diff
@@ -3675,18 +3778,35 @@ fn superset(
...
+    } else if let Some(trailing_zeros) = value.trailing_zeros() {
+        // Negative powers of two don't need an extra sign bit. E.g. `-2 == 0b10`
+        let is_power_of_2 = trailing_zeros == bits_unsigned - 1;
+        bits_unsigned + (!is_power_of_2 as u64)
```
giacomocavalieri: Could you turn this into an `if` expression rather than converting the bool to an integer? Would make the code easier to understand I think:

```suggestion
        let is_power_of_2 = trailing_zeros == bits_unsigned - 1;
        if is_power_of_2 { ... } else { ... }
```

giacomocavalieri:
Looks great! I've left a couple of nits and one question regarding the implementation!

compiler-core/src/exhaustiveness.rs
fruno-bulax: I don't think I understand this comment. Did you misread my comment as `signed` rather than `unsigned` or am I misunderstanding you?

1. In this branch, the number is _signed_ and _non-negative_.
2. `BigInt::bits` returns the minimum number of bits required to represent the number _not considering the sign_. Or in other words, the number of bits required to represent the absolute value as an unsigned integer.
3. For positive numbers, the signed representation needs the `0` sign bits.
4. For zero, `BigInt::bits` returns 0, so we also need to add 1. I failed to make this clear in the implementation.

I agree the implementation is rather confusing! Do you think putting it in a big 'ol match expression makes it clearer?

```rust
#[must_use]
fn representable_with_bits(value: &BigInt, bits: u32, signed: bool) -> bool {
    // No number is representable in 0 bits.
    if bits == 0 {
        return false;
    };

    let required_bits = match (value.sign(), signed) {
        // Zero always needs one bit.
        (Sign::NoSign, _) => 1,

        // `BigInt::bits` does not consider the sign!
        (Sign::Plus, false) => value.bits(),
        // Therefore we need to add the sign bit here: `10` -> `010`
        (Sign::Plus, true) => value.bits() + 1,

        // A negative number must be signed
        (Sign::Minus, false) => return false,
        (Sign::Minus, true) => {
            let bits_unsigned = value.bits();
            let trailing_zeros = value
                .trailing_zeros()
                .expect("trailing_zeros to return a value for non-zero numbers");
            let is_power_of_2 = trailing_zeros == bits_unsigned - 1;

            // Negative powers of two don't need an extra sign bit. E.g. `-2 == 0b10`
            if is_power_of_2 {
                bits_unsigned
            } else {
                bits_unsigned + 1
            }
        }
    };

    match required_bits.to_u32() {
        Some(required_bits) => bits >= required_bits,
        None => false,
    }
}
```

compiler-core/src/exhaustiveness.rs
fruno-bulax: The bytes size may still be larger because it operates on bytes.

For example, for the number 3 we will have the bytes `[0b0000_0011]`. If we have a read size of three, we'd need to slice off the first five bits!

giacomocavalieri: Makes sense, thank you! Could you add this as a comment to have as an explanation for those who might wonder the same thing in future?

compiler-core/src/exhaustiveness.rs
giacomocavalieri: This looks lovely, yeah I think it's a lot easier to understand!

compiler-core/src/exhaustiveness.rs
fruno-bulax: This was technically explained in the little wall of text, but I've reworded it into more digestible chunks at the corresponding match arms.

compiler-core/src/exhaustiveness.rs
fruno-bulax: Done.

compiler-core/src/exhaustiveness.rs
fruno-bulax: pattern matching OP as always.

compiler-core/src/exhaustiveness.rs
fruno-bulax: Done!

fruno-bulax:
I've rebased the PR and addressed the open issues; There's a failing check due to a deprecated image that gears is fixing in #5130

lpil:
Looks good to me! Waiting for Jak's review as he is the expert :)

Issue #5113, opened 2025-11-05
Compiler could warn for more unreachable int and float patterns


fruno-bulax:
Currently in cases like these the second branch is not marked as unreachable:

```gleam
case n {
  10 -> todo
  1_0 -> todo
}

case x {
  1.0 -> todo
  1.00 -> todo
}
```

This happens because the comparison is done on the raw string value rather than a parsed number. Implementers may want to peek at #5001 which tackled the same issue for floats in useless comparison warnings (`1.0 == 1.00`)

There exists a separate issue for bit array patterns: #5013

lpil:
Oh yes! Good call.

abs0luty:
This issue can be closed now due to #5118 I suppose? ☺️

lpil:
Thank you!

PR #5118, opened 2025-11-07
feat: improve exhaustiveness checks for ints and floats

ptdewey:
As noted in #5113, patterns like `10` and `1_0` were incorrectly treated as distinct in exhaustiveness checks, allowing redundant clauses for ints and floats.

I made changes improve exhaustiveness checking, allowing it to correctly detect redundant patterns when integers and floats use different formatting (underscores, scientific notation, trailing decimals). I also added several test cases covering different variants of this situation.

Please reach out with any questions, concerns, or critiques. Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,12 @@
   than stopping at the syntax error.
   ([mxtthias](https://github.com/mxtthias))
 
+- Exhaustiveness checks for ints and floats now correctly handle unreachable
+  cases in which the numbers contain underscores (i.e. `10` and `1_0`).
+  Float exhaustiveness checks also now correctly identify unreachable cases
+  containing scientific notation or trailing zeros (i.e. `100` and `1e2`).
+  ([ptdewey](https://github.com/ptdewey))
+
 ### Build tool
 
 - The help text displayed by `gleam dev --help`, `gleam test --help`, and


compiler-core/src/exhaustiveness.rs
```diff
@@ -443,9 +444,11 @@ pub enum Pattern {
     Discard,
     Int {
         value: EcoString,
```
giacomocavalieri: I think that the only needed change is to remove the string value and only keep the parsed one. Doing so would also simplify the rest of the code to not have to deal with the two representations

compiler-core/src/exhaustiveness.rs
```diff
@@ -443,9 +444,11 @@ pub enum Pattern {
...
+        int_value: BigInt,
     },
     Float {
         value: EcoString,
```
giacomocavalieri: Same for floats!

compiler-core/src/exhaustiveness.rs
```diff
@@ -492,11 +495,11 @@ impl Pattern {
             // out of a branch's checks. So there's no corresponding runtime check
             // we can perform for them.
             Pattern::Discard | Pattern::Variable { .. } | Pattern::Assign { .. } => return None,
-            Pattern::Int { value } => RuntimeCheckKind::Int {
-                value: value.clone(),
+            Pattern::Int { int_value, .. } => RuntimeCheckKind::Int {
+                value: int_value.to_string().into(),
```
giacomocavalieri: Instead of turning it into a string I think we could change the field's type to be a bigint/float

compiler-core/src/exhaustiveness.rs
```diff
@@ -492,11 +495,11 @@ impl Pattern {
...
             },
-            Pattern::Float { value } => RuntimeCheckKind::Float {
-                value: value.clone(),
+            Pattern::Float { float_value, .. } => RuntimeCheckKind::Float {
+                value: float_value.value().to_string().into(),
```
giacomocavalieri: Same for floats!

compiler-core/src/exhaustiveness.rs
```diff
@@ -2484,13 +2487,20 @@ impl<'a> Compiler<'a> {
         &mut self,
         kind: RuntimeCheckKind,
         branch_mode: &BranchMode,
+        pattern: &Pattern,
```
giacomocavalieri: By implementing the other changes mentioned above I think this would no longer be needed

giacomocavalieri:
Amazing, thank you! I've left a couple of comments inline, I'm not sure but I think they could help simplify the implementation :)

ptdewey:
I don't see any comments, did they get published?

giacomocavalieri:
Oh my! I forgot to click the submit review button 😆

compiler-core/src/exhaustiveness.rs
ptdewey: Ah that makes sense, I wasn't completely certain that the string values weren't being used for something else (I probably should have done a bit more digging).

I'll work on making these fixes, thanks for the review!

giacomocavalieri: No worries, totally reasonable to miss that. My approach is very rudimental, I delete the field and see what breaks 😆

ptdewey: Okay, on second pass I think `Pattern` needs to store both values due to the way the data flows through the exhaustiveness system. 

Data flow:
```
TypedPattern (string + numeric)
 > Pattern (string)
  > RuntimeCheckKind (string)
   > RuntimeCheck (string)
    > JS codegen (requires string)
```

Replacing the string in pattern with the numeric value breaks the downstream JS codegen  ([javascript/decision.rs](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/javascript/decision.rs)), which seems to use the string value to preserve the formatting (which is lost if we only pass the numeric value).

I think my original approach is simpler than trying to get around this, since it only requires the additional field in `Pattern` (and an extra string conversion). (Please feel free to correct me if I'm missing/overthinking anything)

giacomocavalieri: I think it would be best to keep only the numeric parsed value and then turn it into a string during codeine when needed.
I don't think maintaining the Gleam formatting in the generated code is that important and also it's not always possible, changing this would also incidentally fix https://github.com/gleam-lang/gleam/issues/5124 I believe

ptdewey: I think that makes sense. I figured preserving the formatting in JS was intended since there a couple of snapshot tests where underscores are expected ~~(although removing the underscores would almost certainly fix that other issue)~~.

ptdewey:
Re this PR potentially addressing #5124: With the changes made here, valid JavaScript is now produced for the case statement referenced in the issue, but not for some other cases.

Case statement now produces valid JS:
```
0 │+----- SOURCE CODE
1 │+
2 │+pub fn main(x) {
3 │+  case x {
4 │+    0._1 -> "bar"
5 │+    _ -> "foo"
6 │+  }
7 │+}
8 │+
9 │+
10 │+----- COMPILED JAVASCRIPT
11 │+export function main(x) {
12 │+  if (x === 0.1) {
13 │+    return "bar";
14 │+  } else {
15 │+    return "foo";
16 │+  }
17 │+}
```

Returning float produces invalid JS (status-quo):
```
0 │+----- SOURCE CODE
1 │+
2 │+pub fn main() {
3 │+  0._1
4 │+}
5 │+
6 │+
7 │+----- COMPILED JAVASCRIPT
8 │+export function main() {
9 │+  return 0._1;
10 │+}
```

Should I try to address the other case here, or is that out of scope of this change? (I'd be happy to look into it either way)

giacomocavalieri:
Let's keep that on a separate PR so this is easier to review, but feel free to take that issue if you feel like it!

PR #5119, opened 2025-11-07
Sort completions by type

### LINKED ISSUE #5058: Type based LS value suggestions
I saw a video on twitter where someone showed off autocompletion in OCaml that went further than "just" names of records, variables and syntax. F.e. auto completing the name of a record and having all of the labels there already. Basically auto completion and a code action combined into just one action.

I think use cases for this will very much be inspired by existing code actions and could shortcut them. For example `case` could let you auto complete a case statement with some paths already filled out.

https://github.com/user-attachments/assets/fa19832b-eba8-4e53-b00e-4ab170c83d3d

Referenced tweet: https://x.com/ekander_/status/1978010164633509969

I am lacking a good term or title for this issue, so feel free to make it more descriptive.

GearsDatapacks:
Closes #5058 

This PR changes the autocomplete provider to prioritise values which match the expected type of the expression being completed.

### DIFF
--- a/compiler-core/src/language_server/completer.rs
+++ b/compiler-core/src/language_server/completer.rs
@@ -128,31 +128,6 @@ impl<'a, IO> Completer<'a, IO> {
         }
     }
 
-    fn match_type(&self, type_: &Type) -> TypeMatch {
-        if let Some(expected_type) = &self.expected_type {
-            // If the type of the value we are completing is unbound, that
-            // technically means that all types match, which doesn't give us
-            // any useful information so we treat it as not knowing what the
-            // type is, which generally is the case.
-            if expected_type.is_unbound() {
-                TypeMatch::Unknown
-            }
-            // We also want to prioritise functions which return the desired type,
-            // as often the user's intention will be to write a function call.
-            else if let Some((_, return_)) = type_.fn_types()
-                && expected_type.same_as(&return_)
-            {
-                TypeMatch::Matching
-            } else if expected_type.same_as(type_) {
-                TypeMatch::Matching
-            } else {
-                TypeMatch::Incompatible
-            }
-        } else {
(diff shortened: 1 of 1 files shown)


compiler-core/src/language_server/completer.rs
```diff
@@ -896,113 +1032,27 @@ fn type_completion(
...
-    }
-}
-
-pub struct LocalCompletion<'a> {
+pub struct LocalCompletion<'a, IO> {
     mod_name: &'a str,
     insert_range: Range,
     cursor: u32,
     completions: HashMap<EcoString, CompletionItem>,
+    completer: &'a Completer<'a, IO>,
```
lpil: It seems not quite right to me that the completion now contains the completer. By this point completion has already taken place, and the simpler completion should not have all the capabilities of a the completer, which it has been given access to.

I think it's only done in order to have access to the `match_type` function, but that could be a free-standing function instead, or belong to some other class.

compiler-core/src/language_server/completer.rs
```diff
@@ -107,6 +124,32 @@ impl<'a, IO> Completer<'a, IO> {
...
+        if let Some(expected_type) = &self.expected_type {
+            // If the type of the value we are completing is unbound, that
+            // technically means that all types match, which doesn't give us
+            // any useful information so we treat it as not knowing what the
+            // type is, which generally is the case.
+            if expected_type.is_unbound() {
+                TypeMatch::Unknown
+            }
+            // We also want to prioritise functions which return the desired type,
+            // as often the user's intention will be to write a function call.
```
lpil: Awesome

lpil:
Sorry Gears, would you mind rebasing on main? 🙏

GearsDatapacks:
Done!

Issue #5124, opened 2025-11-08
💥 Numeric literals with underscore after decimal point


abs0luty:
Gleam accepts underscores as digit separators in numeric literals. When an underscore appears immediately after the decimal point in a float literal (for example `0._1`), the Erlang backend compiles and runs correctly, but the JavaScript backend emits an invalid JS literal and fails at parse time.

```gleam
import gleam/io

pub fn main() {
  let x = 0.1

  case x {
    0._1 -> "bar"
    _ -> "foo"
  }
  |> io.println
}
```

Running this code on erlang target produces:
```
  Compiling test_project
   Compiled in 0.27s
    Running test_project.main
bar
```

Running on the JavaScript target fails during parsing. The generated JS (Node.js v24.5.0) includes:
```
  if (x === 0._1) {
            ^^
```
Node output:
```
SyntaxError: Invalid or unexpected token
    at compileSourceTextModule (node:internal/modules/esm/utils:357:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:108:18)
    at #translate (node:internal/modules/esm/loader:550:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:597:27)
    at async #link (node:internal/modules/esm/module_job:180:19)

Node.js v24.5.0
```

## Expected behavior

Both backends should accept the program and print `bar`. The JS backend should emit a valid JavaScript numeric literal, such as `0.1`, instead of `0._1`.

abs0luty:
- As #5125 was merged, I believe we can close the issue now!

PR #5125, opened 2025-11-08
fix js codegen allowing underscore after decimal

ptdewey:
Addresses #5124

Fixes invalid generation of JavaScript code in cases where floats contain underscores immediately before/after a decimal point (i.e. `1._23`).
The changes made here use the parsed value of the float instead of the string representation in generation code. This does result in semantic float formatting in gleam code no longer being preserved in generated JavaScript.

For case statements, this is fixed as a byproduct of the changes made in #5118. ~~(I'm marking this as a draft until that PR is merged and will add tests for this behavior specifically in a case statement)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -230,3 +230,6 @@
 - Fixed a bug where the formatter would remove the `@deprecated` attribute from
   constants.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fix invalid JavaScript codegen in cases where underscores follow a decimal.
+  ([Patrick Dewey](https://github.com/ptdewey))


ptdewey:
Digging in a bit more, the compiler produces an error (`LexicalErrorType::NumTrailingUnderscore`) if a number has a trailing underscore (i.e. `1000_.1`), but not when there is a preceding underscore. Would it make sense to produce an error for the preceding case as well? (this may be a tangential conversation in relation to this fix)

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__float_scientific_literals.snap
```diff
@@ -20,12 +18,12 @@ pub fn go() {
...
-  -10.01e-0;
-  100.001e523;
-  -100.001e-523;
-  100.001e123_456_789;
-  return -100.001e-123_456_789;
+  0.001;
+  0.01;
+  -1.001;
+  -10.01;
+  inf;
```
lpil: There's no variable called `inf` in scope here!

giacomocavalieri:
> Rather than render the float value to a string could you use the original string value please

But using the original string value wouldn't fix the issue this PR is meant to fix. We're already rendering the float value to a string for patterns instead of using the original string value

lpil:
> But using the original string value wouldn't fix the issue this PR is meant to fix.

Aye, this work is to fix that by changing the output, and the easiest way would be to edit that string.

> We're already rendering the float value to a string for patterns instead of using the original string value

Does that code have the same bugs as this one does? This PR is generating invalid JavaScript currently.

We can use the value here, but we need to ensure that infinity and nan are used in appropriate ways, which requires a bit more work.

ptdewey:
> Does that code have the same bugs as this one does? This PR is generating invalid JavaScript currently.

It may have the same issue, I don't think there is an explicit test against the infinity case. 

> We can use the value here, but we need to ensure that infinity and nan are used in appropriate ways, which requires a bit more work.

When I get a chance later I'll look deeper into the infinity issue.

ptdewey:
Looks like the infinity issue also occurs in the case statements after #5118.

I created a new function for converting from an `f64` to a `Document`, which specifically checks if the value is infinity, and if it is NaN, returning valid JS for each of those cases.
The NaN check is unnecessary with how the function is currently used, but I thought it made sense to add it anyway.

Expressions could likely be handled by modifying the original string, but for consistency I think this approach is cleaner (although there was an alternative version of the exhaustiveness checks in #5118 that still output strings in JS codegen).

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__float_scientific_literals.snap
ptdewey: Great catch, I shouldn't have missed that! 
(Interestingly the node repl autofills `inf` to `infinity`, which is still invalid :thinking:)

lpil:
> Looks like the infinity issue also occurs in the case statements after https://github.com/gleam-lang/gleam/pull/5118.

Ah! What code results in that?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__numbers__inf_float_case_statement.snap
ptdewey, line 1: > > Looks like the infinity issue also occurs in the case statements after #5118.
> 
> Ah! What code results in that?

A very large float, as seen in this test:

```gleam
case x {
  100.001e123_456_789 -> "bar"
    _ -> "foo"
  }
```

lpil: Could you share a full reproduction please? I couldn't reproduce it with that snippet.

giacomocavalieri: This generates invalid code for me with `gleam build -tjs`
```gleam
pub fn main(n: Float) {
  case n {
    100.001e123_456_789 -> "wibble"
    _ -> "wobble"
  }
}
```
The code indeed has that `inf`:
```js
export function main(n) {
  if (n === inf) {
    return "wibble";
  } else {
    return "wobble";
  }
}
```
This happens on main

ptdewey: Yep, the issue is reproducible on main, but I fixed it in f3e449e258d1755b9b205cf645c892da1fbb40e5 in this PR (see below).

compiler-core/src/javascript/expression.rs
```diff
@@ -2353,6 +2353,22 @@ pub fn float(value: &str) -> Document<'_> {
...
+            "-Infinity".to_doc()
+        }
+    } else if value.is_nan() {
+        // NOTE: this case is probably unnecessary, as this function is only
+        // invoked with `LiteralFloatValue` values, which cannot be nan.
+        "NaN".to_doc()
+    } else {
+        value.to_doc()
+    }
+}
```
ptdewey, line 2370: This is fix for the JavaScript infinity special case

PR #5131, opened 2025-11-11
Improve error message for wrong number of subjects

### LINKED ISSUE #5129: Problems with multiple subjects and alternative patterns on case expressions
The following example

```gleam
pub fn f(a, b) {
  case a, b {
    1 | 2, 3 -> 1
    _, _ -> 2
  }
}
```

generates this error

```
error: Incorrect number of patterns
  ┌─ /tmp/a/src/a.gleam:3:5
  │
3 │     1 | 2, 3 -> 1
  │     ^^^^^^^^^^^^^ Expected 2 patterns, got 1

This case expression has 2 subjects, but this pattern matches 1.
Each clause must have a pattern for every subject value.
```

The following example

```gleam
pub fn f(a) {
  case a {
    1 | 2, 3 -> 1
    _ -> 2
  }
}
```

generate this error

```
error: Incorrect number of patterns
  ┌─ /tmp/a/src/a.gleam:3:5
  │
3 │     1 | 2, 3 -> 1
  │     ^^^^^^^^^^^^^ Expected 1 patterns, got 2

This case expression has 1 subjects, but this pattern matches 2.
Each clause must have a pattern for every subject value.
```

I'm using gleam 1.13.0.

fruno-bulax:
I may have jumped the gun with this, but I had too much fun playing with the parser. No harm done if we don't end up going this direction 🙂 

1. Recognize `_ | -> 1` as a syntax error.
2. If a pattern has too many subjects, highlight just the pattern rather than the whole clause.

Closes #5129

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,18 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- When matching the wrong number of subjects, the compiler now pinpoints the
+  error location instead of marking the entire branch.
+  ```
+  case wibble {
+    0, _ -> 1
+    ^^^^ Expected 1 patterns, got 2
+    0 |  -> 1
+      ^ I was expecting a pattern after this
+  }
+  ```
+  ([fruno](https://github.com/fruno-bulax/))
+
 - The performance of `==` and `!=` has been improved for single-variant custom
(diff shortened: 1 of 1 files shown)


giacomocavalieri (approved):
Looks great to me! Could you update the changelog?

PR #5134, opened 2025-11-12
Standardize markdown code blocks with sh and txt

Matvey1109:
Standardize markdown code block syntax:
- Use `sh` instead of `shell` for shell scripts
- Use `txt` instead of `text` for plain text
- Maintains consistency across whole project


compiler-cli/src/new.rs
```diff
@@ -105,7 +105,7 @@ Further documentation can be found at <https://hexdocs.pm/{project_name}>.
 
 ## Development
 
-```sh
+```shell
```
lpil: Did you not need to update tests for this one?

lpil:
The code has been changed from sh to shell, but all projects generated up until this point used the sh, so this would result in inconsistency within the ecosystem. It seems we should standardise on txt and sh because of that. What do you think?

Matvey1109:
I agree with you, rewrote commits for standartize markdown code blocks with txt and sh

PR #5137, opened 2025-11-14
Add 'merge case branches' code action

giacomocavalieri:
With this PR I've implemented the code action to merge consecutive case branches. For now it's quite limited as it only supports branches whose body is a single `todo`; @lpil do you think we should be comparing the ASTs to check if they can be safely merged?

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -136,6 +136,7 @@ const COLLAPSE_NESTED_CASE: &str = "Collapse nested case";
 const REMOVE_UNREACHABLE_CLAUSES: &str = "Remove unreachable clauses";
 const ADD_OMITTED_LABELS: &str = "Add omitted labels";
 const EXTRACT_FUNCTION: &str = "Extract function";
+const MERGE_CASE_BRANCHES: &str = "Merge case branches";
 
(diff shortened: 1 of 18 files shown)


lpil:
> @lpil do you think we should be comparing the ASTs to check if they can be safely merged?

Yes, I think so! Wouldn't be much use otherwise.

giacomocavalieri:
Ok! I'm working on it

giacomocavalieri:
Done! That ended up being quite a bit of boilerplate 😅

compiler-core/src/language_server/tests/action.rs
```diff
@@ -10845,3 +10846,224 @@ fn wibble() -> Nil
...
+        r#"pub fn go(result) {
+  case result {
+    [] | [_, _, ..]-> todo
+    [_] -> todo
+    _ -> 2
+  }
+}"#,
+        find_position_of("[]").select_until(find_position_of("[_]"))
+    );
+}
```
lpil, line 11069: Could you include a test that it is not provided for multiple patterns that define the same variables but of different types please 🙏

lpil (commented):
One request inline!

lpil:
oh I pressed merge by mistake 😵

PR #5139, opened 2025-11-16
🐛 Fix stack overflow in large pattern match

### LINKED ISSUE #4805: Stack overflow when compiling case expression with too many branches
Get a stack overflow when compiling the code in [src.tar.gz](https://github.com/user-attachments/files/21395533/src.tar.gz) (a case expression with 45000 branches)

Versions:
- Gleam: gleam 1.11.0
- OS: openSUSE Tumbleweed Linux 6.15.0-1-default

[build.log](https://github.com/user-attachments/files/21395911/build.log)

### LINKED ISSUE #5138: Stack overflow on large pattern match
The initial investigation happened [in this thread on discord](https://discord.com/channels/768594524158427167/1439376471352148041), but I'll summarize below

When adding [Mörk](https://hexdocs.pm/mork/mork.html) as a dependency, trying to compile a new (blank) project on Windows leads to a stack overflow. The issue is not present on Linux as its stack is 8 times larger (8 MiB).

Stackoverflows don't unwind in Rust, but despite having no idea what I'm doing, I somehow managed to get a stacktrace using `ulimit` to reduce the stack size and [backtrace-on-stackoverflow](https://github.com/matklad/backtrace-on-stack-overflow).
The offending function is [`environment::unify`](https://github.com/gleam-lang/gleam/blob/afcd7e979101c07b00e83e816f1583249a42b8b0/compiler-core/src/type_/environment.rs#L1004).

Meanwhile krig helpfully looked through his Mörk codebase and we determined that [`entities.resolve`](https://git.liten.app/krig/mork/src/branch/main/src/mork/internal/entities.gleam) is the case expression to blame.

Here's the meaningful snippet of the stacktrace (line number are from the `main` branch at time of writing)
```
 378: gleam_core::type_::environment::unify
             at /home/fruno/gleam/compiler-core/src/type_/environment.rs:1004:33
 379: gleam_core::type_::environment::unify
             at /home/fruno/gleam/compiler-core/src/type_/environment.rs:1004:33
 380: gleam_core::type_::environment::unify
             at /home/fruno/gleam/compiler-core/src/type_/environment.rs:1034:61
 381: gleam_core::type_::environment::unify
             at /home/fruno/gleam/compiler-core/src/type_/environment.rs:1004:33
 382: gleam_core::type_::expression::ExprTyper::infer_case
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:2239:33
 383: gleam_core::type_::expression::ExprTyper::infer_or_error
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:490:26
 384: gleam_core::type_::expression::ExprTyper::infer_iter_statements
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:756:49
 385: gleam_core::type_::expression::ExprTyper::infer_statements
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:712:14
 386: gleam_core::type_::expression::ExprTyper::infer_fn_with_known_types::{{closure}}
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:4708:43
 387: gleam_core::type_::expression::ExprTyper::in_new_scope::{{closure}}
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:355:26
 388: gleam_core::type_::expression::ExprTyper::scoped
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:382:40
 389: gleam_core::type_::expression::ExprTyper::in_new_scope
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:354:14
 390: gleam_core::type_::expression::ExprTyper::infer_fn_with_known_types
             at /home/fruno/gleam/compiler-core/src/type_/expression.rs:4649:14
 391: gleam_core::analyse::ModuleAnalyzer<A>::infer_function::{{closure}}
             at /home/fruno/gleam/compiler-core/src/analyse.rs:602:48
 392: gleam_core::type_::environment::Environment::in_new_scope
             at /home/fruno/gleam/compiler-core/src/type_/environment.rs:206:22
 393: gleam_core::analyse::ModuleAnalyzer<A>::infer_function
             at /home/fruno/gleam/compiler-core/src/analyse.rs:575:34
 394: gleam_core::analyse::ModuleAnalyzer<A>::infer_module
             at /home/fruno/gleam/compiler-core/src/analyse.rs:294:63
 395: gleam_core::analyse::ModuleAnalyzerConstructor<A>::infer_module
             at /home/fruno/gleam/compiler-core/src/analyse.rs:184:10
 396: gleam_core::build::package_compiler::analyse
```

fruno-bulax:
Fixes #5138
Fixes #4805

Fix a stack overflow that happens when type checking some very large case expressions. I'll do my best to describe the problem:

To type check a case expression, first the return type is set to be unbound. Afterwards, the return type is unified against each branch:

```gleam
case wibble { // Var(Unbound(0))
  1 -> Ok(1) // Named(Result, [ Var(Link(Named(Int))), Var(Unbound(1)) ])
  // To unify with the unbound type, the inner value of the var is replaced with a link
  // Var(Link(Named(...)))

  // The clause has the same type, just with a different unbound id
  2 -> Ok(2) // Named(Result, [ Var(Link(Named(Int))), Var(Unbound(2)) ])
  // Now that the return type is a complex type, unification recurses into the parameters
  // This means that the Error types will be unified.
  // Now here's the kicker: Just as before, the inner value of the var is replaced with a new link!
  //                                                     vvvvvvvv
  // Var(Link(Named(Result, [ Var(Link(Named(Int))), Var(Link(Var(Unbound(2)))) ])

  // If this repeats thousands of times, we eventually overflow the stack
  // due to the recursive nature of unify. 

  // ... And only when we hit clause that returns Err can we resolve the unbound type parameter
  _ -> Err(":(")
}
```

To avoid accumulating links, we now check the inner value _after_ recursing into a link. If it is another `Var<Link>`, we flatten the nested links into one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,6 +199,10 @@
 - Fixed a typo for the "Invalid number of patterns" error.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a stack overflow when type checking some case expressions with
+  thousands of branches.
+  ([fruno](https://github.com/fruno-bulax/))
+
 - Add a missing BitArray constructor return type in the prelude's TypeScript
   definitions.
   ([Richard Viney](https://github.com/richard-viney))


compiler-core/src/type_/environment.rs
```diff
@@ -995,13 +995,25 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
             Action::Link => {
                 let mut t2 = t2.deref().clone();
                 t2.generalise_custom_type_variant();
+                // This may grow into an infinite linked list... (see next branch)
```
lpil: I think this comment wants to be a comment on the next clause, as it is explaining the behaviour there.

compiler-core/src/type_/environment.rs
```diff
@@ -995,13 +995,25 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
...
+            Action::Unify(t) => {
+                unify(t.clone(), t2)?;
+                // ... therefore we flatten it here if necessary
+                if let Type::Var { type_: var } = t.deref()
+                    && let TypeVar::Link { type_: linked } = var.borrow().deref()
+                {
+                    *type_.borrow_mut() = TypeVar::Link {
+                        type_: linked.clone(),
+                    }
+                }
```
lpil, line 1018: Is this the same as the collapse links function? Any reason not to use that here?

compiler-core/src/type_/environment.rs
```diff
@@ -995,13 +995,25 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
...
                 *type_.borrow_mut() = TypeVar::Link {
                     type_: Arc::new(t2),
                 };
                 Ok(())
             }
 
-            Action::Unify(t) => unify(t, t2),
+            Action::Unify(t) => {
+                unify(t.clone(), t2)?;
+                // ... therefore we flatten it here if necessary
```
lpil: This comment doesn't make sense. Could you replace it with a detailed comment that explains what is doing done here and why please 🙏

lpil (commented):
Thank you!! I've left some notes inline

compiler-core/src/type_/environment.rs
fruno-bulax: It was meant to work together with the other comment, but since this clearly failed I'll try again.

compiler-core/src/type_/environment.rs
fruno-bulax: Ah, I was not aware of that one!

Edit: Apparently it protected me from trying to mutably borrow `type_` while it's already borrowed, investigating...

fruno-bulax: the mutable borrow can actually fail here, as the value may already be borrowed due to recursing into the right-hand side at the start of the function.

I've replaced the panicking borrow with the safe variant.

compiler-core/src/type_/environment.rs
```diff
@@ -1001,7 +1001,23 @@ pub fn unify(t1: Arc<Type>, t2: Arc<Type>) -> Result<(), UnifyError> {
...
 
-            Action::Unify(t) => unify(t, t2),
+            Action::Unify(t) => {
+                unify(t.clone(), t2)?;
+
+                // Note that type_ is always a Link in this branch.
+                // unify may replace t's inner value with another link
+                // (See the Action::Link branch just above)
+                // This can cause the compiler to build up an ever-growing chain of links.
+                // Therefore, we try to collapse the links. However, type_ may already
```
krig: I'm guessing "type_ may already" should be removed here and the next line merged with
this? Or the end of the sentence is missing. :)

fruno-bulax: Aaah, this happens sometimes when I rewrite things too often.
I fixed it now, thank you!

compiler-core/src/type_/environment.rs
lpil: Fantastic, thank you

PR #5141, opened 2025-11-19
🐛 Update manifest entry when dependency source changes

### LINKED ISSUE #5084: `gleam build` sometimes incorrectly verifies dependencies and hits Hex.
I've not worked out what causes this yet but ocassionally my projects end up in a state where `gleam build` will fetch dependencies when run. @yoshi-monster has also observed this behaviour and not been able to determine the cause either.

I've attached the build logs (up to `compile` traces, which i've ommitted) here: [out.txt](https://github.com/user-attachments/files/23174185/out.txt)

This is problematic for workflows that use lustre_dev_tools, vite, radiate, watchexec... to rebuild the project on changes as repeated calls to `gleam build` may trigger Hex's rate limiting forcing builds to fail.

### LINKED ISSUE #5132: Excessive Hex API calls during normal development hitting rate limits
## Problem

Every Gleam command (`gleam test`, `gleam check`, `gleam build`, `gleam run`) re-resolves all dependencies by querying the Hex API, even when `gleam.toml`, `manifest.toml`, and cached packages haven't changed. This causes projects to hit Hex API rate limits during normal development workflows.

## Example

During a typical development session refactoring a medium-sized project (~60 dependencies), we ran:
- `gleam test` ~5 times
- `gleam check` ~8 times  
- `gleam run -m cigogne` (migrations) ~4 times
- `gleam run -m squirrel` (SQL codegen) ~3 times

Each command outputs:
```
INFO download_deps: resolving_versions
INFO download_deps: looking_up_package_release name="gleam_stdlib" version="0.60.0"
INFO download_deps: looking_up_package_release name="mist" version="4.0.7"
[... 60+ more packages ...]
```

**Total:** 20 commands × 60 packages = 1,200+ Hex API calls in ~2 hours

**Result:** `ERROR Failed error=Hex("the rate limit for the Hex API has been exceeded for this IP")`

## Expected Behavior

Gleam should cache dependency resolution results and only re-query Hex when:
- `gleam.toml` dependencies change
- `manifest.toml` doesn't exist or is stale
- Explicit `gleam deps update` command
- Package cache is cleared

## Actual Behavior

Every Gleam command re-resolves the entire dependency graph from Hex, regardless of whether anything changed.

## Possible Contributing Factor

Our project has a local dependency with a relative path in `gleam.toml`:

```toml
[dependencies]
client = { path = "../client" }
```

We haven't verified whether this local dependency is triggering the re-resolution on every run, or if this happens with pure Hex dependencies too. It's possible the presence of a path-based dependency is causing Gleam to invalidate the resolution cache unnecessarily.

## Impact

- Development workflow interrupted by rate limit errors
- Cannot run tests or compile while rate limited
- Wastes Hex API resources
- Forces developers to wait 15-30 minutes between sessions

## Related

Possibly related to #5084 but that issue mentions "sometimes" - this appears to happen on **every** command invocation.

## Steps to Reproduce

1. Create a Gleam project with ~60 transitive dependencies
2. Optionally add a local path dependency: `client = { path = "../client" }`
3. Run `gleam test` multiple times in succession (5-10 times)
4. Observe Hex API rate limit error

## Environment

- **Gleam version:** (run `gleam --version`)
- **OS:** macOS 23.5.0 (Darwin)
- **Erlang/OTP:** 28
- **Project structure:** Workspace with local path dependency
- **Dependencies:** ~60 transitive dependencies from Hex

## Suggested Fix

Cache the resolution result based on hash of `gleam.toml` + `manifest.toml`, similar to how other package managers work:
- npm: `package-lock.json` 
- Cargo: `Cargo.lock`
- Go: `go.sum`

Only re-resolve when these files change or cache is invalidated.

If local path dependencies are causing this, perhaps they should only trigger re-resolution of that specific local package, not the entire dependency tree.

## Trace Logs

Running with `GLEAM_LOG=trace gleam test` shows the resolution happening on every invocation:

```
INFO download_deps: resolving_versions
INFO download_deps: unit_propagation: Id::<String>(0) = 'server'
INFO download_deps: DP chose: Id::<String>(0) = 'server' @ Some(Version { major: 0, minor: 0, patch: 0, pre: [], build: None })
INFO download_deps: add_decision: Id::<String>(0) @ 0.0.0 without checking dependencies
INFO download_deps: unit_propagation: Id::<String>(0) = 'server'
INFO download_deps: DP chose: Id::<String>(1) = 'glisten' @ Some(Version { major: 7, minor: 0, patch: 1, pre: [], build: None })
INFO download_deps: add_decision: Id::<String>(1) @ 7.0.1 without checking dependencies
[... continues for all 60+ packages ...]
INFO download_deps: looking_up_package_release name="gleam_stdlib" version="0.60.0"
INFO download_deps: looking_up_package_release name="mist" version="4.0.7"
INFO download_deps: looking_up_package_release name="pog" version="3.3.0"
[... 60+ API calls to Hex ...]
```

This happens even when:
- `gleam.toml` hasn't been modified
- `manifest.toml` exists and is current
- `build/packages/` contains all cached packages
- Running `gleam test` immediately after a successful `gleam test`

## Reproduction

Minimal reproduction case:

```bash
# First run - should resolve and cache
gleam test

# Second run immediately after - still resolves from Hex
gleam test  

# Check the logs
GLEAM_LOG=trace gleam test 2>&1 | grep "looking_up_package_release" | wc -l
# Shows 60+ lines, indicating 60+ Hex API calls
```

After ~5-10 iterations:
```
ERROR Failed error=Hex("the rate limit for the Hex API has been exceeded for this IP")
```

fruno-bulax:
Fixes #5132
Fixes #5084

When changing the type of source (like from hex to local), the manifest entry was not updated.
This causes gleam to hit hex for version resolution on each operation, quickly running into API limits.

Additionally, the manifest was not updated if there were only changes in requirements but not packages. This could happen if when changing version constraints or adding/removing a direct dependency to a package that is also a transitive dependency.
For this I split the `Resolved` struct up, as some usages _only_ cared about the package changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,11 @@
 
   ([fruno](https://github.com/fruno-bulax/))
 
+- Fixed two bugs that made gleam not update the manifest correctly, causing
+  it to hit hex for version resolution on every operation and quickly reach
+  request limits in large projects.
+  ([fruno](https://github.com/fruno-bulax/))
+
 - The performance of `==` and `!=` has been improved for single-variant custom
   types when compiling to JavaScript. This was done by generating comparison
   code specific to the custom type rather than using the generic equality check


compiler-core/src/manifest.rs
```diff
@@ -599,76 +606,118 @@ impl Resolved {
     }
 }
 
-#[test]
-fn resolved_with_updated() {
+#[cfg(test)]
+mod manifest_update_tests {
```
fruno-bulax, line 642: this diff is very confusing: I just moved the existing test into a module and added the `resolved_with_source_type_change` test below it.

compiler-core/src/manifest.rs
```diff
@@ -554,6 +556,11 @@ impl Resolved {
 
         for new in &new.packages {
             match old.remove(&new.name) {
+                // If the kind of source changed, the packages bear essentially no connection
+                Some(old) if discriminant(&old.source) != discriminant(&new.source) => {
```
lpil: This function is magical and can also result in unexpected behaviour if misused. Please use a normal function with a `match` expression instead 🙏 

```rust
pub fn kind(&self) -> ManifestPackageSourceKind {
  match &self {
    Self::Hex { .. } => ManifestPackageSource::Hex
    // ...
```

fruno-bulax: Trying to be too clever is a habit I'm working to get rid of 😅
Done!

fruno-bulax:
Converting this to a draft until I've addressed a related but different issue: https://github.com/gleam-lang/gleam/issues/5132#issuecomment-3563468844

PR #5144, opened 2025-11-19
Fix formatting of deprecated constants

### LINKED ISSUE #5143: Formatter removes `@deprecated` attribute from constants
Given the following code:
```gleam
@deprecated("some message")
const x = 1
```

The formatter currently rewrites it to:
```gleam
const x = 1
```

This is not correct; the formatter should not remove attributes like this

GearsDatapacks:
Fixes #5143


lpil (approved):
Thank you

PR #4283, opened 2025-02-26
Fix issue #2569 ; Warn of running a deprecated or internal module main function

Lioncat2002:

PR #5146, opened 2025-11-20
Adds Warnings for running a deprecated or an internal module main function

### LINKED ISSUE #2569: Warm of running a deprecated or internal module main function
To enforce that only public APIs are used

kfc35:
A continuation of #4283 
Closes #2569 

To resolve some loose threads from the previous pull request’s comments:
- I took [Louis’s advice to use the `mod_config`](https://github.com/gleam-lang/gleam/pull/4283#discussion_r1976393611) (produced from `find_package_config_for_module` or `root_config`)  to test whether a module is internal, so I removed the new `is_internal` function the previous author put in and was [suggested in the initial task discussion](https://github.com/gleam-lang/gleam/issues/2569#issuecomment-2684715579). Let me know if this is not the right way to go about it and the original comment was correct.
- I did some refactoring to pull out the actual `WarningEmitter` used that manages the counter for the number of warnings [as asked for](https://github.com/gleam-lang/gleam/pull/4283#discussion_r2001379431) and interpreted by me. I think I did this correctly, but if I misinterpreted this request or its usage, please let me know.
- I added snapshot tests for the Warnings, but I did not know if you wanted something more integration-y (i.e. create some test package/dependencies and test for the specific warnings like some of the other tests in the repo).
- The “internal module” warning should not be emitted if the module is in the root package now (if the logic I wrote is correct)

I’m very new to Rust and Gleam, so apologies in advance if something is completely off base.

I would be curious to know how I can test any compiler changes locally and manually with a Gleam project. If there is documentation somewhere, please share.


compiler-cli/src/run.rs
```diff
@@ -122,11 +129,50 @@ pub fn setup(
...
+    match main_function.deprecation {
+        gleam_core::type_::Deprecation::Deprecated { message } => {
+            let warning = Warning::DeprecatedMainFunction {
+                module: module.clone(),
+                message,
+            };
+            // warning_emitter.emit also increments a counter. However, we only want to
+            // increment that counter for root package warnings. For dependencies,
+            // we just print the warning directly to console.
+            match package_kind {
```
kfc35: just highlighting this section to see if this logic is correct re: warning emitter usage

compiler-cli/src/run.rs
```diff
@@ -122,11 +129,50 @@ pub fn setup(
...
+    // Warn if the main function being run has been deprecated
+    match main_function.deprecation {
+        gleam_core::type_::Deprecation::Deprecated { message } => {
+            let warning = Warning::DeprecatedMainFunction {
+                module: module.clone(),
+                message,
+            };
+            // warning_emitter.emit also increments a counter. However, we only want to
+            // increment that counter for root package warnings. For dependencies,
+            // we just print the warning directly to console.
```
lpil: The use of the deprecated is coming from the action of the programmer, not from the dependency, so we do want to count the warning here.

compiler-cli/src/run.rs
```diff
@@ -122,11 +129,50 @@ pub fn setup(
...
+        Rc::new(WarningEmitter::new(console_warning_emitter.clone()));
+    let built = crate::build::main_with_warnings(
+        paths,
+        options,
+        manifest,
+        warning_emitter_with_counter.clone(),
+    )?;
+
+    // Warn if the module being run is internal and NOT in the root package
+    match package_kind {
```
lpil: Could we write this as `if package_kind.is_dependency() && mod_config.is_internal_module(&module)`? 🙏  Bit more compact and easier to read.

lpil (commented):
Great, thank you! If left a note below about the warning emitting which I think would allow us to undo all the changes to how the warning emitter is passed around, making the change simpler.

Sorry this took me so long to review 😓

kfc35:
My latest two commits address your comments! Ready for review again

Hopefully doesn’t take too long that I’ll have to address merge conflicts again 😅

compiler-cli/src/run.rs
```diff
@@ -122,11 +129,40 @@ pub fn setup(
...
+        manifest,
+        warning_emitter_with_counter.clone(),
+    )?;
+
+    // Warn if the module being run is internal and NOT in the root package
+    if package_kind == PackageKind::Dependency && mod_config.is_internal_module(&module) {
+        let warning = Warning::InternalModuleMainFunction {
+            module: module.clone(),
+        };
+        console_warning_emitter.emit_warning(warning);
```
kfc35, line 147: The two different warning emitters are still needed since this case is still handled differently from deprecated main functions, unless this should also just add to the counter?

PR #5147, opened 2025-11-20
Raise error for invalid floats on both targets

giacomocavalieri:
The compiler would raise an error for floating point literals outside of the representable range on the Erlang target. With this PR it raises the same warning on both targets!


fruno-bulax:
Does this not break compatibility since it was previously allowed? Maybe it could be a warning for the javascript target instead?

Either way, could you please update the documentation on `LiteralFloatValue`, since it talks about this overflow into infinity as if it were legitimate.

lpil:
Previously we generated invalid code, so this is a fix. No correctly compiling code would be prevented by this.

PR #5150, opened 2025-11-21
Fix type-checking bug with several mutually recursive functions

### LINKED ISSUE #5149: Invalid program with multiple layers of recursion type checks
```gleam
pub fn collapse(id, events) {
  case events {
    [] -> False
    [_, ..rest] -> collapse_thinking(id, rest)
  }
}

pub fn collapse_thinking(id, events) {
  case events {
    [] -> True
    [1, ..rest] -> collapse(id, rest)
    [_, ..rest] -> collapse_checking(id, rest)
  }
}

pub fn collapse_checking(id, events) {
  case events {
    [] -> True
    [1, ..rest] -> collapse_running(id, rest)
    [_, ..rest] -> collapse_thinking(id, rest)
  }
}

pub fn collapse_running(id, _events) {
  echo "running"
  echo id
  let #(a, b) = id
  echo a
  echo b
  echo "done"
  False
}

pub fn main() {
  echo "start"
  collapse(#("a", "b"), [1,2,1])
  collapse(2, [1,2,1])
}
```
This collapse function should only accept a tuple. The second call in main also type checks, and the program crashes on the line `let #(a,b) = id`

GearsDatapacks:
Fixes #5149 

Seems there was a bug in the `into_dependency_order` function causing some of the functions to be generalised before fully being type-checked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -291,3 +291,7 @@
 - Fixed a bug where the compiler would generate invalid Erlang and TypeScript
   code for unused opaque types referencing private types.
   ([Surya Rose](https://github.com/GearsDatapacks))
+
+- Fixed a bug where the type checker would allow invalid programs when a large
+  group of functions were all mutually recursive.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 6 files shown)


abs0luty:
Brilliant PR!

lpil (approved):
Oops. Manager code from me I presume 😁 

Thank you!!!

PR #5154, opened 2025-11-22
Group typed definitions based on their kind

giacomocavalieri:
This PR introduces a refactoring that I've discussed with Louis.
Right now the compiler represents the typed definitions as a `Vec<TypedDefinition>`. While such a data structure is easy to build, it means that whenever we are looking for a specific definition we have to go over all the definitions, even the ones that are never relevant. That means that in a lot of places we do `filter_map`-ing over definitions like this:
```rs
// Say we're looking for a specific import, we don't care about
// all the other definitions!
definitions.iter.filter_map(|definition| match definition {
  Definition::Import(import) => Some(import),
  _ => None
}).for_each(...)
```
The new representation keeps each kind of definition separate. Having `TypedFunction`s separate from other definitions also means that it will be much easier to add support for tail call optimisation of mutually recursive functions (the piece of work I plan to work on next)!

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -96,7 +96,7 @@ impl TypedModule {
             .find_map(|function| function.find_statement(byte_index))
     }
 
-    pub fn count_definitions(&self) -> usize {
+    pub fn definitions_len(&self) -> usize {
         let TypedDefinitions {
             imports,
             constants,
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__generics__task_typescript.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__generics__task_typescript.snap
@@ -7,6 +7,6 @@ pub type Promise(value)
     pub type Task(a) = fn() -> Promise(a)
 
 ----- TYPESCRIPT DEFINITIONS
-export type Task = () => Promise$<any>;
-
 export type Promise$<I> = any;
+
+export type Task = () => Promise$<any>;
--- a/compiler-core/src/javascript/typescript.rs
+++ b/compiler-core/src/javascript/typescript.rs
@@ -349,15 +349,15 @@ impl<'a> TypeScriptGenerator<'a> {
     fn definitions(&mut self, imports: &mut Imports<'_>) -> Vec<Document<'a>> {
         let mut documents = vec![];
 
-        for type_alias in &self.module.definitions.type_aliases {
-            if let Some(document) = self.type_alias(type_alias) {
-                documents.push(document);
-            }
-        }
-
         for custom_type in &self.module.definitions.custom_types {
             if let Some(mut new_documents) = self.custom_type_definition(custom_type, imports) {
                 documents.append(&mut new_documents);
+            }
+        }
+
+        for type_alias in &self.module.definitions.type_aliases {
+            if let Some(document) = self.type_alias(type_alias) {
+                documents.push(document);
             }
         }
 
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -1827,7 +1827,11 @@ impl<'a> UnqualifiedToQualifiedImportFirstPass<'a> {
         module_name: &EcoString,
         constructor_name: &EcoString,
     ) {
-        self.unqualified_constructor = (self.module.ast.definitions.imports)
+        self.unqualified_constructor = self
+            .module
+            .ast
+            .definitions
+            .imports
             .iter()
             .filter(|import| import.module == *module_name)
             .find_map(|import| {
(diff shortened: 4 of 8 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__generics__task_typescript.snap
```diff
@@ -7,6 +7,6 @@ pub type Promise(value)
     pub type Task(a) = fn() -> Promise(a)
 
 ----- TYPESCRIPT DEFINITIONS
-export type Promise$<I> = any;
-
 export type Task = () => Promise$<any>;
+
+export type Promise$<I> = any;
```
lpil, line 10: Should we put the aliases after the definitions? Might read a bit better.

compiler-core/src/javascript/typescript.rs
```diff
@@ -364,69 +346,48 @@ impl<'a> TypeScriptGenerator<'a> {
...
-    ) -> Vec<Document<'a>> {
-        match definition {
-            Definition::TypeAlias(TypeAlias {
-                alias,
-                publicity,
-                type_,
-                ..
-            }) if publicity.is_importable() => vec![self.type_alias(alias, type_)],
-            Definition::TypeAlias(TypeAlias { .. }) => vec![],
+    fn definitions(&mut self, imports: &mut Imports<'_>) -> Vec<Document<'a>> {
```
lpil, line 349: Looks like this vec is turned immediately into an iterator, so could return one here? Up to you.

compiler-core/src/language_server/code_action.rs
```diff
@@ -1831,49 +1827,41 @@ impl<'a> UnqualifiedToQualifiedImportFirstPass<'a> {
...
+                    })
+            })
     }
 
     fn get_module_import_from_type_constructor(&mut self, constructor_name: &EcoString) {
         self.unqualified_constructor =
-            self.module
-                .ast
-                .definitions
+            (self.module.ast.definitions.imports)
```
lpil: Why are there parens here now?

compiler-core/src/language_server/edits.rs
```diff
@@ -14,20 +14,33 @@ pub fn position_of_first_definition_if_import(
...
         .iter()
-        .min_by(|a, b| a.location().start.cmp(&b.location().start));
-    let import = first_definition.and_then(get_import);
-    import.map(|import| src_span_to_lsp_range(import.location, line_numbers).start)
-}
+        .map(|constant| constant.location)
+        .chain(custom_types.iter().map(|custom_type| custom_type.location))
+        .chain(type_aliases.iter().map(|type_alias| type_alias.location))
+        .chain(functions.iter().map(|function| function.location))
+        .filter(|location| location.lt(&first_import.location))
```
lpil: How come `.lt`?

compiler-core/src/ast.rs
```diff
@@ -71,16 +70,52 @@ impl<Info, Definitions> Module<Info, Definitions> {
...
+        // Statements can only be found inside a module function, there's no
+        // need to go over all the other module definitions.
         self.definitions
+            .functions
             .iter()
-            .find_map(|definition| definition.find_statement(byte_index))
+            .find_map(|function| function.find_statement(byte_index))
     }
+
+    pub fn count_definitions(&self) -> usize {
```
lpil: In Rust I think count means it traverses, this would be `definition_len(gth)`?

lpil (commented):
Few Qs inline!

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__generics__task_typescript.snap
giacomocavalieri: Yeah sure!

compiler-core/src/language_server/code_action.rs
giacomocavalieri: I thought it was formatted a bit nicer, I can remove them though!

lpil: No cheating the formatter ever!

compiler-core/src/language_server/edits.rs
giacomocavalieri: We want to find a definition whose location is _smaller_ than the one of the first import, that means it comes before it and the import is not the first definition in the module (despite being the first import).
I've added it as a comment!

compiler-core/src/javascript/typescript.rs
giacomocavalieri: I couldn't figure out a nice way to make that an iterator from the start, chaining things messed with the borrow checker

compiler-core/src/language_server/edits.rs
lpil: No I mean why the weird `lt` method instead of the usual `<` operator?

compiler-core/src/javascript/typescript.rs
lpil: Rust annoyin

compiler-core/src/language_server/edits.rs
giacomocavalieri: Too Gleam pilled and didn't think I could use operators with custom types

giacomocavalieri:
Something broke when trying to fix a conflict with Gears PR that changed the typescript module, I'll have to look into it tomorrow before this can be merged

giacomocavalieri:
Should be all good now!

PR #5158, opened 2025-11-23
Bugfix: Modules with same original name wrongly being marked as unused

### LINKED ISSUE #5145: False positive `This imported module is never used` when aliasing
```gleam
import bar/foo
import foo as baz

pub fn main() -> Nil {
  foo.go()
  baz.thing()
}
```

When running `gleam check` (or build, or the language server) on this code, it incorrectly states:
```
warning: Unused imported module
  ┌─ /tmp/tmp.0xlJkPoggL/import_shadow/src/import_shadow.gleam:1:1
  │
1 │ import bar/foo
  │ ^^^^^^^^^^^^^^ This imported module is never used

```

See https://github.com/bo0tzz/gleam_reproductions/blob/import-shadow/src/import_shadow.gleam for a minimal project that reproduces this.

seafoamteal:
# Description

This resolves #5145.

Three changes were made:
1. When a module was aliased, the entity for the original name has to be placed in the `Shadowed` layer and not the `Module` layer. This prevents it from shadowing previous imports with the same module name.
2. When registering a module reference from a module access site, the `module_name_to_node` map should not be checked. This map contains full original names of modules (e.g. `wibble/wobble` and `wobble`) and hence isn't guaranteed to point us to the correct node when we have limited information (e.g. just `wobble`). We skip this check and directly look up the entity in the Module layer. The `module_name_to_node` map is only needed/used when registering references from unqualified imports and aliases, since we are guaranteed to have information only about the original name of the module in those cases.
3. This is not directly related to the bugfix, but the `assert_no_warnings!` macro was defined ambiguously. If `src` was any `expr`, then the compiler would not know whether a second module tuple was part of the repeated list of module tuples or `src`. Changing `src` to be a `literal` fixes this.

# Example
## Before

```gleam
import wibble/wobble
import wobble as wob

pub fn main() {
  wobble.wimble()
  wob.womble()
}
```

Here, at `wobble.wimble()`, a lookup in the `module_name_to_node` map would return the top-level `wobble` module that is aliased to `wob`. Hence, no reference to `wibble/wobble` is registered and it is incorrectly marked unused.

A less visible problem is that `wibble/wobble` is added to the `Module` layer as entity with name `wobble`. Then, on the next line, `wobble` is once again added to the `Module` layer with the same name, shadowing the first import.

## After

At `wobble.wimble()`, we look up the `wobble` entity in the `Module` layer instead. This gives us the correct `wibble/wobble` module.

The `Module` layer contains two entities: `wobble` and `wob`. The `Shadowed` layer contains one entity: `wobble`.

### DIFF
--- a/compiler-core/src/reference.rs
+++ b/compiler-core/src/reference.rs
@@ -281,7 +281,12 @@ impl ReferenceTracker {
         // We first record a node for the module being aliased. We use its entire
         // name to identify it in this case and keep track of the node it's
         // associated with.
-        self.register_module(module_name.clone(), module_name.clone(), import_location);
+        self.register_module(
+            module_name.clone(),
+            module_name.clone(),
+            EntityLayer::Shadowed,
+            import_location,
+        );
 
         // Then we create a node for the alias, as the alias itself might be
         // unused!
@@ -289,7 +294,7 @@ impl ReferenceTracker {
         // Also we want to register the fact that if this alias is used then the
         // import is used: so we add a reference from the alias to the import
         // we've just added.
-        self.register_module_reference(module_name.clone());
+        self.register_module_reference_for_unqualified_imports_and_aliases(module_name.clone());
 
         // Finally we can add information for this alias:
         let entity = Entity {
@@ -307,20 +312,30 @@ impl ReferenceTracker {
         );
     }
 
-    pub fn register_module(
+    pub fn register_unaliased_module(
         &mut self,
         used_name: EcoString,
         module_name: EcoString,
         location: SrcSpan,
     ) {
-        self.current_node = self.create_node(used_name.clone(), EntityLayer::Module);
+        self.register_module(used_name, module_name, EntityLayer::Module, location);
+    }
+
+    fn register_module(
+        &mut self,
+        used_name: EcoString,
+        module_name: EcoString,
+        layer: EntityLayer,
+        location: SrcSpan,
+    ) {
+        self.current_node = self.create_node(used_name.clone(), layer);
         let _ = self
             .module_name_to_node
             .insert(module_name.clone(), self.current_node);
 
         let entity = Entity {
             name: used_name,
-            layer: EntityLayer::Module,
+            layer,
         };
 
         _ = self.entity_information.insert(
@@ -344,7 +359,7 @@ impl ReferenceTracker {
             EntityKind::ImportedConstructor { module }
             | EntityKind::ImportedType { module }
             | EntityKind::ImportedValue { module } => {
-                self.register_module_reference(module.clone())
+                self.register_module_reference_for_unqualified_imports_and_aliases(module.clone())
             }
         }
     }
@@ -403,11 +418,19 @@ impl ReferenceTracker {
         _ = self.graph.add_edge(self.current_node, target, ());
     }
 
-    pub fn register_module_reference(&mut self, name: EcoString) {
+    pub fn register_module_reference_for_unqualified_imports_and_aliases(
+        &mut self,
+        name: EcoString,
+    ) {
         let target = match self.module_name_to_node.get(&name) {
             Some(target) => *target,
             None => self.get_or_create_node(name, EntityLayer::Module),
         };
+        _ = self.graph.add_edge(self.current_node, target, ());
+    }
+
+    pub fn register_module_reference(&mut self, name: EcoString) {
+        let target = self.get_or_create_node(name, EntityLayer::Module);
         _ = self.graph.add_edge(self.current_node, target, ());
     }
 
--- a/compiler-core/src/reference.rs
+++ b/compiler-core/src/reference.rs
@@ -51,7 +51,7 @@ pub enum EntityKind {
 /// scope all at once!
 ///
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-enum EntityLayer {
+pub enum EntityLayer {
     /// An entity which exists in the type layer: a custom type, type variable
     /// or type alias.
     Type,
@@ -294,7 +294,7 @@ impl ReferenceTracker {
         // Also we want to register the fact that if this alias is used then the
         // import is used: so we add a reference from the alias to the import
         // we've just added.
-        self.register_module_reference_for_unqualified_imports_and_aliases(module_name.clone());
+        self.register_module_reference_with_module_name(module_name.clone());
 
         // Finally we can add information for this alias:
         let entity = Entity {
@@ -312,16 +312,7 @@ impl ReferenceTracker {
         );
     }
 
-    pub fn register_unaliased_module(
-        &mut self,
-        used_name: EcoString,
-        module_name: EcoString,
-        location: SrcSpan,
-    ) {
-        self.register_module(used_name, module_name, EntityLayer::Module, location);
-    }
-
-    fn register_module(
+    pub fn register_module(
         &mut self,
         used_name: EcoString,
         module_name: EcoString,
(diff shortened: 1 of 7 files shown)


compiler-core/src/reference.rs
```diff
@@ -403,14 +418,22 @@ impl ReferenceTracker {
         _ = self.graph.add_edge(self.current_node, target, ());
     }
 
-    pub fn register_module_reference(&mut self, name: EcoString) {
+    pub fn register_module_reference_for_unqualified_imports_and_aliases(
```
lpil: I don't understand this new name. It seems to say something about the context in which it is used, but not name the function itself, and the meaning is unclear.

Could these two functions be given clearer names please, and documentation comments could be useful 🙏

compiler-core/src/reference.rs
```diff
@@ -307,20 +312,30 @@ impl ReferenceTracker {
         );
     }
 
-    pub fn register_module(
+    pub fn register_unaliased_module(
```
lpil: What does this new name mean? What distinguishes it from the new function with the previous name?

lpil (commented):
Thank you! I've left some questions inline.

compiler-core/src/reference.rs
seafoamteal: You're right, sorry about that. I've given them both better names now, relating to the sort of module name they take as arguments.

compiler-core/src/reference.rs
seafoamteal: I believe I was trying to work around making `EntityLayer` public and changing all the call sites of `register_module`, but it _does_ seem cleaner to avoid the extra wrapper function. I've made the changes necessary to remove it.

seafoamteal:
Hey @lpil, I'm just bumping this because I'm not sure if the inline responses show up as notifications.

compiler-core/src/reference.rs
```diff
@@ -281,15 +281,20 @@ impl ReferenceTracker {
         // We first record a node for the module being aliased. We use its entire
         // name to identify it in this case and keep track of the node it's
         // associated with.
-        self.register_module(module_name.clone(), module_name.clone(), import_location);
+        self.register_module(
+            module_name.clone(),
+            module_name.clone(),
+            EntityLayer::Shadowed,
```
lpil, line 287: This isn't quite right- it's not shadowed.

How about instead we have the `as` node reference the module entity? That seems semantically correct.

compiler-core/src/reference.rs
```diff
@@ -403,14 +409,28 @@ impl ReferenceTracker {
...
 
-    pub fn register_module_reference(&mut self, name: EcoString) {
-        let target = match self.module_name_to_node.get(&name) {
+    /// Register a module reference when we have information about the full
+    /// name of a module e.g. `wibble/wobble` instead of just `wobble`. This
+    /// function can be used when processing imports (aliases, unqualified
+    /// imports) as that is the only time we know the full module name. Since
+    /// full module names are not necessarily the name used to refer to the
+    /// module, this should NOT be used when inferring module access.
+    pub fn register_module_reference_with_module_name(&mut self, module_name: EcoString) {
```
lpil, line 418:
```suggestion
    pub fn register_module_imported(&mut self, module_name: EcoString) {
```

compiler-core/src/reference.rs
```diff
@@ -403,14 +409,28 @@ impl ReferenceTracker {
...
-            None => self.get_or_create_node(name, EntityLayer::Module),
+            None => self.get_or_create_node(module_name, EntityLayer::Module),
         };
         _ = self.graph.add_edge(self.current_node, target, ());
     }
 
+    /// Register a module reference when we only know the name used to refer
+    /// to the module and not the full name of the module. This is the case
+    /// when inferring module access.
+    pub fn register_module_reference_with_used_name(&mut self, used_name: EcoString) {
```
lpil, line 429:
```suggestion
    pub fn register_module_alias_reference(&mut self, used_name: EcoString) {
```

compiler-core/src/reference.rs
```diff
@@ -403,14 +409,28 @@ impl ReferenceTracker {
...
             Some(target) => *target,
-            None => self.get_or_create_node(name, EntityLayer::Module),
+            None => self.get_or_create_node(module_name, EntityLayer::Module),
         };
         _ = self.graph.add_edge(self.current_node, target, ());
     }
 
+    /// Register a module reference when we only know the name used to refer
+    /// to the module and not the full name of the module. This is the case
+    /// when inferring module access.
```
lpil, line 428: The term for the name used to refer to an imported module is an "alias". There's no concept of "name" and "full name" for a module.

compiler-core/src/reference.rs
```diff
@@ -403,14 +409,28 @@ impl ReferenceTracker {
...
     }
 
-    pub fn register_module_reference(&mut self, name: EcoString) {
-        let target = match self.module_name_to_node.get(&name) {
+    /// Register a module reference when we have information about the full
+    /// name of a module e.g. `wibble/wobble` instead of just `wobble`. This
+    /// function can be used when processing imports (aliases, unqualified
+    /// imports) as that is the only time we know the full module name. Since
+    /// full module names are not necessarily the name used to refer to the
+    /// module, this should NOT be used when inferring module access.
```
lpil, line 417: The term for the name used to refer to an imported module is an "alias". There's no concept of "name" and "full name" for a module.

compiler-core/src/reference.rs
```diff
@@ -311,16 +316,17 @@ impl ReferenceTracker {
         &mut self,
         used_name: EcoString,
         module_name: EcoString,
+        layer: EntityLayer,
```
lpil, line 319: This API means that a module could be registered as a type or a value, which isn't right.

lpil (commented):
Nice! Thank you! I think this is nearly ready, I've left some small notes RE semantics.

compiler-core/src/reference.rs
seafoamteal: Oh sorry, I saw the term `module_full_name` in a couple of other places in the codebase and must have gotten confused. Just to clarify, if I import e.g.
```gleam
import lustre/element
```
then `element` is an alias for `lustre/element`?
My understanding was that an alias is something the developer declares explicitly using `as`.

compiler-core/src/reference.rs
seafoamteal: > `as` node reference the module entity
Do you mean a module reference from the alias to the module name? We're already doing that a couple of lines below on line 297.

I agree that it's not quite "shadowed" in the sense that there's another import after it with the same name, but this seems very similar to the reason `EntityLayer::Shadowed` was originally added: to avoid key collisions on `Entity`. It can't be added as a `Module` because that's what led to the bug in the first place — it would incorrectly shadow any previous imports with the same name.

compiler-core/src/reference.rs
seafoamteal: That's a much better name, thank you!

compiler-core/src/reference.rs
seafoamteal: I don't think this name is quite as clear as the other one. I tried it out and reverted it for now as it was confusing to me while I was reading through the code. Would `register_module_reference` be alright instead?

compiler-core/src/reference.rs
seafoamteal: Hmm... yeah, I think that's what I was trying to avoid in my first iteration where there was a wrapper function around it for registering unaliased modules. Would you like me to go back to that implementation?

seafoamteal:
Hi @lpil, sorry about disappearing for a bit there — was bogged down by exams :(
I'm not sure I fully understand your most recent review comments. Could you please clarify?

lpil:
Hi @seafoamteal ! Hope your exams went well.

What comments did you mean, sorry?

seafoamteal:
Hi Louis!

> What comments did you mean, sorry?

The ones in your inline review that I've replied to ^.^

Issue #5152, opened 2025-11-22
cargo-deny fails in CI


fruno-bulax:
cargo-deny currently does not support CVSS 4.0 and thus started failing about an hour ago. See for example [here](https://github.com/gleam-lang/gleam/actions/runs/19594319273/job/56116897969?pr=5141)

I can update it to the latest version once they [fix the issue](https://github.com/EmbarkStudios/cargo-deny/issues/804)

fruno-bulax:
The advisory was changed to a lower CVSS version (see [here](https://github.com/rustsec/advisory-db/pull/2479)) so CI is green again. I'll still open a PR once `cargo-deny` is updated!

lpil:
Thank you! I'll close this issue then, but the PR is very much welcome still.

PR #5161, opened 2025-11-24
👷 Update cargo-deny to support CVSS 4.0

fruno-bulax:
See #5152 for details.

Recently our CI failed for a few hours because the advisory db started publishing some advisories using the CVSS 4.0 standard, which wasn't supported in `cargo-deny` yet.

They quickly converted them back to 3.1, but it's still nice to be ready for when 4.0 does come for real.


lpil (approved):
Thank you!

PR #5162, opened 2025-11-25
Fix bugs with merge case branches code action

giacomocavalieri:
There were a couple of bugs that I didn't catch in the first PR
- no longer allows to merge branches defining variables with the same name but a different type
- allows to merge branches where variables both have unbound types since that will never result in a type error after applying the code action


lpil (approved):
Thank you!!

PR #5165, opened 2025-11-26
Pick better names when generating functions whose arguments are record accesses

giacomocavalieri:
I noticed Rust can do this and thought this is a cute improvement we could have as well!


lpil (approved):
This is cool as heck

PR #5167, opened 2025-11-27
Fix search data generation

### LINKED ISSUE #5083: search_data.json incorrect for indexing with Hex search
Back in Feb the Hex folks changed the format of the file we use for our HTML docs search, so they could index it for the new Hex search. Unfortunately the format doesn't match what their code needs, so the data we have been publishing is not usable by Hex search.

The `doc` field contains the documentation as HTML, but it needs to be as markdown.

We will also need to change the name of `search_data.json`, so they can identify the newer valid files. Perhaps `search-data.json`.

We need to ensure that the search in our HTML docs continues to work with this change.

I wonder if there's value in publishing 2 different JSON files, one for our search, one for theirs. This could help avoid this confusion in future.

GearsDatapacks:
Fixes #5083 
This change doesn't seem to affect our own search system; it still seems to work pretty much identically to before (a few values are swapped around in the order, but not in a way that makes search any worse)
Is there somewhere I should add tests for this?

### DIFF
--- a/compiler-core/src/docs.rs
+++ b/compiler-core/src/docs.rs
@@ -184,13 +184,7 @@ pub fn generate_html<IO: FileSystemReader>(
             content: Content::Text(temp.render().expect("Page template rendering")),
         });
 
-        search_items.push(SearchItem {
-            type_: SearchItemType::Page,
-            parent_title: config.name.to_string(),
-            title: config.name.to_string(),
-            content: escape_html_content(content),
-            reference: page.path.to_string(),
-        })
+        search_items.push(search_item_for_page(&config.name, &page.path, content))
     }
 
     // Generate module documentation pages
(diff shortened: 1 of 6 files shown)


lpil (commented):
Thank you!

It is a bit strange we don't have any tests for this, isn't it? Would adding one be painless?

Was the content field not used anywhere in our code? Just in the Hex code? Let's add a comment to the SearchItem data structure explaining what it contains and where it is used in that case

GearsDatapacks:
Yep, I'll add some tests

The `content` field is used by our search, but we're just searching for words, not parsing the HTML so it shouldn't make a difference as far as I understand it. There was already an inconsistent mix between markdown and HTML in the generated documentation, so clearly our code worked fine, even with that beforehand.

lpil (approved):
Oh that is so much nicer! Thank you very much!

Issue #5168, opened 2025-11-28
"Collapse nested case" produces invalid code when used in a list tail


matiascr:
Currently, the collapsing of nested cases in list tails produces the invalid syntax `..[]` in the result.

```gleam
pub fn main(elems: List(List(Int))) {
  case elems {
    [] -> 0
    [_, ..tail] ->
      case tail {
        [] -> -1
        _ -> 0
      }
    _ -> -1
  }
}
```

produces

```gleam
pub fn main(elems: List(List(Int))) {
  case elems {
    [] -> 0
    [_, ..[]] -> -1
    [_, .._] -> -1
    _ -> -1
  }
}
```

lpil:
Ooops, thank you

abs0luty:
- This issue can be closed now due to #5169 I suppose? ☺️

PR #5169, opened 2025-11-28
Update with fixes to collapsing nested cases with list tails

matiascr:
Addresses https://github.com/gleam-lang/gleam/issues/5168

### DIFF
--- a/compiler-core/src/language_server/code_action.rs
+++ b/compiler-core/src/language_server/code_action.rs
@@ -28,7 +28,6 @@ use ecow::{EcoString, eco_format};
 use im::HashMap;
 use itertools::Itertools;
 use lsp_types::{CodeAction, CodeActionKind, CodeActionParams, Position, Range, TextEdit, Url};
-use regex::Regex;
 use vec1::{Vec1, vec1};
 
 use super::{
@@ -4801,8 +4800,7 @@ impl<'a, IO> PatternMatchOnValue<'a, IO> {
         };
 
         let variable_start = (variable_location.start - clause_location.start) as usize;
-        let variable_end =
-            variable_start + (variable_location.end - variable_location.start) as usize;
+        let variable_end = variable_start + variable_location.len();
 
         let clause_code = code_at(self.module, clause_location);
         let patterns = patterns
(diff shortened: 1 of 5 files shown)


compiler-core/src/language_server/code_action.rs
```diff
@@ -8122,19 +8123,39 @@ impl<'a> CollapseNestedCase<'a> {
...
                     // If the variable is a regular variable we'll have to replace
                     // it entirely with the new pattern taking its place.
+
+                    let variable_span_to_use = if let BoundVariable::ListTail {
+                        location_including_prefix: prefix_and_variable_span,
+                        ..
+                    } = matched_variable
+                    {
+                        // We need to un-nest the list when replacing the tail
+                        let surrounding_brackets = Regex::new(r"(^\[|\]$)").expect("a valid regex");
```
lpil: Can we do this efficiently without regex? It's quite tricky to read this, and we tend not to use it.

If not then the regex will need to be cached in a static variable (currently it is recompiled every time) and it'll need a detailed comment explaining what this does. Thank you.

compiler-core/src/language_server/code_action.rs
```diff
@@ -8122,19 +8123,39 @@ impl<'a> CollapseNestedCase<'a> {
...
+                        new_content = surrounding_brackets
+                            .replace_all(new_content.trim(), "")
+                            .to_string();
+
+                        // We need to also remove the prefix when replacing the variable
+                        // at the tail of the list
```
lpil: What does this mean?

matiascr: Perhaps I could have been more clear... What I meant is that, rather than replacing the span of the variable, we need to use the span from the tail prefix to the end of the variable (prefix_and_variable_span)

lpil: I'm sorry, I don't understand. Could you add a detailed comment with examples please 🙏

lpil:

compiler-core/src/language_server/code_action.rs
```diff
@@ -8264,22 +8263,51 @@ impl<'a> CollapseNestedCase<'a> {
...
+                        // If the selected pattern is not a list, we also replace it normally.
+                        (BoundVariableName::ListTail { .. }, false) => matched_variable_span,
+                        // If the pattern is a list to also remove the list tail prefix.
+                        (BoundVariableName::ListTail { tail_location, .. }, true) => {
+                            // When it's a list literal, we remove the surrounding brackets.
+                            let len = trimmed_contents.len();
+                            if let Some(slice) = new_content.trim().get(1..(len - 1)) {
+                                new_content = slice.to_string()
+                            };
+
```
lpil, line 8298: Fab comments in this function, thanks!

PR #5171, opened 2025-11-30
gleam update should return error if package does not exists

### LINKED ISSUE #4836: Return an error if asked to update a package that does not exist

vshakitskiy:
Fixes #4836.
Continuation of #4887.

The only question is that I don't know how to test ensure_packages_exist_locally. Should I extract it to a standalone function outside of the impl block?


compiler-core/src/error.rs
```diff
@@ -205,6 +205,9 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
     #[error("Packages not exist: {}", packages.iter().join(", "))]
     RemovedPackagesNotExist { packages: Vec<String> },
 
+    #[error("Packages to update not exist: {}", packages.iter().join(", "))]
+    PackagesToUpdateNotExist { packages: Vec<String> },
```
lpil: Use the eco string type please

compiler-cli/src/dependencies/dependency_manager.rs
```diff
@@ -123,6 +125,25 @@ where
         Ok(resolved)
     }
 
+    fn ensure_packages_exist_locally(
+        &self,
+        manifest: &Manifest,
+        packages: Vec<EcoString>,
```
lpil: Take this by reference please 🙏 
```suggestion
        packages: &[EcoString],
```

compiler-cli/src/dependencies/dependency_manager.rs
```diff
@@ -123,6 +125,25 @@ where
         Ok(resolved)
     }
 
+    fn ensure_packages_exist_locally(
+        &self,
```
lpil: Doesn't look like `self` is used, so this should be a module function rather than a method on this struct.

giacomocavalieri (approved):
Looks good to me!

PR #5173, opened 2025-11-30
✨ Allow using a different Hex deployment

### LINKED ISSUE #2817: Allow using hex.pm mirrors!
Using [hex.pm mirrors](https://hex.pm/docs/mirrors) would be useful in situations where [hex.pm](https://hex.pm/) itself is not accessible, like for users in countries under sanctions!

fruno-bulax:
Closes #2817

This PR adds support for hex mirrors using the `HEXPM_REPOSITORY_URL` and `HEXPM_API_URL` environment variables.
If they are set, the respective base URLs will be used read operations like version resolution and fetching of packages. Note that these are only _mirrors_. Fully private repositories are not supported, so package publishing, updating, etc. disregards these variables.

I've manually tested these changes and made sure the only remaining calls to `hexpm::Config::new` happen in the publish, retire, etc. commands, but please let me know if you can think of a way to add automated tests.

Please also let me know of any places where I should document these environment variables!

# (Lack of) Authentication

I initially wanted to implement authentication as well. This would have been easy before, but unfortunately the addition of mirrors complicated things quite a bit. As the repository and api mirror don't necessarily belong together, we would have to juggle up to 3 different authentication tokens ("main" hexpm instance for publishing operations, api mirror, and repository mirror.)

Therefore, I think we would need two additional configuration options to set authentication tokens. Something like `HEXPM_REPOSITORY_MIRROR_API_KEY`. On top of that, `hexpm-rust` requires you pass it two URLs but only a single token, so that would also need to be adapted.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,7 +157,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "serde",
- "syn 2.0.101",
+ "syn",
 ]
 
 [[package]]
@@ -177,9 +177,9 @@ dependencies = [
 
 [[package]]
 name = "asn1-rs"
-version = "0.5.2"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f6fd5ddaf0351dff5b8da21b2fb4ff8e08ddd02857f0bf69c47639106c0fff0"
+checksum = "56624a96882bb8c26d61312ae18cb45868e5a9992ea73c58e45c3101e56a1e60"
 dependencies = [
  "asn1-rs-derive",
  "asn1-rs-impl",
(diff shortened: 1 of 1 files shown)


lpil:
This seems to have extended the scope- it permits the API endpoint to be configured too, which is unrelated to enabling the use of mirrors?

Is your goal here to enable alternative Hex deployments to be used? I think we would need to think more about how the configuration would work in this case.

fruno-bulax:
It was my understanding that the scope grew to encompass alternative deployments of the API endpoint as well.
Mostly based on based on the comments following [this one](https://github.com/gleam-lang/gleam/issues/2817#issuecomment-2118981783) in the original issue.

Please let me know if I've misunderstood! I'm also not entirely convinced this is the best way to go forward — mostly due the authentication complications that I don't think can be adequately addressed with environment variables alone.

lpil:
Ah! Sorry about that, I was not up to date with the issue.

I think this is useful still! @maennchen was just asking for the ability to use a different Hex instance, which this could be the foundation of. I'm not sure if this is enough yet though?

maennchen:
My use case is being able to switch to a different server for Hex entirely. We'll be using that for a security audit on Hex.pm and we'd like to switch to a testing hex server to not pollute production hex.

For that, we'd for sure also need to be able to overwrite the signing public key: https://github.com/gleam-lang/gleam/blob/ba3282809d8fcda3d2d7a7357a8077ab64b9fc9c/compiler-core/src/hex.rs#L15-L24

And we'd need to point to a different server also for publish etc. commands.

fruno-bulax:
During implementation I only had a brief look at what alternative _registries_ would entail. It's certainly a different beast entirely and would require additional discussion.

Both [elixir](https://hexdocs.pm/hex/Mix.Tasks.Hex.Repo.html) and [rust](https://doc.rust-lang.org/cargo/reference/registries.html) go the route of configuring registries via build tool configuration, which I don't think gleam even has right now?

maennchen:
Another option is how rebar solves this. You can specify those options in the `rebar.config`:

```erlang
{hex, [
    {repos, [
        #{
            name => <<"hexpm">>,
            repo_url => <<"https://repo.staging.hex.pm">>,
            api_url => <<"https://staging.hex.pm/api/">>,
            repo_public_key => <<"-----BEGIN PUBLIC KEY-----\n...\n-----END PUBLIC KEY-----\n">>
        }
    ]}
]}.

{rebar_packages_cdn, "https://repo.staging.hex.pm"}.
```

fruno-bulax:
Ah thank you, I forgot about rebar. That's another notch in the "build tool configuration" column!
I think it's best we close this PR and instead open a dedicated discussion about this topic. The final plan may may still end up including similar environment variables to set ad-hoc mirrors, but I wouldn't want us to be stuck with that choice because we rushed it through.

lpil:
Another thing I think we need to do is to restrict local API keys to specific URLs. It would be possible for a malicious actor to configure Gleam to use a Hex endpoint they control, so when the programmer decrypts their API key it would be sent to the malicious fake Hex, where it could be used for anything.

fruno-bulax:
I'll close this PR when and try to compile the requirements into a discussion post.
It'll take me a couple days though, lots of things going before the holidays 😵‍💫

Issue #4771, opened 2025-07-11
application_start_module interferes with gleam run -m command


binajmen:
When `application_start_module` is configured in `gleam.toml`, running a specific module using `gleam run -m <module>` causes the application start module to interfere.

See discord discussion https://discord.com/channels/768594524158427167/1392886040385032245

lpil:
Oh yes! This is a bug, sorry about that.

https://github.com/gleam-lang/gleam/blob/bed57729dfed84c609f5b002c5337a76abcf96ae/compiler-core/templates/gleam%40%40main.erl#L22

We are ensuring the root application is started, but it should be the application that the module being run belongs to.

varnerac:
Would changing the logic in the Erlang template to use `application:get_application/1` on the module work?

lpil:
It'll have to be the name of the application that contains that module rather than module itself.

varnerac:
Yeah, `application:get_application/1` takes a module and returns the application it belongs to. I was thinking of 

```erlang
case application:get_application(Module) of
    {ok, App} -> application:ensure_all_started(App);
    undefined -> ok;
end,
```

instead of something more complicated.

lpil:
I've tried running that in a few projects with various deps and it always returns `undefined` for me. Is there something else we'd need to add to make it work?

J0s3c4rl0s:
Can I pick up where previous efforst have left off and try bringing this to the finish line? @lpil @aayush-tripathi

aayush-tripathi:
@J0s3c4rl0s Sure.

J0s3c4rl0s:
Im a bit inexperienced with erlang/elixir and struggling a bit to create a minimal example to reproduce the bug and test the fix. Could @lpil or @binajmen please point me in the right direction :). 

I assume I can simply set up two modules and set the `application_start_module` to one of them and then try running the other, but this isnt enough to trip up the bug yet 


```
gleam.toml
src
|------ main.gleam 
|------ other.gleam 
```
Running `gleam run -m other` and setting `application_start_module = "main"`. 

Do they need to be split into two separate projects and have one as a dependency? Is there some way to maybe indicate that these should be two "distinct apps"?

lpil:
The module being run must be from a dependency package rather than the root package.

ollie-dot-earth:
> When `application_start_module` is configured in `gleam.toml`, running a specific module using `gleam run -m <module>` causes the application start module to interfere.

[...]

PR #4787, opened 2025-07-16
Fix gleam run -m <module> behaviour

aayush-tripathi:
Closes #4771 

Running a single module with `gleam run -m module` should compile the project (if needed) and invoke module:main/0.

To fix this, I have 

- Added a `use_entrypoint` flag in `run_erlang_command`
- Added unit tests for both cases (with/without -m flag)

PR #5182, opened 2025-12-01
Fix gleam run -m <module> behaviour with OTP entrypoint

J0s3c4rl0s:
Closes https://github.com/gleam-lang/gleam/issues/4771

This is a refactor of the solution in https://github.com/gleam-lang/gleam/pull/4787 to satisfy the CI as well as adding an integration test on top of the unit tests. 

The added test is: 
- Adding two gleam projects: 
  - (base) `test/OTP_run_module` and 
  - (dependency) `test/OTP_run_module_dep`

Base project has an OTP entrypoint and runs module from dependency `cargo run run -m other_mod`. This throws an error with the standard implementation but not with the bugfix.

### DIFF
--- a/compiler-cli/src/run.rs
+++ b/compiler-cli/src/run.rs
@@ -11,6 +11,8 @@ use gleam_core::{
     paths::ProjectPaths,
     type_::ModuleFunction,
 };
+use tracing::field::debug;
+use tracing_subscriber::fmt::format;
 
 use crate::{config::PackageKind, fs::ProjectIO};
 
@@ -92,7 +94,7 @@ pub fn setup(
 
     let use_entrypoint = module.is_none();
     // Determine which module to run
-    let module = module.clone().unwrap_or(match which {
+    let module = module.unwrap_or(match which {
         Which::Src => root_config.name.to_string(),
         Which::Test => format!("{}_test", &root_config.name),
         Which::Dev => format!("{}_dev", &root_config.name),


J0s3c4rl0s:
Oops I accidentally included the build files in one of the commits.... 🤦🏼 


I'll make sure to remove them and probably add a .gitignore file

compiler-cli/src/run.rs
```diff
@@ -51,12 +51,12 @@ pub fn setup(
     arguments: Vec<String>,
     target: Option<Target>,
     runtime: Option<Runtime>,
-    module: Option<String>,
+    module_arg: Option<String>,
```
lpil: Undo this rename please, we don't use abbreviations and this is not one of the arguments. The arguments are in the `arguments` vector.

compiler-cli/src/run.rs
```diff
@@ -174,9 +181,14 @@ fn run_erlang_command(
 
     // gleam modules are separated by `/`. Erlang modules are separated by `@`.
     let module = module.replace('/', "@");
-
+    // Decide what to evaluate
+    let erl_eval = if use_entrypoint {
+        format!("{package}@@main:run({module})")
+    } else {
+        format!("{module}:main(), init:stop()")
+    };
```
lpil, line 192: The entrypoint must always be used! Otherwise the virtual machine does not get booted correctly and the IO systems are not configured as Gleam code expects. Running it like this will cause many or most projects to fail when run.

compiler-cli/src/run.rs
```diff
@@ -91,7 +91,7 @@ pub fn setup(
     let root_config = crate::config::root_config(paths)?;
 
     // Determine which module to run
-    let module = module.unwrap_or(match which {
+    let module = module_arg.clone().unwrap_or(match which {
```
lpil: Do not clone this, instead check if it's `None` before consuming this value.

test/OTP_run_module/gleam.toml
```diff
@@ -0,0 +1,15 @@
+name = "my_app" # <-
+version = "1.0.0"              # <-
+description = "It's very cool" # <-
```
lpil: What are these comments with arrows in them?

CHANGELOG.md
```diff
@@ -211,6 +211,14 @@
 
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- `gleam run -m <module>` now calls the requested module’s `main` function
+  directly rather than first starting the project’s `erlang.application_start_module`.
+    
+  This prevents crashes when the start module isn’t meant to be run in that
+  context.
+
+  ([Aayush Tripathi](https://github.com/aayush-tripathi))
```
lpil: This is a bug fix, so a changelog entry would be in the bug fix section 🙏

compiler-cli/src/run.rs
```diff
@@ -400,24 +412,99 @@ fn get_or_suggest_main_function(
...
-    for mod_name in ["valid", "valid/name", "valid/mod/name"] {
-        assert!(is_gleam_module(mod_name));
+    #[test]
+    fn valid_module_names() {
+        for mod_name in ["valid", "valid/name", "valid/mod/name"] {
+            assert!(is_gleam_module(mod_name));
+        }
+    }
+
+    // Unit tests for #4771 (application_start_module interferes with gleam run -m command)
```
lpil: These tests add inertia to the internal behaviour instead of to the public behaviour that we want to solidify, they're a bit hard to follow, and the assertion in them isn't very clear or precise. I think the integration test you've added will work very nicely, let's remove these and keep that.

test/OTP_run_module/gleam.toml
```diff
@@ -0,0 +1,15 @@
...
+
+[dependencies]
+gleam_stdlib = ">= 0.18.0 and < 2.0.0"
+gleam_erlang = ">= 0.2.0 and < 2.0.0"
+other_mod = {path = "../OTP_run_module_dep"}
+gleam_otp = ">= 1.2.0 and < 2.0.0"
+mist = ">= 5.0.3 and < 6.0.0"
+
+[erlang]
+application_start_module = "main" # <-
```
lpil: I think you may have forgotten to commit the `main.gleam` file that defines the application callbacks? Once added let's also rename it to match the name of this package.

test/OTP_run_module/gleam.toml
```diff
@@ -0,0 +1,15 @@
...
+description = "It's very cool" # <-
+
+target = "erlang"
+
+[dependencies]
+gleam_stdlib = ">= 0.18.0 and < 2.0.0"
+gleam_erlang = ">= 0.2.0 and < 2.0.0"
+other_mod = {path = "../OTP_run_module_dep"}
+gleam_otp = ">= 1.2.0 and < 2.0.0"
+mist = ">= 5.0.3 and < 6.0.0"
```
lpil: These deps seem to be unused. Please remove all unused deps as they greatly increase how long this test takes to run.

test/OTP_run_module/Makefile
```diff
@@ -0,0 +1,3 @@
+test:
```
lpil: This test doesn't get run on CI or run automatically as part of the Rust tests! Instead of this ad-hoc makefile based test how about converting it to be a case in the `test-output` test crate?

test/OTP_run_module/gleam.toml
```diff
@@ -0,0 +1,15 @@
+name = "my_app" # <-
+version = "1.0.0"              # <-
+description = "It's very cool" # <-
+
+target = "erlang"
+
+[dependencies]
+gleam_stdlib = ">= 0.18.0 and < 2.0.0"
+gleam_erlang = ">= 0.2.0 and < 2.0.0"
+other_mod = {path = "../OTP_run_module_dep"}
```
lpil: Put the dependency inside this package's directory please.

test-output/cases/otp_root_run_dependency/external_dependency/src/external_dependency.gleam
```diff
@@ -0,0 +1,5 @@
+import gleam/io
+
+pub fn main() {
+  io.println("Hello from other_mod!")
+}
```
lpil, line 5: `gleam.toml` says there is an application start callback in this module, but none has been defined. The test should fail because of this, but the implementation at the moment doesn't boot the project correctly, so it is never caught.

Makefile
```diff
@@ -26,6 +26,7 @@ test: ## Run the compiler unit tests
 	cd test/hextarball && make test
 	cd test/running_modules && make test
 	cd test/subdir_ffi && make
+	cd test/OTP_run_module && make
```
lpil: No capital letters in file names please

test/OTP_run_module/Makefile
J0s3c4rl0s: I see, so thats where the snapshot tests are. 

I will take a closer look but I assume I can add a case and specify what the CLI args are in the `echo.rs` file (in Rust that is)?

Or just run the rust function directly as opposed to having to go through the CLI?

test/OTP_run_module/gleam.toml
J0s3c4rl0s: Inside the `OTP_run_module` directory? Sure, I assumed the dependency needed to be outside the root, I will make sure to move it.

test-output/cases/otp_root_run_dependency/external_dependency/src/external_dependency.gleam
J0s3c4rl0s: I assume this should also be a problem in the base module `OTP_run_module` since it also has the `application_start_module` set? 

Ill add some boilerplate for this, I used to have it and removed it since it didnt trip up any bugs.

compiler-cli/src/run.rs
J0s3c4rl0s: Agreed, if we need to build a temporary directory we might as well just do an integration test. Removed them.

Makefile
J0s3c4rl0s: Okay, are you happy with the name though, assuming I convert it to lowercase?

compiler-cli/src/run.rs
J0s3c4rl0s: Does it? 

I thought the whole point was that if we run a module outside of the current package we arent supposed to go through the entrypoint since this causes the bug? 

Or is this moreso that if we happen to run a module within this package then it will try to run a module before the VM has been booted (and thus fail)?

Does this relate to the comments you made in the issue?

test/OTP_run_module/Makefile
lpil: That sounds good! I think it would be good to rename it from "echo" as the code doesn't have anything specific to do with `echo`.

Makefile
lpil: `with_root_otp_application_run_dependency` or something like that perhaps?

compiler-cli/src/run.rs
lpil: No, the entrypoint must be run otherwise the application does not get booted.

The entrypoint need to be made to boot the correct OTP application for the module being run https://github.com/gleam-lang/gleam/issues/4771#issuecomment-3061859908

J0s3c4rl0s: Okay so if I have main package A and dependency B if I run `gleam run -m B.module` then the CLI should boot the application B (with its entrypoint) rather than the application A (with its entrypoint), right? 


P.S. Do you have any projects that will fail with this setting so I could use them as an example to compare/experiment with?

Makefile
J0s3c4rl0s: How about 

- `otp_root_application_run_dependency`
- `otp_app_run_dependency` 
- `otp_root_run_dependency`

test/OTP_run_module/Makefile
J0s3c4rl0s: Done! Moved it to the rust tests, and changed the name to "snapshot" as well as made minor changes to make the helper method(s) not so specialized to echo.

compiler-cli/src/run.rs
lpil: That's right!

> P.S. Do you have any projects that will fail with this setting so I could use them as an example to compare/experiment with?

If you make a project and define an application start module for it then it will fail when the dependency is run.

For the case where the dependency has a start module you can make it start a named process and have the function run by `gleam run -m` assert that it is alive.

lpil:

PR #5185, opened 2025-12-05
Add qualify/unqualify code action on constants

### LINKED ISSUE #5180: Unqualify is not available in constants
```gleam

type Result(a) {
  Result(expected: a, actual: Option(a))
}

const zero = Result(0, option.None)
//                       ^
```

Does not offer any code actions.

~ 💜

vshakitskiy:
Closes #5180.

### DIFF
--- a/compiler-core/src/language_server/tests/action.rs
+++ b/compiler-core/src/language_server/tests/action.rs
@@ -4031,7 +4031,7 @@ const mapper = list.map
     assert_no_code_actions!(
         title,
         TestProject::for_source(src).add_hex_module("list", "pub fn map(list, f) { todo }"),
-        find_position_of("map").to_selection(),
+        find_position_of("list.map").to_selection(),
     );
 }
 


compiler-core/src/ast/visit.rs
```diff
@@ -742,7 +742,7 @@ fn visit_typed_constant_string_concatenation<'a, V: Visit<'a> + ?Sized>(
     v.visit_typed_constant(right);
 }
 
-fn visit_typed_constant_var<'a, V: Visit<'a> + ?Sized>(
+pub fn visit_typed_constant_var<'a, V: Visit<'a> + ?Sized>(
```
GearsDatapacks, line 745: What's the reason you made this public? As far as I understand it, trait functions aren't meant to be public like this, at least none of the other methods of this trait are.
I don't actually know what making a trait function public means

compiler-core/src/language_server/code_action.rs
```diff
@@ -1696,6 +1696,65 @@ impl<'ast, IO> ast::visit::Visit<'ast> for QualifiedToUnqualifiedImportFirstPass
...
+        location: &'ast SrcSpan,
+        module: &'ast Option<(EcoString, SrcSpan)>,
+        name: &'ast EcoString,
+        constructor: &'ast Option<Box<ValueConstructor>>,
+        type_: &'ast Arc<Type>,
+    ) {
+        let range = src_span_to_lsp_range(*location, self.line_numbers);
+        if overlaps(self.params.range, range)
+            && let Some((module_alias, _)) = module
+            && let Some(import) = self.get_module_import(module_alias, name, ast::Layer::Value)
```
GearsDatapacks, line 1749: I think we need a check somewhere that this is a record, because we don't want to offer to unqualify functions or other constants. Could you add some tests for that too?

compiler-core/src/language_server/tests/action.rs
```diff
@@ -4373,6 +4443,58 @@ pub fn example() {
...
+"#;
+    let title = "Qualify None as option.None";
+    assert_code_action!(
+        title,
+        TestProject::for_source(src)
+            .add_hex_module("option", "pub type Option(v) { Some(v) None }"),
+        find_position_of("None").nth_occurrence(2).to_selection(),
+    );
+}
+
```
GearsDatapacks, line 4575: Can we get tests for:
- Qualifying unqualified functions
- Qualifying unqualified constants
- Unqualifying a value in a constant also unqualifies references in functions
- Unqualifying a value in a function also unqualifies references in constants

compiler-core/src/ast/visit.rs
vshakitskiy: I made `visit_typed_constant_var` as well as `visit_typed_constant_record` public because I saw some other `visit_typed_*` are public and used when overriding a trait method and needing to continue the default traversal afterward (like on line 2416 or 2481 inside `code_action.rs`). So I tried to match the existing pattern and made them public to call them. I dont really now what's the better approach if its not the correct way 🙏

vshakitskiy: I made visit_typed_constant_var as well as visit_typed_constant_record public because I saw some other visit_typed_* are public and used when overriding a trait method and needing to continue the default traversal afterward (like on line 2416 or 2481 inside code_action.rs). So I tried to match the existing pattern and made them public to call them. I dont really now what's the better approach if its not the correct way 🙏

vshakitskiy:
Added record check, as well as more tests, lemme know if I missed something or did something not quite right 🙏

compiler-core/src/ast/visit.rs
GearsDatapacks: Oh sorry! I misread the code. I thought these were methods inside the trait itself, not top-level functions. This is the correct approach

compiler-core/src/language_server/tests/action.rs
```diff
@@ -3949,6 +3949,154 @@ import option.{type Option}
...
+import list
+
+const mapper = list.map
+"#;
+
+    let title = "Unqualify list.map";
+    assert_no_code_actions!(
+        title,
+        TestProject::for_source(src).add_hex_module("list", "pub fn map(list, f) { todo }"),
+        find_position_of("map").to_selection(),
```
GearsDatapacks: Can you change this to find the position of `list.map` instead? I believe this will currently find the position of the `map` in `mapper`, therefore testing the wrong location.

GearsDatapacks (commented):
Great! One tiny note

compiler-core/src/language_server/tests/action.rs
vshakitskiy: ops, didn't notice that, sorry! FIxed now

PR #5187, opened 2025-12-05
✨  Improve ordering of missing patterns

### LINKED ISSUE #5184: Make "Add missing patterns" code action insert variants in order of definition.
Currently when using the "Add missing patterns" code action, the missing patterns are inserted in alphabetical order:

```gleam
// The variants are defined as such:
type Wibble {
  B
  A
}

// Yet the code action yields:
case wibble {
  A -> todo
  B -> todo
}
```

In my opinion it would be better if they were added in the same order in which they are defined.
Right now I often find myself manually reordering them, as I group "related" variants together to make the code clearer.

According to Gears:
> The "missing patterns" error was sorted alphabetically in order to make testing more deterministic. The "add missing patterns" code action takes data from that error, so it's sorted alphabetically

It seems like the easiest solution might be to order the change the ordering in the error. Do you think that is acceptable or should we laser-target the code action?

fruno-bulax:
Closes #5184.

The implementation of this did _not_ go as expected.

In short, we naturally already construct the missing patterns in a very intuitive order.
However, there can be duplicate patterns involving wildcards, which is probably why the strings were put in a set, losing that order. To then _regain_ a stable order, the patterns were sorted lexicographically.

This PR maintains the original order and keeps the set only to prune duplicates.

As you can see from the updated snapshots, this has the nice side effect of also ordering list patterns by the number of elements being matched on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,12 @@
   request limits in large projects.
   ([fruno](https://github.com/fruno-bulax/))
 
+- Missing patterns in error messages and the "Add missing patterns" code action
+  are no longer sorted lexicographically. Instead, they now consider the order
+  in which variants were defined. As programmers often group "related" variants
+  together, this should mean less reshuffling after inserting missing patterns!
+  ([fruno](https://github.com/fruno-bulax/))
+
 - The performance of `==` and `!=` has been improved for single-variant custom
   types when compiling to JavaScript. This was done by generating comparison
   code specific to the custom type rather than using the generic equality check


giacomocavalieri:
Oh wow now list patterns, bools and results are nicely ordered. Love this 💖

Just left a comment for some clarification

compiler-core/src/exhaustiveness/missing_patterns.rs
```diff
@@ -113,7 +111,17 @@ impl<'a, 'env> MissingPatternsGenerator<'a, 'env> {
                     _ = mapping.insert(step.variable().id, index);
                 }
                 let pattern = self.print_terms(mapping);
-                _ = self.missing.insert(pattern);
+
+                // There may be duplicates, but the enumeration of patterns is ordered
+                // Therefore, it is sufficient to only check the previous pattern!
```
giacomocavalieri: How do we know this is always true? It would be nice to either have an explanation. Or we could change the implementation to keep a set of "seen" patterns to know if one should not be added again while building this list

fruno-bulax: Fair, a set will be more robust if I missed some edge case and I'm not worried about performance here.

fruno-bulax: Let me know if you have a better name than "seen_missing" 😅

lpil: If we keep a set then perhaps it wants to be of the hashes of the values, not clones of the values themselves? Doesn't seem like we need two copies. A better name would be good for sure, it's a bit confusing.

Alternatively we could use a set that preserves insertion order. The `indexmap` package is already a dependency, and it provides one.

```
louis ~/src/gleam/gleam (main) $ cargo tree -i indexmap
indexmap v2.11.4
├── petgraph v0.7.1
│   └── prost-build v0.13.5
│       [build-dependencies]
│       └── hexpm v5.0.0
│           ├── gleam-cli v1.13.0 (/Users/louis/src/gleam/gleam/compiler-cli)
│           │   ├── gleam v1.13.0 (/Users/louis/src/gleam/gleam/gleam-bin)
│           │   └── test-output v1.13.0 (/Users/louis/src/gleam/gleam/test-output)
│           ├── gleam-core v1.13.0 (/Users/louis/src/gleam/gleam/compiler-core)
│           │   ├── gleam-cli v1.13.0 (/Users/louis/src/gleam/gleam/compiler-cli) (*)
│           │   ├── gleam-wasm v1.13.0 (/Users/louis/src/gleam/gleam/compiler-wasm)
│           │   ├── test-helpers-rs v0.1.0 (/Users/louis/src/gleam/gleam/test-helpers-rs)
│           │   │   ├── test-output v1.13.0 (/Users/louis/src/gleam/gleam/test-output)
│           │   │   ├── test-package-compiler v1.13.0 (/Users/louis/src/gleam/gleam/test-package-compiler)
│           │   │   └── test-project-compiler v1.13.0 (/Users/louis/src/gleam/gleam/test-project-compiler)
│           │   ├── test-output v1.13.0 (/Users/louis/src/gleam/gleam/test-output)
│           │   ├── test-package-compiler v1.13.0 (/Users/louis/src/gleam/gleam/test-package-compiler)
│           │   └── test-project-compiler v1.13.0 (/Users/louis/src/gleam/gleam/test-project-compiler)
│           └── gleam-wasm v1.13.0 (/Users/louis/src/gleam/gleam/compiler-wasm)
├── petgraph v0.8.3
│   └── gleam-core v1.13.0 (/Users/louis/src/gleam/gleam/compiler-core) (*)
├── priority-queue v2.3.1
│   └── pubgrub v0.3.0
│       ├── gleam-cli v1.13.0 (/Users/louis/src/gleam/gleam/compiler-cli) (*)
│       ├── gleam-core v1.13.0 (/Users/louis/src/gleam/gleam/compiler-core) (*)
│       └── hexpm v5.0.0 (*)
├── pubgrub v0.3.0 (*)
└── toml_edit v0.22.26
    └── gleam-cli v1.13.0 (/Users/louis/src/gleam/gleam/compiler-cli) (*)
```

fruno-bulax: A set that respects insertion order was actually my first instinct, I just didn't wanna add the linked hash map dep for such a small case.

Since we already have one, I would just go with that. I've added it as a direct dependency to `compiler-core`, please let me know if you think it should be a workspace dependency instead — I'm not really sure on what's best practice here.

lpil: Sounds good to me 👍

compiler-core/src/language_server/tests/snapshots/gleam_core__language_server__tests__action__add_missing_patterns_list.snap
```diff
@@ -20,10 +20,10 @@ pub fn main() {
...
     [] -> todo
-    [_, _, _, _, _, ..] -> todo
-    [_, _, _, _] -> todo
-    [_, _, _] -> todo
-    [_, _] -> todo
     [_] -> todo
+    [_, _] -> todo
+    [_, _, _] -> todo
+    [_, _, _, _] -> todo
+    [_, _, _, _, _, ..] -> todo
```
lpil, line 27: this is so pretty now

PR #5194, opened 2025-12-09
Remove unreferenced insta snapshots

abs0luty:
- While working on #5061, updating test names and trying to remove old snapshots, I found 11 not related insta snapshot files that are no longer referenced by any tests. 

This pull request removes these unused snapshots ~~and adds a CI check to prevent unreferenced snapshots from being introduced in the future~~.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,14 @@ test: ## Run the compiler unit tests
 	cd test/running_modules && make test
 	cd test/subdir_ffi && make
 
+.PHONY: insta-check-unused
+insta-check-unused: ## Check for unused cargo insta snapshots
+	cargo insta test --unreferenced=warn
+
+.PHONY: insta-fix-unused
+insta-fix-unused: ## Remove unused cargo insta snapshots
+	cargo insta test --unreferenced=delete
+
 .PHONY: language-test
 language-test: ## Run the language integration tests for all targets
 	cd test/language && make


.github/workflows/ci.yaml
```diff
@@ -346,6 +346,48 @@ jobs:
...
+        uses: oven-sh/setup-bun@v2
+
+      - name: Install cargo-insta
+        run: cargo install cargo-insta --locked
+
+      - name: Install Gleam
+        run: cargo install --path gleam-bin --debug --locked
+
+      - name: Check for unreferenced snapshots
+        run: cargo insta test --unreferenced=reject --check --workspace
```
lpil: Isn't this running all the tests a second time?

abs0luty: @lpil It does rerun the tests, but insta needs that pass to figure out which snapshots are actually referenced, since it can't know that ahead of time. The snapshot files aren't linked to the code without executing the tests, so this is the only practical way for insta to detect unreferenced snapshots.

We could fold this check into every test-x job, but that would make failures much harder to interpret, since a single unreferenced snapshot would surface in multiple places. 

Keeping it as a separate job keeps the errors in CI isolated and easier to understand. This job runs in parallel with the main test jobs, so it doesn't add to the overall CI runtime.

https://insta.rs/docs/advanced/#:~:text=Handling%20Unused%20Snapshots,-If

lpil: Doubling the amount of CPU time the tests take isn't an option I'm afraid.

abs0luty: Good point! I can put it directly in one of test jobs instead.

.github/workflows/ci.yaml
```diff
@@ -164,6 +171,18 @@ jobs:
...
+      - name: Run tests and check unreferenced insta snapshots
+        uses: clechasseur/rs-cargo@v4
+        with:
+          command: insta
+          args: >-
+            test
+            --unreferenced=reject
+            --workspace
+            --target ${{ matrix.target }}
+        if: ${{ matrix.target == 'x86_64-unknown-linux-gnu' }}
```
lpil: Is there a reason to not always use the insta program? Seems a bit confusing to have the tests here twice but each conditionally run.

lpil:
This is getting a bit complex being done on CI. Maybe we could do this instead with a script that folks use in dev?

.github/workflows/ci.yaml
abs0luty: I actually tried; don't think insta testing can be ran using `cross`. It's not a big issue though, as we don't have target dependent snapshot tests.

abs0luty:
@lpil Well I'm not sure about it, the whole point is to make CI check unreferenced snapshots so this thing doesn't interfere with most of the contributors' work.

lpil:
There's not really any cost to having some outdated text files, so any checking on CI would need to also not have any cost. CI complexity is a headache, so any CI based approach would need to not introduce any cost there

abs0luty:
@lpil What do you think of adding new targets to the Makefile? Or should I just make the PR only remove unused snapshots at this point?

lpil (approved):
That looks fab, let's use that. Thank you!

PR #5201, opened 2025-12-15
Fix record updates discarding unlabelled fields

### LINKED ISSUE #5195: Record update with mixed positional and named fields drops positional fields
The following code

```gleam
pub type T {
  T(Int, a: Int, b: Int)
}

pub fn main() {
  let t = T(1, 2, 3)
  echo T(..t, a: 4)
}
```

outputs in JavaScript (the field b should not be Nil)

```
src/main.gleam:7
T(4, a: 3, b: Nil)
```

but should be

```
src/main.gleam:7
T(1, a: 4, b: 3)
```

and outputs in Erlang (the field b should be present)

```
src/main.gleam:7
T(4, 3)
```

but should be

```
src/main.gleam:7
T(1, 4, 3)
```

I'm using gleam 1.13.0.

GearsDatapacks:
Fixes #5195 

This ended up being a much bigger change than I anticipated

### DIFF
--- a/compiler-core/src/ast/typed.rs
+++ b/compiler-core/src/ast/typed.rs
@@ -105,7 +105,15 @@ pub enum TypedExpr {
         index: u64,
         record: Box<Self>,
         documentation: Option<EcoString>,
-        kind: RecordAccessKind,
+    },
+
+    /// Generated internally for accessing unlabelled fields of a custom type,
+    /// such as for record updates.
+    PositionalAccess {
+        location: SrcSpan,
+        type_: Arc<Type>,
+        index: u64,
+        record: Box<Self>,
     },
 
     ModuleSelect {
@@ -219,7 +227,8 @@ impl TypedExpr {
             | Self::Int { .. }
             | Self::Float { .. }
             | Self::String { .. }
-            | Self::Invalid { .. } => self.self_if_contains_location(byte_index),
+            | Self::Invalid { .. }
+            | Self::PositionalAccess { .. } => self.self_if_contains_location(byte_index),
 
             Self::ModuleSelect {
                 location,
@@ -425,7 +434,8 @@ impl TypedExpr {
             | Self::Float { .. }
             | Self::String { .. }
             | Self::ModuleSelect { .. }
-            | Self::Invalid { .. } => None,
+            | Self::Invalid { .. }
+            | Self::PositionalAccess { .. } => None,
 
             Self::Pipeline {
                 first_value,
@@ -626,6 +636,7 @@ impl TypedExpr {
             | Self::TupleIndex { location, .. }
             | Self::ModuleSelect { location, .. }
             | Self::RecordAccess { location, .. }
+            | Self::PositionalAccess { location, .. }
             | Self::RecordUpdate { location, .. }
             | Self::Invalid { location, .. } => *location,
         }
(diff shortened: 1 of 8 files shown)


compiler-core/src/type_/tests/errors.rs
```diff
@@ -3383,3 +3383,19 @@ pub type Dict(key, value) {
 "#
     );
 }
+
+#[test]
+fn generic_unlabelled_field_in_updated_record_wrong_type() {
+    assert_module_error!(
```
abs0luty, line 3389: This test looks like it should not raise an error because generic type respecialization should resolve it, which matches the earlier point in https://github.com/gleam-lang/gleam/pull/5061#discussion_r2600044018:

```gleam
pub type Wibble(a) {
  Wibble(a, b: Int, c: a)
}

const w = Wibble(1, 2, 3)
const w2 = Wibble(..w, c: False) // Wibble(1, 2, False)
```

Changing how unlabelled fields are handled shouldn't alter this behaviour?

abs0luty (commented):
Left a note!

compiler-core/src/type_/tests/errors.rs
GearsDatapacks: No, because the first positional field is of type `a` also, which is not being updated here, leading to a conflicting `a` type. The first field is of type `Int`, while `c` is of type `Bool`.

abs0luty: You are right, and I am not sure how I missed it! We would need to duplicate this test for constant records as well and the constant version would not produce any errors against expectations. Do you want to dig into it here?

GearsDatapacks: Probably worth opening a separate issue, as they are largely unrelated pieces of code

compiler-core/src/error.rs
```diff
@@ -810,6 +810,15 @@ fn did_you_mean(name: &str, options: &[EcoString]) -> Option<String> {
         .map(|(option, _)| format!("Did you mean `{option}`?"))
 }
 
+fn to_ordinal(value: u32) -> String {
+    match value % 10 {
+        1 => format!("{value}st"),
+        2 => format!("{value}nd"),
```
lpil, line 818: 11st -> 11th
12nd -> 12th
13rd -> 13th

compiler-core/src/ast/typed.rs
```diff
@@ -105,6 +105,7 @@ pub enum TypedExpr {
         index: u64,
         record: Box<Self>,
         documentation: Option<EcoString>,
+        kind: RecordAccessKind,
```
lpil: Record access can never be positional- the data structure mandates that the label is always present.

If we need to modify the AST we'll need to do so in a way that doesn't force invalid data to be added to the fields of a variant.

compiler-core/src/type_/expression.rs
```diff
@@ -3363,20 +3371,107 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+
+                        _ => panic!("Type has already checked to be valid"),
+                    }
+                    .expect("Variant has already checked to be valid");
+
+                let type_ = positional_fields
+                    .get(index as usize)
+                    .expect("Field exists")
+                    .clone();
+                let mut type_vars = im::HashMap::new();
```
lpil, line 3440: Won't this result in rigid type variables not being respected? Without them being passed in they would be instantiated to an unbound variable instead of remaining generic?

lpil (commented):
Thank you! Super fab work!

I've left some Qs inline

compiler-core/src/type_/expression.rs
GearsDatapacks: I'm not sure exactly what you mean by that. I just copied the code that is used for labelled fields.
If you have an example, I can add a test for it

lpil: I wasn't able to write a test that showed incorrect behaviour, so I guess I'm wrong! :)

PR #5204, opened 2025-12-16
Fix formatter removing `@external` annotations from types

GearsDatapacks:
I noticed the formatter would remove `@external` annotations on external custom types. This PR fixes that

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Fixed a bug where the formatter would remove `@external` attributes from
+  custom types.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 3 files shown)


lpil:
Oops. I wonder how we can prevent this mistake, we've made it a couple times before.

GearsDatapacks:
Would be nice to find a way to do that. Not sure what the best way would be though. Maybe destructuring the values in the formatter so that new fields produce an error? That at least forces you to consider it when making changes

lpil:
That's a really good idea! I was thinking of making the formatter run on type tests or integration tests or something, but yours seems easier

GearsDatapacks:
A lot of the non-formatter tests don't follow standard formatting (sometimes on purpose), so that would result in a lot of change needed in tests. I'll make a PR with my suggestion for now, and if that isn't enough, we can come up with something else

PR #5205, opened 2025-12-16
Prevent new annotations from being overlooked in the formatter

GearsDatapacks:
As discussed in #5204, this PR changes the formatter to destructure every definition with no catch-alls, which will force future AST changes to be considered in the formatter, and hopefully prevent this issue from happening in future.


lpil (approved):
Amazing, thank you

PR #5207, opened 2025-12-16
Typecheck unlabelled fields in const record updates

abs0luty:
- No idea how it slipped through, when I was making #5061 😅

While this code produces correct type mismatch error:

```gleam
pub type Wibble(a) {
  Wibble(a: a, b: Int, c: a)
}

const w = Wibble(1, 2, 4)
const w2 = Wibble(..w, c: False) // w is Wibble(Int), incomplete record update
```

This worked without any issues, breaching type safety:

```gleam
pub type Wibble(a) {
  Wibble(a, b: Int, c: a)
}

const w = Wibble(1, 2, 4)
const w2 = Wibble(..w, c: False) // Wibble(1, 2, False)
```

Here is a little explanation: previously `remaining_fields` only contained labelled fields that weren't overridden! So what happens is:
  - `remaining_fields` starts as `{"b": 1, "c": 2}`
  - We override `c`, so it's removed: `remaining_fields = {"b": 1}`
  - The loop only checks field `b` (index `1`)
  - The unlabelled first field (index `0`) is **NEVER** checked!

https://github.com/gleam-lang/gleam/blob/41cb403f0049f1d30c9aad494ac730c79d4a32f4/compiler-core/src/type_/expression.rs#L4026-L4037

- As this pull request is more of an addition to #5201, we will have to wait for it to be merged. Specifically this PR modifies an incomplete record update error message to handle unlabelled fields.

### DIFF
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -4034,11 +4034,16 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                     return self.new_invalid_constant(location);
                 }
 
-                let mut remaining_fields = field_map.fields.clone();
-                let explicit_arguments_count = arguments.len();
+                // Emit warning if no fields are being overridden
+                if arguments.is_empty() {
+                    self.problems
+                        .warning(Warning::NoFieldsRecordUpdate { location });
+                }
 
-                // Type-check explicit override arguments
-                let mut typed_overrides = Vec::new();
+                let mut implicit_labelled_arguments = field_map.fields.clone();
+                let mut update_argument_indices = HashSet::new();
+
+                let mut final_arguments = base_arguments;
                 for argument in arguments {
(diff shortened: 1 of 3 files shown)


compiler-core/src/type_/expression.rs
```diff
@@ -4023,15 +4026,40 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
                             unify(expected_field_type.clone(), field_arg.value.type_())
                     {
-                        self.problems
-                            .error(convert_unify_error(unify_error, location));
+                        let field_name = field_map
+                            .fields
+                            .iter()
+                            .find(|(_, i)| **i == index as u32)
+                            .map(|(name, _)| name.clone())
+                            .unwrap_or_else(|| index.to_string().into()); // Waiting for #5201
```
lpil: It has been merged!

PR #5209, opened 2025-12-18
Enable clippy lint to prevent wildcard matching

GearsDatapacks:
This PR enables a clippy lint which prevents using wildcard (`_`) matches when matching on enums. It also fixes all the places that caused a warning after enabling that lint (which was a lot).

In most cases I expanded the wildcard to all the variants it was implicitly matching, but in some cases where it was clear only one variant was needed (and not future variants), I used `if let` or `matches!` (For example, in the `is_x_variant` functions).


lpil (approved):
Thank you!

PR #5210, opened 2025-12-19
Split language server into its own crate

GearsDatapacks:
This PR splits the code for the language server into its own crate. This should decrease compile times, especially while developing the LS, as only the language server will need to be recompiled, and not the rest of the compiler.

In future we may want to split more parts into their own crates, but other parts of the compiler are more interconnected and would require more complex restructuring of the project.


lpil:
Oh, sorry, your other PR has caused some conflicts

GearsDatapacks:
Rebased to fix the conflicts!

PR #5215, opened 2025-12-22
Fix shadowed variable generation in guards on the JavaScript target

### LINKED ISSUE #5214: Bit arrays in guard clauses do not reference shadowed variables correctly in JavaScript
Small thing I noticed. This code produces different results for js vs erlang.
```gleam
import gleam/io

pub fn main() -> Nil {
  let a = 11
  let a = 12
  let result = case True {
    _ if <<a>> == <<11>> -> "bad"
    _ if <<a>> == <<12>> -> "good"
    _ -> ""
  }
  io.println(result)
}
```

Running `gleam run --target erlang` outputs:
```
good
```

While running `gleam run --target javascript` outputs:
```
bad
```

Version: 1.14.0-rc3

GearsDatapacks:
Fixes #5214

While I was doing this I also removed the `LocalConstant` variant of `ValueConstructor` as I noticed it was not used anywhere, and it made this easier to implement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Fixed a bug where using bit array segments in guard clauses could cause
+  incorrect code to be generated on the JavaScript target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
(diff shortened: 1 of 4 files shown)


lpil (approved):
Thank you!!

PR #5216, opened 2025-12-23
Improve error message when failing to publish a package which already exists

GearsDatapacks:
As discussed [on discord](https://discord.com/channels/768594524158427167/768594524158427170/1453017773411401740), the current error message can be a little confusing when trying to publish a package that already exists, and you don't have permission to modify it. This PR updates it to be more helpful

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Build tool
+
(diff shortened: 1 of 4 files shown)


lpil:
This error message presumes that the person is trying to publish a new package, but that name it taken, but that is not always the case. We would need to make the error message less specific, or identify more precisely when we think this is happening.

Also, please make an issue to discuss designs before making changes 🙏

Issue #5225, opened 2025-12-26
Formatter reoders comments inside function types


abs0luty:
Not sure if it's an issue, but the formatter moves comments inside function parameter lists in a way that changes their position, even though the code itself stays the same. This is confusing when comments are used to annotate or temporarily disable parameters, because the formatter no longer preserves where the comment was written.

For example, the following code places a comment directly above parameter b to indicate intent.
```gleam
fancy_function: fn (
  Int,
  // this is A
  String,
  // this is B
  String,
  // Int,
)
```

After formatting, the comment is moved to the end of the parameter list.
```gleam
fancy_function: fn(
  Int,
  String,
  String,
  // this is A
  // this is B
  // Int,
)
```

I suppose the formatter should keep comments attached to their original location within the parameter list rather than relocating them. 🤔

robertDurst:
Hello 👋

Been creeping for a while hoping to find a "simple enough" issue I could take a stab at. 

Is anyone looking into this? Would it be fine if I assign myself this over the weekend and report back?

If a quick fix is desired ASAP or someone already in the works, let me know!

GearsDatapacks:
Go for it!

robertDurst:
Ok, so making progress here. However, question. This seems to conflict somewhat with the [test case added here](https://github.com/gleam-lang/gleam/blob/3f43218b8431ea9eb3adddd98bc26c28597b2054/compiler-core/src/format/tests/function.rs#L336-L352) as part of this bug: #5004 

That test case asserts:
```gleam
pub fn main() -> #(
  // This is a string
  String, // This is an awesome string
) {
  todo
}
```

Should be rewritten to:
```gleam
pub fn main() -> #(
  String,
  // This is a string
  // This is an awesome string
) {
  todo
}
```

Is this the desired behavior? If so 
```gleam
fancy_function: fn (
  // Int,
  String,
  Int,
  // this is A
  String,
  // this is B
  String,
)
```

should be:
```gleam
fancy_function: fn (
  String,
  Int,
  // this is A
  String,
  // this is B
  String,
  // Int,
)
```

Right? Which feels wrong if `//Int` was just someone temporarily commenting that out (or in the case of laters was a "header of sorts").

Versus what I'd expect is (nothing moved):
```gleam
fancy_function: fn (
  // Int,
  String,
  Int,
  // this is A
  String,
  // this is B
  String,
)
```

***

[...]

PR #5229, opened 2025-12-27
Fix formatter reordering comments in function type parameters

robertDurst:
Address #5225 .

Specifically, this PR:

1. ensures that the positions of comments within nested function params are maintained by popping comments per argument rather than collecting all and appending at the end.
2. adds a unit test to reproduce bug and demonstrate it is fixed (with some bonus edge cases).
3. updates expected comment order in the test `comment_in_tuple_return_type` per conversations in #5225.

Tried my best to follow of convention. Long term goal here is to hopefully continue finding places to contribute, so nit picks of all sizes appreciated 😁

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -805,9 +805,20 @@ impl<'comments> Formatter<'comments> {
             })
             .collect_vec();
 
-        // When comments are present ensure the force_break isn't propagated to parent
         if has_comments {
-            self.wrap_arguments_with_line_breaks(arguments, location.end)
+            // Use line() separators and avoid force_break() to prevent propagation
+            // to parent groups (which would break return type formatting in fn types)
+            break_("(", "(")
+                .append(line())
+                .append(join(arguments, ",".to_doc().append(line())))
+                .append(",")
+                .append(
+                    printed_comments(self.pop_comments(location.end), false)
+                        .map_or(nil(), |c| line().append(c)),
+                )
+                .nest(INDENT)
+                .append(line())
+                .append(")")
         } else {
             self.wrap_arguments(arguments, location.end)
         }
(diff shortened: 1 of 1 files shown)


compiler-core/src/format/tests/function.rs
```diff
@@ -343,12 +343,42 @@ fn comment_in_tuple_return_type() {
 }
 "#,
         r#"pub fn main() -> #(
-  String,
   // This is a string
+  String,
```
robertDurst, line 347: Test fix as per comment in conversation here: https://github.com/gleam-lang/gleam/issues/5225#issuecomment-3694874313

ankddev:
Hey @robertDurst! Could you add a changelog entry (see existing entries for inspiration)?

robertDurst:
> Hey @robertDurst! Could you add a changelog entry (see existing entries for inspiration)?

Done! Thanks for catching this @ankddev ... had the `CONTRIBUTING.md` open in a tab ... longingly waiting for me to read it 🤦

robertDurst:
Ok, cleaned up the commit history, should be GTG for an initial review.

Thanks in advance and happy holidays 🥳 🎄

compiler-core/src/format.rs
```diff
@@ -3049,6 +3065,30 @@ impl<'comments> Formatter<'comments> {
         }
     }
 
+    fn wrap_arguments_with_line_breaks<'a, I>(
```
lpil: How come this is needed when none of the other matching syntaxes need this? Can we not use the same approach for all of them?

lpil (commented):
Thank you!! I've left a small comment inline

compiler-core/src/format.rs
robertDurst: Thanks for the review! Let me take another pass here - had a nice few days off here on holiday, refreshed and with a clear mind 🧘

lpil: Thank you 💜  Happy new year!

compiler-core/src/format.rs
```diff
@@ -790,11 +790,38 @@ impl<'comments> Formatter<'comments> {
...
+                .append(
+                    printed_comments(self.pop_comments(location.end), false)
+                        .map_or(nil(), |c| line().append(c)),
+                )
+                .nest(INDENT)
+                .append(line())
+                .append(")")
+        } else {
+            self.wrap_arguments(arguments, location.end)
+        }
```
lpil, line 824: It looks like this duplicates the code within `wrap_arguments`. Why is it that that code didn't do the job?

I think we should work out what the bug was there so it works for all `(...)`, rather than having a special case where the code is replicated here.

robertDurst:
Optimistically started this during vacation and haven't had enough contiguous blocks of time since to develop a solid enough mental model to be effective. Closing in case someone else is interested in taking this on.

Issue #5218, opened 2025-12-23
Confusing error message when adding a nonexistent package


miguejarias:
When adding a gleam package that does not exist, we get the following error message: 

```shell
error: Dependency resolution failed

An error occured while chosing the version of gleam_regex: resource was not
found
``` 
This error message is a bit confusing, as it suggests that a specific version of the package does not exist, rather than the package itself being missing. Additionally, the wording does not feel as clear or human-friendly as other Gleam error messages.

ankddev:
@lpil This issue can be closed

PR #5230, opened 2025-12-28
tweak the error message for a 404 to be a bit friendlier

Acepie:
Addressing https://github.com/gleam-lang/gleam/issues/5218 mostly just a small tweak to inspect the errors from hexpm (this wording comes from the compiler propagating up https://github.com/gleam-lang/hexpm-rust/blob/main/src/lib.rs#L733-L734 this error when it gets a 404) and convert them to our own enum with more control on the message. I slightly changed the from to not use the normal interface since we might wanna pass other fields from the request in some cases. I just directly passed in the package name but could be the whole request if we saw a need in the future

Image

### DIFF
--- a/compiler-core/src/dependency.rs
+++ b/compiler-core/src/dependency.rs
@@ -241,21 +241,21 @@ impl PackageFetchError {
     pub fn from_api_error(api_error: hexpm::ApiError, package: &str) -> Self {
         match &api_error {
             hexpm::ApiError::NotFound => Self::NotFoundError(package.to_string()),
-            hexpm::ApiError::Json(_) => Self::ApiError(api_error),
-            hexpm::ApiError::Io(_) => Self::ApiError(api_error),
-            hexpm::ApiError::InvalidProtobuf(_) => Self::ApiError(api_error),
-            hexpm::ApiError::UnexpectedResponse(_, _) => Self::ApiError(api_error),
-            hexpm::ApiError::RateLimited => Self::ApiError(api_error),
-            hexpm::ApiError::InvalidCredentials => Self::ApiError(api_error),
-            hexpm::ApiError::InvalidPackageNameFormat(_) => Self::ApiError(api_error),
-            hexpm::ApiError::IncorrectPayloadSignature => Self::ApiError(api_error),
-            hexpm::ApiError::InvalidVersionFormat(_) => Self::ApiError(api_error),
-            hexpm::ApiError::InvalidVersionRequirementFormat(_) => Self::ApiError(api_error),
-            hexpm::ApiError::IncorrectChecksum => Self::ApiError(api_error),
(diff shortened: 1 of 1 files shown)


compiler-core/src/dependency.rs
```diff
@@ -226,20 +226,24 @@ pub trait PackageFetcher {
 
 #[derive(Debug, Error)]
 pub enum PackageFetchError {
+    #[error("The package {0} was not found on Hex")]
```
lpil: Perhaps "in the package repository"? We'll be able to specify other in future, and people may not know what Hex is.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Build tool
 
+- When adding a package that does not exist on Hex, the message is a bit friendlier. ([Ameen Radwan](https://github.com/Acepie))
```
lpil: Wrap this long line please. Credit should be on its own line 🙏

compiler-core/src/dependency.rs
```diff
@@ -226,20 +226,24 @@ pub trait PackageFetcher {
...
-        Self::ApiError(api_error)
-    }
-}
 impl PackageFetchError {
     pub fn fetch_error<T: std::error::Error>(err: T) -> Self {
         Self::FetchError(err.to_string())
     }
+
+    pub fn from_api_error(api_error: hexpm::ApiError, package: &str) -> Self {
+        if let hexpm::ApiError::NotFound = api_error {
```
lpil: Use `match` please 🙏

lpil (commented):
Thank you!

compiler-core/src/dependency.rs
```diff
@@ -226,20 +226,40 @@ pub trait PackageFetcher {
...
+            hexpm::ApiError::IncorrectPayloadSignature => Self::ApiError(api_error),
+            hexpm::ApiError::InvalidVersionFormat(_) => Self::ApiError(api_error),
+            hexpm::ApiError::InvalidVersionRequirementFormat(_) => {
+                Self::ApiError(api_error);
+            }
+            hexpm::ApiError::IncorrectChecksum => Self::ApiError(api_error),
+            hexpm::ApiError::InvalidApiKey => Self::ApiError(api_error),
+            hexpm::ApiError::Forbidden => Self::ApiError(api_error),
+            hexpm::ApiError::NotReplacing => Self::ApiError(api_error),
+            hexpm::ApiError::LateModification => Self::ApiError(api_error),
```
lpil: Could merge these clauses with `|`

Acepie: yeah lemme just do that. i'm being a little silly slowly remembering rust rules so i had them as separate cases destructing and reconstructing at first

PR #5232, opened 2025-12-29
Source maps for Gleam compiled to JavaScript

Acepie:
An initial implementation of https://github.com/gleam-lang/gleam/issues/2249 based off the implementation that was started https://github.com/gleam-lang/gleam/pull/3675 I tried to incorporate the feedback that was left on that pr as best as I could. To separate the sourcemapping logic out of the document printing directly i added a new observer node pattern so that we can call any function/method with the output file position while processing a document. This should theoretically be reusable if we had a need in the future though I'm not entirely sure if there would be one honestly. The bulk of the actual logic is inside the main javascript package where we instrument cursor observer points. essentially whenever we would want to denote a value to map (i.e. an expression, case, etc.) we add a new observer document. I tried writing out the raw mappings to make it a bit easier to line things up but the visualizer https://sokra.github.io/source-map-visualization/# is definitely the easiest way to check this for non trivial programs. I wrote out enough test cases to cover the unique bits but i think we might wanna add more as people try this out/find things that need more robust instrumentation. I think this is in a usable/nice enough state that people could find value in this. I was able to get debuggers working for running tests/a served local web page and it was pretty nice.

Screenshot 2025-12-28 at 5 19 16 PM
Screenshot 2025-12-28 at 8 32 32 PM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,12 +27,12 @@ test: ## Run the compiler unit tests
 	cd test/running_modules && make test
 	cd test/subdir_ffi && make
 
-.PHONY: insta-check-unused
-insta-check-unused: ## Check for unused cargo insta snapshots
+.PHONY: snapshots-check-unused
+snapshots-check-unused: ## Check for unused cargo insta snapshots
 	cargo insta test --unreferenced=warn
 
-.PHONY: insta-fix-unused
-insta-fix-unused: ## Remove unused cargo insta snapshots
+.PHONY: snapshots-fix-unused
+snapshots-fix-unused: ## Remove unused cargo insta snapshots
 	cargo insta test --unreferenced=delete
 
 .PHONY: language-test


compiler-core/src/javascript/tests.rs
```diff
@@ -85,6 +86,31 @@ macro_rules! assert_js {
     }};
 }
 
+#[macro_export]
+macro_rules! assert_source_map {
```
Acepie, line 91: this gets us the raw mappings in a state that is ok enough to line up but honestly its pretty hard to get a good visual without a third party html viewer. if people have ideas on how to improve the rendering lmk

compiler-core/src/javascript/expression.rs
```diff
@@ -214,6 +217,21 @@ impl<'module, 'a> Generator<'module, 'a> {
             scope_position: Position::Tail,
             statement_level: Vec::new(),
             let_assert_always_panics: false,
+            source_map_builder,
+        }
+    }
+
+    pub fn create_cursor_position_observer(&self, start_index: u32) -> Document<'a> {
```
Acepie: this is effectively copy pasted from the top level javascript one. I tried making it generic and not a struct method but couldn't for the life of me get the lifetimes happy. it's just initing the struct so its not really a big deal but kinda annoying

compiler-core/src/javascript/expression.rs
```diff
@@ -133,7 +133,7 @@ impl CurrentFunction {
 pub(crate) struct Generator<'module, 'ast> {
     module_name: EcoString,
     src_path: EcoString,
-    line_numbers: &'module LineNumbers,
+    pub line_numbers: &'module LineNumbers,
```
Acepie: these are pub so that decision can reuse them. this feels slightly off to me but ultimately fine

compiler-core/src/javascript/expression.rs
Acepie: and as soon as i say this it just works xD

compiler-core/src/javascript/tests/sourcemaps.rs
```diff
@@ -0,0 +1,93 @@
+use crate::assert_source_map;
+
+#[test]
```
Acepie, line 3: these were all the cases that were unique enough to handle from what i ran into but we might wanna add more over time as people run into things/the markers get added/expanded

Acepie:
hey @lpil bumping this in case it got lost

compiler-core/src/docs/printer.rs
```diff
@@ -439,7 +439,7 @@ impl Printer<'_> {
             .surround("(", ")")
     }
 
-    fn type_(&mut self, type_: &Type, print_mode: PrintMode) -> Document<'static> {
+    fn type_<'a>(&mut self, type_: &Type, print_mode: PrintMode) -> Document<'a> {
```
lpil: Looks like these 2 functions could still be `'static`?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
```diff
@@ -0,0 +1,47 @@
...
+  return undefined;
+}
+----- SOURCE MAP
+Mappings Raw:
+1:0 -> 5:0
+2:2 -> 6:2
+2:10 -> 6:10
+3:2 -> 7:2
+3:9 -> 7:7
+3:2 -> 8:4
```
lpil: It's cool that the sourcemap is here, but as a human it's pretty hard to tell what these numbers mean. If there was a bug I don't think I would be able to identify it from this.

Is there a good way we could show them in a way where humans can tell what it is? Maybe something like this?

```
----- SOURCE MAP
7| pub fn main() {
⏷
6| export function main() {
-----
8|  let x = True
⏷
7|  let x = true;
```

I don't actually know if that makes sense tbh! Looking at the numbers here I don't understand why some bits of the JS file seem to fall outside this mapping table. i.e. everything after 8:4. Or does everything after that map to 3:2?

Snapshot wizard @giacomocavalieri, do you have any thought on the format?

compiler-core/src/javascript.rs
```diff
@@ -46,6 +110,8 @@ pub struct Generator<'a> {
     module_scope: im::HashMap<EcoString, usize>,
     current_module_name_segments_count: usize,
     typescript: TypeScriptDeclarations,
+    // Debug ignored since SourceMapBuilder doesn't implement debug
+    source_map_builder: DebugIgnore<Option<Rc<RefCell<sourcemap::SourceMapBuilder>>>>,
```
lpil: Put the debug ignore as far down into the type as possible please, so the printable bits are printed.

Alternatively, we could use a custom impl with debug_struct. Would be more code here but would mean we don't need to do `.0` every time this value is used.

compiler-core/src/javascript.rs
```diff
@@ -38,6 +44,64 @@ pub enum JavaScriptCodegenTarget {
...
+    ) -> Self {
+        Self {
+            source_start_location,
+            source_map_builder,
+        }
+    }
+}
+
+impl std::fmt::Debug for SourceMapCursorPositionObserver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
```
lpil: Could you document why there is a custom debug impl please 🙏 

How come we omit one of the fields? Would be good to know it exists

compiler-core/src/javascript/expression.rs
lpil: Instead of this why not expose a method that does what it is always used for? Would make the usages more concise too:

```diff
- create_cursor_position_observer(
-     &generator.source_map_builder.0,
-     generator.line_numbers,
-     location.start
- ),
+ generator.source_map_tracker(location.start)
```

Tried to make the name a bit clearer too, naming it after our domain rather than a design pattern.

compiler-core/src/javascript/typescript.rs
```diff
@@ -32,7 +32,7 @@ use super::{INDENT, concat, import::Imports, join, line, lines, wrap_arguments};
 /// When rendering a type variable to an TypeScript type spec we need all type
 /// variables with the same id to end up with the same name in the generated
 /// TypeScript. This function converts a usize into base 26 A-Z for this purpose.
-fn id_to_type_var(id: u64) -> Document<'static> {
+fn id_to_type_var<'a>(id: u64) -> Document<'a> {
```
lpil: These look like they could all be `'static` still? There's no other changes so there's nothing to specialise them to the less general type of `a'`?

compiler-core/src/type_/pretty.rs
```diff
@@ -144,7 +144,7 @@ impl Printer {
         chars.into_iter().rev().collect()
     }
 
-    fn arguments_to_gleam_doc(&mut self, arguments: &[Arc<Type>]) -> Document<'static> {
+    fn arguments_to_gleam_doc<'a>(&mut self, arguments: &[Arc<Type>]) -> Document<'a> {
```
lpil: Same note about `'static` here

compiler-core/src/erlang.rs
```diff
@@ -641,7 +644,7 @@ fn fun_spec<'a>(
         .group()
 }
 
-fn atom_string(value: EcoString) -> Document<'static> {
+fn atom_string<'a>(value: EcoString) -> Document<'a> {
```
lpil: Same `'static` note in this module

compiler-core/src/erlang.rs
```diff
@@ -3215,7 +3218,7 @@ fn erl_safe_type_name(mut name: EcoString) -> EcoString {
 struct TypePrinter<'a> {
     var_as_any: bool,
     current_module: &'a str,
-    var_usages: Option<&'a HashMap<u64, u64>>,
+    var_usages: Option<HashMap<u64, u64>>,
```
lpil: How come ownership is taken here now?

compiler-core/src/erlang.rs
```diff
@@ -3227,12 +3230,15 @@ impl<'a> TypePrinter<'a> {
         }
     }
 
-    pub fn with_var_usages(mut self, var_usages: &'a HashMap<u64, u64>) -> Self {
-        self.var_usages = Some(var_usages);
-        self
+    pub fn new_with_var_usages(current_module: &'a str, var_usages: HashMap<u64, u64>) -> Self {
```
lpil: How come this has changed?

compiler-core/src/format.rs
```diff
@@ -3261,7 +3261,7 @@ impl<'a> Documentable<'a> for &'a ArgNames {
     }
 }
 
-fn pub_(publicity: Publicity) -> Document<'static> {
+fn pub_<'a>(publicity: Publicity) -> Document<'a> {
```
lpil: Same `'static` note here

compiler-core/src/javascript/tests.rs
lpil: Ah, just seen your comment here. I've started a discussion above

compiler-core/src/pretty.rs
```diff
@@ -477,9 +494,10 @@ pub enum BreakKind {
 fn format(
     writer: &mut impl Utf8Writer,
     limit: isize,
-    mut width: isize,
     mut docs: im::Vector<(isize, Mode, &Document<'_>)>,
 ) -> Result<()> {
+    let mut line: isize = 0;
```
lpil, line 517: This looks like it will be incorrect any time a string has newlines in it?

compiler-core/src/pretty.rs
```diff
@@ -262,6 +271,14 @@ pub enum Document<'a> {
     /// This is useful for additional formatting text which won't be rendered
     /// in the final output, such as ANSI codes or HTML elements.
     ZeroWidthString { string: EcoString },
+
+    /// A node that gets notified of the cursor position as it is being formatted.
+    /// This allows for processes outside of the final output to be notified of
+    /// the cursor position and perform actions based on it, such as recording
+    /// the span of the node in the generated source code for a source mapping.
+    CursorPositionObserver {
+        observer: Rc<RefCell<dyn CursorPositionObserver>>,
```
lpil, line 298: Why is `Rc<RefCell<` used here? Seems like it never gets cloned?

compiler-core/src/docs/printer.rs
Acepie: lemme try and take another pass on the statics i ran into so many issues here with the lifetimes down the chains but maybe future me will be smarter here. the summarized version is that trying to get the observers into the tree made the lifetime requirements really finicky because of the indirection

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
Acepie: the reason for why things fall out of the mapping table is that the spec is kinda weird and only wants the first token of a segment so anything outside the bounds are assumed to be part of the last source map entry. a bit surprising to me but i guess its a space efficiency thing since sourcemaps are already kinda huge even as is

compiler-core/src/pretty.rs
Acepie: its because Document as a whole is clonable and i don't think theres a way to make the trait itself clonable without wrapping it. Switching this to a box gets me something along the lines of 

```
the trait `pretty::CursorPositionObserver` is not dyn compatible
the following types implement `pretty::CursorPositionObserver`:
  javascript::NullCursorPositionObserver
  javascript::SourceMapCursorPositionObserver
```

compiler-core/src/erlang.rs
Acepie: i think when the below methods became not static, module_function started complaining about lifetimes since it returns the results of print. since the only usage of new_with_var_usages was to immediately create a new printer and set the usages on the next line with an instantiated new var_usages i had changed the ownership. that being said, i think i can make the print methods static again so i can just revert a bunch of this

compiler-core/src/javascript/expression.rs
Acepie: that makes sense. i think i tried that at some point and was struggling to get things lined up for some reason though this should be a straightforward lift

compiler-core/src/javascript.rs
Acepie: ahhh yeah i did this because the sourcemap builder we use doesn't implement debug. I could wrap it in debug ignore here also idr why i didn't. I think moving the debugignore inwards might remove the need for this

compiler-core/src/pretty.rs
Acepie: I've added in some tests for new lines and it looks like this is working as intended. Is there a particular case you were thinking of that doesn't work here/that you want me to try?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
Acepie: I added line numbers to the snapshots so that its a bit easier to line up. i'm not entirely sure how much better we can do with just text though

Acepie:
@lpil I think this is good to go for a rereview unless you had further thoughts on the snapshot format

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
lpil: I don't find this any easier to read, I still have to manually check each line by going to the line and then counting the number of columns. This takes such a long time that I don't think it's practically doable.

I think a format like the one I proposed is needed

Acepie: ok i got something that looks about right i think? i tried my best to make it readable but lmk what you think

Acepie:
@lpil bumping this in case it got lost

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
```diff
@@ -0,0 +1,98 @@
...
+19 |  return wibble + wobble;
+20 |}
+
+----- SOURCE MAP
+Mappings:
+----- 
+pub type Wibble {
+⏷
+//# sourceMappingURL=mod.mjs.map
+import { CustomType as $CustomType } from "../gleam.mjs";
```
giacomocavalieri: I don't really understand this mapping, why is it corresponding to the import of `$CustomType`?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
```diff
@@ -0,0 +1,98 @@
...
+1,
+----- 
+2) {
+⏷
+2);
+----- 
+Wibble(wibble, wobble) ->
+⏷
+let wibble = $[0];
+  let wobble = $[1];
```
giacomocavalieri, line 106: Why is the nesting different when it's the same in the original code?

compiler-core/src/javascript/tests/sourcemaps.rs
giacomocavalieri: There's a couple more tests that I would like to see here, for things that sound a bit tricker:
- pipes
- `use`
- recursive functions that get turned into a loop
- lists and tuples
- bit arrays
- a complex case expression with multiple subjects
  ```gleam
  case one, other {
    Ok(1), Error(_) -> "a"
    Ok(_), _ -> "b"
    Error(_), Ok(2) -> "c"
    _, _ -> "d"
  }
  ```

Even if in practice they end up being simpler than expected I think it's still worth it having them explicitly in the tests

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
Acepie: this is probably more of a weirdness of the format plus way its rendering. we don't have any mappings before the first constructor so the import here is implicitly being mapped to everything before the first mapping. I'm not entirely sure what the correct thing for this would be since the type technically has no output. I wonder if long term we should add markers at the end of expressions for things with the same output location to make that more obvious?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
Acepie: i had a trim in the snapshot generating function. lemme remove that

compiler-core/src/javascript/tests/sourcemaps.rs
Acepie: done 👍🏽 should have tests to cover each of these now

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
Acepie: ahh no this is correct technically. since it is saying that the wibble turns into everything starting from let not into the lines that include the let the spaces prior are not included

Acepie:
Not sure why the tests are busted for windows x86 (looks like its busted on main too :') https://github.com/gleam-lang/gleam/actions/runs/23594308997/job/68707004703) but this should be good otherwise now

compiler-core/src/javascript.rs
```diff
@@ -530,10 +642,16 @@ impl<'a> Generator<'a> {
...
+        let head = docvec![
+            self.source_map_tracker(constructor.location.start),
+            head,
+            &constructor.name,
+            " extends $CustomType {"
+        ];
 
         if constructor.arguments.is_empty() {
-            return head.append("}");
+            return docvec![doc, head, "}"];
```
Acepie, line 668: This was actually a bug in the existing behavior. variants with no args were dropping the doc comment in the final output if they had one in the input

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_case_destructure_assignment_statement.snap
Acepie: I added a marker for the start of the custom type definition so that the introduction part gets separately mapped to empty space

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
```diff
@@ -0,0 +1,101 @@
...
+      {
+        kind: "expression",
+        expression: { kind: "expression", value: false, start: 41, end: 42 },
+        start: 34,
+        end: 42,
+        expression_start: 41
+      }
+    )
+  }
+  return undefined;
```
lpil, line 97: Shouldn't this be attributed to the `assert` and not to the `x`?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_use.snap
```diff
@@ -0,0 +1,97 @@
...
+add_3_to_result(
+----- 
+1)
+  
+⏷
+1, 
+----- 
+a + 
+⏷
+(a) => { return a +
```
lpil: Shouldn't this argument map onto the `a` in `use a <-`?

compiler-core/src/pretty.rs
lpil: Could you link me to the tests? I couldn't find them 🙏

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_assert.snap
Acepie: i think this is a quirk of the way i implemented the visualization since it sorts the entries src index. i spent a bit of time reworking the printing to make this properly print when the mappings go backward

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__sourcemaps__sourcemap_use.snap
Acepie: the way we remap use to a fn makes getting a good mapping here really tricky without leaking into the code gen for all functions. i think making the whole function header map back to the whole left of the use is probably the reasonable thing to do here but could in theory remap all the args since we have the data. lemme update to atleast make the function header map

Acepie:
@lpil Ok i think this should address your comments on assert and use mappings. i had to rework the visualization implementation for source mapping to be a bit hairier but result is the same for all the other tests and this should handle cases where the mapping maps to multiple places out of order better

compiler-core/src/pretty.rs
Acepie: https://github.com/gleam-lang/gleam/pull/5232/changes#diff-fc5adbcb3a263cdeaba494d380cacbd9b6dc1968794db2d7289d0f6318120f3cR46-R69 these were the 2 tests i added and https://github.com/gleam-lang/gleam/pull/5232/changes#diff-f62c53658165fc47998715ee0af26c4cec379acf198acfc68e34bfadf7de8458 and https://github.com/gleam-lang/gleam/pull/5232/changes#diff-b4d9ec337220dc78b95469299c9195d5bbe136a7d214c25bd0dfe10b8e3cf0b5 as the output

compiler-core/src/javascript/expression.rs
```diff
@@ -2517,7 +2567,7 @@ pub fn float(value: &str) -> Document<'_> {
     out.to_doc()
 }
 
-pub fn float_from_value(value: f64) -> Document<'static> {
+pub fn float_from_value<'a>(value: f64) -> Document<'a> {
```
lpil: Undo this specialisation please 🙏

PR #5240, opened 2025-12-31
add split alternative pattern code action

### LINKED ISSUE #4856: "Split alternative pattern" code action
Sometimes code starts out as dealing with 2 patterns in the same way, but then later, something different has to happen for one or all of the cases. It could be nice if the language server provided a code action to split an alternative pattern into individual pattern matches.

Before:

```gleam
case next() {
  Ok(response) -> response
  Error(MissingAccessToken) | Error(MissingRefreshToken) -> unauthorized()
  //  ^ split missing pattern
}
```

After:

```gleam
case next() {
  Ok(response) -> response
  Error(MissingAccessToken) -> unauthorized()
  Error(MissingRefreshToken) -> unauthorized()
}
```

To consider:
- split The pattern on the selected pipe; That would make it more flexible, but would require more clicks to fully split a pattern.
- Split out only the currently hovered alternative, leaving the rest in place

~ 💜

vshakitskiy:
Closes #4856.

### DIFF
--- a/language-server/src/tests/snapshots/gleam_language_server__tests__action__split_case_branch_partial_middle.snap
+++ b/language-server/src/tests/snapshots/gleam_language_server__tests__action__split_case_branch_partial_middle.snap
@@ -1,0 +1,25 @@
+---
+source: language-server/src/tests/action.rs
+expression: "\npub fn go(n: Int) {\n  case n {\n    1 | 2 | 3 | 4 -> todo\n    _ -> todo\n  }\n}"
+---
+----- BEFORE ACTION
+
+pub fn go(n: Int) {
+  case n {
+    1 | 2 | 3 | 4 -> todo
+        ▔▔▔▔↑            
+    _ -> todo
+  }
+}
+
+
+----- AFTER ACTION
+
+pub fn go(n: Int) {
+  case n {
+    2 -> todo
+    3 -> todo
+    1 | 4 -> todo
+    _ -> todo
+  }
+}
--- a/language-server/src/tests/snapshots/gleam_language_server__tests__action__split_case_branch_partial_middle.snap
+++ b/language-server/src/tests/snapshots/gleam_language_server__tests__action__split_case_branch_partial_middle.snap
@@ -17,9 +17,9 @@ pub fn go(n: Int) {
 
 pub fn go(n: Int) {
   case n {
+    1 | 4 -> todo
     2 -> todo
     3 -> todo
-    1 | 4 -> todo
     _ -> todo
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@
   variable.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The language server now provides a code action to split alternative case
+  patterns into separate branches.
+  ([Vladislav Shakitskiy](https://github.com/vshakitskiy))
+
 ### Formatter
 
 ### Bug fixes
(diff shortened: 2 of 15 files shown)


language-server/src/tests/snapshots/gleam_language_server__tests__action__split_case_branch_partial_middle.snap
```diff
@@ -0,0 +1,25 @@
...
+
+----- AFTER ACTION
+
+pub fn go(n: Int) {
+  case n {
+    1 | 4 -> todo
+    2 -> todo
+    3 -> todo
+    _ -> todo
+  }
```
giacomocavalieri, line 24: I think when selecting patterns that are in the middle, the alternative should be split by keeping the same order of patterns. So in this test it should become:
```gleam
case n {
  1 -> todo
  2 | 3 -> todo
  4 -> todo
  _ -> todo
}
```
What do you think @lpil?

GearsDatapacks: I would expect to get `2 | 3` and `1 | 4` as patterns here I think

Since neither `1` or `4` is selected here, I definitely wouldn't expect those to be the ones that are split

vshakitskiy: I thought the initial idea is that we split the ones that are selected?  Quoting the example from the issue page:
```gleam
case a {
  AAA | BBB | CCC | DDD -> todo
// ^^^^^^
}

// Becomes:
case a {
  AAA -> todo
  BBB -> todo
  CCC | DDD -> todo
}
```

So I guess it makes sense that if we split the ones on the middle, only them should be affected?

lpil:
Sorry, I lost track here. Is this still work-in-progress?

vshakitskiy:
As far as I remember I was done with the change. The only question was the split behavior on middle cases like this:

```gleam
case n {
  1 | 2 | 3 | 4 -> todo
      ▔▔▔▔↑            
  _ -> todo
}
```
Right now the result is:
```gleam
case n {
  1 | 4 -> todo
  2 -> todo
  3 -> todo
  _ -> todo
}
```

But apparently for most people this is not the right behavior. Sooo should we keep that or change the split logic?

lpil:
Sorry for the delay @vshakitskiy.

My thought is that splitting them all would be the most convenient behaviour for the code action, so that people wouldn't need to highlight them all. What do you think?

vshakitskiy:
So, for this code action, the split should happen for all regardless of which part is highlighted?

```gleam
case n {
  1 | 2 | 3 | 4 -> todo
//    ^^^^^   
  _ -> todo
}

// Becomes:
case n {
  1 -> todo
  2 -> todo
  3 -> todo
  4 -> todo
  _ -> todo
}
```

```gleam
case a {
  AAA | BBB -> todo
  CCC | DDD | EEE -> todo
// ^^
  FFF | GGG -> todo
}

// Becomes:
case a {
  AAA | BBB -> todo
  CCC -> todo
  DDD -> todo
  EEE -> todo
  FFF | GGG -> todo
}
```

GearsDatapacks:
My most common use of this code action in Rust is to remove a single alternative from multiple, so something like:
```rs
case x {
  1 | 2 | 3 | 4 -> todo
  //      ^ Trigger code action
  _ -> todo
}

// Turns into:

case x {
  1 | 2 | 4 -> todo
  3 -> todo
  _ -> todo
}
```

So I would at least expect this to work like that for the single-pattern case. For when multiple patterns are selected, I don't have a particular opinion as I don't really ever use the action like that.

lpil:
Oh! OK then, my assumptions about use is wrong then. Let's have it split out the ones that are currently within the selection.

PR #5255, opened 2026-01-02
Fix bug in the codegen with a logical guard clauses.

### LINKED ISSUE #5254: `case` with `if` can be wrong when JavaScript codegen has `instanceof`
With Gleam v1.14.0, this code:

```gleam
import gleam/io
import gleam/option.{None, Some}

pub fn main() {
  let a = None

  let b = []

  let c = True

  case c == True {
    False if a == Some(1) || b == [] -> "1"
    _else -> "0"
  }
  |> io.println
}
```

Generates this JS:

```js
import * as $option from "../gleam_stdlib/gleam/option.mjs";
import { None, Some } from "../gleam_stdlib/gleam/option.mjs";
import { toList, isEqual } from "./gleam.mjs";

export function main() {
  let a = new None();
  let b = toList([]);
  let c = true;
  let _block;
  let $ = c === true;
  if (!$ && (isEqual(a, new Some(1))) || (isEqual(b, toList([])))) {
    _block = "1";
  } else {
    _block = "0";
  }
  let _pipe = _block;
  return $io.println(_pipe);
}
```

The JS breaks the complex `case` branch due to incorrectly generated parentheses.

Against expectation, `gleam run` prints `1`.

Note, these are initial observations. Changing the order of items in the `if ||` chain can possibly correct the issue (as a workaround), and ordering the chain in a way other than that shown in the minimal repro above, or e.g. adding other simple `Bool` checks, might also generate faulty JS.

vyacheslavhere:
Closes #5254

### DIFF
--- a/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__not_two.snap
+++ b/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__not_two.snap
@@ -13,7 +13,7 @@ pub fn main(x, y) {
 
 ----- COMPILED JAVASCRIPT
 export function main(x, y) {
-  if ((!y && !x)) {
+  if (!y && !x) {
     return 0;
   } else {
     return 1;
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -3,7 +3,9 @@ use super::{
     expression::{self, Generator, Ordering, float, float_from_value},
 };
 use crate::{
-    ast::{AssignmentKind, Endianness, SrcSpan, TypedClause, TypedExpr, TypedPattern},
+    ast::{
+        AssignmentKind, Endianness, SrcSpan, TypedClause, TypedClauseGuard, TypedExpr, TypedPattern,
+    },
     docvec,
     exhaustiveness::{
         BitArrayMatchedValue, BitArrayTest, Body, BoundValue, CompiledCase, Decision,
@@ -298,7 +300,7 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
                 guard,
                 if_true,
                 if_false,
-            } => self.decision_guard(*guard, if_true, if_false),
+            } => self.decision_guard(*guard, if_true, if_false, false),
         }
     }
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -26,3 +26,6 @@
 - The compiler now correctly tracks the minimum required version for expressions
   in `BitArray`s' `size` option to be `>= 1.12.0`.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Clause guard with logical expression now compiles correctly in JS.
```
lpil, line 30: Be more precise with the description please, this sounds like it never worked 🙏

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case_clause_guards__not_two.snap
```diff
@@ -13,7 +13,7 @@ pub fn main(x, y) {
 
 ----- COMPILED JAVASCRIPT
 export function main(x, y) {
-  if (!y && !x) {
+  if ((!y && !x)) {
```
lpil: It would be good to not have these parens here and in other places they are not needed.

compiler-core/src/javascript/tests/case_clause_guards.rs
```diff
@@ -622,3 +622,23 @@ pub fn main() {
...
+
+          let r = case c == True {
+            False if a || b == [] -> \"1\"
+            _else -> \"0\"
+          }
+          assert r == \"0\"
+        }
+        "
+    )
+}
```
lpil, line 644: This test is quite complex, so it's hard to understand. Could you simplify it please 🙏 

Is this testing what the issue reported? The title of the issue says it's related to `instanceof`, but this one doesn't use it? I'm finding it hard to understand this due to the complex test.

lpil (commented):
Thank you! I've left some notes inline 🙏

compiler-core/src/javascript/tests/case_clause_guards.rs
vyacheslavhere: Thanks for the feedback, I'll work on it when I have time.

vyacheslavhere:
I tried to come up with a new solution within the current system. It may be rough, so I am open to suggestions, corrections, etc.

vyacheslavhere:
Just a heads up, the test hasn't been simplified yet.

lpil:
Hi @vyacheslavhere, please mark this as ready-for-review and tag me if you want a review.

PR #5257, opened 2026-01-03
String concatenation in clause guards

### LINKED ISSUE #5238: Compiler error message for wrongly using > instead of >. on floats in a case expression
Example code:

```
pub fn main() {
  let b = 3.0
  case b {
    x if x > 2.0 -> "a"
    _ -> "b"
  }
}
```

which yields:

```
error: Type mismatch
  ┌─ /src/main.gleam:4:10
  │
4 │     x if x > 2.0 -> "a"
  │          ^

Expected type:

    Int

Found type:

    Float

warning: Unreachable pattern
  ┌─ /src/main.gleam:5:3
  │
5 │   _ -> "b"
  │   ^

This pattern cannot be reached as a previous pattern matches the same
values.

```

So it does not mention `>.` like when you do this:

```
pub fn main() {
  echo 3.0 > 2.0
}
```

```
error: Type mismatch
  ┌─ /src/main.gleam:2:12
  │
2 │   echo 3.0 > 2.0
  │            ^ Use >. instead

The > operator can only be used on Ints.

error: Type mismatch
  ┌─ /src/main.gleam:5:10
  │
5 │     x if x > 2.0 -> "a"
  │          ^

Expected type:

    Int

Found type:

    Float
```

abs0luty:
- [x] Closes #5238
- [x] Refactors the binary clause guard AST by removing separate variants for each operator and introducing a single `ClauseGuard::BinOp` node, matching the structure already used in the regular expressions AST.
- [x] Adds support for string concatenation in clause guards, implemented as a direct extension of this unified binary operator representation.


compiler-core/src/javascript/expression.rs
```diff
@@ -2156,116 +2156,83 @@ impl<'module, 'a> Generator<'module, 'a> {
...
+                let operator = match operator {
+                    BinOp::Eq if is_js_scalar(left.type_()) => "===",
+                    BinOp::NotEq if is_js_scalar(left.type_()) => "!==",
+                    BinOp::Eq | BinOp::NotEq => {
+                        let should_be_equal = matches!(
+                            guard,
+                            ClauseGuard::BinaryOperator {
+                                operator: BinOp::Eq,
+                                ..
+                            }
```
lpil: Would be better to match on the operator rather than the `guard` that contains the operator.

Could we make this `let should_be_equal = operator.is_equals()` please

compiler-core/src/parse/token.rs
```diff
@@ -125,9 +125,11 @@ impl Token {
             | Self::GreaterEqualDot
             | Self::GreaterDot => Some(4),
 
-            Self::Plus | Self::PlusDot | Self::Minus | Self::MinusDot => Some(5),
+            Self::LtGt => Some(5),
```
lpil: Could you rename this one to `Concatenate` please 🙏

CHANGELOG.md
```diff
@@ -4,6 +4,19 @@
 
 ### Compiler
 
+- The compiler now reports an error when integer and float binary operators are used incorrectly in case expression guards. ([Adi Salimgereyev](https://github.com/abs0luty))
```
lpil: Wrap long lines please 🙏

compiler-core/src/javascript/expression.rs
abs0luty: Oops, you're right! Forgot to change this match!

PR #5267, opened 2026-01-06
Rename labels

giacomocavalieri:
--- 

### DIFF
--- a/compiler-core/src/reference.rs
+++ b/compiler-core/src/reference.rs
@@ -10,10 +10,63 @@ use petgraph::{
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum ReferenceKind {
+    /// We're referencing a type/value in a qualified manner. For example:
+    ///
+    /// ```gleam
+    /// import wibble
+    ///
+    /// pub fn main() -> wibble.Wobble
+    /// //               ^^^^^^^^^^^^^ This is a qualified reference
+    /// ```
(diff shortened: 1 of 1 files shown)


compiler-core/src/reference.rs
```diff
@@ -10,11 +10,76 @@ use petgraph::{
...
+    ///
     Definition,
+
+    /// This happens when we're referencing a type through an alias.
+    /// For example:
+    ///
+    /// ```gleam
+    /// type Wibble = Wobble
+    /// //            ^^^^^^ This is an alias reference to wobble
+    /// ```
```
GearsDatapacks: `Alias` is slightly different to this. It's when we are referencing a type of value which has been imported using an alias. For example:

```gleam
import wibble.{type Wibble as Wobble}

pub fn main() -> Wobble { ... }
//               ^^^^^^ This is an alias reference
```

For the purposes of renaming, if the name of `Wibble` changed, `Wobble` would remain the same. However, if you want to check for all the pieces of code that reference `Wibble`, this one should be considered, even though it's by a different name.

All the other comments are correct thought!

giacomocavalieri: Thanks!

lpil:
Oh wow I didn't know about this PR 😅

giacomocavalieri:
Ahah I was just exploring and a bit too eager to push. I'm really happy someone else picked it up 😂

PR #5274, opened 2026-01-08
Added empty TypeVariantConstructors record for UtfCodepoint into the Prelude

### LINKED ISSUE #5262: Compiler crashes when trying to pattern match on `UtfCodepoint`
Hello! Came across this a while ago but completely forgot to submit an issue until earlier today. Sorry for the delay ^.^
```
DEBUG compile{package=wibble}: new_module_to_be_compiled module=wibble
  Compiling wibble
 INFO compile{package=wibble}: analysing_modules count=1
DEBUG compile{package=wibble}: Type checking module="wibble"
error: Fatal compiler bug!

Panic: compiler-core/src/exhaustiveness.rs:855
        Custom type variants must exist: ModuleType { name: "UtfCodepoint", module_name: "gleam", type_constructors: ["Int", "Result", "String", "Nil", "BitArray", "List", "UtfCodepoint", "Bool", "Float"], imported_type_as_value: false }
Gleam version: 1.14.0
Operating system: linux

variable set.
```

# [Minimal Reproduction](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIcAtgA4JT4AEA5gDYQCG5A9DPlHEgwB0kQygFcARnQBmSOuVZ8AFAEo6wIXTot6rGDGj0A8gGtFAdzjjxLVQF463XvwB0o/FID6eACYRqffEUAdnRlDTocXQg6CysWNXDNcIBfIWSQZKA)

```gleam
import gleam/string

pub fn main() {
  let assert Ok(wibble) = string.utf_codepoint(71)
  case wibble {
    
  }
}
```
The playground reports `RuntimeError: unreachable executed`.

It was quite confusing the first time I accidentally tried doing this (I now know you can't) and the LS crashed and cut out.

I suppose the expected behaviour would be an error notification to the effect of:
> Can't pattern match on `UtfCodepoint` because it's an FFI type without a constructor.

At least, that's what I think is the cause here.

seafoamteal:
Hello! This resolves #5262.

Without this, the compiler panicked when looking up the type variant constructors for `UtfCodepoint`in the stdlib module's types-to-value-constructors map as there was no entry for it. Now, the lookup is successful and the exhaustiveness checking code knows there's no way to directly construct a value of type `UtfCodepoint`.

Perhaps this would be better off in a separate issue/discussion, but would it be helpful to have a compiler warning/hint telling programmers that you can't pattern match on a type with no variant constructors i.e. external types similar to `UtfCodepoint`?

🩷

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,6 @@
   friendlier.
   ([Ameen Radwan](https://github.com/Acepie))
 
-- The compiler no longer crashes when trying to pattern match on a `UtfCodepoint`.
-  ([Hari Mohan](https://github.com/seafoamteal))
-
 ### Language server
 
 - The language server now allows extracting the start of a pipeline into a
@@ -36,4 +33,7 @@
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
 - The language server no longer recommends the deprecated `@target` attribute.
+  ([Hari Mohan](https://github.com/seafoamteal))
+
+- The compiler no longer crashes when trying to pattern match on a `UtfCodepoint`.
   ([Hari Mohan](https://github.com/seafoamteal))
--- a/compiler-core/src/type_/tests.rs
+++ b/compiler-core/src/type_/tests.rs
@@ -3062,13 +3062,10 @@ fn assert_suitable_main_function_javascript_not_supported() {
             purity: Purity::Impure,
         },
     };
-    assert!(assert_suitable_main_function(
-        &value,
-        &"module".into(),
-        Origin::Src,
-        Target::JavaScript
-    )
-    .is_err(),);
+    assert!(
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 - When adding a package that does not exist on Hex, the message is a bit friendlier.
   ([Ameen Radwan](https://github.com/Acepie))
 
+- The compiler no longer crashes when trying to pattern match on a `UtfCodepoint`.
+  ([Hari Mohan](https://github.com/seafoamteal))
```
lpil: Could you put this in the bugs section please 🙏

seafoamteal: Oh whoops, sorry! Must have been too sleepy to read the headings correctly 😅. Also fixed the `cargo fmt` failure from the CI run.

PR #5276, opened 2026-01-08
fix error with function evaluating arguments in the wrong order

### LINKED ISSUE #5272: Changing order of labels in a function call does not change order of evaluation.
I've noticed that when using labels in a function call, the order arguments are evaulated in does not correspond to the order written in the source code. Instead, the arguments are evaluated in the order they are defined in the function itself.

I have this function:

```gleam
pub fn fit(into viewport: Bounds, box box: Bounds) -> Transform
```

You can see it has two arguments, labelled `into:` and `box:` respectively. If we take this call:

```gleam
transform.fit(
  into: echo bounds.new(..) as "viewport",
  box: echo graph.fold(..) as "box",
)
```

and look at the console output, we can see:

```
[Log] dev/experiments/layout.gleam:40 viewport (layout.mjs, line 246)
#(0, 0, 300, 300)

[Log] dev/experiments/layout.gleam:41 box (layout.mjs, line 246)
#(0, 0, 293.48392537260776, 291.62981208738245)
```

All is expected. `"viewport"` is logged first, then `"box"`.

However, if we flip the order we provide the labels, so now the function call is:

```gleam
transform.fit(
  box: echo graph.fold(..) as "box",
  into: echo bounds.new(..) as "viewport",
)
```

We can see the that the logs have not changed, and `"viewport"` is still logged **first**.

```
[Log] dev/experiments/layout.gleam:51 viewport (layout.mjs, line 246)
#(0, 0, 300, 300)

[Log] dev/experiments/layout.gleam:40 box (layout.mjs, line 246)
#(0, 0, 293.48392537260776, 291.62981208738245)
```

Is this a bug? It is very surprising that side effects can happen in a different order to the one provided in the source code!

giacomocavalieri:
This PR closes #5272 
I'm not so sure about the JS solution, what do you think?

---

### DIFF
--- a/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__functions__call_with_out_of_place_labels.snap
+++ b/compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__functions__call_with_out_of_place_labels.snap
@@ -1,0 +1,43 @@
+---
+source: compiler-core/src/erlang/tests/functions.rs
+expression: "\nfn wibble(a a, b b) { 1 }\nfn fun(arg) { panic as \"impure!\" }\n\npub fn main() {\n  wibble(fun(\"evaluated first\"), a: fun(\"evaluated second\"))\n}\n"
+---
+----- SOURCE CODE
+
+fn wibble(a a, b b) { 1 }
+fn fun(arg) { panic as "impure!" }
+
+pub fn main() {
+  wibble(fun("evaluated first"), a: fun("evaluated second"))
+}
+
+
+----- COMPILED ERLANG
+-module(my@mod).
+-compile([no_auto_import, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch, inline]).
+-define(FILEPATH, "project/test/my/mod.gleam").
+-export([main/0]).
+
+-file("project/test/my/mod.gleam", 2).
+-spec wibble(any(), any()) -> integer().
+wibble(A, B) ->
+    1.
+
+-file("project/test/my/mod.gleam", 3).
+-spec 'fun'(any()) -> any().
+'fun'(Arg) ->
+    erlang:error(#{gleam_error => panic,
+            message => <<"impure!"/utf8>>,
+            file => <<?FILEPATH/utf8>>,
+            module => <<"my/mod"/utf8>>,
+            function => <<"fun"/utf8>>,
+            line => 3}).
+
+-file("project/test/my/mod.gleam", 5).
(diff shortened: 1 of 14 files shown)


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__functions__call_with_out_of_place_labels.snap
```diff
@@ -0,0 +1,43 @@
...
+            line => 3}).
+
+-file("project/test/my/mod.gleam", 5).
+-spec main() -> integer().
+main() ->
+    begin
+        Arg = 'fun'(<<"evaluated first"/utf8>>),
+        Arg@1 = 'fun'(<<"evaluated second"/utf8>>),
+        wibble(Arg@1, Arg)
+    end.
```
lpil, line 43: Can we remove these redundant blocks? 🙏

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__functions__call_with_out_of_place_labels_not_in_tail_position.snap
```diff
@@ -0,0 +1,34 @@
...
+}
+
+function fun(arg) {
+  throw makeError("panic", FILEPATH, "my/mod", 3, "fun", "impure!", {})
+}
+
+export function main() {
+  (() => { let arg = fun("evaluated first");
+    let arg$1 = fun("evaluated second");
+    return wibble(arg$1, arg); })();
```
lpil, line 32: Can we remove these redundant IIFEs? They have a performance cost.

For any that we are to keep, can we fix the formatting please 🙏

lpil (commented):
Thank you! Couple notes inline 🙏 

Looking at this now, would it not make sense to expand calls with labels to a series of statements prior to codegen? This would mean not needing to duplicate the code across both target code generators, and we wouldn't need to add another special case to collapse blocks in codegen.

giacomocavalieri:
Oh I haven't thought of it that's a great idea! I'll change the implementation

lpil:
Sweet, thank you!

jtdowney:
@giacomocavalieri did you mean to close this?

giacomocavalieri:
Hello! Yes. This PR fixed the bug by doing the reordering in the code generators. So it was basically implemented twice. Me and Louis agreed it was not the best and that it should be something that is done earlier so that the code generators don't have to implement the same logic twice separately

PR #5280, opened 2026-01-09
fix: don't perform codegen when exporting package interface

### LINKED ISSUE #5278: Be more precise over when codegen is performed in the CLI
The Gleam CLI often does codegen - in particular calling the beam compiler - in a number of commands where I don't think that step (at least) is necessary. This would speed up those steps dramatically, as the work to generate and write Erlang/Javascript source files and call the Erlang compiler could all be skipped.

From a quick survey these where the commands I found that are affected by this:

- `gleam check`: Performs codegen of dependencies, but not for the root package.
- `gleam fix`: Not actually sure about this command at all, but it does do codegen for all dependencies first.
- `gleam export package-interface`: Always implicitly performs a normal build.
- `gleam docs build`: Always performs full codegen without any cache. This is especially disruptive, as it prevents you from running this command using e.g. `watchexec` effectively  to iterate on the documentation.

I also found [this comment](https://github.com//yoshi-monster/gleam/commit/7f478c8b05eee283a99ed1db8d926300f9fd4bfd#diff-b744e4ee55b136756b9b2a88080fa25baf199b8399aa4d16fd8cc42804181c2bL44-L48) from an older version. I'm not quite sure how the cache affects things, but I would expect the cache to always be used, even if no codegen is performed. The LSP might affect that though.

ankddev:
Closes #5278. I believe that this is fix and not new feature, so added to corresponding section in changelog.
***
- [ ] Tests have been added for new behaviour (N/A)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,5 +48,5 @@
   [Vladislav Shakitskiy](https://github.com/vshakitskiy))
 
 - The build tool now doesn't perform code generation when exporting package
-  interface
+  interface.
   ([Andrey Kozhev](https://github.com/ankddev))


giacomocavalieri:
Thank you, looks good to me!

giacomocavalieri:
Looks like there's a conflict with the CHANGELOG, would you mind rebasing on main?

ankddev:
@giacomocavalieri I'll rebase once I will go to home

giacomocavalieri:
Lovely, thank you!

lpil:
I've rebased!

PR #5287, opened 2026-01-13
fix error with unreachable binary patterns

### LINKED ISSUE #5286: Bug in exhaustiveness matching
The compiler incorrectly considers the second pattern here as unreachable:
```gleam
pub fn main(x) {
   case x {
    <<_, "==">> -> 1
    <<_, _,"=">> -> 2
//  ^^^^^^^^^^^^ Unreachable
    _ -> 3
  }
}
```
I'm working on this!

giacomocavalieri:
This PR closes #5286 

---


compiler-core/src/javascript/tests/case.rs
```diff
@@ -842,22 +842,6 @@ pub fn go(x, y) {
...
-#[test]
-fn interfering_string_pattern_fails_if_succeeding() {
-    assert_js!(
-        r#"
-pub fn wibble(bits) {
-  case bits {
-    <<"aaa", 0, _:bits>> -> 1
-    // If the first one fails, we know this one won't match, so it won't appear
-    // in the final else branch!
-    <<_, "aa", 1, _:bits>> -> 2
```
giacomocavalieri, line 854: This test and the other one in this file I removed were actually wrong! If the first pattern doesn't match the second one actually has a chance of matching. For example if the input were `<<"baa", 1>>` it would not match with the first branch but it would with the second. Don't know how I couldn't notice this 😅

lpil (approved):
Beautiful, thank you!!

PR #5292, opened 2026-01-15
Fix bug with case expressions and assert

giacomocavalieri:
This PR fixes a test that is failing on the main branch.
The compiler no longer generates invalid code when a case expression is used on the right hand side of an equality check in an assert!


lpil (approved):
Thank you!!!

PR #5293, opened 2026-01-16
Enhance the `Extract function` code action to refactor anonymous functions into top-level functions

### LINKED ISSUE #5288: A code-action to refactor anonymous functions to top-level ones
I often start out writing functions inline, but later decide (often after it has grown significantly) that the code would be easier to understand if it was a named top-level function instead.

Right now there's an existing code action called `Extract function` that takes a piece of code and makes a new top-level function for it. Applying this code action to an anonymous creates an extra wrapper around it:

```gleam
pub fn main() {
  let needle = 42
  let haystack = [1, 67, 1337, 42, 69]
  list.filter(haystack, fn(x) { x == needle })
  //.                   ^-------------------- Extract function
}
```

produces

```gleam
fn function(needle: Int) -> fn(Int) -> Bool {
  fn(x) { x == needle }
}
```

I would like for something very similar, but:

- I can trigger the code action on the `fn` keyword without having to select the entire expression
- It un-curries the function and instead partially applies it at the call site.
- It puts the subject first and not last to make the function suitable for piping.

So ideally this would produce this instead:

```gleam
pub fn main() {
  let needle = 42
  let haystack = [1, 67, 1337, 42, 69]
  list.filter(haystack, function(_, needle))
}

fn function(x: Int, needle: Int) -> Bool {
  x == needle
}
```

For anonymous functions with more than 2 arguments this would be equivalent to triggering the `Extract function` code action on the entire body.

Thank you! ~ 💜

seafoamteal:
Resolves #5288.

It's my first time working on a code action, so I'd appreciate any feedback! In particular, I'm a bit unsure of my function and test names because I found it tricky to describe the new behaviour succinctly.

Also, in `extract_unary_anonymous_function`, I've taken care to make sure the produced function is formatted correctly, at the cost of the code being textually longer. The formatter is able to clean up trailing commas in parameter/argument lists, so I can get rid of the formatting if we want the code to be simpler here.

# Checklist


giacomocavalieri:
From a quick read it seems like this is only working for unary and nullity anonymous functions, why not make this work with functions with any arity? Or maybe I am not understanding the code

seafoamteal:
Oh! Thanks for pointing that out, I overlooked the fact that this will also work for any anonymous functions that don't close over variables from an outer scope. Fixed now!

yoshi-monster:
Hey thank you so much!

I think the idiomatic way to write this for non-unary functions with captures would be to do the following instead:

```gleam
fn main() {
  let factor = 2
  list.fold([], 0, fn(acc, value) { acc + factor * value })
}

// --->

fn main() {
  let factor = 2
  list.fold([], 0, fn(acc, value) { function(acc, value, factor) })
}

fn function(acc, value, factor) {
  acc + factor * value
}

```

Not sure if that's better, but that's what I meant with "extract the body instead" :)

seafoamteal:
I think that makes sense, actually. Your example makes it pretty apparent, because I can see how it gets confusing if the original `acc` and `value` arguments get obscured like they do in my current implementation. If any core team members want to weigh in on this, I can quickly make the change!

Just for reference, this is what happens in my current implementation, as opposed to Rebecca's explanation above:

```gleam
pub fn main() {
  let factor = 2
  list.fold([], 0, function(factor))
}

fn function(factor: Int) -> fn(Int, Int) -> Int {
  fn(acc, value) { acc + factor * value }
}

```

lpil:
I'd expect what Yoshi shared there. Thank you 🙏

seafoamteal:
I've just fixed that! The behaviour should now be:

Before:
```gleam
pub fn main() {
  list.each([1, 2, 3], fn(x) { io.println(int.to_string(x)) })

  let needle = 42
  let haystack = [1, 67, 1337, 42, 69]
  list.filter(haystack, fn(x) { x == needle })

  let factor = 2
  list.fold([], 0, fn(acc, value) { acc + value * factor })
}
```

After:
```gleam
pub fn main() {
  list.each([1, 2, 3], function)

  let needle = 42
  let haystack = [1, 67, 1337, 42, 69]
  list.filter(haystack, function_2(_, needle))

  let factor = 2
  list.fold([], 0, fn(acc, value) { function_3(acc, value, factor) })
}

fn function_3(acc: Int, value: Int, factor: Int) -> Int {
  acc + value * factor
}

fn function_2(x: Int, needle: Int) -> Bool {
  x == needle
}

fn function(x: Int) -> Nil {
  io.println(int.to_string(x))
}
```

language-server/src/tests/action.rs
```diff
@@ -11625,3 +11625,54 @@ pub fn wibble() {
...
+ pub fn main() {
+  let outer_scope = 3
+  let wibble = fn(a, b) {
+    a + b + outer_scope
+  }
+}
+        ",
+        find_position_of("fn(").select_until(find_position_of("}"))
+    );
+}
```
giacomocavalieri, line 11832: Could you maybe add a couple of tests where the anonymous function being extracted is passed as an argument to a function?

language-server/src/code_action.rs
```diff
@@ -9452,6 +9490,135 @@ impl<'a> ExtractFunction<'a> {
...
+  {function_body}
+}}"
+        );
+        self.edits.insert(function_end, function);
+    }
+
+    /// Even if an anonymous function captures variables from an external
+    /// scope, it can still be refactored more ergonomically than the default
+    /// _if it only expects a single argument_. In this case, the original
+    /// argument can be replaced with a capture hole.
```
giacomocavalieri: Could you add a code snippet showing what this will look like? Also talking of the "default" could you explicitly say which function is considered the default, that will be handy for people reading this in future

language-server/src/code_action.rs
```diff
@@ -9452,6 +9490,135 @@ impl<'a> ExtractFunction<'a> {
...
+  {function_body}
+}}"
+        );
+        self.edits.insert(function_end, function);
+    }
+
+    /// In the case where a non-unary anonymous function captures variables
+    /// from an external scope, the idiomatic refactoring is to replace the
+    /// _body_ of this function with a newly generated function that accepts
+    /// all the original arguments and captured variables as parameters.
```
giacomocavalieri: Could you add a code snippet to show the desired outcome?

seafoamteal:
I've improved the comments in the source and added some more tests now, including ones which document the behaviour when ignored arguments are present.

giacomocavalieri (approved):
Lovely thank you!

PR #5298, opened 2026-01-18
make sure formatter no longer stack overflows when formatting big lists

### LINKED ISSUE #5297: Formatter stack overflows with long lists
When trying to format a long list with many literal items (from my tests the tipping point seems to be around 8000 items, but I'm guessing it will depend on the specific machine this runs on) the formatter fails with a stack overflow. I'm working on a fix!

giacomocavalieri:
This PR closes #5297.
This seems to be a really straightforward fix: the `docvec!` macro ended up creating deeply nested documents `vec(vec(vec(vec(...), ...), ...), ...)`. Now if the first item passed to the macro is a vector, it behaves the same as appending to that vector rather than creating a new vector every single time!

### DIFF
--- a/compiler-core/src/pretty.rs
+++ b/compiler-core/src/pretty.rs
@@ -68,7 +68,12 @@ macro_rules! docvec {
         // So in case a `docvec!` starts with a `Vec` we flatten it out to avoid
         // having deeply nested documents.
         match $first.to_doc() {
-            first@Document::Vec(_) => first.append(docvec![$($rest.to_doc()),+]),
+            Document::Vec(mut vec) => {
+                $(
+                    vec.push($rest.to_doc());
+                )*
+                Document::Vec(vec)
+            },
             first => Document::Vec(vec![first, $($rest.to_doc()),+])
         }
     };


compiler-core/src/pretty.rs
```diff
@@ -56,8 +56,21 @@ macro_rules! docvec {
...
+    };
+
+    // A docvec![] with multiple elements.
+    ($first:expr, $($rest:expr),+ $(,)?) => {
+        // A document that looks like this: `Vec[Vec[..rest], ..other_rest]`
+        // is exactly the same as a flat: `Vec[..rest, ..other_rest]`.
+        // So in case a `docvec!` starts with a `Vec` we flatten it out to avoid
+        // having deeply nested documents.
+        match $first.to_doc() {
+            first@Document::Vec(_) => first.append(docvec![$($rest.to_doc()),+]),
```
lpil: Why not append all the items to the inner vector instead of recursively calling the macro?

lpil (commented):
Quick question inline!

compiler-core/src/pretty.rs
giacomocavalieri: This way we can flatten all vectors at the top of the macro: 
```
docvec![vec1, vec2, a, b] // becomes =>
vec1.append(vec2.append(vec![a, b]))
// Meaning the first two vectors are flattened out, rather than only the first one
```

lpil: That moves the elements multiples times needlessly, potentially allocating new vectors and copying the existing elements each time. If the `Vec` is worked with directly it allocates at most once, and moves the new elements once. Is there a reason to do more work to restructure the tree?

giacomocavalieri: Mhm you're right and just doing it for the first element seems to be fixing the issue anyway. I'll change it to that!

compiler-core/src/pretty.rs
```diff
@@ -68,7 +68,7 @@ macro_rules! docvec {
         // So in case a `docvec!` starts with a `Vec` we flatten it out to avoid
         // having deeply nested documents.
         match $first.to_doc() {
-            first@Document::Vec(_) => first.append(docvec![$($rest.to_doc()),+]),
+            first@Document::Vec(_) => first.append(Document::Vec(vec![$($rest.to_doc()),+])),
```
lpil: This is still allocating a new vector and copying everything twice! Mutate the inner vec instead and push each item directly to it.

giacomocavalieri: Oooh I see what you meant let me fix that real quick

giacomocavalieri: Should be good for real this time!

PR #5300, opened 2026-01-19
Make gleam.toml consistent

lpil:
These keys were inconsistent in that they used a different case convention to the other keys.

Assuming no bugs this is backwards compatible and will support the old config too.

### DIFF
--- a/compiler-core/src/snapshots/gleam_core__docs__package_config_to_json.snap
+++ b/compiler-core/src/snapshots/gleam_core__docs__package_config_to_json.snap
@@ -20,7 +20,7 @@ gleam = ">= 0.30.0"
 gleam_stdlib = ">= 0.18.0 and < 2.0.0"
 my_other_project = { path = "../my_other_project" }
 
-[dev-dependencies]
+[dev_dependencies]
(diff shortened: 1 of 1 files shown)


compiler-cli/src/remove.rs
```diff
@@ -21,17 +21,19 @@ pub fn command(paths: &ProjectPaths, packages: Vec<String>) -> Result<()> {
...
+                .as_table_like_mut()
+                .and_then(|deps| deps.remove(package_to_remove))
+        };
+
+        // dev-dependencies is the old deprecated name for dev_dependencies
+        let removed = remove(&mut toml, "dependencies")
+            .or_else(|| remove(&mut toml, "dev_dependencies"))
+            .or_else(|| remove(&mut toml, "dev-dependencies"));
+
+        if removed.is_none() {
```
giacomocavalieri, line 36: Doesn't this change the behaviour of this piece of code? Previously both the remove from dependencies and from dev_dependencies would run and then the check is performed. Does it make any difference?

lpil: It does change in that way, but that config would have been invalid and rejected by the build tool

PR #5304, opened 2026-01-20
 Support renaming string prefix pattern variables

### LINKED ISSUE #5270: LS: Rename string prefix pattern alias and suffix variable (`"prefix" as prefix_alias <> suffix`)
Currently, it's not possible to rename either the alias of a string prefix pattern (`as prefix_alias`) or its suffix variable (`<> suffix`).

Code examples:

```gleam
pub fn main() {
  let wibble = "1-wibble"
  case wibble {
    "1" as digit <> rest -> digit <> rest
    other -> other
  }
}
```

```gleam
pub fn main() {
  let assert "1" as digit <> rest = "1-wibble"
  digit <> rest
}
```

IgorCastejon:
Closes #5270.

### DIFF
--- a/language-server/src/tests/hover.rs
+++ b/language-server/src/tests/hover.rs
@@ -1816,3 +1816,93 @@ pub fn wibble(w: Wibble) {
         find_position_of("w.wibble").under_char('l')
     );
 }
+
+#[test]
+fn hover_for_string_prefix_pattern() {
+    assert_hover!(
+        "
+pub fn main() {
+  case \"wibble\" {
+    \"wib\" as alias <> suffix -> alias <> suffix
+    other -> other
+  }
+}
+",
+        find_position_of("<>")
+    );
+}
+
+#[test]
+fn hover_for_string_prefix_pattern_prefix_alias() {
+    assert_hover!(
+        "
+pub fn main() {
+  case \"wibble\" {
+    \"wib\" as alias <> suffix -> alias <> suffix
+    other -> other
+  }
+}
+",
+        find_position_of("alias").nth_occurrence(1)
+    );
+}
+
+#[test]
+fn hover_for_string_prefix_pattern_suffix_variable() {
+    assert_hover!(
+        "
+pub fn main() {
+  case \"wibble\" {
+    \"wib\" as alias <> suffix -> alias <> suffix
+    other -> other
+  }
+}
+",
+        find_position_of("suffix").nth_occurrence(1)
+    );
+}
+
+#[test]
+fn hover_for_string_prefix_pattern_prefix_alias_alternative_definition() {
+    assert_hover!(
+        "
+pub fn main() {
+  case \"wibble\" {
+    \"wib\" as prefix <> rest | \"wob\" as prefix <> rest -> prefix <> rest
+    other -> other
+  }
+}
+",
+        find_position_of("prefix").nth_occurrence(2)
+    );
+}
+
+#[test]
+fn hover_for_string_prefix_pattern_suffix_variable_alternative_definition() {
+    assert_hover!(
+        "
+pub fn main() {
+  case \"wibble\" {
+    \"wib\" <> rest | \"wob\" <> rest -> rest
+    other -> other
+  }
+}
+",
+        find_position_of("rest").nth_occurrence(2)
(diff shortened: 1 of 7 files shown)


language-server/src/tests/reference.rs
```diff
@@ -48,21 +48,22 @@ fn show_references(code: &str, position: Option<Position>, ranges: &[Range]) ->
...
+                .any(|range| range.start <= current_position && current_position < range.end);
+
+            if Some(current_position) == position {
+                underline_empty = false;
+                underline.push('↑');
+            } else if is_in_range {
+                underline_empty = false;
+                underline.push('▔');
+            } else {
+                underline.push(' ');
```
IgorCastejon, line 66: This change was necessary because otherwise multiple references in the same line would not be printed.

language-server/src/reference.rs
```diff
@@ -510,6 +545,27 @@ impl FindVariableReferences {
...
+                self.update_alternative_origin(*location);
+
+                _ = self.references.insert(VariableReference {
+                    location: *location,
+                    kind: VariableReferenceKind::Variable,
+                });
+            }
+            AlternativeVariable::Track | AlternativeVariable::Ignore => {}
+        }
+    }
```
IgorCastejon, line 541: Extracted this logic as-is. I kept the code identical to the original implementation to avoid diff noise.

language-server/src/engine.rs
```diff
@@ -640,50 +640,57 @@ where
...
-                        ..
+            Ok(
+                match reference_for_ast_node(found, &current_module.name, byte_index) {
+                    Some(Referenced::LocalVariable {
+                        location, origin, ..
+                    }) if location.contains(byte_index) => match origin.map(|origin| origin.syntax)
+                    {
+                        Some(VariableSyntax::Generated) => None,
+                        Some(
+                            VariableSyntax::Variable(label) | VariableSyntax::LabelShorthand(label),
```
IgorCastejon: Note: Mostly formatting changes in this file. The only logic change is passing the new `byte_index` argument to `reference_for_ast_node`.

CHANGELOG.md
```diff
@@ -32,6 +32,18 @@
   variable.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- It's now possible to find references and rename variables in string prefix
+  patterns.
+
+  ```gleam
+  case wibble {
+    "1" as digit <> rest -> digit <> rest
+    //     ^^^^^    ^^^^ You can now trigger "Find references" and "Rename" from here
```
giacomocavalieri: I try and keep the changelog lines below 80 chars of length
```suggestion
    //     ^^^^^    ^^^^
    // You can now trigger "Find references" and "Rename" from here
```

language-server/src/reference.rs
```diff
@@ -231,6 +232,40 @@ pub fn reference_for_ast_node(
             name_kind: Named::CustomTypeVariant,
             target_kind: RenameTarget::Definition,
         }),
+        Located::Pattern(Pattern::StringPrefix {
+            left_side_assignment,
+            right_side_assignment,
+            right_location,
+            ..
+        }) => {
```
giacomocavalieri: I'm thinking rather than updating this `reference_for_ast_node` function adding the position as a parameter the code returning the `Located::Pattern` should return some more specific data structure for this case to tell us we're inside the left/right side assignment or not

language-server/src/reference.rs
giacomocavalieri: Nice refactoring!

language-server/src/reference.rs
IgorCastejon: Does it make sense to add a new `PatternStringPrefixVariable` variant to the `Located` type?

giacomocavalieri: I think it could! Best wait and hear what @lpil has to say

lpil: Sounds good to me!

IgorCastejon: @giacomocavalieri @lpil I've done this change! Due to this new variant, I also had to change the hover and autocomplete logic - please let me know if I need to add or change anything.

PR #5309, opened 2026-01-20
Fix erlang-shipment including dev dependencies

### LINKED ISSUE #5308: `gleam export erlang-shipment` includes dev dependencies
### Discussed in https://github.com/gleam-lang/gleam/discussions/5307

<sup>Originally posted by **jtdowney** January 20, 2026</sup>
I tried to look through the issues in other history, but I wasn't able to determine whether this was intentional or a bug.</div>

jtdowney:
The `gleam export erlang-shipment` command was incorrectly including dev dependencies in the exported shipment.

The fix filters packages at shipment creation time by computing the set of production dependencies and only copying those to the output directory.

Closes #5308

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -238,6 +238,13 @@ jobs:
         env:
           GLEAM_COMMAND: gleam
 
+      - name: test/erlang_shipment_no_dev_deps
+        run: ./test.sh
+        working-directory: ./test/erlang_shipment_no_dev_deps
+        if: ${{ matrix.run-integration-tests }}
+        env:
+          GLEAM_COMMAND: gleam
+
       - name: test/project_javascript
         run: |
           gleam run
--- a/compiler-core/src/build/project_compiler.rs
+++ b/compiler-core/src/build/project_compiler.rs
@@ -177,12 +177,12 @@ where
 
     fn retain_only_production_packages(&mut self) {
         let mut production = HashSet::new();
-        let mut queue: Vec<_> = self.config.dependencies.keys().cloned().collect();
+        let mut queue: Vec<_> = self.config.dependencies.keys().collect();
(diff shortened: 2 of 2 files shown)


jtdowney:
Thanks for the feedback. I agree that is a much cleaner way to handle it and I pushed an updated fix.

giacomocavalieri:
Remember to un-draft the PR once it's ready to be reviewed again!

giacomocavalieri:
Hello! There seems to be merge conflicts, would you mind rebasing on main to fix them?

jtdowney:
Done, I've rebased from main

lpil:
Looks like the tests are failing, could you have a look at those please? 🙏 

I'm happy to continue with this approach, but it is a bit wasteful that it compiles the dev dependencies and then does extra work to remove them. Instead the `PackageCompiler` could skip over dev deps when `Mode` is `Prod`, making the build faster and making it impossible for dev code to be used or included. If this is something you'd be interested in doing here then that's fab, if not then could you include a comment and then we can make a new issue for that later.

jtdowney:
Thanks for the pointer regarding `PackageCompiler`, that ended up being even cleaner. I've pushed d371e81 with the new changes.

I dug deeper into the previous test failures. My current thinking is that when it was copying the packages, it wasn't copying the Elixir package, so it failed in some cases. `Error(Elixir(#([110, 111, 32, 115, 117, 99, 104, 32, 102, 105, 108, 101, 32, 111, 114, 32, 100, 105, 114, 101, 99, 116, 111, 114, 121], [101, 108, 105, 120, 105, 114, 46, 97, 112, 112])))` is `no such file or directory` and `elixir.app`.

This new approach likely doesn't have that problem, and I wasn't able to recreate the error with the new diff on the few platforms I tried.

PR #5314, opened 2026-01-23
Fix typescript generation of generic types

### LINKED ISSUE #5312: Generic in TypeScript types are not propagated to subtypes
Hi!

I figured out type-checking is lost in TypeScript declarations for nested types. Let's take an example. If I write in Gleam:

```gleam

pub type Example(value) {
  Example(
    field1: value, 
    field2: Option(value),
  )
}
```

It gets generated to:

```typescript
export class Example<PBN> extends _.CustomType {
  /** @deprecated */
  constructor(field1: PBN, field2: $option.Option$<any>);
  /** @deprecated */
  field1: PBN;
  /** @deprecated */
  field2: $option.Option$<any>;
}
export function Example$Example<PBN>(
  field1: PBN,
  field2: $option.Option$<any>,
): Example$<PBN>;
```

While it should be compiled to:

```typescript
export class Example<PBN> extends _.CustomType {
  /** @deprecated */
  constructor(field1: PBN, field2: $option.Option$<PBN>);
  /** @deprecated */
  field1: PBN;
  /** @deprecated */
  field2: $option.Option$<PBN>;
}
export function Example$Example<PBN>(
  field1: PBN,
  field2: $option.Option$<PBN>,
): Example$<PBN>;
```

You can note it's correctly propagated for direct children of the record, but not for nested types.

GearsDatapacks:
Fixes #5312

I also made sure generic parameters are sorted, as I noticed when reviewing the snapshot that some of the parameters were `<I, J>`, while some were `<J, I>`.

### DIFF
--- a/compiler-core/src/javascript/typescript.rs
+++ b/compiler-core/src/javascript/typescript.rs
@@ -849,7 +849,7 @@ impl<'a> TypeScriptGenerator<'a> {
     /// Converts a Gleam type into a TypeScript type string
     ///
     pub fn print_type(&mut self, type_: &Type) -> Document<'static> {
-        self.do_print(type_, None, false)
+        self.do_print(type_, GenericPrinting::AsAny)
     }
 
     /// Helper function for generating a TypeScript type string after collecting
@@ -860,7 +860,7 @@ impl<'a> TypeScriptGenerator<'a> {
         type_: &Type,
         generic_usages: &HashMap<u64, u64>,
     ) -> Document<'static> {
-        self.do_print(type_, Some(generic_usages), true)
+        self.do_print(type_, GenericPrinting::FromUsage(generic_usages))
     }
 
     /// Get the locally used name for a module. Either the last segment, or the
@@ -883,13 +883,10 @@ impl<'a> TypeScriptGenerator<'a> {
     fn do_print(
         &mut self,
         type_: &Type,
-        generic_usages: Option<&HashMap<u64, u64>>,
-        force_generic_id: bool,
+        generic_printing: GenericPrinting<'_>,
     ) -> Document<'static> {
         match type_ {
-            Type::Var { type_ } => {
-                self.print_var(&type_.borrow(), generic_usages, force_generic_id)
-            }
+            Type::Var { type_ } => self.print_var(&type_.borrow(), generic_printing),
 
             Type::Named {
                 name,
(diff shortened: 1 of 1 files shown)


compiler-core/src/javascript/typescript.rs
```diff
@@ -58,6 +58,7 @@ fn name_with_generics<'a>(
     let generic_usages = collect_generic_usages(HashMap::new(), types);
     let generic_names: Vec<Document<'_>> = generic_usages
         .keys()
+        .sorted()
```
giacomocavalieri, line 61: Shouldn't we make sure that they appear in the same order they appear in the Gleam definition? Is this guaranteed by just using `.sorted()`?

compiler-core/src/javascript/typescript.rs
```diff
@@ -883,32 +884,35 @@ impl<'a> TypeScriptGenerator<'a> {
         &mut self,
         type_: &Type,
         generic_usages: Option<&HashMap<u64, u64>>,
+        force_generic_id: bool,
```
giacomocavalieri: Could you add a comment explaining what this does?

compiler-core/src/javascript/typescript.rs
```diff
@@ -974,6 +980,7 @@ impl<'a> TypeScriptGenerator<'a> {
         name: &str,
         arguments: &[Arc<Type>],
         generic_usages: Option<&HashMap<u64, u64>>,
+        force_generic_id: bool,
```
giacomocavalieri: Same as above, a comment would be useful!

compiler-core/src/javascript/typescript.rs
```diff
@@ -1023,6 +1032,7 @@ impl<'a> TypeScriptGenerator<'a> {
         arguments: &[Arc<Type>],
         module: &str,
         generic_usages: Option<&HashMap<u64, u64>>,
+        force_generic_id: bool,
```
giacomocavalieri: Same here!

compiler-core/src/javascript/typescript.rs
GearsDatapacks: I think it should work fine like this, seeing as it will sort by numeric value, and the compiler generates the new IDs in the same order as they are in the definition.

GearsDatapacks:
I've replaced the boolean parameter with a custom type which is documented to explain what it does

PR #5316, opened 2026-01-23
Add best-effort zero value generation to `Generate Dynamic Decoder` code action

### LINKED ISSUE #5299: Generate dynamic decoder could pick suitable zero value when possible
It would be nice if the "generate dynamic decoder" code action could pick a suitable zero value when generating the decoder if the selected type has a constructor with no arguments:
```gleam
pub type ReviewOutcome {
  Approved
  Rejected
  Commented
}

fn review_outcome_decoder() -> decode.Decoder(ReviewOutcome) {
  use variant <- decode.then(decode.string)
  case variant {
    "approved" -> decode.success(Approved)
    "rejected" -> decode.success(Rejected)
    "commented" -> decode.success(Commented)
    _ -> decode.failure(todo as "zero value for ReviewOutcome", "ReviewOutcome")
  }
}
```
What if, seeing a constructor with no values it could use that directly?
```diff
fn review_outcome_decoder() -> decode.Decoder(ReviewOutcome) {
  use variant <- decode.then(decode.string)
  case variant {
    "approved" -> decode.success(Approved)
    "rejected" -> decode.success(Rejected)
    "commented" -> decode.success(Commented)
-    _ -> decode.failure(todo as "zero value for ReviewOutcome", "ReviewOutcome")
+    _ -> decode.failure(Approved, "ReviewOutcome")
  }
}
```

seafoamteal:
Resolves #5299.

Hi! I've got something that is functional, and would like some feedback on the design.

As I've implemented it now, the code action finds the variant constructor with the fewest fields and attempts to generate a zero value for it. If this is not possible, we revert to the `todo as "Zero value for Type"` fallback that exists currently. I considered some alternative designs such as generating the zero value for every constructor for which it is possible, then choosing the shortest (textually), but wasn't sure about adding that complexity without feedback.

As for when the zero value generation can fail, it only generates zero values for Prelude and stdlib types that the code action can currently generate _decoders_ for (with the addition of `Nil`). By failing on all other types, we don't have to worry about the possibility of recursive types, or loops in type definitions.

### DIFF
--- a/language-server/src/tests/action.rs
+++ b/language-server/src/tests/action.rs
@@ -12176,6 +12176,7 @@ fn interpolate_string_does_not_add_empty_string_right_at_the_start() {
     );
 }
 
+#[test]
 fn generate_dynamic_decoder_produces_zero_values_for_user_defined_type_in_the_same_package_1() {
     let src = r#"
 import wibble.{type Wibble}


language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_1.snap
```diff
@@ -0,0 +1,41 @@
+---
+source: language-server/src/tests/action.rs
+expression: "\nimport wibble.{type Wibble}\n\npub type Wobble {\n  Wobble(value: Wibble)\n  Wimble(a: Int, b: Int)\n}\n    "
+---
+----- BEFORE ACTION
+
+import wibble.{type Wibble}
+                          ↑
```
giacomocavalieri: How come the cursor is here and the action is triggered anyway for the wobble type?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_2.snap
```diff
@@ -0,0 +1,41 @@
...
+source: language-server/src/tests/action.rs
+expression: "\nimport wibble.{type Wibble}\n\npub type Wobble {\n  Wobble(value: option.Option(Wibble))\n  Wimble(a: Int, b: Int)\n}\n    "
+---
+----- BEFORE ACTION
+
+import wibble.{type Wibble}
+                          ↑
+
+pub type Wobble {
+  Wobble(value: option.Option(Wibble))
```
giacomocavalieri: I think it would be nicer if we generated zero values for stdlib types, even if they are not in the prelude

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
```diff
@@ -0,0 +1,41 @@
...
+---
+----- BEFORE ACTION
+
+import wibble.{type Wibble}
+                          ↑
+
+pub type Wobble {
+  Wobble(value: Wibble)
+  Wimble(a: Int, b: Int)
+}
```
giacomocavalieri, line 15: I think it would be really useful for it to also generate zero values for importable user defined values too, in this case using `wibble.Wibble`. @lpil, @GearsDatapacks what do you think?

GearsDatapacks: Yes I think that would be useful

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_2.snap
seafoamteal: It does so for stdlib types, yeah, just in this case the type parameter was not a stdlib type (`Wibble`).

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
seafoamteal: So just types defined in the same package?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_1.snap
seafoamteal: My bad, sorry, I figured selections only moved forward and didn't specify selecting the second `}` in the source.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
seafoamteal: I've been playing around with an implementation of this just to see how it would work, and it's able to generate zero values for non-stdlib and non-prelude types as well, now. It doesn't handle cycle detection yet, so recursive and mutually recursive types are still a problem, but I'm working on that.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_1.snap
lpil: Sorry, I don't understand. What does that mean? It should be able to be triggered from an import.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_user_defined_type_in_the_same_package_1.snap
```diff
@@ -0,0 +1,41 @@
+---
+source: language-server/src/tests/action.rs
+expression: "\nimport wibble.{type Wibble}\n\npub type Wobble {\n  Wobble(value: option.Option(Wibble))\n  Wimble(a: Int, b: Int)\n}\n    "
+---
+----- BEFORE ACTION
+
+import wibble.{type Wibble}
```
lpil, line 7: This shouldn't be able to be triggered from here 🙏

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_2.snap
lpil: Could we still add the optional decoder if not the inner decoder?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_prelude_and_stdlib_types.snap
```diff
@@ -0,0 +1,125 @@
...
+  use variant <- decode.field("type", decode.string)
+  case variant {
+    "wobble" -> {
+      use bit_array <- decode.field("bit_array", decode.bit_array)
+      use int <- decode.field("int", decode.int)
+      use float <- decode.field("float", decode.float)
+      use bool <- decode.field("bool", decode.bool)
+      use list <- decode.field("list", decode.list(decode.int))
+      use string <- decode.field("string", decode.string)
+      use nil <- decode.field("nil", todo as "Decoder for Nil")
```
lpil: What should we do for `nil`? Always succeed? That's what Haskell does

language-server/src/code_action.rs
```diff
@@ -4389,6 +4463,70 @@ impl<'a, 'b> DecoderPrinter<'a, 'b> {
...
+
+    /// Returns the zero values for all Prelude and stdlib types except Result.
+    /// Returns None for all other types.
+    fn zero_value_for_type(&mut self, type_: &Type) -> Option<EcoString> {
+        if type_.is_bit_array() {
+            Some("<<>>".into())
+        } else if type_.is_bool() {
+            Some("False".into())
+        } else if type_.is_float() {
+            Some("0.".into())
```
lpil: The canonical way to write this would be `0.0`

language-server/src/code_action.rs
```diff
@@ -4389,6 +4463,70 @@ impl<'a, 'b> DecoderPrinter<'a, 'b> {
             module = self.printer.print_module(DECODE_MODULE)
         )
     }
+
+    /// Returns the zero values for all Prelude and stdlib types except Result.
+    /// Returns None for all other types.
+    fn zero_value_for_type(&mut self, type_: &Type) -> Option<EcoString> {
```
lpil: This function has lots of nesting, making it harder to read. Please use early returns in each clause to reduce it to a single level of nesting, so the reader doesn't have to hold so much context in their mind. 🙏

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_prelude_and_stdlib_types.snap
seafoamteal: I think `decode.success(Nil)` makes the most sense here too. But if we're adding "support" for `Nil` to the decoder code action, should we also add it to the encoder action? iIRC Haskell encodes `()` to an empty list `[]`; but of course, the value doesn't matter and we can choose whatever is idiomatic in BEAM-land.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_non_prelude_and_stdlib_types_1.snap
seafoamteal: I started the selection at `pub type Wobble` and selected up until a closing brace, except I forgot to use `nth_of_kind` to specify which closing brace in the source I wanted to select. As a result, the selection cursor moved backwards to the closing brace in the import by mistake.

> It should be able to be triggered from an import.

This is not the current behaviour in the stable release of Gleam. This code action is only available when you have selected a type definition and not an import of a type.

seafoamteal:
I've had a go at implementing the enhancement that @giacomocavalieri and @GearsDatapacks suggested: producing zero values for user-defined importable types also. It's currently limited to user-defined types in the same package, but opening it up to all user-defined types would only be a tiny change.
It turned out to be  bit more involved than I thought, but I've tried to structure the code nicely, taking into account the feedback about nesting and early returns.
As for the `Nil` decoder and encoder, I wasn't sure if this PR is the right place for them. I wouldn't mind making those changes once we've settled on a design, but maybe I should open a separate issue for it?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_recursively_defined_constructors_1.snap
```diff
@@ -0,0 +1,40 @@
...
+    "wobble" -> {
+      use inside <- decode.field("inside", wobble_decoder())
+      decode.success(Wobble(inside:))
+    }
+    "dummy" -> {
+      use a <- decode.field("a", decode.int)
+      use b <- decode.field("b", decode.int)
+      decode.success(Dummy(a:, b:))
+    }
+    _ -> decode.failure(todo as "Zero value for Wobble", "Wobble")
```
giacomocavalieri: I think the code action should always produce a zero value if a variant exists. In this case the second one is not recursive and could be a suitable candidate. I'm thinking going over all variants (from least number of fields to most number of fields) until a suitable one is found to be used as a zero value. Does that make sense to you?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_user_defined_type_in_the_same_package_1.snap
```diff
@@ -0,0 +1,44 @@
...
+    "wobble" -> {
+      use value <- decode.field("value", todo as "Decoder for Wibble")
+      decode.success(Wobble(value:))
+    }
+    "dummy" -> {
+      use a <- decode.field("a", decode.int)
+      use b <- decode.field("b", decode.int)
+      decode.success(Dummy(a:, b:))
+    }
+    _ -> decode.failure(Wobble(value: wibble.Wibble), "Wobble")
```
giacomocavalieri, line 42: Really nice!

language-server/src/code_action.rs
```diff
@@ -4389,6 +4469,186 @@ impl<'a, 'b> DecoderPrinter<'a, 'b> {
...
+
+        let type_module_interface = if !type_is_inside_current_module {
+            self.compiler.get_module_interface(custom_type_module)?
+        } else {
+            current_module_interface
+        };
+
+        // We only try and generate zero values for user-defined types in the current
+        // package. If you are expanding the scope of this functionality to
+        // remove this limitation, make sure to check for internal modules.
```
giacomocavalieri, line 4719: Lovely comment

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_recursively_defined_constructors_1.snap
seafoamteal: Yes, makes sense! I can do that.

giacomocavalieri: Lovely, thank you!

seafoamteal:
Checking every constructor instead of just the smallest one was a really simple change! I think I spent more time modifying the tests around this new behaviour :)

language-server/src/code_action.rs
```diff
@@ -4598,56 +4598,62 @@ impl<'a, 'b, IO> DecoderPrinter<'a, 'b, IO> {
...
+        // to construct our zero value to reduce visual noise, but this might not always
+        // be possible. So we check all constructors in increasing order of size to
+        // find the first one that succeeds, and then short circuit.
+        constructors
             .variants
             .iter()
-            .min_by_key(|v| v.parameters.len())?;
+            .sorted_by_key(|v| v.parameters.len())
+            .find_map(|zero_constructor| {
+                path.push((custom_type_module.clone(), custom_type_name.clone()));
```
giacomocavalieri: This is maybe getting a bit too nested, do you think this lambda could be extracted into its own named function?

seafoamteal: Sorry about that, I had a slight suspicion it might be too nested, but wasn't sure enough. Will keep it in mind from now.

giacomocavalieri: No worries! Your code is so good honestly, it's just a minor nitpick :)

giacomocavalieri:
Looks good to me!

seafoamteal:
Hi, I'm so sorry to make a change after you approved the PR, but I was having one last look at it and noticed some redundant logic from the previous implementation sticking around. Don't know how it happened, but I've cleaned it up now 🙏🏽

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
```diff
@@ -0,0 +1,44 @@
...
+    "wobble" -> {
+      use value <- decode.field("value", todo as "Decoder for Wibble")
+      decode.success(Wobble(value:))
+    }
+    "dummy" -> {
+      use a <- decode.field("a", decode.int)
+      use b <- decode.field("b", todo as "Decoder for Wibble")
+      decode.success(Dummy(a:, b:))
+    }
+    _ -> decode.failure(todo as "Zero value for Wobble", "Wobble")
```
lpil, line 42: Should this be this?
```suggestion
    _ -> decode.failure(Wobble(todo as "Zero value for Wibble"), "Wobble")
```

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_generates_todo_for_zero_value_when_all_constructors_fail.snap
```diff
@@ -0,0 +1,40 @@
...
+
+----- AFTER ACTION
+import gleam/dynamic/decode
+
+pub type Wobble {
+  Wobble(nope: Wobble)
+  Wibble(not: Int, again: Wobble)
+}
+
+fn wobble_decoder() -> decode.Decoder(Wobble) {
```
lpil, line 26: I wonder if we should be offering to generate decoders for types that are impossible to create the values of?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_prelude_and_stdlib_types.snap
lpil: Let's use `null`, as that's semantically closer.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
giacomocavalieri: I think if it can't fully generate a zero value it won't try and make a partially built one for you. I think we could go with the current implementation and see how it feels

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_generates_todo_for_zero_value_when_all_constructors_fail.snap
giacomocavalieri: I think making the code more complex to check for those might not be worth the effort

seafoamteal: I agree with Jak on the implementation part, but I also think that since there's no feedback on _why_ the code action isn't available, this might be confusing for developers. Whereas if we leave them a todo, they might more quickly realise that it's impossible while trying to do it manually.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
seafoamteal: Implementation-wise, this is actually the same work as your second comment.

Right now, zero generation fails in three cases:
1. Opaque types
2. Types from a different package
3. Recursive types
Of these, only the last one is truly impossible.

Differentiating between "we won't generate a zero value here because we don't _know_ how to, and so will insert a `todo`" and "we won't generate a zero value here because we _can't_" requires us to now keep track of an extra failure outcome when checking each constructor: success, impossible, and success with todo.

This raises questions like: if all constructor variants will have a `todo` in them, how do we choose which one? The one with the least `todo`s?

I think that at least for now, this would be more work for a user experience that might be less friendly.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_produces_zero_values_for_prelude_and_stdlib_types.snap
seafoamteal: Makes sense! Is it okay if I add it to this PR or would you like a separate one?

seafoamteal:
While thinking about your review comments, I had an idea. Instead of refusing to show this code action to users if it's not possible, why not allow them to try executing it and return an error response if we're unable to, explaining why and what they need to change? In #5093, @giacomocavalieri added better scaffolding to let us return error responses on code actions, which we can use here.

I commented the other day on Discord that we need better documentation on when this action is available (i.e. all fields labeled, no external types). Perhaps being able to tell the user directly in their editor would be a better experience? 

With this, maybe we could also add error messages for un-constructable types.

lpil:
> why not allow them to try executing it and return an error response if we're unable to, explaining why and what they need to change?

I don't think we should have code actions presented if they're not going to work. Offering it says we can do it, so it's misleading and disappointing to then reveal that we cannot.

It would be good to be able to surface more information about when we can do certain actions, but I don't think LSP has a good way to do this yet.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_generates_todo_for_zero_value_when_all_constructors_fail.snap
lpil: OK, sounds good 👍

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_dynamic_decoder_does_not_produce_zero_values_for_types_from_other_packages.snap
lpil: Works for me 👍

seafoamteal:
Makes sense, I get that. I was just working on adding `Nil` to the decoder and encoder code actions, keeping in mind what you said about encoding `Nil` as `"null"`. I have it working, but I realised that there's a problem with this just as I was about to push it. The problem is evident below:

```gleam
import gleam/dynamic/decode
import gleam/json
import gleam/option.{type Option, Some}

pub type OhNo {
  OhNo(val: Option(Nil))
}

fn oh_no_to_json(oh_no: OhNo) -> json.Json {
  let OhNo(val:) = oh_no
  json.object([
    #("val", case val {
      option.None -> json.null()
      Some(_) -> json.null()
    }),
  ])
}

fn oh_no_decoder() -> decode.Decoder(OhNo) {
  use val <- decode.field("val", decode.optional(decode.success(Nil)))
  decode.success(OhNo(val:))
}

pub fn main() {
  let oops = OhNo(val: Some(Nil))
  let encoded_oops = oh_no_to_json(oops) |> json.to_string
  let assert Ok(decoded_oops) = json.parse(encoded_oops, oh_no_decoder())

  assert oops == decoded_oops as "uh oh"
}
```

Because we use `"null"` to represent `None` also, `Some(Nil)` will be incorrectly decoded as `None` after being encoded to `"null"`.

Should we use `[]` like Haskell, then? i.e. `json.preprocessed_array([])`.

Once I fix that, is there anything else this PR is waiting on?

lpil:
`Option(Nil)` isn't representable in JSON like this, so no need to worry about it.

> Should we use [] like Haskell, then? i.e. json.preprocessed_array([]).

Definitely not, it is not an empty list.

Is there anything else or is this ready to review now?

seafoamteal:
Okay! I've just pushed the `Nil` encoder/decoder update, so I believe this is ready for review now.

seafoamteal:
Looks like that job is failing not because of any tests but because the step that fetches Deno timed out.

PR #5322, opened 2026-01-24
Move Hex crate into core repo

lpil:
This reduces the overhead of working with this code, and also puts us in a good position for the Hex OAuth work.

### DIFF
--- a/hexpm/src/lib.rs
+++ b/hexpm/src/lib.rs
@@ -9,12 +9,12 @@ use crate::proto::{signed::Signed, versions::Versions};
 use bytes::buf::Buf;
 use flate2::read::GzDecoder;
 use http::{Method, StatusCode};
-use lazy_static::lazy_static;
 use prost::Message;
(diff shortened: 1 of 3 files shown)


hexpm/src/lib.rs
```diff
@@ -954,6 +954,7 @@ pub struct Dependency {
 static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), " (", env!("CARGO_PKG_VERSION"), ")");
 
 fn validate_package_and_version(package: &str, version: &str) -> Result<(), ApiError> {
+    // TODO: replace lazy_static with OnceCell
```
GearsDatapacks: Is there any reason not to make this change in this PR? Seems like a fairly straightforward change to me, and `TODO` comments rarely get addressed once put into the code from my experience.

lpil: I'll do that now

PR #5336, opened 2026-01-30
Fix completions

giacomocavalieri:
This PR fixes a little quirk with completions. Now accepting a completion for something that has been already (partially) typed won't generate invalid code.
For example, say you're qualifying a type, or fixing a typo:
```gleam
list.f|ter
//    ^ I'm typing here
// If I accept the completion for `list.filter` the resulting code will be:
import gleam/list

list.filter

// Previously it would be:
import gleam/list

list.filterter
```
---

- [X] The changes in this PR have been discussed beforehand with Louis

### DIFF
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -75,10 +75,10 @@ pub fn case<'a>(
                 .into_doc(),
         ),
 
-        tree @ Decision::Run { .. }
-        | tree @ Decision::Guard { .. }
-        | tree @ Decision::Switch { .. }
-        | tree @ Decision::Fail => printer.decision(tree).into_doc(),
+        tree @ (Decision::Run { .. }
+        | Decision::Guard { .. }
+        | Decision::Switch { .. }
+        | Decision::Fail) => printer.decision(tree).into_doc(),
     };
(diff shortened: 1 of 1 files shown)


language-server/src/tests/completion.rs
```diff
@@ -161,8 +161,8 @@ fn format_completion_results(completions: Vec<CompletionItem>) -> EcoString {
         let edit = |buffer: &mut EcoString, e: lsp_types::TextEdit| {
             let a = e.range.start.line;
             let b = e.range.start.character;
-            let c = e.range.start.line;
-            let d = e.range.start.character;
+            let c = e.range.end.line;
+            let d = e.range.end.character;
```
giacomocavalieri, line 165: I noticed the text displayed for completions in snapshot tests has been wrong for quite a while. The fact no one has ever noticed before is a sign this is not a particularly useful format. Maybe we should actually render the completion and show what the code would look like after applying it. What do you think? Could be a different PR to avoid making this one blow up!

lpil: Sounds fab, really nice idea!

Papipo:
Lol, this happens to me all the time, thanks! 🙇🏽

PR #5355, opened 2026-02-06
Add support for workspace symbol queries

### LINKED ISSUE #5191: Support workspace symbols in language server
Would be nice to be able to navigate to types and functions across the whole workspace: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_symbol

I had a brief look at the code and the immediate hurdle seems to be the routing within the language server. For document-level requests we determine the project to operate on from the document's file path. For workspace requests this path is simply not present.

I'd be happy to work on this if we can find a way around that!

seafoamteal:
Resolves #5191.

I've gotten a first draft of workspace symbol support working, as discussed in the linked issue. I'd like some advice on how to go about testing it, however.

The current implementation of the `TestProject` harness is based on the expectation of one single project that lives at the root of the in-memory test file-system. However, the workspace symbol query implementation needs to act on all available engines (since, as discussed, the query params don't include a path) and so acts on all projects known to the `Router`.

Do we want to test this behaviour? If so, I would need to modify the test harness to allow `TestProject`s to be placed at arbitrary locations in the file-system so we can have a test router with multiple projects. If not, and we only want to test this on one open project, then it's straightforward and the tests are almost identical to the document symbol tests.


ankddev (commented):
Hello! What a lovely feature! While I know that it's draft, could we show modifiers and syntax highlight items, like ra does?

Image

For instance, e.g. luals doesn't:

изображение

With current changes, it looks like this:

Image

seafoamteal:
This is the way I see it in Helix.

Rust:
Screenshot of workspace symbols picker in Helix for Rust-Analyser showing symbols alongside information about what kind they are, i.e. struct, enum, package

Gleam:
Screenshot of workspace symbols picker in Helix for Gleam LS showing the same information as Rust-Analyser

The current implementation does send back information about the `kind` of the symbol. In fact, we seem to be sending back _more_ information than RA. I'm not sure what's happening here, to be honest. Syntax highlighting, I'm pretty sure, is an editor/plugin feature, but Neovim (I believe that is what you are using) should be able to pick up the `SymbolKind` information. I will try to set up Neovim and investigate.

ankddev:
The editor on screenshots is Zed, but I will test on Helix and Neovim

seafoamteal:
@lpil, just bumping this. Would it be helpful if I made a PoC by editing the `TestProject` harness to support multiple Gleam projects in a single in-memory filesystem?
I've been playing around with a `TestWorkspace` implementation; if it would help decide hash out the design, I can polish it up and add it to the draft PR.

ankddev:
Sorry for such a long waiting, really forgot about this.

> The current implementation does send back information about the `kind` of the symbol. In fact, we seem to be sending back _more_ information than RA. I'm not sure what's happening here, to be honest. Syntax highlighting, I'm pretty sure, is an editor/plugin feature, but Neovim (I believe that is what you are using) should be able to pick up the `SymbolKind` information. I will try to set up Neovim and investigate.

I checked it recently and this highlighting seems to be a feature we would want to implement in the Gleam Zed extension, and not in the LS

PR #5360, opened 2026-02-10
fix erlang code generation with 0 literal denominator

giacomocavalieri:
We discussed this on Discord, a quick and easy fix!


lpil (approved):
Oh much better! Thank you!!

PR #5362, opened 2026-02-11
Improve "Interpolate String" code action

### LINKED ISSUE #5263: LS: "Interpolate string" code action could allow more string selections
Currently, the "Interpolate string" code action only allows the interpolation of selections that match a valid Gleam identifier (e.g `name`). Because of this, it doesn't allow the extraction of record accesses, function calls, references to other module's constants, and so on.

I believe allowing any valid expression within the selection would be the most robust solution, but even adding support for specific cases like record access (`user.name`) or function calls (`process_string(my_string, options:)`) would be a significant improvement to the code action.

seafoamteal:
This resolves #5263.

I've changed the code action so that you can now interpolate any part of string, not just valid lowercase identifiers.

While working on the above, I noticed that if you interpolate an expression at the very start of/very end of a string, a redundant empty string was left in the resulting code; I made some changes to get rid of this. I hope that's okay; if not, I can revert it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,11 +155,12 @@
   ([Vladislav Shakitskiy](https://github.com/vshakitskiy))
 
 - The "Interpolate String" code action now lets the user "cut out" any portion
-  of a string, regardless of whether it is a valid Gleam identifier or not. In
-  addition, when interpolating an expression at the very start or the very end
-  of a string, redundant empty string are no longer added before/after the
-  interpolated expression.
+  of a string, regardless of whether it is a valid Gleam identifier or not.
+  ([Hari Mohan](https://github.com/seafoamteal))
 
+- When interpolating an expression at the very start or the very end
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -154,6 +154,14 @@
 
   ([Vladislav Shakitskiy](https://github.com/vshakitskiy))
 
+- The "Interpolate String" code action now lets the user "cut out" any portion
+  of a string, regardless of whether it is a valid Gleam identifier or not. In
+  addition, when interpolating an expression at the very start or the very end
+  of a string, redundant empty string are no longer added before/after the
+  interpolated expression.
```
giacomocavalieri: Can you make this a separate changelog entry?

seafoamteal: Makes sense, I've split them apart now!

giacomocavalieri:
Lovely, thank you!

PR #5365, opened 2026-02-12
Improve the look of the hover snapshots

giacomocavalieri:
Reading some of the hover snapshots I figured we could do a much better job at displaying their content in the snapshot tests. So I tried to improve them: now the content is rendered as a regular string (rather than on a single line with all characters escaped) and we show what kind of content we're dealing with (markdown or plain text)


lpil (approved):
Much better! Thank you

PR #5367, opened 2026-02-13
Make sure the compiler always rejects duplicate modules

### LINKED ISSUE #5352: Gleam compiles even if two packages provide the same imports
In the following eamples, moden should't compile:

https://github.com/weedonandscott/lustre_platform/tree/main/examples/04-applications/01-routing

giacomocavalieri:
This PR closes #5352 
Now the compiler should always detect when a dependency is defining a module with the same name as the root project, or even when two imported dependencies are defining a module with the same name.

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -3,7 +3,7 @@ mod tests;
 
 use crate::analyse::{ModuleAnalyzerConstructor, TargetSupport};
 use crate::build::package_loader::CacheFiles;
-use crate::config::PackageKind;
+
 use crate::error::DefinedModuleOrigin;
 use crate::inline;
 use crate::io::files_with_extension;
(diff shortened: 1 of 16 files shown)


compiler-core/src/build/package_compiler.rs
```diff
@@ -73,6 +80,7 @@ where
     IO: FileSystemReader + FileSystemWriter + CommandExecutor + BeamCompiler + Clone,
 {
     pub fn new(
+        package_kind: PackageKind,
```
lpil: 🔴 This is introducing build-tool level concepts to the lower level package compiler API. There is no such thing as a package kind at this low level, they are all just packages! This is important because this code is used outside of the build tool too.

Any behaviour needs to be configured in a way that makes sense at this level, as you can see with the other configuration this class has, like `write_entrypoint`.

I've not reviewed fully yet, but I think this is only used for constructing an error message. If that's right we could instead store the package name always.

giacomocavalieri: Ok, I've removed it!

PR #5369, opened 2026-02-15
Fix crash on record update if constructor definition is invalid

### LINKED ISSUE #5296: Gleam Language Server crash on Visual Studio Code open
Upon opening VS Code I get this error:

>>> The Gleam Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted. See the output for more information.

And this is the output for more information: [Pastebin](https://pastebin.com/wuvf9E34)

And the version of VS Code I'm using, if that helps:

Version: 1.108.1 (Universal)
Commit: 585eba7c0c34fd6b30faac7c62a42050bfbc0086
Date: 2026-01-14T14:55:44.241Z
Electron: 39.2.7
ElectronBuildId: 12953945
Chromium: 142.0.7444.235
Node.js: 22.21.1
V8: 14.2.231.21-electron.0
OS: Darwin arm64 24.6.0

seafoamteal:
Fixes #5296.

Replacing the `unwrap` by silently skipping over the error seemed weird at first, but I thought about it more, and given that this can only be caused by an error in the type definition itself, I thought that it was okay to not surface an error to the user here. My reasoning was that we are already reporting the misplaced positional field, which is the root error, so it's best not to clutter diagnostics with too many downstream errors.

### DIFF
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_record_update_when_constructor_definition_is_invalid.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_record_update_when_constructor_definition_is_invalid.snap
@@ -22,11 +22,3 @@ error: Unlabelled argument after labelled argument
   │                          ^^^^
 
 All unlabelled arguments must come before any labelled arguments.
-
-error: Unlabelled argument after labelled argument
-  ┌─ /src/one/two.gleam:8:13
-  │
-8 │   let two = Wibble(..one, b: 1)
-  │             ^^^^^^^^^^^^^^^^^^^
-
-All unlabelled arguments must come before any labelled arguments.
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -3256,13 +3256,14 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
                 // all correspond to positional fields. If this is not the case, then the
                 // user currently has a mistake in the definition of their custom type.
                 //
-                // For example, when a user starts adding a new labelled field to a custom
-                // type constructor, the field label will be parsed as a custom type name
-                // until they type the colon i.e. as a new positional field.
-                let type_ = positional_fields
-                    .get(index as usize)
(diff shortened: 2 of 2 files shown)


seafoamteal:
I slept on it and realised quashing the error like that didn't seem like a good idea at all, so I went back to reporting an error. The only `Error` that made sense, because it's actually the root cause, is `UnlabelledAfterLabelled`.
My only concern is that when you just look at the error at the record update site, it's not immediately obvious what's going on. Would it make sense to make it a _syntax_ error if there's an unlabelled field after a labelled field? Something like:

```gleam
pub type Wibble {
  Wibble(Int, a: Int, Bool)
//                    ^^^^
//                    Syntax error: Expected one of
//                    - ')'
//                    - A labelled field 
}
```

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__no_crash_on_record_update_when_constructor_definition_is_invalid.snap
```diff
@@ -0,0 +1,32 @@
...
+
+All unlabelled arguments must come before any labelled arguments.
+
+error: Unlabelled argument after labelled argument
+  ┌─ /src/one/two.gleam:8:13
+  │
+8 │   let two = Wibble(..one, b: 1)
+  │             ^^^^^^^^^^^^^^^^^^^
+
+All unlabelled arguments must come before any labelled arguments.
```
giacomocavalieri: I think if the constructor definition is wrong there's no point in showing this error message where it's used. Also this looks very confusing in the record update syntax.

giacomocavalieri (commented):
I think skipping it is better. The example test you have is really confusing. I think in this case it's fine to skip the error and only report the one happening in the constructor.

seafoamteal:
Thanks for confirming my initial intuition. I've gone back to the version that skips showing the error message at this site.

The reason I thought showing an error is the correct course of action is because I was unsure if returning an `Ok` result with incomplete inferred arguments would cause a similar crash elsewhere in the type inference pipeline. I spent some time walking through the references of this field again, and I think it should be fine actually.

PR #5372, opened 2026-02-15
Do not allow publishing with default readme

giacomocavalieri:
As we discussed, we want to discourage publishing packages with the default README. That is just a placeholder, and a carefully written README shouldn't be an afterthought for packages published on Hex, hopefully this will reduce the number of published packages doing this!


GearsDatapacks:
Make sure you `.gitignore` the `build` directory of new projects, because currently you're including the entire stdlib in the PR.

PR #5373, opened 2026-02-18
Support various LSP operations on clause guards

### LINKED ISSUE #4943: Cannot go to definition, find references, etc. from clause guard
While fixing #4935 I noticed clause guards had the same issue:

```gleam
case wibble {
  Wibble if wobble > 2 -> Nil
  //         ^- Cannot go to do definition, find references from here
}
```

To fix this, one would have to

1. Add an `or_else` case to traverse the clause guard after the alternative patterns in [`TypedClause.find_node`](https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/ast.rs#L1616) (easy peasy)
2. Implement `find_node` for `ClauseGuard`

Step 2 is where I was as a bit out of my depth, but maybe someone else will have more success.

GearsDatapacks:
Closes #4943

This PR adds goto definition, hover, rename and find references functionality to guards. I haven't implemented autocomplete just yet as that requires making guards fault tolerant which would introduce loads of extra changes, and I felt that this PR was getting quite big already.

For hover, I just did the same thing we do for constants which is to only show the type of the hovered value. Both could do with improving to show documentation and other details which are shown for expressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,10 @@
   now generate a decoder and an encoder, respectively, for `Nil` values.
   ([Hari Mohan](https://github.com/seafoamteal))
 
+- The language server now supports renaming, go to definition, hover, and
+   finding references from expressions in case clause guards.
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Formatter
 
 ### Bug fixes


giacomocavalieri (approved):
Beautiful!! Could you rebase on main?

GearsDatapacks:
Oops, I somehow only just saw this. Rebased!

PR #5375, opened 2026-02-19
Delete compile file after compilation

### LINKED ISSUE #4474: Don't put gleam@@compile.erl in the _gleam_artefacts dir

ankddev:
Closes https://github.com/gleam-lang/gleam/issues/4474. I hope that it's right place to place, but I'm not sure, since why error while deleting file should fail the entire run?
***


compiler-cli/src/beam_compiler.rs
```diff
@@ -152,3 +155,8 @@ impl Drop for BeamCompiler {
 fn escape_path<T: AsRef<str>>(path: T) -> String {
     path.as_ref().replace("\\", "\\\\")
 }
+
+fn get_escript_path(out: &Utf8Path) -> Utf8PathBuf {
+    out.join(paths::ARTEFACT_DIRECTORY_NAME)
+        .join("gleam@@compile.erl")
+}
```
lpil, line 204: Could you move this to the paths module please 🙏  Bit naughty that it wasn't in there to start with, I guess I made a mistake.

compiler-cli/src/beam_compiler.rs
```diff
@@ -66,6 +66,11 @@ impl BeamCompiler {
         let mut buf = String::new();
         let mut accumulated_modules: Vec<String> = Vec::new();
         while let (Ok(_), Ok(None)) = (inner.stdout.read_line(&mut buf), inner.process.try_wait()) {
+            let escript_path = get_escript_path(out);
+
+            // Delete unnecessary after compilation escript file
```
lpil: Delete unnecessary?

test/delete_escript_after_compilation/Makefile
```diff
@@ -0,0 +1,10 @@
+# TODO: migrate to Rust shell commands, possibly ./compiler-cli/src/fs/tests.rs
+test:
+	# clean build directory && build the project
+	cargo run clean && cargo run build
+
+	# fail if file exists
+	@if find build -name "gleam@@compile.erl" | grep -q .; then \
+		echo "Error: gleam@@compile.erl file(s) still exist after build"; \
+		exit 1; \
+	fi
```
lpil: Write integration tests in shell please, using the `test.sh` convention 🙏

lpil (commented):
Thank you! Few tiny things inline

compiler-cli/src/beam_compiler.rs
ankddev: Yeah, my mistake as well that I haven't placed it there!

test/delete_escript_after_compilation/Makefile
ankddev: Between `Makefile` variant and `test.sh` variant, I have chosen the former, don't know why :) I will rewrite it

compiler-core/src/paths.rs
```diff
@@ -133,6 +133,10 @@ impl ProjectPaths {
     }
 }
 
+pub fn compilation_escript_path(out: &Utf8Path) -> Utf8PathBuf {
+    out.join(ARTEFACT_DIRECTORY_NAME).join("gleam@@compile.erl")
+}
+
```
ankddev, line 151: @lpil I've moved this function to `paths` module. I've thought about placing it under `ProjectPaths`, but `BeamCompiler` and its callers don't receive `ProjectPaths`, they receive just `out`, so this would require some refactoring. What are your thoughts?

ankddev:
Oh, it seems that I forgot to add a changelog entry, will do it!

ankddev:
It's done now!

.github/workflows/ci.yaml
```diff
@@ -481,7 +481,11 @@ jobs:
         run: make test
         working-directory: ./test/hextarball
 
-      - name: test/running_modules
+      - name: Test deletion of escript compile file after compilation
+        run: ./test.sh
+        working-directory: ./test/delete_escript_after_compilation
+
+      - name: Test running modules
```
lpil: Undo the renaming of this step please

compiler-cli/src/beam_compiler.rs
```diff
@@ -66,6 +66,11 @@ impl BeamCompiler {
         let mut buf = String::new();
         let mut accumulated_modules: Vec<String> = Vec::new();
         while let (Ok(_), Ok(None)) = (inner.stdout.read_line(&mut buf), inner.process.try_wait()) {
+            let escript_path = paths::compilation_escript_path(out);
+
+            // Delete escript file, which is unnecessary after compilation
+            io.delete_file(&escript_path)?;
```
lpil: This attempts to delete the file on every single line of output! We only want to do it once.

compiler-core/src/paths.rs
lpil: Nice, thank you!

lpil (commented):
Very nice, just some tiny notes inline 🙏

ankddev:
Thank you for review, I will do it tomorrow!

ankddev:
@lpil Done!

lpil (commented):
Thank you! This is currently deleting the file after each batch of compilation, but really we only want to delete it once. It seems the best time to do this would be after it has been spawned but before it has been used to compile anything.

ankddev:
@lpil

> Thank you! This is currently deleting the file after each batch of compilation, but really we only want to delete it once. It seems the best time to do this would be after it has been spawned but before it has been used to compile anything.

Sorry, but I don't quite understand this. Do you mean that it should be deleted right after spawning a process? But it will still delete it for each compilation, won't it?

PR #5376, opened 2026-02-19
Support textDocument/foldingRange for imports and multiline top-level definitions

### LINKED ISSUE #5370: Support textDocument/foldingRange in language server
It would be nice if you could fold long import statements. 

Example:
```gleam
import gleam/int
import gleam/json
import gleam/list
```

Turned into:
```gleam
import gleam/int ⋯
```

aayush-tripathi:
Closes #5370 

This PR adds textDocument/foldingRange support to the Gleam language server.

It currently supports folding for:

- contiguous import blocks

- multiline function bodies
- multiline custom type definitions
- multiline constant definitions
- multiline type aliases

**Sample Screenshot**

_Before_
Screenshot 2026-02-19 213107

_After_
Screenshot 2026-02-19 221623

### DIFF
--- a/language-server/src/engine.rs
+++ b/language-server/src/engine.rs
@@ -619,7 +619,7 @@ where
         params: lsp::FoldingRangeParams,
     ) -> Response<Vec<FoldingRange>> {
         self.respond(|this| {
-            let mut ranges: Vec<(u32, FoldingRange)> = vec![];
+            let mut ranges: Vec<FoldingRange> = vec![];
             let Some(module) = this.module_for_uri(&params.text_document.uri) else {
                 return Ok(vec![]);
             };
@@ -635,7 +635,7 @@ where
                     continue;
                 };
 
-                ranges.push((import.start, range));
+                ranges.push(range);
             }
 
             for type_ in &module.ast.definitions.custom_types {
@@ -643,7 +643,7 @@ where
                 let Some(range) = folding_range_for_span(span, &line_numbers, None) else {
                     continue;
                 };
-                ranges.push((span.start, range));
+                ranges.push(range);
             }
 
             for constant in &module.ast.definitions.constants {
@@ -651,7 +651,7 @@ where
                 let Some(range) = folding_range_for_span(span, &line_numbers, None) else {
                     continue;
                 };
-                ranges.push((span.start, range));
+                ranges.push(range);
             }
 
             for alias in &module.ast.definitions.type_aliases {
@@ -659,7 +659,7 @@ where
                 let Some(range) = folding_range_for_span(span, &line_numbers, None) else {
                     continue;
                 };
-                ranges.push((span.start, range));
+                ranges.push(range);
             }
 
             for function in &module.ast.definitions.functions {
@@ -671,11 +671,11 @@ where
                 let Some(range) = folding_range_for_span(span, &line_numbers, None) else {
                     continue;
                 };
-                ranges.push((span.start, range));
+                ranges.push(range);
             }
 
-            ranges.sort_by_key(|(start, _)| *start);
-            Ok(ranges.into_iter().map(|(_, range)| range).collect())
+            ranges.sort_by_key(|range| range.start_line);
+            Ok(ranges)
         })
     }
 


aayush-tripathi:
[test (stable, x86_64-apple-darwin)](https://github.com/gleam-lang/gleam/actions/runs/22192455208/job/64183805510?pr=5376#logs) The test failed after not being able to install gleam for 30 minutes. How do I re-run this?

language-server/src/tests/snapshots/gleam_language_server__tests__folding_range__does_not_fold_single_line_top_level_definitions.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: language-server/src/tests/folding_range.rs
+expression: "pub type W { W(Int) }\npub const x = 1\npub type Pair = #(Int, Int)\npub fn main() { 1 }"
+snapshot_kind: text
+---
+----- Code -----
+pub type W { W(Int) }
+pub const x = 1
+pub type Pair = #(Int, Int)
+pub fn main() { 1 }
```
giacomocavalieri: Could you split these in four different tests? One for each kind of top level definition? And the one for functions is already defined I think

language-server/src/tests/snapshots/gleam_language_server__tests__folding_range__folds_import_block.snap
```diff
@@ -0,0 +1,19 @@
...
+import two
+import three
+
+pub fn main() { 1 }
+
+----- Ranges -----
+1. lines 0..2 kind: imports
+
+----- Fold 1 (lines 0..2) -----
+import one ...
```
giacomocavalieri, line 17: Really nice snapshot, well done!

language-server/src/tests/folding_range.rs
```diff
@@ -0,0 +1,221 @@
...
+        response.result.unwrap()
+    })
+}
+
+fn kind_name(kind: &Option<FoldingRangeKind>) -> &'static str {
+    match kind {
+        Some(kind) if *kind == FoldingRangeKind::Imports => "imports",
+        Some(kind) if *kind == FoldingRangeKind::Comment => "comment",
+        Some(kind) if *kind == FoldingRangeKind::Region => "region",
+        Some(_) => "other",
```
giacomocavalieri: Pattern match on the value directly rather than using a case guard. That way the compiler can help detecting unmatched values and using a catch all is no longer needed:

```suggestion
        Some(FoldingRangeKind::Imports) => "imports",
        Some(FoldingRangeKind::Comment) => "comment",
        Some(FoldingRangeKind::Region) => "region",
```

language-server/src/tests/folding_range.rs
```diff
@@ -0,0 +1,221 @@
...
+
+#[test]
+fn folds_multiline_function_body() {
+    let code = "pub fn main() {
+  let x = 1
+  x
+}";
+
+    let output = folding_snapshot_output(TestProject::for_source(code));
+    insta::assert_snapshot!(insta::internals::AutoName, output, code);
```
giacomocavalieri: Could you define a macro for writing these tests in the same spirit of the various `assert_code_action!`, `assert_hover!` and such? So it could look something like this:

```suggestion
    assert_folding!("pub fn main() {
 let x = 1
 x
}");
```

language-server/src/engine.rs
```diff
@@ -1109,6 +1174,74 @@ Unused labelled fields:
...
+    let mut current_start = first_import.location.start;
+    let mut current_end = first_import.location.end;
+    let mut current_len = 1;
+
+    for import in imports {
+        let previous_span = SrcSpan::new(current_end, current_end);
+        let next_span = SrcSpan::new(import.location.start, import.location.start);
+        let previous_line = src_span_to_lsp_range(previous_span, line_numbers)
+            .start
+            .line;
```
giacomocavalieri: Rather than keeping track of the `current_end` couldn't you keep track of the `previous_line` directly? I think it might make this a bit easier to follow

aayush-tripathi:
Thanks. I have made some changes

language-server/src/tests/folding_range.rs
```diff
@@ -81,23 +80,30 @@ fn folding_snapshot_output(project: TestProject<'_>) -> String {
     pretty_folding_output(src, &ranges)
 }
 
+macro_rules! assert_folding {
+    ($src:literal $(,)?) => {
+        let output = folding_snapshot_output(TestProject::for_source($src));
+        insta::assert_snapshot!(insta::internals::AutoName, output, $src);
+    };
+    ($project:expr, $src:expr $(,)?) => {
+        let output = folding_snapshot_output($project);
```
giacomocavalieri: Rather than passing the `src` around along with the test project you could instead access it from the `test_project.src` field. This way the api ends up being a bit nicer and you don't have to pass the src around twice

language-server/src/tests/snapshots/gleam_language_server__tests__folding_range__does_not_fold_single_line_constant.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: language-server/src/tests/folding_range.rs
+expression: pub const x = 1
+snapshot_kind: text
+---
+----- Code -----
+pub const x = 1
+
+----- Ranges -----
+(none)
```
giacomocavalieri: Sorry I just realised this, but for these tests were we expect to be no foldable ranges could you write a macro `assert_no_folding!` that asserts the array of ranges is empty rather than making a snapshot? I think it would make it easier to check and also it follows the compiler's convention (like with `assert_warning`/`assert_no_warnings`, `assert_code_action`/`assert_no_code_action` and such)

giacomocavalieri:
Awesome work! Looks good to be merged to me

PR #5378, opened 2026-02-20
Fallback to hardlinking if can't symlink on Windows w/o dev mode

### LINKED ISSUE #3015: `gleam build` doesn't have permission to link `priv` on Windows
On Windows running `gleam build` on a project with a `priv` directory will error. This seems to be caused by a Windows-specific symlink behaviour.
https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/

You can enable "Developer Mode", which will grant the symlink permission.
image
If there's no easy way to fix this, then printing out a warning to enable this setting would probably be okay.

Using Gleam v1.1.0

Reproducing:
```
gleam new project_name
cd project_name
mkdir priv
gleam build
```
```
> gleam build
Downloading packages
 Downloaded 2 packages in 0.02s
  Compiling gleam_stdlib
  Compiling gleeunit
  Compiling project_name
error: File IO failure

An error occurred while trying to link this file:

    C:\Users\USERNAME\Desktop\project_name\build\dev\erlang\project_name\priv

    A required privilege is not held by the client. (os error 1314)
```

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/3015

Hardlinking is only supported for files on most filesystems, so it falls back to copying directory.
I haven't written a test, since it's hard to write that test and the logic is quite trivial.
***

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -705,10 +705,18 @@ pub fn copy_dir(path: impl AsRef<Utf8Path>, to: impl AsRef<Utf8Path>) -> Result<
     .map(|_| ())
 }
 
-pub fn symlink_dir(src: impl AsRef<Utf8Path>, dest: impl AsRef<Utf8Path>) -> Result<(), Error> {
+/// Symlink directory.
+/// If it couldn't symlink directory on Windows because of error 1314, which
+/// occurs without Developer Mode enabled, it falls back to hardlinking each
+/// file in directory.
+///
+pub fn symlink_dir(
+    src: impl AsRef<Utf8Path> + Debug,
+    dest: impl AsRef<Utf8Path> + Debug,
+) -> Result<(), Error> {
+    tracing::debug!(src=?src, dest=?dest, "symlinking");
     let src = src.as_ref();
     let dest = dest.as_ref();
-    tracing::debug!(src=?src, dest=?dest, "symlinking");
     let src = canonicalise(src)?;
 
     #[cfg(target_family = "windows")]
@@ -727,7 +735,7 @@ pub fn symlink_dir(src: impl AsRef<Utf8Path>, dest: impl AsRef<Utf8Path>) -> Res
         Err(err) => Err(Error::FileIo {
             action: FileIoAction::Link,
             kind: FileKind::File,
-            path: Utf8PathBuf::from(dest.as_ref()),
+            path: Utf8PathBuf::from(dest),
             err: Some(err.to_string()),
         }),
     }
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -719,8 +719,9 @@ pub fn symlink_dir(src: impl AsRef<Utf8Path>, dest: impl AsRef<Utf8Path>) -> Res
     match result {
         Ok(()) => Ok(()),
 
-        // Fallback to hardlinking if we can't symlink. This occurs on Windows without Developer Mode enabled.
-        // We match on the raw OS code, since this error has no specific error kind.
+        // Fallback to hardlinking if we can't symlink. This occurs on Windows
+        // without Developer Mode enabled. We match on the raw OS code, since
+        // this error has no specific error kind.
         #[cfg(target_family = "windows")]
         Err(err) if err.raw_os_error() == Some(1314) => hardlink_dir(src, dest),
         Err(err) => Err(Error::FileIo {
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -198,6 +198,10 @@ impl FileSystemWriter for ProjectIO {
         hardlink(from, to)
     }
 
+    fn hardlink_dir(&self, from: &Utf8Path, to: &Utf8Path) -> Result<(), Error> {
+        hardlink_dir(from, to)
+    }
+
     fn symlink_dir(&self, from: &Utf8Path, to: &Utf8Path) -> Result<(), Error> {
         symlink_dir(from, to)
     }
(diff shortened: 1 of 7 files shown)


lpil:
Copying is too expensive, it needs to be hard-linking!

I think we also should only hard-link when the symbolic-linking fails due for this specific reason on Windows, instead of for any IO error on any OS.

Files that are removed will also need to be removed from the build directory, otherwise folks can get cryptic bugs due to drift between their source directory and their build directory.

ankddev:
@lpil Thanks for review, I have few questions on this.

> Copying is too expensive, it needs to be hard-linking!

So it should walk directory and hardlink each file?

> I think we also should only hard-link when the symbolic-linking fails due for this specific reason on Windows, instead of for any IO error on any OS.

This seems reasonable to me.

> Files that are removed will also need to be removed from the build directory, otherwise folks can get cryptic bugs due to drift between their source directory and their build directory.

Is it good enough just deleting directory each time and re-hardlinking? Or it should check each file and just delete unnecessary and hardlink new/changed?

lpil:
> So it should walk directory and hardlink each file?

I think that may be the only option in this case.

> Is it good enough just deleting directory each time and re-hardlinking? Or it should check each file and just delete unnecessary and hardlink new/changed?

Do you know how each of these compare in terms of performance? I'm not sure myself!

ankddev:
@lpil

> Do you know how each of these compare in terms of performance? I'm not sure myself!

I will implement both and experiment with it, but probably on small dirs the former will be faster (and waaaay easier to implement), while on very large ones the latter will be faster.

ankddev:
@lpil I've implemented hardlinking with first approach and it works quite fast. I've checked second as well and I don't think if it's worth implemnting it.

compiler-cli/src/fs.rs
```diff
@@ -657,17 +661,24 @@ pub fn symlink_dir(
...
-        kind: FileKind::File,
-        path: Utf8PathBuf::from(dest.as_ref()),
-        err: Some(err.to_string()),
-    })?;
-    Ok(())
+    match result {
+        Ok(()) => Ok(()),
+
+        // Fallback to hardlinking if we can't symlink. This occurs on Windows without Developer Mode enabled.
+        // We match on the raw OS code, since this error has no specific error kind.
```
lpil: Wrap long comment lines please.

compiler-cli/src/fs.rs
```diff
@@ -685,6 +696,50 @@ pub fn hardlink(
         .map(|_| ())
 }
 
+/// Hardlink directory.
+/// This is done by deleting destination directory if it exists and recursively
+/// creating directories and hardlinking files inside the directory.
+///
+pub fn hardlink_dir(
```
lpil: Let's avoid abbreviations rather than copying the old code that uses patterns we want to avoid

```suggestion
pub fn hardlink_directory(
```

compiler-cli/src/fs.rs
```diff
@@ -685,6 +696,50 @@ pub fn hardlink(
...
+    src: impl AsRef<Utf8Path> + Debug,
+    dest: impl AsRef<Utf8Path> + Debug,
+) -> Result<(), Error> {
+    tracing::trace!(src=?src, dest=?dest, "hardlinking_dir");
+
+    let src = src.as_ref();
+    let dest = dest.as_ref();
+
+    // Recreate destination directory to avoid getting it in weird state.
```
lpil: Explain what a "weird state" is please.

compiler-cli/src/fs.rs
```diff
@@ -685,6 +696,50 @@ pub fn hardlink(
...
+
+        let file_name = path.file_name().ok_or_else(|| Error::NonUtf8Path {
+            path: path.to_path_buf().into(),
+        })?;
+
+        let destination = &dest.join(file_name);
+
+        if path.is_dir() {
+            // Recursively hardlink directory.
+            hardlink_dir(path, destination)?;
```
lpil: Use a vec for this rather than the stack please 🙏

compiler-cli/src/fs.rs
```diff
@@ -657,17 +661,24 @@ pub fn symlink_dir(
...
+    let result = std::os::unix::fs::symlink(&src, dest.as_ref());
 
-    result.map_err(|err| Error::FileIo {
-        action: FileIoAction::Link,
-        kind: FileKind::File,
-        path: Utf8PathBuf::from(dest.as_ref()),
-        err: Some(err.to_string()),
-    })?;
-    Ok(())
+    match result {
```
lpil, line 727: This function is called "symlink directory", but it can create hardlinks!

lpil (commented):
Great, thank you.

This design works for files that exist the first time the compiler is run, but it looks like for any new files to be picked up the programmer has to run `gleam clean && gleam run`. Is that right?

ankddev:
> This design works for files that exist the first time the compiler is run, but it looks like for any new files to be picked up the programmer has to run `gleam clean && gleam run`. Is that right?

No, every time it walks original directory, deletes linked dirctory and then hardlinks every file.

Above we have talked about another approach: don't delete the entire directory and check files that need to be added and that need to be deleted, but it will be harder to implement.

compiler-cli/src/fs.rs
ankddev: Is `symlink_or_hardlink` good? I can't imagine about better name 😆

lpil:
Sorry, I don't understand how that happens, as the modified function only gets called once. What am I missing?

ankddev:
@lpil Oh, sorry, I checked it now and it requires cleaning. I don't know why I've thoughts that it should work... Should we leave it as is?

compiler-cli/src/fs.rs
ankddev: If I understood you correctly, then it's done. I've refactored it to use `VecDeque` instead of being recursive. I hope that it's right

compiler-cli/src/fs.rs
ankddev: I've added a doc comment explaining this

giacomocavalieri:
Is this ready for review now?

ankddev:
@giacomocavalieri I think that it is!

compiler-cli/src/fs.rs
lpil: It's still called symlink directory!

lpil (commented):
With this implementation if a file is removed from the source directory it will still be present in the created directory, potentially resulting in invalid results until `gleam clean` is ran.

Some amount of bookkeeping is going to need to be performed to avoid this problem, but there is a cost associated with that. Perhaps we want to track this for directories that will change (root, path deps) but not for those that do not (Elixir installation, Hex and git dependencies). Unsure if this is the best approach.

ankddev:
@lpil Hello! Thank you for review and sorry for so long waiting.

> Some amount of bookkeeping is going to need to be performed to avoid this problem, but there is a cost associated with that. Perhaps we want to track this for directories that will change (root, path deps) but not for those that do not (Elixir installation, Hex and git dependencies). Unsure if this is the best approach.

So, what do you think about calling function to symlink or hardlink directory (`priv`) on each build for root package and path deps? Or will it be better to check if content differs, and only re-call function in that case? I'll experiment with this too.

PR #5381, opened 2026-02-21
LSP: cleanup when a client has closed all files in a project

### LINKED ISSUE #3205: LSP: cleanup when a client has closed all files in a project
When the client has closed all files for a Gleam project:

- Unset all the diagnostics for that project
- Discard the engine for that project

TrippleCCC:
Closes #3205 

The language server now keeps track of open files and removes the project engine once all open project files have been closed. 

The issue also mentions clearing all diagnostics; I believe this should already be the case since `discard_in_memory_cache` returns an empty feedback.

It looks like `LanguageServer` is not set up for testing. I would be happy to create tests for this if it is necessary.

---

- [x] Tests have been added for new behavior

### DIFF
--- a/language-server/src/server.rs
+++ b/language-server/src/server.rs
@@ -149,22 +149,25 @@ where
     }
 
     fn attempt_engine_cleanup(&mut self, path: &Utf8PathBuf, feedback: &mut Feedback) {
-        if let Some(project_path) = self.router.project_path(path)
-            && self
+        if let Some(project_path) = self.router.project_path(path) {
+            let all_files_closed = self
                 .opened_files
                 .iter()
                 .filter(|(path, _)| path.starts_with(&project_path))
-                .all(|(_, is_open)| !*is_open)
-        {
-            self.router.delete_engine_for_path(&project_path);
-            _ = self.changed_projects.remove(&project_path);
-            self.opened_files
-                .extract_if(|path, _| path.starts_with(&project_path))
-                .map(|(path, _)| path)
-                .into_iter()
-                .for_each(|path| {
+                .all(|(_, is_open)| !*is_open);
+
+            if all_files_closed {
+                self.router.delete_engine_for_path(&project_path);
+                _ = self.changed_projects.remove(&project_path);
+
+                for (path, _) in self
+                    .opened_files
+                    .extract_if(|path, _| path.starts_with(&project_path))
+                    .into_iter()
+                {
                     feedback.unset_existing_diagnostics(path);
-                });
+                }
+            }
         }
     }
 


CHANGELOG.md
```diff
@@ -175,6 +175,9 @@
   now generate a decoder and an encoder, respectively, for `Nil` values.
   ([Hari Mohan](https://github.com/seafoamteal))
 
+- The language server now cleans up compilation engines when all project
+  files are closed ([Nonso Chukwurah](https://github.com/TrippleCCC))
```
ankddev: Could you place your name on a new line, like other entries?

language-server/src/server.rs
```diff
@@ -41,6 +41,7 @@ pub struct LanguageServer<'a, IO> {
     outside_of_project_feedback: FeedbackBookKeeper,
     router: Router<IO, ConnectionProgressReporter<'a>>,
     changed_projects: HashSet<Utf8PathBuf>,
+    open_files: HashSet<Utf8PathBuf>,
```
ankddev: This should be `opened_files`, shouldn't it?

language-server/src/server.rs
```diff
@@ -131,12 +133,35 @@ where
...
+        _ = self.open_files.remove(path);
+        if let Some(project_path) = self.router.project_path(path) {
+            if self
+                .open_files
+                .iter()
+                .any(|path| path.starts_with(&project_path))
+            {
+                self.router.delete_engine_for_path(&project_path);
+            }
+        }
```
ankddev, line 171: Maybe collapse it? (This is suggested by clippy as well)
```suggestion
        if let Some(project_path) = self.router.project_path(path)
            && self
                .open_files
                .iter()
                .any(|path| path.starts_with(&project_path))
            {
                self.router.delete_engine_for_path(&project_path);
            }
```

ankddev (commented):
Hello! I've left some notes inline.

Also, it should clean all diagnsotic when I close all files in project, am I right? For me it doesn't work, I still see diagnostics (Zed):

Image

TrippleCCC:
> Hello! I've left some notes inline.
> 
> Also, it should clean all diagnsotic when I close all files in project, am I right? For me it doesn't work, I still see diagnostics (Zed):
> 
I initially miss understood the procedure for clearing diagnostics. The latest changes should fix this.

language-server/src/messages.rs
```diff
@@ -90,9 +90,13 @@ impl Request {
 #[derive(Debug)]
 pub enum Notification {
     /// A Gleam file has been modified in memory, and the new text is provided.
-    SourceFileChangedInMemory { path: Utf8PathBuf, text: String },
+    SourceFileChangedInMemory {
+        path: Utf8PathBuf,
+        opened: bool,
```
lpil: What does this represent, and why do we need to track it?

Could you add documentation please 🙏

language-server/src/messages.rs
```diff
@@ -115,6 +120,7 @@ impl Notification {
                 let notification = Notification::SourceFileChangedInMemory {
                     path: super::path(&params.text_document.uri),
                     text: params.content_changes.into_iter().next_back()?.text,
+                    opened: false,
```
lpil: When is it possible to have files edited in memory but not open? Isn't that impossible?

language-server/src/messages.rs
```diff
@@ -90,9 +90,13 @@ impl Request {
...
     /// A Gleam file has been modified in memory, and the new text is provided.
-    SourceFileChangedInMemory { path: Utf8PathBuf, text: String },
+    SourceFileChangedInMemory {
+        path: Utf8PathBuf,
+        opened: bool,
+        text: String,
+    },
     /// A Gleam file has been saved or closed in the editor.
-    SourceFileMatchesDisc { path: Utf8PathBuf },
+    SourceFileMatchesDisc { path: Utf8PathBuf, closed: bool },
```
lpil: Split this into a SourceFileSaved and SourceFileChanged please 🙏

language-server/src/server.rs
```diff
@@ -41,6 +41,7 @@ pub struct LanguageServer<'a, IO> {
     outside_of_project_feedback: FeedbackBookKeeper,
     router: Router<IO, ConnectionProgressReporter<'a>>,
     changed_projects: HashSet<Utf8PathBuf>,
+    opened_files: HashMap<Utf8PathBuf, bool>,
```
lpil, line 59: What does this represent?

language-server/src/server.rs
```diff
@@ -131,12 +133,46 @@ where
...
+                        .map(|is_open| *is_open = false);
+                    self.attempt_engine_cleanup(&path, &mut feedback);
+                }
+                feedback
+            }
+            Notification::SourceFileChangedInMemory { path, opened, text } => {
+                let feedback = self.cache_file_in_memory(&path, text);
+                if opened && feedback.diagnostics.is_empty() {
+                    let _ = self.opened_files.insert(path, true);
+                }
```
lpil: How come we only record the file as opened if there's no diagnostics?

language-server/src/server.rs
```diff
@@ -131,12 +133,46 @@ where
...
-            Notification::SourceFileMatchesDisc { path } => self.discard_in_memory_cache(path),
-            Notification::SourceFileChangedInMemory { path, text } => {
-                self.cache_file_in_memory(path, text)
+            Notification::SourceFileMatchesDisc { path, closed } => {
+                let mut feedback = self.discard_in_memory_cache(&path);
+                if closed {
+                    let _ = self
+                        .opened_files
+                        .get_mut(&path)
+                        .map(|is_open| *is_open = false);
```
lpil: What's the difference between the value being absent and being false?

language-server/src/server.rs
```diff
@@ -131,12 +133,46 @@ where
...
+    }
+
     fn handle_notification(&mut self, notification: Notification) {
         let feedback = match notification {
             Notification::CompilePlease => self.compile_please(),
-            Notification::SourceFileMatchesDisc { path } => self.discard_in_memory_cache(path),
-            Notification::SourceFileChangedInMemory { path, text } => {
-                self.cache_file_in_memory(path, text)
+            Notification::SourceFileMatchesDisc { path, closed } => {
+                let mut feedback = self.discard_in_memory_cache(&path);
```
lpil: Previously this was a router-like function, dispatching to different behaviour functions depending on what the the notification was. Now it mixes up the "what should be done" and the "how it should be done", making it less clear. Could you go back to having that split please 🙏

language-server/src/server.rs
```diff
@@ -131,12 +133,46 @@ where
...
+        {
+            self.router.delete_engine_for_path(&project_path);
+            _ = self.changed_projects.remove(&project_path);
+            self.opened_files
+                .extract_if(|path, _| path.starts_with(&project_path))
+                .map(|(path, _)| path)
+                .into_iter()
+                .for_each(|path| {
+                    feedback.unset_existing_diagnostics(path);
+                });
```
lpil: Prefer for loops for side effects please 🙏

language-server/src/server.rs
```diff
@@ -131,12 +133,46 @@ where
...
     }
 
+    fn attempt_engine_cleanup(&mut self, path: &Utf8PathBuf, feedback: &mut Feedback) {
+        if let Some(project_path) = self.router.project_path(path)
+            && self
+                .opened_files
+                .iter()
+                .filter(|(path, _)| path.starts_with(&project_path))
+                .all(|(_, is_open)| !*is_open)
+        {
```
lpil: Prefer guard clauses and named variables to to `if let` and method chains please 🙏

language-server/src/messages.rs
TrippleCCC: Moved this into separate notification enums. Should be a bit more clear

language-server/src/messages.rs
TrippleCCC: `SourceFileChangedInMemory` was used for both open and change notifications; the flag was used to diff between the two. I went ahead an separated like you suguested.

language-server/src/messages.rs
TrippleCCC: Done

language-server/src/server.rs
TrippleCCC: Added a comment to explain a bit more. TLDR this allows us to keep diagnostics for closed files up when not all project files are open.

language-server/src/server.rs
TrippleCCC: I removed this check. We actually should still track that the file is open even if we don't return diagnostics.

language-server/src/server.rs
TrippleCCC: Added a comment to the `opened_files` field

language-server/src/server.rs
TrippleCCC: Done

language-server/src/server.rs
TrippleCCC: Done

giacomocavalieri:
Hello @TrippleCCC! If you think the PR is ready to be reviewed again remember to un-draft it

language-server/src/server.rs
TrippleCCC: Done

language-server/src/server.rs
lpil: I see now! Thank you.

Seems this exists for feedback book-keeping, but that should be in the feedback book keeper class!

It would also be more efficient and potentially nicer to have each project be responsible for tracking what files they have open. That way we wouldn't need to have this global state that gets traversed over each time to find the files for a given project. Instead when a file is closed the project can indicate that it no longer is in use and can be discarded.

PR #5393, opened 2026-02-24
Add more documentation to CLI

lpil:
Hopefully should make Jak happy!

### DIFF
--- a/compiler-cli/src/lib.rs
+++ b/compiler-cli/src/lib.rs
@@ -311,9 +311,9 @@ enum Command {
     #[command(trailing_var_arg = true)]
     Run {
         #[arg(short, long, ignore_case = true, help = target_doc())]
-/// Which compilation target to use
-#[arg(short, long, ignore_case = true, help = target_doc())]
-target: Option<Target>,
+        /// Which compilation target to use
+        #[arg(short, long, ignore_case = true, help = target_doc())]
+        target: Option<Target>,
 
         #[arg(long, ignore_case = true, help = runtime_doc())]
         runtime: Option<Runtime>,
@@ -335,9 +335,9 @@ target: Option<Target>,
     #[command(trailing_var_arg = true)]
     Test {
         #[arg(short, long, ignore_case = true, help = target_doc())]
-/// Which compilation target to use
-#[arg(short, long, ignore_case = true, help = target_doc())]
-target: Option<Target>,
+        /// Which compilation target to use
+        #[arg(short, long, ignore_case = true, help = target_doc())]
+        target: Option<Target>,
 
         #[arg(long, ignore_case = true, help = runtime_doc())]
         runtime: Option<Runtime>,
@@ -351,9 +351,9 @@ target: Option<Target>,
     #[command(trailing_var_arg = true)]
     Dev {
         #[arg(short, long, ignore_case = true, help = target_doc())]
-/// Which compilation target to use
-#[arg(short, long, ignore_case = true, help = target_doc())]
-target: Option<Target>,
+        /// Which compilation target to use
+        #[arg(short, long, ignore_case = true, help = target_doc())]
+        target: Option<Target>,
 
         #[arg(long, ignore_case = true, help = runtime_doc())]
         runtime: Option<Runtime>,


compiler-cli/src/lib.rs
```diff
@@ -180,12 +187,91 @@ enum Command {
...
     Docs(Docs),
 
     /// Work with dependency packages
-    #[command(subcommand)]
+    ///
+    /// The packages and the acceptable version ranges are specified in
+    /// `gleam.toml`. You can edit this file manually, or use the `gleam add`
+    /// and `gleam remove` commands.
+    ///
+    /// Packge versions follow semantic versioning: MAJOR.MINOR.PATCH.
```
ankddev:
```suggestion
    /// Package versions follow semantic versioning: MAJOR.MINOR.PATCH.
```

compiler-cli/src/lib.rs
```diff
@@ -180,12 +187,91 @@ enum Command {
...
+    /// - Major updates include breaking changes, either type changes or
+    ///   semantic changes.
+    /// - Minor updates include new functionality, but no breaking changes.
+    /// - Patch updates include only bug fixes.
+    ///
+    /// Dependency resolution will be performed automatically by any command
+    /// that builds your project. Once versions have been selected they are
+    /// written to `manifest.toml`, which locks the package to those versions,
+    /// make your build deterministic. You should not edit this file manually.
```
ankddev: I think `making` is better here
```suggestion
    /// making your build deterministic. You should not edit this file manually.
```

compiler-cli/src/lib.rs
```diff
@@ -180,12 +187,91 @@ enum Command {
...
+    ///
+    /// To upgrade the dependencies you can use the `gleam update` command,
+    /// which will select the newest versions compatible with the requirements
+    /// in `gleam.toml`.
+    ///
+    /// The `[dependencies]` section of `gleam.toml` holds the production
+    /// dependencies. These are included in your production application, or
+    /// are used as the dependencies when published as a library. The
+    /// `[dev_dependencies]` section holds dependencies that are only used
+    /// during development. e.g. code used for testing the package.
```
ankddev: Shouldn't it be a comma here?
```suggestion
    /// during development, e.g. code used for testing the package.
```

compiler-cli/src/lib.rs
```diff
@@ -180,12 +187,91 @@ enum Command {
...
+    ///    [dependencies]
+    ///    wibble = ">= 1.2.0 and < 2.0.0 and != 1.4.1"
+    ///
+    /// Permit a range, but deny a specific version within that range. This
+    /// could be useful if there is a version known to have a bug.
+    ///
+    ///    [dependencies]
+    ///    wibble = { git = "https://example.com/wibble.git", ref = "a8b3c5d82" }
+    ///
+    /// A dependency fetched from git instead of from Hex. This is useful
```
ankddev: I think, the sentence case is better here
```suggestion
    /// A dependency fetched from Git instead of from Hex. This is useful
```

compiler-cli/src/lib.rs
```diff
@@ -200,22 +286,23 @@ enum Command {
 
     /// Format source code
     Format {
-        /// Files to format
+        /// The files or directories to format
         #[arg(default_value = ".")]
         files: Vec<String>,
 
-        /// Read source from STDIN
+        /// Read source from standard-input
```
ankddev, line 293: This reads better for me
```suggestion
        /// Read source from the standard input
```

compiler-cli/src/lib.rs
```diff
@@ -268,15 +355,44 @@ enum Command {
...
+    /// Add the package "wibble", requiring version v2.0.0 or higher:
+    ///     gleam add wibble@2
+    ///
+    /// Add the package "wibble", requiring version v2.5.1 or higher:
+    ///     gleam add wibble@2.5.1
+    ///
+    /// Add multiple packages:
+    ///     gleam add wibble@2 warble@1
+    ///
+    /// Add a package as a non-production dependency.
```
ankddev: To be consistent with others
```suggestion
    /// Add a package as a non-production dependency:
```

compiler-cli/src/lib.rs
ankddev, line 340:
```suggestion
/// Which compilation target to use
#[arg(short, long, ignore_case = true, help = target_doc())]
target: Option<Target>,
```
Don't think that this one is needed since it already adds help, but you changed it above, so left it to be consistent

ankddev, line 356:
```suggestion
/// Which compilation target to use
#[arg(short, long, ignore_case = true, help = target_doc())]
target: Option<Target>,
```

ankddev, line 316:
```suggestion
/// Which compilation target to use
#[arg(short, long, ignore_case = true, help = target_doc())]
target: Option<Target>,
```
Don't think that this one is needed since it already adds help, but you changed it above, so left it to be consistent

ankddev (commented):
This looks really nice, I like documented CLIs. Left some notes inline

compiler-cli/src/lib.rs
```diff
@@ -268,15 +355,44 @@ enum Command {
...
+    ///
+    /// The newest compatible version of the package is determined, and then
+    /// `gleam.toml` is updated to require at least that version, with a range
+    /// permitting future patch and minor updates.
+    ///
+    /// Add the package "wibble":
+    ///     gleam add wibble
+    ///
+    /// Add the package "wibble", requiring version v2.0.0 or higher:
+    ///     gleam add wibble@2
```
giacomocavalieri: v2 or higher to me sounds like it could also add a v3. Maybe this could make it clearer it will be lower than 3 as well

compiler-cli/src/lib.rs
lpil: "standard input" is the name of that file descriptor. I think that "the standard input" is a bit vaguer, meaning some input that is standard.

ankddev: Oh, didn't think about that!

lpil:
Thanks folks

PR #5394, opened 2026-02-24
fix: exclude dev dependencies from erlang-shipment export

Harry-kp:
## Summary

- Fixes `gleam export erlang-shipment` including dev dependencies in the output
- Filters manifest packages to only those transitively reachable from production dependencies when building in `Mode::Prod`

**Closes** #5308

## Root cause

`download_dependencies` resolves with `Mode::Dev` (hardcoded), so the manifest always contains all packages including dev-only dependencies. When `erlang_shipment` calls `build::main` with `Mode::Prod`, all manifest packages were still passed to `ProjectCompiler::new` and compiled into `build/prod/erlang/`, then copied into the shipment.

## Fix

Added `prod_only_packages()` in `compiler-cli/src/build.rs` that performs a BFS from the root config's `[dependencies]` (excluding `[dev-dependencies]`) through each package's transitive requirements. In `main_with_warnings`, when mode is `Prod`, the manifest packages are filtered before being passed to the compiler.

## Test plan

- [x] All 6 `test-project-compiler` tests pass (including `with_dev_dep_prod`)
- [x] `cargo check -p gleam` compiles cleanly

### DIFF
--- a/compiler-cli/src/build.rs
+++ b/compiler-cli/src/build.rs
@@ -1,9 +1,10 @@
-use std::{rc::Rc, time::Instant};
+use std::{collections::HashSet, rc::Rc, time::Instant};
 
 use gleam_core::{
     Result,
-    build::{Built, Codegen, NullTelemetry, Options, ProjectCompiler, Telemetry},
-    manifest::Manifest,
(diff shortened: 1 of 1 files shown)


Harry-kp:
Closing — I missed that #5309 already addresses this. Apologies for the duplicate!

PR #5397, opened 2026-02-26
Hide internal types structure

### LINKED ISSUE #5388: Stop suggesting field access completions for internal types
As reported by gungun [here](https://discord.com/channels/768594524158427167/814516237173784607/1475415367911346217) they ended up accidentally using an internal type's field without even knowing it was internal because the LS suggested it in completions. I think not suggesting field access completions for those outside of the module they're defined in would be good. Otherwise it would be easy to inadvertently start depending on internal APIs.

giacomocavalieri:
This PR fixes #5388 

- The "inexhaustive patterns" error messages now show `_` rather than leaking the structure of internal types outside their module
- And the language server no longer show field access completions for internal types outside their module

Should make it much harder to inadvertently rely on an internal type's internal shape by accident now!

---

### DIFF
--- a/compiler-core/src/exhaustiveness/printer.rs
+++ b/compiler-core/src/exhaustiveness/printer.rs
@@ -64,7 +64,7 @@ impl<'a> Printer<'a> {
                 let is_defined_in_current_module = *module == self.current_module;
                 let is_internal = variable
                     .type_
-                    .publicity()
+                    .named_type_publicity()
                     .unwrap_or(Publicity::Public)
                     .is_internal();
 
--- a/compiler-core/src/type_.rs
+++ b/compiler-core/src/type_.rs
@@ -531,7 +531,7 @@ impl Type {
 
     /// If the type is named, return its publicity.
     ///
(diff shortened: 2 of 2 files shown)


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__incomplete_pattern_does_not_show_structure_of_internal_type_outside_of_its_module.snap
```diff
@@ -0,0 +1,29 @@
+---
+source: compiler-core/src/type_/tests/errors.rs
+expression: "import wibble.{type Wibble}\n\npub fn go(wibble: Wibble) {\n  case wibble {}\n}"
+---
+----- SOURCE CODE
+-- wibble.gleam
+@internal
+            pub type Wibble { Wibble Wobble Woo }
```
lpil, line 8: Nice indentation :P

lpil (approved):
Thank you!!!

PR #5399, opened 2026-02-27
feat: delete build caches after removing dependencies

### LINKED ISSUE #5363: `gleam remove` could invalidate the build cache
Hi hello!

When removing a dependency with `gleam remove`, the build cache is not invalidated, so `gleam check` (and `build` and `run`) will continue to report no compile errors until a file in the project has been edited.

Reproduction:

```bash
$ gleam new wibble
$ cd wibble
$ gleam run
# ...
Hello from wibble!
$ gleam remove gleam_stdlib
$ gleam run
Hello from wibble!
```

I thought a dependency in a library was unused and did something similar tot he above to check if the tests still run without the dependency, only to realise that it's actually required after trying to `gleam publish`.

~ 💜

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5363
* Now it deletes the entire build directory after removing dependencies via `gleam remove`
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -397,3 +397,7 @@
 - The formatter now properly formats binary operations in bit array size
   segments.
   ([Andrey Kozhev](https://github.com/ankddev))
+
+- Fixed a bug where after removing dependencies with `gleam remove` if a
+  removed dependency is still used the build would succeed, resulting in
+  runtime crash due to missing files.
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -299,18 +299,23 @@ pub fn cleanup<Telem: Telemetry>(paths: &ProjectPaths, telemetry: Telem) -> Resu
     let changes = PackageChanges::between_manifests(&old_manifest, &manifest);
     telemetry.resolved_package_versions(&changes);
 
-    // Invalidate build caches after removing package
-    let target = config.target;
-    let name = config.name.clone();
+    // Cleanup build cache of the root package if there are some changes.
+    // Without this, if a removed dependency is still used, teh build will
+    // succeed, resulting in runtime crash due to missing files.
+    if changes.any_changes() {
+        tracing::debug!("cleaning_root_package_build_cache");
+        for mode in Mode::iter() {
+            for target in Target::iter() {
+                let lock = BuildLock::new_target(paths, mode, target)?;
+                let _guard = lock.lock(&telemetry)?;
 
-    // We need to clean build directory for the top level package for both dev
-    // and prod modes
-    for mode in [Mode::Dev, Mode::Prod] {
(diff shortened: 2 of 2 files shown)


yoshi-monster:
Isn't running `gleam clean` afterwards way to broad since it also recompiles all dependencies? Wouldn't we just want to invalidate the top-level package?

ankddev:
Yeah, I was thinking about that, drafted PR for now. Initially I wanted to implemebt it like in the LS, where we clean build directory for LSP mode, current target and current package. It's the information that we don't have in cleanup code. I will try to implement it anyway.

ankddev:
@yoshi-monster Now it removes build directory only for the toplevel package, current target, and both dev and prod modes. What do you think?

yoshi-monster:
Hello!!

Sorry, it looks like my github is still broken and won't send me emails.

Yeah, looks good to me! I don't know enough of the compiler if we could delete the cached type information but keep the generated code, but I don't think it matters that much. Thank you! 💜

lpil (commented):
Hello! Shouldn't we be removing the metadata for the removed package, not the root package?

ankddev:
Makes sense to me, will do it tomorrow!

ankddev:
@lpil I tested and found that removing cache for root package is needed here.

For instance, it already removes cache for dependencies here:

https://github.com/ankddev/gleam/blob/b325e95380ffb535e27029d54ce17408bc796d17/compiler-cli/src/dependencies.rs#L526-L563

From my tests issue is resolved only with cleaning cache of root package.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
   `package.hexdocs.pm` rather than `hexdocs.pm/package`.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- `gleam remove` now cleans build caches after removing dependencies, which
+  ensures that dependencies are actually deleted.
+  ([Andrey Kozhev](https://github.com/ankddev))
```
lpil, line 49: This isn't quite right, the dependencies were always deleted. Could you explain the problem in terms of the user facing behaviour please

Also move this to the bug fixes section 🙏

compiler-cli/src/dependencies.rs
```diff
@@ -301,6 +301,18 @@ pub fn cleanup<Telem: Telemetry>(paths: &ProjectPaths, telemetry: Telem) -> Resu
     let changes = PackageChanges::between_manifests(&old_manifest, &manifest);
     telemetry.resolved_package_versions(&changes);
 
+    // We need to clean build directory for the top level package for all modes
```
lpil: Explain why this is needed and only do it if there are any changes please. A logger message would be good too.

lpil (commented):
Lovely! Thank you for the investigation and the explanation. I understand why it wasn't working before now.

I've left a few tiny notes inline

lpil (approved):
Thank you!!

PR #5420, opened 2026-03-03
Warn when Erlang shipment is run on a different OTP version

### LINKED ISSUE #4769: Have Erlang shipments print a warning if being run with a different OTP major version
**Update:** It looks like the error was caused because the newer docker images use Erlang 28 and I was trying to run the exported shipment with Erlang 27. Maybe a check could be added to the entrypoint script to make sure you're running it with the correct Erlang version since the current error message is pretty cryptic.

---

When I run `gleam export erlang-shipment` on a fresh project in the `v1.11.1-erlang` docker image and then try to do `./entrypoint.sh run`, the app crashes with the following error:

```
Error! Failed to eval: docker_gleam_test@@main:run(docker_gleam_test)

Runtime terminating during boot ({undef,[{'docker_gleam_test@@main',run,[docker_gleam_test],[]},{erl_eval,do_apply,7,[{file,"erl_eval.erl"},{line,915}]},{init,start_it,1,[]},{init,start_em,1,[]},{init,do_boot,3,[]}]})

Crash dump is being written to: erl_crash.dump...done
```

The same issue occurs with `v1.11.0-erlang` but not with `v1.10.0-erlang`.

nootr:
When an Erlang shipment is built with one OTP version and run with a different one, the program crashes with a cryptic `undef` error. This commit embeds a version check into the generated entrypoint scripts that prints a clear warning to stderr when a mismatch is detected, without booting the Erlang VM twice.

Closes #4769

## Test plan

- [x] `cargo check` / `cargo clippy` / `cargo fmt` pass
- [x] `cargo test -p gleam-core -p gleam-cli` passes
- [x] Created a test project, exported shipment, verified OTP version is embedded in both scripts
- [x] Test POSIX entrypoint, with matching version and version mismatch
- [x] Test PowerShell entrypoint, with matching version and version mismatch

### DIFF
--- a/compiler-cli/src/beam_compiler.rs
+++ b/compiler-cli/src/beam_compiler.rs
@@ -24,6 +24,7 @@ pub struct BeamCompilerInstance {
     process: Child,
     stdin: Option<ChildStdin>,
     stdout: BufReader<ChildStdout>,
+    pub otp_version: u64,
     // A guard held for cleaning up the temporary file used to start the BEAM instance.
     _source: tempfile::NamedTempFile,
 }
@@ -131,11 +132,27 @@ impl BeamCompilerInstance {
 
         let stdin = process.stdin.take().expect("could not get child stdin");
         let stdout = process.stdout.take().expect("could not get child stdout");
+        let mut stdout = BufReader::new(stdout);
+
+        let mut version_line = String::new();
+        let _ = stdout
+            .read_line(&mut version_line)
+            .map_err(|e| Error::ShellCommand {
+                program: "escript".into(),
+                reason: ShellCommandFailureReason::IoError(e.kind()),
+            })?;
+        let otp_version = version_line
+            .trim()
+            .strip_prefix("gleam-otp-version:")
+            .expect("BEAM compiler did not report OTP version")
+            .parse()
+            .expect("BEAM compiler reported invalid OTP version");
 
         Ok(Self {
             process,
             stdin: Some(stdin),
-            stdout: BufReader::new(stdout),
+            stdout,
+            otp_version,
             _source: escript_file,
         })
     }
--- a/compiler-cli/src/export.rs
+++ b/compiler-cli/src/export.rs
@@ -174,11 +174,17 @@ pub(crate) fn erlang_shipment(paths: &ProjectPaths) -> Result<()> {
         }
     }
 
+    let otp_version = built
+        .otp_version
+        .expect("OTP version not available after Erlang compilation");
+    let otp_version = otp_version.to_string();
+
     // PowerShell entry point script.
     write_entrypoint_script(
         &out.join(ENTRYPOINT_FILENAME_POWERSHELL),
         ENTRYPOINT_TEMPLATE_POWERSHELL,
         &built.root_package.config.name,
+        &otp_version,
     )?;
 
     // POSIX Shell entry point script.
@@ -186,6 +192,7 @@ pub(crate) fn erlang_shipment(paths: &ProjectPaths) -> Result<()> {
         &out.join(ENTRYPOINT_FILENAME_POSIX_SHELL),
         ENTRYPOINT_TEMPLATE_POSIX_SHELL,
         &built.root_package.config.name,
+        &otp_version,
     )?;
 
     crate::cli::print_exported(&built.root_package.config.name);
@@ -206,10 +213,13 @@ one of the following scripts:
 
 fn write_entrypoint_script(
     entrypoint_output_path: &Utf8PathBuf,
-    entrypoint_template_path: &str,
+    entrypoint_template: &str,
     package_name: &str,
+    otp_version: &str,
 ) -> Result<()> {
-    let text = entrypoint_template_path.replace("$PACKAGE_NAME_FROM_GLEAM", package_name);
+    let text = entrypoint_template
+        .replace("$PACKAGE_NAME_FROM_GLEAM", package_name)
+        .replace("$OTP_VERSION_FROM_GLEAM", otp_version);
     fs::write(entrypoint_output_path, &text)?;
     fs::make_executable(entrypoint_output_path)?;
     Ok(())
--- a/compiler-cli/templates/erlang-shipment-entrypoint.ps1
+++ b/compiler-cli/templates/erlang-shipment-entrypoint.ps1
@@ -3,19 +3,35 @@ $ErrorActionPreference = "Stop"
 $PackageName = "$PACKAGE_NAME_FROM_GLEAM"
 $BaseDirectory = $PSScriptRoot
 $ScriptCommand = $args[0]
+$OtpVersion = "$OTP_VERSION_FROM_GLEAM"
+$OtpVersionCheck = "case erlang:system_info(otp_release) of `"$OtpVersion`" -> ok; V -> io:format(standard_error, `"warning: This Erlang shipment was built with Erlang/OTP $OtpVersion but you are running Erlang/OTP ~s.~nThe shipment may fail to run. Please use Erlang/OTP $OtpVersion or rebuild the shipment with your current version.~n`", [V]) end"
 
 $CodePath = Join-Path -Path $BaseDirectory -ChildPath "\*\ebin" -Resolve
+
+function CheckOtpVersion {
+  erl `
+    -noshell `
+    -eval $OtpVersionCheck `
+    -eval 'halt().'
+
+  if ($LASTEXITCODE -ne 0) {
+    exit $LASTEXITCODE
+  }
+}
 
 function Run {
   erl `
     -pa $CodePath `
+    -eval $OtpVersionCheck `
     -eval "$PackageName@@main:run($PackageName)" `
     -noshell `
     -extra $args
 }
 
 function Shell {
-  erl -pa $CodePath
+  CheckOtpVersion
+  erl `
+    -pa $CodePath
 }
 
 switch ($ScriptCommand) {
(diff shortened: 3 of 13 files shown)


compiler-cli/src/export.rs
```diff
@@ -108,15 +114,48 @@ one of the following scripts:
...
     crate::fs::make_executable(entrypoint_output_path)?;
     Ok(())
 }
 
+fn detect_otp_version() -> Option<String> {
+    let output = std::process::Command::new("erl")
+        .args([
+            "-noshell",
+            "-eval",
+            "io:format(erlang:system_info(otp_release)), halt().",
```
lpil: Rather than boot the VM again could the long-running BEAM-compiler print the version when it starts? 🙏  We could use it in other places too then.

compiler-cli/src/export.rs
```diff
@@ -76,18 +76,24 @@ pub(crate) fn erlang_shipment(paths: &ProjectPaths) -> Result<()> {
         }
     }
 
+    let version_check = detect_otp_version()
+        .map(|v| format!("-eval '{}'", otp_version_check(&v)))
+        .unwrap_or_default();
```
lpil: Don't silently discard errors please 🙏 Panicking here would be good

compiler-cli/src/export.rs
```diff
@@ -108,15 +114,48 @@ one of the following scripts:
...
+    if version.is_empty() || !version.chars().all(|c| c.is_ascii_alphanumeric()) {
+        return None;
+    }
+
+    Some(version)
+}
+
+fn otp_version_check(version: &str) -> String {
+    format!(
+        r#"case erlang:system_info(otp_release) of "{version}" -> ok; V -> io:format(standard_error, "warning: This Erlang shipment was built with Erlang/OTP {version} but you are running Erlang/OTP ~s.~nThe shipment may fail to run. Please use Erlang/OTP {version} or rebuild the shipment with your current version.~n", [V]) end"#
```
lpil: Move the code into the entrypoint files please

lpil (commented):
Great, thank you! Left some notes inline.

Could you rebase on main too please? Should fix CI

nootr:
Hi, @lpil! Thank you very much for your code review. I think the latest changes cover your feedback. Please let me know if this is what you meant.

compiler-cli/templates/erlang-shipment-entrypoint.ps1
```diff
@@ -9,13 +9,16 @@ $CodePath = Join-Path -Path $BaseDirectory -ChildPath "\*\ebin" -Resolve
 function Run {
   erl `
     -pa $CodePath `
+    -eval 'case erlang:system_info(otp_release) of "$OTP_VERSION_FROM_GLEAM" -> ok; V -> io:format(standard_error, "warning: This Erlang shipment was built with Erlang/OTP $OTP_VERSION_FROM_GLEAM but you are running Erlang/OTP ~s.~nThe shipment may fail to run. Please use Erlang/OTP $OTP_VERSION_FROM_GLEAM or rebuild the shipment with your current version.~n", [V]) end' `
```
lpil: Assign this to a variable rather than using the same string literal twice please.

language-server/src/tests.rs
```diff
@@ -257,6 +257,10 @@ impl BeamCompiler for LanguageServerTestIO {
     ) -> Result<Vec<String>> {
         panic!("compile_beam({out:?}, {lib:?}, {modules:?}, {stdio:?}) is not implemented")
     }
+
+    fn otp_version(&self) -> Option<String> {
+        None
```
lpil: Call `unimplemented!()` here please

language-server/src/files.rs
```diff
@@ -175,4 +175,8 @@ where
     ) -> Result<Vec<String>, Error> {
         panic!("The language server is not permitted to create subprocesses")
     }
+
+    fn otp_version(&self) -> Option<String> {
+        None
```
lpil: Call `unimplemented!()` here please

compiler-wasm/src/wasm_filesystem.rs
```diff
@@ -35,6 +35,10 @@ impl BeamCompiler for WasmFileSystem {
     ) -> Result<Vec<String>, Error> {
         Ok(Vec::new()) // Always succeed.
     }
+
+    fn otp_version(&self) -> Option<String> {
+        None
```
lpil: Call `unimplemented!()` here please

compiler-core/src/io.rs
```diff
@@ -326,6 +326,8 @@ pub trait BeamCompiler {
         modules: &HashSet<Utf8PathBuf>,
         stdio: Stdio,
     ) -> Result<Vec<String>, Error>;
+
+    fn otp_version(&self) -> Option<String>;
```
lpil: It's invalid for this to ever be `None`, so instead of propagating the null pointer have this always return `&Utf8String` please

compiler-core/src/io/memory.rs
```diff
@@ -446,6 +446,10 @@ impl BeamCompiler for InMemoryFileSystem {
     ) -> Result<Vec<String>, Error> {
         Ok(Vec::new()) // Always succeed.
     }
+
+    fn otp_version(&self) -> Option<String> {
+        None
```
lpil: Call unimplemented!() here please

compiler-cli/templates/erlang-shipment-entrypoint.sh
```diff
@@ -8,13 +8,16 @@ COMMAND="${1-default}"
 run() {
   exec erl \
     -pa "$BASE"/*/ebin \
+    -eval 'case erlang:system_info(otp_release) of "$OTP_VERSION_FROM_GLEAM" -> ok; V -> io:format(standard_error, "warning: This Erlang shipment was built with Erlang/OTP $OTP_VERSION_FROM_GLEAM but you are running Erlang/OTP ~s.~nThe shipment may fail to run. Please use Erlang/OTP $OTP_VERSION_FROM_GLEAM or rebuild the shipment with your current version.~n", [V]) end' \
```
lpil: Assign this to a variable rather than using the same string literal twice please.

compiler-cli/src/beam_compiler.rs
```diff
@@ -130,13 +131,32 @@ impl BeamCompiler {
...
             process,
             stdin,
-            stdout: BufReader::new(stdout),
+            stdout,
+            otp_version,
         })
     }
+
+    pub fn otp_version(&self) -> Option<String> {
+        self.inner.as_ref().map(|inner| inner.otp_version.clone())
```
lpil: Make `otp_version` a property on the beam compiler struct rather than a method please. It must exist, so the spawn function can return the "inner" and the OTP version.

Please also make the version an unsigned int rather than a string, this more precise data-type secures us against possibility of code injection attacks.

compiler-cli/src/beam_compiler.rs
```diff
@@ -21,6 +21,7 @@ struct BeamCompilerInner {
     process: Child,
     stdin: ChildStdin,
     stdout: BufReader<ChildStdout>,
+    otp_version: String,
```
lpil: We use Utf8String rather than String 🙏

lpil (commented):
Thank you!! This is looking really good. I've added some small notes inline.

Sorry this took me so long to review. 😓

nootr:
Hi @lpil, thank you very much for taking the time to review this.

I just wanted to let you know that I haven’t had as much time as usual lately, and I also won’t have much time over the next month, as I’m getting married next week and going on a honeymoon :)

After that, I’ll finish this, but it may take a little while.

nootr:
Hi @lpil,

I've addressed your feedback and also fixed an additional bug where `run` printed the OTP mismatch warning, but `shell` did not. The new OTP version check helper fixes this.

I verified locally that both `run` and `shell` now warn on an OTP mismatch and stay silent when the OTP version matches.

compiler-cli/src/export.rs
```diff
@@ -174,18 +174,25 @@ pub(crate) fn erlang_shipment(paths: &ProjectPaths) -> Result<()> {
         }
     }
 
+    let otp_version = built
+        .otp_version
+        .expect("OTP version not available after Erlang compilation");
```
lpil, line 179: This message seems backwards: It says we expect the OTP version to not be available after Erlang compilation.

compiler-cli/templates/erlang-shipment-entrypoint.sh
```diff
@@ -7,17 +7,29 @@ set -eu
...
+    -eval "$OTP_VERSION_CHECK" \
     -eval "$PACKAGE@@main:run($PACKAGE)" \
     -noshell \
     -extra "$@"
 }
 
 shell() {
-  erl -pa "$BASE"/*/ebin
+  check_otp_version
+  erl \
```
lpil, line 31: This is now booting the BEAM twice. Please only do it once.

compiler-core/src/io/memory.rs
```diff
@@ -462,6 +462,10 @@ impl BeamCompilerIO for InMemoryFileSystem {
     ) -> Result<Vec<String>, Error> {
         Ok(Vec::new()) // Always succeed.
     }
+
+    fn otp_version(&self) -> u64 {
+        unimplemented!()
```
lpil, line 467: I ran the tests and got failures coming from this panicking.

compiler-core/src/build/project_compiler.rs
```diff
@@ -231,10 +232,17 @@ where
             });
         }
 
+        let otp_version = if self.target().is_erlang() && self.options.codegen == Codegen::All {
```
lpil, line 235: Leave a comment explaining what this is doing please.

This seems error prone. The types say it is always present, but it either of those options above are not set like so then it will panic. This implies to me that it is being handled the wrong way. The BEAM compiler could by started by the export function, or the OTP version could be written to a file instead of being threaded through the build tool.

lpil (commented):
Thank you! I've left a handful of comments inline.

Please undraft and tag me when ready for a review

Issue #5424, opened 2026-03-04
Set read-write permissions for the Hex credentials file when writing


lpil:
Currently we use the default umask, which may include it being world-readable. A best practice would be to set permissions to read-write for the user (0600 on Unix-like systems).

bala-bhargav:
@lpil can i try this issue??

GearsDatapacks:
Hello! #5431 is already being worked on to solve this issue, so it would be better if you picked something else.

pushcx:
@bala-bhargav You can take it. None of those tests failed for me locally and the reviewer wanted a test I already said I don't know how to write.

salespaulo:
Hi! I'd like to work on this issue.
My plan is to add a small Unix-only helper for setting `0o600` permissions and use it when writing the stored Hex OAuth credentials file. How are you planning to do it @bala-bhargav ?

bala-bhargav:
Hi @salespaulo! I'm already working on this issue — @pushcx handed it off to me . I've got a working implementation with a 
make_private  helper that sets 0o600 permissions after writing the credentials file, along with a test. I'll be opening a PR shortly!

PR #5431, opened 2026-03-05
set mode 0x600 on hex credentials

pushcx:
Implements #5424. I'm new to Rust but it looked enough like Gleam that I could muddle through based on the compiler errors.

Existing tests pass, but I didn't see test that named `encrypt_and_store_oauth_refresh_token` or `credentials.toml`. Writing one from scratch is beyond me, so I marked this PR to 'allow edits by maintainers' if you'd like to. I won't be bothered if you close this to wait for a PR with tests, this was a fun puzzle.

### DIFF
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -1,5 +1,7 @@
 use crate::{cli, http::HttpClient};
 use ecow::EcoString;
+use std::fs::Permissions;
+use std::os::unix::fs::PermissionsExt;
 use gleam_core::{
     Error, Result, encryption,
     error::{FileIoAction, FileKind},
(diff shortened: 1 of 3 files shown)


compiler-cli/src/hex/auth.rs
```diff
@@ -1,5 +1,7 @@
 use crate::{cli, http::HttpClient};
 use ecow::EcoString;
+use std::fs::Permissions;
+use std::os::unix::fs::PermissionsExt;
```
jtdowney, line 4: This is going to be a problem on non-Unix builds like Windows. You can look at https://github.com/gleam-lang/gleam/blob/e40d30212c9a613160c4144666f589604e7a08e7/compiler-cli/src/fs.rs#L319-L338 for the pattern of how to do this.

compiler-cli/src/hex/auth.rs
```diff
@@ -133,6 +135,8 @@ impl<'runtime> HexAuthentication<'runtime> {
             },
         };
         let toml = toml::to_string(&credentials).expect("OAuth credentials TOML encoding");
+        let perm = Permissions::from_mode(0o600);
```
jtdowney, line 138: I'd recommend extracting this into a function, then adding tests for that function that operate on a file in a temp directory and verify that the permissions are set correctly.

lpil:
Hello @pushcx ! Thanks for looking into this. If you'd like any pointers or to continue in another PR lmk. Cheers

PR #5454, opened 2026-03-13
Set read-write permissions for the Hex credentials file when writing

bala-bhargav:
After writing the Hex OAuth credentials file (`credentials.toml`), set its permissions to `0600` (owner read-write only) on Unix systems, preventing other users from reading it.
#5424

### DIFF
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -133,8 +133,7 @@ impl<'runtime> HexAuthentication<'runtime> {
             },
         };
         let toml = toml::to_string(&credentials).expect("OAuth credentials TOML encoding");
-        crate::fs::write(&path, &toml)?;
-        crate::fs::make_private(&path)?;
+        crate::fs::write_private(&path, &toml)?;
         Ok(())
     }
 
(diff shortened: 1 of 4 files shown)


bala-bhargav:
@lpil
 can you review the PR for the issue #5454

compiler-cli/src/hex/auth.rs
```diff
@@ -134,6 +134,7 @@ impl<'runtime> HexAuthentication<'runtime> {
         };
         let toml = toml::to_string(&credentials).expect("OAuth credentials TOML encoding");
         crate::fs::write(&path, &toml)?;
+        crate::fs::make_private(&path)?;
```
lpil: How about instead of this we have a `write_private` function which does the same as write but with `.mode(0o600)`?

lpil (commented):
Thank you! I've left a note inline.

compiler-cli/src/hex/auth.rs
bala-bhargav: made the corresponding changes @lpil

ankddev:
@bala-bhargav Once you are ready to review, please undraft the PR and ping Louis

lpil:
Hi @bala-bhargav , are you still working on this?

PR #5461, opened 2026-03-15
Fix BitArray$BitArray's TypeScript definition

### LINKED ISSUE #5460: prelude: TypeScript definition for `BitArray$BitArray` marks all fields required
Currently, Gleam's `prelude.d.mts` marks the function signature of `BitArray$BitArray` as:

```ts
export function BitArray$BitArray(
  buffer: Uint8Array,
  bitSize: number,
  bitOffset: number,
): BitArray;
```

Checking the function definition in the corresponding JS:

```js
export const BitArray$BitArray = (buffer, bitSize, bitOffset) =>
  new BitArray(buffer, bitSize, bitOffset);

/// BitArray constructor

/**
 * Constructs a new bit array from a `Uint8Array`, an optional size in
 * bits, and an optional bit offset.  
 *
 * If no bit size is specified it is taken as `buffer.length * 8`, i.e. all
 * bytes in the buffer make up the new bit array's data.
 *
 * If no bit offset is specified it defaults to zero, i.e. there are no unused
 * high bits in the first byte of the buffer.
 *
 * @param {Uint8Array} buffer
 * @param {number} [bitSize]
 * @param {number} [bitOffset]
 */
constructor(buffer, bitSize, bitOffset)
```

The TypeScript declaration marks the `bitSize` and `bitOffset` fields as required, when they are optional according to the JSDoc and the code.

Using Gleam v1.14.0

acandoo:
Fixes #5460

This pull request makes a small change to the `BitArray$BitArray` function signature in `prelude.d.mts`, marking the `bitSize` and `bitOffset` parameters optional instead of required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+- Marked the `bitSize` and `bitOffset` fields of `BitArray$BitArray` TypeScript
+  definition as optional.
+  ([acandoo](https://github.com/acandoo))
+
 ## v1.15.0-rc1 - 2026-03-04
 
 ### Compiler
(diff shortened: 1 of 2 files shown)


acandoo:
Something like this?

PR #5469, opened 2026-03-17
make sure extract variable code action picks right name

giacomocavalieri:
I noticed the extract variable code action would previously produce strange names. Trying not to shadow names of variables in unrelated scopes:
```gleam
case wibble {
  Ok(_) -> {
    let int = 1
    int + 1
  }
  Error(_) -> {
    wobble(11)
//         ^^ Triggering the code action here
  }
}
```

```diff
case wibble {
  Ok(_) -> {
    let int = 1
    int + 1
  }
  Error(_) -> {
+   let int_2 = 11
+   wobble(int_2)
//         ^^^^^ This should have been `int`!
  }
}
```

This PR fixes it!

---

### DIFF
--- a/language-server/src/code_action.rs
+++ b/language-server/src/code_action.rs
@@ -6863,13 +6863,13 @@ fn labels_are_correct<A>(arguments: &[CallArg<A>]) -> bool {
 
 #[derive(Clone)]
 struct NameGenerator {
-    used_names: HashSet<EcoString>,
+    used_names: im::HashSet<EcoString>,
 }
 
 impl NameGenerator {
     pub fn new() -> Self {
         NameGenerator {
-            used_names: HashSet::new(),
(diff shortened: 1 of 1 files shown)


language-server/src/code_action.rs
```diff
@@ -3289,6 +3288,30 @@ impl<'a> ExtractVariable<'a> {
         action
     }
 
+    fn inside_new_scope<F>(&mut self, fun: F)
+    where
+        F: Fn(&mut Self),
+    {
+        let names = self.name_generator.clone();
```
lpil, line 3295: If we're going to be cloning the name generator a lot let's use `im`'s hashset rather than stdlib, to make it cheap.

giacomocavalieri: Ah I didn't think about that, nice catch. I've fixed it!

Issue #5472, opened 2026-03-17
Show correct error message from hexpm 422 unprocessable entity


D-matz:
Currently all hexpm status 422 check for error not replacing then fall through to error late modification. https://github.com/gleam-lang/gleam/blob/f2cfd1ec9142d55cc4b8cbf2c9d9da1aa12509fe/hexpm/src/lib.rs#L549

However 422 can have different errors, resulting in an incorrect “Can only modify a release up to one hour after publication” message when the problem is actually something else.

I tripped on this when the error message said late modification but the real error was a too big package. (Incidentally hexpm may have added new size checks recently, although idk if this is related as I have no clue how hexpm works: https://github.com/hexpm/hex_core/commit/4383e1b08aaf7d6dd325154d02f91e37f55b16b4)

I fixed the error message in my local lib.rs and would make a pull request if it would be good, although idk the preferred way:

-could add a new ApiError TooLarge specifically for too large

-could add a general Unprocessable(String) that includes the body message

I would kind of lean towards the second, idk if all the various reasons for 422 unprocessable entity are documented somewhere, but given that the result will just be an error message, just showing error to user seem simplest/most useful for diagnosing problem publishing package, no matter why hex returned 422. But lmk if I’m missing anything.

lpil:
Oh yes! Let's make it show the specific errors for known bodies and then use a default unexpected response one otherwise.

D-matz:
Sounds good I’ll try tonight, falling through to existing unexpected_response.

D-matz:
https://github.com/gleam-lang/gleam/pull/5476

D-matz:
closed by https://github.com/gleam-lang/gleam/pull/5476 thanks!

PR #5476, opened 2026-03-18
on hexpm error 422 show user error msg text instead of late modification

D-matz:

Currently publish with hex.pm error 422 defaults to showing user "Can only modify a release up to one hour after publication" even if the actual error message is something else, issue: https://github.com/gleam-lang/gleam/issues/5472

This change defaults to showing error message instead. It still checks for LateModification but by a string which might be fragile if hex.pm decides to change the error string. It'd be nice to match more precisely on all possible hex.pm 422 errors, but I'm not sure how or where they're documented, and as a user if my package fails to publish the main thing I care about is seeing the error message why.

Late modification looks the same, other 422 errors show unexpected response with hex status and error:

hexpm2
hexpm1

### DIFF
--- a/hexpm/src/lib.rs
+++ b/hexpm/src/lib.rs
@@ -547,17 +547,17 @@ pub fn api_publish_package_response(response: http::Response<Vec<u8>>) -> Result
         StatusCode::UNAUTHORIZED => Err(unauthorised_response(&parts.headers)),
         StatusCode::FORBIDDEN => Err(ApiError::Forbidden),
         StatusCode::UNPROCESSABLE_ENTITY => {
-            let body = String::from_utf8_lossy(&body).to_string();
+            let body = String::from_utf8_lossy(&body);
             if body.contains("--replace") {
-                return Err(ApiError::NotReplacing);
+                Err(ApiError::NotReplacing)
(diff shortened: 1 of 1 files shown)


hexpm/src/lib.rs
```diff
@@ -547,11 +547,17 @@ pub fn api_publish_package_response(response: http::Response<Vec<u8>>) -> Result
         StatusCode::UNAUTHORIZED => Err(unauthorised_response(&parts.headers)),
         StatusCode::FORBIDDEN => Err(ApiError::Forbidden),
         StatusCode::UNPROCESSABLE_ENTITY => {
-            let body = &String::from_utf8_lossy(&body).to_string();
-            if body.contains("--replace") {
+            let body_str = String::from_utf8_lossy(&body);
```
lpil: No type suffixes or abbreviated words (str) please, revert this renaming and use `&body` when a string is needed. Thank you 🙏

D-matz:
I reverted body_str, tried to only call from_utf8_lossy once and to_string only if needed for msg, but tbh I'm not super confident in rust, lmk if wrong (I mean it compiles/passes test, but idk maybe could be cleaner or more efficient). Also changed early returns to one if statement. Thank you for your patience!

Issue #5389, opened 2026-02-23
Code action to generate new type


giacomocavalieri:
The language server already allows to create new functions when needed. It would be nice if it allowed to do the same thing with types too. For example:
```gleam
type Wibble {
  Wibble(Wobble)
//       ^^^^^^ Doesn't exist, trigger code action
}
```
```diff
type Wibble {
  Wibble(Wobble)
}

+ type Wobble
```

giacomocavalieri:
I've started having a look at this internally and it seems like we don't carry around enough information in the typed ast to tell if a written type (a `TypeAst`) could be resolved to a type or resulted in an error. Do you think we should change it to carry around its type as well like the typed/untyped versions?
```diff
// Something like this, each variant of the TypeAst could 
// hold onto a reference to its type
type UntypedTypeAst = TypeAst<()>
type TypedTypeAst = TypeAst<Arc<Type>>
type TypeAst<Type> {

+  type_: Type
}
```

lpil:
Seems somewhat redundant to me.

Why is it that we would need that information in the type AST? Doesn't the error contain all the required information (that it's an unknown type, what it is called, what module it is referenced in)? I guess it would need the number of parameters too, but that's all I think?

giacomocavalieri:
Oh yeah totally I just didn't think of that! 😂

giacomocavalieri:
I'm working on this!

lupodevelop:
@giacomocavalieri , great to see you’re working on this! 👍🏻 

If it’s useful, I actually have an implementation for this sitting on a private branch. I didn’t open a PR yet because I was finishing a couple of other things first.

If you want to take a look, you can find it here:
[#5389-LS-Code-action-to-generate-new-type](https://github.com/lupodevelop/gleam/tree/%235389-LS-Code-action-to-generate-new-type)

giacomocavalieri:
Please go ahead and open the PR! There's no point in me spending time on something that's been solved already by someone else 😊

PR #5477, opened 2026-03-18
#5389 ls code action to generate new type

lupodevelop:

This pull request introduces a new "Generate type" code action to the language server, allowing users to quickly generate missing type definitions for unknown types directly from the editor. The implementation tracks the arity of unknown types, ensuring generated types have the correct parameterization. 

New code action feature:

* Added `code_action_generate_type` function to generate type definitions for unknown types, including proper handling of type parameters based on arity.
* Integrated the new code action into the language server's code action pipeline, so it is offered alongside other quick fixes. [[1]](diffhunk://#diff-8bfba7ed626aeef6d13e57616558399e9eb1846aeeaf22ca7bf1a719212ea269L50-R51) [[2]](diffhunk://#diff-8bfba7ed626aeef6d13e57616558399e9eb1846aeeaf22ca7bf1a719212ea269R417)

Issue: #5389

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -120,15 +120,15 @@
   is not defined:
 
   ```gleam
-  pub fn wibble(argument: Wibble) { todo }
+  pub fn run(data: Wibble(Int)) { todo }
   ```
 
   The code action will generate:
 
   ```gleam
-  pub type Wibble
+  pub type Wibble(a)
 
-  pub fn wibble(argument: Wibble) { todo }
+  pub fn run(data: Wibble(Int)) { todo }
   ```
 
   ([Daniele Scaratti](https://github.com/lupodevelop))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -115,6 +115,24 @@
 
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The language server now offers a code action to generate a missing type
+  definition when an unknown type is referenced. For example, if `Wibble`
+  is not defined:
+
+  ```gleam
+  pub fn wibble(argument: Wibble) { todo }
+  ```
+
+  The code action will generate:
+
+  ```gleam
+  pub type Wibble
+
+  pub fn wibble(argument: Wibble) { todo }
+  ```
+
+  ([Daniele Scaratti](https://github.com/lupodevelop))
+
 ### Formatter
 
 ### Bug fixes
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -178,6 +178,7 @@ pub enum Error {
         location: SrcSpan,
         name: EcoString,
         hint: UnknownTypeHint,
+        parameter_names: Vec<EcoString>,
     },
 
     /// This happens when someone writes `module_name.` with no type name after
@@ -1574,12 +1575,14 @@ pub fn convert_get_type_constructor_error(
     error: UnknownTypeConstructorError,
     location: &SrcSpan,
     module_location: Option<SrcSpan>,
+    parameter_names: Vec<EcoString>,
 ) -> Error {
     match error {
         UnknownTypeConstructorError::Type { name, hint } => Error::UnknownType {
             location: *location,
             name,
             hint,
+            parameter_names,
         },
 
         UnknownTypeConstructorError::Module { name, suggestions } => Error::UnknownModule {
--- a/compiler-core/src/type_/hydrator.rs
+++ b/compiler-core/src/type_/hydrator.rs
@@ -170,10 +170,12 @@ impl Hydrator {
                 } = environment
                     .get_type_constructor(&module, name)
                     .map_err(|error| {
+                        let parameter_names = generate_parameter_names(arguments);
                         convert_get_type_constructor_error(
                             error,
                             location,
                             module.as_ref().map(|(_, location)| *location),
+                            parameter_names,
                         )
                     })?
                     .clone();
@@ -320,6 +322,7 @@ impl Hydrator {
                             name: name.clone(),
                             location: *location,
                             hint,
+                            parameter_names: vec![],
                         })
                     }
                 }
@@ -373,4 +376,55 @@ impl Hydrator {
 pub struct CreatedTypeVariable {
     pub type_: Arc<Type>,
     pub usage_count: usize,
+}
+
+/// Names for the parameters of an unknown type, used to suggest a definition.
+/// Type variables keep their name; other parameters get a generated one.
+fn generate_parameter_names(arguments: &[TypeAst]) -> Vec<EcoString> {
+    // Reserve the type variable names so generated ones don't collide with them.
+    let mut used: HashSet<EcoString> = arguments
+        .iter()
+        .filter_map(|argument| match argument {
+            TypeAst::Var(TypeAstVar { name, .. }) => Some(name.clone()),
+            TypeAst::Constructor(_) | TypeAst::Fn(_) | TypeAst::Tuple(_) | TypeAst::Hole(_) => None,
+        })
+        .collect();
+
+    let mut next = 0;
+    arguments
+        .iter()
+        .map(|argument| match argument {
+            TypeAst::Var(TypeAstVar { name, .. }) => name.clone(),
+            TypeAst::Constructor(_) | TypeAst::Fn(_) | TypeAst::Tuple(_) | TypeAst::Hole(_) => {
+                loop {
+                    let candidate = type_parameter_name(next);
+                    next += 1;
+                    if used.insert(candidate.clone()) {
+                        break candidate;
+                    }
+                }
+            }
+        })
+        .collect()
+}
+
+/// `0 -> "a"`, `25 -> "z"`, `26 -> "aa"`, ...
+fn type_parameter_name(index: usize) -> EcoString {
+    let alphabet_length = 26;
+    let char_offset = 97;
+    let mut chars = vec![];
+    let mut rest = index;
+
+    loop {
+        let n = rest % alphabet_length;
+        rest /= alphabet_length;
+        chars.push((n as u8 + char_offset) as char);
+
+        if rest == 0 {
+            break;
+        }
+        rest -= 1;
+    }
+
+    chars.into_iter().rev().collect()
 }
--- a/language-server/src/code_action.rs
+++ b/language-server/src/code_action.rs
@@ -1282,6 +1282,102 @@ pub fn code_action_import_module(
     }
 }
 
+pub fn code_action_generate_type(
+    module: &Module,
+    line_numbers: &LineNumbers,
+    params: &CodeActionParams,
+    error: &Option<Error>,
+    actions: &mut Vec<CodeAction>,
+) {
+    let uri = &params.text_document.uri;
+    let Some(errors) = type_errors_for_module(error, module) else {
+        return;
+    };
+
+    for error in errors {
+        let type_::Error::UnknownType {
+            location,
+            name,
+            parameter_names,
+            ..
+        } = error
+        else {
+            continue;
+        };
+
+        let range = src_span_to_lsp_range(*location, line_numbers);
+        if !within(params.range, range) {
+            continue;
+        }
+
+        // Insert the new type stub before the top-level definition that
+        // contains the error, so it appears close to where it is used.
+        let (insert_at, is_public) =
+            definition_start_and_publicity_containing(&module.ast.definitions, location.start)
+                .unwrap_or((module.code.len() as u32, false));
+        let insert_range = src_span_to_lsp_range(
+            SrcSpan {
+                start: insert_at,
+                end: insert_at,
+            },
+            line_numbers,
+        );
+
+        let pub_prefix = if is_public { "pub " } else { "" };
+        let new_text = if parameter_names.is_empty() {
+            format!("{pub_prefix}type {name}\n\n")
+        } else {
+            let parameters = parameter_names.join(", ");
+            format!("{pub_prefix}type {name}({parameters})\n\n")
+        };
+
+        let edit = TextEdit {
+            range: insert_range,
+            new_text,
+        };
+
+        CodeActionBuilder::new("Generate type")
+            .kind(CodeActionKind::QuickFix)
+            .changes(uri.clone(), vec![edit])
+            .preferred(true)
+            .push_to(actions);
+    }
+}
+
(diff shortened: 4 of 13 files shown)


compiler-core/src/type_/hydrator.rs
```diff
@@ -136,11 +136,15 @@ impl Hydrator {
...
+                        let mut err = convert_get_type_constructor_error(
                             e,
                             location,
                             module.as_ref().map(|(_, location)| *location),
-                        )
+                        );
+                        if let Error::UnknownType { ref mut arity, .. } = err {
+                            *arity = arguments.len();
+                        }
+                        err
```
giacomocavalieri: Why is this needed here?

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_type_works_for_unknown_type.snap
```diff
@@ -0,0 +1,21 @@
...
+
+----- AFTER ACTION
+
+pub fn main() {
+  let x: Wobble = todo
+}
+
+type Wobble {
+  Wobble
+}
```
giacomocavalieri: Only generate `type Wobble`, the code action shouldn't come up with constructors as well, there's a separate code action for that

language-server/src/tests/action.rs
```diff
@@ -8145,6 +8146,45 @@ pub fn main() -> Bool {
...
+        GENERATE_TYPE,
+        "
+type Wibble {
+  Wibble(Wobble(Int, String))
+}
+",
+        find_position_of("Wobble").to_selection()
+    );
+}
+
```
giacomocavalieri, line 8802: Can you also add a test for an argument annotation like this one:
```gleam
pub fn wibble(arg: Wibble(some, generics)) { todo }
//                    ^^^^^^ Here
```

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+            name,
+            arity,
+            ..
+        } = error
+        else {
+            continue;
+        };
+
+        let range = src_span_to_lsp_range(*location, line_numbers);
+        if !overlaps(params.range, range) {
```
giacomocavalieri: Use `within`, not overlaps otherwise the code action would pop up even when selecting big unrelated chunks of code

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+            continue;
+        }
+
+        // Avoid offering duplicate actions for the same type name.
+        if !seen_names.insert(name.clone()) {
+            continue;
+        }
```
giacomocavalieri: When using `within`, this would no longer be needed as the code action would be offered for a single type at a time

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+        let insert_at = module.code.len() as u32;
+        let insert_range =
+            src_span_to_lsp_range(SrcSpan { start: insert_at, end: insert_at }, line_numbers);
+
+        // Ensure proper spacing before the new type definition.
+        let prefix = if module.code.ends_with('\n') {
+            "\n"
+        } else {
+            "\n\n"
+        };
```
giacomocavalieri: Rather than adding the type to the end of the module, I'd add it before the definition the error happens in:
```diff
+ type Wibble

pub fn wibble() -> Wibble {
//                  ^^^^ Triggered here
}
```
You can take other code actions like `generate function` as inspiration

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+
+        let new_text = if *arity == 0 {
+            format!("{prefix}type {name} {{\n  {name}\n}}\n")
```
giacomocavalieri: Don't try and generate constructors as well
```suggestion
            format!("{prefix}type {name}")
```

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+        } else {
+            let params: Vec<_> = (0..*arity)
+                .map(|i| {
+                    let letter = (b'a' + (i % 26) as u8) as char;
+                    let suffix = i / 26;
+                    if suffix == 0 {
+                        letter.to_string()
+                    } else {
+                        format!("{letter}{suffix}")
```
giacomocavalieri: I think there's a type printer to come up with generics names, don't reimplement this logic here.

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,83 @@ pub fn code_action_import_module(
...
+                    }
+                })
+                .collect();
+            let params_str = params.join(", ");
+            format!("{prefix}type {name}({params_str}) {{\n  {name}({params_str})\n}}\n")
```
giacomocavalieri: Same as above
```suggestion
            format!("{prefix}type {name}({params_str})")
```

language-server/src/engine.rs
```diff
@@ -391,10 +391,10 @@ where
     ) -> Response<Option<Vec<CodeAction>>> {
         self.respond(|this| {
             let mut actions = vec![];
-            let Some(module) = this.module_for_uri(&params.text_document.uri) else {
+            let module_uri = &params.text_document.uri;
+            let Some(module) = this.module_for_uri(module_uri) else {
```
giacomocavalieri: Why was this changed? It seems like the code coming after it doesn't use this newly declared variable

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 
 ### Language server
 
+- The language server now offers a code action to generate a missing type
+  definition when an unknown type is referenced. The generated stub includes
+  the correct number of type parameters.
+  ([Daniele Scaratti](https://github.com/lupodevelop))
```
giacomocavalieri: No need to explain how the code it generates is correct, that's expected.
I'd add an example though.
```suggestion
  definition when an unknown type is referenced.
  ([Daniele Scaratti](https://github.com/lupodevelop))
```

giacomocavalieri (changes requested):

language-server/src/engine.rs
lupodevelop: You're right, that's a leftover from an earlier draft.

compiler-core/src/type_/hydrator.rs
lupodevelop: get_type_constructor always returns arity: 0 in its error because at that point it doesn't see the call site's arguments. The hydrator is the only place where both the error and arguments.len() are in scope at the same time, so we patch the arity there before propagating it.

lupodevelop:
@giacomocavalieri I think I’ve covered everything you pointed out 😃  let me know if I missed anything.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_type_works_for_unknown_type_with_parameters.snap
```diff
@@ -0,0 +1,18 @@
...
+         ↑                   
+}
+
+
+----- AFTER ACTION
+
+type Wibble {
+  Wibble(Wobble(Int, String))
+}
+type Wobble(a, b)
```
giacomocavalieri: Why is this after while all the other ones put it before? I'd make it consistent with the other ones

CHANGELOG.md
```diff
@@ -8,6 +8,11 @@
 
 ### Language server
 
+- The language server now offers a code action to generate a missing type
+  definition when an unknown type is referenced.
+
```
giacomocavalieri, line 133: I would add an example here

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,77 @@ pub fn code_action_import_module(
...
+            .kind(CodeActionKind::QUICKFIX)
+            .changes(uri.clone(), vec![edit])
+            .preferred(true)
+            .push_to(actions);
+    }
+}
+
+/// Returns the source offset of the top-level definition that contains `pos`,
+/// so the caller can insert code before it.
+fn definition_start_containing(defs: &TypedDefinitions, pos: u32) -> Option<u32> {
```
giacomocavalieri: No shorthand names please:
```suggestion
fn definition_start_containing(definitions: &TypedDefinitions, position: u32) -> Option<u32> {
```

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,77 @@ pub fn code_action_import_module(
...
+    defs.functions
+        .iter()
+        .map(|f| SrcSpan { start: f.location.start, end: f.end_position })
+        .chain(defs.custom_types.iter().map(|ct| ct.location))
+        .chain(defs.type_aliases.iter().map(|ta| ta.location))
+        .chain(defs.constants.iter().map(|c| c.location))
```
giacomocavalieri: Same here
```suggestion
        .map(|function| SrcSpan { start: f.location.start, end: f.end_position })
        .chain(defs.custom_types.iter().map(|custom_type| ct.location))
        .chain(defs.type_aliases.iter().map(|type_alias| ta.location))
        .chain(defs.constants.iter().map(|constant| c.location))
```

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,77 @@ pub fn code_action_import_module(
...
+    }
+}
+
+/// Returns the source offset of the top-level definition that contains `pos`,
+/// so the caller can insert code before it.
+fn definition_start_containing(defs: &TypedDefinitions, pos: u32) -> Option<u32> {
+    defs.functions
+        .iter()
+        .map(|f| SrcSpan { start: f.location.start, end: f.end_position })
+        .chain(defs.custom_types.iter().map(|ct| ct.location))
```
giacomocavalieri: I think you should use `.full_location()` for functions and custom types

giacomocavalieri (approved):

compiler-core/src/type_/error.rs
```diff
@@ -173,6 +173,7 @@ pub enum Error {
         location: SrcSpan,
         name: EcoString,
         hint: UnknownTypeHint,
+        arity: usize,
```
lpil: Let's call this `parameter_count` as types don't have an arity, functions and constructors do.

compiler-core/src/type_/error.rs
```diff
@@ -1511,6 +1512,7 @@ pub enum UnknownTypeConstructorError {
     Type {
         name: EcoString,
         hint: UnknownTypeHint,
+        arity: usize,
```
lpil: Here too

compiler-core/src/type_/hydrator.rs
lpil: The parameter count information should be passed in earlier then, we should not be constructing invalid data.

language-server/src/tests/snapshots/gleam_language_server__tests__action__generate_type_works_for_unknown_type_in_argument_annotation.snap
```diff
@@ -0,0 +1,15 @@
...
+---
+----- BEFORE ACTION
+
+pub fn wibble(arg: Wibble(some, generics)) { todo }
+                   ↑                               
+
+
+----- AFTER ACTION
+
+type Wibble(a, b)
```
lpil: This would need to be public as it is being returned by a public function

language-server/src/code_action.rs
```diff
@@ -1170,6 +1170,97 @@ pub fn code_action_import_module(
...
+                end: insert_at,
+            },
+            line_numbers,
+        );
+
+        let new_text = if *arity == 0 {
+            format!("type {name}\n\n")
+        } else {
+            let mut printer = Printer::new_without_type_variables(&module.ast.names);
+            let params_str = printer.fresh_generic_names(*arity).join(", ");
```
lpil: No need to use the printer here. It's a new definition so there's no possibility of collisions with any existing type parameters. We can use `a, b, c`. OR even better, we could use names that were used in the original code, so `pub fn wibble(arg: Wibble(some, generics)) { todo }` generates `pub type Wibble(some, generics)`.

No `str` abbreviation please 🙏

lpil (commented):
Great, thank you! I've left some notes inline.

language-server/src/code_action.rs
lupodevelop: sorry for str 😅

giacomocavalieri:
Could you rebase so there's no merge commit please?

giacomocavalieri:
Hello! Would you mind rebasing on main to get rid of the conflicts?

lupodevelop:
> Hello! Would you mind rebasing on main to get rid of the conflicts?

Sorry for the delay. Anyway, I did it.

CHANGELOG.md
```diff
@@ -79,6 +79,24 @@
...
+  ```gleam
+  pub fn wibble(arg: Wibble) { todo }
+  ```
+
+  The code action will generate:
+
+  ```gleam
+  type Wibble
+
+  pub fn wibble(arg: Wibble) { todo }
```
lpil: This has produced invalid code! A private type cannot be used in a public signature.

compiler-core/src/type_/hydrator.rs
```diff
@@ -167,10 +167,22 @@ impl Hydrator {
...
                     .map_err(|error| {
+                        let parameter_names = arguments
+                            .iter()
+                            .enumerate()
+                            .map(|(i, arg)| match arg {
+                                TypeAst::Var(TypeAstVar { name, .. }) => name.clone(),
+                                TypeAst::Constructor(_)
+                                | TypeAst::Fn(_)
+                                | TypeAst::Tuple(_)
+                                | TypeAst::Hole(_) => EcoString::from(char::from(b'a' + i as u8)),
```
lpil: This will produce invalid code in these scenarios:

1. When there is more than 26 parameters.
2. When the user has written a parameter with a single letter variable name, such as `Wibble(b, Int)`.

Check out the `next_letter` method in `type_::pretty` for some code that solves problem 1.

lpil (commented):
Looking great! Thank you!

I've spotted a couple bugs, but once they are fixed this is good to go.

nit: `argument` rather than `arg` please 🙏

giacomocavalieri:
Hello! Remember to un-draft the PR if you think it is ready to be reviewed again!

lpil (approved):
Thank you!!

Issue #5456, opened 2026-03-15
JS codegen: subtraction is right-associative in bit array size expressions


Alge:
The JS backend generates right-associative subtractions in bit arrays. This passes with the Erlang backend but not with the JS one.

`a - b - c` in `bytes-size()` generates `a - (b - c)` instead of `(a - b) - c` on the JavaScript target.

## Versions for Gleam, OS, the runtime and any other software used
 - Gleam: 1.14.0
 - OS: Ubuntu 25.10

## Steps and code to reproduce the problem
```
import gleam/io

pub fn main() -> Nil {
  // Bug: subtraction in bit array size is right-associative on JS
  // a - b - c becomes a - (b - c) instead of (a - b) - c
  let a = 10
  let b = 3
  let c = 2
  // (10 - 3) - 2 = 5, but JS generates 10 - (3 - 2) = 9
  let data = <<0:size(5 * 8)>>
  case data {
    <<_:bytes-size(a - b - c)>> -> io.println("PASS")
    _ -> io.println("FAIL")
  }
}

```

```
$ gleam run --target javascript
   Compiled in 0.02s
    Running bitarray_bug.main
FAIL
$ gleam run --target erlang
   Compiled in 0.03s
    Running bitarray_bug.main
PASS
```

## Logs

```
DEBUG download_deps: locking_directory path="/home/alge/repos/gleam_bitarray_bug/build/packages/gleam.lock"
TRACE download_deps: reading_file path="/home/alge/repos/gleam_bitarray_bug/gleam.toml"
DEBUG download_deps: reading_manifest_toml
TRACE download_deps: reading_file path="/home/alge/repos/gleam_bitarray_bug/manifest.toml"
TRACE download_deps: reading_file path="/home/alge/repos/gleam_bitarray_bug/build/packages/packages.toml"
DEBUG download_deps: locking_directory path="/home/alge/repos/gleam_bitarray_bug/build/gleam-dev-erlang.lock"
DEBUG download_deps: locking_directory path="/home/alge/repos/gleam_bitarray_bug/build/gleam-dev-javascript.lock"
DEBUG download_deps: locking_directory path="/home/alge/repos/gleam_bitarray_bug/build/gleam-prod-erlang.lock"
DEBUG download_deps: locking_directory path="/home/alge/repos/gleam_bitarray_bug/build/gleam-prod-javascript.lock"
[...]

PR #5478, opened 2026-03-18
#5456 fix right associative subtraction bit array

lupodevelop:

This pull request fixes a bug in the JavaScript backend where subtraction in bit array size expressions was incorrectly parsed as right-associative instead of left-associative. As a result, expressions like `a - b - c` are now correctly evaluated as `(a - b) - c`. 

Issue: #5456

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -3844,7 +3844,8 @@ where
             let Some((op_s, token, op_e)) = self.tok0.take() else {
                 break;
             };
-            let Some(prec) = token_to_bit_array_size_operator(&token).map(|op| op.precedence()) else {
+            let Some(prec) = token_to_bit_array_size_operator(&token).map(|op| op.precedence())
+            else {
                 self.tok0 = Some((op_s, token, op_e));
                 break;
             };
--- a/test/language/test/language/bit_array_dynamic_size_test.gleam
+++ b/test/language/test/language/bit_array_dynamic_size_test.gleam
@@ -32,7 +32,20 @@ pub fn size_sub_add_chain_is_left_associative_test() {
   let b = 3
   let c = 5
   // (a - b) + c = 12, but a - (b + c) = 2
-  let assert <<value:bytes-size(a - b + c)>> = <<1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12>>
+  let assert <<value:bytes-size(a - b + c)>> = <<
+    1,
+    2,
+    3,
+    4,
+    5,
+    6,
+    7,
+    8,
+    9,
+    10,
+    11,
+    12,
+  >>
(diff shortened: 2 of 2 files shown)


lupodevelop:
Good point on `* `and `/`. I kept this PR intentionally minimal,  just fixing the associativity bug for `+/-` but you're right that the same issue would apply there.

If you're open to it, I could follow up with a separate PR that replaces the ad-hoc sub-parser with the same mechanism the expression parser already uses. It would be a fairly small refactor, a precedence map and a loop reusing `handle_op` but it would handle all operators uniformly and remove `bit_array_size_high_precedence_operator/bit_array_size_additive_operator `entirely.

lupodevelop:
@lpil  Hey! Your comment really made me stop and think 🤯 ! I ended up going pretty deep and refactored the whole thing.
I’ve now unified the operator handling using the same approach `as parse_expression_inner`, so it’s all consistent and a lot cleaner.
Would you prefer I push this into this PR, or should I open a separate one just for the refactor?

lpil:
I don't mind! Whatever is easiest for you :)

compiler-core/src/parse.rs
```diff
@@ -4866,6 +4732,87 @@ fn tok_to_binop(t: &Token) -> Option<BinOp> {
         | Token::Use => None,
     }
 }
+
+fn tok_to_bit_array_size_op(t: &Token) -> Option<IntOperator> {
```
lpil:
```suggestion
fn token_to_bit_array_size_operator(t: &Token) -> Option<IntOperator> {
```

compiler-core/src/parse.rs
```diff
@@ -4888,6 +4835,27 @@ fn do_reduce_clause_guard(op: Spanned, estack: &mut Vec<UntypedClauseGuard>) {
     }
 }
 
+/// Simple-Precedence-Parser, perform reduction for bit array size expressions
+fn do_reduce_bit_array_size((_, token, _): Spanned, estack: &mut Vec<BitArraySize<()>>) {
```
lpil:
```suggestion
fn reduce_bit_array_size((_, token, _): Spanned, estack: &mut Vec<BitArraySize<()>>) {
```

lpil:
Don't worry about the Windows build atm, it's broken on main.

compiler-core/src/parse.rs
lupodevelop: yeah, better 👍🏻

compiler-core/src/parse.rs
lupodevelop: Done. 👍🏻

lupodevelop:
> Don't worry about the Windows build atm, it's broken on main.

I saw that the "ci / validate" workflow also fails. 🥲

PR #5479, opened 2026-03-19
fix bug where add missing type parameter would trigger on invalid types

giacomocavalieri:
I noticed the "add missing type parameters" code action could be triggered on types that do not exists rather than type variables, generating invalid code.
This PR fixes the bug!

wrong_action

---


lpil (approved):
Thank you!

PR #5485, opened 2026-03-19
Warning for ints segments over 52 bits on the js target

### LINKED ISSUE #5458: LSP warning when attempting to destructure bit array to size(>52) on JS target
On the JS target, this code

```gleam
let assert <<first:size(8), rest:size(152)>> = sha
echo "sha: " <> bit_array.base16_encode(sha)
let first_byte = int.to_base16(first)
let rest_bytes = int.to_base16(rest)
echo "first byte: " <> first_byte
echo "rest bytes: " <> rest_bytes
```

outputs, for instance
```
"sha: 67E363C7DC6F87AEB1AEE833FAB587B8AEE2AE78"
"first byte: 67"
"rest bytes: E363C7DC6F87B0000000000000000000000000"
```

`rest_bytes` is silently truncated to 52 bits and zero-padded to the specified length. This is because the max int length on JS is 52 bits.

It might be helpful if the LSP showed a warning when a user attempts to construct an int > 52 bits via bit array destructuring on JS, instead of being confused by the silent truncation. It should probably suggest that the user use `:bytes` or `:bits` instead. (Also, the limit on BEAM as of OTP 26 is [around 4 Mb](https://elixirforum.com/t/is-there-an-integer-size-limit/65647/8) – wonder if we also might want to warn on _that_ 😱 )

[Discord discussion here](https://discord.com/channels/768594524158427167/768594524158427170/1482754317327798302)

giacomocavalieri:
This PR fixes #5458 
I'm not sure about the warning, let me know what you think!

### DIFF
--- a/compiler-core/src/warning.rs
+++ b/compiler-core/src/warning.rs
@@ -256,8 +256,7 @@ impl Warning {
                 text: wrap(
                     "Module names must begin with a lowercase letter and \
 contain only lowercase alphanumeric characters or underscores.",
-                )
-                .into(),
+                ),
                 level: diagnostic::Level::Warning,
                 location: None,
                 hint: Some(wrap_format!(
@@ -1365,7 +1364,7 @@ resulting in the value {truncated_into}."
 
                     Diagnostic {
                         title: "Truncated bit array segment".into(),
-                        text: text,
+                        text,
                         hint: None,
                         level: diagnostic::Level::Warning,
                         location: Some(Location {


compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__prefer_list_is_empty_over_0_eq_list_length.snap
```diff
@@ -23,8 +23,7 @@ warning: Inefficient use of `list.length`
 6 │             let _ = 0 == list.length(a_list)
   │                     ^^^^^^^^^^^^^^^^^^^^^^^^
 
-The `list.length` function has to iterate across the whole
-list to calculate the length, which is wasteful if you only
-need to know if the list is empty or not.
-
+The `list.length` function has to iterate across the whole list to
+calculate the length, which is wasteful if you only need to know if the
+list is empty or not.
```
lpil, line 28: Do you think it's easier to read without the empty line before the hint?

giacomocavalieri: I liked it better with the newline, but I think having it consistent is better overall. We could change the error formatting so a newline is added before the Hints, which one do you like better?

lpil: IIRC originally all the hand written ones had new lines, so if it's a consistency thing I say we return to that.

I think no newline came in with the parse errors maybe? Doesn't matter 😁

giacomocavalieri: Sure! Do you want me to change that here?

lpil: Don't mind :)

PR #5486, opened 2026-03-19
Adding special handling for when hex OAuth session get revoked by hex

### LINKED ISSUE #5423: Add special handling for when a Hex OAuth session has been revoked by Hex
If a session is revoked from Hex the refresh token we store will no longer be usable, resulting in an error.

Improve the UX to explain the situation, and to restart the OAuth flow to get a new token.

404salad:
fixes [#5423](https://github.com/gleam-lang/gleam/issues/5423) 

### DIFF
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -188,10 +188,9 @@ It will be used to locally encrypt your Hex API tokens.
 
         match self.read_and_decrypt_and_refresh_stored_tokens() {
             Ok(Some(tokens)) => return Ok(tokens.as_credentials()),
-            Ok(None) => (),
+            Ok(None) => {}
             Err(Error::HexSessionRevoked) => {
-                println!("\nYour Hex session has been revoked or has expired.");
-                println!("Restarting the authentication flow to get a new token...\n");
+                // we get this error when refresh token can’t be used anymore
             }
             Err(e) => return Err(e),
         }
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -186,8 +186,14 @@ It will be used to locally encrypt your Hex API tokens.
             return Ok(key);
         }
 
-        if let Some(tokens) = self.read_and_decrypt_and_refresh_stored_tokens()? {
-            return Ok(tokens.as_credentials());
+        match self.read_and_decrypt_and_refresh_stored_tokens() {
+            Ok(Some(tokens)) => return Ok(tokens.as_credentials()),
+            Ok(None) => (),
+            Err(Error::HexSessionRevoked) => {
+                println!("\nYour Hex session has been revoked or has expired.");
+                println!("Restarting the authentication flow to get a new token...\n");
+            }
+            Err(e) => return Err(e),
         }
 
         self.create_and_store_new_credentials_via_oauth()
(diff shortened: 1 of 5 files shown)


compiler-cli/src/hex/auth.rs
```diff
@@ -186,8 +186,14 @@ It will be used to locally encrypt your Hex API tokens.
...
         }
 
-        if let Some(tokens) = self.read_and_decrypt_and_refresh_stored_tokens()? {
-            return Ok(tokens.as_credentials());
+        match self.read_and_decrypt_and_refresh_stored_tokens() {
+            Ok(Some(tokens)) => return Ok(tokens.as_credentials()),
+            Ok(None) => (),
+            Err(Error::HexSessionRevoked) => {
+                println!("\nYour Hex session has been revoked or has expired.");
+                println!("Restarting the authentication flow to get a new token...\n");
```
lpil: Do not print inside this function please 🙏 Any printing should only be at the top level.

Thinking about it, do we need to print anything in this case? Logging seems like it would be enough for debugging.

Could you add a comment here explaining that this error happens when the refresh token couldn't be used please

lpil (commented):
Nice! Thank you!! I've left a comment inline

compiler-cli/src/hex/auth.rs
404salad: Hey thanks for the feedback! I don't think we need to print anything I'll fix this and the CI failure

404salad: I updated the pr!

compiler-core/src/error/tests.rs
```diff
@@ -35,3 +35,17 @@ fn test_shell_program_not_found_error() {
         }
     }
 }
+
+#[test]
+fn hex_session_revoked() {
+    let err = Error::HexSessionRevoked.to_diagnostics();
+    assert_snapshot!("hex_session_revoked", err[0].text);
```
lpil: Could you snapshot the whole error instead of just the test field please 🙏

compiler-cli/src/hex/auth.rs
```diff
@@ -186,8 +186,13 @@ It will be used to locally encrypt your Hex API tokens.
             return Ok(key);
         }
 
-        if let Some(tokens) = self.read_and_decrypt_and_refresh_stored_tokens()? {
-            return Ok(tokens.as_credentials());
+        match self.read_and_decrypt_and_refresh_stored_tokens() {
+            Ok(Some(tokens)) => return Ok(tokens.as_credentials()),
+            Ok(None) => {}
+            Err(Error::HexSessionRevoked) => {
+                // we get this error when refresh token can’t be used anymore
```
lpil: Comments start with a capital letter!

Please include in this documentation when this could happen, why we consider it an acceptable error, what will happen as a result of ignoring this error, and that because the refresh token isn't usable we don't need to revoke it. Thank you.

lpil (commented):
Fab! Left a few small things inline. Thank you

lpil:
btw really like your username 😁

compiler-core/src/error/tests.rs
404salad: working on this!

lpil (approved):
Thank you! Very nice, super clean!

404salad:
> Thank you! Very nice, super clean!

Thank you for your guidance Louis, you have created a wonderful language and community :)

Issue #5396, opened 2026-02-26
Add an owner to a package on hex


CrowdHailer:
The gleam tool supports transferring ownership but not adding owners to a package.

It would be helpful to have this otherwise users need mix installed.

lpil:
Good idea, thank you

nkxxll:
I would like to take that PR gonna work on that tomorrow. Sounds like a nice introduction to the codebase. 😄

lpil:
Great, thank you! Do open an early PR if you need any pointers with anything

nkxxll:
> The gleam tool supports transferring ownership but not adding owners to a package.
> 
> It would be helpful to have this otherwise users need mix installed.

@CrowdHailer I think this issue can be safely closed as completed, right?

PR #5487, opened 2026-03-20
feat(cli): added hex owner add

nkxxll:
- [x] Tests have been added for new behaviour (NA: because its a CLI change tested manually)
- [x] clippy checked

related issue #5396

Implementation is heavily oriented on `gleam hex owner transfer`.

Though I did not add tests. I did test the functionality by adding you @lpil to a test package I created on hex.pm. The only thing I did not test is the distinction between `"owner"` and `"maintainer"` (because I actually don't know how I can remove someone from the package :smile:).

Help output:
```bash
$ gleam hex owner add --help
Adds a new owner to the given package on Hex

This command uses this environment variable:

- HEXPM_API_KEY: (optional) A Hex API key to authenticate against the Hex package manager.

Usage: gleam hex owner add [OPTIONS] --user <USERNAME_OR_EMAIL> <PACKAGE>

Arguments:
  <PACKAGE>


Options:
      --user <USERNAME_OR_EMAIL>
          The username or email of the new owner

      --level <LEVEL>
          The ownership level: "full" or "maintainer"

          [default: full]
          [possible values: full, maintainer]

  -h, --help
          Print help (see a summary with '-h')
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,10 @@
 
 ### Build tool
 
+- The `gleam hex owner add` command has been added, which allows adding
+  owners to the package.
+  ([Niklas Kirschall](https://github.com/nkxxll))
+
 - When publishing, the package manager now uses the full term instead of the
   shorthand "MFA" in the prompt and error message.
   ([Luka Ivanović](https://github.com/luka-hash))
--- a/compiler-cli/src/lib.rs
+++ b/compiler-cli/src/lib.rs
@@ -591,6 +591,25 @@ enum Hex {
 
 #[derive(Subcommand, Debug)]
 enum Owner {
+    /// Adds a new owner to the given package on Hex
+    ///
+    /// This command uses this environment variable:
+    ///
+    /// - HEXPM_API_KEY: (optional) A Hex API key to authenticate against the Hex package manager.
+    ///
+    #[command(verbatim_doc_comment)]
+    Add {
+        package: String,
+
+        /// The username or email of the additional owner
+        #[arg(long = "user")]
+        username_or_email: String,
+
+        /// The ownership level
+        #[arg(long, default_value = "maintainer")]
+        level: hexpm::OwnerLevel,
+    },
+
     /// Transfers ownership of the given package to a new Hex user
     ///
     /// This command uses this environment variable:
@@ -816,6 +835,12 @@ fn parse_and_run_command() -> Result<(), Error> {
             let paths = find_project_paths()?;
             hex::revert(&paths, package, version)
         }
+
+        Command::Hex(Hex::Owner(Owner::Add {
+            package,
+            username_or_email,
+            level,
+        })) => owner::add(package, username_or_email, level),
 
         Command::Hex(Hex::Owner(Owner::Transfer {
             package,
--- a/hexpm/Cargo.toml
+++ b/hexpm/Cargo.toml
@@ -28,6 +28,7 @@ http-auth-basic = "0.3"
 prost = "0.13.5"
 
 base16 = { workspace = true, features = ["alloc"] }
+clap = { workspace = true, features = ["derive"] }
 ecow = { workspace = true, features = ["serde"] }
 http.workspace = true
 pubgrub.workspace = true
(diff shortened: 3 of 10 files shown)


nkxxll:
...`From` trait looks nicer.

CHANGELOG.md
```diff
@@ -21,6 +21,10 @@
 
 ### Build tool
 
+- The `gleam hex owner add` command has been added.
+  You can now add owners to your package through the gleam CLI.
+  ([Niklas Kirschall](https://github.com/nkxxll))
+
```
ankddev: I think that wording here is bad, so it could be written like following, IMO it is better (but not ideal :smile:):
```suggestion
- The `gleam hex owner add` command has been added, which allows adding
  owners to the package.
  ([Niklas Kirschall](https://github.com/nkxxll))

```

compiler-cli/src/lib.rs
```diff
@@ -591,6 +591,25 @@ enum Hex {
...
+    ///
+    #[command(verbatim_doc_comment)]
+    Add {
+        package: String,
+
+        /// The username or email of the new owner
+        #[arg(long = "user")]
+        username_or_email: String,
+
+        /// The ownership level: "full" or "maintainer"
```
ankddev: What's the sense to enumerate values here, if clap adds them to `ValueEnum` values automatically?

compiler-cli/src/lib.rs
```diff
@@ -591,6 +591,25 @@ enum Hex {
...
+        #[arg(long, default_value = "full")]
```
giacomocavalieri: I think the default should be "maintainer", best give the least amount of permissions by default

CHANGELOG.md
nkxxll: very fair 🤣

compiler-cli/src/lib.rs
nkxxll: You mean in the comment. Yes this is duplicate. Could have seen this in the help output. Will fix 👍 .

compiler-cli/src/lib.rs
nkxxll: fair

nkxxll:
Thanks for the feedback! I have implemented the necessary fixes. If they work properly now, I'll squash them. 😉

compiler-cli/src/lib.rs
```diff
@@ -607,6 +626,21 @@ enum Owner {
     },
 }
 
+#[derive(ValueEnum, Debug, Clone, Copy)]
+enum OwnerLevel {
```
lpil: Use the hexpm type directly please 🙏 Can add the new trait impl to it there

nkxxll: jep on it. I have to add clap the the hexpm deps then because clap is a "local" dep of the compiler-cli crate, right?

ankddev: @nkxxll I don't think that it is fine to add new dependency to `hexpm`. Firstly, it is published for public use, IIRC. So, clap will be built on every build. Secondly, this just contradicts with the principle of separation of concepts. At least, we can add it behind feature, though I don't like it a lot.

lpil: @nkxxll Yup use it as a workspace dependency.

That's not correct @ankddev, it is not published, and clap is already always built.

hexpm/Cargo.toml
```diff
@@ -16,6 +16,8 @@ categories = ["api-bindings"]
 url = "2.2"
 # Byte collections
 bytes = "1"
+# Command line interface
+clap = { version = "4", features = ["derive"] }
```
lpil: Make it a workspace dependency please, we don't want to have multiple copies of the same crate in the build tree.

compiler-cli/src/lib.rs
ankddev: @lpil Okay, sorry for confusion, and thank you for clarification!

hexpm/Cargo.toml
nkxxll: today I learned something about workspaces in cargo a good day :smile:. Hope it is fine now ;)

PR #5494, opened 2026-03-21
Add unicode warning for Hex publication

### LINKED ISSUE #5490: Emoji descriptions cause Hex publication to fail
`hex.pm` does not like unicode characters in the title or the description of your gleam package. The publish process will fail with `500 Internal Server Error`. But then you have to search what is wrong. Wouldn't it be nice (and very easy to add) if the gleam cli would warn you about the unicode characters instead.

I would love to implement this as well just asking for permission. 😄

qianchongyang:
## Problem

Publishing a Gleam package to Hex fails with a 500 Internal Server Error when the package title or description contains unicode characters (like emoji). Users have no advance warning and must troubleshoot the cause themselves.

## Solution

Added a validation check in the Gleam CLI that warns users about unicode characters in package title and description before attempting to publish. This helps users fix the issue proactively instead of dealing with cryptic server errors.

## Validation

```bash
gleam publish
# Warning: Package title contains unicode characters that may cause Hex publication to fail
# Proceed? [y/N]
```

Fixes #5490

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -29,7 +29,8 @@ pub fn command(paths: &ProjectPaths, replace: bool, i_am_sure: bool) -> Result<(
 
     let should_publish = check_for_gleam_prefix(&config)?
         && check_for_version_zero(&config)?
-        && check_repo_url(&config, i_am_sure)?;
+        && check_repo_url(&config, i_am_sure)?
+        && check_for_non_ascii_characters(&config)?;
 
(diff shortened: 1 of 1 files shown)


lpil:
Hello! Thank you for your contribution, but this is not a feature we want.

Sorry for the confusion, but an issue from someone other than myself does not mean something is to be implemented. You'll need to read the thread to see what I have said needs to be done. I rename issues to make it clearer too.

Issue #4524, opened 2025-04-29
Optimise decision tree for bit arrays


giacomocavalieri:
Right now when generating code for bit arrays we might end up doing something like this:
```js
if (size > n) {
  if (size === n + 10) {
    // ...
  } else {
    // body_1
  }
} else {
  // body_1
}
``` 
If all else branches have the same body it would be really nice to make the tree shorter and skip the first check entirely:
```js
if (size === n + 10) {
  // ...
} else {
  // body_1
}
```

This might be quite challenging and could require reading more on how to reduce the tree size: this could be a very good resource for that https://user.it.uu.se/~kostis/Papers/JFP_06.pdf

lupodevelop:
### This looks like a pretty neat little challenge! 🤯 

I can see two natural ways of "attacking" it:

- “Prune” the tree after it’s generated. A small pass that walks the pattern and merges branches whose bodies are identical, thus eliminating redundant size tests.
- Get your hands dirty and make the pivot/split logic smarter so the redundant test never gets inserted in the first place (the _Gustafsson‑Sagonas paper_).

I haven’t dug into the code yet, but those are the obvious points of entry. I’ll take a look at `exhaustiveness.rs`  soon and see how hard each approach would be and which one makes more sense.

giacomocavalieri:
Sorry what paper are you talking about? Which section in particular? I'd love to read that!

lupodevelop:
> Sorry what paper are you talking about? Which section in particular? I'd love to read that!

I'm referring to exactly this:
> This might be quite challenging and could require reading more on how to reduce the tree size: this could be a very good resource for that https://user.it.uu.se/~kostis/Papers/JFP_06.pdf

I should have more material somewhere, perhaps annotated.
This is the first thing that comes to mind:

http://moscova.inria.fr/~maranget/papers/ml05e-maranget.pdf

😃 Right now I'm pretending to be working profitably on an old issue. Maybe I'll take a look at this when I finish.

giacomocavalieri:
Sorry did you actually take a look at this or not? How are you proposing specific solutions to an issue you haven't had a look at?

lupodevelop:
Yes. I looked at the problem. I have some notes on what I could do. But before saying "I'll work on it," I want to finish the things I have open. Also because otherwise in the meantime I might discourage someone from taking up this issue. Yes, I know I'm a bit cryptic.

[...]

PR #5497, opened 2026-03-22
#4524 optimise decision tree for bit arrays

lupodevelop:

ISSUE: #4524 

This pull request improves the compiler's handling of bit array pattern matching for the JavaScript backend. The main focus is on simplifying the generated decision trees by eliminating redundant size and modulo checks, leading to more efficient and readable compiled code. It also introduces new tests.

**Pattern Matching and Decision Tree Simplification:**

* The compiler now generates simpler decision trees for bit array pattern matching by removing redundant size and modulo checks when multiple patterns overlap. This is achieved by normalizing and simplifying checks like `(bitSize - N) % 8 === 0` to `bitSize % 8 === 0` where possible, and by collapsing equivalent branches in the decision tree. 
* The logic for splitting branches during compilation now prefers to split on `CatchAllIsBytes` checks, which helps to produce simpler and more optimal decision trees for bit array patterns.

**Implementation note:**

The issue mentions an in-generation approach (described in the Gustafsson–Sagonas paper linked in the issue). That was explored first, but the results weren't satisfying. The final implementation instead generates the tree normally and then applies a small post-processing pass (`simplify`) that removes branches made redundant by a sibling check. This keeps the changes localised and avoids touching the core compilation logic, at the cost of a single extra tree walk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,13 @@
   raised in.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler now normalizes remaining-bytes bit-array checks for the JavaScript
+  backend so `(bitSize - c) % 8 === 0` becomes `bitSize % 8 === 0` when the
+  constant offset `c` is congruent modulo 8. This produces more uniform
+  generated code for byte-aligned patterns.
+
+  ([Daniele Scaratti](https://github.com/lupodevelop))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -3501,9 +3501,17 @@ impl CaseToCompile {
             let is_last_segment = i + 1 == segments_count;
             match &segment_size {
                 ReadSize::RemainingBits => (),
-                ReadSize::RemainingBytes => tests.push_back(BitArrayTest::CatchAllIsBytes {
-                    size_so_far: previous_end.clone(),
-                }),
+                ReadSize::RemainingBytes => {
+                    // `(bitSize - c) % 8 === 0` is equivalent to
+                    // `(bitSize - (c % 8)) % 8 === 0` for any constant c,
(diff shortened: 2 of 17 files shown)


lpil (commented):
Hello!

I really like the `%` related optimisation there, but I'm not sure about the other one. In my tests it seems to produce larger code that isn't clearly better performing.

Could we extract the `%` optimisation and merge that, as to not have it be blocked by the other optimisation in this branch.

Did the other one come from a particular paper? What's the theory behind it?

Thank you

lupodevelop:
Hi @lpil, thanks for the review!

You’re right, the two parts of the PR have different tradeoffs.

`%` normalization is safe arithmetic: it replaces `(bitSize - c) % 8 === 0` with `(bitSize - (c % 8)) % 8 === 0` (so offsets that differ by multiples of 8 normalize to the same check). It doesn't come from a paper is purely modular arithmetic and reduces redundant modulo checks across the board.

Regarding the larger code you observed: I was able to reproduce that. It comes from changing the split order to prioritize `CatchAllIsBytes`, which in some cases produces larger trees with simpler checks. Since `simplify() `only collapses branches that are structurally identical, it can’t always recover from that change in shape.

My plan would be to:
	•	**extract only the % normalization into this PR**
	•	drop the split-order change
	•	keep `simplify() `out for now

Then we could evaluate `simplify()` separately (without changing the tree construction order), where its impact should be easier to reason about.

Let me know if that direction works for you.

giacomocavalieri:
Yeah only doing the `%` optimisation in this PR would be great. Go for it!

lpil (commented):
Thank you! Could you update the changelog please

lupodevelop:
done! Sorry

giacomocavalieri (approved):

giacomocavalieri:
Hello! Would you mind rebasing on main to fix the conflicts?

lupodevelop:
> Hello! Would you mind rebasing on main to fix the conflicts?

yup!

giacomocavalieri (approved):

lpil (approved):
Thank you!!

lupodevelop:
> Thank you!!

It's always a pleasure to contribute to this project! 💜

PR #5499, opened 2026-03-23
feat: support .jsx, .tsx, .mts, and .cts as FFI file extensions

### LINKED ISSUE #5368: Adding support for `.mts`, `.cts`, `.tsx` & `.jsx` as FFI files
### Current Issue

Currently, the compiler handle FFI code in JavaScript in a simple way: copy-pasting the external sources [matching a list of extensions](https://github.com/gleam-lang/gleam/blob/4b952446071c2c630f923910d4d1cd79324722f7/compiler-core/src/io.rs#L432). Before diving in the code, I naively thought the compiler was analyzing the imports, and importing everything appearing in an `@external` declaration. This led me to some issues, when doing some interoperability with React and JSX code (and probably with all JSX-friendly frameworks).

In the current JavaScript ecosystem, Vite and modern bundlers distinguish JS and JSX based on the file extension. Because JSX is an extension of JS, acting as a superset, every JSX files should now be written as `[filename].jsx`. Writing JSX code in a `.js` file leads to an error in Vite. This makes sense, as `.jsx` files require a special parser to handle them correctly. However, when writing some React code alongside Gleam code in a frontend codebase, it's not possible to use `@external(javascript, "./[filename].jsx", "function")` in Gleam, as the `.jsx` file will not be exported from the sources. 
The solution to that situation would be to add `.jsx` extension to the list of native file extensions.

### Would it make sense?

Because I know adding anything is a compiler is not an easy task and should be thought out well, let me present some arguments.
I think adding `.jsx` and `.tsx` extension as a supported native file extension would make sense for several reasons.
- The native file detection already includes non-JS files, i.e. `.ts` files. In Node < 24, native support for TS is not there. TS support in the browser does not exist neither. But TS support in the native files make sense to help users write TS code, because while TS files are not natively supported everywhere, they can be supported in some places, like Node >= 24, or even any bundler. Most users in frontend are using a bundler nowadays (whether Vite, ESBuild — inside Lustre Devtools for example — or even Webpack). [While in 2016, only half frontend developers were using bundlers, in 2025, at least 85 % of frontend developers are using a bundler, according to State of JS 2025](https://2025.stateofjs.com/en-US/libraries/build-tools/#build_tools_ratios). All bundlers correctly support TS files. So while TS files are not compatible with all runtimes, the Gleam compiler still allows them.
- [TS compiler handles JSX out-of-the-box](https://www.typescriptlang.org/docs/handbook/jsx.html). If the compiler already allows `.ts` files as native file, it does not make sense to not support `.tsx` files too, files which are already taken into account by the TypeScript compiler.
- [Bun runtime handles JSX out-of-the-box too](https://bun.com/docs/runtime/jsx). Because JSX and TSX are more and more accepted within the JS & TS developers, runtimes are slowly adapting to their usage. Adding them in Gleam would help leveraging Bun and modern runtimes capabilities.
- Accepting them could mean users would be able to publish packages with some `.jsx` on Hex, leading to a somewhat unusable package. While it's true, it's already the case with TypeScript. [TypeScript strongly advises to publish any package containing TS code in its compiled form](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html). So allowing `.ts` files in the Gleam native files could end up with unusable code too. `.jsx` would be identical: if users want to publish them, they should compile the file, and only publish the resulting artifacts, not the source code.
- Accepting `.jsx` and `.tsx` files would also be transparent for most users. It's not a breaking change or anything dangerous. Just like `.cjs` files are allowed: CommonJS is an old format, and is slowly dying. JavaScript newcomers could not even know what CommonJS is, and could use only `import/export` syntax, exactly like Gleam. Though `.cjs` are still supported, because it can be important for interoperability.

### Sumup

To be exhaustive, I think Gleam would also gain to formally accept `.mts` and `.cjs`, which are equivalent for `.mjs` and `.cjs`. If the latter are accepted, I think the former would be more consistency to the native file importer.

I'm clearly advocating to add support for `.jsx` & `.tsx` files. Partly because our frontend at work is built using Redraw — I can't deny that point. (Currently, [Redraw needs an additional loader](https://github.com/ghivert/vite-plugins?tab=readme-ov-file#using-redraw-plugin) in Vite to properly handle JSX interop. It's working, and so it's not bothering for that Gleam does not offer support for `.jsx` and `.tsx`, but it would be a really nice addition.) But also because I find it would make sense to support all flavours of JavaScript out there. I know JSX is not so-much loved in the community, but it's now part of major frameworks (React, [Vue.js](https://vuejs.org/guide/extras/render-function.html#jsx-tsx) or [Solid.js](https://www.solidjs.com/) for example.), and we could improve interoperability by allowing those two formats.

---

I hope that issue is sufficiently detailed and thought out to be helpful for the Gleam design! Tell me if it does not suit the minimum quality requirement, or if I'm completely out of subject!

mvanhorn:
## Summary

Adds `.jsx`, `.tsx`, `.mts`, and `.cts` to the FFI file extension whitelist. These files are now copied alongside existing `.js`, `.mjs`, `.cjs`, and `.ts` files.

## Why this matters

Modern JS/TS tooling (Vite, esbuild, Bun) uses these extensions. JSX files in particular are required by React/Solid/Vue frameworks. The compiler already supports `.ts` - adding `.tsx` and the module-type variants (`.mts`, `.cts`) is consistent with that. See #5368 for the full rationale from @ghivert.

## Changes

- `compiler-core/src/io.rs`: Added `"mts" | "cts" | "jsx" | "tsx"` to the `matches!` macro in `is_native_file_extension()` (line 435).

## Testing

`cargo check -p gleam-core` passes.

Closes #5368

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/compiler-core/src/io.rs
+++ b/compiler-core/src/io.rs
@@ -432,7 +432,7 @@ pub trait TarUnpacker {
 pub fn is_native_file_extension(extension: &str) -> bool {
     matches!(
         extension,
-        "erl" | "hrl" | "ex" | "js" | "mjs" | "cjs" | "ts"
+        "erl" | "hrl" | "ex" | "js" | "mjs" | "cjs" | "mts" | "cts" | "ts" | "jsx" | "tsx"
     )
 }
 


lpil:
Hi @mvanhorn ! Thanks for the PR.

The "good first issue" tickets are intended for humans to learn with, so we can up-skill and enroll more contributors. Having the work performed by AI rather defeats that purpose, and removes opportunity for people to use them.

If you'd like to do this yourself, to learn your way around the compiler, that'd be cool! Otherwise I'm afraid we'd like to leave this for someone else.

mvanhorn:
> Hi @mvanhorn ! Thanks for the PR.
> 
> The "good first issue" tickets are intended for humans to learn with, so we can up-skill and enroll more contributors. Having the work performed by AI rather defeats that purpose, and removes opportunity for people to use them.
> 
> If you'd like to do this yourself, to learn your way around the compiler, that'd be cool! Otherwise I'm afraid we'd like to leave this for someone else.

it's Mr. Gleam himself! Hi! thank for creating Gleam.

I do use ai in my process (clearly labeled co-authored) I'll close this PR then unless you tell me to re-open it.  Do you want contributions on non-good-first-issues projects in the future from me?  

Would love to learn / dog food around the complier myself but I do it with ai helping.

thanks agian for the reply!

lpil:
So long as you're learning and it's not the program committing then that sounds good to me!

PR #5501, opened 2026-03-23
Better hint formatting

giacomocavalieri:
This completes the work we were talking about to report hints as such in the `Diagnostic` data structure and add an empty newline before each. All the changes:
- now there's an empty line before hints in an error/warning message
- all hints start with an uppercase letter (previously, some did not)
- hints are now properly wrapped at the line limit, previously the first line wouldn't take the `"hint: "` label into account, meaning the text could overflow


lpil:
Tests are failing!

giacomocavalieri:
Oops should be good now!

PR #5506, opened 2026-03-24
more consistent cli commands, docs

giacomocavalieri:
This PR closes #5505 and makes the cli a bit more consistent. The commands that require a package/version will all accept it as a required flag.
I've also added docs where it was missing!

### DIFF
--- a/compiler-cli/src/lib.rs
+++ b/compiler-cli/src/lib.rs
@@ -586,8 +586,8 @@ enum Hex {
         #[arg(long)]
         package: Option<String>,
 
-        #[arg(long)]
         /// The version to revert
+        #[arg(long)]
         version: Option<String>,
     },
 


compiler-cli/src/lib.rs
```diff
@@ -181,8 +181,10 @@ enum Command {
     ///
     #[command(verbatim_doc_comment)]
     Publish {
+        /// Replace the latest release with this one.
         #[arg(long)]
         replace: bool,
+        /// Automatically accept all prompts.
```
ankddev: Same as above
```suggestion
        /// Automatically accept all prompts
```

compiler-cli/src/lib.rs
```diff
@@ -181,8 +181,10 @@ enum Command {
     ///
     #[command(verbatim_doc_comment)]
     Publish {
+        /// Replace the latest release with this one.
```
ankddev: Most docs don't end with a period, so for consistency it's better to be removed
```suggestion
        /// Replace the latest release with this one
```

compiler-cli/src/lib.rs
```diff
@@ -574,10 +582,12 @@ enum Hex {
     ///
     #[command(verbatim_doc_comment)]
     Revert {
+        /// The name of the package to revert
         #[arg(long)]
         package: Option<String>,
 
         #[arg(long)]
+        /// The version to revert
```
ankddev:
```suggestion
        /// The version to revert
        #[arg(long)]
```

giacomocavalieri:
Thank you!

compiler-cli/src/lib.rs
```diff
@@ -488,7 +487,7 @@ pub struct NewOptions {
 #[derive(Args, Debug)]
 pub struct CompilePackage {
     /// The compilation target for the generated project
-    #[arg(long, ignore_case = true)]
+    #[arg(long, ignore_case = true, help = target_doc())]
```
ankddev, line 490: Oh, while working on a fix for other issue, I found that this one doesn't have short argument, while other `--target`s have

PR #5508, opened 2026-03-25
Optimise single character matching on JavaScript

### LINKED ISSUE #5400: Optimise JS codegen for single-char string prefix matches
I while ago we discovered that the JavaScript `startsWith` function can be used with an additional index argument, allowing us to potentially optimise string prefix matches:

```js
"hello".startsWith('ll', 2) // --> true
```

this would allow us to organise string prefix matches into a trie, only ever checking for every prefix exactly once:

```gleam
case s {
  "wibble" -> todo
  "wobble" -> todo
}
```

```js
if (s.startsWith('w')) {
  if(s.startsWith('ibble', 1)) {
    // todo
  } else if (s.startsWith('obble', 1)) {
    // todo
  }
}
```

I've been experimenting with that, and it turns out that most performance-critical string pattern matching code never actually really does this -- in practice, the hot parser loops check for single-character or 2-3 character prefixes most often, consuming strings one character at a time.

It turns out, we can get extremely clos (like 98% close!) in real code with a much simpler optimisation:

if a case expression checks for single utf-16 code-unit string prefixes (so anything that fits in a Javascript string with `length === 1`), call `charCodeAt(0)` once, and then check the resulting integer value. If the pattern match is for the empty string, replace it with a check for `s.length === 0`. Most of the speedup is in caching of the variable, circumventing calling the slower `startsWith` machinery repeatedly.

A robot-crafted version of glexer with this optimisation applied results in a 2x performance improvement compared to 1.14:

```
Input               Function                       IPS           Min           P99
before x1           lex                       186.3663        4.7665        8.8307
before x10          lex                        15.5402       59.9722       79.1750
before x100         lex                         1.1485      797.6688      942.1426

Input               Function                       IPS           Min           P99
after x1            lex                       312.9216        2.6653        6.3808
after x10           lex                        23.2026       38.9667       52.1681
after x100          lex                         1.5833      576.2196      715.1019
```

> (Note: the maximum I could achieve with the full "trie"-style optimisation was 318 IPS)

It's relatively easy (and might even result in simpler code!) to extend this optimisation to support all unicode codepoints, by checking if any of the branches contains a codepoint that would require a surrogate pair, and in that case emit a `codePointAt(0)` call instead of a `charCodeAt` call.

I tried hacking that in myself and it was a surprisingly small change (like 30 lines). While I think parsing on the Javascript target is a bit niche a 2x improvement isn't too bad either :)

What do you think? ~ rebecca 💜

GearsDatapacks:
Close #5400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,22 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler now emits more efficient code when matching on single-character
+  string prefixes on the JavaScript target. For example, the `glance` package
+  is now nearly 30% faster on the JavaScript target:
+
+  ```
+  # before:
+  min: 10.8ms, max: 365.82ms, median: 14.74ms, mean: 14.76ms
+  warmup: 100/1.5s, total post-warmup: 1000/14.76s
+
+  # after:
+  min: 8.96ms, max: 143.76ms, median: 10.72ms, mean: 11.06ms
+  warmup: 100/1.24s, total post-warmup: 1000/11.06s
+  ```
+
+  ([Surya Rose](https://github.com/GearsDatapacks))
+
 ### Build tool
 
 - When publishing, the package manager now uses the full term instead of the
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -408,6 +408,20 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
             assignments.push(let_doc(name, value.to_doc()))
         };
 
+        // Variable storing the character code for the first character of a string.
+        // This is only declared if multiple patterns match on just the first
+        // character, as it allows us to avoid calling `.startsWith` multiple times
+        // and just call `.charCodeAt(0)` once, which is much faster.
+        let first_character_variable = if multiple_single_character_prefix_checks(choices) {
+            let name = self.variables.next_local_var(&ASSIGNMENT_VAR.into());
+            let string = self.variables.get_value(var);
+            let first_character = docvec![string, ".charCodeAt(0)"];
+            assignments.push(let_doc(name.clone(), first_character));
+            Some(name)
+        } else {
+            None
+        };
+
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -49,6 +49,10 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler now emits more efficient code when matching on single-character
+  string prefixes on the JavaScript target.
```
giacomocavalieri: It would be cool if the changelog also had real world numbers. Maybe we could test this with glance parsing the files of a gleam project and see how much faster that is

giacomocavalieri:
I'm no JS string expert but it looks good to me! I've left a small nit in the changelog, and it looks like there's some snapshots that haven't been updated!

giacomocavalieri:
Also could you add some tests to the language tests as well? Just to make sure this also words at runtime. I'd like to see a test for both cases of the `.charCodeAt` being inlined, and `.charCodeAt` assigned to a variable

GearsDatapacks:
> it looks like there's some snapshots that haven't been updated!

Nope, there was a bug in the code but when I fixed it I broke some of the already passing tests and didn't check properly before pushing

> Also could you add some tests to the language tests as well?

I've added one for multiple single-char prefix patterns. The existing tests already cover when it is called inline.

yoshi-monster:
Glance parsing itself:

```
# before:
pocket_watch [glance.parse]: min: 10.8ms, max: 365.82ms, median: 14.74ms, mean: 14.76ms
                             warmup: 100/1.5s, total post-warmup: 1000/14.76s

# after:
pocket_watch [glance.parse]: min: 8.96ms, max: 143.76ms, median: 10.72ms, mean: 11.06ms
                             warmup: 100/1.24s, total post-warmup: 1000/11.06s
```

GearsDatapacks:
Ah thank you! I was just going to do that. So that's about a 25-30% speed increase then?

yoshi-monster:
For glance, yeah!

GearsDatapacks:
Okay, I've added that to the changelog. Not 100% sure what the best way to present the information is

Issue #3136, opened 2024-05-14
Incorrect formatting in `case` expression arm


maxdeviant:
While working on https://github.com/massivefermion/birl/pull/24 I noticed some weird formatting applied by `gleam format` in Gleam v1.1.

I pointed it out to @giacomocavalieri and he asked me to open an issue here.

Here is a slightly more minimized reproduction ([repo](https://github.com/maxdeviant/gleam-issue-repros/blob/975cde493b3efe5c53b1e01d9898245395152bd1/formatter_issue_2024_05_13/src/formatter_issue_2024_05_13.gleam)):

```gleam
fn wibble(value: String) {
  case [], [], [] {
    [day_string, time_string], _, _
    | _, [day_string, time_string], _
    | _, _, [day_string, time_string]
    -> Ok(#(day_string, time_string))
    [_], [_], [_] -> Ok(#(value, "00"))
    _, _, _ -> Error(Nil)
  }
}
```

I would expect something more like this:

```gleam
fn wibble(value: String) {
  case [], [], [] {
    [day_string, time_string], _, _
    | _, [day_string, time_string], _
    | _, _, [day_string, time_string] ->
      Ok(#(day_string, time_string))
    [_], [_], [_] -> Ok(#(value, "00"))
    _, _, _ -> Error(Nil)
  }
}
```

### Version Info

```
λ gleam --version
gleam 1.1.0

λ rebar3 --version
rebar 3.23.0 on Erlang/OTP 25 Erts 13.2.2.9

λ erl
Erlang/OTP 26 [erts-14.2.5] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [jit]
```

### Gleam Logs

```
λ GLEAM_LOG=trace gleam format
DEBUG built glob set; 1 literals, 1 basenames, 2 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
TRACE reading_file path="./test/formatter_issue_2024_05_13_test.gleam"
DEBUG ignoring ./.gitignore: Ignore(IgnoreMatch(Hidden))
DEBUG ignoring ./build: Ignore(IgnoreMatch(Gitignore(Glob { from: Some("./.gitignore"), original: "/build", actual: "build", is_whitelist: false, is_only_dir: false })))
TRACE reading_file path="./src/formatter_issue_2024_05_13.gleam"
 INFO Successfully completed
```

lpil:
Thanks

[...]

PR #5512, opened 2026-03-26
add some more formatter tests for case expressions

giacomocavalieri:
I was looking at #3136 and noticed I commented with some tests that should succeed but they had never been included in the codebase to make sure there's no regressions. So this PR adds those back.

Also, I think that issue could be closed, I think the formatting there is reasonable and consistent. Changing it for all branches but the last one would be oddly not consistent I think


lpil (approved):
Thank you!

PR #5513, opened 2026-03-26
feat: make CLI options docs more consistent

### LINKED ISSUE #5492: task: make enumerations of enum's possible values more consistent
### The issue

I find enumeration of possible values in CLI inconsistent. For example, `--template` flag of `gleam new` uses clap's `ValueEnum`, so output looks like this:

```
--template <TEMPLATE>  The template to use [default: erlang] [possible values: erlang, javascript]
```

The ongoing https://github.com/gleam-lang/gleam/pull/5487/ uses same approach. On the other hand, `target` and some others use custom documentation implementation, so output like this:

```
-t, --target <TARGET>     The platform to target (erlang|javascript)
```

I find this very inconsistent.

#### Use custom implementation everywhere

Just use custom implementation everywhere. This one is the simplest option.

#### Use `ValueEnum` everywhere

This one is harder. Since e.g. `target` is defined in `core`, we will need to define newtype with derive of `ValueEnum` or another one enum, which is same, with conversion (or `transmute` 😄) and implementation of `ValueEnum`

***

It's just a small niche, so IDK if it makes sense to implement.

ankddev:
* Closes #5492
* Now all enums in CLI implement clap's `ValueEnum`
* I've simplified some attributes on those enums
* Docs for `--target` and `--runtime` have been improved and includes more details
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,8 +107,8 @@
   possible values.
   ([Andrey Kozhev](https://github.com/ankddev))
 
-- Documentation for `--target` and `--runtime` options has been improved to
-  include more details.
+- Documentation for `--target` option has been improved to include more
+  details.
   ([Andrey Kozhev](https://github.com/ankddev))
 
 ### Language server


compiler-cli/src/lib.rs
```diff
@@ -431,12 +432,14 @@ fn template_doc() -> &'static str {
     "The template to use"
 }
 
-fn target_doc() -> String {
-    format!("The platform to target ({})", Target::VARIANTS.join("|"))
+fn target_doc() -> &'static str {
+    "The platform to target. On the JavaScript target you can also select \
+    the runtime with the `--runtime` option"
```
giacomocavalieri: I don't think this is a nice addition, when showing options one can already see that there's a `runtime` option as well if available.

PR #5517, opened 2026-03-26
Show licence in deps list

### LINKED ISSUE #5515: Show licence in `gleam deps list`
In this issue https://github.com/gleam-lang/gleam/issues/4219 you said it would be nice to show licences and a link to hexdiff when doing `gleam deps list`.
So I opened a separate issue for that.
Do you have any ideas on how these information could be presented?

XiaoPengMei:
Closes #5515

## Summary
- add a `License` column to `gleam deps list`
- read licence information from downloaded dependency `gleam.toml` files when available, leaving the column blank if a package config is not present

## Testing
- `cargo fmt --all --check`
- `cargo test -p gleam-cli list_manifest_format`
- manual QA: ran `cargo run --manifest-path /Users/davidmei/issue-workspaces/gleam-5515/Cargo.toml -p gleam -- deps list` in a temporary project and verified the output included `Package   Version  License` and `demo_dep  1.2.3    MIT`

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -117,18 +117,15 @@ fn list_manifest_packages<W: std::io::Write>(
     paths: &ProjectPaths,
     manifest: Manifest,
 ) -> Result<()> {
-    let packages = manifest
-        .packages
-        .into_iter()
-        .map(|package| {
-            vec![
-                package.name.to_string(),
-                package.version.to_string(),
-                package_licence(paths, &package),
-            ]
-        })
-        .collect_vec();
-    let out = space_table(&["Package", "Version", "License"], packages);
+    let runtime = tokio::runtime::Runtime::new().expect("Unable to start Tokio async runtime");
+    let config = hexpm::Config::new();
+    let http = HttpClient::new();
+    let out = list_manifest_packages_with(paths, manifest, |package_name| {
+        runtime
+            .block_on(hex::get_package(package_name, &config, &http))
+            .map(|package| package.meta.licenses.join(", "))
+            .unwrap_or_default()
+    })?;
 
     write!(buffer, "{out}").map_err(|e| Error::StandardIo {
         action: StandardIoAction::Write,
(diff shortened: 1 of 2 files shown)


compiler-cli/src/dependencies/tests.rs
```diff
@@ -54,15 +79,46 @@ fn list_manifest_format() {
...
--------  -------
+        trim_line_endings(std::str::from_utf8(&buffer).unwrap()),
+        trim_line_endings(
+            "Package  Version  License
+-------  -------  -------
 root     1.0.0
-aaa      0.4.2
-zzz      0.4.0
 "
+        )
```
lpil, line 120: Looks like the licence is missing here!

compiler-cli/src/dependencies.rs
```diff
@@ -112,20 +112,49 @@ fn get_manifest_details(paths: &ProjectPaths) -> Result<(PackageConfig, Manifest
...
 
+fn package_licence(paths: &ProjectPaths, package: &ManifestPackage) -> String {
+    let config_path = match &package.source {
+        ManifestPackageSource::Local { path } => paths.root().join(path).join("gleam.toml"),
+        ManifestPackageSource::Hex { .. } | ManifestPackageSource::Git { .. } => {
+            paths.build_packages_package_config(&package.name)
+        }
+    };
+
+    crate::config::read(config_path)
```
lpil, line 178: We need to report the licences for all dependency packages, not just for Gleam ones. We'll need to work out how to get this information. The linked issue would be a good place to share any ideas

compiler-cli/src/dependencies.rs
```diff
@@ -112,20 +112,49 @@ fn get_manifest_details(paths: &ProjectPaths) -> Result<(PackageConfig, Manifest
...
+        .map(|package| {
+            vec![
+                package.name.to_string(),
+                package.version.to_string(),
+                package_licence(paths, &package),
+            ]
+        })
         .collect_vec();
-    let out = space_table(&["Package", "Version"], packages);
+    let out = space_table(&["Package", "Version", "License"], packages);
```
lpil: typo here!
```suggestion
    let out = space_table(&["Package", "Version", "Licence"], packages);
```

compiler-cli/src/dependencies/tests.rs
warwickmei: Good catch. I added the missing coverage and expanded the test setup so it exercises local, git, and Hex package licence lookups.

compiler-cli/src/dependencies.rs
warwickmei: Updated this to fetch licences for Hex dependencies from the package metadata endpoint rather than relying on downloaded `gleam.toml` files, so non-Gleam packages are covered as well. I also added coverage for the missing-licenses case and kept local/git deps on their package config path.

compiler-cli/src/dependencies.rs
warwickmei: Fixed, thanks.

compiler-cli/src/dependencies.rs
lpil: Fetching this information from Hex is quite a lot of work! We need to figure out what would be the best way to do this to avoid that work.

Best do design work before implementation, to avoid rework.

Issue #2003, opened 2023-02-14
Add integration test for typescript type definitions (d.ts)


JohnBjrk:
Issue #1998 was found when trying to use the type definitions with the TypeScript compiler. Maybe it would be a good idea to add add a small integration test that can be run in CI for testing the type definitions. A minimal JavaScript project that uses some type definitions generated from Gleam code and tries to compile with the TypeScript compiler. Probably this doesn't need to be done on all the cross-compilation targets, but unsure what is the easiest way to integrate this in the current CI pipeline.

jtdowney:
Can this be closed now that #5521 is merged?

lpil:
Oh yes, thank you

PR #5521, opened 2026-03-27
test: add integration test for TypeScript declarations

nkxxll:

Added typescript declaration smoke test for ci.
All credit to @ankddev because he actually made the change the only problem was that npx was not present in `ci.yaml` as far as I can tell. Now the test uses `bunx` because `bun` is installed for the tests and some stricter flags.

I hope this fits the use case better now. If not I am happy to iterate on that I think this smoke test would be nice to have.

#2003


.github/workflows/ci.yaml
```diff
@@ -477,6 +477,11 @@ jobs:
         run: make
         working-directory: ./test/javascript_prelude
 
+      - name: Test generated TypeScript declarations
+        run: make test
+        needs: Install Bun
```
ankddev: According to.the GitHub documentation, `needs` exists only for jobs and not for steps

CHANGELOG.md
```diff
@@ -67,6 +67,9 @@
 
 ### Build tool
 
+- Added TypeScript type declaration smoke test for ci.
```
ankddev: This is internal change, so no changelog entry is needed

test/typescript_declarations/README.md
```diff
@@ -0,0 +1,3 @@
+# typescript_declarations
+
+Check that generated TypeScript declarartions are correct. It uses `tsc` via `npx`.
```
ankddev:
```suggestion
Check that generated TypeScript declarartions are correct. It uses `tsc` via `bunx`.
```

ankddev (commented):
Thank you for picking this up! Looks nice! I've left some notes inline

.github/workflows/ci.yaml
nkxxll: ohh damn, that is right I think steps run in sequence anyways

CHANGELOG.md
nkxxll: ok fine didn't know

nkxxll:
fixed the comments @ankddev ;) thanks for the feedback

compiler-core/src/javascript/decision.rs
```diff
@@ -741,7 +741,7 @@ fn single_character_prefix_code(check: &RuntimeCheck) -> Option<u32> {
     match check {
         // On JavaScript, a single "character" is one that can be represented as
         // a single UTF-16 codepoint.
-        RuntimeCheck::StringPrefix { prefix, rest } if utf16_no_escape_len(prefix) == 1 => {
+        RuntimeCheck::StringPrefix { prefix, rest: _ } if utf16_no_escape_len(prefix) == 1 => {
```
ankddev: Just weent through this PR and found that while this change is unrlated to the PR, it's better to use `..` operator here

```suggestion
        RuntimeCheck::StringPrefix { prefix, .. } if utf16_no_escape_len(prefix) == 1 => {
```

test/typescript_declarations/README.md
```diff
@@ -0,0 +1,3 @@
+# typescript_declarations
+
+Check that generated TypeScript declarartions are correct. It uses `tsc` via `bunx`.
```
ankddev:
```suggestion
Check that generated TypeScript declarations are correct. It uses `tsc` via `bunx`.
```

ankddev (commented):
Randomly went through this PR and found two small things. Also you could rebase your branch, so Windows CI isn't failing

compiler-core/src/javascript/decision.rs
nkxxll: sure this was just a clippy issue that had nothing to do with my change I had to fix it for ci. Didn't pay attention...

nkxxll:
cleaned those up thank you!

lpil (approved):
Thank you!

PR #5522, opened 2026-03-28
fix: correct supplied argument count in use<- arity error

cuiweixie:
## Summary

Fixes the plural branch of the `UseFnIncorrectArity` diagnostic: it used `given` (total arity including the `use` callback) when describing how many arguments were supplied, instead of the count of non-callback arguments (`given - 1`).

## Checklist


lpil:

PR #5523, opened 2026-03-28
fix: check max bit inference size before representability in int_to_bits

cuiweixie:
## Summary

In `int_to_bits`, the `MAX_BITS_INTERFERENCE` check ran after `representable_with_bits`. For oversized read sizes where the value is not representable in that width, the compiler could return `IntToBitsError::Unrepresentable` instead of `ExceedsMaximumSize`. This change checks the maximum inference width first.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Compiler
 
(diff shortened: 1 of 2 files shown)


lpil:

PR #5524, opened 2026-03-28
fix: use digit index for integer thousands separators in formatter

cuiweixie:
## Summary

`underscore_integer_string` walks the reversed string with two counters: iterator index `i` and digit index `j` (skipping `_`). The condition for inserting thousands separators mixed `i` and `j`, so when the input already contained underscores, `i` and `j` could diverge and `_` ended up in the wrong places. The guard now uses `j` consistently, matching digit grouping.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ### Compiler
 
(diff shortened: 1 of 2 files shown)


lpil:

Issue #5418, opened 2026-03-03
Improve experience when publishing packages in a gleam monorepo


daniellionel01:
This is a collection of issues I ran into when I tried to publish my own packages from a gleam monorepo and ideas on how we could improve that.

To be clear, I am talking about a project setup like this:
```sh
.
├── package_a
│   ├── gleam.toml
│   ├── manifest.toml
│   ├── README.md
│   └── src
│       └── app.gleam
├── package_b
│   ├── gleam.toml
│   ├── manifest.toml
│   ├── README.md
│   └── src
│       └── app.gleam
├── README.md
└── .git
```

## Warning when there is no tag prefix or path in `repository` in `gleam.toml`

To make the hexdocs code references work, you need to make sure you link the correct path of your package inside of the git repository:
```toml
# gleam.toml
repository = { type = "github", user = "...", repo = "...", path = "package_a", tag-prefix = "pck-a-" }
```

If there is a mismatch between location of the `.git` directory (see point below) and the location of the gleam package, we could issue a warning that code references in hexdocs might not work if they do not match.

## Make .git directory discovery more robust

When I published my package I did not get the `git tag` message after running `gleam publish`.

Having looked at the current source code, I think we can make the logic for detecting the `.git` directory more robust:
<https://github.com/gleam-lang/gleam/blob/4bcd14f500a4b7259eacec51fb7c93c23b588ec4/compiler-cli/src/publish.rs#L116-L132>

Instead, we could run `git rev-parse --show-toplevel` to know if we are in a .git directory.

As it states on <https://git-scm.com/docs/git-rev-parse>:
> Unless otherwise specified, most of the options and operation modes require you to run this command inside a git repository or a working tree that is under the control of a git repository, and will give you a fatal error otherwise.

## Command for checking hexdocs references

This might be kind of out there, but judging from the reaction in Discord, invalid hexdocs references are fairly common, especially among new users. So it would be cool if there was a command that could check if any of the hexdocs links run into 404 or github (or other providers) pages that do not exist.

Definitely something someone (me) could provide as a separate package, but wanted to bring up the idea anyways.

---

Let me know what y'all think! Obviously there is a lot to consider with different providers and lots of edge cases that aren't on my map.

lpil:
Sounds like a good idea!!

Do we even path the path at that point? We could infer it entirely.

ankddev:
I'm working on better `.git` discovery!

ankddev:
[...]

PR #5525, opened 2026-03-28
Better Git repo discovery in `gleam publish`

ankddev:
* Related to #5418 (that issue contains some other items, so this PR won't close it)
* We now use `git rev-parse` with flags to check presence of repo and tag
* Now we don't use paths here entirely, though other items in #5418 will probably require to get path of the `.git` directory.
***
- [ ] Tests have been added for new behaviour (N/A)

### DIFF
--- a/compiler-cli/src/publish.rs
+++ b/compiler-cli/src/publish.rs
@@ -18,7 +18,11 @@ use gleam_core::{
 use hexpm::version::{Range, Version};
 use itertools::Itertools;
 use sha2::Digest;
-use std::{collections::HashMap, io::Write, path::PathBuf};
+use std::{
+    collections::HashMap,
+    io::Write,
+    process::{Command, Stdio},
+};
 
 use crate::{build, cli, docs, fs, http::HttpClient, new::default_readme};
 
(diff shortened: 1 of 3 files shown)


compiler-cli/src/publish.rs
```diff
@@ -116,19 +120,40 @@ pub fn command(paths: &ProjectPaths, replace: bool, i_am_sure: bool) -> Result<(
...
-        println!(
-            "
-Please push a git tag for this release so source code links in the
-HTML documentation will work:
-
-    git tag {tag_name}
-    git push origin {tag_name}
-"
-        )
+    let is_repository = Command::new("git")
```
lpil: Rather than shelling out multiple times could we keep the logic as before but walk up the file tree ourselves as we do in `compiler-cli/src/fs.rs`'s `get_project_root` please 🙏 

A `get_git_repository_root` or such. This could be tested also!

lpil (commented):
Thank you! I've left a note inline

compiler-cli/src/publish.rs
ankddev: OK, will do it tomorrow!

ankddev:
@lpil I've implemnted back this fucntionality using paths and not shell commands, also added few tests for discovering of repo root.

compiler-cli/src/publish.rs
```diff
@@ -116,23 +116,43 @@ pub fn command(paths: &ProjectPaths, replace: bool, i_am_sure: bool) -> Result<(
...
+            return Some(path);
+        }
+
+        path = match path.parent() {
+            Some(path) => path.into(),
+            None => return None,
+        }
+    }
+}
+
```
giacomocavalieri: I thought this was what the `Paths` data structure was for, how come we're not using that?

lpil: What's the `Paths` data structure?

lpil: Ah, ProjectPaths? That one is pure, it doesn't do file system IO.

lpil (approved):
Thank you!!!

PR #5531, opened 2026-03-29
Fault tolerant compilation

### LINKED ISSUE #5371: Gleam LSP don't show my new module in LSP actions
Okay I really like Gleam as a language but sometimes I just want to hate it's LSP when it come to do refactoring 😅
Often I have a type with some functions I need to make them common for multiple module so what I want to do is move this code to a new file than in the original file replace the function and types with imports.

My issue is while doing this my LSP is kind of dead and refuse to auto import my new module if I don't have an import.
I made three videos:
- I extract my `Context` to a new file and I first try to import it with LSP than I manually import and fix the errors https://asciinema.org/a/TZ2WExH4mX9Uw7Ww 
- I extract my `Context` to a new file but I make sure to not have any errors in this module and after I try in my mai to use LSP to import https://asciinema.org/a/LmBbsmk01yHQ3vvK
- I already have a working project and I remove the import, LSP work, close my editor and reopen and try to import it don't work anymore. https://asciinema.org/a/jqQvP2SjrqpyE8sw

```
gleam --version
gleam 1.14.0
```

I don't have logs for the Gleam LSP but I think it's easily reproducible with my videos linked above.

giacomocavalieri:
This PR fixes #5371 

This PR makes the compiler not stop at the first module with an error it finds.
Now the compiler reports errors from all unrelated modules, rather than stopping at the first one it finds.
The compiler will not analyse a module if a module it depends on has errors, so this should not lead to false positives.

I couldn't find an issue specifically talking about this, let me know if I missed it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -179,6 +179,10 @@
 
 ### Language server
 
+- The language server will now show a diagnostic if you have a file open that
+  could not be analysed due to its dependencies failing to compile.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 - The language server now offers code actions to wrap a function reference in an
   anonymous function, or to remove a trivial anonymous function, leaving its
   contents. For example:


language-server/src/code_action.rs
```diff
@@ -581,6 +582,22 @@ pub fn code_action_inexhaustive_let_to_case(
...
 }
 
+pub(crate) fn type_errors_for_module<'a>(
+    error: &'a Option<Error>,
+    module: &Module,
+) -> Option<&'a Vec1<type_::Error>> {
+    let Some(Error::Type { failed_modules }) = error else {
+        return None;
+    };
+    failed_modules.iter().find_map(|failed_module| {
```
giacomocavalieri: Since we now have to go over failed modules in the language server, do you think it would make more sense to have them stored as a map `path -> failed_module` for faster lookup?

compiler-core/src/error.rs
```diff
@@ -1693,3106 +1699,3077 @@ The error from the encryption library was:
...
-                        importing_module,
-                        imported_module,
-                        kind: InvalidImportKind::SrcImportingTest,
-                    } => {
-                        let text = wrap_format!(
-                            "The application module `{importing_module}` \
-is importing the test module `{imported_module}`.
+                .sorted_by_key(|failed_module| &failed_module.path)
+                .flat_map(failed_module_diagnostics)
+                .collect_vec(),
```
giacomocavalieri, line 1807: This diff is super messy! All I did in this module was:
- move the code turning `TypeError`s into diagnostics into its own function `failed_module_diagnostics`, this is what really messed the diff
- add a `wrap` to a long error message we forgot about

compiler-core/src/build/package_compiler.rs
```diff
@@ -596,47 +610,65 @@ fn analyse(
...
+                    },
+                );
             }
         };
     }
 
-    Outcome::Ok(modules)
+    // Now we need to check if any module has failed and return the appropriate
+    // outcome.
+    let errors = Vec1::try_from_vec(failed_modules.into_values().collect_vec());
```
lpil: Can we collect directly to the vec1 instead of allocating a vec and then a vec1?

compiler-core/src/error.rs
```diff
@@ -69,6 +69,18 @@ pub struct DefinedModuleOrigin {
...
+#[error("type checking failed")]
+pub struct FailedModule {
+    pub path: Utf8PathBuf,
+    pub src: EcoString,
+    pub errors: Vec1<crate::type_::Error>,
+    pub names: Box<Names>,
+}
+
+/// TODO)) Maybe it's best to wrap the parse, type, unknown import errors into a
+/// variant since it's only those the ones that can happen multiple times now
```
lpil: TODO here! I'm not sure what it means though 😅

language-server/src/code_action.rs
lpil: Sounds reasonable to me!

compiler-core/src/build/package_compiler.rs
```diff
@@ -540,6 +543,16 @@ fn analyse(
...
         tracing::debug!(module = ?name, "Type checking");
 
+        // We first need to check if the module can actually be compiled.
+        // If we weren't able to compile one of the modules it depends on, then
+        // we have to skip this one to avoid reporting false errors.
+        let depends_on_failed_module = dependencies
+            .iter()
+            .any(|(dependency, _)| failed_modules.contains_key(dependency));
+        if depends_on_failed_module {
+            continue;
```
lpil, line 587: Don't we need to insert this module into the failed modules in this path?

e.g:

A imports B imports C

A fails

B skipped

C will attempt compilation due to B not being marked as failed.

compiler-core/src/error.rs
giacomocavalieri: Ooh was from a previous iteration on this that did it a different way. It's no longer relevant for the code you're reading 😄

compiler-core/src/build/package_compiler.rs
giacomocavalieri: Yeah, switching to a HashMap this will no be needed anyway 👍

giacomocavalieri:
I should have addressed all the comments and made it so the LS can show the error message on open files that were skipped. Let me know what you think of the implementation!

D-matz:
uYxzEnv

PR #5533, opened 2026-03-29
Implement go-to-definition and renaming for record fields

### LINKED ISSUE #4358: LS: Rename record label
So Gleam programmers can change their record labels quickly and easily.

### LINKED ISSUE #5534: LSP: Support go-to-definition on record fields
I'd like to be able to use go-to-definition with record fields

```rs
pub type Thing {
  Thing(a: String)
}

pub fn work() {
  let x = get_thing()
  x.a // go-to-def here!
}
```

I started working on this in #5533, with some Discussion in the Gleam Discord [here](https://discord.com/channels/768594524158427167/768594524158427170/1487827603812388924)

alii:
Fixes #5534
Fixes #4358


compiler-core/src/ast/typed.rs
```diff
@@ -1217,6 +1217,40 @@ impl TypedExpr {
         }
     }
 
+    /// If this expression refers to a record constructor, returns the name of
+    /// that constructor. This is the unqualified name for `Var` and the label
+    /// for a qualified `ModuleSelect`.
+    ///
+    pub fn record_constructor_name(&self) -> Option<EcoString> {
```
GearsDatapacks: I'm not sure `record_constructor_name` is the best description of this function, as it doesn't only apply to record constructors. Perhaps just `name`?

compiler-core/src/reference.rs
```diff
@@ -25,6 +25,25 @@ pub struct Reference {
...
+    /// constructor is unknown (e.g. `record.field`).
+    pub constructor: Option<EcoString>,
+    /// Whether the label was written using the shorthand syntax (`label:`).
+    /// Renaming the field then has to expand it so the value keeps its name:
+    /// `wibble:` becomes `wobble: wibble`.
+    pub shorthand: bool,
+}
+
+/// Key: (type_module, type_name, label_name)
+pub type LabelReferenceMap = HashMap<(EcoString, EcoString, EcoString), Vec<LabelReference>>;
```
GearsDatapacks: Instead of having a tuple with three elements of the same type as a key, I think it would be clearer if you defined a custom struct with labels and used that as a key instead. That way you could also eliminate this type alias.

compiler-core/src/reference.rs
```diff
@@ -392,6 +414,43 @@ impl ReferenceTracker {
...
+        constructor: Option<EcoString>,
+        shorthand: bool,
+    ) {
+        let location = if shorthand {
+            argument_location
+        } else {
+            SrcSpan {
+                start: argument_location.start,
+                end: argument_location.start + label.len() as u32,
+            }
```
GearsDatapacks: Instead of doing this check here, could we just pass in `location` when we call the function in the first place? That would probably simplify this a bit and let the call-site decide on which location is relevant.

GearsDatapacks (commented):
This is great, thank you! I've left a couple of minor style comments inline.

I also have a question (open to anyone who has an opinion) on how renaming should work for fields that are shared between variants. For example:
```gleam
pub type Shape {
  Circle(colour: String, radius: Float)
  Rectangle(colour: String, width: Float, height: Float)
  //                ^ Rename here
}

pub fn colour(shape: Shape) -> String {
  shape.colour
}
```

If we rename one of the fields, what should happen? If we just rename the `Rectangle` one then the `shape.colour` expression will fail to compile because it's now named differently to the `Circle` one. We could rename both, but that could be annoying if that's not what was intended.

alii:
Thank you for the review :pray:
Will get to this ASAP.

alii:
Resolved review comments in 5659ea29b692e9bab59630a0e03d1cab43bc9e62 :pray:

--

Re/ renaming fields shared between variants, my preference (and what's implemented thus far) is to rename across all variants with the same name. To me, this feels the most sensible. For example: all shapes have a colour, and if I rename `colour` to `shade` on one shape, I'd expect all shapes to now have a shade instead. Furthermore renaming across all variants breaks less code: renaming just one would break any `shape.colour` accessor, since those only exist while every variant shares the field.

GearsDatapacks:
Sounds good to me. Could you add a test for that case then? Unless there's already one and I missed it.

alii:
Thanks! I did add a test in 5659ea29b692e9bab59630a0e03d1cab43bc9e62, [please see here](https://github.com/gleam-lang/gleam/blob/5659ea29b692e9bab59630a0e03d1cab43bc9e62/language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_record_field_shared_between_variants.snap).

GearsDatapacks (approved):
Great! Looks good to me now. Just need Louis to review it before merging; probably will have to wait until after 1.17 is released.

alii:
Exciting! Thank you. Are we able to unblock CI?

GearsDatapacks:
Need @lpil to do it, but yes

compiler-core/src/ast/typed.rs
```diff
@@ -1217,6 +1217,40 @@ impl TypedExpr {
         }
     }
 
+    /// The name of the value this expression refers to, if any. This is the
+    /// unqualified name for `Var` and the label for a qualified
+    /// `ModuleSelect`.
+    ///
```
lpil, line 1250: This is introducing a new meaning for the word "name" which I don't quite understand. Could you expand on this concept please? It is confusing to me that expressions now have a name.

Where it is used it is called a `constructor_name`, but that seems to be a different thing.

compiler-core/src/type_/expression.rs
```diff
@@ -1225,6 +1225,25 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
 
         self.purity = self.purity.merge(fun.called_function_purity());
 
+        // Register field label references for record constructor calls.
+        if fun.is_record_constructor_function()
+            && let Some(type_name) = type_.named_type_name()
+        {
+            let constructor_name = fun.name();
```
lpil: This is repeating the logic of the `is_record_constructor_function` function I believe. Instead of doing the same thing twice, return the required data from the first call.

compiler-core/src/type_/expression.rs
```diff
@@ -1225,6 +1225,25 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            for arg in &arguments {
```
lpil:
```suggestion
            for argument in &arguments {
```

compiler-core/src/reference.rs
```diff
@@ -392,6 +422,35 @@ impl ReferenceTracker {
...
+    ///
+    /// `type_` is the `(module, name)` of the type the label belongs to, as
+    /// returned by [`Type::named_type_name`].
+    pub fn register_label_reference(
+        &mut self,
+        type_: (EcoString, EcoString),
+        label: EcoString,
+        location: SrcSpan,
+        kind: ReferenceKind,
+        constructor: Option<EcoString>,
```
lpil: What is this argument? Labels don't have constructors, so I'm not sure.

compiler-core/src/type_/expression.rs
```diff
@@ -2984,12 +3003,22 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             type_,
             documentation,
         } = self.infer_known_record_access(
-            record_type,
+            record_type.clone(),
```
lpil: Sorry, maybe I'm being slow. How come this is cloned now?

compiler-core/src/type_/expression.rs
```diff
@@ -2984,12 +3003,22 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
             label,
         )?;
+        if let Some(type_) = record_type.named_type_name() {
+            self.environment.references.register_label_reference(
+                type_,
+                label.clone(),
+                label_location,
+                ReferenceKind::Unqualified,
+                None,
+                false,
```
lpil: Not very clear what this `false` is. Use an descriptive enum instead please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -4077,7 +4119,11 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
 
                 let mut final_arguments = base_arguments;
                 for argument in arguments {
-                    if argument.uses_label_shorthand() {
+                    // Captured before `argument.value` is moved below, so they
+                    // are still available when we register the label reference.
```
lpil: No need to explain Rust memory management. Remove this comment please 🙏

compiler-core/src/build.rs
```diff
@@ -475,7 +476,15 @@ pub enum Located<'a> {
         location: SrcSpan,
     },
     UnqualifiedImport(UnqualifiedImport<'a>),
-    Label(SrcSpan, std::sync::Arc<Type>),
+    Label {
```
lpil, line 483: Is this any label or just record labels? If just records, please call this `RecordLabel`.

Rather than introduce the concept of an "owner" (which I would want to rename to make it clearer), how about we split this into two variants? `RecordLabelDefinition` and `RecordLabelUsage`?

compiler-core/src/build.rs
```diff
@@ -475,7 +476,15 @@ pub enum Located<'a> {
...
+        location: SrcSpan,
+        /// The type of the labelled argument value (used for hover).
+        type_: std::sync::Arc<Type>,
```
lpil:
```suggestion
        field_type: std::sync::Arc<Type>,
```

compiler-core/src/type_/expression.rs
```diff
@@ -4449,6 +4506,27 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
             .collect_vec();
 
+        // Register a reference to each labelled field so the language server can
+        // offer go-to-definition, find-references and rename on record fields. We
+        // do this before adding back the ignored arguments below, as those are
+        // synthetic placeholders without a real value.
+        if let Some(type_) = expected_return.named_type_name() {
+            for argument in &typed_arguments {
+                if let (Some(label), Some(label_location)) =
+                    (&argument.label, argument.label_location())
```
lpil: Use let chains rather than tuple wrapping please 🙏

compiler-core/src/type_/pattern.rs
```diff
@@ -1229,6 +1229,24 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                             constructor_field_map,
                         );
 
+                        if let Some(type_) = return_.named_type_name() {
+                            for arg in &pattern_arguments {
+                                if let (Some(label), Some(label_location)) =
```
lpil: Use let chains rather than tuple wrapping please 🙏

compiler-core/src/type_/pattern.rs
```diff
@@ -1229,6 +1229,24 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                             constructor_field_map,
                         );
 
+                        if let Some(type_) = return_.named_type_name() {
+                            for arg in &pattern_arguments {
```
lpil:
```suggestion
                            for argument in &pattern_arguments {
```

compiler-core/src/ast.rs
```diff
@@ -1712,8 +1730,20 @@ impl CallArg<TypedExpr> {
                 }
                 Some(located) => Some(located),
                 None => {
-                    if self.location.contains(byte_index) && self.label.is_some() {
-                        Some(Located::Label(self.location, self.value.type_()))
+                    if self.location.contains(byte_index) && let Some(label) = &self.label {
+                        let fn_type = called_function.type_();
+                        let record_type = fn_type.return_type().unwrap_or(fn_type);
```
lpil: `return_type` should never fail fail to return ok in this context, right? Let's use `expect` to enforce that property rather than silently failing.

compiler-core/src/ast.rs
```diff
@@ -1712,8 +1730,20 @@ impl CallArg<TypedExpr> {
...
+                        Some(Located::Label {
+                            location: self.location,
+                            type_: self.value.type_(),
+                            label: label.clone(),
+                            owner: called_function.name().map(|c| {
```
lpil: I'm confused by this code. It says it gets the name of the function, but I don't think this function does that, and the value is then being called "constructor", but constructors and function names are different things.

compiler-core/src/ast.rs
```diff
@@ -1764,12 +1794,27 @@ impl CallArg<TypedExpr> {
...
-                if self.location.contains(byte_index) && self.label.is_some() {
-                    Some(Located::Label(self.location, self.value.type_()))
+                if self.location.contains(byte_index)
+                    && let Some(label) = &self.label
+                {
+                    Some(Located::Label {
+                        location: self.location,
+                        type_: self.value.type_(),
+                        label: label.clone(),
+                        owner: constructor.map(|c| LabelOwner::Usage {
```
lpil: Same here as above

compiler-core/src/ast.rs
```diff
@@ -1779,12 +1824,27 @@ impl CallArg<TypedPattern> {
...
```
lpil: Same here as above.

compiler-core/src/ast.rs
```diff
@@ -1779,12 +1824,27 @@ impl CallArg<TypedPattern> {
 }
 
 impl CallArg<TypedConstant> {
-    pub fn find_node(&self, byte_index: u32) -> Option<Located<'_>> {
+    pub fn find_node(
+        &self,
+        byte_index: u32,
+        record_type: &Arc<Type>,
+        constructor: Option<&EcoString>,
```
lpil: What is this argument? It is called constructor, but it is a string. It's not clear to me how a constructor can be represented by a string, so either it is not correct, or it needs documentation.

compiler-core/src/build.rs
```diff
@@ -506,6 +534,36 @@ impl<'a> Located<'a> {
...
+            label: label.clone(),
+        };
+        let references = module.references.label_references.get(&key)?;
+        let definition = references.iter().find(|r| {
+            r.kind == reference::ReferenceKind::Definition
+                && match (&r.constructor, constructor) {
+                    (Some(a), Some(b)) => a == b,
+                    _ => true,
+                }
+        })?;
```
lpil: Could you add comments explaining the logic of this function please 🙏

How come the "constructor"s have to match if they are both present, but not if one is absent?

compiler-core/src/build.rs
```diff
@@ -519,6 +577,10 @@ impl<'a> Located<'a> {
             }),
             Self::Statement(statement) => statement.definition_location(),
             Self::FunctionBody(statement) => None,
+            Self::Expression {
+                expression: TypedExpr::RecordAccess { record, label, .. },
+                ..
+            } => self.label_definition_location(importable_modules, &record.type_(), label, None),
```
lpil: Rather than adding logic to this dispatching function to further specialise the located data, create a new variant in the `Located` type for `RecordAccessLabel`

compiler-core/src/build.rs
```diff
@@ -566,7 +628,30 @@ impl<'a> Located<'a> {
...
+            ),
+            // Already at the declaration; go-to-definition jumps to itself.
+            Self::Label {
+                owner: Some(LabelOwner::Definition { .. }),
+                location,
+                ..
+            } => Some(DefinitionLocation {
+                module: None,
+                span: *location,
+            }),
```
lpil, line 665: Same as above, it would be good to avoid having extra specialisation logic in this function. My suggestion about splitting this one variant into two would help here.

compiler-core/src/reference.rs
```diff
@@ -25,6 +25,33 @@ pub struct Reference {
...
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct LabelReference {
+    /// The location of the label. For a shorthand (`label:`) this spans the
+    /// whole `label:`, mirroring how variable references record label
+    /// shorthands. For everything else it is just the label itself.
+    pub location: SrcSpan,
+    pub kind: ReferenceKind,
+    /// The constructor this label belongs to. `None` for usages where the
+    /// constructor is unknown (e.g. `record.field`).
+    pub constructor: Option<EcoString>,
```
lpil: Ah! I understand what a constructor is now. The word you are looking for is "variant"

compiler-core/src/reference.rs
```diff
@@ -25,6 +25,33 @@ pub struct Reference {
 
 pub type ReferenceMap = HashMap<(EcoString, EcoString), Vec<Reference>>;
 
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct LabelReference {
+    /// The location of the label. For a shorthand (`label:`) this spans the
+    /// whole `label:`, mirroring how variable references record label
+    /// shorthands. For everything else it is just the label itself.
+    pub location: SrcSpan,
+    pub kind: ReferenceKind,
```
lpil: I believe this is stored so that the `label_definition_location` function can identify the one that is the definition by iterating across all of the references, and all but `Unqualified` and `Definition` are unused.

It's incorrect to say that a label is unqualified (as there is no qualification of labels), and storing this extra data + doing the traverse is wasteful, so how about instead storing a map of label definitions alongside the label references? It would simplify the tracking of references a little too

language-server/src/engine.rs
```diff
@@ -839,6 +842,13 @@ where
 
                 Some(Referenced::TypeVariable { location, name: _ }) => success_response(location),
 
+                Some(Referenced::Label {
+                    location, label, ..
+                }) if location.contains(byte_index) => success_response(SrcSpan {
```
lpil: How come this new clause needs to check the location contains the byte index while none of the other ones do?

language-server/src/engine.rs
```diff
@@ -1048,6 +1073,31 @@ where
                     ))
                 }
             },
+            Some(Referenced::Label {
+                type_module,
+                type_name,
+                label,
+                location,
+            }) if location.contains(byte_index) => match search_scope {
```
lpil: Same here. If a label has been located then we don't need to check the location a second time, surely?

language-server/src/reference.rs
```diff
@@ -370,14 +377,64 @@ pub fn reference_for_ast_node(
...
+                },
+            ..
+        } => record
+            .type_()
+            .named_type_name()
+            .map(|(type_module, type_name)| Referenced::Label {
+                type_module,
+                type_name,
+                label: label.clone(),
+                // `field_start` is the start of the whole `record.field`
```
lpil: What's `field_start`? I can't see it here?

lpil (commented):
This is absolutely fantastic work!!! Thank you so much! Really excited to merge this one in.

I've left a smattering of notes and question inline. Thanks again

alii:
Thank you both so much for taking the time to review. Will get to these ASAP 🙏🏻

language-server/src/reference.rs
alii: Ah left over from earlier code! - the clause is gone now

compiler-core/src/ast/typed.rs
alii: This was originally something that was used in a few places where name was somewhat accurate, but it's now `record_constructor_variant_name`

compiler-core/src/type_/expression.rs
alii: infer_known_record_access previously took ownership but we can remove it

compiler-core/src/type_/expression.rs
alii: Agreed! Added a LabelSyntax which is either Longhand or Shorthand

compiler-core/src/ast.rs
alii: renamed to variant

compiler-core/src/reference.rs
alii: aAgree! done

language-server/src/engine.rs
alii: When the span covered more than just the label, although this was kind of a hack and now Located carries the correct span anyway so this is outdated now

alii:
Thank you for the in-depth review. Resolved your comments.

lpil:
Wow, that was fast! I'll review again shortly.

alii:
Is the dependency check CI failure something I should be fixing?

ankddev:
@alii You could rebase on main to get rid of this failure!

alii:
Ah thanks!

lpil (commented):
This looks great!! In my testing I found a bug relating to record updates:

```gleam
pub type Example {
  Example(a: Int, b: Int)
}

pub fn example(value: Example) -> Example {
  Example(..value, b: 100)
}
```

If the field `b` is renamed then it incorrectly changes `..value` to the new name. `..value` also shows as a reference when using show-references.

alii:
Great catch thanks. Ill fix that shortly and look for potentially similar problems

alii:
Just rebased since the module renaming PR landed, which caused some conflicts here.

lpil (approved):
Brilliant work! Really high quality!! Thank you.

This is going to be such a huge improvement to the Gleam developer experience 💜

alii:
Thank you Louis!!!!!!! So exciting!

PR #5545, opened 2026-03-31
Support `textDocument/reference` for imports

### LINKED ISSUE #5544: LSP: Support triggering find references from imports
In the following example, triggering find references on any of the underlined locations should show all of the uses of the type `Foo`. It currently works everywhere but in the `import` line.

```gleam
// -- foo.gleam
pub type Foo
//       ^^^

// -- app.gleam
import foo.{type Foo}
//               ^^^ ← this line shows up in all results, but triggering here fails
fn main() {
  let _: Foo = todo
  //     ^^^
}
```

### LINKED ISSUE #5864: LS: allow to trigger rename on import alias
For example,

```gleam
import wibble.{Wibble as Wobble}
//                    ^^^^^^^^^ Allow to trigger here
```

gavinmorrow:
Closes #5544.
Closes #5864.


language-server/src/reference.rs
```diff
@@ -361,13 +376,48 @@ pub fn reference_for_ast_node(
...
+            } else {
+                Some(Referenced::ModuleValue {
+                    module: module.clone(),
+                    name: name.clone(),
+                    location: *imported_name_location,
+                    name_kind: if is_upname {
+                        Named::CustomTypeVariant
+                    } else {
+                        // QUESTION: is this okay? it seems to be done above (eg lines 172 and 188).
+                        Named::Function
```
gavinmorrow: is this okay? (just setting `name_kind` to `Named::Function` b/c that info isn't easily available.) it seems to be done above (eg lines 172 and 188).

language-server/src/reference.rs
```diff
@@ -264,18 +272,25 @@ pub fn reference_for_ast_node(
...
                     }
                     ast::TypeAst::Fn(_)
                     | ast::TypeAst::Var(_)
                     | ast::TypeAst::Tuple(_)
-                    | ast::TypeAst::Hole(_) => (RenameTarget::Unqualified, ast.location()),
+                    | ast::TypeAst::Hole(_) => (
+                        // QUESTION: when would this ever be run
+                        RenameTarget::Unqualified { as_name: todo!() },
+                        ast.location(),
+                    ),
```
gavinmorrow: When would this ever be run?

lpil (commented):
Hello! Are you waiting for a review? Mark as ready for review and tag me if so 🙏

gavinmorrow:
Not on this particular one yet, there are still a few things I'm working out.

I do have two other pull requests that are ready for review though (and are marked as such in GitHub); should I tag you on them?

gavinmorrow:
@lpil It's ready for review!

gavinmorrow:
I rebased to resolve the conflicts.

CHANGELOG.md
```diff
@@ -143,6 +143,20 @@
...
+- The language server can now find references for and rename items when
+  triggered from an import statement:
+
+  ```gleam
+  import foo.{type Foo}
+  //               ^^^ Trigger find references or rename here
+
+  pub fn main() {
+    let _ = Foo
+  }
```
giacomocavalieri: Could you use `wibble` instead of `foo` for code examples?

compiler-core/src/ast.rs
```diff
@@ -1277,8 +1295,10 @@ pub struct UnqualifiedImport {
     pub location: SrcSpan,
     /// The location excluding the potential `as ...` clause, or the `type` keyword
     pub imported_name_location: SrcSpan,
+    pub as_name_location: Option<SrcSpan>,
```
giacomocavalieri: Can you add a comment showing in a code example what this span covers?

compiler-core/src/build.rs
```diff
@@ -420,7 +420,10 @@ pub struct UnqualifiedImport<'a> {
     pub name: &'a EcoString,
     pub module: &'a EcoString,
     pub is_type: bool,
+    pub is_upname: bool,
     pub location: &'a SrcSpan,
+    /// The location excluding the potential `as ...` clause, or the `type` keyword
+    pub imported_name_location: &'a SrcSpan,
```
giacomocavalieri: Can you add a code example showing the span this covers?

compiler-core/src/reference.rs
```diff
@@ -12,7 +12,7 @@ use petgraph::{
 pub enum ReferenceKind {
     Qualified,
     Unqualified,
-    Import,
+    Import { as_name_location: Option<SrcSpan> },
```
giacomocavalieri: Is this the right place to add `as_name_location`? All other references are the same irrespective of their span, so this would be the only exception

language-server/src/rename.rs
```diff
@@ -258,8 +258,19 @@ fn alias_references_in_module(
             ReferenceKind::Unqualified | ReferenceKind::Alias => {
                 edits.replace(reference.location, params.new_name.clone())
             }
-            ReferenceKind::Import => {
-                edits.insert(reference.location.end, format!(" as {}", params.new_name));
+            ReferenceKind::Import { as_name_location } => {
+                match as_name_location {
+                    Some(as_name_location) if params.new_name == *name => edits.delete(
+                        // cut 4 chars earlier to include the " as "
+                        SrcSpan::new(as_name_location.start - 4, as_name_location.end),
```
giacomocavalieri: Mhm this means that it wouldn't work on code that is not formatted with the formatter, for example: `wibble     as     wobble`. For these kind of situations we tend to add the additional required positions to the data structure rather than hard coding the number of spaces we expect to see

giacomocavalieri (commented):
Hello! I've left some comments

compiler-core/src/reference.rs
gavinmorrow: I'm not quite sure. It's hard, because the reference location is still the original name (ie when triggering find reference on a `Wobble` elsewhere, it'll select the `Wibble` in `type Wibble as Wobble`).

Even if we decided that it should select the `Wobble` instead of `Wibble`, there's the case of when it's renamed to the original name (and so the ` as Wobble` needs to be removed). If we only store the location of the `Wobble`, then it's difficult to remove the ` as ` (as you noted in your comment below).

I'm currently playing around a little with having a `ReferenceKind::AliasedImport` (inspired by the way renaming modules works), but I'm not quite sure yet if it's working well/better than this approach.

gavinmorrow: I'm also possibly exploring using `module.ast.definitions.imports`. Haven't gotten far at all with that yet though.

gavinmorrow: The ast worked out pretty well imo! Tell me what you think.

language-server/src/rename.rs
gavinmorrow: Good catch, I'll add a test for that.

language-server/src/rename.rs
```diff
@@ -259,7 +259,51 @@ fn alias_references_in_module(
...
+                            .find(|unqualified_import| unqualified_import.name == *name)
+                    })
+                    .expect("import should exist in ast");
+
+                if let Some(as_name_location) = import.as_name_location {
+                    // `as ...` clause exists
+                    if import.name == params.new_name {
+                        // Remove the `as ...` clause
+                        edits.delete(SrcSpan {
+                            start: as_name_location.start - 1,
```
giacomocavalieri: Rather than this wouldn't it be better to delete everything from the end of the imported name to the end of the `as_name_location`?

giacomocavalieri (commented):
Sorry it took me so long to get back to this, I kind of lost it in my inbox!

language-server/src/rename.rs
gavinmorrow: Oh yeah that's a much smarter solution, thanks.

gavinmorrow:
@giacomocavalieri No worries, thanks for the review! I made changes based on your comment and rebased to resolve conflicts.

giacomocavalieri (approved):

compiler-core/src/ast.rs
```diff
@@ -1293,10 +1311,15 @@ impl<A, B, C> Definition<A, B, C> {
...
-    /// The location excluding the potential `as ...` clause, or the `type` keyword
+    /// The location excluding the potential `as ...` clause, or the `type` keyword.
+    /// For example, in `type Wibble as Wobble`, it covers `Wibble`.
     pub imported_name_location: SrcSpan,
+    /// The location of the `as ...` clause.
+    /// For example, in `type Wibble as Wobble`, it covers `as Wobble`.
+    pub as_name_location: Option<SrcSpan>,
     pub name: EcoString,
     pub as_name: Option<EcoString>,
+    pub is_upname: bool,
```
lpil: This has been designed so that there can be an alias name but no location, or an alias location but no name. These are invalid states, so please combine them into a single optional. It could be a struct named `Alias`

compiler-core/src/ast.rs
lpil: The first position in this span is redundant as the end of the imported name location can be used.

Let's change this `UnqualifiedImport` type to be more in keeping with the rest of the codebase by having a `location` cover the code from the start of the name to the end of the alias. The location of the name and the alias can be calculated by adding or subtracting the byte length of the string from the start or end positions.

compiler-core/src/build.rs
```diff
@@ -420,7 +420,11 @@ pub struct UnqualifiedImport<'a> {
     pub name: &'a EcoString,
     pub module: &'a EcoString,
     pub is_type: bool,
+    pub is_upname: bool,
     pub location: &'a SrcSpan,
+    /// The location excluding the potential `as ...` clause, or the `type` keyword.
+    /// For example, in `type Wibble as Wobble`, it covers `Wibble`.
+    pub imported_name_location: &'a SrcSpan,
```
lpil: And the feedback for the other unqualified import struct applied to this class too please 🙏

lpil (commented):
Lovely work!! I've left a tiny bit of feedback inline. Thank you!

compiler-core/src/ast.rs
ankddev: In my PR I got this location directly with this:

https://github.com/gleam-lang/gleam/pull/5863/changes#diff-a30c847e65e9aedbd447132d28c0f2cccc8c85eaac4f49ced6f5115f2a0c4f94R122-R126

gavinmorrow: oh when I did that before Giacomo said to do it differently lol
(https://github.com/gleam-lang/gleam/pull/5545#discussion_r3241299178)

lpil:
Hello! I've got a bit confused with the comments. Are you waiting for something from me?

gavinmorrow:
@lpil Sorry, I wasn't waiting on you. I'd gotten a bit confused by the whole mess, but I finally had the time to sit down and work through it; now it's cleaner+rebased and should address your comments :]

compiler-core/src/ast.rs
```diff
@@ -1310,10 +1328,12 @@ impl<A, B, C> Definition<A, B, C> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct UnqualifiedImport {
     pub location: SrcSpan,
-    /// The location excluding the potential `as ...` clause, or the `type` keyword
+    /// The location excluding the potential `as ...` clause, or the `type` keyword.
+    /// For example, in `type Wibble as Wobble`, it covers `Wibble`.
     pub imported_name_location: SrcSpan,
     pub name: EcoString,
     pub as_name: Option<EcoString>,
+    pub is_upname: bool,
```
lpil: Is this field redundant? Seems it could be a method that does `self.name.chars().next().map_or(false, |c| c.is_uppercase())`

language-server/src/reference.rs
```diff
@@ -41,13 +41,27 @@ pub enum Referenced {
         module: EcoString,
         name: EcoString,
         location: SrcSpan,
+        /// The span that is considered to be referencing this name.
+        ///
+        /// This should be the same value as `location` in all instances except
+        /// when triggering on an aliased import, e.g. `wibble as wobble`. Then,
+        /// `location` will cover `wobble` while `trigger_location` will cover
+        /// `wibble as wobble`.
+        trigger_location: SrcSpan,
```
lpil: The start of this span is always the same as the start of `location`, so it can be replaced with just the end position. The names are a bit confusing to me. Perhaps `location` to cover all of them, and `name_end` for the position of the end of `wibble` in `wibble as wobble`. Would that be clearer?

And the same for the `ModuleType` variant too please.

lpil (commented):
Thank you!! Couple tiny notes inline.

language-server/src/reference.rs
gavinmorrow: I think it's the end that's always the same.

There are five cases:

1. Non-aliased import of a value: `import m.{wibble}` → location: `wibble`, trigger: `wibble`
2. Non-aliased import of a type: ` import m.{type Wibble}` → location: `Wibble`, trigger: `type Wibble`
3. Aliased import of a value: `import m.{wibble as wobble}` → location: `wobble`, trigger: `wibble as wobble`
4. Aliased import of a type: `import m.{type Wibble as Wobble}` → location: `Wobble`, trigger: `type Wibble as Wobble`
5. Any other value/type not in an import statement. location and trigger are the same.

Writing this out, I think that trigger location could be replaced with location, and then what's currently called location can probably be calculated by taking the last `name.len()` characters of that. (Although I'm not sure at the moment if name refers to aliased name or the original one.)

lpil: Oh yes, that's much better than what I suggested!

gavinmorrow: I had to go with `name_start`, because the `name` stored in the struct is the original name and not the aliased name.

gavinmorrow:
@lpil I addressed your comments.

language-server/src/reference.rs
lpil: It's still using a span rather than a position!

gavinmorrow: Which piece of code are you talking about? The new field in [`ModuleValue`](https://github.com/gleam-lang/gleam/pull/5545/changes#diff-7d52d97f527b91ecefa5f18dc9c0427c2ee09356f9380d2335f764b4952ab6f6R44-R49) and [`ModuleType`](https://github.com/gleam-lang/gleam/pull/5545/changes#diff-7d52d97f527b91ecefa5f18dc9c0427c2ee09356f9380d2335f764b4952ab6f6R57-R64) is a u32.

If you're talking about in [`build::UnqualifiedImport`](https://github.com/gleam-lang/gleam/pull/5545/changes#diff-746727aa30255b43313e9a0abd778e66d380280227bb41ba4039b36e9a8117ccR429-R431) or [`ast::UnqualifiedImport`](https://github.com/gleam-lang/gleam/pull/5545/changes#diff-cf64077b634ff405777db0547679ca9c43af162510737304e50b3507839e4f1eR1327-R1329)—that needs to be a span, because there is an unknown amount of characters before it (there could be the `type` keyword) and an unknown amount of characters after it (there could be an `as ...` clause).

lpil: We only need to know the position at the start of `Wibble` in `type Wibble as Wobble` because the end can be calculated like so:

```rust
pub fn name_location(&self) -> SrcSpan {
  SrcSpan::new(self.name_position, self.name_position + self.name.len())
}
```

The `is_type` field is also redundant with the addition of this data as an `is_type` method can be implemented like this:

```rust
pub fn is_type(&self) -> bool {
  // If the position is not at the start location then that
  // means that there is a `type` keyword before it, along
  // with an amount of whitespace and optional comments
  self.name_position != self.location.start
}
```

gavinmorrow: Oh yes I see what you mean. Thank you!!

gavinmorrow:
@lpil Thank you for the explanation! I believe that I fixed it.

lpil (approved):
Thank you!!!

PR #5556, opened 2026-04-03
LSP: Support references when triggered from alias

### LINKED ISSUE #5542: LSP: Support references when triggered from alias
For example, in the program

```gleam
import option.{map as m}
//             ^^^ A  ^ B

fn main() {
  let _ = option.map(option.Some(Nil), fn(_) { todo })
  //             ^^^ C
  let _ = m(option.Some(Nil), fn(_) { todo })
  //      ^ D
}
```

Moving the cursor over position C and triggering go to references will show positions A, C, and D. However, doing the same over position D will fail. At the very least, it should have the same behavior as triggering over position C. In my opinion, it would make sense for triggering over either to highlight all of A through D, although I can see either way.

Interestingly, this seems to work fine for types and type variants, although not for functions or constants.

gavinmorrow:
Fixes #5542.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,7 +122,18 @@
   ([Cory Forsstrom](https://github.com/tarkah))
 
 - The language server now supports finding references when triggered on an
-  aliased import (i.e. something that was imported and aliased via `as`).
+  aliased import. For example, in the following snippet, moving the cursor
+  over `log` and triggering "find references" will show all references of
+  `io.println()`.
+
+  ```gleam
+  import gleam/io.{println as log}
+  fn main() {
+    log("Hello, world!")
+  //^^^ trigger here
+  }
+  ```
+
   ([Gavin Morrow](https://github.com/gavinmorrow))
 
 ### Formatter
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -234,6 +234,10 @@
   qualified type.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The language server now supports finding references when triggered on an
+  aliased import (i.e. something that was imported and aliased via `as`).
```
giacomocavalieri: Rather than explaining it I would should an example that highlights the
references. You can take some of the other changelog entries as reference
```suggestion
  aliased import. For example, in the following snippet...
```

giacomocavalieri (commented):
Really nice! I left a small note

giacomocavalieri (approved):

lpil (approved):
Thank you!!!

lpil (approved):
Thank you!

lpil (approved):
Thank you!!!

lpil (approved):
Thank you!!!

PR #5563, opened 2026-04-04
Formatter

giacomocavalieri:
I was investigating the formatting of some long patterns, turns out we could already handle that but it was missing a test so I added one for it.
I also noticed the pretty printer could avoid boxing in some places and renamed some variables since I was modifying the module

### DIFF
--- a/compiler-core/src/format.rs
+++ b/compiler-core/src/format.rs
@@ -387,7 +387,7 @@ impl<'comments> Formatter<'comments> {
 
             Definition::TypeAlias(alias) => self.type_alias(alias),
 
-            Definition::CustomType(ct) => self.custom_type(ct),
+            Definition::CustomType(custom_type) => self.custom_type(custom_type),
 
             Definition::Import(Import {
                 module,
@@ -462,7 +462,7 @@ impl<'comments> Formatter<'comments> {
                     .append(name.as_str());
                 let head = match annotation {
                     None => head,
-                    Some(t) => head.append(": ").append(self.type_ast(t)),
+                    Some(type_) => head.append(": ").append(self.type_ast(type_)),
                 };
                 head.append(" = ").append(self.const_expr(value).group())
             }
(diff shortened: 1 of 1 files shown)


compiler-core/src/pretty.rs
```diff
@@ -764,7 +764,23 @@ impl<'a> Document<'a> {
     /// outermost group is broken, any sub-groups might be rendered broken
     /// or unbroken, depending on whether they fit on a single line.
     pub fn group(self) -> Self {
-        Self::Group(Box::new(self))
+        match self {
+            // Grouping a group doesn't change how it will be formatted so we
+            // can avoid boxing it.
+            Document::Group(_)
```
lpil, line 770: Sorry, quick question: Do we have code that does this today? It wraps groups in groups?

giacomocavalieri: I don't have exact numbers because it's a bit hard to tell just by looking at the code but there's a few places where we build documents by piecing together smaller ones like this `some_fun().group()` and it can end up wrapping the document in more groups than necessary. A few examples:
https://github.com/gleam-lang/gleam/blob/768b1ee5efc6f17a1e2e178edd41be8e3b540076/compiler-core/src/erlang.rs#L1281
Here we build statements and group each one; each statement in turn is produced by the following function:
```rs
fn statement<'a>(
    statement: &'a TypedStatement,
    env: &mut Env<'a>,
    position: Position,
) -> Document<'a> {
    match statement {
        Statement::Expression(e) => expr(e, env),
        Statement::Assignment(a) => assignment(a, env, position),
        Statement::Use(use_) => expr(&use_.call, env),
        Statement::Assert(a) => assert(a, env),
    }
}
```
And each of those calls, especially `expr` can end up producing groups based on the kind of expression that's being printed:
```rs
fn expr<'a>(expression: &'a TypedExpr, env: &mut Env<'a>) -> Document<'a> {
    match expression {
        // this one doesn't produce a group
        TypedExpr::Panic {
            location, message, ..
        } => panic(*location, message.as_deref(), env),
        // this one does
        TypedExpr::Fn { .. } => ...,
        // and so on...
    }
}
```

So yes we might end up doing that quite frequently, but it's basically impossible to avoid the way documents are meant to be composed together

giacomocavalieri: Actually I do have some numbers! I tried this on my `squirrel` repo to try it out on some decently sized codebase and it avoids boxing 6796 times!

lpil: ahahaha amazing

PR #5571, opened 2026-04-05
Work around Node bug on Windows

lpil:


giacomocavalieri:
Seems like https://github.com/gleam-lang/gleam/pull/5565 fixes this, so no need to disable the tests! 🎉

PR #5597, opened 2026-04-09
Avoid false target-support edges from imported module aliases

### LINKED ISSUE #5507: False edges in the call graph can cause calls to unsupported functions to typecheck and be emitted
Given the following module in a **dependency:**

```gleam
import gleam/option

@external(javascript, "../gleam_stdlib/gleam/function.mjs", "identity")
pub fn id(a: a) -> a

fn wibble() -> option.Option(Int) {
  option.Some(id(42))
}

pub fn option() -> option.Option(Int) {
  wibble()
}

```

calling `wibble.option()` from the root package type-checks even on the Erlang target, but errors in the erlang compiler with a undefined function `wibble/0` error.

`option` should not be supported on the Erlang target, since wibble calls id, which is Javascript-only. We hit this bug in practice in the Tiramisu release candidate, which primarily targets Javascript, but includes some functions to be used with server-side Lustre and support both targets. I think @renatillas solved this by adding a bunch of `@target` annotations instead. 

What's happening here is that the `option.Some` call creates a false edge in the call graph to the local `option` function. When the compiler analyses target support, all functions start out as being supported on all targets. Because of the false edge, `wibble` and `option` create a fake circular dependency, so here `option` is analysed before `wibble`. It sees `wibble` is still supported on both targets as that's the default and is emitted in the resulting Erlang file. After, `wibble` of course is not supported and is not emitted, but `option` still includes a call to a now missing `wibble` local function.

~ 💜

officialasishkumar:

Fixes #5507

This excludes imported module aliases from the local call graph used for dependency ordering, preventing module-select expressions such as `option.Some(...)` from creating false edges to same-named local functions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@
   finds.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler no longer treats module-select expressions as dependencies on
+  same-named local functions, which avoids false target-support ordering in
+  dependency packages.
+  ([Asish Kumar](https://github.com/officialasishkumar))
+
 - The compiler now shows a better error message when trying to use the record
   update syntax with variants that have no labelled fields.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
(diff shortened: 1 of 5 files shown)


lpil:
Hello, are you a robot?

lpil:
Closing this as there was no reply. Please feel free to reopen!

Issue #5605, opened 2026-04-10
1.16.0-rc1: Wrap in anonymous functions does weird things in combination with `use`


yoshi-monster:
```gleam

pub fn main() {
  use a <- apply(5)
  a * 1
  // ^ Wrap in anonymous function
}

fn apply(x, k) {
  k(x)
}
```

results in

```gleam
  fn(int) { a * 1(int) }
```

It also wraps the `use` statement with multiple `use`s, but I think that's the same thing. I was originally confused where the braces went :)
It is offered multiple times, which is confusing maybe.

lpil:
oh whoa, that is super messed up. I'm not sure exactly what is happening here, but I'm guessing we should not be offering the code action at all here.

ankddev:
@lpil this issue could be closed

PR #5608, opened 2026-04-11
LSP: don't suggest 'Wrap in anonymous function' with 'use' body

GioMaz:
Issue: #5605

This makes it so the `Use` function literal kind is also excluded from the `Wrap in anonymous function` action 


ankddev (commented):
Hello @GioMaz! I think that you should add a changelog entry, otherwise it looks good to me!

lpil (approved):
Looks fab! Thank you!!

And welcome to Gleam!

PR #5609, opened 2026-04-13
Fix formatting of binary operations in bit array size segments

### LINKED ISSUE #5606: Formatter unnecessarily wraps expressions inside bit array size segments in a block.
Given this like:

```gleam
  let slice = <<bits:bits-size(count * marshal_zone_size)-unit(8)>>
```

The formatter will modify it to:

```gleam
  let slice = <<bits:bits-size({ count * marshal_zone_size })-unit(8)>>
```

Even though the original is less than 80 characters wide.

ankddev:
* Closes #5606

***

### DIFF
--- a/compiler-core/src/format/tests/bit_array.rs
+++ b/compiler-core/src/format/tests/bit_array.rs
@@ -510,7 +510,7 @@ fn simple_single_line_bit_array_with_no_trailing_comma_is_split_one_item_per_lin
 
 // https://github.com/gleam-lang/gleam/issues/5606
 #[test]
-fn dont_wrap_expression_size_segments_if_line_is_short() {
+fn dont_wrap_binary_operators_size_segments_if_line_is_short() {
     assert_format!(
         "pub fn main() {
   let wubble = <<bits:bits-size(wibble * wobble)-unit(8)>>
@@ -521,7 +521,7 @@ fn dont_wrap_expression_size_segments_if_line_is_short() {
 
 // https://github.com/gleam-lang/gleam/issues/5606
 #[test]
-fn dont_wrap_expression_size_segments_if_line_is_quite_short() {
+fn dont_wrap_binary_operators_size_segments_if_line_is_quite_short() {
     assert_format_rewrite!(
         "pub fn main() {
   let some_long_variable_name = <<bits:bits-size(some_other_variable * wibbling_wobbling)-unit(8)>>
@@ -538,7 +538,7 @@ fn dont_wrap_expression_size_segments_if_line_is_quite_short() {
 
 // https://github.com/gleam-lang/gleam/issues/5606
 #[test]
-fn wrap_expression_size_segments_if_line_is_long() {
+fn wrap_binary_operators_size_segments_if_line_is_long() {
     assert_format_rewrite!(
         "pub fn main() {
   let some_long_variable_name = <<bits:bits-size(some_other_long_variable_name * and_some_other_long_variable_name)-unit(8)>>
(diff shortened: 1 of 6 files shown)


compiler-core/src/format/tests/bit_array.rs
```diff
@@ -507,3 +507,52 @@ fn simple_single_line_bit_array_with_no_trailing_comma_is_split_one_item_per_lin
...
+  let some_long_variable_name = <<bits:bits-size(some_other_long_variable_name * and_some_other_long_variable_name)-unit(8)>>
+}
+",
+        "pub fn main() {
+  let some_long_variable_name = <<
+    bits:bits-size({
+      some_other_long_variable_name
+      * and_some_other_long_variable_name
+    }
+    )-unit(8),
```
ankddev, line 553: I couldn't fix this newline after `}`, since there seems no way to insert something when line is too long, but don't insert newline. 

CC @giacomocavalieri since you've worked on a formatter a lot, as I've seen, and may know what should we do here

giacomocavalieri: Oh I totally missed the ping, sorry! I'll think the way to go is use a `break_` document for that but I'm not really sure, I'll have to have a more in depth look into this

ankddev: No worries! I already use `break_` here, but it emits break document that is formatted with newline if broken, while here we want just document without newline.

giacomocavalieri: Aha I think I found the fix: it's `bit_array_segment_expr` that adds the unnecessary wrapping, what we should do is remove that entirely and replace it with the regular `self.expr` and everything should be fine.
At this point I'd also add some tests to make sure we respect the programmer choice of using curly braces or not:
```rs
#[test]
fn bit_array_with_size_is_not_unnecessarily_wrapped() {
    assert_format!(
        "pub fn main() {
  let some_variable = <<bits:bits-size(some * other)>>
}
"
    );
}

#[test]
fn formatter_respects_user_inserted_parentheses() {
    assert_format!(
        "pub fn main() {
  let some_variable = <<bits:bits-size({ some * other })>>
}
"
    );
}
```
With the suggested change all tests should be green, let me know how that goes!

ankddev: Yeah, now everything seems to work, thank you a lot! I think that I misunderstood issue a bit intially.

giacomocavalieri: Thank you for looking into this!

compiler-core/src/format/tests/bit_array.rs
```diff
@@ -507,3 +507,62 @@ fn simple_single_line_bit_array_with_no_trailing_comma_is_split_one_item_per_lin
...
+#[test]
+fn dont_insert_curly_braces_with_binary_operators_when_line_is_long() {
+    assert_format_rewrite!(
+        "pub fn main() {
+  let some_long_variable_name = <<bits:bits-size(some_other_long_variable_name * and_some_other_long_variable_name)-unit(8)>>
+}
+",
+        "pub fn main() {
+  let some_long_variable_name = <<
+    bits:bits-size(some_other_long_variable_name
```
giacomocavalieri: Mhm this does look a bit off to me, maybe we should fix it in this same pr, I think it should be formatted as:
```gleam
<<
  bits:bits-size(
    some_other_long_variable_name
    * and_some_...
  )-unit(8),
>>
```
Basically the same as if there were curly braces. Do you think you could try and have a look at this as well?

ankddev: Yeah, I thought that it's strange too. Looking at the code, it seems simple, so I'll do this!

ankddev: @giacomocavalieri Done!

giacomocavalieri (approved):
Brilliant, thank you!

lpil (approved):
Lovely!!! Thank you

PR #5614, opened 2026-04-17
Share dependency licences 

lpil:
- [x] Licences are kept in the repo
- [x] Dependencies are credited in the repo
- [x] A HTML page detailing all the licences is generated
- [x] Dependencies and their licences are included in releases


licence-bundler/test/licence_bundler_test.gleam
```diff
@@ -0,0 +1,13 @@
...
+  gleeunit.main()
+}
+
+// gleeunit test functions end in `_test`
+pub fn hello_world_test() {
+  let name = "Joe"
+  let greeting = "Hello, " <> name <> "!"
+
+  assert greeting == "Hello, Joe!"
+}
```
giacomocavalieri, line 5: Should this package have tests? Should tests for this package also run in CI?

licence-bundler/test/licence_bundler_test.gleam
```diff
@@ -0,0 +1,13 @@
+import gleeunit
+
+pub fn main() -> Nil {
+  gleeunit.main()
+}
+
+// gleeunit test functions end in `_test`
+pub fn hello_world_test() {
```
giacomocavalieri: I'd get rid of the default test

licence-bundler/src/licence_bundler.gleam
```diff
@@ -0,0 +1,220 @@
...
+    )
+
+  let rust_dependencies =
+    output
+    |> string.split("\n")
+    |> list.filter(fn(line) { line != "" })
+    |> list.map(trim_line_prefix)
+    |> list.filter_map(dependency_information(_, gleam_crates, active_licences))
+    |> set.from_list
+    |> set.to_list
```
giacomocavalieri: It doesn't make any difference in practice, but this is what `list.unique` does more efficiently!

licences/MPL-2.0.txt
```diff
@@ -0,0 +1,319 @@
+Mozilla Public License, version 2.0
+
+Copyright (c) {YEAR}, {AUTHOR}
```
giacomocavalieri, line 3: I'm not sure how all of this works, but should these licenses have the `{YEAR}` and `{AUTHOR}` placeholders filled in with some actual values?

licence-bundler/src/licence_bundler.gleam
```diff
@@ -0,0 +1,220 @@
...
+  let assert Ok(output) =
+    shellout.command(
+      run: "cargo",
+      with: [
+        "tree",
+        "--format",
+        "{p} {l}",
+        "--edges",
+        "no-dev,no-build,no-proc-macro",
+      ],
```
giacomocavalieri, line 35: I would like to have a comment explaining why this particular combination of rust incantations is used, and what format it's expected to produce. It would also make it easier to review the later parsing

licence-bundler/src/licence_bundler.gleam
```diff
@@ -0,0 +1,220 @@
...
+    active_licences
+    |> set.to_list
+    |> list.sort(string.compare)
+    |> list.try_map(fn(identifier) {
+      simplifile.read("../licences/" <> identifier <> ".txt")
+      |> result.map(Licence(identifier:, text: _))
+    })
+
+  let h = htmb.h
+  let t = htmb.text
```
giacomocavalieri, line 64: This is just a shortcut to type less, I would use `htmb.text` and `htmb.h` directly! Or at most import those two functions unqualified

licence-bundler/src/licence_bundler.gleam
```diff
@@ -0,0 +1,220 @@
...
+    })
+  set.from_list(crates)
+}
+
+type Crate {
+  Crate(
+    name: String,
+    version: String,
+    licence_identifier: String,
+    used_licences: List(String),
```
giacomocavalieri, line 154: What's the difference between `licence_identifier` and `used_licences`? Is the identifier the license of the create, and the used ones the licenses of its dependencies? A comment would help here

LICENCE
```diff
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright 2016 - present Louis Pilfold
+   Copyright [yyyy] [name of copyright owner]
```
jtdowney, line 189: Looks like this was an accidental revert to the template

licence-bundler/src/licence_bundler.gleam
```diff
@@ -0,0 +1,220 @@
...
+
+  let html =
+    h("html", [#("lang", "en")], [
+      h("head", [], []),
+      h("body", [], [
+        h("h1", [], [t("Gleam dependency licences")]),
+        h("section", [], [
+          h("h2", [], [t("Rust dependency packages")]),
+          h_ul(rust_dependencies, fn(package) {
+            assert licences != [] as { package.name <> " missing licences" }
```
jtdowney, line 81: Should this be `package.used_licences`?

LICENCE
lpil: I switched back to the default speculatively to make it easier to update lots of repos using future automated tooling, but that was a bit hasty of me and I should have left it until I was actually doing that work.

licence-bundler/src/licence_bundler.gleam
lpil: What would be the advantage of that over this scoped shortcut?

giacomocavalieri: This doesn't quite feel gleam-y I'd expect to see the fully qualified function being used rather than this shortcut

licence-bundler/test/licence_bundler_test.gleam
lpil: We can just run the program 👍

licences/MPL-2.0.txt
lpil: Not unless we want to have thousands of copies of this same file!

licence-bundler/src/licence_bundler.gleam
lpil: One is a SPDX expression of the possible and required licences, one is the licences actually used by us.

licence-bundler/src/licence_bundler.gleam
lpil: Well spotted, thank you. Copy paste fail

PR #5616, opened 2026-04-19
Todo in constants

giacomocavalieri:
This PR closes #5496 
It is now possible to use `todo`s in constants!

---

Reminder for me: once this is merged we should update the syntax highlighting grammar to allow this!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
   analyse the fields that are being provided.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- It is now possible to use the `todo` keyword in constants, this will result in
+  an helpful error message rather than a syntax error.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
--- a/compiler-core/src/ast/constant.rs
+++ b/compiler-core/src/ast/constant.rs
@@ -90,6 +90,11 @@ pub enum Constant<T, RecordTag> {
         /// states.
         extra_information: Option<InvalidExpression>,
     },
+
+    Todo {
+        location: SrcSpan,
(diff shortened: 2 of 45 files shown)


lpil (commented):
Do we want to support `as "blah"`?

giacomocavalieri:
Would that only be a literal string given it's four constants and only a compile time error?

lpil:
Fair enough :)

test-project-compiler/src/generated_tests.rs
```diff
@@ -3,6 +3,39 @@
 
 use gleam_core::build::Mode;
 
+#[rustfmt::skip]
+#[test]
+fn .DS_Store_dev() {
+    let output = crate::prepare("./cases/.DS_Store", Mode::Dev);
```
lpil: What's this? 🤔

giacomocavalieri: Oh wow that's so funny. That must be macOS putting its .DS_Store file in the `cases` directory and rust generating a test for it as well

giacomocavalieri:
So I've changed the code so that it would also accept a message, in the end I had the message be any constant expression that is then type checked to be a string meaning that we can reuse a lot of the existing code and if you type `todo as [1, 2, 3]` you get a nice type mismatch error rather than a syntax error. The only catch is that this means one can also write this:
```gleam
const message = "wibble" 
const other = todo as message
```
And that is a single error for the incomplete code, but not an error for using another constant in the message. It would be pretty straightforward to add an additional check to also raise an error for that. What do you think we should do?

lpil:
Sorry, I'm a bit confused. Why would there be an error for the constant message?

giacomocavalieri:
> Sorry, I'm a bit confused. Why would there be an error for the constant message?

No specific reason, just in case we didn't want to allow putting anything in there except for a literal string. Right now there's no error message for String constants

lpil:
Using a constant seems fine to me!

PR #5620, opened 2026-04-20
Fix dependency order sorting

### LINKED ISSUE #5618: Type inference breaks with recursive functions (?)
The following program type-checks:

```gleam
pub fn main() {
  // TypeError: undefined is not an object
  let _ = step(Leaf, Leaf)
  Nil
}

type Tree {
  Leaf
  Node(List(Tree))
}

fn step(left: Tree, right: Tree) -> Int {
  // here, we pass a List(a) as "effect" ...
  sum(diff(left, right, []))
}

// sum is here to force the return type to List(Int)
fn sum(xs: List(Int)) -> Int {
  case xs {
    [] -> 0
    [x, ..xs] -> x + sum(xs)
  }
}

fn diff(left, right, effect) {
  case left, right {
    Node(olds), Node(news) -> diff_batch(olds, news, effect)
    _, _ -> start(right, effect)
  }
}

fn diff_batch(left, right, effect) {
  case left, right {
    [old, ..left], [new, ..right] ->
      diff_batch(left, right, diff(old, new, effect))
    _, _ -> effect
  }
}

fn start(node, effect) {
  case node {
    // But here, the return type should be forced to String!
    Leaf -> "wat"
    Node(children) -> start_batch(children, effect)
  }
}

fn start_batch(children, effect) {
  case children {
    [] -> effect
    [child, ..rest] -> start_batch(rest, start(child, effect))
  }
}
```

I tried minifying it as much as possible, I couldn't find anything else that i can remove from this program.

Hovering the functions, the LSP shows the correct types for all of them, but the call side expression has the wrong return type.

~ 💜

GearsDatapacks:
Fixes #5618 

Turns out the `petgraph` crate has a function that does exactly what we need here, so I've entirely removed our buggy implementation. I've had to reshuffle some of our tests as that function produces a slightly different output than our did, but I've checked and it's only inconsequential changes like swapping the order of two leaves.


lpil (approved):
Amazing work as always. Thank you !

Issue #5622, opened 2026-04-20
Sourcemaps are not valid without further asset processing


lpil:
The sourcemap functionality was originally tested with nodejs, but unfortunately it does work in the browser without further processing. Gleam should always output artefacts that can be used without modification, so we want to make it work with any tooling.

The problems are:
1. The source map `sources` key uses an absolute file system path instead of a relative HTTP path.
2. The source map references the input Gleam file, but that file is not in the output directory, so it not accessible over HTTP.

Proposed solution:

1. Change the absolute path to a relative path to the `.gleam` file in the same directory. e.g. the source map for `wibble.mjs` would specify `"sources":["lesson.gleam"]`
2. When source map generation is enabled hard-link the source file into the output directory, to reside next to the generated JavaScript file and the source map file.

We would need to work out some system for deciding when to remove the source file from the directory when the original file has been deleted or replaced. Perhaps this could be an issue for after the v1.16 release?

PR #5623, opened 2026-04-20
Use URL for source-map source

lpil:
Part of https://github.com/gleam-lang/gleam/issues/5622


compiler-core/src/javascript.rs
```diff
@@ -141,13 +141,12 @@ impl<'a> Generator<'a> {
             typescript,
             source_map_builder: if source_map {
                 let module_name = module.name.clone();
-                // Need the absolute path to the module to be able to map the
-                // destination location back to the source location.
-                let src_path_str = path.as_str();
                 let output_path = format!("{module_name}.mjs");
+                let module_alias = module_name.split('/').next_back().unwrap_or(&module_name);
+                let input_path = format!("{module_alias}.gleam",);
```
Acepie, line 146: this seems like it would cause more issues until the gleam source is copied over. if the issue before was that the file path was not being respected because its not a url, why don't we append the file protocol for now and wait until we are also copying the source into the output directory to change this to relative. As is this seems like it would cause more issues than it fixes since the path would definitely not resolve

Acepie: I tested out just adding that protocol and that seems to fix the resolution issues for a lustre dev tools project (i used https://github.com/giacomocavalieri/polish_practice as the test case)

lpil: File protocol only works in NodeJS etc, so it's not an option.

Yup having the source file available is part 2 of the linked proposal

Acepie (approved):
This all makes sense to me. Something that occurred to me as a bit today problem but maybe for the long term is that since this now includes the gleam source in the build output there might be an argument to having a separation between dev and prod output since in prod you generally want to explicitly strip out the source and mapping. Not for this pr but something to consider here long term

compiler-core/src/build/package_compiler.rs
```diff
@@ -448,9 +448,44 @@ where
...
 
+    /// Link the module source .gleam directory to the output directory.
+    /// This is done if we are generating source maps.
+    ///
+    fn link_module_source_file_to_out(&mut self, module: &Module) -> Result<(), Error> {
+        let source = module.input_path.as_path();
+        let destination = self.out.join(module.name.as_str()).with_extension("gleam");
+        let file_at_destination = self.io.exists(&destination);
+
+        // If the file does not exist then linking is needed.
```
giacomocavalieri: When can it happen that the file exists already? Is it from previous runs of the compiler? A comment explaining it would be nice!

compiler-core/src/build/package_compiler.rs
```diff
@@ -448,9 +448,44 @@ where
...
+                self.link_module_source_file_to_out(module)?;
+            }
+        }
+
         Ok(())
     }
 
+    /// Link the module source .gleam directory to the output directory.
+    /// This is done if we are generating source maps.
+    ///
```
giacomocavalieri: I'm having a hard time parsing this sentence, what does this mean "Link the module source .gleam directory", is it linking the `.gleam` files in the `src/` directory?

compiler-core/src/javascript/tests.rs
```diff
@@ -270,7 +270,14 @@ pub fn append_line_numbers(src: &str) -> String {
 // Pretty-print a sourcemap to a string that might be readable by humans.
 pub fn source_map_to_string(src: &str, compiled: &str, source_map: SourceMap) -> String {
     let mut output = String::new();
-    output.push_str("Mappings:\n");
+    output.push_str("File: ");
+    output.push_str(source_map.get_file().unwrap_or_default());
```
giacomocavalieri: Maybe I'm not understanding this right, but shouldn't the test fail if the file is not there? Getting a default value seems easier to forget to check in the snapshot

compiler-core/src/build/package_compiler.rs
lpil: Oops, I have written "directory" instead of "file" at one point by mistake.

PR #5635, opened 2026-04-24
Improve error wording

lpil:


compiler-core/src/config.rs
```diff
@@ -71,7 +71,7 @@ impl<'de> serde::de::Visitor<'de> for SpdxLicenseVisitor {
     {
         match spdx::license_id(value) {
             None => Err(serde::de::Error::custom(format!(
-                "{value} is not a valid SPDX License ID"
+                "{value} is not a known SPDX License identifier"
```
00-kat, line 74: Maybe a link to https://spdx.org/licenses might help.

PR #5637, opened 2026-04-25
Fix extract variable used on records with 0 arity

### LINKED ISSUE #5629: Extract variable code action does not work for variants with no fields
Image

giacomocavalieri:
This PR closes #5629

### DIFF
--- a/language-server/src/code_action.rs
+++ b/language-server/src/code_action.rs
@@ -3476,6 +3476,37 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
         });
     }
 
+    fn visit_typed_expr_call(
+        &mut self,
+        location: &'ast SrcSpan,
+        type_: &'ast Arc<Type>,
+        fun: &'ast TypedExpr,
+        arguments: &'ast [TypedCallArg],
+        open_parenthesis: &'ast Option<u32>,
+    ) {
+        // Function calls need some extra care. If we're inspecting a record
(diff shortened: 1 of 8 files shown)


language-server/src/code_action.rs
```diff
@@ -3568,6 +3603,15 @@ impl<'ast> ast::visit::Visit<'ast> for ExtractVariable<'ast> {
...
                 });
             }
+
+            // Expressions that don't make sense to extract
+            TypedExpr::Panic { .. }
+            | TypedExpr::Echo { .. }
+            | TypedExpr::Block { .. }
+            | TypedExpr::ModuleSelect { .. }
+            | TypedExpr::Invalid { .. }
+            | TypedExpr::PositionalAccess { .. }
```
lpil, line 3613: I think extracting positional access and module select could make sense? I might want to extract a long chain of them to a variable for clarity, like `one.two.three.four`

giacomocavalieri: That's what the `TypedExpr::RecordAccess` represents, module select is for calling a function `list.map`, I don't think that extracting it ever makes much sense, for example:
```
list.map([], int.parse)
//           ^^^ Extract
// Producing:
let function = int.parse
list.map([], function)
```
And `PositionalAccess` is generated internally for unlabelled fields.
We already support your example, I've added a couple of tests just to make sure!

lpil: Oh doh! I mixed up record and module 🤦

lpil (approved):
Thank you!!

PR #5642, opened 2026-04-26
Fix code generation for guards with lists with a tail

### LINKED ISSUE #5641: Invalid generated code for list constants used in guards
The following code:
```gleam
pub fn go(x: List(Int), y: List(Int)) {
  case x {
    [1, ..x] if [1, 2, ..x] == y -> True
    _ -> False
  }
}
```
Doesn't produce the correct code on either target:
```erl
go(X, Y) ->
    case X of
        [1 | X@1] when [1, 2] =:= Y ->
%                      ^^^^^^ Missing the X tail
            true;

        _ ->
            false
    end.
```
```js
export function go(x, y) {
  if (x instanceof $Empty) {
    return false;
  } else {
    let $ = x.head;
    if ($ === 1) {
      let x$1 = x.tail;
      if (isEqual(toList([1, 2]), y)) {
//                       ^^^^^^ Missing the x tail!
        return true;
      } else {
        return false;
      }
    } else {
      return false;
    }
  }
}
```
I'm working on a fix!

giacomocavalieri:
This PR closes #5641 
Working on constants I noticed that the newly added lists with a tail were not handled properly during code generation, leading to invalid code that could result in ignoring the tail segment. Quite tricky stuff!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,3 +81,7 @@
 - Fixed a bug where the language server would suggest the "wrap in anonymous"
   code action when hovering over a record update.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the compiler would generate invalid code for guards using
+  lists with a tail.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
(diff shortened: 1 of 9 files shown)


lpil (approved):
Thank you!!

PR #5643, opened 2026-04-26
Fix version tracking for constant list prepending

giacomocavalieri:
I noticed we forgot to update the minimum version tracking for prepending in list constants that was added in 1.16. This fixes it!


lpil (approved):
Thank you!!

PR #5645, opened 2026-04-27
Completions for list tails and record updates

giacomocavalieri:
I noticed the language server would not show completions when typing a record update or a list tail:
```gleam
Wibble(..wi|)
//         ^ No help here
[1, 2, ..t|]
//        ^ Nor here
```
I thought it would be an easy fix, but it kind of lead down a rabbit hole and fixing this required a couple of other things to get fixed first. Here's a recap:
- record updates needed to be more fault tolerant so that if the record being updated doesn't exist (e.g. `Wibble(..i_do_not_exist)`) the compiler doesn't entirely give up
- the language server was always suggesting all prelude values despite their type being incompatible with the current context

I've made sure to keep each item of work onto a separate commit and would recommend reviewing this going over each individual one!

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,3 +28,8 @@
 - Fixed a bug where the language server wouldn't let one extract record
   constructors as variables.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the language server would suggest completions for values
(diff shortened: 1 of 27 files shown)


lpil (approved):
What a lovely clean series of commits! Really super easy to review. Thank you!

PR #5646, opened 2026-04-27
Fix extract variable bug

giacomocavalieri:
I noticed I introduced a little bug with the recent changes to the "extract variable" code action, that would allow extracting a record update like this:
```gleam
Wibble(..wibble, a: 1)
// ^^ Trigger here
let function = Wibble
function(..wibble, a: 1)
// That's a bit silly
```

So I've added a check to make sure it doesn't happen, I haven't updated the changelog since the bug wasn't in any released version


lpil (approved):
Thank you!!!!!

PR #5647, opened 2026-04-27
Fix wrap in anonymous function code action applied on record updates

giacomocavalieri:
I noticed a small bug in the newly added "wrap in anonymous function" code action. The LS would allow to trigger it on record update expressions generating invalid code. This fixes it!


lpil (approved):
Thank you!!

PR #5649, opened 2026-04-27
generate valid code when matching on the rhs of another let

### LINKED ISSUE #5648: Pattern match on value inside an anonymous function produces invalid code
When using the "pattern match on value" code action on a value defined inside an anonymous function that is also on the right hand side of an assignment, the code action ends up putting the generated code outside of the anonymous function:
```gleam
pub fn main() {
  let outcome = apply(#(Ok(1), Nil), fn(pair) {
    let #(result, nil) = pair
    //    ^^^^^^ Triggering it here
  })
  // Will add the code here!
}

fn apply(a, f) { f(a) }
```
I'm working on a fix!

giacomocavalieri:
This PR fixes #5648 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,11 @@
   or git dependency on Hex when running `gleam update`.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the "pattern match on value" code action would generate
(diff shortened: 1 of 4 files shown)


lpil (approved):
Wow you're working fast. Thank you!!

giacomocavalieri:
I'm on a roll!

PR #5655, opened 2026-04-29
Use relative symlinks for priv dirs

### LINKED ISSUE #5430: Compilation breaks when renaming project directory
Renaming a gleam project directory causes a compilation error. Running the following shell commands (on Arch Linux):

```sh
gleam new repro
cd repro
gleam add lustre
gleam build
cd ..
mv repro repro_broken
cd repro_broken
gleam build
```

Result in the following error:

```
error: File IO failure

An error occurred while trying to link this file:

    /tmp/tmp.QzSnQdYrRy/repro_broken/build/dev/erlang/lustre/priv

    File exists (os error 17)
```

Running `gleam build` a second time seems to fix the error.

I routinely run into this because I run gleam projects via bind mounts in various development containers. It's a bit of a weird setup :D

I didn't do a super deep investigation but it seems related to the symlinks to `priv`, e.g. in `build/dev/erlang/lustre`:

```sh
ll build/dev/erlang/lustre
Permissions Size User   Group  Date Modified Name
...
lrwxrwxrwx     - rafael rafael  5 Mär 11:50  priv -> /tmp/tmp.QzSnQdYrRy/repro/build/packages/lustre/priv
```

Would it be an option to use relative links instead? Or to find broken symlinks and recreate them?

In both cases, this seems like an easy issue to fix, if someone is interested in giving some hints I'd like to make a PR :)

raffomania:

Fixes #5430.

Notes:

- I'm not super happy with extending the ProjectIO trait but think it's better than e.g. adding a flag to the existing `symlink_dir` 
- Would love to make the project test work offline, atm it has to download a dependency. I'm not sure how to trigger the symlinking behaviour without it

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -277,7 +277,8 @@ where
         let priv_build = self.out.join("priv");
         if self.io.is_directory(&priv_source) && !self.io.is_directory(&priv_build) {
             tracing::debug!("linking_priv_to_build");
-            self.io.symlink_dir(&priv_source, &priv_build)?;
+            self.io
+                .symlink_dir(&self.io.canonicalise(&priv_source)?, &priv_build)?;
         }
 
         let copier = NativeFileCopier::new(
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -4,8 +4,10 @@ mod tests;
 use crate::analyse::{ModuleAnalyzerConstructor, TargetSupport};
 use crate::build::package_loader::CacheFiles;
 
-use crate::error::{DefinedModuleOrigin, FailedModule, SkipReason, SkippedModule};
-use crate::io::files_with_extension;
+use crate::error::{
+    DefinedModuleOrigin, FailedModule, FileIoAction, FileKind, SkipReason, SkippedModule,
+};
+use crate::io::{files_with_extension, make_relative};
 use crate::line_numbers::{self, LineNumbers};
 use crate::type_::PRELUDE_MODULE_NAME;
 use crate::{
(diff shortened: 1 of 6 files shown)


lpil (commented):
Thank you! Very nice.

Rather than add a new relative version to the project IO trait could you instead add a new trait method for canonicalising a path, and then have the remaining logic be kept within the core compiler. This keeps the trait small, is more generally useful, and removes the need to duplicate the logic between all implementations + for them to use code from core.

Could you also add unit tests for the function that makes the new relative path, and a test-package-compiler test for this new functionality. We avoid those big integration tests as much as possible, and this could be tested sufficiently with those two.

compiler-core/src/build/package_compiler.rs
```diff
@@ -277,7 +277,7 @@ where
         let priv_build = self.out.join("priv");
         if self.io.is_directory(&priv_source) && !self.io.is_directory(&priv_build) {
             tracing::debug!("linking_priv_to_build");
-            self.io.symlink_dir(&priv_source, &priv_build)?;
+            self.io.symlink_dir_relative(&priv_source, &priv_build)?;
```
lpil: Could you add a documentation comment explaining what this is doing and why please 🙏 It'll be forgotten otherwise.

raffomania:
Thanks for the review!

> Rather than add a new relative version to the project IO trait could you instead add a new trait method for canonicalising a path

Just to make sure I got you right, do you mean to extract the existing call to `canonicalise` from the existing version of `fs::symlink_dir` into a separate trait method, so `fs::symlink_dir` will accept a relative path as the `src` arg? This would then require to check/change existing calls to `symlink_dir` to see if they need a canonicalised input or not.

raffomania:
@lpil I've added unit tests, lifted the canonicalise functionality from the `symlink_dir` function, added an explanatory comment. I've also split the changes into multiple commits for easier review. Let me know if you want them squashed. 

The `test-package-compiler` case would require some more work: [copy_native_files (the functionality doing the symlinking) is disabled in those tests](https://github.com/raffomania/gleam/blob/5a5447079c47caaee798154c11b20dc7fdb83f0e/test-package-compiler/src/lib.rs#L61), and the in-memory filesystem does not support symlinks or canonicalising yet. I'm not sure if I have the necessary experience with the codebase to make those changes without extensive review work from your side. If you think it'll work, I'm happy to tackle this, but would like to ask some questions on the approach.

PR #5656, opened 2026-04-29
Allow removing dependencies without manifest

### LINKED ISSUE #5654: More helpful error message when removing a dependency and `manifest.toml` is missing
```sh
error: File IO failure

    /app/manifest.toml

```

For example if you do this:
```sh
$ gleam new app
$ cd app
$ gleam remove gleam_stdlib
error: File IO failure

    /app/manifest.toml

```

arjunrawal1:

This updates `gleam remove` so that if `manifest.toml` is missing it still removes the dependency from `gleam.toml` and exits successfully instead of trying to run manifest cleanup and reporting a file IO error.

Closes #5654

Testing:
- Added `remove_dependency_without_manifest` regression coverage.
- `git diff --check`
- Could not run `cargo test -p gleam-cli remove::tests::remove_dependency_without_manifest` locally because this machine does not have `cargo`/`rustc` installed.

AI assistance disclosure: I used OpenAI Codex GPT-5.5 to inspect the issue, implement the scoped change, and prepare this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,10 @@
 
 ### Build tool
 
+- The `gleam remove` command now updates `gleam.toml` when `manifest.toml` is
+  missing rather than reporting a file IO error.
+  ([Arjun Rawal](https://github.com/arjunrawal1))
+
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
   flag is passed, no progress information is printed.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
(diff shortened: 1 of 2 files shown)


ankddev:
Hello @arjunrawal1. Thank you for your work, but good first issues are not for LLMs, they are for humans to learn about compiler's internals and processes. There's totally no sense in doing them with LLMs, as you don't learn anything. Because of this, if you don't want to do it yourself, I think that it's better to leave it for someone else.

lpil:
Yup that's right, please do not use AI for issues intended for training humans. Having a path for people to learn these skills is important to the health of the project.

If you would like to do this yourself and learn how Gleam works then we've love to have your contribution!

arjunrawal1:
Closing this — I understand and respect that good first issues are meant as a learning path for humans getting familiar with the project. Thanks for calling that out.

PR #5659, opened 2026-04-30
Make unknown config keys emit a warning

### LINKED ISSUE #5301: have `gleam check`, `gleam build` validate the gleam.toml schema
Currently the inclusion of a spurious field in gleam.toml goes silently unremarked upon.

This caused me some grief just now as I didn't spot that a local dependency list should appear under `[dependencies]` not above it.

The proposal would be for `gleam build` to report the presence of unrecognized/unsupported fields in gleam.toml before it tries to build anything, i.e., for `gleam build` to validate the gleam.toml schema. (As versioned by the Gleam version.)

ankddev:
* Closes #5301
***

### DIFF
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -5,7 +5,7 @@ use crate::io::ordered_map;
 use crate::manifest::Manifest;
 use crate::requirement::Requirement;
 use crate::version::COMPILER_VERSION;
-use crate::{Error, Result};
+use crate::{Error, Result, Warning};
 use camino::{Utf8Path, Utf8PathBuf};
 use ecow::EcoString;
 use globset::{Glob, GlobSetBuilder};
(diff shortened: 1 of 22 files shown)


compiler-core/src/config.rs
```diff
@@ -319,6 +328,71 @@ impl PackageConfig {
         Ok(())
     }
 
+    /// Return list of warnings for unknown keys in config
+    pub fn generate_unknown_key_warnings(&self) -> Vec<Warning> {
```
ankddev, line 330: tbh I don't like this function, it looks a bit weird 😅

compiler-core/src/config.rs
```diff
@@ -319,6 +328,71 @@ impl PackageConfig {
...
+                        unknown
+                            .into_iter()
+                            .map(|(key, _)| format!("repository.{key}"))
+                    }),
+            )
+            .chain(
+                self.links
+                    .iter()
+                    .flat_map(|link| &link.unknown)
+                    .map(|(key, _)| format!("links.{key}")),
```
ankddev, line 367: I don't think that this format is good, but I have no ideas how to make it better

compiler-core/src/snapshots/gleam_core__config__unknown_keys_warning.snap
```diff
@@ -0,0 +1,68 @@
...
+warning: Unknown config key
+
+Config key 'repository.wibble' isn't allowed.
+
+Hint: See the `gleam.toml` schema at
+https://gleam.run/writing-gleam/gleam-toml.
+
+warning: Unknown config key
+
+Config key 'links.wibble' isn't allowed.
```
ankddev, line 44: As I stated above, I don't think that this is good style

ankddev (commented):
I didn't added this warning for invalid deps, since iirc it was intentionally made error (with `deny_unknown_fields`). Note that ignoring of warnings in many places is intentional, since in most cases warnings are already emitted while dependency resolution. And few notes inline

lpil (commented):
I don't like how the configuration data structure's definition now contains lots of invalid data that it should not have. What are some other approaches to this?

ankddev:
@lpil Neither do I, tbh. Approach I have used here is used in Pyrefly. I have few alternative ideas:

0. Rename it to `_unknown`: this still looks ugly, but denotes that it is internal and not data-related
1. Somewhere I've seen approach with `RawConfig` with this field, which is then used to genrate diagnsotic and to make real `Config`
2. Maybe some custom serde deserialiser?

What do you think about these?

lpil:
Would the raw-config approach mean having 2 copies of the data structure? Is there a way to avoid that?

I've been looking at serde and I couldn't see a way to configure this behaviour at runtime.

ankddev:
> Would the raw-config approach mean having 2 copies of the data structure? Is there a way to avoid that?

Yeah, there will be one raw struct with these invalid values, which will then be converted to another structure, dropping these fields.

> I've been looking at serde and I couldn't see a way to configure this behaviour at runtime.

Yeah, there seems  to be no way to do this better ATM.

ankddev:
Closing as we didn't come up with good solution at the moment.

PR #5660, opened 2026-04-30
fix detection of integer segments over the safe js limit

### LINKED ISSUE #5674: Warning about ints > 52 bits triggers incorrectly on JS target
In Gleam 1.16.0 the new "Truncated bit array segment" warning on the JavaScript target triggers in at least three cases it shouldn't:

1. When the function has an @external(javascript, ...) defined.
2. When pattern matching to a 64-bit float type.
3. When pattern matching to the `bytes` type and more than 52 bytes are being matched.

The following code illustrates the three cases:

```gleam
import gleam/bit_array
import gleam/list

pub fn main() -> Nil {
  echo wibble(<<1, 2, 3, 4, 5, 6, 7, 8>>)

  let assert <<f:big-float-size(64)>> = <<1, 2, 3, 4, 5, 6, 7, 8>>
  echo f

  let assert <<b:bytes-64>> = <<"0">> |> list.repeat(64) |> bit_array.concat
  echo bit_array.inspect(b)

  Nil
}

@external(javascript, "./ffi.mjs", "wibble")
fn wibble(bits: BitArray) -> Int {
  let assert <<x:64>> = bits
  x
}
```

Expected results is no compiler warnings from the above code when targeting JavaScript.

Thanks!

giacomocavalieri:
This PR fixes #5674 

This PR fixes two bugs with safe integer segment detection on the JS target:
- One where the compiler would incorrectly warn on byte segments (that Hayleigh reported on discord)
- And another one I found by chance where the compiler wouldn't take the unit into account when computing the safe integer limit

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -218,6 +218,14 @@
   patterns and guards of a case arm.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the compiler would not warn for integers over the safe
+  JavaScript limit in `BitArray` segments with a unit option.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
+- Fixed a bug where the compiler would incorrectly warn for integers over the
+  safe JavaScript limit in `BitArray` byte segments that aren't integers.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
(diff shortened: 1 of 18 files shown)


lpil (commented):
Use the term "int" rather than "integer" for consistency please 🙏

lpil (approved):
Thanks!!!

PR #5675, opened 2026-05-03
fix(javascript): escape reserved property names in TypeScript declarations

### LINKED ISSUE #5613: Invalid TypeScript Declarations for property named `constructor` and other javascript keywords
Given this gleam code:
```gleam
pub type Box {
  Box(constructor: Int)
}
```

And this `gleam.toml`:
```toml
name = "app"
version = "1.0.0"

target = "javascript"

[dependencies]
gleam_stdlib = ">= 0.44.0 and < 2.0.0"

[dev_dependencies]

[javascript]
```

The generated typescript declarations are invalid with the following error:
Image

Similarly when you use other property names such as `prototype` or `then`, the generated typescript declarations are valid, however do not match the generated JavaScript classes.

For example:
```gleam
pub type Box {
  Box(prototype: Int)
}
```
```javascript
export class Box extends $CustomType {
  constructor(prototype) {
    super();
    this.prototype$ = prototype;
  }
}
```
```typescript
export class Box extends _.CustomType {
  /** @deprecated */
  constructor(prototype: number);
  /** @deprecated */
  prototype: number;
}
```

Maybe this is not important, because they are deprecated, but wanted to bring it up anyways.

This is on `gleam 1.15.4`

Metbcy:
Closes #5613.

### Summary

The generated TypeScript declarations for record fields were using `maybe_escape_identifier`, which does not escape JavaScript reserved property names like `constructor`, `prototype`, `__proto__`, and `then`. The corresponding JavaScript class generation uses `maybe_escape_property`, which *does* escape them by appending `$`.

Result: the `.d.ts` advertised fields like `prototype: number` while the runtime actually exposes them as `prototype$`, and produced outright invalid declarations for `constructor`.

### Fix

Switch the field declaration in `compiler-core/src/javascript/typescript.rs` to `maybe_escape_property` so the `.d.ts` output mirrors the actual JS class shape (e.g. `constructor$`, `prototype$`, `then$`).

The constructor parameter list and `Box$Box(...)` factory function still use `maybe_escape_identifier`, matching the unescaped parameter names emitted by the JS code generator.

### Test plan

- New regression test `typescript_record_with_javascript_keyword_labels` in `compiler-core/src/javascript/tests/custom_types.rs` covering `constructor`, `prototype`, `then`, and a control field.
- New insta snapshot showing the corrected `.d.ts` output.
- `cargo test -p gleam-core` — 3816 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,13 @@
 
 ### Bug fixes
 
+- Fixed a bug where the generated TypeScript declarations for record fields
+  using JavaScript reserved property names (`constructor`, `prototype`,
+  `__proto__`, `then`) did not match the names used in the generated JavaScript
+  classes, leading to incorrect TypeScript declarations and misleading
+  autocompletion.
+  ([Metbcy](https://github.com/Metbcy))
+
 - Fixed a bug where the build tool would check for new major versions of a local
   or git dependency on Hex when running `gleam update`.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
--- a/compiler-core/src/javascript/tests/custom_types.rs
+++ b/compiler-core/src/javascript/tests/custom_types.rs
@@ -204,6 +204,18 @@ pub fn go() {
     );
 }
 
+// https://github.com/gleam-lang/gleam/issues/5613
+#[test]
+fn typescript_record_with_javascript_keyword_labels() {
+    assert_ts_def!(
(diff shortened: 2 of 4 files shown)


lpil:
Thank you but we do not accept AI generated contributions for good-first-issue tasks. Without these we have no way for humans to learn the codebase, which would be damaging to the project.

PR #5684, opened 2026-05-04
Fix code action overlap

giacomocavalieri:
A language server code action normally triggers when the cursor selection falls within a region of interest. I noticed there's a few code action that do something slightly different (and a bit annoying): they trigger when the cursor selection contains a region of interest. In practice this means that for those code actions, if I have a selection that covers a big chunk of code I'll see a rather confusing list pop up. For example:
 
Screenshot 2026-05-04 alle 20 49 35

This makes it impossible to know which action we should be triggering, and what portion of code they're going to update. I've updated those actions so that they work the same way as all the other ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -201,6 +201,10 @@
   decoder" code action even when not explicitly hovering a custom type.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the language server would suggest the "generate json
+  encoder" code action even when not explicitly hovering a custom type.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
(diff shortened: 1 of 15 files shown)


giacomocavalieri:
Another example I ran into of how using `overlap` instead of `within` results in confusing suggestions most of the time:
Screenshot 2026-05-04 alle 15 27 59
Here I wanted to extract a variable but I got the unqualify, and wrap in anonymous function suggestions for stuff that was inside it

lpil (approved):
Thank you!! Gosh that's a lot of changelog entries 😁

giacomocavalieri:
I wasn't sure if I wanted to add just a single entry or one for each but having more makes my brain happier 😂

PR #5685, opened 2026-05-04
Fix bit arrays with `unit` and `bytes` options

### LINKED ISSUE #5208: Inconsistent behaviour with `bytes` and `unit` options
When using both the `bytes` and `unit` options when pattern matching on a bit array, the compiled code is inconsistent across targets, and even by ordering of options.

On the Erlang target, the `unit` option overrides the `bytes` option completely, allowing you to pattern match bit arrays that are not a whole number of bytes:
```gleam
let assert <<x:bytes-unit(2)-size(3)>> = <<1:6>>
// Runs successfully
```

On the JavaScript target, the same behaviour occurs, but only if the `unit` option comes first:
```gleam
let assert <<x:unit(2)-bytes-size(3)>> = <<1:6>>
// Runs successfully
```
However, if the `bytes` option comes first, the `unit` option is ignored:
```gleam
let assert <<x:bytes-unit(2)-size(3)>> = <<1, 2, 3>>
// Runs successfully
```

Which behaviour do we want when combining these options? My expectation would be that the the unit would be multiplied by the implicit unit of 1 byte/8 bits, meaning `x:bytes-unit(2)-size(3)` would match 6 bytes.

GearsDatapacks:
Fixes #5208

Based on some testing, the correct Erlang behaviour is to have `unit` override `bytes`, so that's what I've made it do.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,3 +186,7 @@
   build directory was manually deleted. The compiler now automatically rebuilds
   the project when this configuration changes.
   ([daniellionel01](https://github.com/daniellionel01))
+
+- Fixed a bug where using the `bytes` and `unit` options together on a bit array
+  segment could generate incorrect code on the JavaScript target.
+  ([Surya Rose](https://github.com/GearsDatapacks))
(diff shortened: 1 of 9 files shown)


giacomocavalieri (approved):

lpil (approved):
Thank you!!!!

PR #5687, opened 2026-05-05
fix suggest extract function code action

giacomocavalieri:
I noticed the language server would suggest the extract function even when selecting multiple branches of a case expression, or pattern/guards of a branch:
Screenshot 2026-05-04 alle 17 24 20

This PR fixes it!


lpil (approved):
Thank you!!!

PR #5688, opened 2026-05-05
#2234 type alias and internal leak

lupodevelop:
## Checklist

Hopefully this will close the issue: #2234.

_EDIT: Rewrote the summary to give it some soul. Hopefully it makes sense now._

## Summary

Currently, `@internal` types are hidden from documentation but can still leak into public APIs without any compiler feedback. This PR aims to enforce that boundary and make such leaks visible.

* Emit warnings when a public API exposes an internal type.
* Prevent `gleam publish` if these warnings are present.
* Provide an escape hatch via public aliases:

As said, this is not a full finished proposal, but a concrete starting point.

To support the alias escape hatch, I introduced `Type::Alias` into the type system so that alias information is preserved instead of being immediately expanded.

However, as @lpil pointed out, there’s an issue in the current implementation, likely due to a misunderstanding on my part. It’s nothing that can’t be resolved, but it does affect error reporting: in some cases, the actual structure of types (e.g. tuples) is hidden, which makes debugging harder and goes against Gleam’s focus on clarity.

## Open Question

To improve diagnostics, I’m considering showing both forms in error messages when they differ. For example:

```
Expected: Pair(a) (i.e. #(a, a))
Found: #(a, b)
```

Does this feel like the right direction?

I wanted to check before diving deeper into reworking the diagnostics.

## What’s Implemented

* Internal type leak detection (compiler + LSP warnings)
* `gleam publish` guard
* Initial plumbing for alias tracking in the type system

Given this, I’ll wait for feedback to validate whether the proposed approach (aside from the issue mentioned above) is useful for addressing the problem. I’m of course happy to defer to your judgment on the direction.
Also, apologies for the previous message on the PR, my attempt to over-explain likely made things more complicated than necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,39 @@
   raised in.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The compiler now warns when a public function, constructor, or constant
+  exposes an `@internal` type. For example:
+
+  ```gleam
+  @internal
+  pub type Internal { ... }
+
+  pub fn process(value: Internal) -> Internal {
+    todo
+  }
+  ```
+
+  Will produce:
+
+  ```txt
+  warning: Internal type used in public interface
+    ┌─ /src/wrn.gleam:5:1
+    │
+  5 │ pub fn process(value: Internal) -> Internal {
+    │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+  The following type is internal, but is being used by this public export.
+
+      Internal
+
+  Internal types should not be used in a public facing function since
+  they are hidden from the package's documentation.
+  ```
+
+  Using `pub type Alias = Internal` is a safe escape hatch: the alias is
+  public, so using it in a public signature does not trigger the warning.
+  ([Daniele Scaratti](https://github.com/lupodevelop))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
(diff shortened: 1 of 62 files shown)


lpil:
Hello! Please do not generate AI summaries of PRs, they are very verbose and they are not useful. Instead please write them yourself so we can understand your thoughts, and so less time is spent reading.

The generated text says that this PR implements the agreed design, but no design has been agreed upon. Looking at the tests the behaviour of the alias system is not what we want- it hides the information about the underlying types in error messages. For example, the test for `alias Pair(a) = #(a, a); Pair(a) == #(a, b)` hides that `Pair` is a tuple, making debugging very difficult.
Clarity is of the utmost importance in Gleam, so these degraded error messages are not possible.

There are also other very impactful UX factors that we would need to consider when making such a large change to the language. This branch is useful for exploring this new language expansion, but very large amounts of design and test work needs to be done on the path to it being accepted.

lupodevelop:
Hi @lpil, thanks for the read.

I wrote the summary myself, so English isn't my first language, so I
tend toward the dry, structured tone you're seeing. I'll rewrite it
tighter. It's funny that after 2 days of work writing it, it was flagged like this 😄 .

On the substantive point: I misread the "no degraded error messages"
constraint as "preserve the alias name". You're saying the opposite.
The underlying structure must stay visible at the diagnostic site, or
the user can't see *why* the types don't match.

Direction I'd take, before re-implementing: in mismatch errors, show
both forms when they differ, e.g.

    Expected: Pair(a) (i.e. #(a, a))
    Found:    #(a, b)

Is that the shape you want? I'd rather check than rebuild blindly.

_I'll write you the Pull Request in the next few hours, without formalities_

lupodevelop:
One more thing: I don’t expect this PR to be accepted or closed right away... I’ve been digging into this for months and I'm aware that I might have missed some key context or misread the architectural vision. I just wanted to put forward a concrete attempt at solving the alias issue that's been open since 2023.

Also, for the record: 

I tend to oscillate between a 'mechanical' cold tone and a 'cryptic' one. It’s just how I navigate English. You should actually consider yourselves lucky: in my native Italian, I am so wordy and chaotic that reading my technical summaries feels like a psychedelic trip you’d need a long vacation to recover from. 

Consider my current 'robotic' coldness a courtesy to your mental health! 😄

I'm ready to dismantle and rebuild whatever is needed based on your feedback.

PR #5690, opened 2026-05-07
fix: escape double-quotes in beam compiler path formatting

maennchen:
escape_path only escaped backslashes, not double-quote characters. Since paths are embedded in an Erlang string literal sent to the compile escript, a filename containing " could break out of the string delimiter and produce a malformed module list.


lpil (approved):
Thank you!!!

Issue #2708, opened 2024-03-14
Error sending request for url (https://repo.hex.pm/packages/gleeunit)


corani:
This is most likely a local issue, but I haven't been able to get things to work due to some networking issue:

```bash
$ gleam --version
gleam 1.0.0
$ erl -version
Erlang (SMP,ASYNC_THREADS) (BEAM) emulator version 14.2.3

$ gleam new vars
$ cd vars
$ gleam run
  Resolving versions
error: Dependency resolution failed

An error occurred while determining what dependency packages and
versions should be downloaded.
The error from the version resolver library was:

Unable to determine package versions: error sending request for url
(https://repo.hex.pm/packages/gleeunit): error trying to connect: dns
error: failed to lookup address information: Name does not resolve

$ curl https://repo.hex.pm/packages/gleeunit -o out
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1822  100  1822    0     0   3884      0 --:--:-- --:--:-- --:--:--  3876
```

So I can download the file from hex using `curl` just fine, but `gleam run` runs into a DNS error.

(note: both gleam and erlang are installed using ASDF on WSL2/Ubuntu)

lpil:
Thank you

corani:
Yes, consistently. `gleam run` will terminate with the same message in milliseconds each time.

Interestingly, even if I add `repo.hex.pm` to my `hosts` file it'll still fail with the same DNS error.

lpil:
I've just tried adjusting our HTTP client. Tomorrow or later would you be able to download the [nightly](https://github.com/gleam-lang/gleam/releases/tag/nightly) build and see if it works? Thank you

lpil:
Here's the log from another WSL user who got "too many redirects". [client_logs_nc.txt](https://github.com/gleam-lang/gleam/files/14614476/client_logs_nc.txt)

corani:
@lpil I have no clue what's going on. Today after rebooting it suddenly started working. 

I've tried both `1.0.0` and `nightly` a couple of times, and both are fetching from hex without issue now...

Sorry for wasting your time, feel free to close this.

lpil:
Oh dear, it gets even more mysterious... Ok let's close this for now and keep an eye open for any future problems like this. Thank you

andho:
I often run into this problem previously with elixir. And now with gleam as well.

```bash
$ gleam deps update
[...]

PR #5691, opened 2026-05-07
update reqwest from 0 -> 0.13

apsoras:
      https://github.com/gleam-lang/gleam/issues/2708
       https://github.com/gleam-lang/gleam/issues/2708
       No new behaviour
       No user-facing changes

Apologies for the cargo.lock churn, again not super familiar with working on very large rust projects but all I did was bump reqwest in `compiler-cli/Cargo.toml` from `reqwest = { version = "0", default-features = false, features = ["rustls-tls"] }` to `reqwest = { version = "^0.13", default-features = false, features = ["rustls"] }` and run cargo build - the rest are dependencies for the new way reqwest has modernized/unified its tls/crypto stack!

Benefit is that reqwest now references the system trust store by default in front of bundled certs, so those working behind a proxy (with a self-signed cert) were previously failing to make the tls handshakes necessary to do `gleam <anything>`.  As mentioned in the issue, astral uv and tealdeer both have ways to work around this as well either with command line flags or configuration settings to use system certs, but reqwest default doesn't require that anymore thankfully.

The error that gleam bubbles up when the tls handshake/request fails maybe obfuscated this a bit (for example, mine showed this when running gleam build: `An error occurred while choosing the version of gleam_stdlib: error sending request for url (https://repo.hex.pm/packages/gleam_stdlib)` and I was getting a similar sort of error when trying to run the language server in vs code) as opposed to outright saying the client builder or tls handshake failed, etc, but hopefully no future problems so maybe not worth worrying about!

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -284,3 +284,8 @@
   fail to build when they depended on Gleam packages with transitive
   dependencies.
   ([Logan Bresnahan](https://github.com/LoganBresnahan))
+
+- Fixed a bug where cli would fail to complete https connections from behind a proxy
+  with self-signed certificates. The cli now defaults to using system trust stores
+  for trusted CAs, allowing use in proxied network environments.
+  ([apsoras][https://github.com/apsoras])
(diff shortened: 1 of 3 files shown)


lpil (commented):
Thank you!! Could you update the changelog to describe the bug fix please? This would be written descriptively for the benefit of the users 🙏

lpil (approved):
Thanks once again!!

PR #5692, opened 2026-05-07
Decrease JS verbosity when destructuring exhaustive patterns with `let`

### LINKED ISSUE #5679: Destructuring with `let` produces slightly verbose javascript
I triggered the "Pattern match on variable" code action, which caused the compiler to generate this line:

```gleam
let Model(tick:, app:, mounted:, state:, connected_clients:) = model
```

however, this compiles to the following JavaScript:

```js
let tick;
let app;
let mounted;
let state$1;
let connected_clients;
tick = model.tick;
app = model.app;
mounted = model.mounted;
state$1 = model.state;
connected_clients = model.connected_clients;
```

when instead using a case expression, the JavaScript matches more what I would expect:

```gleam
case model {
 Model(tick:, app:, mounted:, state:, connected_clients:) -> todo
}
```

compiles to

```js
let tick = model.tick;
let app = model.app;
let mounted = model.mounted;
let state$1 = model.state;
let connected_clients = model.connected_clients;
throw makeError(...)
```

in theory, this could use the Javascript object destructuring syntax as well, but at least those 2 could be equivalent I think?

~ 💜

gavinmorrow:
Fixes #5679.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,11 @@
 
   ([Daniele Scaratti](https://github.com/lupodevelop))
 
+- The code generated for destructuring exhaustive patterns is now less verbose
+  on the JavaScript target.
+
+  ([Gavin Morrow](https://github.com/gavinmorrow))
+
 ### Build tool
 
 - The `gleam dev` command now accepts the `--no-print-progress` flag. When this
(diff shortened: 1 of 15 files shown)


lpil (approved):
Wow, lovely!! Thank you!!

PR #5696, opened 2026-05-09
Deduplicate bitArraySliceToFloat calls in JS codegen

### LINKED ISSUE #4658: Avoid calling `bitArraySliceToFloat` in generated JS for float patterns
Today when using a `:float` bit array pattern the `bitArraySliceToFloat` function is called twice with the same arguments. This is wasted work!

Change the code generator to do it only once, saving the result to a variable to use twice.

kimjune01:
Closes #4658

The JS code generator was calling `bitArraySliceToFloat` twice with the same arguments when matching on float bit array patterns — once for the `Number.isFinite` check and again for the body binding. This caches the result in a variable during the check (`$ = bitArraySliceToFloat(...)`) so the body binding can reuse it.

### DIFF
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -432,6 +432,7 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
             let old_names = self.variables.scoped_variable_names.clone();
             let old_segments = self.variables.segment_values.clone();
             let old_segment_names = self.variables.scoped_segment_names.clone();
+            let old_cached_float = self.variables.cached_float_var.clone();
 
             let result = self.decision(fallback);
 
@@ -459,6 +460,7 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
             self.variables.scoped_variable_names = old_names;
             self.variables.segment_values = old_segments;
             self.variables.scoped_segment_names = old_segment_names;
+            self.variables.cached_float_var = old_cached_float;
             return result;
         }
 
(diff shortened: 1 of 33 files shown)


compiler-core/src/javascript/decision.rs
```diff
@@ -1821,6 +1859,22 @@ impl<'generator, 'module, 'a> Variables<'generator, 'module, 'a> {
...
             check_assignments.push(let_doc(variable_name, segment_value))
         }
+
+        // If this is a `SegmentIsFiniteFloat` check, declare a variable to
+        // cache the result of the float read. The variable is declared here
+        // (before the if-condition) but assigned inside the condition via an
+        // assignment expression `$N = bitArraySliceToFloat(...)`. This way
+        // the float is only computed after earlier checks (like size) pass,
+        // and the result can be reused in the body binding.
+        if let RuntimeCheck::BitArray {
```
jtdowney, line 1869: This pushes unconditionally but only pops on reads. I cloned your branch and added a test that shows this.

```rust
#[test]
fn case_match_float_discard_first() {
    assert_js!(
        r#"
pub fn go(x) {
  case x {
    <<_:float-32, b:float-32>> -> "float"
    _ -> "other"
  }
}
"#,
    );
}
```

The JS code produced is:

```js
import { bitArraySliceToFloat } from "../gleam.mjs";

export function go(x) {
  if (x.bitSize >= 32) {
    let $;
    if (
      Number.isFinite($ = bitArraySliceToFloat(x, 0, 32, true)) &&
      x.bitSize === 64
    ) {
      let $1;
      if (Number.isFinite($1 = bitArraySliceToFloat(x, 32, 64, true))) {
        let b = $;
        return "float";
      } else {
        return "other";
      }
    } else {
      return "other";
    }
  } else {
    return "other";
  }
}
```

As you can see `b` is assigned the value from `$` and not `$1`.

lpil:

ankddev:
@lpil Looking at their profile this is a LLM agent. I've seen quite a lot contributions by them in other projects.

PR #5703, opened 2026-05-11
Improve error for pattern match prefix and suffix

### LINKED ISSUE #5695: Add better error message when pattern matching on string prefix and suffix
When attempting to pattern match on both the suffix and prefix of a string, the error message could be friendlier.

For example, with
```gleam
case "my string!" {
  "my " <> object <> "!" -> todo
  start <> "!" -> todo
}
```

The current error is:
```gleam
error: Syntax error
  ┌─ /private/tmp/match_string_end/src/match_string_end.gleam:5:21
  │
5 │     "my " <> object <> "!" -> todo
  │                     ^^ I was not expecting this

Found `<>`, expected one of:
- `->`

```

Compare to the error if you don't try to match on the string's prefix:
```gleam
error: Syntax error
  ┌─ /private/tmp/match_string_end/src/match_string_end.gleam:6:5
  │
6 │     start <> "!" -> todo
  │     ^^^^^ This must be a string literal

We can't tell what size this prefix should be so we don't know
how to handle this pattern.

If you want to match one character consider using `pop_grapheme`
from the stdlib's `gleam/string` module.
```

Thanks to [@the_tank_man_](https://discord.com/channels/768594524158427167/768594524158427170/1502644925253030060) from the discord for suggesting this :]

Also, I can start working on this.

gavinmorrow:
Fixes #5695

An example of the new error:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:2:23
  │
2 │     "prefix" <> infix <> "suffix" -> infix
  │                       ^^^^^^^^^^^ This pattern is not allowed

A string pattern can only match on a literal string prefix.

Matching on a literal suffix is not possible, because `infix` would have an
unknown size.
```


gavinmorrow:
@lpil This is ready, except that I'm not sure if/where to add tests. Do error messages get tests?

lpil:
Could you put some in here please 🙏  https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/parse/tests.rs

gavinmorrow:
@lpil it's ready for review!

compiler-core/src/parse/snapshots/gleam_core__parse__tests__assign_infix_of_concat_pattern.snap
```diff
@@ -0,0 +1,21 @@
...
+  ┌─ /src/parse/error.gleam:2:17
+  │
+2 │     "prefix" <> infix <> "suffix" -> infix
+  │                 ^^^^^ This must be a string literal prefix
+
+We can't tell what size this infix should be so we don't know
+how to handle this pattern.
+
+If you want to match one character consider using `pop_grapheme`
+from the stdlib's `gleam/string` module.
```
giacomocavalieri: I think I'd slightly change the wording, and I wouldn't recommend `pop_grapheme`, doesn't seem all that useful and it's easy to write very slow code with that.

```suggestion
----- ERROR
error: Syntax error
  ┌─ /src/parse/error.gleam:2:17
  │
2 │     "prefix" <> wibble <> "suffix" -> infix
  │                        ^^^^^^^^^^^ This pattern is not allowed

A string pattern can only match on a literal string prefix.
Matching on a literal suffix is not possible, because `wibble` would have an unknown size.
```

gavinmorrow: That's much better wording, ty!

compiler-core/src/parse.rs
gavinmorrow, line 1: Is the repetition acceptable, or should I find a way to put it in a function?

compiler-core/src/parse/error.rs
```diff
@@ -555,6 +560,18 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
                 extra_labels: vec![],
             },
 
+            ParseErrorType::ConcatPatternVariableWithSuffix  { name}=> ParseErrorDetails {
```
lpil: Formatting looks off here

compiler-core/src/parse/error.rs
```diff
@@ -555,6 +560,18 @@ utf16_codepoint, utf32_codepoint, signed, unsigned, big, little, native, size, u
...
+                text: [
+                    "A string pattern can only match on a literal string prefix.",
+                    "",
+                    &format!("Matching on a literal suffix is not possible, because `{}` would have an unknown size.", name),
```
lpil: Wrap long lines please. The width of this is variable, so the `wrap` function or `wrap_format` macro would need to be used here.

lpil (commented):
Thank you!! Couple small notes inline

gavinmorrow:
@lpil thanks for the comments! I made some changes.

giacomocavalieri (approved):
Lovely! Would you mind rebasing on main?

gavinmorrow:
@giacomocavalieri I rebased, thank you!

compiler-core/src/parse.rs
lpil: It's fine :)

lpil (approved):
Very nice!! Thank you very much!!

PR #5706, opened 2026-05-12
fix bug with create module being too eager to create stuff

### LINKED ISSUE #5705: Create module code action should not trigger for importable modules
The "create module" code action triggers even for modules that exist and are already importable in the project, I think that is never the intended behaviour and the action should not pop up if the module is importable
Image

giacomocavalieri:
This PR closes #5705 

- [ ] The changelog has been updated for any user-facing changes: this action wasn't released yet so no changelog entry is needed


lpil (approved):
Thank you!!

PR #5707, opened 2026-05-12
reorder code actions

giacomocavalieri:
I've noticed in some cases the language server ordering of code actions is not ideal, for example:
missing add labels code action
In this case the quick fix to fill in the missing labels (what I actually wanted was third), with the two other options not being immediately useful.

I've reordered the code actions so that fixes for errors and warnings are always presented first, before "refactoring" code actions: this means that when my cursor is over an error or a warning the language server should now first present the quick fix, then the other less relevant refactorings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -151,6 +151,10 @@
 
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- The language server now presents quick fix code actions before refactoring
+  ones.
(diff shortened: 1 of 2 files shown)


lpil (commented):
Oh! That's interesting. I never thought about that!

Is there a way we could make this clearer? Maybe just comments explaining the order? Or maybe we should sort the vector based on the type of the code action?

ankddev:
I think that sorting would be nice!

lpil (approved):
Thank you!!!

PR #5713, opened 2026-05-13
Improve parser error messages

### LINKED ISSUE #5711: Confusing syntax error when writing bit array size
I was writing a constant bit array with a size option and a variable name in it, this is not allowed and results in a syntax error that is a bit confusing as it is telling me it's expecting a variable name, when that is the thing causing the error in the first place
Image

giacomocavalieri:
This PR fixes #5711 and other various parser error messages I found while fixing it

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -4293,9 +4293,15 @@ functions are declared separately from types.";
 
     // Expect a String else error
     fn expect_string(&mut self) -> Result<(u32, EcoString, u32), ParseError> {
-        match self.next_tok() {
-            Some((start, Token::String { value }, end)) => Ok((start, value, end)),
-            _ => self.next_tok_unexpected(vec!["a string".into()]),
+        match self.tok0.take() {
+            Some((start, Token::String { value }, end)) => {
+                self.advance();
+                Ok((start, value, end))
+            }
+            tok0 => {
+                self.tok0 = tok0;
+                self.next_tok_unexpected(vec!["a string".into()])
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


compiler-core/src/parse.rs
```diff
@@ -3823,7 +3824,10 @@ where
                                     }),
                                 }
                             }
-                            _ => self.next_tok_unexpected(vec!["positive integer".into()]),
+                            tok0 => {
+                                self.tok0 = tok0;
+                                self.next_tok_unexpected(vec!["A positive integer".into()])
```
lpil:
```suggestion
                                self.next_tok_unexpected(vec!["A positive int".into()])
```

compiler-core/src/parse.rs
```diff
@@ -3922,18 +3939,27 @@ where
                     inner: Box::new(inner),
                 })
             }
-            _ => self.next_tok_unexpected(vec!["A variable name or an integer".into()]),
+            tok0 => {
+                self.tok0 = tok0;
+                self.next_tok_unexpected(vec!["A variable name or an integer".into()])
```
lpil:
```suggestion
                self.next_tok_unexpected(vec!["A variable name or an int".into()])
```

compiler-core/src/parse.rs
```diff
@@ -3922,18 +3939,27 @@ where
...
+                self.advance();
+                Ok(Constant::Int {
+                    location: SrcSpan { start, end },
+                    value,
+                    int_value,
+                })
+            }
+            tok0 => {
+                self.tok0 = tok0;
+                self.next_tok_unexpected(vec!["An integer".into()])
```
lpil:
```suggestion
                self.next_tok_unexpected(vec!["An int".into()])
```

lpil (approved):
Thank you!! Few tiny thing inline

lpil (approved):
Thank you!!!!!

PR #5714, opened 2026-05-13
Allow pattern matching on discard in case branches

giacomocavalieri:
This PR allows to trigger the "match on value" code action on discard patterns as well!

- [X] The changes in this PR have been discussed beforehand with Louis


lpil (approved):
Thank you!!

PR #5717, opened 2026-05-14
Improve hex revert missing version error

### LINKED ISSUE #5709: Error message when reverting version that doesn't exist could be improved
When trying to revert a version that doesn't exist we get the following error message:
```sh
> gleam hex revert --package birdie --version 1.234234.2
error: Hex API failure

    Resource was not found
```
I think a much better error would look something like this:
```
> gleam hex revert --package birdie --version 1.234234.2

error: Hex API failure

There is no version 1.234234.2 of the package birdie.
```

fizyxbt:
Closes #5709.

This adds a specific error for `gleam hex revert` when Hex reports that the requested release does not exist, so the command can show the package and version directly instead of the generic Hex client message.

Tests:
- `cargo fmt --all -- --check`
- `cargo check -p gleam-cli`
- `cargo test -p gleam-core hex_release_not_found`
- `cargo test -p gleam-cli hex`

### DIFF
--- a/compiler-cli/src/hex.rs
+++ b/compiler-cli/src/hex.rs
@@ -111,7 +111,13 @@ pub fn revert(
     )
     .map_err(Error::hex)?;
     let response = runtime.block_on(http.send(request))?;
-    hexpm::api_revert_release_response(response).map_err(Error::hex)?;
+    match hexpm::api_revert_release_response(response) {
+        Ok(()) => (),
+        Err(hexpm::ApiError::NotFound) => {
+            return Err(Error::HexReleaseNotFound { package, version });
+        }
+        Err(error) => return Err(Error::hex(error)),
+    }
 
     // Done!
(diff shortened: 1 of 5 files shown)


compiler-cli/src/hex.rs
```diff
@@ -111,7 +111,13 @@ pub fn revert(
     )
     .map_err(Error::hex)?;
     let response = runtime.block_on(http.send(request))?;
-    hexpm::api_revert_release_response(response).map_err(Error::hex)?;
+    match hexpm::api_revert_release_response(response) {
```
giacomocavalieri, line 114: I wonder if the api actually has more information in the response body as to tell us if it's the package or the version that doesn't exist. Just converting would produce confusing error messages if it's the package the one that doesn't exist

lpil: I agree! Can the hexpm library tell what the precise problem is from the response body or the headers?

PR #5719, opened 2026-05-14
improve doc tables on small screens

giacomocavalieri:
I noticed tables look a bit cramped on mobile, often with columns that are just a single character wide:
Screenshot 2026-05-14 alle 15 50 14

I've changed it so that, like code blocks, tables can be scrolled horizontally rather than getting too narrow. This is what it looks like:

And there's no changes if the table has enough space:
Screenshot 2026-05-14 alle 15 49 17

Hope this is good!


lpil (approved):
Thank you!!

PR #5720, opened 2026-05-14
Escripts

lpil:
A new `gleam export escript` command, for making single file Erlang programs.


CHANGELOG.md
```diff
@@ -130,6 +130,11 @@
   improves suggestions to push a tag, if it doesn't exists.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- The `gleam export escript` command has been added for the creation
+  [escripts](https://www.erlang.org/doc/apps/erts/escript_cmd.html), BEAM
```
ankddev:
```suggestion
- The `gleam export escript` command has been added for the creation of
  [escripts](https://www.erlang.org/doc/apps/erts/escript_cmd.html), BEAM
```

ankddev (commented):
Found typo in changelog. And small niche because you have touched few errors: in some places main function is referred as `` `main` `` and in others as `main`, is that intentional?

compiler-core/src/error.rs
```diff
@@ -1186,11 +1192,13 @@ target, so it cannot be run.",
...
                 text: wrap_format!(
-                    "`{module}:main` should have an arity of 0 to be run but its arity is {arity}."
+                    "`{module}.main` should take no arguments, but it takes {arity}."
                 ),
                 level: Level::Error,
                 location: None,
-                hint: Some("Change the function signature of main to `pub fn main() {}`.".into()),
+                hint: Some(
+                    "Change the function signature of main to `pub fn main() -> Nil`.".into(),
+                ),
```
giacomocavalieri, line 1201: Love this change

compiler-core/src/error.rs
```diff
@@ -1598,6 +1606,41 @@ This was error from the tar library:
...
             }
 
+            Error::ZipAdd { path, error: err } => {
+                let text = wrap_format!(
+                    "There was a problem when attempting to add the file {path} \
+to a zip archive.
+
+This was error from the zip library:
+
+    {err}"
```
giacomocavalieri, line 1616: Why would someone care about if the compiler is using a "zip library". I would just say:
```suggestion
to a zip archive:

    {err}"
```

compiler-core/src/error.rs
```diff
@@ -1598,6 +1606,41 @@ This was error from the tar library:
...
+                }]
+            }
+
+            Error::ZipFinish(detail) => {
+                let text = format!(
+                    "There was a problem when creating a zip archive.
+
+This was error from the zip library:
+
+    {detail}"
```
giacomocavalieri, line 1633: Same as above

compiler-cli/src/export.rs
```diff
@@ -14,8 +17,100 @@ static ENTRYPOINT_TEMPLATE_POWERSHELL: &str =
...
+    let package_name = &built.root_package.config.name;
+
+    // The main function must exist for the escript to call. This will return an
+    // error if it could not be found.
+    let _: ModuleFunction = built.get_main_function(package_name, target)?;
+
+    // Create the zip archive for the code
+    let mut zip = ZipArchive::new(Cursor::new(Vec::new()));
+
+    for entry in fs::read_dir(&build)?.filter_map(Result::ok) {
```
giacomocavalieri, line 52: I don't understand what this `.filter_map(Result::ok)` is for

compiler-cli/src/export.rs
```diff
@@ -14,8 +17,100 @@ static ENTRYPOINT_TEMPLATE_POWERSHELL: &str =
...
+        let ebin = entry.path().join("ebin");
+
+        // We want the ebin code directories for each package
+        if !ebin.is_dir() {
+            continue;
+        }
+
+        for entry in fs::read_dir(&ebin)?.filter_map(Result::ok) {
```
giacomocavalieri, line 60: Same here!

giacomocavalieri (commented):
Just a couple of comments!

compiler-cli/src/export.rs
lpil: It skips over errors from files that cannot be iterated from a directory that could be. Could change it in future but this is how all the directory iteration is written today, so I maintained the existing style.

giacomocavalieri: mhm I see

compiler-cli/Cargo.toml
```diff
@@ -33,6 +33,8 @@ toml_edit = "0"
 fslock = "0"
 # Provides a way to determine if two files are the same using filesystem node ids
 same-file = "1"
+# Creation of zip files
+zip = { version = "8", features = ["deflate"] }
```
jtdowney: If you set `default-features = false` here, you can cut down ~30 extra crates being pulled in, including `bzip2` which means it is no longer needed in the licenses. I'd expect just the `deflate` feature to be enough for escripts.

compiler-core/src/paths.rs
```diff
@@ -124,6 +124,34 @@ impl ProjectPaths {
             .join("ebin")
     }
 
+    pub fn build_package_dot_app(
```
jtdowney: I wasn't able to find any reference to this method in the code or tests. Was this created for some future expansion? Or did I miss it?

compiler-core/src/paths.rs
```diff
@@ -124,6 +124,34 @@ impl ProjectPaths {
...
+        mode: Mode,
+        target: Target,
+        package_name: &str,
+    ) -> Utf8PathBuf {
+        self.build_package_ebin(mode, target, package_name)
+            .join(package_name)
+            .with_extension("app")
+    }
+
+    pub fn build_package_beam(
```
jtdowney: Same issue as `build_package_dot_app`

compiler-cli/Cargo.toml
lpil: John, you are worth every penny of your Gleam core team salary

compiler-core/src/paths.rs
lpil: Good catch, it was lost in an earlier refactoring. Thank you

compiler-core/src/error.rs
lpil: That's what they all say. It was to originally to make clear the boundary into some other system, which I think was taken from Elm. If we change this we should change the others too, which I won't do here.

PR #5721, opened 2026-05-16
remove: don't fail when manifest.toml is missing

### LINKED ISSUE #5654: More helpful error message when removing a dependency and `manifest.toml` is missing

c-tonneslan:
Ran into this trying to reproduce #5654 locally. In a fresh project (or after temporarily `rm manifest.toml` to untangle dependency versions), `gleam remove` bails with:

```
error: File IO failure

    /path/to/manifest.toml

```

The remove command updates `gleam.toml` and then calls `dependencies::cleanup` to prune the manifest, but cleanup reads `manifest.toml` unconditionally. If there's no manifest there's also nothing to prune, so just skip the cleanup. Next build will regenerate it.

Repro before the fix:

```
$ gleam new myproj && cd myproj
$ gleam add gleam_stdlib
$ rm manifest.toml
$ gleam remove gleam_stdlib
error: File IO failure
...
```

After: `gleam.toml` gets updated and the command exits cleanly.

Closes #5654.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -190,6 +190,11 @@
 
 ### Bug fixes
 
+- Fixed a bug where `gleam remove` would fail with a confusing File IO error
+  if `manifest.toml` didn't exist yet (e.g. in a freshly-created project or
+  after the manifest had been deleted).
+  ([Charlie Tonneslan](https://github.com/c-tonneslan))
+
 - Fixed a bug where the build tool would check for new major versions of a local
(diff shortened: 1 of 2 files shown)


lpil (approved):
Lovely work! Thank you for the detailed comment there, really nice!!

PR #5722, opened 2026-05-17
fix: use property escaping in TypeScript declarations for reserved field names

### LINKED ISSUE #5613: Invalid TypeScript Declarations for property named `constructor` and other javascript keywords

arkh-node:
Fixes #5613

## Problem

Fields named `constructor`, `prototype`, `then`, or `__proto__` are renamed with a `$` suffix in the JavaScript output to avoid collisions with built-in JS properties. However, the TypeScript declaration generator was using `maybe_escape_identifier` instead of `maybe_escape_property` for field names, so the `.d.ts` output did not reflect this rename.

This caused two problems:

1. A field named `constructor` produced **invalid TypeScript** — the class ended up with a duplicate `constructor` overload which TypeScript rejects.
2. Fields named `prototype` or `then` produced **valid but wrong TypeScript** — the declared property name didn't match the actual JS property name, so e.g. `box.prototype` was suggested by autocomplete when the real property is `box.prototype$`.

## Fix

Make `maybe_escape_property` pub(super) and use it in the three places in `typescript.rs` that emit field names:
- constructor parameter names in the class definition
- class body field declarations  
- parameters in standalone constructor functions

## Tests

Added three snapshot tests for `constructor`, `then`, and `prototype` field names. The generated `.d.ts` now correctly uses `constructor$`, `then$`, and `prototype$` to match the JavaScript output.

### DIFF
--- a/compiler-core/src/javascript.rs
+++ b/compiler-core/src/javascript.rs
@@ -1241,7 +1241,7 @@ fn maybe_escape_identifier(word: &str) -> EcoString {
     }
 }
 
-fn maybe_escape_property(label: &str) -> EcoString {
+pub(super) fn maybe_escape_property(label: &str) -> EcoString {
     if is_usable_js_property(label) {
         EcoString::from(label)
     } else {
--- a/compiler-core/src/javascript/tests/custom_types.rs
+++ b/compiler-core/src/javascript/tests/custom_types.rs
@@ -634,6 +634,39 @@ pub fn main() {
     );
 }
 #[test]
+fn record_with_field_named_constructor_typescript() {
+    assert_ts_def!(
+        r#"
+pub type Box {
+  Box(constructor: Int)
+}
+"#,
+    );
+}
+
+#[test]
+fn record_with_field_named_then_typescript() {
+    assert_ts_def!(
+        r#"
+pub type Box {
+  Box(then: Int)
+}
+"#,
+    );
+}
+
+#[test]
+fn record_with_field_named_prototype_typescript() {
+    assert_ts_def!(
+        r#"
+pub type Box {
+  Box(prototype: Int)
+}
+"#,
+    );
+}
+
+#[test]
 fn record_access_in_guard_with_reserved_field_name() {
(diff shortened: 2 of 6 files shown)


lpil (commented):
Hello! Are you a person or a robot?

arkh-node:
Hi! Definitely a person — I'm Alexey, 56 years old, based in Mexico. No CS degree, but programming has been a lifelong fascination. With AI as a collaborator I finally get to do this properly: contributing to real projects, relearning math from scratch, building agents. Gleam caught my attention because of the type system and the BEAM — exactly the kind of thing I want to understand deeply. Happy to adjust anything in the PR if needed.

ankddev:
> With AI as a collaborator I finally get to do this properly: contributing to real projects, relearning math from scratch, building agents.

Hey! You shouldn't use LLMs to fix "good first issues", because these are for people to learn about codebase and processes. When LLM does most of the work, you don't learn anything.

arkh-node:
Closing — not pursuing this further.

PR #5723, opened 2026-05-17
Add helpful error message for prefix+suffix string pattern matching

### LINKED ISSUE #5695: Add better error message when pattern matching on string prefix and suffix

arkh-node:
Fixes #5695

## What

When a user writes a string pattern that attempts to match both a prefix and a suffix:

```gleam
case "my string!" {
  "my " <> object <> "!" -> todo
}
```

Previously this produced a confusing generic error:

```
error: Syntax error
  ┌─ src/main.gleam:3:21
  │
3 │   "my " <> object <> "!" -> todo
  │                   ^^ I was not expecting this

Found `<>`, expected one of:
- `->`

```

Now it produces a clear, actionable error:

```
error: Syntax error
  ┌─ src/main.gleam:3:21
  │
3 │   "my " <> object <> "!" -> todo
  │                   ^^ This is not allowed in a string pattern

String patterns can only match on a prefix, not both a prefix and a suffix.

If you want to match on a suffix consider using `string.ends_with`
from the stdlib's `gleam/string` module.
```

## How

Added a `ConcatPatternWithSuffix` parse error variant and a check in the parser: after successfully parsing `"prefix" <> var`, if the next token is another `<>`, emit the new specific error. The check covers both the plain form and the `as`-assignment form (`"prefix" as x <> var <> "suffix"`).

Two snapshot tests added.

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -1364,6 +1364,18 @@ where
                             Some((_, Token::Concatenate, _)) => {
                                 self.advance();
                                 let (r_start, right, r_end) = self.expect_assign_name()?;
+                                // "prefix" as x <> var <> "suffix" is not supported
+                                if let Some((suffix_start, Token::Concatenate, suffix_end)) =
+                                    self.tok0
+                                {
+                                    return Err(ParseError {
+                                        error: ParseErrorType::ConcatPatternWithSuffix,
+                                        location: SrcSpan {
+                                            start: suffix_start,
+                                            end: suffix_end,
+                                        },
+                                    });
+                                }
                                 Pattern::StringPrefix {
                                     location: SrcSpan { start, end: r_end },
                                     left_location: SrcSpan {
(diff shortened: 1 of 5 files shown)


giacomocavalieri:
This is a duplicate of #5703

arkh-node:
Closing as duplicate of #5703 — thanks for the pointer @giacomocavalieri!

PR #5724, opened 2026-05-18
Improve Hex API key error message

### LINKED ISSUE #5710: Error message for inserting the wrong local Hex API key could be improved
Right now when typing the wrong Hex API key (for any command that requires it) shows the following error message:
```
error: Failed to decrypt local Hex API key

Unable to decrypt the local Hex API key with the given password.

    unable to decrypt message: Decryption failed
```
It's telling me the same thing three times: "failed to decrypt local hex api key", "unable to decrypt the local hex api key", "unable to decrypt message: decryption failed". It's also telling me about some "encryption library" I don't really care about as a user!
I think a better error message should be more straight to the point:
```
error: Failed to decrypt local Hex API key

Unable to decrypt the local Hex API key with the given password.
```
Or maybe add an hint if you feel like people would need further direction:
```
error: Failed to decrypt local Hex API key

Unable to decrypt the local Hex API key with the given password.
Hint: try again and make sure to typed the correct local password
```
Or something along those lines

MoritzBoehme:
This PR fixes #5710

Regarding the tests I deliberately did not add one as I could not find other such tests, and they are in my experience very brittle and do not add a lot of value.

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -1918,8 +1918,8 @@ Did you make a typo? Please try again."
                     title: "Failed to decrypt local Hex API key".into(),
                     text,
                     hint: Some(
-                        "If you have forgotten your local password you can create
-new credentials with `gleam hex authenticate`."
+                        "If you have forgotten your local password you can create a new one
+with `gleam hex authenticate`."
                             .into(),
                     ),
                     level: Level::Error,
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -419,7 +419,7 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
     FailedToEncryptLocalHexApiKey { detail: String },
 
     #[error("Failed to decrypt local Hex API key")]
-    FailedToDecryptLocalHexApiKey { detail: String },
+    FailedToDecryptLocalHexApiKey,
 
     #[error("Cannot add a package with the same name as a dependency")]
     CannotAddSelfAsDependency { name: EcoString },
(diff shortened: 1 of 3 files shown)


compiler-core/src/error.rs
giacomocavalieri: Rather than putting the hint in the text itself, could you add it as the `hint` field here?
```suggestion
"
);
vec![Diagnostic {
title: "Failed to decrypt local Hex API key".into(),
text,
hint: Some("If you have forgotten your local password you can create
new credentials with `gleam hex authenticate`.".into()),
```

giacomocavalieri:
Amazing, thank you! I just left a little comment

compiler-core/src/error.rs
MoritzBoehme: Ah, I totally glossed over this. Will fix.

MoritzBoehme:
Should be good now :)

giacomocavalieri (approved):
Thank you!

compiler-core/src/error.rs
```diff
@@ -1767,17 +1767,15 @@ The error from the encryption library was:
                 }]
             }
 
-            Error::FailedToDecryptLocalHexApiKey { detail } => {
+            Error::FailedToDecryptLocalHexApiKey { detail: _detail } => {
```
lpil: This `detail` value is no longer used, could you remove it from the definition of `Error::FailedToDecryptLocalHexApiKey` please 🙏

lpil (commented):
Thank you!! I've left one small comment inline. Could you also run `cargo fmt` please 🙏

MoritzBoehme:
Sorry, as always happens, I have/had a lot to do at $dayjob. But I hope to get to finish this tomorrow and hopefully look at other issues this weekend.

MoritzBoehme:
Ran `cargo fmt` and removed the detail.
Thanks to both of you for your patience :)

~~Edit: I just noticed the merge conflict. I will rebase and fix this.~~

ankddev:
@MoritzBoehme We do not use merge commits in Gleam, so you need to rebase to get rid of that merge commit.

giacomocavalieri (approved):

lpil (approved):
Thank you!!

PR #5728, opened 2026-05-18
Improve error message for reverting old release

### LINKED ISSUE #5708: Bad error message when trying to revert an old release
When running `gleam hex revert`, if the release is too old (and can't be reverted) the error message we get back from Hex shows some raw json object rather than a nice error message:

Image

MoritzBoehme:
Fixes #5708 

~~I still have to do some clean-up and refactoring. Not really satisfied with the way I handled formatting. I would like to use space_table, but this is implemented in compiler_cli :/~~
I think this is quite okay now.

Adding a snapshot test for the error formatting might also be nice. Insta is not yet a dependency of hexpm, so I did not do so for now.

### DIFF
--- a/hexpm/src/lib.rs
+++ b/hexpm/src/lib.rs
@@ -591,8 +591,14 @@ pub fn api_revert_release_response(response: http::Response<Vec<u8>>) -> Result<
         StatusCode::TOO_MANY_REQUESTS => Err(ApiError::RateLimited),
         StatusCode::UNAUTHORIZED => Err(unauthorised_response(&parts.headers)),
         StatusCode::FORBIDDEN => Err(ApiError::Forbidden),
+        StatusCode::UNPROCESSABLE_ENTITY if is_late_deletion(&body) => Err(ApiError::LateDeletion),
         status => Err(ApiError::unexpected_response(status, body)),
     }
+}
+
+fn is_late_deletion(body: &[u8]) -> bool {
+    String::from_utf8_lossy(body)
+        .contains("can only delete a release up to one hour after publication")
 }
 
 /// See: https://github.com/hexpm/hex/blob/main/lib/mix/tasks/hex.owner.ex#L47
@@ -761,6 +767,9 @@ pub enum ApiError {
 
     #[error("Can only modify a release up to one hour after publication")]
     LateModification,
+
+    #[error("Can only delete a release up to one hour after publication")]
+    LateDeletion,
 
     #[error("The oauth request wasn't approved in time")]
     OAuthTimeout,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,9 @@
   `package.hexdocs.pm` rather than `hexdocs.pm/package`.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- More readable error message when trying to revert an old release.
+  ([Moritz Böhme](https://github.com/MoritzBoehme))
+
 ### Language server
 
 - The "pattern match on value" code action can now be used to pattern match on
--- a/compiler-core/src/dependency.rs
+++ b/compiler-core/src/dependency.rs
@@ -272,6 +272,7 @@ impl PackageFetchError {
             | hexpm::ApiError::IncorrectChecksum
             | hexpm::ApiError::Forbidden
             | hexpm::ApiError::NotReplacing
+            | hexpm::ApiError::LateDeletion
             | hexpm::ApiError::LateModification => Self::ApiError(api_error),
         }
     }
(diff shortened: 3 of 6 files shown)


D-matz:
@MoritzBoehme I'm not totally sure about this but would it make sense to use the new nicely printed validation error for all of the response handling functions? https://github.com/gleam-lang/gleam/issues/5708#issuecomment-4502222616

MoritzBoehme:
Could be, but I don't know if the added complexity is justified. I just adjusted this PR and I think it is quite neat as it is.

MoritzBoehme:
I think I misunderstood your previous comment. This could be used for any validation error returned by the hex API. But I would not want to do so in this PR but rather look at other cases and see if it makes sense then.

giacomocavalieri (approved):
Looks good to me!

hexpm/src/lib.rs
```diff
@@ -796,6 +807,25 @@ impl ApiError {
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let errors: String = self
+            .errors
+            .iter()
+            .map(|(key, value)| format!("{}: {}", key, value))
+            .collect::<Vec<_>>()
+            .join("\n");
+        write!(f, "{}", errors)
+    }
+}
```
lpil: We do not implement traits unless unavoidable, and this is not a correct use of the display trait. Not all displaying of this type would have this particular format.

Please move this error printing logic into the existing error printing function 🙏 

I think that collect may be unnecessary also.

hexpm/src/lib.rs
```diff
@@ -777,13 +778,23 @@ pub enum ApiError {
...
     fn unexpected_response(status: StatusCode, body: Vec<u8>) -> Self {
         ApiError::UnexpectedResponse(status, String::from_utf8_lossy(&body).to_string())
     }
 
+    fn validation_error(body: Vec<u8>) -> Self {
+        match serde_json::from_slice(&body) {
+            Ok(body) => ApiError::ValidationError(body),
+            Err(_) => Self::unexpected_response(StatusCode::UNPROCESSABLE_ENTITY, body),
+        }
+    }
```
lpil: The name of this function is `validation_error`, but it does not always return a validation error.

In the path where it does return the `ValidationError` variant it checks that it can be deserialisation to the `ValidationErrorBody` structure, but it doesn't check that it actually is a validation error. The deserialisation will accept any Hex JSON error at all, so it's likely that the `ValidationErrorBody` structure is misnamed.

hexpm/src/lib.rs
```diff
@@ -796,6 +807,25 @@ impl ApiError {
     }
 }
 
+#[derive(Debug, PartialEq, Eq, serde::Deserialize)]
+pub struct ValidationErrorBody {
+    pub message: String,
+    pub status: u16,
```
lpil: The `message` and `status` fields seem to be unused.

hexpm/src/lib.rs
```diff
@@ -777,13 +778,23 @@ pub enum ApiError {
 
     #[error("The supplied one-time-password was not correct")]
     IncorrectOneTimePassword,
+
+    #[error("Validation error(s):\n{0}")]
+    ValidationError(ValidationErrorBody),
```
lpil: Have the data held directly by this variant rather than having it hold a structure please.

lpil (commented):
Thank you! A good start. I've left some notes inline, please un-draft this PR and tag me when it is ready for another review.

More generally, it would be great to stick with our pattern of making custom error messages for specific situations, so for this issue we want to focus on that old release error case. The generic improved formatting of generic errors from Hex is good too, but we want to make sure this specific one is as good as possible!

hexpm/src/lib.rs
MoritzBoehme: I see, this is currently confusing. I will change this.

Regarding the naming of the ValidationErrorBody I named it such because this is the only client error with an errors field according to the [specification](https://github.com/hexpm/specifications/blob/main/apiary.apib). But in regard to using the most specific error I will try to remove the ValidationErrorBody structure anyway and then add a specific error for the too old release.

hexpm/src/lib.rs
MoritzBoehme: Yeah makes sense, this also felt a bit weird to implement 😅

MoritzBoehme:
Thanks for the review! I will address everything over the weekend.

MoritzBoehme:
I removed the generic pretty printing in favour of a more specific error message. I could make a new PR for the generic case, but I agree that it is probably nicer to have specific messages for individual errors.

MoritzBoehme:
I just realized that by tag me you probably meant in a message not a request to review @lpil  😅

lpil (approved):
Lovely work! Very clean!! Thank you

Your other idea of having better rendering for generic errors sounds good too, please make an issue for that one 🙏

lpil:
No need to worry about that validation failure, I've fixed it on the main branch.

PR #5729, opened 2026-05-18
Fault tolerant guards

giacomocavalieri:
While fixing a language server bug I realized the proper fix needed clause guards to be fault tolerant. That is quite the big change on its own so I'm putting this into its own PR to make it easier to review!

With this clause guards in case expressions are fault tolerant, an error in a part of one is no longer blocking the compiler from analyzing the rest of it, or reporting correct errors for missing patterns


lpil (approved):
Fab!!! Thank you!!

PR #5737, opened 2026-05-19
Fill labels in constant records

### LINKED ISSUE #4730: Support "fill labels" code action for const
Fill labels code action is already available for variant constructors in functions, but it's not for consts:

Image

giacomocavalieri:
This PR fixes #4730

This required actually so much more work than I expected. Allowing `todo`s in constants was just the tip of the iceberg!

I tried my best to keep the history of changes nice and tidy and split it in piece meal commits. I would strongly recommend going over the changes one commit at a time when reviewing!


compiler-core/src/ast/constant.rs
```diff
@@ -476,6 +496,23 @@ impl TypedConstant {
             | Constant::Invalid { .. } => None,
         }
     }
+
+    /// If the constant is a record or record update this returns its tag.
+    /// It might return `None` if the record constructor couldn't be inferred.
```
lpil, line 501: When could it not be inferred?

compiler-core/src/type_/expression.rs
```diff
@@ -4527,6 +4291,191 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
...
+            ValueConstructorVariant::Record { field_map, .. } => field_map.clone(),
+
+            ValueConstructorVariant::ModuleFn { .. }
+            | ValueConstructorVariant::LocalVariable { .. } => {
+                self.problems
+                    .error(Error::NonLocalClauseGuardVariable { location, name });
+                return self.new_invalid_constant(location);
+            }
+
+            // TODO: remove this clone. Could be an rc instead
```
lpil: What clone?

lpil (commented):
Wow! That was quite involved! Very nice work, I couldn't find any issues at all.

I left a couple of tiny notes on some of the comments.

compiler-core/src/type_/expression.rs
giacomocavalieri: Oh this is probably a leftover TODO from the old code that I forgot to delete 😄

compiler-core/src/ast/constant.rs
giacomocavalieri: I think the main case is when we write a variant that doesn't exist:
```
pub const a = ThisDoesNotExist(1, 2)
```
In that case the compiler won't be able to find any constructor for it

giacomocavalieri: I've added an example in the docs!

lpil (approved):
Thank you!!!!

PR #5738, opened 2026-05-19
Fix binop in clauses

giacomocavalieri:
I noticed it was not possible to fix wrong binary operators in guards (while it is possible in expressions). This PR fixes it!


compiler-core/src/ast.rs
```diff
@@ -2260,6 +2260,12 @@ pub enum ClauseGuard<Type, RecordTag> {
     BinaryOperator {
         location: SrcSpan,
         operator: BinOp,
+        /// This is the span covering the operator itself. For example:
+        /// ```gleam
+        /// _ if 1.0 +. 2.3 == 3.0 -> todo
+        /// //       ^^ This!
+        /// ```
+        operator_location: SrcSpan,
```
lpil: Use a position please, the second int of the span is redundant as operators have a fixed fix 🙏

lpil (commented):
Thank you! One tiny note inline

lpil (approved):
Thank you!!!

PR #5739, opened 2026-05-20
Add support for multiline strings with triple-quote syntax

cfsmi:
First commit!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ### Compiler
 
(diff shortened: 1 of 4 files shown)


lpil (commented):
Thank you, but we do not accept unplanned changes, as indicated by the pull-request template. Please open a discussion if you wish to make a proposal.

This specific change is not acceptable as it includes breaking changes to the language syntax, changing the meaning of existing valid Gleam programs such as this one.

```gleam
pub fn main() {
  """
  Example
  """
}
```

PR #5741, opened 2026-05-22
Simplify the failed Hex API key decryption error message

### LINKED ISSUE #5710: Error message for inserting the wrong local Hex API key could be improved

c-tonneslan:
Closes #5710.

When you give the wrong password for the local Hex API key, the error currently repeats itself and then dumps the raw error from the encryption library:

```
error: Failed to decrypt local Hex API key

Unable to decrypt the local Hex API key with the given password.

    unable to decrypt message: Decryption failed
```

That last bit is always the same generic "Decryption failed" for a bad password, so it's just noise, and "the encryption library" isn't something a user cares about. This drops it and adds a hint instead:

```
error: Failed to decrypt local Hex API key

Unable to decrypt the local Hex API key with the given password.
Hint: Check that you entered the correct local password and try again.
```

Since the `detail` field on the error was only used for that line, I removed it too. Added a snapshot test alongside the existing `hex_session_revoked` one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -139,6 +139,11 @@
   improves suggestions to push a tag, if it doesn't exists.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- The error shown when the local Hex API key can't be decrypted no longer
+  prints the raw encryption library error, and now hints to check the
+  password.
+  ([Charlie Tonneslan](https://github.com/c-tonneslan))
+
 ### Language server
 
 - The language server now offers a "Fill labels" code action on constants to
--- a/compiler-cli/src/hex/auth.rs
+++ b/compiler-cli/src/hex/auth.rs
@@ -229,9 +229,7 @@ It will be used to locally encrypt your Hex API tokens.
             encrypted_credentials.refresh_token.as_bytes(),
             &local_password,
         )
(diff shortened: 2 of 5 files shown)


lpil:
Thank you, but "good first issue" tagged issues are intended to be used for humans to learn and get introduced to the codebase. If we use genAI to close them then we do not have a way for humans to learn, which is a risk to the long-term sustainability of the project.

PR #5742, opened 2026-05-22
Show a readable error when reverting a release that's too old

### LINKED ISSUE #5708: Bad error message when trying to revert an old release

c-tonneslan:
Closes #5708.

`gleam hex revert` on a release that's too old to revert (Hex only lets you modify a release within an hour of publishing) printed the raw JSON error body straight from Hex instead of a readable message.

`api_publish_package_response` already handles this: on a 422 it checks the body for "can only modify a release up to one hour after publication" and returns `ApiError::LateModification`, which has a proper message. `api_revert_release_response` just never grew the same arm, so the response fell through to the `unexpected_response` fallback and the JSON leaked out.

This adds the matching 422 handling to the revert path. Reverting a too-old release now surfaces the existing "Can only modify a release up to one hour after publication" error. Test mirrors the existing `modify_package_late` one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -330,6 +330,10 @@
 
 ### Bug fixes
 
+- Fixed a bug where `gleam hex revert` printed a raw JSON error body when a
+  release was too old to be reverted, instead of a readable message.
+  ([Charlie Tonneslan](https://github.com/c-tonneslan))
+
 - Fixed a bug where `gleam remove` would fail with a confusing File IO error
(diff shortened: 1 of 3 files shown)


lpil:
Thank you, but "good first issue" tagged issues are intended to be used for humans to learn and get introduced to the codebase. If we use genAI to close them then we do not have a way for humans to learn, which is a risk to the long-term sustainability of the project.

PR #5746, opened 2026-05-23
Fix duplicate let declarations for a case in a pipeline

### LINKED ISSUE #5743: Invalid JavaScript generated for case in a pipeline
- gleam 1.16.0
- target JS

This code gets compiled to an invalid JS file.

```gleam
import gleam/function
import gleam/io

pub fn main() {
  0
  |> case True {
    True -> function.identity |> echo
    False -> fn(i) { i + 1 }
  }
  |> fn(i) { i + 1 }
}
```

Compiled to a JS file with this error:
`SyntaxError: Identifier '_pipe$1' has already been declared`

```js
export function main() {
  let _pipe = 0;
  let _block;
  let $ = true;
  let _pipe$1 = $function.identity;
  _block = echo(_pipe$1, undefined, "src/tmp_gleam.gleam", 7);
  let _pipe$1 = _block(_pipe);
  return ((i) => { return i + 1; })(_pipe$1);
}
```

eyupcanakman:
When a `case` expression is used as a step in a pipeline, the JavaScript code generator could declare the same `_pipe$N` variable twice in one scope. That is a `SyntaxError`.

A directly matching `case` branch generates its code into the enclosing scope, so the counters for the compiler's internal variables have to survive past it. Before this only the `$` case subject counter was kept. The `_pipe` counter got reset, so the next step in the outer pipeline reused the name.

Fixes #5743.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,11 @@
   function would throw an Error using the wrong class.
   ([John Downey](https://github.com/jtdowney))
 
+- Fixed a bug where the JavaScript code generator could produce duplicate `let`
+  declarations for internal variables when a `case` expression was used as a
+  step in a pipeline.
+  ([Eyup Can Akman](https://github.com/eyupcanakman))
+
 ## v1.17.0-rc1 - 2026-05-23
 
 ### Compiler
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -1,6 +1,6 @@
 use super::{
     INDENT, bit_array_segment_int_value_to_bytes,
-    expression::{self, Generator, Ordering, float, float_from_value},
+    expression::{self, Generator, Ordering, Scope, float, float_from_value},
 };
 use crate::{
     ast::{AssignmentKind, Endianness, SrcSpan, TypedClause, TypedExpr, TypedPattern},
@@ -23,7 +23,7 @@ use itertools::Itertools;
 use num_bigint::BigInt;
 use std::{collections::HashMap, sync::OnceLock};
 
-pub static ASSIGNMENT_VAR: &str = "$";
+pub const ASSIGNMENT_VAR: &str = "$";
 
 pub fn case<'a>(
     compiled_case: &'a CompiledCase,
@@ -427,14 +427,20 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
 
             // We can't use `inside_new_scope` here without care: the
             // code we generate goes directly into the enclosing scope
-            // (there's no wrapping if-else block), so the `$` variable
-            // counter must not be reset or later code could redeclare
-            // one of the `$N` variables introduced here.
-            let old_scope = match &self.kind {
+            // (there's no wrapping if-else block). User variables bound
+            // in the branch go out of scope at its end, so their counters
+            // are reset and later references resolve to the outer bindings.
+            // Compiler synthesised variables (the `$` case subjects, `_pipe`,
+            // `_block`, ...) can't be referenced by user code and their
+            // declarations leak into this scope. Restoring only the user-variable
+            // counters leaves the synthesised counters advanced, so later code
+            // can't redeclare one of them.
+            let old_user_variables = match &self.kind {
                 DecisionKind::Case { .. } => self
                     .variables
                     .expression_generator
-                    .current_scope_vars
+                    .current_scope
+                    .user_variables()
                     .clone(),
                 DecisionKind::LetAssert { .. } => Default::default(),
             };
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -485,3 +485,8 @@
   would not be offered when the cursor was on an inexhaustive `case` expression
   if another inexhaustive `case` appeared earlier in the same module.
   ([John Downey](https://github.com/jtdowney))
+
+- Fixed a bug where the JavaScript code generator could produce duplicate `let`
+  declarations for internal variables when a `case` expression was used as a
+  step in a pipeline.
+  ([Eyup Can Akman](https://github.com/eyupcanakman))
```
lpil: This has been added to v1.17.0-rc1'c changelog, but that has already been released

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__no_duplicate_pipe_variable_with_case_in_pipeline.snap
```diff
@@ -0,0 +1,41 @@
+---
+source: compiler-core/src/javascript/tests/case.rs
+expression: "\npub fn main() {\n  0\n  |> case True {\n    True -> add_one |> identity\n    False -> add_one\n  }\n  |> add_one\n}\n\nfn identity(x) {\n  x\n}\n\nfn add_one(x) {\n  x + 1\n}"
+---
+----- SOURCE CODE
+
+pub fn main() {
+  0
+  |> case True {
```
lpil, line 9: Could you replace this `True` with a parameter given to this function? Otherwise any future constant propagation would prevent this test from doing anything.

test/language/test/language/directly_matching_case_subject_test.gleam
```diff
@@ -46,3 +54,15 @@ pub fn no_duplicate_let_after_case_with_same_named_variable_with_declaration_bef
   let x = x
   assert x == 2
 }
+
+// https://github.com/gleam-lang/gleam/issues/5743
+pub fn no_duplicate_pipe_variable_with_case_in_pipeline_test() {
+  let result =
+    0
+    |> case True {
```
lpil, line 66: As above, could you not use a constant subject please 🙏

compiler-core/src/javascript/decision.rs
```diff
@@ -446,13 +454,31 @@ impl<'a> CasePrinter<'_, '_, 'a, '_> {
...
+                        .filter_map(|&name| {
+                            let counter = self
+                                .variables
+                                .expression_generator
+                                .current_scope_vars
+                                .get(name)
+                                .copied()?;
+                            Some((name.into(), counter))
+                        })
+                        .collect();
```
lpil: It's a shame we have to do all these hash map lookups to get these values, and then to later set them again. We would replace this hash map with a class that contains the hashmap, but holds these other values as fields.

Do you think that'd be a good idea?

lpil (commented):
Thank you!!! Some notes inline.

CHANGELOG.md
eyupcanakman: Good catch, moved it up to a new Unreleased section.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__no_duplicate_pipe_variable_with_case_in_pipeline.snap
eyupcanakman: Ah, I tried this. A parameter subject actually stops the test catching the bug, because the duplicate `let` only happens when the subject is known at compile time and the case folds to a single branch. With a parameter it compiles to a block-scoped `if`/`else` where the names can't clash, so it passes even before the fix. Kept `True` here, but happy to change it if you'd prefer.

test/language/test/language/directly_matching_case_subject_test.gleam
eyupcanakman: Same here, so I kept the constant subjects but made the two branches return different values, so the assert actually checks which branch ran.

compiler-core/src/javascript/decision.rs
eyupcanakman: Yeah, much nicer. Done. There's now a `Scope` struct holding the user-variable map plus the synthesised counters (`$`, `_pipe`, `_block`, ...) as fields, so restoring only the user variables leaves the synthesised ones advanced and we don't have to look them up and set them again afterwards.

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__no_duplicate_pipe_variable_with_case_in_pipeline.snap
lpil: I see! Thank you!

lpil (approved):
Thank you very much!!!

PR #5732, opened 2026-05-18
Fix add missing patterns action skipping later cases

jtdowney:
Something I noticed earlier when doing a bigish refactoring, and was surprised by.

- [ ] The changes in this PR have been discussed beforehand in an issue
- [ ] The issue for this PR has been linked
- [x] Tests have been added for new behaviour
- [x] The changelog has been updated for any user-facing changes

PR #5747, opened 2026-05-24
Fix inexhaustive let-to-case skipping later cases

jtdowney:
This is similar to #5732 in that I encountered it during a refactoring, when I noticed an action wasn't available except on the first instance.


lpil (approved):
Thank you!!

PR #5753, opened 2026-05-25
Fix error message on Windows on publish with no README

### LINKED ISSUE #5752: bug: bad error message when trying to publish package with no README on Windows

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5752
***

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -4,7 +4,9 @@
 use gleam_core::{
     Result, Warning,
     build::{NullTelemetry, Target},
-    error::{Error, FileIoAction, FileKind, OS, ShellCommandFailureReason, parse_os},
+    error::{
+        Error, FileIoAction, FileIoFailure, FileKind, OS, ShellCommandFailureReason, parse_os,
+    },
     io::{
         BeamCompilerIO, Command, CommandExecutor, Content, DirEntry, FileSystemReader,
         FileSystemWriter, OutputFile, ReadDir, Stdio, WrappedReader, is_native_file_extension,
@@ -37,7 +39,11 @@ pub fn get_current_directory() -> Result<Utf8PathBuf, Error> {
         kind: FileKind::Directory,
         action: FileIoAction::Open,
         path: ".".into(),
-        err: Some(e.to_string()),
+        err: if let io::ErrorKind::NotFound = e.kind() {
+            FileIoFailure::NotFound
+        } else {
+            FileIoFailure::Other(e.to_string())
+        },
     })?;
     Utf8PathBuf::from_path_buf(curr_dir.clone()).map_err(|_| Error::NonUtf8Path { path: curr_dir })
 }
@@ -163,7 +169,11 @@ fn is_same_file(left: &Utf8Path, right: &Utf8Path) -> Result<bool, Error> {
         action: FileIoAction::ReadMetadata,
         kind: FileKind::File,
         path: left.to_path_buf(),
-        err: Some(e.to_string()),
+        err: if let io::ErrorKind::NotFound = e.kind() {
+            FileIoFailure::NotFound
+        } else {
+            FileIoFailure::Other(e.to_string())
+        },
     })
 }
 
@@ -174,7 +184,11 @@ pub fn modification_time(path: &Utf8Path) -> std::result::Result<SystemTime, Err
             action: FileIoAction::ReadMetadata,
             kind: FileKind::File,
             path: path.to_path_buf(),
-            err: Some(e.to_string()),
+            err: if let io::ErrorKind::NotFound = e.kind() {
+                FileIoFailure::NotFound
+            } else {
+                FileIoFailure::Other(e.to_string())
+            },
         })
 }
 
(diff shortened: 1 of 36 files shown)


lpil:
Thank you! I think this problem highlights that the string-based approach here isn't a good one. Instead of comparing the string contained by this error, let's check if the file exists and return an error if it does not.

jtdowney:
Agreed, I'd be worried that an OS locale setting would cause the error message to not match the string in some cases. Not what I've seen with Rust by default on macOS/Linux, but I could see it happening on Windows. It seems like adding an `Error::FileNotFound` variant that is constructed when `err.kind() == io::ErrorKind::NotFound` is probably best.

ankddev:
> It seems like adding an `Error::FileNotFound` variant that is constructed when `err.kind() == io::ErrorKind::NotFound` is probably best.

I think that this approach is not nice, because it needs dedicated variant instead of exisitng I/O error. One approach I think is fine is to include kind of I/O error and not only message.

jtdowney:
That works too. It's sort of a case of pick your poison. You either add a field or you add a variant. In either case, you're going to have to handle it in a number of places where you previously hadn't bothered to consider the variant, or where it's irrelevant. The reason I suggested the variant is that I expected it would be a smaller overall change.

ankddev:
Makes sense to me, thank you!

lpil:
Undraft and tag me when ready for a review, thank you!

ankddev:
@lpil @jtdowney I think that it's done now!

compiler-cli/src/fs.rs
```diff
@@ -708,16 +701,11 @@ pub fn symlink_dir(src: impl AsRef<Utf8Path>, dest: impl AsRef<Utf8Path>) -> Res
...
-    let result = std::os::unix::fs::symlink(src, dest);
+    let result = std::os::unix::fs::symlink(&src, dest);
 
-    result.map_err(|err| Error::FileIo {
-        action: FileIoAction::Link,
-        kind: FileKind::File,
-        path: Utf8PathBuf::from(dest),
-        err: Some(err.to_string()),
-    })?;
+    result.map_err(|err| Error::io(err, FileKind::File, src.to_path_buf(), FileIoAction::Link))?;
```
jtdowney: The path reported here was `dest` but this changes it to `src`. I am not sure which is expected, it is just something I noticed.

compiler-cli/src/fs.rs
```diff
@@ -575,7 +578,7 @@ pub fn mkdir(path: impl AsRef<Utf8Path> + Debug) -> Result<(), Error> {
 
     std::fs::create_dir_all(path.as_ref()).map_err(|err| Error::FileIo {
```
jtdowney, line 594: Should this use `Error::io`?

compiler-cli/src/fs.rs
ankddev: Since PR that adds destination to errors was merged, this will be resolved after rebase

compiler-cli/src/fs.rs
ankddev: I think that I forgot to add it, because `NotFound` error probably isn't possible here.

compiler-cli/src/fs.rs
ankddev: Now it uses only source path when the file is not found and both source and destination when there's other error.

compiler-cli/src/fs.rs
ankddev: Done!

ankddev:
@jtdowney I've addressed your comments!

lpil (commented):
Nice, thank you!

This new error has much less detail than the previous one, making it much harder to understand and debug. How about instead we continue to use the existing file-io variant, and changing the `err` field to a `FileIoFailure`

```rust
pub enum FileIoFailure {
    NotFound,
    Other(String)
}
```

That way we keep all the detail, and we can remove new less-clear `io` function and keep using the named fields of the FileIo error variant.

ankddev:
Yeah, this is a nice idea, thank you!

ankddev:
@lpil @jtdowney Now it uses suggested approach!

compiler-core/src/error.rs
```diff
@@ -1850,10 +1857,14 @@ Erlang modules must have unique names regardless of the subfolders where their
                 err,
             } => {
                 let err = match err {
-                    Some(e) => {
+                    FileIoFailure::NotFound => {
+                        "\nThe error message from the file IO library was:\n\n    File Not Found\n"
```
ankddev, line 1861: I'm not sure about this format, open for ideas!

lpil (commented):
There's a lot of duplicated code for constructing the error value, could you extract it into a `FileIoFailure::from_io_error` function please 🙏

compiler-core/src/error.rs
lpil: Looks good! Thought there's an extra empty line in the error

lpil:
> There's a lot of duplicated code for constructing the error value, could you extract it into a FileIoFailure::from_io_error function please 🙏

This hasn't been done!

ankddev:
Yeah, I only rebased it, sorry!

lpil:
Oops! I'll draft this so I don't confuse myself again

ankddev:
@lpil Done!

PR #5758, opened 2026-05-27
JS: fix singleton equality checks with imported aliased variant

### LINKED ISSUE #5757: Equality with an aliased variant silently returns False on JavaScript
- **Gleam**: 1.16.x
- **Target**: `javascript` only — Erlang is unaffected

On the JavaScript target, comparing a value to a no-args variant that you imported under an alias can produce the wrong result. The same code compiled to Erlang works correctly.

```gleam
// src/aliased_variant_bug/foo.gleam
pub type FooStatus { Active Inactive }

// src/aliased_variant_bug/bar.gleam
pub type BarStatus { Active Disabled }

// src/aliased_variant_bug.gleam
import aliased_variant_bug/bar.{type BarStatus, Active as BarActive}
import aliased_variant_bug/foo.{type FooStatus, Active}

pub fn is_foo_active(s: FooStatus) -> Bool { s == Active }
pub fn is_bar_active(s: BarStatus) -> Bool { s == BarActive }
```

Generated `aliased_variant_bug.mjs`:

```js
import * as $bar from "./aliased_variant_bug/bar.mjs";
import { Active as BarActive } from "./aliased_variant_bug/bar.mjs";
import * as $foo from "./aliased_variant_bug/foo.mjs";
import { Active } from "./aliased_variant_bug/foo.mjs";

export function is_foo_active(s) { return s instanceof Active; }
export function is_bar_active(s) { return s instanceof Active; }  // wrong — should be BarActive
```

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5757
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@
   step in a pipeline.
   ([Eyup Can Akman](https://github.com/eyupcanakman))
 
+- Fixed a bug where the JavaScript code generator would use wrong constructor
+  name in equality comparisons for imported variant with zero arguments and
+  import alias.
+  ([Andrey Kozhev](https://github.com/ankddev))
+
 ## v1.17.0-rc1 - 2026-05-23
 
 ### Compiler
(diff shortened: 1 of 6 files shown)


GearsDatapacks (approved):
Thank you! Funnily enough I found this exact bug (and fixed it the same way) in #5759, then came over to Github and found that there was already an issue and PR.

PR #5759, opened 2026-05-27
Use singleton values for fieldless variants on the JavaScript target

### LINKED ISSUE #5756: Use singleton values for fieldless variants and the empty list on JavaScript
The empty list and variants without fields are singletons, and all instances/uses of it are exactly the same. The Erlang implementation takes advantage of this so there is only ever one of each, but on the JavaScript implementation a new object is allocated for each use. This results in a small but pervasive performance cost.

Change the JavaScript implementation create a single instance of each of these values and to always use them.

Here is a proposed design. The exact design is open to discussion and can be changed.

## Fieldless variant

```gleam
pub type Wibble {
  Wobble
}

pub const constant = Wobble

pub fn expression() -> Wibble {
  Wobble
}
```
```js

export class Wobble extends $CustomType {}
export const Wibble$Wobble$const = new Wobble(); // <- New
export const Wibble$Wobble = () => Wibble$Wobble$const; // <- Changed
export const Wibble$isWobble = (value) => value instanceof Wobble;

export const constant = /* @__PURE__ */ Wibble$Wobble$const; // <- Changed

export function expression() {
  return Wibble$Wobble$const; // <- Changed
}
```

## Empty list

```gleam
pub const constant = []

pub fn expression() -> List(a) {
  []
}
```
```js
import { List$NonEmpty$const } from "../gleam.mjs";

export const constant = List$NonEmpty$const; // <- Changed

export function expression() {
  return List$NonEmpty$const; // <- Changed
}
```

The same change to the external API would be made for the prelude for the empty list.

No differences are to be made to equality testing as existing code may be using the constructor directly. Once all code is using the correct API we can further optimise equality testing.

In a previous discussion it was suggested we could store the singleton instance as a static property on the class object, but I opted for a const as I believe this will minify better (as property names do not get mangled).

GearsDatapacks:
Closes #5756

I've also made empty lists consistent with other empty variants so that expressions in the form `x == []` will be compiled to `x instanceof $Empty` instead of calling the much slower `isEqual` function.

I wasn't 100% sure about how to rename imported constants for aliased variants; happy to hear suggestions for alternative naming schemes.

### DIFF
--- a/compiler-core/src/javascript.rs
+++ b/compiler-core/src/javascript.rs
@@ -520,26 +520,32 @@ impl<'a> Generator<'a> {
     ) -> Document<'a> {
         let class_definition = self.variant_class_definition(constructor, publicity);
 
-        // If the custom type is private or opaque, we don't need to generate API
-        // functions for it.
-        if publicity.is_private() {
-            return class_definition;
-        }
-
+        // Singleton constants are used internally by the compiler, so they aren't
+        // part of the public JS API and need to be generated even for private
+        // types.
         let constructor_singleton = if constructor.arguments.is_empty() {
-            self.variant_constructor_constant(constructor, type_name)
-                .append(line())
+            docvec![
+                line(),
+                self.variant_constructor_constant(constructor, type_name, publicity)
+            ]
         } else {
             nil()
         };
+
+        // If the custom type is private or opaque, we don't need to generate API
+        // functions for it.
+        if publicity.is_private() {
+            return class_definition.append(constructor_singleton);
+        }
+
         let constructor_definition = self.variant_constructor_definition(constructor, type_name);
         let variant_check_definition = self.variant_check_definition(constructor, type_name);
         let fields_definition = self.variant_fields_definition(constructor, type_name);
 
         docvec![
             class_definition,
-            line(),
             constructor_singleton,
+            line(),
             constructor_definition,
             line(),
             variant_check_definition,
@@ -555,10 +561,16 @@ impl<'a> Generator<'a> {
         &self,
         constructor: &'a TypedRecordConstructor,
         type_name: &'a str,
+        publicity: Publicity,
     ) -> Document<'a> {
+        let keyword = if publicity.is_importable() {
+            "export const "
+        } else {
+            "const "
+        };
         docvec![
             self.source_map_tracker(constructor.location.start),
-            "export const ",
+            keyword,
             type_name,
             "$",
             constructor.name.as_str(),
--- a/compiler-core/src/javascript.rs
+++ b/compiler-core/src/javascript.rs
@@ -6,7 +6,7 @@ mod tests;
 mod typescript;
 
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::rc::Rc;
 
 use debug_ignore::DebugIgnore;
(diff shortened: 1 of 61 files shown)


compiler-core/src/javascript.rs
```diff
@@ -223,6 +223,14 @@ impl<'a> Generator<'a> {
             self.register_prelude_usage(&mut imports, "Empty", Some("$Empty"));
         };
 
+        if self.tracker.list_empty_const_used {
+            self.register_prelude_usage(
+                &mut imports,
+                "List$Empty$const",
+                Some("$List$Empty$const"),
```
lpil, line 230: I don't think we need to add a dollar prefix here? The canonical `List$Empty$const` can't collide with anything else, no?

lpil (commented):
Beautiful work!!! I've left one little comment inline.

I've also asked Hayleigh if she thinks the design needs any tweaks.

Do we have any benchmarks? With the empty-list change maybe a Lustre view function could be good to measure.

> I wasn't 100% sure about how to rename imported constants for aliased variants; happy to hear suggestions for alternative naming schemes.

Not totally sure what you mean by this- could you expand on that?

compiler-core/src/javascript.rs
GearsDatapacks: It can, for example:
```gleam
// list.gleam
pub type List(a) {
  Empty
  NonEmpty(a, List(a))
}

// main.gleam
import list.{Empty}

pub fn main() {
  #(Empty, [])
}
```

Here, both the `Empty` value and the empty list have the name `List$Empty$const`.

GearsDatapacks:
> Not totally sure what you mean by this- could you expand on that?

For example, see the code produces in [this snapshot](https://github.com/gleam-lang/gleam/pull/5759/changes#diff-b241eb06e2425b1e32f7e3a0925716ec8106ca1ca988238326cb2dee8d70e8cc). Are you okay with this or would you prefer something different for the aliased variants?

compiler-core/src/javascript.rs
lpil: Ah!!! I understand now. Thank you very much

lpil:
I'm happy with that variant handling, thank you

yoshi-monster:
Gears working hard to make me write Gleam code like a normal person again, thank you ❤️ 

Anyways I do this optimisation manually in Lustre because of course I do, so I also have benchmarks for our diffing algorithm (red is with that optimisation manually applied for `[]` and `Error(Nil)`, green is idiomatic Gleam):

image

one place that would also be nice to change is in the preludes `toList` / `List.fromArray` function, since with that change `[item, ..[]]` would be faster than `[item]` :)

lpil:
Like this @yoshi-monster ?

https://github.com/gleam-lang/gleam/pull/5759/changes/2f6c88fb6550bf57427ec4f1df473ad5dfb45a84

yoshi-monster:
Yes perfect, thank you so much :D 💜

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__functions__labelled_argument_ordering.snap
```diff
@@ -41,12 +41,12 @@ function wibble(a, b, c, d) {
...
 export function main() {
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  wibble(new A(), new B(), new C(), new D());
-  return wibble(new A(), new B(), new C(), new D());
+  wibble(A$A$const, B$B$const, C$C$const, D$D$const);
```
lpil, line 48: I don't think `A$A$const` is defined in this module?

GearsDatapacks: Ah yes! This is a bug. I believe it is because we only generate the new JS API for public types, but since we're using the `$const` value internally, we need to always generate it and not treat it like the public API.

GearsDatapacks:
Okay I've fixed that now.

lpil (approved):
Amazing work as always!!! Rebecca will be super pleased 😁

PR #5764, opened 2026-05-29
JS: fix missing `;` after some expressions in statement position

### LINKED ISSUE #5680: Pattern matching on tuples or custom types produces broken JavaScript
The following Gleam code 

```gleam
pub fn main() {
  case #(1, 2) {
    #(a, _) -> a
  }
  {2 + 4} * 2
  let a = 42
  Nil
}
```

produces this Javascript:

```js
export function main() {
  let $ = [1, 2];
  let a = $[0];
  a
  (2 + 4) * 2;
  let a = 42;
  return undefined;
}
```

Here, 2 things go wrong:

1. the `a` binding does not reserve the name in the enclosing scope, causing a parse  error.
2. The expression inside the arm of the pattern is not followed by a semicolon. The line after it is constructed in such a way that this is now parsed as a function call, changing the meaning of this program.

Both of these can be triggered independently.

~ 💜

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5680 (resolves its second part)
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   every inexhaustive `let` assignment in a module other than the first one.
   ([John Downey](https://github.com/jtdowney))
 
+- Fixed a bug where the JavaScript code generator would not emit semicolon
+  after some expressions.
+  ([Andrey Kozhev](https://github.com/ankddev))
+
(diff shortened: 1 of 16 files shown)


lpil (approved):
Fantastic work, thank you!

Issue #5763, opened 2026-05-28
`BitArray$BitArray$data` returns the wrong bits for un-aligned bit arrays


yoshi-monster:
I think this is a bug, I'm just a bit unsure which direction :D

`BitArray$BitArray$data` only rejects bit arrays whose `bitSize` is not a multiple of 8. The exception it throws says its rejecting un-aligned BitArrays, but that's not what's happening - it does not check `bitOffset`.

Basically it means a bit array can be byte-sized but not byte-aligned. In that case, `BitArray$BitArray$data` returns a DataView over the raw backing buffer rather than the logical bytes represented by the bit array.

```gleam
import gleam/int
import gleam/io

pub fn main() {
  let original = <<170, 85>>
  let assert <<_:size(1), shifted:bits-size(8), _:bits>> = original

  echo shifted
  io.println("first byte from ffi: " <> int.to_string(first_data_byte(shifted)))
}

@external(javascript, "./bitarray_data_repro_ffi.mjs", "first_data_byte")
fn first_data_byte(bits: BitArray) -> Int
```
```js
import { BitArray$BitArray$data } from "./gleam.mjs";

export function first_data_byte(bits) {
  return BitArray$BitArray$data(bits).getUint8(0);
}
```

```
src/bitarray_data_repro.gleam:8
<<84>>
first byte from ffi: 170
```

~ 💜

jtdowney:
Agree that this is clearly a bug. As for the direction, I'd expect it to match the deprecated `buffer` method.

jtdowney:
The more I stare at it, the less sure I am which way it should go. There is `bitArraySliceToFloat` which copies unaligned bytes, so maybe `BitArray$BitArray$data` should also copy unaligned bytes.

lpil:
The intention was for it to never copy, and for it to always pass a reference to the internal buffer. For some reason I/we thought that there would never be offset at the start, but looking now I don't know why I thought that.

Having to do a copy here is a real shame. Is there a way we could avoid that cost? I can't spot it.

PR #5766, opened 2026-05-29
Throw the global JS error instead of the Result subclass Error

jtdowney:
While looking at #5763, I noticed that this code was throwing the local class `Error` that is defined as a subclass of `Result` instead of the global JS class `Error`. The rest of the code was already using `globalThis.Error`.


lpil (approved):
Oops, thank you! Good catch.

PR #5771, opened 2026-05-30
publish: check README existence directly instead of matching OS error

### LINKED ISSUE #5752: bug: bad error message when trying to publish package with no README on Windows

c-tonneslan:
Closes #5752.

The check matched the raw IO error string \`"No such file or directory"\` to recognize a missing README, but Windows reports something like \`"The system cannot find the file specified"\`. On Windows the friendly \`Cannot publish with no README\` error never fired and the raw IO error showed up instead, as caught in a CI run on a different PR.

Switched to checking whether the path exists before reading, which is OS-independent and a bit clearer:

\`\`\`rust
let readme_path = paths.readme();
if !readme_path.exists() {
    return Err(Error::CannotPublishWithInvalidReadme {
        reason: InvalidReadmeReason::Missing,
    });
}
let project_readme = fs::read(readme_path)?;
\`\`\`

All 93 existing gleam-cli tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,12 @@
   function would throw an Error using the wrong class.
   ([John Downey](https://github.com/jtdowney))
 
+- Fixed a bug where publishing a package without a README would show a raw
+  IO error on Windows instead of the friendly "Cannot publish with no README"
+  message, because the check matched on the OS-emitted error string which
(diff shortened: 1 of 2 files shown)


lpil:
Sorry @c-tonneslan, but there already is a PR for this issue 🙏

PR #5773, opened 2026-05-31
Update imports when a file is renamed

### LINKED ISSUE #4370: LSP: Rename modules
Given a file called `foo.gleam`, (module name `foo`), and lots of other modules `import foo` and also call `foo.stuff()`.

When `foo.gleam` gets renamed, we can give the user a choice to auto-rename all references and imports.

We may also want to have a rename module code action when on top of a module reference or import, which would need to also rename the file.

GearsDatapacks:
Closes #4370

This PR updates `import` statements and module selects when a Gleam file is renamed. I've needed to reimplement the module-reference-finding code as a tracker the same as values and types rather than an AST visitor, as we don't have access to the ASTs of files which aren't currently open by the user.

The current behaviour is:
- If a module is imported without an alias, all references are update to match the new name
- If a module is imported with an alias, the alias is not touched and only the import path is updated

This makes the most sense to me, but let me know if you think a different behaviour is preferred.


compiler-core/src/reference.rs
```diff
@@ -8,21 +8,39 @@ use petgraph::{
...
 pub enum ReferenceKind {
-    Qualified,
+    Qualified {
+        module_alias: EcoString,
+        module_location: SrcSpan,
+    },
     Unqualified,
     Import,
     Definition,
     Alias,
```
giacomocavalieri, line 105: It's not totally related to this PR but could you maybe add a comment on the `Qualified` variant explaining what that is with a couple examples? Same goes for the newly added fields, would be great to have an example comment showing what those are and what that span covers

compiler-core/src/reference.rs
```diff
@@ -8,21 +8,39 @@ use petgraph::{
...
 
-#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct Reference {
     pub location: SrcSpan,
     pub kind: ReferenceKind,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub enum ModuleNameReference {
```
giacomocavalieri, line 120: Could you add a doc comment explaining what a `ModuleNameReference` is?

compiler-core/src/reference.rs
```diff
@@ -8,21 +8,39 @@ use petgraph::{
...
+    AliasedModuleSelect(SrcSpan),
+    Import {
+        module_location: SrcSpan,
+        import_end: u32,
+    },
+    AliasedImport {
+        module_location: SrcSpan,
+        alias_location: SrcSpan,
+        alias: EcoString,
+    },
```
giacomocavalieri, line 172: Could you add some doc comments for each variant with examples for each?

GearsDatapacks:
Added some more documentation! Hopefully it's clear what everything means now.

lpil (approved):
This is so nice! If only I could work out how to use it in neovim 😆 

Thank youu!!!

PR #5774, opened 2026-05-31
Fix heading level

GearsDatapacks:
I noticed that the "Bug fixes" section of the changelog was an h2 instead of an h3.


lpil (approved):
Thank you!

PR #5783, opened 2026-06-01
Use new hexdocs URL format

GearsDatapacks:

This PR updates the build tool to generate code using the new Hexdocs URL format.

### DIFF
--- a/compiler-core/src/docs.rs
+++ b/compiler-core/src/docs.rs
@@ -110,8 +110,8 @@ pub fn generate_html<IO: FileSystemReader>(
             path,
         });
 
-    let host = if is_hex_publish == DocContext::HexPublish {
-        "https://hexdocs.pm"
+    let url = if is_hex_publish == DocContext::HexPublish {
+        &format!("https://{}.hexdocs.pm", config.name.replace("_", "-"))
     } else {
         ""
     };
@@ -175,7 +175,7 @@ pub fn generate_html<IO: FileSystemReader>(
             project_version: &config.version.to_string(),
             content: rendered_content,
             rendering_timestamp: &rendering_timestamp,
-            host,
+            url,
             unnest: &unnest,
         };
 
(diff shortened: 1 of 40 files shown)


test/publishing_default_readme/README.md
```diff
@@ -14,7 +14,7 @@ pub fn main() -> Nil {
 }
 ```
 
-Further documentation can be found at <https://hexdocs.pm/default_readme>.
+Further documentation can be found at <https://default_readme.hexdocs.pm>.
```
ankddev:
```suggestion
Further documentation can be found at <https://default-readme.hexdocs.pm>.
```

compiler-core/src/docs.rs
```diff
@@ -110,10 +110,10 @@ pub fn generate_html<IO: FileSystemReader>(
             path,
         });
 
-    let host = if is_hex_publish == DocContext::HexPublish {
-        "https://hexdocs.pm"
+    let url = if is_hex_publish == DocContext::HexPublish {
+        format!("https://{}.hexdocs.pm", config.name.replace("_", "-"))
```
ankddev: nitpick: maybe borrow here to avoid cloning later? This will require changing everything back to borrowed, of course.
```suggestion
        &format!("https://{}.hexdocs.pm", config.name.replace("_", "-"))
```

ankddev (commented):
Left few thoughts, otherwise looks nice to me :+1:

test/publishing_default_readme/README.md
GearsDatapacks: Oops! Thanks

compiler-core/src/docs.rs
GearsDatapacks: Oh yes! I didn't even try that because I assumed the lifetimes wouldn't work, but seems like it only needs to survive until the end of the function which is fine.

lpil (approved):
Thank you!!!

PR #5787, opened 2026-06-01
LS: add "Discard unused variable" code action

### LINKED ISSUE #4950: LSP code action to ignore label in pattern match

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/4950
***


compiler-core/src/type_/error.rs
```diff
@@ -807,8 +807,10 @@ pub enum VariableSyntax {
     Variable(EcoString),
     /// A variable from label shorthand syntax, which can be ignored with an underscore: `label: _`
     LabelShorthand(EcoString),
-    /// A variable from an assignment pattern, which can be ignored by removing `as name`,
-    AssignmentPattern,
+    /// A variable from an assignment pattern, which can be ignored by removing `as name`.
+    /// Contained location is the location of assignment from the end of the pattern
+    /// to the end of assignment itself, so it includes the `as` keyword.
+    AssignmentPattern(SrcSpan),
```
ankddev, line 828: Hope that this approach is good

giacomocavalieri: Could you add an example to show what this covers in those different cases? Both when `as` is present and when it's not

giacomocavalieri (commented):
Looks great to me, just one minor comment about docs

compiler-core/src/type_/error.rs
ankddev: I believe that it's good now!

giacomocavalieri: Beautiful thank you!

giacomocavalieri (approved):

lpil (approved):
Oh very nice, that is super clean. Thank you very much!

PR #5788, opened 2026-06-02
Fix stale module cache detection

ORESoftware:
## Summary

Fix module cache invalidation when a source file timestamp differs from the cached timestamp but moves backwards rather than forwards.

## Root cause

The module loader only rechecked the source fingerprint when the source mtime was newer than the cached mtime. If a checkout or timestamp-preserving copy changed file contents while giving the source an older timestamp, Gleam could incorrectly reuse stale cached module metadata.

## Changes

- Recheck the source fingerprint whenever the source mtime differs from the cached mtime.
- Keep the fast cached path for matching mtimes.
- Ensure LSP incomplete modules are invalidated even when mtimes match.
- Add regression tests for backward-mtime source changes and same-mtime LSP invalidation.

## Validation

- `cargo test -p gleam-core build::module_loader::tests --quiet`
- `cargo test -p gleam-core --quiet`
- `cargo fmt --check`
- `cargo test --quiet --workspace --exclude test-output`

Full `cargo test --quiet` was also attempted, but local `test-output` tests require `bun`, which is not installed in this environment.

### DIFF
--- a/compiler-core/src/build/module_loader.rs
+++ b/compiler-core/src/build/module_loader.rs
@@ -66,15 +66,18 @@ where
             return read_source(name).map(Input::New);
         }
 
-        // If the timestamp of the source is newer than the cache entry and
-        // the hash of the source differs from the one in the cache entry,
-        // then we need to recompile.
-        if meta.mtime < source_mtime {
+        let lsp_incomplete_module =
+            self.mode == Mode::Lsp && self.incomplete_modules.contains(&name);
+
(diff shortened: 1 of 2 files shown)


giacomocavalieri:
Hello, what is the issue this is solving? Was this discussed in an issue beforehand?

lpil:
I'm not understanding the scenario here I'm afraid. How could the new time be further back in time than the previous one?

PR #5795, opened 2026-06-02
Fix duplicate let when rebinding after a directly matching case

### LINKED ISSUE #5748: Duplicate let when a variable is rebound after a directly matching case
- gleam 1.17.0-rc1
- target javascript

A variable shadowed inside a directly matching `case` branch and then rebound after the case compiles to two `let` declarations with the same name, which is a `SyntaxError`.

```gleam
pub fn go() -> Int {
  let n = 1
  case True {
    True -> {
      let n = 99
      n
    }
    False -> 0
  }
  let n = n + 5
  n
}
```

Compiles to:

```javascript
export function go() {
  let n = 1;
  let $ = true;
  let n$1 = 99;
  n$1
  let n$1 = n + 5;
  return n$1;
}
```

`SyntaxError: Identifier 'n$1' has already been declared`.

This is the same duplicate `let` family as #3379 and #5612, which were fixed for pattern bound variables. It still happens when the inner binding is a plain `let` inside the branch block. I have a fix that tracks the names already declared in a function so a leaked branch declaration isn't reused, which also lets the #5493 same-name workaround go. Happy to open a PR if that approach sounds right.

eyupcanakman:
When a variable is shadowed inside a directly matching `case` branch and then rebound after the case, the JavaScript code generator could declare the same `let` variable twice in one scope. That is a `SyntaxError`.

A directly matching branch generates its code into the enclosing scope, so a `let` inside it leaks out. The user variable counter is reset on branch exit so later references resolve to the outer binding, but a later `let` then declared the leaked name again. The generator now keeps the highest counter used for a name and does not go below it, which removes the extra check added in #5493.

Fixes #5748.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -243,3 +243,8 @@
   ```
 
   ([Gavin Morrow](https://github.com/gavinmorrow))
+
+- Fixed a bug where the JavaScript code generator could produce duplicate `let`
+  declarations when a variable was reassigned after being shadowed inside a
+  directly matching `case` branch.
+  ([Eyup Can Akman](https://github.com/eyupcanakman))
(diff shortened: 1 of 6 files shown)


lpil (approved):
Oh that is very nice!! Thank you very much!!

PR #5803, opened 2026-06-03
fix completions for types

giacomocavalieri:
I noticed that completions were a bit messed up when accepting a completion for a type after having it written in full: `set.|Set` would turn out to be `set.set.Set`. This PR fixes it!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -94,8 +94,8 @@
 
   ([Gavin Morrow](https://github.com/gavinmorrow))
 
-- Fixed a bug where the language server would insert the wrong text when
-  triggering a completion on types that are halfway complete.
+- Work around an ambiguity of the language server protocol that resulted in
+  editors like Zed inserting the wrong text when accepting type completions.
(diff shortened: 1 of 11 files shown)


CHANGELOG.md
```diff
@@ -33,3 +15,7 @@
 ### Formatter
 
 ### Bug fixes
+
+- Fixed a bug where the language server would insert the wrong text when
+  triggering a completion on types that are halfway complete.
```
lpil: This isn't what's happening! It's the client that decides how to insert the content, not the server, so it's a problem with Zed specifically. This isn't fixing a bug in the Gleam LS, it's working around ambiguity in the LSP specification and an unfortunate behaviour in Zed that doesn't match other popular implementations.

lpil (commented):
Thank you!! Left a note inline

CHANGELOG.md
giacomocavalieri: Changed it!

lpil (approved):
Thank you!!

PR #5804, opened 2026-06-03
Remove to string traits

giacomocavalieri:
This PR gets rid of some trait implementations we forgot to get rid of in the codebase in favour of regular functions

### DIFF
--- a/hexpm/src/lib.rs
+++ b/hexpm/src/lib.rs
@@ -630,10 +630,10 @@ pub fn api_add_owner_request(
 
 /// Turns an `OwnerLevel` into a string that can be used with the json payload
 /// of requests to the Hex API expecting an owner level.
-fn owner_level_to_json_string(level: OwnerLevel) -> String {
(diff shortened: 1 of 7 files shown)


hexpm/src/lib.rs
```diff
@@ -638,6 +628,15 @@ pub fn api_add_owner_request(
         .expect("add_owner_request request")
 }
 
+/// Turns an `OwnerLevel` into a string that can be used with the json payload
+/// of requests to the Hex API expecting an owner level.
+fn owner_level_to_json_string(level: OwnerLevel) -> String {
```
lpil: Return a static str please, no need to allocate strings 🙏

lpil (commented):
Brill, thank you! One tiny note inline.

hexpm/src/lib.rs
giacomocavalieri: Fixed!

lpil (approved):
Thank you!!

PR #5811, opened 2026-06-04
add tests for unreachable use statement after panic

### LINKED ISSUE #5683: Warnings for unreachable code in `use` expression aren't always accurate.
While developing a feature to suggest variables when the compiler get an unknown variable (#4810), I stumbled on what I think is a bug. Warnings for unreachable code in `use` expression do not seem do be handled correctly.

This code correctly emit a warning for unreachable code:

```gleam
fn wibble(_) {
  Nil
}

pub fn main() -> Nil {
  panic
  use <- wibble
  1
}
```
```
warning: Unreachable code
  ┌─ /path/to/project/project.gleam:7:10
  │
7 │   use <- wibble
  │          ^^^^^^

This code is unreachable because it comes after a panic.
```

But this code does not:

```gleam
pub fn main() -> Nil {
  panic
  use <- module.wibble
  1
}
```

And this code seems to emit a warning on the wrong expression:

```gleam
pub fn main() -> Nil {
  panic
  use <- fn(_) { Nil }
  1
}
```

```
warning: Unreachable code
  ┌─ /path/to/project/project.gleam:4:3
  │
4 │   1
  │   ^

This code is unreachable because it comes after a panic.
```

This is not a huge bug. I already tried to answer that [here](https://github.com/gleam-lang/gleam/pull/4810/changes#r2704404919), but I don't know if it is the right answer. The test coverage is not sufficient. The only case tested is the first one presented, it can be found [here](https://github.com/gleam-lang/gleam/blob/45290f3f32d19c38a5023803031256cbb3e4c166/compiler-core/src/type_/tests/warnings.rs#L1978-L1990).

Gleam version : 1.16.0

giacomocavalieri:
This PR fixes #5683 
Actually I can't seem to replicate the bad error message, it seems like something has changed after the issue was created. The warning seems to look good to me highlighting the use block. So I've just added some tests to make sure it remains like this!

### DIFF
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__unreachable_use_fn_statement_after_panic.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__warnings__unreachable_use_fn_statement_after_panic.snap
@@ -1,0 +1,22 @@
+---
+source: compiler-core/src/type_/tests/warnings.rs
+expression: "\npub fn main() -> Nil {\n  panic\n  use <- fn(_) { Nil }\n  1\n}\n"
+---
+----- SOURCE CODE
+
+pub fn main() -> Nil {
+  panic
+  use <- fn(_) { Nil }
+  1
+}
+
+
+----- WARNING
+warning: Unreachable code
+  ┌─ /src/warning/wrn.gleam:4:3
+  │  
+4 │ ╭   use <- fn(_) { Nil }
+5 │ │   1
+  │ ╰───^
+
+This code is unreachable because it comes after a `panic`.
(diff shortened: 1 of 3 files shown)


lpil (approved):
Thank you!!!

realraphrous:
Thank you for the tests !

I already solved that with my prior PR: [link to the solution found](https://github.com/gleam-lang/gleam/pull/4810/changes#r2704404919)

I wasn't sure it was a good idea to include this fix in this PR so i opened an issue eheh. With my PR merged, there are no more problems !

Could it be a good idea to had one more test with module @lpil @giacomocavalieri ?

```gleam
import module
pub fn main() -> Nil {
  panic
  use <- module.wibble
  1
}
```

giacomocavalieri:
Sure thing! Would you mind making a PR?

realraphrous:
Forget what I said, it's already been done [here](https://github.com/gleam-lang/gleam/blob/9bc9c7d6baebbf55cf037b2b492c7da7b168405e/compiler-core/src/type_/tests/warnings.rs#L2021-L2034).

I completely forgot what I had dev ...

giacomocavalieri:
Ahah I forget all the time! Thank you for looking into it!!

PR #5819, opened 2026-06-06
Improve the error message when a Hex release is too large to publish

### LINKED ISSUE #5744: Improve error message for a file being too large when publishing to Hex
Similar to https://github.com/gleam-lang/gleam/issues/5708 which is for making api_revert_package_response error msg nicer

api_publish_package_response currently dumps the raw json eg

hexpm1

assuming https://github.com/gleam-lang/gleam/issues/5708 gets https://github.com/gleam-lang/gleam/pull/5728 and adds hexpm::ApiError::ValidationError that same type could be used for the api_publish_package_response case StatusCode::UNPROCESSABLE_ENTITY https://github.com/gleam-lang/gleam/blob/d09b84d289458b39caa304496a02e7412643bcf0/hexpm/src/lib.rs#L542

kyo5uke:
This improves the error shown by `gleam publish` when a release fails to upload
to Hex because the tarball is too large. Previously the build tool fell through
to the generic "unexpected response" handling and printed the raw `422` JSON
body; now it reports a clear message that the release tarball is too large to be
published.

Hex returns a `422 Unprocessable Entity` validation error in this case (the
`@tarball_max_size` limit). `api_publish_package_response` already inspects the
`422` body for the `--replace` and late-modification cases, so this follows the
same style and adds one more arm: a `"too big"` substring maps to a new
`ApiError::ReleaseTooLarge` variant. The substring matches both the server's
`{"errors": {"tar": "too big"}, ...}` form and the `"file too big: <name>"`
form, and is only checked inside the `422` branch after the existing `--replace`
and modify checks, so the collision risk is low.

The pre-existing `unknown_error_422` test used a `"file too big: ..."` body to
exercise the fall-through `UnexpectedResponse` path. Since that body is now
recognised as a too-large error, its body has been changed to a different,
still-unhandled `422` validation error so it keeps testing the fall-through.

A possible follow-up (left out to keep this focused) would be a dedicated
diagnostic that mentions the size limit, modelled on `HexPublishReplaceRequired`,
rather than reusing the generic "Hex API failure" diagnostic.

Verified locally with `cargo test -p hexpm` and `cargo test -p gleam-core`,
`cargo clippy --workspace` (with `-D warnings`), and `cargo fmt --check`, all
passing.

Closes #5744

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,10 @@
   `package.hexdocs.pm` rather than `hexdocs.pm/package`.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
+- The build tool now shows a clearer error message when publishing to Hex
+  fails because the release tarball is too large.
+  ([kyo5uke](https://github.com/kyo5uke))
+
 ### Language server
 
 - The "pattern match on value" code action can now be used to pattern match on
--- a/compiler-core/src/dependency.rs
+++ b/compiler-core/src/dependency.rs
@@ -272,7 +272,8 @@ impl PackageFetchError {
             | hexpm::ApiError::IncorrectChecksum
             | hexpm::ApiError::Forbidden
             | hexpm::ApiError::NotReplacing
-            | hexpm::ApiError::LateModification => Self::ApiError(api_error),
+            | hexpm::ApiError::LateModification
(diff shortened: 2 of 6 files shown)


lpil (commented):
Hello! Are you a human?

PR #5820, opened 2026-06-06
remove redundant lifetime from formatter

giacomocavalieri:
While working on the formatter I noticed that all the various functions used more lifetimes than necessary. A single lifetime is plenty enough!


lpil (approved):
Thank you!!

Issue #2251, opened 2023-06-29
Use memory arena for pretty printing algebra


lpil:
Pretty printing, algebra. Data structure is represented using a recursive type that is reference counted. Create a new version of this that uses a memory arena (likely a reference based one) and benchmark it to see if there is a notable performance improvement.

If this is determined as being significantly faster, we can convert the formatter and code generators over to use this new version. This would be quite a long manual job.

ankddev:
@lpil This issue can be closed!

giacomocavalieri:
Not yet! We haven't ported all the places using the old pretty printer yet

ankddev:
Oh, sorry, totally missed this!

PR #5824, opened 2026-06-07
Faster pretty printing

giacomocavalieri:
This PR starts addressing #2251
This introduces a `pretty_arena` module with the new pretty printing algebra based on an arena. Introduction will need to be gradual, for now the new arena is only used by the language formatter.

The results are really encouraging!
- I've measured `gleam format` to be up to 13% faster on real projects like `squirrel` and `lustre`
- The memory footprint of the formatter has been reduced too, for example with Lustre  `gleam format` goes from having a peak memory footprint of 8.4MB to 7.6MB!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -113,6 +113,11 @@
 
 ### Formatter
 
+- Performance of the formatter has been improved.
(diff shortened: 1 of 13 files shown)


lpil (approved):
Amazing!! Thank you!!

PR #5847, opened 2026-06-13
fix remove unreachable clauses

giacomocavalieri:
Hello! I noticed the "remove unreachable cases" code action wouldn't trigger on alternative patterns so I went and fixed it. It was actually a lot trickier than I anticipated! 

### DIFF
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1915,6 +1915,11 @@ impl TypedClause {
         }
     }
 
+    /// Returns an iterator over all the alternative patterns of a case clause.
+    pub fn alternatives(&self) -> impl Iterator<Item = &Vec<TypedPattern>> {
+        std::iter::once(&self.pattern).chain(self.alternative_patterns.iter())
+    }
+
     pub fn find_node(&self, byte_index: u32) -> Option<Located<'_>> {
         self.pattern
             .iter()
--- a/compiler-core/src/ast.rs
+++ b/compiler-core/src/ast.rs
@@ -1915,8 +1915,8 @@ impl TypedClause {
         }
     }
 
-    /// Returns an iterator over all the alternative patterns of a case clause.
-    pub fn alternatives(&self) -> impl Iterator<Item = &Vec<TypedPattern>> {
+    /// Returns an iterator over all the patterns of a case clause.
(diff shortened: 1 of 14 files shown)


compiler-core/src/ast.rs
```diff
@@ -1915,6 +1915,11 @@ impl TypedClause {
         }
     }
 
+    /// Returns an iterator over all the alternative patterns of a case clause.
+    pub fn alternatives(&self) -> impl Iterator<Item = &Vec<TypedPattern>> {
```
lpil: The name and doc comment say that it iterates over the alternatives, but it is iterating over all patterns and not just the alternative ones.

language-server/src/code_action.rs
```diff
@@ -10127,25 +10127,110 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveUnreachableCaseClauses<'ast> {
...
+            return;
+        }
+
+        for clause in clauses {
+            let mut all_alternatives_are_unreachable = true;
+            let mut unreachable_alternatives = vec![];
+            let mut previous_alternative_end = None;
+            let mut all_previous_alternatives_were_deleted = true;
+
+            for alternative in clause.alternatives() {
```
lpil: Not sure this should be called an alternative, as it iterates over all of them, not just the alternative ones.

language-server/src/code_action.rs
```diff
@@ -10127,25 +10127,110 @@ impl<'ast> ast::visit::Visit<'ast> for RemoveUnreachableCaseClauses<'ast> {
...
+                //   1, 2 | 1, 2 -> todo
+                // //       ^^^^ just this one should be deleted
+                // }
+                self.clauses_to_delete.extend(&unreachable_alternatives);
+            }
+        }
     }
 }
 
+fn alternative_location(alternative: &[Pattern<Arc<Type>>]) -> SrcSpan {
```
lpil: This also works with any multi-pattern rather than being specific to alternatives.

lpil (commented):
Oooh very good! Thank you. I've left some small notes inline

giacomocavalieri:
Should have addressed all the review comments!

lpil (approved):
Thank you!!!

PR #5853, opened 2026-06-15
Do not write compile escript to build directory

### LINKED ISSUE #4474: Don't put gleam@@compile.erl in the _gleam_artefacts dir

lpil:
Closes https://github.com/gleam-lang/gleam/issues/4474

### DIFF
--- a/compiler-cli/src/fs.rs
+++ b/compiler-cli/src/fs.rs
@@ -6,7 +6,7 @@ use gleam_core::{
     build::{NullTelemetry, Target},
     error::{Error, FileIoAction, FileKind, OS, ShellCommandFailureReason, parse_os},
     io::{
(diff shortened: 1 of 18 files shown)


compiler-cli/src/fs.rs
```diff
@@ -102,6 +102,15 @@ impl ProjectIO {
     pub fn boxed() -> Box<Self> {
         Box::new(Self::new())
     }
+
+    pub(crate) fn initialise_beam_compiler(&self) -> Result<(), Error> {
+        let mut guard = self
+            .beam_compiler
+            .lock()
+            .expect("could not lock beam_compiler");
+        *guard = Some(BeamCompilerInstance::new(self)?);
```
giacomocavalieri: Should this first check that `guard` is not already a `Some` value?

lpil: Sure why not

PR #5855, opened 2026-06-16
Ignore generated files on Windows

jtdowney:
I noticed while testing a branch on Windows that `cargo test` leaves files behind so a second `cargo test` would fail because `all_files_have_copyright_notice` would pick them up and fail.


lpil (approved):
Good catch, thank you

PR #5857, opened 2026-06-16
Move formatter to its own crate

giacomocavalieri:
Continuing with the splitting things into their own crate to make compilation faster.
After moving the arena to its own crate, moving the formatter is also pretty straightforward!

I noticed the `fix` module was just dead code so I removed it in the process!


lpil (approved):
Wicked! Thank you!!!

PR #5858, opened 2026-06-17
Don't rename record fields from other packages

jtdowney:
While trying out the great new features from #5533, I noticed that I was able to rename a record field defined in a dependency. This worked because the downloaded source in `build/` was edited along with my code. However it obviously breaks after a clean build.

### DIFF
--- a/language-server/src/engine.rs
+++ b/language-server/src/engine.rs
@@ -878,11 +878,16 @@ where
                 // spans the whole `label:`, so we trim it down to just the
                 // label.
                 Some(Referenced::Label {
-                    location, label, ..
-                }) => success_response(SrcSpan {
-                    start: location.start,
-                    end: location.start + label.len() as u32,
-                }),
+                    location,
+                    label,
+                    type_module,
+                    ..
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -248,3 +248,8 @@
   declarations when a variable was reassigned after being shadowed inside a
   directly matching `case` branch.
   ([Eyup Can Akman](https://github.com/eyupcanakman))
+
+- Fixed a bug where the language server would offer to rename a record field
+  whose type is defined in a dependency, editing the dependency's source in the
+  build directory.
+  ([John Downey](https://github.com/jtdowney))
```
GearsDatapacks: We don't need to add a changelog entry here because the feature hasn't been released yet so this doesn't affect any of our users.

GearsDatapacks (approved):
Thanks!

CHANGELOG.md
jtdowney: Dropped

alii:
Oh, great catch. Thank you. :pray:

lpil (approved):
Oops 😁 Thank you!

PR #5866, opened 2026-06-18
Use a struct for the erlang source code printer

giacomocavalieri:
There was a very old TODO left at the beginning of the erlang code generation module to refactor it to use a struct with methods rather than passing an env around. With this PR I've done that! There's actually a lot more that I think could be done with the code, we could try and clean up some of the old functions, use better variable names, etc etc.

For now all I've done is basically just copy and paste the functions into methods and updated the code to work with that. I think that's plenty enough as a start (and it will make it easier for me to start working on compiling to Erlang Abstract Form)


lpil (approved):
Awesome, thank you!!

PR #5868, opened 2026-06-19
Fix incorrect Javascript generated for overlapping strings

### LINKED ISSUE #5856: String prefix pattern matches with guards behave differently on JavaScript and Erlang
```gleam
pub fn classify(input: String, flag: Bool) -> String {
  case input {
    "aa" <> _rest if flag -> "wibble"
    "a" <> _rest if flag -> "wobble"
    "a" <> _rest -> "woo"
    _ -> "other"
  }
}

pub fn main() {
  assert classify("aa", False) == "woo"
}
```

throws on Javascript, but passes as expected on Erlang.

It compiles to the following JavaScript:

```js
export function classify(input, flag) {
  if (input.startsWith("aa")) {
    if (flag) {
      return "wibble";
    } else {
      let $ = input.slice(2); // this should be slice(1) i think?
      if ($.charCodeAt(0) === 97) { 
        return "woo";
      } else {
        return "other";
      }
    }
  } // ....
}
```

It's not the recent change with `charCodeAt`, it's broken in the same way on 1.16 too (btw the playground is still on 1.16).

~ 💜

jtdowney:
This ended up getting decently hairy, but I am fairly happy with where it ended up. The previous code assumed that prefixes went in one direction (shorter->longer or `wib` then `wibble`). Which makes sense in pattern matches without guards, since the shorter one wins first. However, with guards, this assumption breaks down. This PR changes it to handle the other direction (longer->shorter), including correctly binding the prefix name and the rest when control falls through to the shorter pattern.

After that, I noticed some redundant checks in some branches, like checking `startsWith("wib")` inside a branch where we already checked `startsWith("wibble")`. There was no way to slice a string in a match without issuing a check. This adds a new Pattern variant for an unconditional string slice and allows cleanups like https://github.com/gleam-lang/gleam/commit/9b2c116fa128dff1f1de4ff29e7f45904d372feb#diff-647aa27870a77b993d46306c988f30d8d3f62b143f9063375acbe88d5ab317e0.

Closes #5856 

### DIFF
--- a/compiler-core/src/exhaustiveness.rs
+++ b/compiler-core/src/exhaustiveness.rs
@@ -336,6 +336,26 @@ impl Branch {
                         None => return false,
                     },
 
+                    // A guaranteed-match slice binds the prefix name (if any) and
+                    // the rest variable (if any) to the correctly sliced string.
+                    Pattern::StringPrefixSlice {
+                        prefix,
+                        prefix_name,
+                        rest_name,
+                    } => {
+                        let prefix = prefix.clone();
+                        let prefix_name_val = std::mem::take(prefix_name);
+                        let rest_name_val = std::mem::take(rest_name);
+                        if let Some(name) = prefix_name_val {
+                            self.body.assign_literal_string(name, prefix.clone());
+                        }
+                        if let Some(name) = rest_name_val {
+                            self.body
+                                .assign_string_slice(name, check.var.clone(), prefix);
+                        }
+                        return false;
+                    }
+
                     // All other patterns are not unconditional, so we just keep them.
                     Pattern::Int { .. }
                     | Pattern::Float { .. }
@@ -396,6 +416,13 @@ pub enum BoundValue {
         bit_array: Variable,
         read_action: ReadAction,
     },
+
+    /// `let a = subject.slice(N)`
+    ///
+    StringSlice {
+        subject: Variable,
+        prefix: EcoString,
+    },
 }
 
 impl Body {
@@ -415,6 +442,11 @@ impl Body {
     fn assign_literal_string(&mut self, variable: EcoString, value: EcoString) {
         self.bindings
             .push((variable, BoundValue::LiteralString(value)));
+    }
+
+    fn assign_string_slice(&mut self, variable: EcoString, subject: Variable, prefix: EcoString) {
+        self.bindings
+            .push((variable, BoundValue::StringSlice { subject, prefix }));
     }
 
     fn assign_bit_array_slice(
@@ -477,6 +509,11 @@ pub enum Pattern {
         prefix_name: Option<EcoString>,
         rest: Id<Pattern>,
     },
+    StringPrefixSlice {
+        prefix: EcoString,
+        prefix_name: Option<EcoString>,
+        rest_name: Option<EcoString>,
+    },
     Assign {
         name: EcoString,
         pattern: Id<Pattern>,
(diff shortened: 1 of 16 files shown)


compiler-core/src/javascript/tests/strings.rs
```diff
@@ -256,3 +256,167 @@ pub fn main() {
...
+pub fn classify(input: String) -> String {
+  case input {
+    "🫥a" <> rest if rest == "x" -> rest
+    "🫥" <> rest -> rest
+    _ -> "other"
+  }
+}
+"#,
+    );
+}
```
giacomocavalieri, line 422: It would be nice if all of these tests were also written as assertions in the language tests to make sure this will compile, run, and produce the expected output

compiler-core/src/exhaustiveness.rs
```diff
@@ -2702,24 +2752,60 @@ impl<'a> Compiler<'a> {
...
+                        vec![rest0.is(*rest1)]
+                    } else {
+                        vec![rest0.is(self.string_prefix_pattern(remaining, *rest1))]
+                    }
+                } else if prefix_name.is_none() && matches!(self.pattern(*rest1), Pattern::Discard)
+                {
+                    // The prefix we've already checked is longer than the pattern's prefix, so
+                    // the pattern is guaranteed to match. As it binds nothing there's nothing
+                    // left to do.
+                    vec![]
```
giacomocavalieri, line 2790: I don't think this is right: what we know here is that the prefix0 check has succeeded and that `prefix1.strip_prefix(prefix0)` has returned `None` and there's not prefix name nor rest binding for the new check. So we might be in this case as well:
```gleam
"aaa" <> _ -> todo
// We know this has succeeded

"b" <> _ -> todo
// We are now on to this pattern
```
In this case it's not guaranteed to succeed as the comment is saying

compiler-core/src/exhaustiveness.rs
```diff
@@ -3122,28 +3230,115 @@ impl BranchSplitter {
...
 }
 
-fn ancestors_values(trie: &Trie<String, usize>, key: &str) -> impl Iterator<Item = usize> {
+/// Describes how a branch's runtime check relates to the choices discovered so
+/// far when splitting branches.
+struct CheckOverlaps {
+    /// The indices of the existing choices this branch is relevant to and so
+    /// must be added to
+    overlapping: Vec<usize>,
+    /// Whether the check also needs to become a new choice of its own
```
giacomocavalieri: I don't understand what this means, could you add an example of when this would be the case here? What does it mean for a runtime check to become a choice of its own?

giacomocavalieri (commented):
I have a couple of questions!

compiler-core/src/javascript/tests/strings.rs
jtdowney: Good call, done. I've added executable versions of all these cases to `test/language/test/language/string_pattern_matching_test.gleam` in 1bc5e2526fa4efdc739703de230357e48acc22f6.

compiler-core/src/exhaustiveness.rs
jtdowney: I took a stab at improving the comment and adding an example in https://github.com/gleam-lang/gleam/pull/5868/changes#diff-5433a92804b5dfd584f7ff2615067cceb2dd33d7ea910b3c5cae8ade594b3ca0R3321-R3328. Let me know if that helps.

A choice is a `(RuntimeCheck, branches)` pair in `BranchSplitter.choices`.

compiler-core/src/exhaustiveness.rs
jtdowney: To even get to this point, it's already been established that there is an overlap and that `prefix0.starts_with(prefix1)`. The previous code also assumed this in the `unwrap_or` part, which was the root of the bug.

I made an update to the framing comment in https://github.com/gleam-lang/gleam/pull/5868/changes#diff-5433a92804b5dfd584f7ff2615067cceb2dd33d7ea910b3c5cae8ade594b3ca0R2757-R2764 and added a test in a202a2ce8faa4fc5cac9eb15658df936250b3d08 to help clarify.

giacomocavalieri (approved):
Beautiful work! And thanks for the lovely comments, they made it a lot easier to review

lpil:
Hey @jtdowney , how come you closed this?

jtdowney:
Because I was absent-mindedly cleaning up my git branches and did this again. Sorry. I really need to get more tolerant of a messy git.

lpil (approved):
Thank you!!!!

PR #5872, opened 2026-06-21
Fix bug where singleton constants are not used in Gleam constants

### LINKED ISSUE #5871: JS codegen: module-level const builds a fieldless variant with `new None()` but omits its import
I have a `gleam` build in my path that was built from main, and noticed that a project with `modem` started failing even after a clean build. I believe it's related to the work in #5759.

```gleam
import gleam/option.{type Option, None}

pub const nothing: Option(Int) = None

pub fn get() -> Option(Int) {
  None
}

pub fn main() {
  echo nothing
}
```

This codegens

```javascript
import * as $stdlib$dict from "../gleam_stdlib/gleam/dict.mjs";
import * as $option from "../gleam_stdlib/gleam/option.mjs";
import { Option$None$const } from "../gleam_stdlib/gleam/option.mjs";
import {
  Empty as $Empty,
  NonEmpty as $NonEmpty,
  CustomType as $CustomType,
  bitArraySlice,
  bitArraySliceToInt,
  BitArray as $BitArray,
  List as $List,
  UtfCodepoint as $UtfCodepoint,
} from "./gleam.mjs";

export const nothing = /* @__PURE__ */ new None();

export function get() {
  return Option$None$const;
}

export function main() {
  return echo(nothing, undefined, "src/scratch.gleam", 10);
}

...
```

And when run I get

```
❯ gleam run -tjs
  Compiling scratch
   Compiled in 0.01s
    Running scratch.main
file:///Users/jtdowney/code/scratch/build/dev/javascript/scratch/scratch.mjs:15
export const nothing = /* @__PURE__ */ new None();
                                       ^

ReferenceError: None is not defined
    at file:///Users/jtdowney/code/scratch/build/dev/javascript/scratch/scratch.mjs:15:40
    at ModuleJob.run (node:internal/modules/esm/module_job:439:25)
    at async node:internal/modules/esm/loader:646:26
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5)

Node.js v26.3.1
```

GearsDatapacks:
Fixes #5871

I also found a somewhat related bug that I've fixed here to avoid having to make a separate issue and PR, where constants referencing variants would ignore the qualifier, which could sometimes lead to incorrect code being generated (e.g. in [this snapshot](https://github.com/gleam-lang/gleam/blob/28e51ca0cf733fd62f3aea3d00844709f049f2c6/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__consts__imported_ok.snap#L24), where the constant is incorrectly referencing the local `Ok` instead of `gleam.Ok`).

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -2281,26 +2281,25 @@ impl<'module, 'a> Generator<'module, 'a> {
                     }
                 }
 
-                // If there's no arguments and the type is a function that takes
-                // arguments then this is the constructor being referenced, not the
-                // function being called.
-                if let Some(arity) = type_.fn_arity()
-                    && arguments.is_none()
-                    && arity != 0
-                {
-                    let arity = arity as u16;
-                    return self.record_constructor(type_.clone(), None, &tag, name, arity);
+                // If there's no arguments, then this is either a constructor
+                // which takes arguments being referenced rather than called,
+                // or a variant with no fields at all.
+                if arguments.is_none() {
+                    // If the constructor is not a function that takes arguments,
+                    // it effectively has zero arity.
+                    let arity = type_.fn_arity().unwrap_or(0) as u16;
+                    return self.record_constructor(
+                        type_.clone(),
+                        module.as_ref().map(|(name, _)| name.as_str()),
+                        &tag,
+                        name,
+                        arity,
+                    );
                 }
 
-                // Otherwise we're always constructing a record! Even if there's
-                // no argument list:
-                // ```gleam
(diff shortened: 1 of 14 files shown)


compiler-core/src/javascript/expression.rs
jtdowney: Is this comment stale now?

jtdowney (approved):

compiler-core/src/javascript/expression.rs
GearsDatapacks: Oops, yes! Fixed now.

lpil (approved):
Brilliant!! Thank you!!!

PR #5874, opened 2026-06-22
change inefficient regex usage

giacomocavalieri:
Profiling the compiler on `gleam build -tjs` I noticed that `is_gleam_path` was taking an unusual big percentage of time constantly rebuilding the regex. Turns out it should have been `static` and not just a `const`!


jtdowney:
Seems odd to have identical `is_gleam_path` in both `package_loader` and `fs`. Should one of them go away and the other become `pub` and be the canonical implementation?

lpil (approved):
Thank you!!

PR #5879, opened 2026-06-23
remove redundant clones

giacomocavalieri:
This PR enables the Clippy `redundant_clone` lint and fixes it!

### DIFF
--- a/compiler-cli/src/dependencies.rs
+++ b/compiler-cli/src/dependencies.rs
@@ -90,7 +90,7 @@ pub fn tree(paths: &ProjectPaths, options: TreeOptions) -> Result<()> {
 
     // Get the manifest packages and add the root package to the vec
     let mut packages = manifest.packages.iter().cloned().collect_vec();
-    packages.append(&mut vec![root_package.clone()]);
+    packages.push(root_package);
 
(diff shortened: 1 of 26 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1880,7 +1880,7 @@ fn generalise_module_constant(
         deprecation,
         implementations,
     } = constant;
-    let type_ = type_.clone();
+    let type_ = type_;
```
ankddev: Do we need this line at all?

compiler-cli/src/dependencies.rs
```diff
@@ -90,7 +90,7 @@ pub fn tree(paths: &ProjectPaths, options: TreeOptions) -> Result<()> {
 
     // Get the manifest packages and add the root package to the vec
     let mut packages = manifest.packages.iter().cloned().collect_vec();
-    packages.append(&mut vec![root_package.clone()]);
+    packages.append(&mut vec![root_package]);
```
ankddev: Maybe just `.push()` or append to iterator directly?

ankddev (commented):
Few niches from me

giacomocavalieri:
Oops that's what I get for not checking Clippys output, thank you!!

lpil (approved):
This is so nice!

PR #5886, opened 2026-06-24
Preserve return type when extracting statements

### LINKED ISSUE #5826: Extracting non-tail statement to function returns Nil
For example, in

```gleam
fn main() {
  echo "line 2"
  echo "line 3"
  echo "line 4"
}
```

Using the extract function code action on lines 2 and/or 3 will result in

```gleam
fn function() {
  echo "line 2"
  echo "line 3"
  Nil
}
```

The final trailing Nil is unnecessary.

The correct output would be

```gleam
fn function() -> String {
  echo "line 2"
  echo "line 3"
}
```

gavinmorrow:
Fixes #5826.

### DIFF
--- a/language-server/src/tests/snapshots/gleam_language_server__tests__action__extract_function_first_few_statements_inside_of_use.snap
+++ b/language-server/src/tests/snapshots/gleam_language_server__tests__action__extract_function_first_few_statements_inside_of_use.snap
@@ -1,30 +1,30 @@
 ---
 source: language-server/src/tests/action.rs
-expression: "\npub fn wibble() {\n    use wobble <- result.map(todo)\n    echo wobble as \"1\"\n    echo wobble as \"2\"\n    echo wobble as \"3\"\n    echo wobble as \"4\"\n}\n"
+expression: "\npub fn wibble() {\n  use wobble <- result.map(todo)\n  echo wobble as \"1\"\n  echo wobble as \"2\"\n  echo wobble as \"3\"\n  echo wobble as \"4\"\n}\n"
 ---
 ----- BEFORE ACTION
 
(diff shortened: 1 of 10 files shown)


giacomocavalieri:
Hello! If this is ready for review please un-draft it

language-server/src/tests/snapshots/gleam_language_server__tests__action__extract_function_first_few_statements_inside_of_use.snap
```diff
@@ -24,8 +24,7 @@ pub fn wibble() {
     echo wobble as "4"
 }
 
-fn function(wobble: a) -> Nil {
+fn function(wobble: a) -> a {
   echo wobble as "1"
     echo wobble as "2"
```
ankddev: This identation is weird. Looks like Gleam code in these tests is idented with four spaces, could you change it to two spaces?

ankddev (commented):
Hello! Lovely work! Found bad identation in snapshots, could you fix it?

language-server/src/tests/snapshots/gleam_language_server__tests__action__extract_function_first_few_statements_inside_of_use.snap
gavinmorrow: oops, thanks!

lpil (approved):
Wonderful work!! Thank you very much!!

0xda157:
Shouldn't [the example here](https://github.com/gleam-lang/gleam/blob/main/CHANGELOG.md?plain=1#L403) be updated?

gavinmorrow:
Oh yup, you're right. #5903

PR #5893, opened 2026-06-24
remove usages of `Default::default()`

giacomocavalieri:
This PR removes all usages of `Default::default()` and adds a clippy lint to make sure we don't use it again. This way the code is more explicit!


lpil (approved):
Thank you!!! I hope this was a clippy auto fix

giacomocavalieri:
Unfortunately not! But it made for good stream content 😂

PR #5895, opened 2026-06-24
use the arena for printing documentation

giacomocavalieri:
With this PR the documentation pretty printer now uses the new pretty printing arena!

I haven't added any line to the changelog because I don't have any reliably measurable improvement to any user facing command! `gleam docs build` peak memory footprint and execution time hasn't changed in any meaningful way. The pretty printing time was already a very small fraction of the total time I guess!


lpil (approved):
Wooo!!!!

PR #5899, opened 2026-06-25
Move type printing to arena

giacomocavalieri:
With this PR types are now pretty printed using the new arena-based pretty printing.
This PR is stacked on top of #5895, so will have to wait for that one to be merged first.


lpil (approved):
Jak, you're unstoppable!

PR #5905, opened 2026-06-25
change Erlang code generation

### LINKED ISSUE #2251: Use memory arena for pretty printing algebra

### LINKED ISSUE #5935: `let assert` with an aliased string prefix pattern generates invalid Erlang
`let assert` with an aliased string prefix pattern produces invalid Erlang.

```gleam
pub fn go() {
  let assert "Hello " as greeting <> name = "Hello Mike"
  #(greeting, name)
}
```

```
$ gleam build --target erlang
src/repro.gleam:5:50: variable 'Greeting' is unbound
error: Shell command failure

```

The alias is bound after the generated `case`, but the success clause already returns it:

```erlang
go() ->
    {Name@1, Greeting@1} = case <<"Hello Mike"/utf8>> of
        <<"Hello "/utf8, Name/binary>> -> {Name, Greeting};   % `Greeting` not bound yet
        _assert_fail -> erlang:error(...)
    end,
    Greeting = <<"Hello "/utf8>>,
    {Greeting@1, Name@1}.
```

Gleam 1.17.0 and current `main`, Erlang/OTP 29, macOS.

### LINKED ISSUE #5942: Arguments of integer and float division are evaluated in the wrong order on Erlang
Right now, integer and float division evaluate their arguments in the wrong order: denominator first, and numerator second.
For example this code panics with the message "second":
```gleam
pub fn main() {
  panic as "first" / panic as "second"
}
```
I'm working on a fix!

### LINKED ISSUE #5962: Redundant warning from Erlang compiler
When compiling the following piece of code:
```gleam
pub fn main() {
  1
  2
}
```
We get a warning from the Gleam compiler about the unused `1`, but then we also get a similar warning from the Erlang compiler complaining about an unused term being constructed. Since the Gleam compiler is already reporting errors, I think we should silence these kind of warnings coming from the Erlang compiler. That can be done with the `nowarn_ignored` compile flag.

giacomocavalieri:
With this PR we're switching the Erlang code generation to the current pretty printing approach to a new code generator that is better suited to later be used to produce the binary Erlang Abstract Format!

- This PR changes the Erlang code generator to use a different interface for producing textual Erlang, that is closer reflecting the shape of the Erlang Abstract Form. This is an intermediate step that will allow us to then start generating the EAF binary format rather than textual Erlang files
- This PR closes #2251 as well! Now that the Erlang codegen doesn't need any pretty printing at all I was able to rip out the old `pretty` module!
- This PR closes #5935 too, changing how let assert expression are compiled
- This PR closes #5942 too, changing how int and float divisions are compiled
- This PR closes #5962 too, adding the `nowarn_ignored` compilation flag


erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,1052 @@
...
+        self.annotation();
+        self.etf.atom("module");
+        self.etf.atom(module_name);
+    }
+
+    fn annotation(&mut self) {
+        let list = self.etf.start_list();
+        self.etf.small_tuple(2);
+        self.etf.atom("location");
+        self.etf.usize(1);
```
lpil: A placeholder I presume?

giacomocavalieri: Yeah! Annotations are still very much a wip

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__documentation__function_with_multiline_documentation.snap
```diff
@@ -11,23 +11,13 @@ pub fn documented() { 1 }
...
+-compile([no_auto_import, nowarn_ignored, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch, inline]).
 -export([documented/0]).
 
--if(?OTP_RELEASE >= 27).
--define(MODULEDOC(Str), -moduledoc(Str)).
--define(DOC(Str), -doc(Str)).
--else.
--define(MODULEDOC(Str), -compile([])).
--define(DOC(Str), -compile([])).
--endif.
```
giacomocavalieri, line 24: Gleam officially supports OTP 28 and 29, and it no longer supports OTP27. So I thought we could make this simpler and drop the `-if` `-else` macros since both 28 and 29 support `-doc`

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__binops__float_divide_side_effecting_function_by_zero.snap
```diff
@@ -0,0 +1,17 @@
...
+----- COMPILED ERLANG
+-module(my@mod).
+-compile([no_auto_import, nowarn_ignored, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch, inline]).
+-export([main/1]).
+
+-file("project/test/my/mod.gleam", 1).
+-spec main(fun(() -> float())) -> float().
+main(A) ->
+    A(),
+    +0.0.
```
lpil, line 17: This is cool.

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__case__multiple_alternative_branches_with_same_variable_names.snap
```diff
@@ -0,0 +1,50 @@
...
+---
+----- SOURCE CODE
+
+pub fn go(x) {
+  case x {
+    1 as x | x -> {
+      let a = 1 + x
+      Nil
+    }
+    3 as x | x -> {
```
lpil, line 13: I wonder if we should use the exhaustiveness information to trim out unreachable branches on Erlang too

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__documentation__function_with_multiline_documentation.snap
```diff
@@ -11,23 +11,13 @@ pub fn documented() { 1 }
...
-
 -file("project/test/my/mod.gleam", 5).
-?DOC(
-    " Function doc!\n"
-    " Hello!!\n"
-).
 -spec documented() -> integer().
+-doc(<<" Function doc!
+ Hello!!
+"/utf8>>).
```
lpil: I think this is invalid Erlang?

lpil: Apparently not!

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__echo__echo_in_a_pipeline.snap
```diff
@@ -28,7 +27,7 @@ wibble(N) ->
 -spec main() -> list(integer()).
 main() ->
     _pipe = [1, 2, 3],
-    echo(_pipe, nil, 4),
-    wibble(_pipe).
+    _pipe@1 = echo(_pipe, nil, ~"project/test/my/mod.gleam", 4),
```
lpil, line 30: Interesting the return value is used here while it wasn't before

giacomocavalieri: Yeah the tricky bit is the step after echo is expecting to be passed a variable whose location corresponds to the one generate by the previous step (the echo one). Previously we were referencing things by name, so we just got away with it by being able to reference the one two steps above; but since now we're using the location to resolve each variable we really need that step's variable to actually be defined and available

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+}
+
+impl Drop for List {
+    fn drop(&mut self) {
+        assert!(self.used, "list not closed");
+    }
+}
+
+impl Etf {
```
lpil: Don't put other definitions between the definition for a type and its impl blocks please 🙏 Move this up to be with `Etf`.

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+
+#[cfg(test)]
+#[macro_use]
+extern crate pretty_assertions;
+
+/// A data structure used to encode values into the Erlang Term Format:
+/// https://www.erlang.org/doc/apps/erts/erl_ext_dist.html.
+///
+#[derive(Debug)]
+pub struct Etf {
```
lpil: We generally avoid abbreviation names. `erlang_term_format::Builder` perhaps?

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+            used: false,
+        }
+    }
+
+    pub fn consume(mut self) {
+        self.used = true;
+    }
+}
+
+impl Drop for List {
```
lpil: Should we include this + `used` always of just in dev+test? If we won't want it in the release build we could use `#[cfg(debug_assertions)]` on both.

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+        self.bytes.extend(bytes);
+    }
+
+    /// Pushes a single raw byte.
+    ///
+    pub fn raw_byte(&mut self, byte: u8) {
+        self.push(byte);
+    }
+
+    /// Pushes the etf of an empty list.
```
lpil: Capitalise acronyms please.

That said, probably don't need to explain that this code is related to ETF. How about this?
```suggestion
    /// Pushes an empty list.
```

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+        List::new(size_index)
+    }
+
+    pub fn end_list(&mut self, list: List, items: u32) {
+        self.nil();
+        self.bytes[list.size_index..list.size_index + 4].copy_from_slice(&items.to_be_bytes());
+        list.consume();
+    }
+
+    /// Pushes the most compact etf representation of the given atom.
```
lpil:
```suggestion
    /// Pushes the most compact representation of the given atom.
```

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+    pub fn atom(&mut self, atom: &str) {
+        if atom.len() <= 255 {
+            self.small_atom_utf8(atom);
+        } else {
+            self.atom_utf8(atom);
+        }
+    }
+
+    /// Pushes the most compact etf representation of the given bigint number.
```
lpil:
```suggestion
    /// Pushes the most compact representation of the given big int.
```

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+        if let Some(value) = value.to_u8() {
+            self.small_integer(value);
+        } else if let Some(value) = value.to_i32() {
+            self.integer(value);
+        } else {
+            self.small_big(value);
+        }
+    }
+
+    /// Pushes the most compact etf representation of the given usize number.
```
lpil:
```suggestion
    /// Pushes the most compact representation of the given usize int.
```

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    /// https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#large_tuple_ext
+    pub fn large_tuple(&mut self, arity: u32) {
+        self.push(105);
+        self.extend(arity.to_be_bytes());
+    }
+
+    /// https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#nil_ext
+    fn nil(&mut self) {
```
lpil: Why does both this an `empty_list` exist?

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#atom_utf8_ext
+    fn atom_utf8(&mut self, name: &str) {
+        self.push(118);
+        self.extend((name.len() as u16).to_be_bytes());
+        self.extend(name.bytes());
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
lpil: Maybe a comment here explaining how these assertions were made, so folks know how to verify them and how to make more in future.

erlang-term-format/src/lib.rs
```diff
@@ -0,0 +1,573 @@
...
+    }
+
+    /// Pushes the most compact etf representation of the given atom.
+    pub fn atom(&mut self, atom: &str) {
+        if atom.len() <= 255 {
+            self.small_atom_utf8(atom);
+        } else {
+            self.atom_utf8(atom);
+        }
+    }
```
lpil, line 153: Perhaps one for later, but could be good to implement [ATOM_CACHE_REF](https://www.erlang.org/doc/apps/erts/erl_ext_dist.html#atom_cache_ref). Might be able to get slightly faster ETF parsing out of `erlc` if we use it.

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+// SPDX-FileCopyrightText: 2026 The Gleam contributors
+
+use ecow::EcoString;
+use itertools::Itertools;
+use num_bigint::BigInt;
+use num_traits::Zero;
+use regex::Regex;
+use std::sync::OnceLock;
+
+macro_rules! pretty_printing_error {
```
lpil: More decriptive name for this macro and the function it calls please. The fact it's an error and it is pretty isn't the important bit.

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+///
+/// With this those bugs cannot happen, and since this is a newtype wrapper its
+/// not adding any overhead over using plain strings.
+///
+pub struct ErlangModuleName(EcoString);
+
+impl ErlangModuleName {
+    #[inline]
+    /// Creates a new erlang module name from a Gleam module name.
+    pub fn new(gleam_module_name: EcoString) -> Self {
```
lpil: This only calls replace on the input, so it doesn't need to take ecostring, but doing so means that any caller with a str or a reference to an ecostring is forced to create one, which is wasted work.

Swap this to taking a `&str`, and remove the `From` implementation to remove the implicit casting.

Alternatively could have it have that as-str trait.

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+///
+/// That's because this API has been made primarily to be able to generate
+/// the Erlang Abstract Format data structure. All the methods almost map 1:1
+/// to how Erlang constructs are represented in that format.
+///
+/// You might want to keep a reference to it open as you go over this module,
+/// it's gonna be handy:
+/// https://www.erlang.org/doc/apps/erts/absform.html
+///
+pub trait Eaf<Output> {
```
lpil: This name is an acronym, but unlike ETF it doesn't have the advantage of being widely known, and it looks rather alien to me.

It also seems a bit unclear, because it's not actually anything concretely to do with Erlang terms forms, it seems to be an interface for building Erlang code generally.

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+    /// ```
+    ///
+    fn atom(&mut self, name: &str);
+}
+
+/// A structure that implements the EAF trait but rather than producing the
+/// Erlang abstract format binary, it will produce a nice and readable Erlang
+/// source string that can be used for testing.
+#[derive(Debug)]
+pub struct PrettyEaf {
```
lpil: The name says it produces pretty Erlang term format (unclear what pretty means in this context), but it produces Erlang source.

`ErlangSourceBuilder`, perhaps?

This crate seems to be similarly misnamed if it's the home of this class and the Erlang builder interface. Perhaps `erlang-generation` or something?

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+        exported: impl IntoIterator<Item = (Name, usize)>,
+    ) {
+        // If there's no item in the iterator we don't add the attribute at all.
+        let mut exported = exported.into_iter().peekable();
+        if exported.peek().is_none() {
+            return;
+        }
+
+        self.new_top_level_form();
+        self.code.push_str(&format!(
```
lpil: Avoid using `format!` to then push to the buffer. It allocates a new buffer, writes to it, reads from it, and then deallocates it.

Instead push directly to the buffer.

erlang-abstract-format/src/lib.rs
```diff
@@ -0,0 +1,3645 @@
...
+    } else {
+        format!("{number}")
+    }
+}
+
+const INDENT: usize = 4;
+
+impl PrettyEaf {
+    fn dummy_list() -> erlang_term_format::List {
+        erlang_term_format::List::new(0)
```
lpil: This class doesn't have anything to do with ETF, but it is forced to construct "dummy" values relating to it.

The builder interface trait could parameterise this data structure, so the abstract forms builder could use the ETF list data structure, and the source builder could use `()`.

erlang-term-format/src/lib.rs
giacomocavalieri: I forgot to delete one of the two 😂

erlang-term-format/src/lib.rs
giacomocavalieri: Yeah I think we can wait on this one and see what performance is like. I'll add a comment though!

erlang-abstract-format/src/lib.rs
giacomocavalieri: Since right now all of those fields would all be unused I ended up deleting all of them! Once we implement the binary format as well we can introduce this generic

compiler-core/src/erlang.rs
```diff
@@ -3274,7 +3274,7 @@ pub fn record_definition(record_name: &str, fields: &[(&str, Arc<Type>)]) -> Str
         type_printer.type_(&mut builder, field_type);
     }
 
-    builder.end_record_attribute(record);
+    builder.end_record_attribute(());
```
giacomocavalieri: This means this won't work for any implementation of `start_record_builder` which returns something that is not the empty tuple (the binary one will have to return something)

lpil: bad clippy!

lpil (approved):
Jak this rules!!!

also wow I've never had a worse time with github's UI

PR #5953, opened 2026-07-08
Authenticate Hex package requests with HEXPM_READ_API_KEY

### LINKED ISSUE #3651: Allow gleam hex.user.auth and hex.user.deauth to set a logged in user
Otherwise many times `gleam add wibble` where wibble is different packages will result in

`the rate limit for the Hex API has been exceeded for this IP`

jtdowney:
Use the API key from the `HEXPM_READ_API_KEY` environment variable, when set, to authenticate requests to Hex during dependency resolution and download.

Closes #3651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -87,6 +87,13 @@
   informative and helpful.
   ([Moritz Böhme](https://github.com/MoritzBoehme))
 
+- The build tool can now authenticate requests to Hex with the API key from the
+  `HEXPM_READ_API_KEY` environment variable, when resolving and downloading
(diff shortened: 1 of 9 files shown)


lpil (commented):
Do you think it's worth using a different environment variable? So that it doesn't get accidentally used when doing things like `gleam publish`. Then folks could use a read-only key for this and not worry about it interfering with write commands.

jtdowney:
I think that is a good idea, something like `HEXPM_READ_API_KEY`?

lpil (approved):
Super nice, as always!! Thank you!

PR #5954, opened 2026-07-09
Make Hex owner CLI commands more consistent

lpil:
Following on from Jak's previous similar change, making the CLI more consistent.

What do you think @giacomocavalieri ?


giacomocavalieri (approved):
Looks nice to me!

PR #5964, opened 2026-07-11
Improve error message for custom type with fields but no constructor

### LINKED ISSUE #5377: Report error about wrong C-style type definition for every argument
### Problem

Currently with following code:

```gleam
pub type User {
  name: String,
  surname: String
}
```

Gleam gives this error:

```txt
error: Syntax error
  ┌─ C:\Users\user\projects\deps_test\src\deps_test.gleam:2:3
  │
2 │   name: String
  │   ^^^^ I was not expecting this

Each custom type variant must have a constructor:

pub type User {
  User(
    name: String,
  )
}
```

As you see, it says only about one field, which is wrong, I think.

### Desired behaviour

It should say about all fields in the error.

### Motivation

Firstly, this will improve DX, since user will see full error.
Secondly, this will allow implementation of #3479, which is blocked by this.

### Possible solution

I don't know, if this is possible inside parser, since it's fault tolerance skips the entire expression, I think. In case it's possible, we need to say about each field in error.

### Things to consider

0. Is it possible in parser?
1. How to display errors about each field (just one `I was not expecting this` or for each one?)?

### LINKED ISSUE #5907: Provide uppercase hint for type variant names
Type names have a hint to uppercase them:

```
5 │ type foo {
  │      ^^^ I'm expecting a type name here

Hint: Type names start with a uppercase letter, and can contain a-z, A-Z,
or 0-9.
```

It would be nice if record variants got this error in addition to their existing one. Something like this:

```
6 │   bar
  │   ^^^ I was not expecting this

Each custom type variant must have a constructor:

type Foo {
  Foo(
    bar: Type,
  )
}

Hint: Type variant names start with a uppercase letter, and can contain a-z, A-Z,
or 0-9.
```

giacomocavalieri:
This PR fixes #5907, this PR fixes #5377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -490,6 +490,10 @@
   segments when compiling a function with a JavaScript external.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where the compiler would produce a confusing error message when
+  writing a constructor with a lowercase name.
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 - A `gleam@@compile.erl` is no longer left in the build output of
   `gleam compile-package`.
   ([Louis Pilfold](https://github.com/lpil))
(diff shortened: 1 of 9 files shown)


ankddev:
This PR seems to fix https://github.com/gleam-lang/gleam/issues/5377 too!

giacomocavalieri:
Aha, nice catch!

lpil (approved):
Lovely! Thank you!!!

PR #5968, opened 2026-07-14
Clippy

giacomocavalieri:
- This PR enables the clippy lint that warns for `.push_str(&format(...))`, as suggested `write!(...)` is used instead to save an allocation
- I've also had clippy reformat the code so that `;` usage is consistent: expressions returning nothing end with a `;` but blocks like `if {} else {}` do not. That's already what we're doing in most of the codebase
- Though I've not enabled the lint to warn about that, it would be quite annoying to have CI fail because of that and have to wait loads, I think this is fine as a one-of

### DIFF
--- a/compiler-core/src/warning.rs
+++ b/compiler-core/src/warning.rs
@@ -22,6 +22,7 @@ use debug_ignore::DebugIgnore;
 use ecow::EcoString;
 use itertools::Itertools;
 use std::{
+    fmt::Write as _,
     io::Write,
     sync::{Arc, atomic::Ordering},
 };
@@ -1100,9 +1101,12 @@ Either change the pattern or use `panic` to unconditionally fail.",
                     };
                     let mut text = format!("`{name}` is not a function");
                     match arguments {
-                        0 => text.push_str(&format!(
-                            ", you can just write `{name}` instead of `{name}()`."
-                        )),
+                        0 => {
+                            let _ = write!(
+                                text,
+                                ", you can just write `{name}` instead of `{name}()`."
+                            );
+                        }
                         1 => text.push_str(
                             " and will crash before it can do anything with this argument.",
                         ),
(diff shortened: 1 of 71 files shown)


compiler-core/src/warning.rs
```diff
@@ -1100,16 +1101,19 @@ Either change the pattern or use `panic` to unconditionally fail.",
                     };
                     let mut text = format!("`{name}` is not a function");
                     match arguments {
-                        0 => text.push_str(&format!(
-                            ", you can just write `{name}` instead of `{name}()`."
-                        )),
+                        0 => {
+                            let _ = write!(
```
lpil: What's this `let _ =`?

giacomocavalieri: It's what the clippy lint suggests: https://rust-lang.github.io/rust-clippy/master/?search=push_str#format_push_string

`write!` returns a result that can't fail in this case

lpil: Adding results and error handling for cases that don't exist seems pretty bad to me, and we never discard errors with `let _ =`. If we're going to get rid of that extra format allocation I think we should do it with either some macro that doesn't return an error (I've not found one) or with `push_str`

giacomocavalieri: I find push_str really hard to read the way it forces us to split it into multiple function calls. Maybe we could write a macro ourselves wrapping the `let _ = write!(...)`?

lpil: Hiding incorrect domain modelling doesn't make it more correct, and we never throw away errors so a macro that makes makes it easy to accidentally do that is definitely not an option.

Let's use push_str, it's the fastest, most correct, and easiest to understand if more annoying to read.

lpil (approved):
Thank you!!!

lpil:
oops, there's a conflict

giacomocavalieri:
Fixed!

lpil (approved):
Thanks Jak!!

PR #5972, opened 2026-07-15
Be smarter when building documents with comments

giacomocavalieri:
By profiling a run of the compiler I noticed we spent an abnormal amount of time formatting comment strings. I realised we were wasting a lot of time building `EcoString` documents out of comments while we could have turned them into `ZeroWidthStrings`: a comment document is never broken and is always on its own on an output line, so it makes no sense to count its graphemes!

With this change we shaved off ~4/5ms from JavaScript compilation, and from formatting too!


lpil (approved):
Thank you!!!

PR #5974, opened 2026-07-15
git merge conflict indicator error message

### LINKED ISSUE #5906: Show a helpful message for git merge conflict annotations
Instead of the current generic error:

```
7 │ <<<<<<< HEAD
  │       ^ I was not expecting this

Found `<`, expected one of:
- `>>`
- a bit array segment
```

we could provide a specific message asking the user to resolve the conflict(s).

0xda157:

Fixes #5906

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -47,6 +47,7 @@ pub enum LexicalErrorType {
         token: char,
     },
     InvalidTripleEqual,
+    MergeConflictIndicator,
     /// Operators that commonly exist in procedural language, but do
     /// not exist in Gleam. e.g. `++`, `-=`, `/=`, `%=`, `+=`, `*=`
     UnsupportedProceduralOperator,
@@ -969,6 +970,23 @@ impl LexicalError {
                 vec![
                     "Gleam uses `==` to check for equality between two values.".into(),
                     "See: https://tour.gleam.run/basics/equality".into(),
+                ],
+            ),
+            LexicalErrorType::MergeConflictIndicator => (
+                "Merge conflict marker found",
+                vec![
+                    "Your code appears to have a merge conflict. Resolve it by removing
+the code you do not want to keep.
+
+If you are in a merge, the top chunk is the checked-out code, and the
+bottom chunk is the new code being merged.
+
+If you are in a rebase, the top chunk is the code being rebased onto,
+and the chunk section is the checked-out code which is being rebased.
+
+More information can be found in the official git documentation:
+https://git-scm.com/book/en/v2/Git-Tools-Advanced-Merging#_checking_out_conflicts"
+                        .into(),
                 ],
             ),
             LexicalErrorType::UnsupportedProceduralOperator => (
--- a/compiler-core/src/parse/lexer.rs
+++ b/compiler-core/src/parse/lexer.rs
@@ -196,26 +196,59 @@ where
                 self.emit(string);
             }
             '=' => {
-                let token_start = self.get_position();
+                let mut token_start = self.get_position();
                 let _ = self.next_char();
-                match self.char0 {
-                    Some('=') => {
-                        let _ = self.next_char();
-                        let token_end = self.get_position();
-                        if let Some('=') = self.char0 {
-                            return Err(LexicalError {
-                                error: LexicalErrorType::InvalidTripleEqual,
-                                location: SrcSpan {
-                                    start: token_start,
-                                    end: token_end + 1,
-                                },
-                            });
-                        }
-                        self.emit((token_start, Token::EqualEqual, token_end));
+
+                if self.char0 != Some('=') {
+                    let token_end = self.get_position();
+                    self.emit((token_start, Token::Equal, token_end));
+                    return Ok(());
+                }
+
+                let _ = self.next_char();
+
+                if self.char0 != Some('=') {
+                    self.emit((token_start, Token::EqualEqual, self.get_position()));
+                    return Ok(());
+                }
+
+                let _ = self.next_char();
+                let mut seen_equals = 3;
+
+                if self.char0 != Some('=') {
+                    return Err(LexicalError {
+                        error: LexicalErrorType::InvalidTripleEqual,
+                        location: SrcSpan {
+                            start: token_start,
+                            end: self.get_position(),
+                        },
+                    });
+                }
+
+                loop {
+                    if seen_equals >= 7 {
+                        return Err(LexicalError {
+                            error: LexicalErrorType::MergeConflictIndicator,
+                            location: SrcSpan {
+                                start: token_start,
+                                end: self.get_position(),
+                            },
+                        });
                     }
-                    _ => {
-                        let token_end = self.get_position();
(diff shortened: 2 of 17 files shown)


compiler-core/src/parse/error.rs
```diff
@@ -952,6 +953,10 @@ impl LexicalError {
                     "See: https://tour.gleam.run/basics/equality".into(),
                 ],
             ),
+            LexicalErrorType::MergeConflictIndicator => (
+                "I don't know how to handle this",
+                vec!["Hint: resolve merge conflicts".into()],
```
0xda157: I didn't really have any good ideas for the error message, suggestions would be appreciated.

compiler-core/src/parse/tests.rs
```diff
@@ -354,6 +354,162 @@ fn triple_equals_with_whitespace() {
...
+
+#[test]
+fn four_lt_not_merge_conflict() {
+    assert_error!(
+        "let wobble: Int = 32
+        <<<<",
+        ParseError {
+            error: ParseErrorType::UnexpectedEof,
+            location: SrcSpan { start: 0, end: 0 },
+        }
```
0xda157: for some reason `GtGT` provides a nice error message but `LtLt` provides this unhelpful message. any advice on this would be great.

lpil: Sorry- which inputs do you mean? What do the error look like?

0xda157: `>>>>` results in a `UnexpectedToken` error with `expected: ["An import, const, type, or function."],` but `<<<<` just gives unexpected end of file

ankddev: I think that it's because parser expects bit array  when it encounters `<<`, so when it encounters end of file in test, it emits `unexpected EOF` error

compiler-core/src/parse/lexer.rs
```diff
@@ -838,6 +795,158 @@ where
...
+            }
+            _ => {
+                let tok_end = self.get_pos();
+                self.emit((tok_start, Token::Less, tok_end));
+            }
+        }
+        Ok(())
+    }
+
+    // once we've found a `>`
```
ankddev:
```suggestion
    // Once we've found a `>`
```

compiler-core/src/parse/snapshots/gleam_core__parse__tests__five_lt_not_merge_conflict.snap
```diff
@@ -0,0 +1,18 @@
...
+----- SOURCE CODE
+let wobble: Int = 32
+<<<<<
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:6
+  │
+2 │ <<<<<
+  │      ^ I was not expecting this
```
giacomocavalieri: How is this pointing to the empty space after `<`, but says it has found `<`?

compiler-core/src/parse/snapshots/gleam_core__parse__tests__merge_conflict_equals.snap
```diff
@@ -0,0 +1,16 @@
...
+=======
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:1
+  │
+2 │ =======
+  │ ^^^^^^^ I don't know how to handle this
+
+Hint: resolve merge conflicts
```
giacomocavalieri: It says resolve merge conflicts but there's no merge conflict in the code? A series of `=====` doesn't seem like an indicator of that? What if different tools use different indicators to mark conflicts?

compiler-core/src/parse/snapshots/gleam_core__parse__tests__merge_conflict_full.snap
```diff
@@ -0,0 +1,19 @@
...
+>>>>>>> main
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:1
+  │
+1 │ <<<<<<< HEAD
+  │ ^^^^^^^ I don't know how to handle this
+
+Hint: resolve merge conflicts
```
giacomocavalieri: I don't think this is a useful or actionable hint, I'd probably get rid of it. Or maybe just add a regular message (not an hint) saying something along the lines of "This looks like the start of a merge conflict ..."

compiler-core/src/parse/snapshots/gleam_core__parse__tests__merge_conflict_equals.snap
ankddev: > It says resolve merge conflicts but there's no merge conflict in the code? A series of `=====` doesn't seem like an indicator of that?

It has been said in https://github.com/gleam-lang/gleam/issues/5906#issuecomment-4878954910 by @lpil that we should only check for presence of each marker.

> What if different tools use different indicators to mark conflicts?

Most tools I've seen only distinguish Git conflict markers. Maybe we could also add support for default and snapshot jj markers? I am not aware of other kinds of markers.

lpil (commented):
Lovely, thank you!

Could we have an error like this please:

```
error: Syntax error
  ┌─ /src/parse/error.gleam:2:1
  │
2 │ <<<<<<< HEAD
  │ ^^^^^^^ merge conflict marker found

Your code appears to have a merge conflict. Resolve it by removing
the code you do not want to keep.

If you are in a merge, the top chunk is the checked-out code, and the
bottom chunk is the new code being merged.

If you are in a rebase, the top chunk is the code being rebased onto,
and the chunk section is the checked-out code which is being rebased.

More information can be found in the official git documentation:
https://git-scm.com/book/en/v2/Git-Tools-Advanced-Merging#_checking_out_conflicts
```

0xda157:
> Could we have an error like this please:
> 
> [...]

Ooo that does look a lot better, I've implemented it.

compiler-core/src/parse/lexer.rs
```diff
@@ -194,26 +194,59 @@ where
                 self.emit(string);
             }
             '=' => {
-                let tok_start = self.get_pos();
+                let mut tok_start = self.get_pos();
```
giacomocavalieri: No abbreviations please
```suggestion
                let mut token_start = self.get_pos();
```

compiler-core/src/parse/lexer.rs
```diff
@@ -194,26 +194,59 @@ where
...
                 let _ = self.next_char();
-                match self.chr0 {
-                    Some('=') => {
+
+                if self.chr0 != Some('=') {
+                    let tok_end = self.get_pos();
```
giacomocavalieri:
```suggestion
                    let token_end = self.get_pos();
```

compiler-core/src/parse/lexer.rs
0xda157: oh sorry, I was just copying the style of the existing code. I'll update it.

giacomocavalieri: Thank you!

giacomocavalieri (approved):

lpil (approved):
Thank you!!!

PR #5981, opened 2026-07-16
Fix formatting of commented guards

giacomocavalieri:
As Hayleigh pointed out guards with comments were not formatted properly. This fixes it!


lpil (approved):
Thank you!!

Issue #5983, opened 2026-07-16
Provide helpful error messages for C-style bitwise operators


0xda157:
For example:

```
error: Syntax error
  ┌─ /private/var/folders/sj/1h85_4mj28l5r065r16wdz980000gp/T/tmp.Hh2wP1UTI9/src/demo.gleam:3:18
  │
3 │   let wobble = 1 & 2
  │                  ^ I can't figure out what to do with this character

Hint: Is it a typo?
```

Instead of just printing a generic error such as "I can't figure out what to do with this character," we could provide a more specific error message directing the user to the bit array syntax and/or the bitwise functions in `gleam/int` 

Somewhat related to #5915

lpil:
Great idea!

jackprogramsjp:
Hello, I would like to work on this issue. I'll check it out.

PR #5987, opened 2026-07-17
provide specific error message for invalid assignment shorthand

### LINKED ISSUE #5915: Provide better error messages for invalid assignment shorthands
Having better error messages for `x++`, `+=`, etc that explicit call out that these syntaxes are not supported would be great! similar to what is done for `===` and `=` without a `let`. Currently the error provided is 

```
This operator has no value on its right side
Hint: Remove it or put a value after it.
```

0xda157:

Resolves #5915

### DIFF
--- a/compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_assignment_shorthand_divide_equals.snap
+++ b/compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_assignment_shorthand_divide_equals.snap
@@ -1,0 +1,17 @@
+---
+source: compiler-core/src/parse/tests.rs
+expression: "let wobble: Int = 32\nwobble /= 1"
+---
+----- SOURCE CODE
+let wobble: Int = 32
+wobble /= 1
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:8
+  │
+2 │ wobble /= 1
+  │        ^^ Invalid assignment shorthand
+
+Gleam does not support this assignment shorthand syntax, nor mutation
+Hint: variables may be rebound with `let` and `=`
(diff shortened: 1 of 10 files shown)


lpil (commented):
Nice, thank you! Could you add a snapshot test showing what it looks like when rendered please 🙏

jackprogramsjp:
Has this been solved for https://github.com/gleam-lang/gleam/issues/5983 as well or is that issue still open for a PR?

0xda157:
> Has this been solved for #5983 as well or is that issue still open for a PR?

this pr does not address #5983, feel free to work on that issue

compiler-core/src/parse/tests.rs
```diff
@@ -354,6 +354,116 @@ fn triple_equals_with_whitespace() {
     );
 }
 
+#[test]
```
ankddev: Could you add snapshot test for all the cases?

0xda157: is that really necessary? it seems like a lot of duplicated test cases. maybe we either keep the tests as is or switch them to be snapshots instead of duplicating them?

ankddev: Most of the parser tests are not snapshot ones, so it makes sense to keep them as is

lpil: One snapshot test for each of the new errors messages please. 🙏 

We don't use example tests for error messages as the snapshot tests test more thoroughly and clearly, while also being less work to maintain (as we don't have to manually edit them when the error changes).

0xda157: oh, alright, I've updated the tests. I was just following what the triple equals test does.

giacomocavalieri (approved):

compiler-core/src/parse/snapshots/gleam_core__parse__tests__invalid_assignment_shorthand_divide_equals.snap
```diff
@@ -0,0 +1,17 @@
...
+wobble /= 1
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:2:8
+  │
+2 │ wobble /= 1
+  │        ^^ Invalid assignment shorthand
+
+Gleam does not support this assignment shorthand syntax, nor mutation
```
lpil: This sounds like we do support _some_ assignment shorthands. I'll reword this

lpil (approved):
Thank you!!! Looks great

PR #5994, opened 2026-07-19
fix compilation of negate binary operators

giacomocavalieri:
I just noticed some tests in the standard library were failing with the nightly build of the compiler, I did not compile unary negation properly!


lpil (approved):
Thank you!

PR #5996, opened 2026-07-20
Improve error message for unexpected '&'

### LINKED ISSUE #5983: Provide helpful error messages for C-style bitwise operators

jackprogramsjp:

Resolves https://github.com/gleam-lang/gleam/issues/5983

### DIFF
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -917,6 +917,14 @@ impl LexicalError {
                 "Unexpected single quote",
                 vec!["Hint: Strings are written with double quotes.".into()],
             ),
+            LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
+                "Unexpected `&`",
+                vec![
+                    "Hint: Gleam does not have a bitwise '&' operator.".into(),
+                    "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`.".into(),
+                    "See: https://hexdocs.pm/gleam_stdlib/gleam/int.html".into(),
+                ],
+            ),
             LexicalErrorType::UnrecognizedToken { .. } => (
                 "I can't figure out what to do with this character",
                 vec!["Hint: Is it a typo?".into()],
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -937,6 +937,14 @@ impl LexicalError {
                 "Unexpected single quote",
                 vec!["Hint: Strings are written with double quotes.".into()],
             ),
+            LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
+                "Unexpected `&`",
+                vec![
+                    "Hint: Gleam does not have a bitwise '&' operator.".into(),
+                    "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`.".into(),
+                    "See: https://hexdocs.pm/gleam_stdlib/gleam/int.html".into(),
+                ],
+            ),
             LexicalErrorType::UnrecognizedToken { .. } => (
                 "I can't figure out what to do with this character",
                 vec!["Hint: Is it a typo?".into()],
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -937,11 +937,12 @@ impl LexicalError {
                 "Unexpected single quote",
                 vec!["Hint: Strings are written with double quotes.".into()],
             ),
-            LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
+            LexicalErrorType::UnrecognizedToken { token } if *token == '&' => (
                 "Unexpected `&`",
                 vec![
                     "Hint: Gleam does not have a bitwise '&' operator.".into(),
-                    "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`.".into(),
+                    "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`."
+                        .into(),
                     "See: https://gleam-stdlib.hexdocs.pm/gleam/int.html#bitwise_and".into(),
                 ],
             ),
--- a/compiler-core/src/parse/error.rs
+++ b/compiler-core/src/parse/error.rs
@@ -942,7 +942,7 @@ impl LexicalError {
                 vec![
                     "Hint: Gleam does not have a bitwise '&' operator.".into(),
                     "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`.".into(),
-                    "See: https://hexdocs.pm/gleam_stdlib/gleam/int.html".into(),
+                    "See: https://gleam-stdlib.hexdocs.pm/gleam/int.html#bitwise_and".into(),
                 ],
             ),
             LexicalErrorType::UnrecognizedToken { .. } => (
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,10 @@
   in a consistent alphabetical order.
   ([Zbyněk Juřica](https://github.com/zbyju))
 
+- The compiler now gives a clearer error message when `&` is used
+  so it points to `gleam/int` bitwise functions and bit array syntax.
+  ([Jack Programs](https://github.com/jackprogramsjp))
+
 ### Build tool
 
 - The build tool now generates Hexdocs URLs using the new format of
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,10 @@
   merge conflict indicator
   ([0xda157](https://github.com/0xda157))
 
+- The compiler now gives a clearer error message when `&` is used
+  so it points to `gleam/int` bitwise functions and bit array syntax.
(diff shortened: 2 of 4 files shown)


0xda157 (commented):
Hello! this does not resolve https://github.com/gleam-lang/gleam/issues/5983, as this pr only creates an error message for `&`, while the issue applies to all bitwise operators. Also, I think we should either make the hint more general or do more syntactic analysis, as providing an error  about bitwise operators to someone trying to preform pointer operators (e.g. `let wibble = &wobble`) doesn't make much sense.

jackprogramsjp:
Thank you for clarifying this.
I didn't think through that for whatever reason. That's my fault.

I'm assuming for the `&` operation, we'd have to let it go accepted into the lexer and then based on how it's parsed, we'd reject it due to `1 & 2` being different from `&variable` for example.

I also know that it can also be used as a reference, in C++, like `int& x;` for example. But not sure if that's too out of scope here.

I'm also wondering if we should detect `&=` in the lexer.

What do you think?

0xda157:
> Thank you for clarifying this. I didn't think through that for whatever reason. That's my fault.
> 
> I'm assuming for the `&` operation, we'd have to let it go accepted into the lexer and then based on how it's parsed, we'd reject it due to `1 & 2` being different from `&variable` for example.

Yes, unless the core team is apposed to that, which cases I'd say we should just add another line to the existing message about points.

> I also know that it can also be used as a reference, in C++, like `int& x;` for example. But not sure if that's too out of scope here.

Are you asking if we should try to provide a message specifically for C style variable declaration? I'm not sure that worth the effort just for an improved error message, but ultimately it's up to the core team.    

> I'm also wondering if we should detect `&=` in the lexer.

I think that would be a good idea. Originally I was going to do that in #5987 but it seemed out of scope for that pr.

jackprogramsjp:
I'll add `&=` in the lexical analysis to reject error in this PR because it seems more commonly only used in compound operations and nothing else so, but correct me if I'm wrong.

I'm wondering if we have to ping the core team here to ask about accepting this in the lexer and checking in the parser. Or it's better to make a new issue off of this? Or if I can just straight away do it now.

0xda157:
> I'll add `&=` in the lexical analysis to reject error in this PR because it seems more commonly only used in compound operations and nothing else so, but correct me if I'm wrong.

sorry, I don't quite follow, what is reject error?

> I'm wondering if we have to ping the core team here to ask about accepting this in the lexer and checking in the parser. Or it's better to make a new issue off of this? Or if I can just straight away do it now.

I believe the core team watches this repo so they will already have been notified of the discussion in this pr. You can ping them if you'd like, but they're probably just busy working on other things at the moment.

jackprogramsjp:
> sorry, I don't quite follow, what is reject error?

I worded it bad!
So I meant to reject these specific bitwise operations in the lexer only (so the parser is not involved). And then obviously error according to their definition or use-case and provide the alternative.
* `|`
* `^`
* `~`
* `<<`
* `>>`
* `&=`
* `|=`
* `^=`
* `<<=`
* `>>=`

But the `&` operator will be rejected in the parser if allowed because it can also be used as an memory address like you said.

ankddev:
> But the `&` operator will be rejected in the parser if allowed because it can also be used as an memory address like you said.

I'm not quite sure what you mean. Gleam doesn't have syntax for memory addresses/pointers!

0xda157:
> > sorry, I don't quite follow, what is reject error?
> 
> I worded it bad! So I meant to reject these specific bitwise operations in the lexer only (so the parser is not involved). And then obviously error according to their definition or use-case and provide the alternative.
> 
> But the `&` operator will be rejected in the parser if allowed because it can also be used as an memory address like you said.

This sounds good and was what I was thinking as well.

> > But the `&` operator will be rejected in the parser if allowed because it can also be used as an memory address like you said.
> 
> I'm not quite sure what you mean. Gleam doesn't have syntax for memory addresses/pointers!

I believe they mean providing a more helpful error message. For example: `wibble & wobble` <- here we want to provide an error message about bitwise operators, but here `let wibble = &wobble` we want to provide an error message about memory addresses/pointers.

giacomocavalieri:
If we want this error message to be accurate for bitwise operators that do not exist in gleam, I think it should happen when parsing a binary expression. In that case, when parsing the operator, if we see one of the unexpected bitwise operators then we can show an error

lpil:
For operators that do not exist in Gleam today we can do it in the lexer, for the ones that do we can do it in the parser.

jackprogramsjp:
@0xda157 lpil stated that operators that don't exist today will be done in lexer, but operators do, can be done in parser. `&` is not an existing Gleam operator, so that means that we won't be able to detect how `&` will be used in the parser.

0xda157:
In that case I'm happen with the current implementation on this branch for `&`, but the others should still be implemented.

compiler-core/src/parse/error.rs
```diff
@@ -917,6 +917,14 @@ impl LexicalError {
                 "Unexpected single quote",
                 vec!["Hint: Strings are written with double quotes.".into()],
             ),
+            LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
+                "Unexpected `&`",
+                vec![
+                    "Hint: Gleam does not have a bitwise '&' operator.".into(),
+                    "Hint: Use `int.bitwise_and` from `gleam/int`, or bit array syntax `<<>>`.".into(),
+                    "See: https://hexdocs.pm/gleam_stdlib/gleam/int.html".into(),
```
0xda157:
```suggestion
                    "See: https://gleam-stdlib.hexdocs.pm/gleam/int.html#bitwise_and".into(),
```

jackprogramsjp:
Do you all think it might be better to add a general Invalid bitwise operator enum for the `pub enum LexicalErrorType`?
```rust
    InvalidBitwiseOperator {
        operator: &'static str,
    },
```

Or should I just do multiple instead for `pub enum LexicalErrorType`:
```rust
InvalidBitwiseLeftShift,
InvalidBitwiseRightShift,
```
etc.....

0xda157:
> Do you all think it might be better to add a general Invalid bitwise operator enum for the `pub enum LexicalErrorType`?
> 
> ```rust
>     InvalidBitwiseOperator {
>         operator: &'static str,
>     },
> ```

I think this is the way to go, with an added `stdlib_int_function` field to point to the suggested stdlib function (e.g. `"bitwise_and"`)

giacomocavalieri:
I think the current solution is fine for the time being. Would you mind rebasing on main? It seems like there's some failing checks

jackprogramsjp:
What do you mean? Certain bitwise operations like &= have two characters so have to detect those as well, that's why I thought of adding like InvalidBitwiseOperator that way it's more organized

lpil:
Hi @jackprogramsjp ! Could you rebase on main to remove that merge commit please 🙏

ankddev:
Looks like there's compile error ATM (maybe while conflict resolution; fields got renamed in another PR):

```
error[E0026]: variant `LexicalErrorType::UnrecognizedToken` does not have a field named `tok`
   --> compiler-core/src/parse/error.rs:940:51
    |
940 |             LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
    |                                                   ^^^
    |                                                   |
    |                                                   variant `LexicalErrorType::UnrecognizedToken` does not have this field
    |                                                   help: `LexicalErrorType::UnrecognizedToken` has a field named `token`

error[E0027]: pattern does not mention field `token`
   --> compiler-core/src/parse/error.rs:940:13
    |
940 |             LexicalErrorType::UnrecognizedToken { tok } if *tok == '&' => (
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing field `token`
    |
help: include the missing field in the pattern
    |
940 |             LexicalErrorType::UnrecognizedToken { tok, token } if *tok == '&' => (
    |                                                      +++++++
help: if you don't care about this missing field, you can explicitly ignore it
    |
940 |             LexicalErrorType::UnrecognizedToken { tok, token: _ } if *tok == '&' => (
    |                                                      ++++++++++
help: or always ignore missing fields here
    |
940 |             LexicalErrorType::UnrecognizedToken { tok, .. } if *tok == '&' => (
    |                                                      ++++

Some errors have detailed explanations: E0026, E0027.
For more information about an error, try `rustc --explain E0026`.
error: could not compile `gleam-core` (lib) due to 2 previous errors
```

jackprogramsjp:
Thanks for letting me know ankkdev. I accidentally rebased the main against my own fork repository and thought I had updated my own main. That's my fault.

I rebased it and adjusted it again. Now it should compile correctly according to the main branch. 

I'm still awaiting a response on whether we can add errors for two operators instead of one only (like `&=`, `<<`, `>>`, etc.), and hopefully we can quickly finish this!

PR #5997, opened 2026-07-20
Fix grammar in error message

spenserblack:
When one or more files already exist in the target directory for `gleam new`, the error message would be one of the below:

- File already exist...
- Files already exists...

This reverses the verb forms to be one of the below:

- File already exists...
- Files already exist...

As an English-speaker (en-US) this just naturally felt incorrect to me, but I've Googled and provided a source just in case: https://www.yourdictionary.com/articles/singular-plural-verb-chart

### DIFF
--- a/compiler-core/src/error/snapshots/gleam_core__error__tests__output_files_already_exist_singular.snap
+++ b/compiler-core/src/error/snapshots/gleam_core__error__tests__output_files_already_exist_singular.snap
@@ -1,0 +1,9 @@
+---
+source: compiler-core/src/error/tests.rs
+expression: err.pretty_string()
+---
+error: File already exists in target directory
+
+  - wibble
+
+If you want to overwrite this file, delete it and run the command again.
--- a/compiler-core/src/error/tests.rs
+++ b/compiler-core/src/error/tests.rs
@@ -54,6 +54,24 @@ fn hex_error_conversion() {
     );
 }
 
+#[test]
+fn test_output_files_already_exist() {
+    let file_names = vec![
+        ("singular", vec![Utf8PathBuf::from("wibble")]),
+        (
+            "multiple",
+            vec![Utf8PathBuf::from("wibble"), Utf8PathBuf::from("wobble")],
+        ),
+    ];
+    for (group, file_names) in file_names {
+        let err = Error::OutputFilesAlreadyExist { file_names };
+        assert_snapshot!(
+            format!("output_files_already_exist_{group}"),
+            err.pretty_string(),
+        );
+    }
+}
+
 // There are 2 separate tests for Windows and for Unix, because on Windows note
 // about inability to create symlinks without Developer Mode is shown, so
 // snapshots differ.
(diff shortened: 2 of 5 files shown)


spenserblack:
BTW, not really related, but GitHub detects and highlights `*.snap` files as *Jest* snapshot files. Considering that almost any string of text can come after the front matter in an *Insta* snapshot, perhaps `.gitattributes` should have this? :thinking:

```gitattributes
*.snap linguist-language=Text
```

compiler-core/src/error/tests.rs
```diff
@@ -54,6 +54,15 @@ fn hex_error_conversion() {
     );
 }
 
+#[test]
+fn test_output_files_already_exist() {
+    let file_names = vec![("singular", vec![Utf8PathBuf::from("foo")]), ("multiple", vec![Utf8PathBuf::from("foo"), Utf8PathBuf::from("bar")])];
```
lpil: No foo and bar in this codebase please, it's a rather dark WW2 reference.

compiler-core/src/error/tests.rs
```diff
@@ -54,6 +54,15 @@ fn hex_error_conversion() {
...
+    for (group, file_names) in file_names {
+        let err = Error::OutputFilesAlreadyExist { file_names };
+        assert_snapshot!(format!("output_files_already_exist_{group}"), err.to_diagnostics()[0].title);
+    }
+}
```
lpil, line 73: Snapshot the whole error please, it's hard to understand what it is showing without the rest of the context. 🙏

lpil (commented):
Thank you! Couple tiny notes inline

spenserblack:
Thanks for the feedback (and educating me on the origin of those words)!

I see you have a few extra words configured to be allowed by the spell checker -- I think certain words can be configured to be denied, too.

lpil:
Ooops sorry, I've been unclear! I would like to snapshot the printed error that gets shown to the user rather than the data structure that produces it. Sorry about the confusion!

spenserblack:
Ah, okay, sorry for misunderstanding! No need for you to apologize 🙂
I'll look into snapshotting the error message instead

spenserblack:
Alright, I think these snapshot tests are more representative of what the user sees. Hope this is what you were looking for :slightly_smiling_face:

compiler-core/src/error/snapshots/gleam_core__error__tests__output_files_already_exist_singular.snap
```diff
@@ -0,0 +1,8 @@
+---
+source: compiler-core/src/error/tests.rs
+expression: err.pretty_string()
+---
+error: File already exists in target directory
+
+  - wibble
+If you want to overwrite these files, delete them and run the command again.
```
lpil: Aha! It has uncovered a problem! This part is plural still, which doesn't match the changed part of the message.

Could we add an empty line between `- wibble` and this text too please 🙏

spenserblack: Oh, good catch! I missed that :sweat_smile:

lpil (approved):
Thank you!!!!

PR #6022, opened 2026-07-22
Fix panic wrapping multibyte char at width boundary

jtdowney:
Currently, this program crashes the compiler:

```gleam
@deprecated("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaé")
pub fn old() -> Int {
  1
}

pub fn main() -> Int {
  old()
}
```

This is because it attempts to split the message for wrapping in the middle of the é.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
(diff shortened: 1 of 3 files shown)


lpil (approved):
Good catch! Thank you!!!

PR #6023, opened 2026-07-22
Fix invalid BEAM code for bits + unit bit-array segments

### LINKED ISSUE #5984: Bug: bits paired with a unit multiplier generates invalid Erlang
Gleam version: 1.17.0

When writing a bit array match that combines the bits specifier with a unit multiplier, the Gleam compiler accepts the syntax but produces invalid Erlang code.

When writing this Gleam code:
```
case bits {
  <<reserved:size(8)-unit(8)-bits>> -> ..
}
```
Generated Erlang:
`<<Reserved:8/bitstring-unit:8>> -> ..`

pairing a bitstring type with a unit multiplier other than 1 is invalid.

error log shows:
`conflict in unit specification for bit field: '8' and '1'`

Because this isn't caught initially, escript fails and reports an error pointing to a different line number, which makes this bug difficult to catch.

ssenthilnathan3:
Fixes #5984

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -472,6 +472,33 @@ fn type_parameter_name(type_: &Type) -> EcoString {
     }
 }
 
+fn fold_bits_unit_value<Value>(options: &[BitArrayOption<Value>]) -> Option<u8> {
+    let mut has_bits = false;
+    let mut unit_value = None;
+    for option in options {
+        match option {
+            BitArrayOption::Bits { .. } => has_bits = true,
+            BitArrayOption::Unit { value, .. } => unit_value = Some(value),
+            BitArrayOption::Bytes { .. }
+            | BitArrayOption::Int { .. }
+            | BitArrayOption::Float { .. }
+            | BitArrayOption::Utf8 { .. }
+            | BitArrayOption::Utf16 { .. }
+            | BitArrayOption::Utf32 { .. }
+            | BitArrayOption::Utf8Codepoint { .. }
+            | BitArrayOption::Utf16Codepoint { .. }
+            | BitArrayOption::Utf32Codepoint { .. }
+            | BitArrayOption::Signed { .. }
+            | BitArrayOption::Unsigned { .. }
+            | BitArrayOption::Big { .. }
+            | BitArrayOption::Little { .. }
+            | BitArrayOption::Native { .. }
+            | BitArrayOption::Size { .. } => {}
+        }
+    }
+    if has_bits { unit_value.copied() } else { None }
+}
+
 impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
     pub fn new(
         function: &'a TypedFunction,
@@ -2349,17 +2376,7 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedConstantBitArraySegment,
     ) {
-        // Fold unit for constant bits segments
-        let mut has_bits = false;
-        let mut bits_unit_value = None;
-        for option in &segment.options {
-            match option {
-                BitArrayOption::Bits { .. } => has_bits = true,
-                BitArrayOption::Unit { value, .. } => bits_unit_value = Some(*value),
-                _ => {}
-            }
-        }
-        let bits_unit_value = if has_bits { bits_unit_value } else { None };
+        let bits_unit_value = fold_bits_unit_value(&segment.options);
 
         builder.bit_array_segment();
         self.inlined_constant(builder, &segment.value);
@@ -2806,16 +2823,7 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         } else {
             // If the bit array segment doesn't need any special handling we use the
             // regular printing functions to format its value and options.
-            let mut has_bits = false;
-            let mut bits_unit_value = None;
-            for option in &segment.options {
-                match option {
-                    BitArrayOption::Bits { .. } => has_bits = true,
-                    BitArrayOption::Unit { value, .. } => bits_unit_value = Some(*value),
-                    _ => {}
-                }
-            }
-            let bits_unit_value = if has_bits { bits_unit_value } else { None };
+            let bits_unit_value = fold_bits_unit_value(&segment.options);
             builder.bit_array_segment();
 
             // Fold unit into size for bits segments with unit
@@ -2845,9 +2853,10 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         // results in a runtime error. We can't do that in Gleam! So any
         // negative value must be turned to zero instead.
         //
-        // When `bits_unit_value` is set the unit has been folded into
-        // the size so the unit specifier can be omitted. For example,
-        // size=8 with unit=8 produces 64, emitting <<X:64/bitstring>>
+        // And when `bits_unit_value` is set the unit has been folded into
+        // the size so the unit specifier can be omitted.
+        //
+        // For example, size=8 with unit=8 produces 64, emitting <<X:64/bitstring>>
         // instead of <<X:8/bitstring-unit:8>>.
         if let Some(unit) = bits_unit_value {
             // Multiply the size by the unit
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -2341,22 +2341,38 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedConstantBitArraySegment,
     ) {
+        // fold unit for constant bits segments
+        let unit_fold = segment
+            .has_bits_option()
+            .then(|| segment.options.iter().find_map(|o| match o {
+                BitArrayOption::Unit { value, .. } => Some(*value),
+                _ => None,
+            }))
+            .flatten();
+
         builder.bit_array_segment();
         self.inlined_constant(builder, &segment.value);
-        match segment.size() {
-            Some(TypedConstant::Int { int_value, .. }) if int_value.is_negative() => {
+        match (segment.size(), unit_fold) {
+            (Some(TypedConstant::Int { int_value, .. }), _) if int_value.is_negative() => {
                 builder.int(BigInt::ZERO);
             }
-            Some(size) => self.inlined_constant(builder, size),
-            None => builder.atom("default"),
+            (Some(TypedConstant::Int { int_value, .. }), Some(unit)) => {
+                builder.int(int_value * unit);
+            }
+            (Some(size), _) => self.inlined_constant(builder, size),
+            (None, _) => builder.atom("default"),
         }
-        self.bit_array_segment_specifiers(builder, segment);
+        self.bit_array_segment_specifiers(builder, segment, unit_fold.is_some());
     }
 
+    // When fold_unit is true, the unit was already multiplied into the
+    // size (e.g. size=8, unit=8 became 64), so skip emitting the unit
+    // specifier. BEAM rejects unit with the bitstring type.
     fn bit_array_segment_specifiers<Output, Expr>(
         &self,
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a BitArraySegment<Expr, Arc<Type>>,
+        fold_unit: bool,
     ) {
         let options = segment.options.iter();
         builder.bit_array_segment_specifiers(options.filter_map(|option| match option {
@@ -2378,6 +2394,7 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
             BitArrayOption::Big { .. } => Some(BitArraySegmentSpecifier::Big),
             BitArrayOption::Little { .. } => Some(BitArraySegmentSpecifier::Little),
             BitArrayOption::Native { .. } => Some(BitArraySegmentSpecifier::Native),
+            BitArrayOption::Unit { .. } if fold_unit => None,
             BitArrayOption::Unit { value, .. } => Some(BitArraySegmentSpecifier::Unit(*value)),
             BitArrayOption::Size { .. } => None,
         }));
@@ -2776,10 +2793,19 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         } else {
             // If the bit array segment doesn't need any special handling we use the
             // regular printing functions to format its value and options.
+            let unit_fold = segment
+                .has_bits_option()
+                .then(|| segment.options.iter().find_map(|o| match o {
+                    BitArrayOption::Unit { value, .. } => Some(*value),
+                    _ => None,
+                }))
+                .flatten();
             builder.bit_array_segment();
+
+            // Folds unit for expression bits segments
             self.maybe_block_expr(builder, &segment.value);
-            self.bit_array_expression_segment_size(builder, segment);
-            self.bit_array_segment_specifiers(builder, segment);
+            self.bit_array_expression_segment_size(builder, segment, unit_fold);
+            self.bit_array_segment_specifiers(builder, segment, unit_fold.is_some());
         }
     }
 
@@ -2792,6 +2818,7 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         &mut self,
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedExprBitArraySegment,
+        unit_fold: Option<u8>,
     ) {
         let Some(size) = segment.size() else {
             builder.atom("default");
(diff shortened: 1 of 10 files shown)


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__bit_arrays__unit_with_bits_option_constant.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: compiler-core/src/erlang/tests/bit_arrays.rs
+assertion_line: 546
+expression: "\npub const bits = <<1, 2, 3>>\npub const more_bits = <<bits:bits-size(3)-unit(8)>>\n"
+---
+----- SOURCE CODE
+
+pub const bits = <<1, 2, 3>>
+pub const more_bits = <<bits:bits-size(3)-unit(8)>>
```
lpil, line 8: Could you use these constants please 🙏  Otherwise we can't see what code is generated

ssenthilnathan3:
Closing this because #6021 addresses it.

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,31 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
+                                    segment.size(),
+                                    Some(Pattern::BitArraySize(
+                                        BitArraySize::Int { .. }
+                                    ))
+                                );
+                                if is_literal_size { Some(*value) } else { None }
+                            }
+                            _ => None,
+                        })
+                    }).flatten();
```
lpil: This iterates over the segment options three times, can we do it one pass please 🙏 

Favour `match` over `matches!` and `if`.

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,31 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
             Pattern::BitArray { segments, .. } => {
                 let bit_array = builder.start_bit_array_pattern();
                 for segment in segments {
+                    // Only fold literal ints, patterns don't have runtime ops
```
lpil: What's runtime ops? No abbreviations please, it's not clear what they mean.

compiler-core/src/erlang.rs
```diff
@@ -2341,22 +2341,38 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedConstantBitArraySegment,
     ) {
+        // fold unit for constant bits segments
```
lpil: Capital letters at the start of sentences please

compiler-core/src/erlang.rs
```diff
@@ -2341,22 +2341,38 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
...
+        let unit_fold = segment
+            .has_bits_option()
+            .then(|| segment.options.iter().find_map(|o| match o {
+                BitArrayOption::Unit { value, .. } => Some(*value),
+                _ => None,
+            }))
+            .flatten();
```
lpil: Only one pass across the options please

compiler-core/src/erlang.rs
```diff
@@ -2776,10 +2793,19 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         } else {
             // If the bit array segment doesn't need any special handling we use the
             // regular printing functions to format its value and options.
+            let unit_fold = segment
+                .has_bits_option()
+                .then(|| segment.options.iter().find_map(|o| match o {
+                    BitArrayOption::Unit { value, .. } => Some(*value),
+                    _ => None,
+                }))
```
lpil: And there too! :)

compiler-core/src/erlang.rs
```diff
@@ -2792,6 +2818,7 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         &mut self,
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedExprBitArraySegment,
+        unit_fold: Option<u8>,
```
lpil: What's a unit fold?

compiler-core/src/erlang.rs
```diff
@@ -2800,8 +2827,27 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
...
         // Sizes need some care: in Erlang, having a negative segment size
         // results in a runtime error. We can't do that in Gleam! So any
-        // negative value must be turned to zero instead:
-        if let TypedExpr::Int { int_value, .. } = &size {
+        // negative value must be turned to zero instead.
+        //
+        // Also see the `unit_fold` parameter above: when set,
+        // we fold the unit into the size here.
+        if let Some(unit) = unit_fold {
+            // size * unit
```
lpil: What does this mean?

lpil (commented):
Thank you! I've left some comments and questions inline

lpil:
Hi @ssenthilnathan3, did you mean to close this?

ssenthilnathan3:
There were some conflicts with #6021. And I think that PR addressed it.

lpil:
@giacomocavalieri did you fix https://github.com/gleam-lang/gleam/issues/5984 too in your PR?

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,27 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
             Pattern::BitArray { segments, .. } => {
                 let bit_array = builder.start_bit_array_pattern();
                 for segment in segments {
+                    // Only fold unit into size for literal integer sizes,
+                    // patterns cannot use binary operators at runtime.
```
lpil: Sorry, what does this comment mean? I'm finding this code hard to understand and the comment doesn't give much context.

"int" rather than "integer" please.

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,27 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
+                    let is_literal_size = matches!(
+                        segment.size(),
+                        Some(Pattern::BitArraySize(BitArraySize::Int { .. }))
+                    );
+                    let bits_unit_value = fold_bits_unit_value(&segment.options)
+                        .filter(|_| is_literal_size);
```
lpil: Rather than iterate twice, please do this in a single pass with a single `match` expression in the loop.

compiler-core/src/erlang/pattern.rs
```diff
@@ -392,6 +408,31 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
 
+    /// For bits segments with a unit option and a literal integer size,
+    /// multiply the size by the unit so we can drop the unit specifier.
+    fn bit_array_pattern_segment_size_with_unit<Output>(
+        &mut self,
+        builder: &mut impl ErlangBuilder<Output>,
+        segment: &'a TypedPatternBitArraySegment,
+        unit: u8,
+    ) {
+        let Some(size) = segment.size() else {
```
lpil: This is iterating a third time, but the size is already in scope when this function is called, so let's use that instead. It could be passed into the other method too.

lpil (commented):
Nice, thank you!

I've left some comments inline. Could you also add a test for a non-literal size please 🙏 

```gleam
pub fn main(x) {
  let size = 8
  <<x:bits-size(size)-unit(8)>>
}
```

After that can you also add some tests that run the code to make sure it has the correct behaviour? Perhaps in this file:

https://github.com/gleam-lang/gleam/blob/97b74f828020337dd5e78a11f837d8d59f6ad9ec/test/language/test/language/bit_array_test.gleam

compiler-core/src/erlang/pattern.rs
```diff
@@ -392,6 +423,15 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
         self.bit_array_size(builder, size);
     }
 
+    fn bit_array_pattern_segment_size_with_unit<Output>(
+        &mut self,
+        builder: &mut impl ErlangBuilder<Output>,
+        int_value: &BigInt,
+        unit: u8,
+    ) {
+        builder.int(int_value * unit);
```
lpil: Inline this function please, seeing as it's only one call itself.

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,42 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
             Pattern::BitArray { segments, .. } => {
                 let bit_array = builder.start_bit_array_pattern();
                 for segment in segments {
+                    // When a bits segment has both size and unit as literal
+                    // ints, multiply them so the unit specifier can be
+                    // dropped. The BEAM rejects unit with bitstring.
+                    let int_value = match segment.size() {
```
lpil: This is iterating twice here when it calls size. Move the size handling into the loop and `match` below 🙏

lpil (commented):
Hello! It seems now there is some duplication between the code on line 240 of pattern.rs and the new `fold_bits_unit_value`. Could they be the same code? 

If `fold_bits_unit_value` is the way forward, could we give it a clearer name? It's not the fold function, so the current name seems a bit off. A name that clearly describes what it does would be best, along with a documentation comment.

compiler-core/src/erlang/pattern.rs
```diff
@@ -240,18 +240,23 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
-                        Some(Pattern::BitArraySize(BitArraySize::Int { int_value, .. })) => {
-                            Some(int_value)
-                        }
-                        _ => None,
-                    };
                     let mut has_bits = false;
                     let mut bits_unit_value = None;
+                    let mut int_value = None;
                     for option in &segment.options {
+                        #[allow(clippy::wildcard_enum_match_arm)]
```
lpil: Never disable clippy please.

compiler-core/src/erlang/pattern.rs
```diff
@@ -240,18 +240,23 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
                         match option {
                             BitArrayOption::Bits { .. } => has_bits = true,
                             BitArrayOption::Unit { value, .. } => bits_unit_value = Some(*value),
+                            BitArrayOption::Size { value, .. } => {
+                                if let Pattern::BitArraySize(BitArraySize::Int {
```
lpil: Inline this into the pattern above please

compiler-core/src/erlang.rs
```diff
@@ -472,6 +472,26 @@ fn type_parameter_name(type_: &Type) -> EcoString {
...
 
+/// Extract Bits, Unit and Size options from a bit array segment and
+/// Returns a tuple of (has_bits, unit_value, size_value).
+fn collect_bit_array_options<Value>(
+    options: &[BitArrayOption<Value>],
+) -> (bool, Option<u8>, Option<&Box<Value>>) {
+    let mut has_bits = false;
+    let mut unit_value = None;
+    let mut size_value = None;
+    #[allow(clippy::wildcard_enum_match_arm)]
```
lpil: Never disable clippy please.

compiler-core/src/erlang.rs
```diff
@@ -472,6 +472,26 @@ fn type_parameter_name(type_: &Type) -> EcoString {
...
+    for option in options {
+        match option {
+            BitArrayOption::Bits { .. } => has_bits = true,
+            BitArrayOption::Unit { value, .. } => unit_value = Some(*value),
+            BitArrayOption::Size { value, .. } => size_value = Some(value),
+            _ => {}
+        }
+    }
+    (has_bits, unit_value, size_value)
```
lpil: This looks like the same code as in the pattern module?

compiler-core/src/erlang.rs
```diff
@@ -472,6 +472,26 @@ fn type_parameter_name(type_: &Type) -> EcoString {
     }
 }
 
+/// Extract Bits, Unit and Size options from a bit array segment and
+/// Returns a tuple of (has_bits, unit_value, size_value).
+fn collect_bit_array_options<Value>(
+    options: &[BitArrayOption<Value>],
+) -> (bool, Option<u8>, Option<&Box<Value>>) {
```
lpil: Return `&Value` rather than `&Box<Value>`, to avoid the reference to a reference.

compiler-core/src/erlang.rs
```diff
@@ -2349,22 +2369,33 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         builder: &mut impl ErlangBuilder<Output>,
         segment: &'a TypedConstantBitArraySegment,
     ) {
+        let (has_bits, unit_value, size_value) = collect_bit_array_options(&segment.options);
+        let bits_unit_value = if has_bits { unit_value } else { None };
```
lpil: Couldn't this be in the helper function? Seems to always be done afterwards?

compiler-core/src/erlang.rs
```diff
@@ -2786,10 +2818,15 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         } else {
             // If the bit array segment doesn't need any special handling we use the
             // regular printing functions to format its value and options.
+            let (has_bits, unit_value, size_value) = collect_bit_array_options(&segment.options);
+            let bits_unit_value = if has_bits { unit_value } else { None };
+            let size = size_value.map(|v| v.as_ref());
             builder.bit_array_segment();
+
+            // Fold unit into size for bits segments with unit
```
lpil: What does this comment mean?

lpil (commented):
Thanks! Few more notes inline 🙏

ssenthilnathan3:
Sorry for the churn on this. I'll address all the review comments and get back to you. Thanks for the thorough review! 🙏🏾

compiler-core/src/erlang/tests/bit_arrays.rs
```diff
@@ -532,6 +532,60 @@ pub fn go(wibble: String) {
...
+}
+
+// https://github.com/gleam-lang/gleam/issues/5984
+#[test]
+fn unit_with_bits_option_variable_size() {
+    assert_erl!(
+        "
+pub fn main(x) {
+  let size = 8
+  <<x:bits-size(size)-unit(8)>>
```
lpil, line 566: Could we get a variable size test for patterns too please 🙏  Do we need to test `const`s too?

Rather than having size be a local variable lets take it as an argument, as a variable could be inlined in future, making it no longer variable.

compiler-core/src/erlang/pattern.rs
```diff
@@ -237,11 +237,30 @@ impl<'a, 'generator, 'module> PatternGenerator<'a, 'generator, 'module> {
...
+                    // ints, multiply them so the unit specifier can be
+                    // dropped. The BEAM rejects unit with bitstring.
+                    let (bits_unit_value, size_value) = collect_bit_array_options(&segment.options);
+                    let int_value = size_value.and_then(|v| match v {
+                        Pattern::BitArraySize(BitArraySize::Int { int_value, .. }) => {
+                            Some(int_value)
+                        }
+                        _ => None,
+                    });
+                    let bits_unit_value = bits_unit_value.filter(|_| int_value.is_some());
```
lpil: Does this line do nothing? It seems to be, but I could be misunderstanding as it's a bit cryptic.

compiler-core/src/erlang.rs
```diff
@@ -2810,10 +2830,15 @@ impl<'a, 'generator> FunctionGenerator<'a, 'generator> {
         } else {
             // If the bit array segment doesn't need any special handling we use the
             // regular printing functions to format its value and options.
+            let (bits_unit_value, size_value) = collect_bit_array_options(&segment.options);
             builder.bit_array_segment();
+
+            // Multiply the unit into the size for bits segments so the
+            // unit specifier can be dropped from the Erlang output.
+            // The BEAM rejects unit with the bitstring type.
```
lpil: I'm not sure this comment is in the right place, there's no multiplication here?

lpil (commented):
Thank you!! Few more small things inline 🙏

ankddev (commented):
Could you rebase to get rid of merge commits?

compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__bit_arrays__unit_with_bits_option_constant.snap
lpil: This has not been done! 🔴

ssenthilnathan3:
Hi @lpil thanks for the review. I have addressed the changes. Should be ready for review now!

ankddev (commented):
Looks like Clippy is failing:

```
error: wildcard match will also match any future added variants
   --> compiler-core/src/erlang/pattern.rs:286:25
    |
286 |                         _ => None,
    |                         ^ help: try: `Pattern::Int { .. } | Pattern::Float { .. } | Pattern::String { .. } | Pattern::Variable { .. } | Pattern::BitArraySize(_) | Pattern::Assign { .. } | Pattern::Discard { .. } | Pattern::List { .. } | Pattern::Constructor { .. } | Pattern::Tuple { .. } | Pattern::BitArray { .. } | Pattern::StringPrefix { .. } | Pattern::Invalid { .. }`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.98.0/index.html#wildcard_enum_match_arm
    = note: `-D clippy::wildcard-enum-match-arm` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::wildcard_enum_match_arm)]`
```

test/language/test/language/bit_array_test.gleam
```diff
@@ -219,4 +219,23 @@ pub fn float_16_subnormal_rounds_ties_to_even_test() {
 pub fn zero_size_int_segment_test() {
   assert <<0:size(0), 1:size(8)>> == <<1>>
   assert bit_array.bit_size(<<0:size(0)>>) == 0
+
```
ankddev, line 223: Looks like while conflict resolution you got missing `}`

PR #6026, opened 2026-07-22
Remove references to "foobar" in codebase

spenserblack:

Based on conversation in https://github.com/gleam-lang/gleam/pull/5997, I figured this could be helpful. This:

- Forbids `foo` and `foobar` in the cspell config and allows alternatives, which should help some contributors catch this mistake before making a PR.
- Removes references to `foo` (and the following `bar`) and `foobar` that I found with `git grep '\bfoo\b'` and `git grep '\bfoobar\b'`.

Forbidding `bar` seemed too extreme since it has way more uses than just as a placeholder word. I think that forbidding the first word in the "foo bar baz etc." sequence is enough to encourage developers to use an alternative sequence.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -27,7 +27,13 @@
     "Telem",
     "Unretire",
     "Untar",
-    "walkdir"
+    "walkdir",
+    // Alternatives to "foo bar"
+    "!foo",
+    "!foobar",
+    "wibble",
+    "wobble",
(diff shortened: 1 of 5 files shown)


0xda157 (approved):
LGTM

ankddev (approved):

lpil (approved):
Thank you!

PR #6028, opened 2026-07-23
Improve error message when reverting a non-existent Hex release

### LINKED ISSUE #5709: Error message when reverting version that doesn't exist could be improved

KevesDev:
When running gleam hex revert with an invalid version, the error message only reports Resource was not found, which doesn't provide any context about whether the package or version was missing.

To improve this, I added a new error variant in compiler-core/src/error.rs that includes the package name and version. In compiler-cli/src/hex.rs, the NotFound case from the hexpm crate is now caught and converted into this more specific error instead of returning the generic one.

I followed the existing pattern used by HexPublishAccessDenied and HexPublishReplaceRequired. The CLI layer already has the package and version information available, while the hexpm crate does not, so this context belongs at the caller level.

While investigating this, I looked into api_revert_release_response() in the hexpm crate and found that it discards the 404 response body entirely. This means there is no way to distinguish between a missing package and a missing version, which was the question raised in PR #5717. Because of this limitation, the error message reports that "version X of package Y" was not found rather than attempting to make a more specific distinction.

Closes #5709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,11 @@
   when building large projects.
   ([John Downey](https://github.com/jtdowney))
 
+- `gleam hex revert` now shows a clearer error when the specified package
+  version does not exist on Hex, including the package name and version in the
+  message rather than the generic "Resource was not found".
+  ([KevesDev](https://github.com/KevesDev))
+
 ### Language server
 
 - The language server now supports go-to-definition, find-references and rename
--- a/compiler-cli/src/hex.rs
+++ b/compiler-cli/src/hex.rs
@@ -114,7 +114,13 @@ pub fn revert(
     )
     .map_err(Error::hex)?;
     let response = runtime.block_on(http.send(request))?;
-    hexpm::api_revert_release_response(response).map_err(Error::hex)?;
+    match hexpm::api_revert_release_response(response) {
+        Ok(()) => (),
+        Err(hexpm::ApiError::NotFound) => {
+            return Err(Error::HexReleaseNotFound { package, version });
+        }
+        Err(error) => return Err(Error::hex(error)),
+    }
 
     // Done!
     println!("{package} {version} has been removed from Hex");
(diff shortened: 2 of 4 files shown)


ankddev (commented):
@KevesDev Hello! Looks like you used an LLM to fix this issue. You shouldn't use it to fix "good first issues", which are for people to learn.

KevesDev:
Hey @lpil, first contribution here! I fixed the error message for gleam hex revert when the requested version doesn't exist. earlier it just returned Resource was not found, which didn't give much context.

I also looked into the hexpm crate to check the question raised in PR #5717 about whether the API can distinguish between a missing package and a missing version. It looks like it can't, since the 404 response body is discarded, so the error now reports the package and version that couldn't be found instead.

Happy to make any changes if anything needs adjusting! This project is a pretty cool idea. :)

KevesDev:
> @KevesDev Hello! Looks like you used an LLM to fix this issue. You shouldn't use it to fix "good first issues", which are for people to learn.

Hi @ankddev, thanks for the feedback. I'm part of the CodePath AI301 capstone program, where using AI tools as a learning aid is part of the curriculum. The goal is to teach students how to contribute to open source while learning how to use these tools responsibly. I reviewed the changes in this PR and understand what was added, and even made sure to write it out clearly myself. I ran each test and I own all the code written!

I also disclosed the AI usage in the PR description. I understand if the project has different expectations around AI contributions, and I'm happy to discuss it with @lpil!

Edit: Regarding your earlier question, Yep, the error was the same in both cases. "There is no version X of the package Y on Hex." When I looked into api_revert_release_response() in the hexpm crate, I found it discards the 404 response body completely, so there's no real way to distinguish the two cases from the API response. That was actually the first thing from PR #5717 that I looked into before writing this fix, funny enough. That said, If the Hex API ever starts returning distinguishable error bodies for the two cases, the hexpm crate could be updated to expose that, and it's not a bad idea for scalability!

lpil:
Thank you, but I'm afraid we do not accept AI-authored contributions, especially not if they tagged as good-first-issue. If AI is used for them then we do not have any material with which to train human contributors, and those contributors are needed for the long-term health of the project.

PR #6039, opened 2026-07-25
Fix 16-bit encoding for -0.0 on JavaScript

### LINKED ISSUE #6036: Incorrect encoding for -0.0 for 16-bit float bit array segments on JavaScript
On JavaScript, `-0.0` encodes correctly for both 32-bit and 64-bit floats:
```gleam
// These both succeed
assert <<-0.0:32>> == <<0x80000000:32>>
assert <<-0.0:64>> == <<0x8000000000000000:64>>
```
However, for 16-bit floats, it gets encoded incorrectly. It becomes the same as just `0.0`:
```gleam
// Fails
assert <<-0.0:16>> == <<0x8000>>
```

I would expect consistent behaviours across sizes and targets here.

GearsDatapacks:
Fixes #6036 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
   bit array pattern to extract a `Float` and had a constructor called `Number`.
   ([Surya Rose](https://github.com/GearsDatapacks))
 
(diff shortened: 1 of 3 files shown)


lpil (approved):
Oh wow!!! Thank you

PR #6041, opened 2026-07-25
fix crash in ls

### LINKED ISSUE #6038: LS crash with v1.18.0-rc1
While extending a file, say:

```gleam
pub fn main() {
  1
}
```

And I start typing

```gleam
fn wibble() {
  let wobble =
}
```

Then the language server seems to crash because it has encountered an error in the new lines I typed.

```
error: Fatal compiler bug!

Panic: language-server/src/code_action.rs:12997
	Line number should be valid
Gleam version: 1.18.0-rc1
Operating system: macos

variable set.
```

giacomocavalieri:
This fixes #6038 
I couldn't find a way to write tests for this, since it relies on the state of previous compilations to trigger.

I've also added a changelog entry for a fix I forgot to include in the changelog in a previous PR

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,14 @@
   were ignored.
   ([Francesco Cappetti](https://github.com/frakappa))
 
+- Fixed a bug where the language server would crash on code that contains syntax
+  errors.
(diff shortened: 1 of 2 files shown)


lpil (approved):
Thank you!!

Issue #5884, opened 2026-06-23
Renaming constructors referenced in constants results in incorrect code


GearsDatapacks:
If I have some code:
```gleam
pub type Wibble {
  Wibble(Int)
}

pub const zero = Wibble(0)
```
And I rename the `Wibble` constructor to `Wobble`, the language server produces:
```gleam
pub type Wibble {
  Wobble(Int)
}

pub const zero = Wobble
```
It's replacing the entire call to `Wibble` with `Wobble`, while it should only replace the name.

lpil:
Good find, thank you

igordiogobp:
Hello, can i work on this issue?

ankddev:
@igordiogobp Sure, go for it!

khalidbelk:
Can be closed : ) ! @GearsDatapacks

PR #6044, opened 2026-07-25
fix: renaming a record constructor deletes its arguments in constants

khalidbelk:
Hey : ) ! This PR fixes issue #5884 

When renaming the `Wibble` constructor to `Wobble`, now it behaves as expected like this : 

Before renaming
```
pub type Wibble {
  Wibble(Int) // or even Wibble(wibble: Int)
}

pub const zero = Wibble(0)
```

After:
```
pub type Wibble {
  Wobble(Int)
}

pub const zero = Wobble(0)
```
whereas before it would incorrectly produce `pub const zero = Wobble`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,10 @@
   single or multiple conflicting files existed in the target directory.
   ([Spenser Black](https://github.com/spenserblack))
 
+- Fixed a bug where renaming constructors referenced in constants
+  results in incorrect code.
+  ([Khalid Belkassmi E.H.](https://github.com/khalidbelk))
+
 ## v1.18.0-rc2 - 2026-07-25
 
 ### Bug fixes
--- a/compiler-core/src/ast/constant.rs
+++ b/compiler-core/src/ast/constant.rs
@@ -42,7 +42,7 @@ pub enum Constant<T> {
 
     Record {
         location: SrcSpan,
-        constructor_location: SrcSpan,
+        arguments_start_position: u32,
         module: Option<(EcoString, SrcSpan)>,
         name: EcoString,
         /// These are the arguments used when calling the record.
--- a/compiler-core/src/ast/constant.rs
+++ b/compiler-core/src/ast/constant.rs
@@ -42,6 +42,7 @@ pub enum Constant<T> {
 
     Record {
         location: SrcSpan,
+        constructor_location: SrcSpan,
         module: Option<(EcoString, SrcSpan)>,
         name: EcoString,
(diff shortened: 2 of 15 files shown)


compiler-core/src/ast/constant.rs
```diff
@@ -42,6 +42,7 @@ pub enum Constant<T> {
 
     Record {
         location: SrcSpan,
+        constructor_location: SrcSpan,
```
lpil: The first position of this is always the same as the first position of `location`, so it is redundant. Replace this with an `arguments_start_position` please 🙏

lpil (commented):
Nice! Thank you!! One small note inline 🙏

compiler-core/src/ast/constant.rs
khalidbelk: Should be good now ; )  [a09a6c7](https://github.com/gleam-lang/gleam/pull/6044/commits/a09a6c7928a2c15b8880a212aa58cab5d928d4d0)

GearsDatapacks:
Can you rebase to remove the merge commits? We don't allow them in this repo.

khalidbelk:
> Can you rebase to remove the merge commits? We don't allow them in this repo.

Done ! @GearsDatapacks

CHANGELOG.md
```diff
@@ -12,6 +12,9 @@
 - Fixed a bug where the compiler would generate invalid code when using constant
   strings in bit array segments.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+- Fixed a bug where renaming constructors referenced in constants
```
ankddev, line 34: Newline between entries, please

khalidbelk: Was mistakenly removed when accepting both changes during **rebase**. It's been fixed [7735440](https://github.com/gleam-lang/gleam/pull/6044/commits/7735440128782dbe5728931f47352b7d14e976ff)

lpil (approved):
Lovely, thank you! Could you add the missing `.` in the changelog and run the formatter please 🙏 

I would do it but I can never work out how to push to your branch when it is also called `main` 😅

khalidbelk:
> Lovely, thank you! Could you add the missing `.` in the changelog and run the formatter please 🙏

Should be good ; ) !

> I would do it but I can never work out how to push to your branch when it is also called `main` 😅

Np ^^

Just rebased as well (to resolve conflicts), since the CHANGELOG file had some additions in the meantime

lpil (approved):
Thank you!!

PR #5632, opened 2026-04-24
Add path field support for git dependencies

jtdowney:
Allow specifying a subdirectory within a git
repository using an optional `path` field.

Closes #4556

- [x] The changes in this PR have been discussed beforehand in an issue
- [x] The issue for this PR has been linked
- [x] Tests have been added for new behaviour
- [x] The changelog has been updated for any user-facing changes

PR #6048, opened 2026-07-26
feat: add git dep support to gleam add

### LINKED ISSUE #5678: Add git deps with `gleam add`
As previously discussed here: #5619.

We can add a `--git` flag to `gleam add` to denote that we want to install the package from Git instead of Hex. For example:
```sh
gleam add --git https://github.com/gleam-lang/stdlib
```

The default behaviour is that the tip of the default branch is used. An optional `--ref` flag will let the user specify which branch, tag, or commit they want to use. This is, as I said in the original discussion post, essentially identical to the syntax used in `gleam.toml` to represent Git deps.
```sh
# branch
gleam add --git https://github.com/lustre-labs/ui --ref headless-redux-redux

# commit
gleam add --git https://tangled.org/haripm.com/snowball_stemmer -r a2dc4eff

# tag
gleam add --git ssh://git@git.:/some/path --ref v1.0.1
```

### Note for the future

Once support for monorepos/path field in git deps (#5632) is merged, we should also add an optional `--path` field (as suggested by @GearsDatapacks) that lets users specify where in the repository the package they want lives. Right now, with single-package repos, we can use the `gleam.toml` at the repo root to identify the package name.

> Re. monorepo support: I checked how Cargo does it again; instead of supporting a `--path` flag, you instead specify the package name right after `cargo add`, and the corresponding package is fetched from the `crates` directory. However, I think a flag is more consistent with the overall design, especially since Gleam does not have a similar established convention for monorepos (yet) (#5653).

I am starting work on this! Let me know if we want any changes to this design.

seafoamteal:
Closes #5678.

This adds new `--git`, `--ref`, and `--path` arguments to the `gleam add` subcommand, as discussed in the [linked issue](https://github.com/gleam-lang/gleam/issues/5678) and [initial discussion](https://github.com/gleam-lang/gleam/discussions/5619).
`--ref` and `--path` are both optional arguments. Without `--ref`, the dep is pinned to the tip of the default branch.

I only added one test; similar functions in the existing code are not tested either (presumably because they do network I/O). However, already-tested utility functions were reused and the bulk of the changes are in the params passed down from the command handler to the existing resolution+download code path. I might be wrong here, so I would appreciate suggestions on tests to add if required. 

I went ahead and added `--path`, but I'm also wondering—what will this look like if/when we want to add support for path deps to `gleam add`? Would we just reuse the `--path` argument and make it context-sensitive?

`gleam add --help` has been updated so it looks like this:

```
Usage:
        gleam add [OPTIONS] [PACKAGES]
        gleam add [OPTIONS] --git <URI> [--ref <REF>] [--path <PATH>]

Arguments:
  [PACKAGES]...
          The names of Hex packages to add

Options:
      --dev
          Add the packages as dev-only dependencies

  -h, --help
          Print help (see a summary with '-h')

Git Dependencies:
      --git <URI>
          Add a package from a remote git repository

      --ref <REF>
          The tag, branch, or commit to check out

      --path <PATH>
          The path inside the git repo where the package is located
```

`gleam help deps` has also been updated: 1. with information about the new behaviour and 2. with information about the `path` field in `gleam.toml` that was previously missing.

H/T @jtdowney for your fantastic work in #5632, it laid the groundwork for a lot of this!

🩷 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,9 @@
   making the cache files noticeably smaller.
   ([Andrey Kozhev](httos://github.com/ankddev))
 
+- Added support for installing Git dependencies via `gleam add`.
+  ([Hari Mohan](https://github.com/seafoamteal))
+
 ### Language server
 
 - The "Generate dynamic decoder" code action is now only offered when the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,13 @@
   making the cache files noticeably smaller.
   ([Andrey Kozhev](httos://github.com/ankddev))
 
-- Added support for installing Git dependencies via `gleam add`.
+- The `gleam add` command now supports a `--git` flag that allows the user to
+  add Git dependencies without manually editing `gleam.toml`. Optional `--ref`
+  and `--path` flags can be provided; these behave the same as `path` and `ref`
+  in the `gleam.toml` Git dependency syntax. When `--ref` is omitted,
+  `gleam add` pins the dependency to the commit at the tip of the default
+  branch; the user may opt in to tracking a branch instead by explicitly
+  providing `--ref <branch_name>`.
   ([Hari Mohan](https://github.com/seafoamteal))
 
 ### Language server
--- a/compiler-cli/src/add.rs
+++ b/compiler-cli/src/add.rs
@@ -2,6 +2,7 @@
 // SPDX-FileCopyrightText: 2021 The Gleam contributors
 
 use camino::{Utf8Path, Utf8PathBuf};
+use ecow::EcoString;
 use std::fmt::Write as _;
 
 use gleam_core::{
@@ -18,60 +19,20 @@ use crate::{
     fs,
 };
 
-pub enum PackagesToAdd {
-    Hex(Vec<String>),
-    Git {
-        repository: String,
-        ref_: Option<String>,
-        path: Option<Utf8PathBuf>,
-    },
-}
-
-pub fn command(paths: &ProjectPaths, packages_to_add: PackagesToAdd, dev: bool) -> Result<()> {
+pub fn hex_dependencies(
+    paths: &ProjectPaths,
+    packages_to_add: Vec<String>,
+    dev: bool,
+) -> Result<()> {
     let config = crate::config::root_config(paths)?;
+    if packages_to_add.iter().any(|name| name == &config.name) {
+        return Err(Error::CannotAddSelfAsDependency { name: config.name });
+    }
 
-    let new_package_requirements = match &packages_to_add {
-        PackagesToAdd::Hex(packages_to_add) => {
-            if packages_to_add.iter().any(|name| name == &config.name) {
-                return Err(Error::CannotAddSelfAsDependency { name: config.name });
-            }
-
-            let mut new_package_requirements = Vec::with_capacity(packages_to_add.len());
-            for specifier in packages_to_add {
-                new_package_requirements.push(parse_gleam_add_specifier(specifier)?);
-            }
-
-            new_package_requirements
-        }
-        PackagesToAdd::Git {
-            repository,
-            ref_,
-            path,
-        } => {
-            // When a user adds a git dependency, we do NOT know its name! The rest
-            // of this dependency resolution code path needs to know the names of the
-            // packages it's working with, and we don't want to maintain a second,
-            // near-identical copy. Instead, we fetch the repo and resolve the name
-            // of the package upfront, letting us continue as usual from here.
-            //
-            // When we get to the bit where we actually download and provide the package,
-            // we can simply reuse the clone from here. A second small request will be made
-            // to fetch the refs, but since they (almost certainly) haven't changed, no more
-            // deltas will be downloaded.
-            let (package_to_add, requirement) = dependencies::resolve_git_requirement(
-                repository,
-                ref_.as_deref(),
-                path.as_deref(),
-                paths,
-            )?;
-
-            if package_to_add == config.name {
-                return Err(Error::CannotAddSelfAsDependency { name: config.name });
-            }
-
-            vec![(package_to_add, requirement)]
-        }
-    };
+    let mut new_package_requirements = Vec::with_capacity(packages_to_add.len());
+    for specifier in packages_to_add {
+        new_package_requirements.push(parse_gleam_add_specifier(&specifier)?);
+    }
 
     // Insert the new packages into the manifest and perform dependency
     // resolution to determine suitable versions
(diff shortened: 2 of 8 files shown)


ankddev:
> I went ahead and added `--path`, but I'm also wondering—what will this look like if/when we want to add support for path deps to `gleam add`? Would we just reuse the `--path` argument and make it context-sensitive?

Yeah, I find this consistent with fields in gleam.toml.

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@
 
 ### Build tool
 
+- Added support for installing Git dependencies via `gleam add`.
```
lpil: More detail here please! So the reader can understand what this means.

compiler-core/src/error.rs
```diff
@@ -2287,15 +2283,9 @@ manifest.toml and a version range specified in gleam.toml:
...
-                path,
-                repo,
-            } => {
-                let text = format!(
-                    "The path `{path}` does not exist in the git repository `{repo}` \
-for package `{package}`."
-                );
+            Error::GitDependencyPathNotFound { path, repo } => {
+                let text =
+                    format!("The path `{path}` does not exist in the git repository `{repo}`.");
```
lpil, line 2288: No unrelated changes please, revert this.

compiler-cli/src/lib.rs
```diff
@@ -615,9 +653,68 @@ impl Command {
...
+            Self::Add {
+                packages,
+                dev,
+                git: git_uri,
+                ref_: git_ref,
+                path,
+            } => {
+                let mut cmd = Self::command();
+
+                if packages.is_some() == git_uri.is_some() {
```
lpil: This function is a router, it maps variants of `Command` onto functions. Do not put any other logic inline in this function please 🙏 

Favour using clap to express the API instead of doing further validation after parsing.

compiler-cli/src/lib.rs
```diff
@@ -416,19 +426,47 @@ pub enum Command {
...
+        gleam add [OPTIONS] [PACKAGES]
+        gleam add [OPTIONS] --git <URI> [--ref <REF>] [--path <PATH>]
+    "
+    )]
     Add {
         /// The names of Hex packages to add
-        #[arg(required = true)]
-        packages: Vec<String>,
+        #[arg()]
+        packages: Option<Vec<String>>,
```
lpil, line 450: How come this is now wrapped in an option?

compiler-cli/src/lib.rs
```diff
@@ -416,19 +426,47 @@ pub enum Command {
...
+        #[arg(long, value_name = "URI", help_heading = "Git Dependencies")]
+        git: Option<String>,
+
+        /// The tag, branch, or commit to check out
+        #[arg(long, alias = "ref", help_heading = "Git Dependencies")]
+        ref_: Option<String>,
+
+        /// The path inside the git repo where the package is located
+        #[arg(long, help_heading = "Git Dependencies")]
+        path: Option<String>,
```
lpil, line 476: Is it possible to write this as an enum with one variant for each action? I don't think clap as something like serde's `flatten`, so I'm not sure if it is supported.

compiler-cli/src/add.rs
```diff
@@ -17,16 +18,60 @@ use crate::{
     fs,
 };
 
-pub fn command(paths: &ProjectPaths, packages_to_add: Vec<String>, dev: bool) -> Result<()> {
+pub enum PackagesToAdd {
```
lpil: Use two different functions rather than one that takes an enum please 🙏

compiler-cli/src/add.rs
```diff
@@ -59,13 +103,63 @@ pub fn command(paths: &ProjectPaths, packages_to_add: Vec<String>, dev: bool) ->
...
+                    ref_: ref explicit_arg_ref,
+                    ..
+                } = packages_to_add
+                else {
+                    unreachable!(
+                        "new_package_requirements can only contain a Git requirement if packages_to_add is Git"
+                    )
+                };
+
+                let ref_: String = if let Some(ref_) = explicit_arg_ref {
```
lpil: Favour EcoString over String if possible please

compiler-cli/src/add.rs
```diff
@@ -83,20 +177,18 @@ pub fn command(paths: &ProjectPaths, packages_to_add: Vec<String>, dev: bool) ->
...
                     canonical_name
                 };
-                gleam_toml[name][&added_package] = toml_edit::value(range.clone());
+                gleam_toml[name][&added_package] = specifier.clone();
             } else {
                 if !gleam_toml.as_table().contains_key("dependencies") {
                     gleam_toml["dependencies"] = toml_edit::table();
                 }
-                gleam_toml["dependencies"][&added_package] = toml_edit::value(range.clone());
+                gleam_toml["dependencies"][&added_package] = specifier.clone();
```
lpil: How come the manifest is no longer being updated?

lpil (commented):
Thank you!! I've left a bunch of questions as I couldn't quite tell what approach has been taken here and why the changes to the surrounding code had been made.

compiler-core/src/error.rs
seafoamteal: It's not an unrelated change. I'm sorry, I should have explained my approach more clearly.
The big difference between someone doing `gleam add --git <url>` and `gleam add lustre` or manually editing `gleam.toml` is that we don't know the package name initially. To get the package name, you have to clone the repository and resolve the provided path, which might result in this `GitDependencyPathNotFound` error. I've just removed the package name from the diagnostic message because it is not always guaranteed to be there now.

compiler-cli/src/lib.rs
seafoamteal: It's not possible to derive `Args` on an `enum`. I did try making a `struct` to hold the git options, but there's [a bug](https://github.com/clap-rs/clap/issues/5092) which means that using `flatten` on an `Option` type (since we would want `Option<GitFlags` or equivalent) makes all the fields required, ignoring the `Option`. It seems like the only way to do this is just to list all the fields out and use a bunch of `conflicts_with` and `required_unless_present` to make it work.

compiler-cli/src/add.rs
seafoamteal: `dependencies::resolve_and_download` calls `dependency_manager::resolve_and_download_versions`, which writes `manifest.toml` after resolving any changes to it (line 206 in `dependency_manager.rs`). After this, the manifest does not change, so we don't need to write it again.

I suppose it is an unrelated change, so I'll put that line back in if you'd like.

compiler-cli/src/lib.rs
seafoamteal: I thought it was more idiomatic to represent a lack of Hex packages (as is the case when the user only specific a Git package) with `None` rather than an empty `Vec`.

seafoamteal:
Hi Louis, thank you for the review! I'm sorry, I realise now that I didn't explain enough in the PR description. I hope some of the decisions I made are clearer now. I'll work on addressing the comments i've 👍🏽-ed to.

seafoamteal:
I've made the changes requested:
- Split up `add::command` into two functions, one for hex deps and one for Git deps
- Updated the changelog
- Replaced the custom validation with `clap`'s validation utilities

seafoamteal:
Forgot to note this also fixes a bug where the `--tags` option is not provided to `git fetch` in `prepare_git_clone`, resulting in not all tags being fetched.

PR #6056, opened 2026-07-27
Migrate from bincode to bitcode

### LINKED ISSUE #5827: Migrate from bincode as it has been archived.
https://rustsec.org/advisories/RUSTSEC-2025-0141.html

We need to pick a replacement. Which one is best?

Unless I am misremembering @GearsDatapacks did some initial investigation into alternatives.

Thanks to our using serde for this we can switch between options relatively easily, so while we do want to pick the best option for now, we will be able to migrate again with similar ease in future.

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5827
* Note that it requires `gleam clean` if you have build cache with `bincode` build.
* I think that `bitcode` works fine here, and migration was so easy
* After local tests everything seems to work 
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,8 @@
 
 ### Build tool
 
-- Cache size has been significantly reduced.
+- The build tool now stores its build cache in a more compact binary format,
+  making the cache files noticeably smaller.
   ([Andrey Kozhev](httos://github.com/ankddev))
 
 ### Language server
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@
 
 ### Build tool
 
+- Cache size has been significantly reduced.
+  ([Andrey Kozhev](httos://github.com/ankddev))
+
 ### Language server
 
 - The "Generate dynamic decoder" code action is now only offered when the
--- a/compiler-core/Cargo.toml
+++ b/compiler-core/Cargo.toml
@@ -24,7 +24,7 @@ dirs-next = "2"
 # SPDX license parsing
 spdx = "0"
 # Binary format de-serialization
-bincode = { version = "2", features = ["alloc", "serde"] }
+bitcode = { version = "0.6", features = ["serde"] }
 # cross platform single glob and glob set matching
 globset = { version = "0", features = ["serde1"] }
 # Checksums
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -133,6 +133,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "69f7f8c3906b62b754cd5326047894316021dcfe5a194c8ea52bdd94934a3457"
 
 [[package]]
+name = "arrayvec"
+version = "0.7.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
+
+[[package]]
 name = "askama"
 version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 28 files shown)


GearsDatapacks (approved):
Wow, very simple change! I noticed all the `.cache_meta` files decreased in size; I wonder if this meaningfully affects the size of the main `.cache` binaries.

edit: Just tested on a random lustre project of mine, and I'm seeing about a 25% decrease in cache size. Not that the cache was very big anyway

ankddev:
@GearsDatapacks Yeah! From my tests on small project I use for tests, there are these changes (I know exact names don't matter here, just for instance):
* `build/dev/erlang/testgleam/_gleam_artefacts/testgleam.cache` - 6,5K -> 2,4K
* `build/dev/erlang/testgleam/_gleam_artefacts/wobble.cache` - 1,8K -> 696B
* `build/dev/erlang/gleeunit/_gleam_artefacts/gleeunit.cache` - 9,5K -> 4,1K
* `build/dev/erlang/gleam_stdlib/_gleam_artefacts/gleam@float.cache` - 25,2K -> 14,1K

GearsDatapacks:
Good to know that we're heading away from Rust-levels of cache rather than towards it :p

giacomocavalieri:
I think such a big reduction of cache size is worth of including in the user facing changes of the changelog!

lpil (commented):
Thank you!! I didn't realise it could result in smaller files. I wonder if there's a performance difference too.

Oh yes, good idea Jak. Let's have a changelog entry for this.

ankddev:
> I wonder if there's a performance difference too.

In my environment, it is a bit faster (0,1-0,2 secs)

ankddev:
@lpil @giacomocavalieri I'm not sure how to write changelog entry better, hope that it's good enough 😅 .

CHANGELOG.md
```diff
@@ -33,6 +33,9 @@
 
 ### Build tool
 
+- Cache size has been significantly reduced.
```
jtdowney: Maybe

> The build tool now stores its build cache in a more compact binary format, making the cache files noticeably smaller.

compiler-core/Cargo.toml
```diff
@@ -24,7 +24,7 @@ dirs-next = "2"
 # SPDX license parsing
 spdx = "0"
 # Binary format de-serialization
-bincode = { version = "2", features = ["alloc", "serde"] }
+bitcode = { version = "0.6", features = ["serde"] }
```
jtdowney, line 27: Given the library is < 1.0 and they could change formats with a minor release, it would be better to explicitly list the version testing compatibility when bumping it.

jtdowney (commented):
Given this will probably trip people up when they upgrade Gleam, I think it's worth adding a better error suggesting a `gleam clean` when decoding fails if they have recently upgraded.

GearsDatapacks:
> Given this will probably trip people up when they upgrade Gleam

This won't be a problem; we automatically discard cache from previous versions to avoid this. The only friction might be upgrading during nightly

ankddev:
> This won't be a problem; we automatically discard cache from previous versions to avoid this. The only friction might be upgrading during nightly

I think that it's still good to add a hint when decoding fails, because there is some chance of files corruption, and as you said, it would help when upgrading between nightly versions.

CHANGELOG.md
ankddev: Thank you, much better!

compiler-core/Cargo.toml
ankddev: > Given the library is < 1.0 and they could change formats with a minor release, it would be better to explicitly list the version testing compatibility when bumping it.

Sorry, I'm not sure what you mean with "explicitly list the version testing compatibility when bumping it" 😞

jtdowney: I meant to list out the full, exact version like `bitcode = "=0.6.9"`. However, the more I thought about it, the more I thought it might be too restrictive and probably not worth the potential future hassle. I know very little about the dependency and their history with stable formats. My hope is they wouldn't break formats with a 0.6.10 release.

ankddev: I think that this makes sense, thank you! And I'll check their release history to check how often they changed formats.

lpil:
I don't think we should be adding error messages that tell people work-arounds for bugs. It would result in people using them instead of reporting the bugs, which is already a problem without built-in encouragement.

compiler-core/Cargo.toml
lpil: This is an application and we use a lockfile, so being specific in `Cargo.toml` is redundant unless we want to prevent any future upgrades.

CHANGELOG.md
lpil: Perfect!

lpil (approved):
Thank you!!

PR #6070, opened 2026-07-29
`Convert to documentation comment` code action is not offered if there are newlines between comment and definition

### LINKED ISSUE #6052: v1.18.0: LS: `Convert to documentation comment` is offered even when there're newlines between it and definition
```gleam
const answer = 0

// A comment
// ^^^^^^^^^ There should be no code action here!

fn function() {
  todo
}
```

I think that this code action should not be offered with newlines between it and a definition. This happens even with several newlines, but doesn't happen when there's another comment between.

DanielVenable:
Fixes [https://github.com/gleam-lang/gleam/issues/6052](https://github.com/gleam-lang/gleam/issues/6052)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@
 
 # Changelog
 
+## Unreleased
+
+### Language Server
+
+- The "Convert to documentation comment" code action now will not trigger if
+  there is a newline between the comment and definition.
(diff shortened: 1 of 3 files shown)


lpil (commented):
Thank you, but this behaviour is already correct. It is a valid position for a doc comment, even if it is not the canonical Gleam code formatting style.

PR #6074, opened 2026-07-29
Clean up warnings in gleam_core::build

jtdowney:
Most of the fixes after dropping the allow directive were mechanical. Two that weren't:

- `gleam_core::build::package_compiler::Input` - the variants of the enum were wildly different in sizes; `New` was 640 bytes and `Cached` was 72 bytes according to rustc, and it suggested boxing `New` to cut that down.
- `gleam_core::build::package_compiler::CachedModule` - the `line_numbers` field was unused.
- `gleam_core::build::project_compiler::Built` - the `compiled_dependency_modules` field was unused, but the call that populated it was left for any side effects.

### DIFF
--- a/compiler-core/src/build.rs
+++ b/compiler-core/src/build.rs
@@ -1,8 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2021 The Gleam contributors
 
-#![allow(warnings)]
-
 mod elixir_libraries;
 mod module_loader;
 mod native_file_copier;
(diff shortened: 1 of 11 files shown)


compiler-core/src/build/module_loader/tests.rs
```diff
@@ -24,7 +24,7 @@ fn no_cache_present() {
     let file = GleamFile::new("/src".into(), "/src/main.gleam".into());
     let result = loader.load(file).unwrap();
 
-    assert!(result.is_new());
+    assert_matches!(result, Input::New(..));
```
lpil: Undo these changes please, we prefer the methods.

compiler-core/src/build.rs
```diff
@@ -136,24 +130,17 @@ impl Codegen {
...
-    Clone,
-    Copy,
-    PartialEq,
-    Eq,
+    Debug, Serialize, Deserialize, EnumString, VariantNames, ValueEnum, Clone, Copy, PartialEq, Eq,
 )]
 #[clap(rename_all = "lower")]
 #[serde(rename_all = "lowercase")]
 #[strum(serialize_all = "lowercase")]
+#[derive(Default)]
```
lpil: Merge this with the above `derive` attribute please

lpil (commented):
Fantastic work! Thank you! Couple tiny notes inline

compiler-core/src/build/module_loader/tests.rs
jtdowney: Restored in 53e04173d5467a7d3d23a43421f41f8e5558e337. The methods are only used in the tests, so I left them and marked them `#[cfg(test)]`.

compiler-core/src/build.rs
jtdowney: Oops, thanks. Fixed in 53e04173d5467a7d3d23a43421f41f8e5558e337.

lpil (approved):
Thank you!!!

PR #6075, opened 2026-07-29
LS: add code action to remove redundant list prepend

### LINKED ISSUE #5918: Code action to remove redundant list prepend
I often run into a situation where I have some code like this, where I'm mapping a list and adding one or more other items to it:
```gleam
[wibble, ..list.map(some_list, todo)]
```

Then, for some reason I decide to remove the prepended `wibble`:
```gleam
[..list.map(some_list, todo)]
```
This leave me with a compiler error, as prepending nothing to lists is not allowed. If the list spans multiple lines, it can be a little tedious to clean this up as you need to delete the opening bracket, the `..` and the closing bracket, all of which are on separate lines.

So it would be nice to have a code action which rewrites `[..x]` to just `x`.

0xda157:
Closes https://github.com/gleam-lang/gleam/issues/5918

### DIFF
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -242,7 +242,7 @@ pub enum Error {
         reason: UnsafeRecordUpdateReason,
     },
 
-    UnnecessarySpreadOperator {
+    RecordUpdateWithoutChanges {
         location: SrcSpan,
         arity: usize,
     },
@@ -1336,7 +1336,7 @@ impl Error {
             | Error::UnknownRecordField { location, .. }
             | Error::IncorrectArity { location, .. }
             | Error::UnsafeRecordUpdate { location, .. }
-            | Error::UnnecessarySpreadOperator { location, .. }
+            | Error::RecordUpdateWithoutChanges { location, .. }
             | Error::IncorrectTypeArity { location, .. }
             | Error::TypeUsedAsAConstructor { location, .. }
             | Error::CouldNotUnify { location, .. }
--- a/compiler-core/src/type_/pattern.rs
+++ b/compiler-core/src/type_/pattern.rs
@@ -1019,7 +1019,7 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                             // record's fields throws an error
                             if pattern_arguments.len() == field_map.arity as usize {
                                 {
-                                    self.problems.error(Error::UnnecessarySpreadOperator {
+                                    self.problems.error(Error::RecordUpdateWithoutChanges {
                                         location: spread_location,
                                         arity: field_map.arity as usize,
                                     });
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -3290,13 +3290,13 @@ being used as a type constructor."
             }
         }
 
-        TypeError::UnnecessarySpreadOperator { location, arity } => {
+        TypeError::RecordUpdateWithoutChanges { location, arity } => {
             let text = wrap_format!(
                 "This record has {arity} fields and you have already \
 assigned variables to all of them."
             );
             Diagnostic {
-                title: "Unnecessary spread operator".into(),
+                title: "Record update without changes".into(),
                 text,
                 hint: None,
                 level: Level::Error,
--- a/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unnecessary_spread_operator.snap
+++ b/compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__unnecessary_spread_operator.snap
@@ -14,7 +14,7 @@ fn main() {
 }
 
 ----- ERROR
-error: Unnecessary spread operator
+error: Record update without changes
   ┌─ /src/one/two.gleam:8:23
   │
 8 │   let Triple(a, b, c, ..) = triple


compiler-core/src/type_/snapshots/gleam_core__type___tests__redundant_list_prepend_constant.snap
```diff
@@ -0,0 +1,18 @@
...
+----- ERROR
+error: Redundant list spread
+  ┌─ /src/one/two.gleam:2:16
+  │
+2 │ const wobble = [..wibble]
+  │                ^^^^^^^^^^ This spread does nothing
+
+See: https://tour.gleam.run/basics/lists/
+
+Hint: Try prepending some elements [1, 2, ..list]
```
giacomocavalieri: I don't think this is a useful hint. I'd get rid of it, or reword it to say something like `you can write <THAT_VALUE> directly`

compiler-core/src/type_/error.rs
```diff
@@ -708,6 +708,22 @@ pub enum Error {
...
+    /// ```gleam
+    /// const wibble = [1, 2]
+    ///
+    /// const wobble = [..wibble]
+    ///              // ^^^^^^^^ This spread is redundant!
+    /// ```
+    ///
+    ListSpreadWithoutElements {
+        location: SrcSpan,
+        // We hold onto this for the code action.
```
giacomocavalieri: This should be a doc comment, also can you explain what `tail_location` is with an example? Does it cover the `..` part as well? Or is that the location of just what comes after `..`
```suggestion
        /// We hold onto this for the code action.
```

language-server/src/tests/snapshots/gleam_language_server__tests__action__remove_redundant_list_prepend_const.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: language-server/src/tests/action.rs
+expression: "const wibble = [1] const wobble = [..wibble]"
+---
+----- BEFORE ACTION
+const wibble = [1] const wobble = [..wibble]
```
giacomocavalieri: Can you write this on two lines? Otherwise it's a bit hard to review

language-server/src/tests/snapshots/gleam_language_server__tests__action__remove_redundant_list_prepend_variable.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: language-server/src/tests/action.rs
+expression: "pub fn main() { let wibble = [1] let wobble = [..wibble]}"
+---
+----- BEFORE ACTION
+pub fn main() { let wibble = [1] let wobble = [..wibble]}
```
giacomocavalieri: Same as above, please write properly formatted gleam code for these tests

language-server/src/tests/snapshots/gleam_language_server__tests__action__remove_redundant_list_prepend_variable_literal.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: language-server/src/tests/action.rs
+expression: "pub fn main() { let wibble = [..[1]] }"
+---
+----- BEFORE ACTION
+pub fn main() { let wibble = [..[1]] }
```
giacomocavalieri: Formatting!

language-server/src/tests/action.rs
```diff
@@ -15846,3 +15847,114 @@ fn convert_to_int_has_nicely_separated_digits() {
...
+#[test]
+fn no_remove_list_prepend_no_prepend() {
+    assert_no_code_actions!(
+        REMOVE_REDUNDANT_LIST_PREPEND,
+        "pub fn main {
+  let wibble = [1, 2]
+}",
+        find_position_of("[1, 2]").to_selection()
+    );
+}
```
giacomocavalieri, line 16267: I'd like to see an additional test where the list is not properly formatted, something like `[   ..    [wibble]  ]`, does the code action still behave correctly in that case?

CHANGELOG.md
```diff
@@ -13,6 +13,24 @@
 
 ### Language server
```
giacomocavalieri, line 192: I'd also add an entry to the language section explaining how analysis is now more fault tolerant when writing invalid expressions like `[..wibble]`, and it doesn't just stop at that error

0xda157: Shouldn't it go under the compiler section because it's a change to the compiler core?

GearsDatapacks: That's what jak meant by "language section", yeah

compiler-core/src/type_/expression.rs
```diff
@@ -1273,6 +1273,16 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             }
             None => None,
         };
+
+        if let Some(ref tail) = tail
```
lpil, line 1284: A comment showing code this is checking for please 🙏

compiler-core/src/type_/expression.rs
```diff
@@ -4680,6 +4690,15 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
             None
         };
 
+        if let Some(ref tail) = tail
```
lpil: A comment showing code this is checking for please 🙏

compiler-core/src/type_/pattern.rs
```diff
@@ -1019,7 +1019,7 @@ impl<'a, 'b> PatternTyper<'a, 'b> {
                             // record's fields throws an error
                             if pattern_arguments.len() == field_map.arity as usize {
                                 {
-                                    self.problems.error(Error::UnnecessarySpreadOperator {
+                                    self.problems.error(Error::UnnecessaryRecordSpreadOperator {
```
lpil: Oh dear, this was incorrect before too. Would you mind also changing this to something like `RecordUpdateWithoutChanges` or something like that 🙏

language-server/src/code_action.rs
```diff
@@ -10454,6 +10454,75 @@ impl<'a> RemoveRedundantRecordUpdate<'a> {
     }
 }
 
+pub struct RemoveRedundantListPrepend<'a> {
```
lpil, line 10479: A documentation comment explaining what this code action does please 🙏

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Compiler
 
+<<<<<<< HEAD
```
lpil: Merge marker here!

lpil (commented):
There's no such thing as a "spread" in Gleam, it's call a list prepend or prepending to a list. Could you remove the term "spread" please 🙏

0xda157:
> There's no such thing as a "spread" in Gleam, it's call a list prepend or prepending to a list. Could you remove the term "spread" please 🙏

I believe I've removed all the references to spreading a list from this PR, but there are a lot more [references to spreading a list/record in the codebase](https://github.com/search?q=repo%3Agleam-lang%2Fgleam+spread&type=code). I'd rather not deal with them as I'm not sure what the best phrasing is in most cases.

compiler-core/src/type_/snapshots/gleam_core__type___tests__redundant_list_prepend_constant.snap
```diff
@@ -0,0 +1,18 @@
...
+----- ERROR
+error: Redundant list prepend
+  ┌─ /src/one/two.gleam:2:16
+  │
+2 │ const wobble = [..wibble]
+  │                ^^^^^^^^^^ This prepend does nothing
+
+See: https://tour.gleam.run/basics/lists/
+
+Hint: `list` can be used directly instead of `[..list]`
```
giacomocavalieri: Could this show the value being written by the user rather than a generic `list` variable? So here it would be `wibble` can be used directly instead of `[..wibble]`.

CHANGELOG.md
```diff
@@ -28,6 +28,10 @@
   to optimise.
   ([John Downey](https://github.com/jtdowney))
 
+- The compiler is now fault tolerant when provider an error for a redundant list
```
```suggestion
- The compiler is now fault tolerant when providing an error for a redundant list
```

compiler-core/src/type_/snapshots/gleam_core__type___tests__redundant_list_prepend_constant.snap
0xda157: I'm not sure how to go about this because the tail can be any `TypedExpr`, do you have any points?

giacomocavalieri: I think it would amount to taking a the slice of the module's code string corresponding to the expressions span. I'm pretty sure there's other places where we do that but I can't seem to find them... maybe it's in the language server

0xda157: I'm working on implementing this but I'm kinda stuck. When creating the error we can the current module as an `EcoString` from `self.enviroment.current_module`, but then we need to slice that by our `SrcSpan`, which causes problems because `SliceIndex<str>` is not implemented for `Range<u32>` (only `Range<i32>`) so we can't use `EcoString::get` without casting each part of the `SrcSpan` to a i32 which is failable.

<details>
<summary>my (not working) code so far</summary>

```diff
diff --git a/compiler-core/src/type_/error.rs b/compiler-core/src/type_/error.rs
index 9b8fcea90..6edae62f9 100644
--- a/compiler-core/src/type_/error.rs
+++ b/compiler-core/src/type_/error.rs
@@ -719,6 +719,8 @@ pub enum Error {
     /// ```
     ///
     ListPrependWithoutElements {
+        list_text: EcoString,
+        tail_text: EcoString,
         location: SrcSpan,
         /// We hold onto the location of the tail expression for the code action.
         ///
diff --git a/compiler-core/src/type_/expression.rs b/compiler-core/src/type_/expression.rs
index 408c78c29..59976e277 100644
--- a/compiler-core/src/type_/expression.rs
+++ b/compiler-core/src/type_/expression.rs
@@ -1280,9 +1280,25 @@ impl<'a, 'b> ExprTyper<'a, 'b> {
         if let Some(ref tail) = tail
             && inferred_elements.is_empty()
         {
+            let tail_location = tail.location();
+
             self.problems.error(Error::ListPrependWithoutElements {
                 location,
-                tail_location: tail.location(),
+                tail_location,
+                list_text: self
+                    .environment
+                    .current_module
+                    .get(i64::from(location.start)..i64::from(location.end))
+                    .expect("list location should be valid")
+                    .clone()
+                    .into(),
+                tail_text: self
+                    .environment
+                    .current_module
+                    .get(i64::from(tail_location.start)..i64::from(tail_location.end))
+                    .expect("tail location should be valid")
+                    .clone()
+                    .into(),
             })
         }
```
</details>

ankddev: Won't this work? I've used similar in LS.

```rust
self.environment.current_module[location.start as usize..location.end as usize]
```

0xda157: Well that works type-wise, but it turns out `self.environment.current_module` is just the name of the module, not its source. After poking around a little I can't find the source anywhere else but I can look a little more.

0xda157: never mind, I got it, turns out we have access to the Src when creating the error message.

Issue #5945, opened 2026-07-06
compile-package command additions for use in Elixir's Mix


lpil:
There is a PR to add support for Gleam dependencies to the Mix build tool: https://github.com/elixir-lang/elixir/pull/14262

In order for it to be merged some additional functionality is required:

- [ ] `gleam compile-package` to compile _all_ source code in the package to BEAM, not just Gleam code. Any Elixir or Erlang code needs to be compiled too.
- [ ] `.app` file generation. This command historically does not do this because it does not know about all the BEAM code in the package, it only knows about the Gleam code. Since this command is being upgraded to compile non-Gleam code too we can also generate a `.app` file for the package.
- [ ] The ability to specify whether to compile `dev` and `test` directories as well as `src`.

Is there anything else?

Papipo:
@lpil can those be done separately? I can start working on the easiest one.

hiSandog:
Splitting the Gleam/Mix requirements into separate pieces seems wise: compiling non-Gleam BEAM sources, `.app` generation, and dev/test directory selection can each be validated independently. A tiny Mix fixture that depends on a Gleam package with Gleam, Erlang, and Elixir files would make the contract very clear.

lpil:
Yup all three can be done in different PRs, that would be preferable!

> A tiny Mix fixture that depends on a Gleam package with Gleam, Erlang, and Elixir files would make the contract very clear.

No need for that, we can test the API directly and leave the testing of Mix for the Elixir team.

Papipo:
> `gleam compile-package` to compile _all_ source code in the package to BEAM, not just Gleam code. Any Elixir or Erlang code needs to be compiled too.

Will this be the new behavior or do we require a new argument?

>`.app` file generation. This command historically does not do this because it does not know about all the BEAM code in the package, it only knows about the Gleam code. Since this command is being upgraded to compile non-Gleam code too we can also generate a `.app` file for the package.

Again, default behavior or the suggested `--app` parameter?

>  The ability to specify whether to compile `dev` and `test` directories as well as `src`.

How would this work exactly? Again, I am probably asking about parameters.

lpil:
We can make it the default for both.

> How would this work exactly? Again, I am probably asking about parameters.

A new flag

Papipo:
I'll work on this when I come back from vacations (a bit more than one week)

Papipo:
[...]

PR #6081, opened 2026-07-30
Add test for erlang and elixir package compilation to beam

Papipo:
The changes in this PR have been discussed beforehand [in an issue](https://github.com/gleam-lang/gleam/issues/5945)

I wanted to tackle the first item on the issue checklist but it seems that this is the current behaviour, so I added a test.


test/compile_package2/src/three.ex
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: Apache-2.0
```
jtdowney: This isn't valid Elixir, so it failed when I tried to run `make`.

jtdowney (changes requested):
I noticed this doesn't add the test to `.github/workflows/ci.yaml` so it wouldn't run in CI.

test/compile_package2/src/three.ex
jtdowney:
```
❯ make
# Remove any previously compiled code
rm -rf out
cargo run -- compile-package --out out --target erlang --lib . --package .
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.36s
warning: the following packages contain code that will be rejected by a future version of Rust: proc-macro-error2 v2.0.1
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `/Users/jtdowney/code/gleam/target/debug/gleam compile-package --out out --target erlang --lib . --package .`

== Compilation error in file out/_gleam_artefacts/three.ex ==
** (SyntaxError) invalid syntax found on out/_gleam_artefacts/three.ex:1:17:
    error: syntax error before: 'Identifier'
    │
  1 │ // SPDX-License-Identifier: Apache-2.0
    │                 ^
    │
    └─ out/_gleam_artefacts/three.ex:1:17
    (elixir 1.20.2) lib/kernel/parallel_compiler.ex:548: anonymous fn/5 in Kernel.ParallelCompiler.spawn_workers/8
error: Shell command failure

make: *** [build] Error 1
```

Papipo: Ah yes, that was force pushed after CI complained about the comment being missing.

lpil:
Hello! What's the purpose of this new test? We already have tests for compilation of Elixir and Erlang.

We are moving away from shell-script based tests, so I don't want to add more without clear reason.

Papipo:
> Hello! What's the purpose of this new test? We already have tests for compilation of Elixir and Erlang.

I couldn't find a test that checked that elixir was being compiled to beam 😅 

> 
> We are moving away from shell-script based tests, so I don't want to add more without clear reason.

Ok, so what I can do is use this new fixture to test for `.app` file generation instead. Unless there is a better way of testing this, which seems to be the case.

Papipo:
I am closing this in favor of working on the .app file generation.

PR #6085, opened 2026-07-30
LS: only offer generate dynamic decoder code action when stdlib is available

### LINKED ISSUE #6067: LS: do not suggest code actions with dependency on stdlib when it is unavailable
For example, if I have project without `gleam_stdlib` and call code action like `Generate dynamic decoder`, the LS will generate code which uses non-installed stdlib. I think that we could check for it's existence and don't show code action if it's not there. By greping I found that only `Generate dynamic decoder` depends on this so far, but this applies to #6059 too.

0xda157:
Fixes https://github.com/gleam-lang/gleam/issues/6067

### DIFF
--- a/language-server/src/code_action.rs
+++ b/language-server/src/code_action.rs
@@ -4541,7 +4541,16 @@ impl<'a, IO> GenerateDynamicDecoder<'a, IO> {
     }
 
     pub fn code_actions(&mut self) {
-        self.visit_typed_module(&self.module.ast);
+        let project_config = &self.compiler.project_compiler.config;
+        if project_config
+            .dependencies
+            .contains_key(STDLIB_PACKAGE_NAME)
+            || project_config
+                .dev_dependencies
+                .contains_key(STDLIB_PACKAGE_NAME)
+        {
+            self.visit_typed_module(&self.module.ast);
+        }
     }
 
     fn custom_type_decoder_body(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@
 
 ### Language server
 
+- The "Generate dynamic decoder" code action is now only offered when the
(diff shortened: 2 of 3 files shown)


language-server/src/code_action.rs
```diff
@@ -4541,7 +4543,16 @@ impl<'a, IO> GenerateDynamicDecoder<'a, IO> {
...
 
     pub fn code_actions(&mut self) {
-        self.visit_typed_module(&self.module.ast);
+        let project_config = &self.compiler.project_compiler.config;
+        if project_config
+            .dependencies
+            .contains_key(GLEAM_STDLIB_PACKAGE_NAME)
+            || project_config
+                .dev_dependencies
+                .contains_key(GLEAM_STDLIB_PACKAGE_NAME)
```
ankddev: This will show code action in case when stdlib is added as a dev dependency, but we are editing in `src`.

language-server/src/code_action.rs
```diff
@@ -108,6 +108,8 @@ impl CodeActionBuilder {
     }
 }
 
+const GLEAM_STDLIB_PACKAGE_NAME: &str = "gleam_stdlib";
```
ankddev: This constant is imported at the top of module!

language-server/src/code_action.rs
0xda157: I just copied the behavior of the Json code action, but I can try to figure out if we're in `src` `dev` or `test` if we think that's worth it. I do think it might be a little confusing if the code action is offered in `dev` but just does not appear in `src`

giacomocavalieri: I think copying what we do for json is fine! 👍

giacomocavalieri (approved):

lpil (approved):
Thank you!!

PR #6087, opened 2026-07-31
Fix underline span for function calls in constants

### LINKED ISSUE #6053: diagnostics: better underline for functions in constants
With following code:

```gleam
pub const wibble = wubble(wobble)
```

The compiler gives following error message:

```text
error: Syntax error
  ┌─ /data/data/com.termux/files/home/tgleam/src/wobble.gleam:1:20
  │
1 │ pub const wibble = wubble(wobble)
  │                    ^^^^^^^ Functions can only be called within other functions
```

I think that this underline is weird and should be on either of:
* Function name (so only `wubble` is highlighted)
* Full function call (so `wubble(wobble)` is highlighted)

Probably, it comes from the fact that parser finds that this is a function call and immediately emits an error.

mmustafasenoglu:
The diagnostic underline for function calls in const definitions previously included the opening parenthesis, producing something like:

```
3 │ const wib: Int = wibble()
  │                  ^^^^^^^ Functions can only be called within other functions
```

Now it only highlights the function name:

```
3 │ const wib: Int = wibble()
  │                  ^^^^^^ Functions can only be called within other functions
\```

Closes #6053

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -3533,16 +3533,8 @@ where
 
                         match self.tok0 {
                             Some((_, Token::LeftParen, paren_end)) => {
-                                self.advance(); // pop the LeftParen
-                                let call_end = self.find_matching_paren();
-                                let call_end = if call_end == 0 { paren_end } else { call_end };
-                                parse_error(
-                                    ParseErrorType::UnexpectedFunction,
-                                    SrcSpan {
-                                        start,
-                                        end: call_end,
-                                    },
-                                )
+                                self.advance();
+                                self.function_call_in_constant(start, paren_end)
                             }
                             _ => Ok(Some(Constant::Var {
                                 location: SrcSpan { start, end },
@@ -3572,16 +3564,8 @@ where
 
                 match self.tok0 {
                     Some((_, Token::LeftParen, paren_end)) => {
-                        self.advance(); // pop the LeftParen
-                        let call_end = self.find_matching_paren();
-                        let call_end = if call_end == 0 { paren_end } else { call_end };
-                        parse_error(
-                            ParseErrorType::UnexpectedFunction,
-                            SrcSpan {
-                                start,
-                                end: call_end,
-                            },
-                        )
+                        self.advance();
+                        self.function_call_in_constant(start, paren_end)
                     }
                     _ => Ok(Some(Constant::Var {
                         location: SrcSpan { start, end },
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ankddev** [comment]
Hello @mmustafasenoglu! Thank you! As you can see in https://github.com/gleam-lang/gleam/issues/6053#issuecomment-5090607513, we should highlight the whole call and not only function name, could you do it?

**Comment by mmustafasenoglu** [comment]
Updated! Now the underline covers the entire function call including parentheses and arguments:

```
wibble()       -> ^^^^^^^^  (was ^^^^^^)
wibble(1)      -> ^^^^^^^^^  (was ^^^^^^)
wibble(1, "a") -> ^^^^^^^^^^^^^^^^^^^  (was ^^^^^^)
```

This uses `find_matching_paren()` to scan ahead for the closing `)` in the token stream.

File: compiler-core/src/parse.rs [hunk 1]
```diff
@@ -3566,13 +3566,14 @@ where
...
-                    Some((_, Token::LeftParen, _)) => parse_error(
-                        ParseErrorType::UnexpectedFunction,
-                        SrcSpan {
-                            start,
-                            end: end + 1,
-                        },
-                    ),
+                    Some((_, Token::LeftParen, _)) => {
+                        // Scan ahead to find the matching closing paren
+                        let call_end = self.find_matching_paren();
```
giacomocavalieri: Shouldn't we pop the left paren here rather than in the `find_matching_paren` function? Also what happens if the parentheses are not matched? What would the error look like in those cases?

mmustafasenoglu:
Thanks for the review @giacomocavalieri! Good catch -- I've updated the code to address both points:

1. **LeftParen is now popped by the caller** via advance() before calling find_matching_paren, so the function no longer manages the opening paren internally. This makes the contract clearer -- it simply scans from the current token position looking for the matching ).

2. **Unmatched parens** are now handled gracefully: if find_matching_paren reaches EOF without finding a closing paren, the error span falls back to the opening paren position (instead of pointing to byte 0).

giacomocavalieri:
Hello, are you using copilot? Could you please disable those automatic reviews they get quite annoying, thanks!

mmustafasenoglu:
Hi @giacomocavalieri, sorry about that — I had Copilot enabled and didnt realize it was firing off automatic reviews on my PRs. Ive disabled it now. Wont happen again!

mmustafasenoglu:
Hi @giacomocavalieri — unfortunately I don't have permission to delete Copilot's review comments from the API. They were left before I disabled the bot. Please feel free to ignore them.

Regarding your earlier question about popping LeftParen and unmatched parens: the caller pops LeftParen via `advance()` before calling `find_matching_paren()`, which assumes the opening paren is already consumed. For unmatched parens (EOF), `find_matching_paren` returns 0 and the caller falls back to `paren_end`, producing a span up to the opening paren — which is a reasonable error span.

Let me know if you'd like any further changes!

compiler-core/src/parse.rs
```diff
@@ -4508,6 +4518,54 @@ functions are declared separately from types.";
...
+                    }
+                }
+                None => break,
+            }
+        }
+
+        // Note: we leave the parser state somewhat disrupted here,
+        // but since this is called right before parse_error (which
+        // returns Err immediately), it doesn't matter.
+        last_end
```
lpil: Good comment! Thank you.

To make it safer to use let's have this function construct and return the error, that way it is more clear when it can be used, and less change of leaving the parser in a bad state.

lpil: I'm doing this now

lpil:
Oh rats, I forgot about the changelog. I'll make another PR

Issue #6079, opened 2026-07-30
Merge alternative patterns in JS decision trees


jtdowney:
While thinking about #6027 the other day, I noticed this optimization. Currently, the decision tree used to codegen for JS is conservative when it comes to alternative patterns like:

```gleam
pub type Wibble {
  Wibble(Int)
  Wobble(Int)
  Woo
}

pub fn go(a: Wibble) {
  case a {
    Wibble(a) | Wobble(a) -> a
    Woo -> 2
  }
}
```

Generates

```javascript
export function go(a) {
  if (a instanceof Wibble) {
    let a$1 = a[0];
    return a$1;
  } else if (a instanceof Wobble) {
    let a$1 = a[0];
    return a$1;
  } else {
    return 2;
  }
}
```

The code could be more optimally generated as

```javascript
export function go(a) {
  if (a instanceof Wibble || a instanceof Wobble) {
    let a$1 = a[0];
    return a$1;
  } else {
    return 2;
  }
}
```

I plan to investigate this and see what can be done. This was previously discussed back in #1147, but that predates the decision tree work, so I think it would be easier to pull off now. This is somewhat related to #5475 but is a different problem.

giacomocavalieri:
I'm curious to see if you find a way to do so, given how the decision tree works I'm afraid it wouldn't be quite straightforward as it seems!

PR #6089, opened 2026-07-31
Fix JavaScript clause guards with top level or

jtdowney:
While investigating #6079 I noticed [this snapshot](https://github.com/gleam-lang/gleam/blob/4bc44632f47377cc4962f922251829ff910068ad/compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__case__case_branches_guards_are_wrapped_in_parentheses.snap#L23) picked up a bug from #4678. This code

```gleam
pub fn anything() -> a {
  case [] {
    [a] if False || True -> a
    _ -> anything()
  }
}
```

Was generating

```javascript
export function anything() {
  while (true) {
    let $ = $List$Empty$const;
    if (!($ instanceof $Empty)) {
      let $1 = $.tail;
      if ($1 instanceof $Empty && false || true) {
        let a = $.head;
        return a;
      }
    }
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,3 +16,7 @@
 ### Formatter
 
 ### Bug fixes
+
+- Fixed a bug where on the JavaScript target a case clause whose guard's top
+  level operator was `||` could run for a subject its pattern did not match.
+  ([John Downey](https://github.com/jtdowney))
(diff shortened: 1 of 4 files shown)


giacomocavalieri (approved):

lpil (approved):
Thank you!!

PR #6092, opened 2026-07-31
erl: fix miscompilation of some consts in bit arrays

### LINKED ISSUE #6086: Constant aliasing another module's constant miscompiles in a string concatenation (Erlang target)
Constant aliasing another module's constant miscompiles in a string concatenation (Erlang target)

Gleam 1.18.0, Erlang/OTP 28, macOS.

```gleam
// src/repro/other.gleam
pub const value = "a"
```

```gleam
// src/repro.gleam
import repro/other

const aliased = other.value

pub fn main() {
  aliased <> "b"
}
```

`gleam run` crashes with `erlang:error(Badarg)`.

Generated Erlang, 1.18.0:

```erlang
main() ->
    <<"a"/binary, "b"/utf8>>.
```

`"a"` is an Erlang string (a list), so the `/binary` specifier is a type error.
erlc also warns at compile time:
`Warning: binary construction will fail because of a type mismatch`.

1.16.0 emitted the correct form:

```erlang
main() ->
    <<(<<"a"/utf8>>)/binary, "b"/utf8>>.
```

Only the alias form is affected. Using `other.value` directly in the
concatenation, or a local `const x = "a"`, both emit `/utf8` correctly.
The JavaScript target is unaffected (`aliased + "b"`).

Looks like the same family as #6045 (fixed by #6049), but a different code
path: that one was a local constant in a bit array segment. The repro from
#6045 runs clean on 1.18.0, while the one above still crashes on the same
binary.

ankddev:
* Closes #6086 
* It also fixes similar bug in clause guards
***

### DIFF
--- a/compiler-core/src/erlang.rs
+++ b/compiler-core/src/erlang.rs
@@ -3362,33 +3362,25 @@ fn type_export(custom_type: &TypedCustomType) -> (EcoString, usize) {
 
 /// This returns true if the given expression is going to be compiled to a
 /// single literal Erlang string.
-/// This is not true just for literal Gleam strings like `"abc"`, but also
+/// This is not only true for literal Gleam strings like `"abc"`, but also for
 /// variables referencing string constants (as those are inlined)
 fn produces_literal_string(value: &TypedExpr) -> bool {
     match value {
-        TypedExpr::String { .. }
+        TypedExpr::String { .. } => true,
         // Constants are inlined on the Erlang target, so we need to check if
-        // those are literal strings too!
-        | TypedExpr::ModuleSelect {
-            constructor:
-                ModuleValueConstructor::Constant {
-                    literal: Constant::String { .. },
-                    ..
-                },
+        // those produce literal strings too!
+        TypedExpr::ModuleSelect {
+            constructor: ModuleValueConstructor::Constant { literal, .. },
             ..
         }
         | TypedExpr::Var {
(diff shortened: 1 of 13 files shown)


compiler-core/src/erlang.rs
```diff
@@ -3394,21 +3386,49 @@ fn produces_literal_string(value: &TypedExpr) -> bool {
...
+        | Constant::RecordUpdate { .. }
+        | Constant::BitArray { .. }
+        | Constant::StringConcatenation { .. }
+        | Constant::Invalid { .. }
+        | Constant::Todo { .. }
+        | Constant::Var { .. } => false,
+    }
+}
+
+/// This returns true if the given expression is going to be compiled to a
```
giacomocavalieri:
```suggestion
/// This returns true if the given guard is going to be compiled to a
```

compiler-core/src/erlang.rs
```diff
@@ -3394,21 +3386,49 @@ fn produces_literal_string(value: &TypedExpr) -> bool {
 
 /// This returns true if the given expression is going to be compiled to a
```
giacomocavalieri:
```suggestion
/// This returns true if the given constant is going to be compiled to a
```

ankddev:
@giacomocavalieri Done!

lpil (approved):
Thank you!!

PR #6109, opened 2026-08-01
set proper permission for built tar of docs

### LINKED ISSUE #5992: Set file permissions in documentation tarballs
Gleam's documentation tarballs give files a mode of `0`. Hex works around this, but a [recent regression](https://github.com/hexpm/hexpm/issues/1725#issuecomment-5011703341) in that workaround caused successful documentation uploads to return 404s.

Regular Hex package entries set an explicit mode https://github.com/gleam-lang/gleam/blob/beb4da4dc916d8115612bfc6f3356f445a7af7d3/compiler-cli/src/publish.rs#L755

While `create_tar_archive` does not call `set_mode`, thus it leaves the mode at zero.

It would be preferable that Gleam matched the rest of the Hex ecosystem instead of relying on a workaround to be maintained.

giacomocavalieri:
This should fix #5992, I've changed the `create_tar_archive` method directly since it's always used just for docs. Maybe it could make sense to pull that method out of the `fs` crate and give that a name that refers to docs instead?
I'm not sure if it would always make sense to set the permission like that for every tar archive we might be building in future using that function.
Or rather we could take the permission as an argument?

Let me know what you think!


lpil (commented):
Changelog entry and tests for this one?

ankddev (commented):
I think that you committed test project from another PR, didn't you?

giacomocavalieri:
> Changelog entry and tests for this one?

I'm not sure how I could write a test for this? Do I check the generated file has the proper permissions? What do you think we should write in the CHANGELOG?

lpil:
> I'm not sure how I could write a test for this? Do I check the generated file has the proper permissions?

Yup a normal example test with some assertions.

> What do you think we should write in the CHANGELOG?

Could be something like "Fixed a bug where file permissions were not set in Hex tarballs"?

lpil (approved):
Thank you!!!

PR #6110, opened 2026-08-01
ts overload

### LINKED ISSUE #5715: Improve TS type narrowing
In the following issue, the example are made with a simple `Option(a)` in Gleam.

When using an `Option(a)` to TS, we know for sure that the expected type is `Option$`, so we can type functions signatures as `(value: $option.Option$<EL>) => ....`.
When using the new API, "pattern matching" is defined as `$option.Option$isSome(value))`. 
The type narrowing works, but will refine value from `Option$<EL>` to `Option$<unknown>`, because `isSome` is defined as `export function Option$isSome<EL>(value: any): value is Option$<unknown>;`. 

Thanks to TS and function overloading, we could do way better, and keep the generic type, while narrowing to the correct type. We could write an overload `export function Option$isSome<EL>(value: Option$<EL>): value is Some<EL>`. That way, it would keep the generic.

giacomocavalieri:
This PR fixes #5715 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,26 @@
   to optimise.
   ([John Downey](https://github.com/jtdowney))
 
+- When producing TypeScript annotations, the compiler now produces overloads to
+  narrow the type of variants of a custom type. For example, given the following
+  type:
+
+  ```gleam
+  pub type Box(value) {
+    Full(value)
+    Empty
+  }
+  ```
+
+  Will now produce these two additional overloads:
+
+  ```ts
+  export function Box$isFull<I>(value: Box$<I>): value is Full<I>;
+  export function Box$isEmpty<I>(value: Box$<I>): value is Empty;
+  ```
+
+  ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
+
 ### Build tool
 
 - The build tool now stores its build cache in a more compact binary format,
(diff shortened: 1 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,26 @@
...
+  type:
+
+  ```gleam
+  pub type Box(value) {
+    Full(value)
+    Empty
+  }
+  ```
+
+  Will now produce these two additional functions:
```
lpil: No additional functions are produced, the existing functions have a new overload!

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__custom_types__generic_type_parameter_used_in_field.snap
```diff
@@ -34,6 +34,10 @@ export function Wibble$Wibble<I, J>(
 export function Wibble$isWibble<I, J>(
   value: any,
 ): value is Wibble$<unknown, unknown>;
+export function Wibble$isWibble<I, J>(value: Wibble$<I, J>): value is Wibble<
+  I,
+  J
+>;
```
lpil, line 37: There's a bug here! Overloads are tried top-to-bottom, so due to the catch-all `any` going first this new specialised overload is unreachable.

Would be great to have a TypeScript test to verify it gets type checked the way we intend it to be.

giacomocavalieri: Oh TIL we have typescript tests! I've added them!

lpil (commented):
Don't make a new project for typescript tests please, we already have one and we are trying to move away from shell script based tests 🙏

giacomocavalieri:
Oops should all be good now, and I've moved tests to the existing typescript project

lpil (approved):
Excellent! Thank you!!!

PR #6114, opened 2026-08-01
fix: replace undeclared variable for UnsupportedProceduralOperator message

khalidbelk:

Seems like the example variable is wrong (the variable `bounce` isn't declared before, it would give an out of scope error), I guess it went under the radar since both words start with letter '**b**'.

Related to **PR** #5987


lpil (approved):
Oops! thank you!!

PR #6118, opened 2026-08-03
Skip binding in JavaScript when a switch value is used once

jtdowney:
This is another item I noticed while investigating #6079 that seemed like a nice optimization. The code generator was very conservative about creating a binding for values, even if they were only ever used once. This prevented the existing if merging from creating a better condition. For example, this code:

```gleam
pub fn go(x) {
  case x {
    Wibble(1, 2) -> 1
    _ -> 2
  }
}
```

Currently generates

```javascript
export function go(x) {
  if (x instanceof Wibble) {
    let $ = x[0];
    if ($ === 1) {
      let $1 = x[1];
      if ($1 === 2) {
        return 1;
      } else {
        return 2;
      }
    } else {
      return 2;
    }
  } else {
    return 2;
  }
}
```

But with this change, it now generates

```javascript
export function go(x) {
  if (x instanceof Wibble && x[0] === 1 && x[1] === 2) {
    return 1;
  } else {
    return 2;
  }
}
```

The change explores how many times a value is used, and if it is ever used more than once, a binding is still preferred. That way, more expensive operations like string slices aren't duplicated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,13 @@
   expression now points to the entire function call.
   ([mmustafasenoglu](https://github.com/mmustafasenoglu))
 
+- Pattern matching on the JavaScript target now generates flatter code, with
+  nested `if` statements collapsed into a single condition and fewer
+  intermediate variables. This makes little difference to the size of a bundled
+  application, but the resulting code has fewer branches for JavaScript engines
+  to optimise.
+  ([John Downey](https://github.com/jtdowney))
+
 ### Build tool
 
 - The build tool now stores its build cache in a more compact binary format,
--- a/compiler-core/src/javascript/decision.rs
+++ b/compiler-core/src/javascript/decision.rs
@@ -23,7 +23,10 @@ use ecow::{EcoString, eco_format};
 use itertools::Itertools;
 use num_bigint::BigInt;
 use pretty_arena::*;
-use std::{collections::HashMap, sync::OnceLock};
+use std::{
+    collections::{HashMap, HashSet},
+    sync::OnceLock,
+};
 
 pub const ASSIGNMENT_VAR: &str = "$";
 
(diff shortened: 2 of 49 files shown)


compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__assignments__assert1.snap
```diff
@@ -11,21 +11,7 @@ import { makeError } from "../gleam.mjs";
...
-        FILEPATH,
-        "my/mod",
-        1,
-        "go",
-        "Pattern match failed, no pattern matched the value.",
-        { value: x, start: 15, end: 37, pattern_start: 26, pattern_end: 33 }
-      )
-    }
-  } else {
+  if (!(x[0] === 1 && x[1] === 2)) {
```
giacomocavalieri, line 14: What a nice improvement!!

CHANGELOG.md
```diff
@@ -17,6 +17,58 @@
...
+  export function go(x) {
+    if (x instanceof Wibble && x[0] === 1 && x[1] === 2) {
+      return 1;
+    } else {
+      return 2;
+    }
+  }
+  ```
+
+  ([John Downey](https://github.com/jtdowney))
```
giacomocavalieri, line 29: I think the explanation is a bit technical and not super relevant for the changelog. I'd just focus on the user facing effects like how the generated code is smaller (maybe with some measures of how much for a real project like a lustre application? See if it also changes the size of the bundled app) rather than explaining how that's done or when that triggers in the compiler

compiler-core/src/javascript/decision.rs
```diff
@@ -390,6 +393,112 @@ impl<'a, 'doc> CasePrinter<'_, '_, 'a, '_, 'doc> {
...
+            // A guarded body is different: even when it returns a subject, the
+            // bindings its check needs are used before the check itself, so
+            // those still count.
+            Decision::Guard {
+                if_true, if_false, ..
+            } => {
+                self.count_body_value_uses(if_true, derived)
+                    + self.count_decision_value_uses(if_false, derived)
+            }
+            Decision::Switch {
```
giacomocavalieri, line 457: Could you add a comment for switch as well? It would be good to have something that explains why it's computed this way for a switch, especially that `in_branches + 1` that comes later

compiler-core/src/javascript/decision.rs
```diff
@@ -390,6 +393,112 @@ impl<'a, 'doc> CasePrinter<'_, '_, 'a, '_, 'doc> {
...
+                    in_branches + 1
+                } else {
+                    in_branches
+                }
+            }
+        }
+    }
+
+    /// Whether the clause `body` belongs to is rebuilding and returning one of
+    /// the case subjects, in which case its own bindings are never used.
```
giacomocavalieri: I don't understand what that "its own bindings" is referring to. Whose bindings are we talking about?

compiler-core/src/javascript/decision.rs
```diff
@@ -390,6 +393,112 @@ impl<'a, 'doc> CasePrinter<'_, '_, 'a, '_, 'doc> {
...
+        };
+
+        clauses
+            .get(body.clause_index)
+            .expect("invalid clause index")
+            .returned_subject()
+            .is_some()
+    }
+
+    fn count_body_value_uses(&self, body: &'a Body, derived: &HashSet<usize>) -> usize {
```
giacomocavalieri: Can you add a doc comment explaining what this does and what that `derived` set represents?

compiler-core/src/javascript/decision.rs
```diff
@@ -390,6 +393,112 @@ impl<'a, 'doc> CasePrinter<'_, '_, 'a, '_, 'doc> {
...
+            .iter()
+            .map(|(_check, decision)| decision)
+            .chain([fallback])
+            .any(|decision| self.count_decision_value_uses(decision, &derived) > 0)
+    }
+
+    /// The number of times `decision` uses the value of one of the `derived`
+    /// variables.
+    ///
+    fn count_decision_value_uses(&self, decision: &'a Decision, derived: &HashSet<usize>) -> usize {
```
giacomocavalieri: Do we need to count it, or would a bool telling us if it's used/not used suffice? Would also enable us to early return in many cases without inspecting everything if that's all we need

giacomocavalieri (commented):
What an amazing improvement in the generated code! I've left a couple of comments

CHANGELOG.md
jtdowney: I pushed an updated one in 34d65231d7f652f73a34fee8f400acce765467a7

compiler-core/src/javascript/decision.rs
jtdowney: Done in 34d65231d7f652f73a34fee8f400acce765467a7

compiler-core/src/javascript/decision.rs
jtdowney: Thanks, I tried to clarify in 34d65231d7f652f73a34fee8f400acce765467a7

compiler-core/src/javascript/decision.rs
jtdowney: Done in 34d65231d7f652f73a34fee8f400acce765467a7

compiler-core/src/javascript/decision.rs
jtdowney: Great point, and you are correct. I overthought it on an earlier pass, and it reduces to a bool now. Changed in 34d65231d7f652f73a34fee8f400acce765467a7

lpil (approved):
Amazing work! I am so impressed!!!

Did we have any measurements of any performance improvements? So I can perhaps use them in the release post.

jtdowney:
All of my attempts to benchmark the change from a speed perspective have not shown anything outside of the noise window. I would say it's neutral on speed, which isn't surprising; JS engines are very good. It is -0.4% on generated JS code size for a full lustre app with dependencies. There are fewer registers allocated and the bytecode is smaller. The rest of what I am investigating for #6079 will likely be a bigger difference.

PR #6126, opened 2026-08-03
Remove abbreviations from parser

lpil:
Bringing parser more into our regular style

### DIFF
--- a/compiler-core/src/parse.rs
+++ b/compiler-core/src/parse.rs
@@ -197,7 +197,7 @@ pub fn parse_module(
 pub fn parse_statement_sequence(src: &str) -> Result<Vec1<UntypedStatement>, ParseError> {
     let lex = lexer::make_tokenizer(src);
     let mut parser = Parser::new(lex);
-    let expr = parser.parse_statement_seq();
+    let expr = parser.parse_statement_sequence();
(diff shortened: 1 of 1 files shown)


compiler-core/src/parse.rs
```diff
@@ -1134,22 +1134,24 @@ where
...
-            _ => ParseError {
-                error: ParseErrorType::ExpectedValue,
-                location: SrcSpan {
-                    start: eq_s,
-                    end: eq_e,
+                _ => ParseError {
+                    error: ParseErrorType::ExpectedValue,
+                    location: SrcSpan {
+                        start: eq_s,
+                        end: eq_e,
```
GearsDatapacks, line 1151: We still have a lot of abbreviations in the parser like these two here – mostly start and end positions of tokens. Do we want to expand those now as well or should we wait until a later date?

GearsDatapacks (commented):
Ah brilliant, I've been wanting to do this for ages now!

compiler-core/src/parse.rs
lpil: If you'd like to do another pass over this code and fix more that would be fab. I'll merge this to avoid risk of merge conflicts with others, and folks can do additional PRs.

PR #6134, opened 2026-08-04
Improve type printing using type aliases

### LINKED ISSUE #6133: Support printing type alias re-exports of specialised generic types
Right now, the type printer can print type aliases if it deems them to be _re-exports_, which are type aliases that expose an internal type publicly. However, it only supports named re-exports that wrap another named type directly, and pass through the exact same type parameters.

For example:

```gleam
// works:
import lustre/vdom/vnode
pub type Element(message) = vnode.Element(message)

// doesn't work:
import gleam/http/request
@internal pub type Connection
pub type Request = request.Request(Connection)
```

The rule to determine re-exports could be expanded to be any public type alias that includes an internal type on its right hand side.

yoshi-monster:
fix #6133

This extends alias resolution to use the type checking mechanism to determine matching types. If the right-hand-side of a type alias declaration can unify with some usage place where we want to print a type, the alias can be considered instead.

It also expands what it means to be a re-export to include all aliases that mention an internal type on their right-hand-side, so the following types would also be considered re-exports under the new rule:

```gleam
@internal pub type Internal

pub type Internals = List(Internal)
pub type InternalConsumer = fn(Internals) -> Nil
pub type InternalPair(a) = #(Internal, a)
pub type InternalCounter = InternalPair(Int)
```

The following types are checked in order to find a matching type alias:

1. A re-export in the local module 
2. A re-export that is imported unqualified
3. A re-export in another imported module
4. When printing docs, a re-export from any other module, as long as it resolves to a unique type alias.

In addition to being used in the type printer, the new mechanism is also used
to match up local aliases for the language server and error printing.
Global names are not scanned in that case to make types in errors and in
the language server follow real Gleam syntax that is valid in the current module.

Since I just came up with some rules of how it should work I'd suggest we discuss them first though.

--- 


yoshi-monster:
One difference for example is that i couldn't understand the language server rules for imported aliases. From what I got from #3749 printing aliases is wanted there in theory, but in practice was very limited because types would not be matched up properly.

I personally don't like printing aliases that were imported unqualified at all really, but it looks to me like that is the behaviour that is wanted?

compiler-core/src/docs/tests.rs
```diff
@@ -1153,6 +1153,41 @@ pub fn do_thing(value: internal.Internal) -> External {
...
 }
 
+#[test]
+fn function_uses_generic_reexport_of_internal_type() {
+    assert_documentation!(
+        ("thepackage/internal", "pub type Wibble(a)"),
+        "
+import thepackage/internal
+
+pub type Wobble(a) = internal.Wibble(List(a))
```
lpil: Place the right hand side of `=` on a new line for type aliases please

compiler-core/src/docs/tests.rs
```diff
@@ -1176,6 +1211,103 @@ pub fn do_thing(value: something.External) {
...
+}
+
+#[test]
+fn function_only_uses_unique_global_reexport_of_internal_type() {
+    assert_documentation!(
+        (
+            "thepackage/internal",
+            "
+@internal pub type Wibble(a)
+@internal pub type Wobble(a)
```
lpil: Module attributes go on their own line please

compiler-core/src/docs/tests.rs
```diff
@@ -1153,6 +1153,41 @@ pub fn do_thing(value: internal.Internal) -> External {
...
+
+#[test]
+fn function_uses_structural_reexports_of_internal_type() {
+    assert_documentation!(
+        ("thepackage/internal", "pub type Wibble(a)"),
+        "
+import thepackage/internal
+
+pub type Handler(a) = fn(internal.Wibble(a)) -> Nil
+pub type Pair(a) = #(internal.Wibble(a), Int)
```
lpil: Empty line between definitions please

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__error_uses_type_alias_with_internal_generic_argument.snap
```diff
@@ -0,0 +1,30 @@
...
+8 │   request == [1]
+  │              ^^^
+
+Expected type:
+
+    Request
+
+Found type:
+
+    List(Int)
```
lpil, line 31: 🔴 This error message has become extremely unclear, there's no way for the programmer to know what the difference between these two types is the request parameter. Any further use of aliases must never obscure debugging information in error messages.

lpil (commented):
Hello! I have not looked at the implementation fully, but while reading the tests I found a blocker that means this specific behaviour couldn't be accepted. Please see my inline comment 🙏

compiler-core/src/type_/tests/snapshots/gleam_core__type___tests__errors__error_uses_type_alias_with_internal_generic_argument.snap
yoshi-monster: Hello!

This is already a behaviour of the compiler, I didn't really change anything there:

```gleam
pub type Wibble = String
pub fn main(wibble: Wibble) {
  assert [1] == wibble
}
```

errors with

```
error: Type mismatch
  ┌─ /src/main.gleam:3:17
  │
3 │   assert [1] == wibble
  │                 ^^^^^^

Expected type:

    List(Int)

Found type:

    Wibble
```

As I said I think those exact rules should be discussed and because I expanded them to be applied more consistently there may be a bunch of new cases now where the current behaviour isn't what we'd want (in that case, I expanded it to work with a generic type on the RHS, so `List(Connection)` doesn't fall-back to printing the underlying type). I'd move this to the discussion though :D

lpil: That is not the same thing. This change is not acceptable as it further hides important debugging information, replicating the Elm problem. Avoiding that is a primary goal of this work and far more important and the other benefits.

yoshi-monster:
Hello!

I've kept type printing the same for docs and also for errors, but errors are now smart about expanding type aliases and gain an additional hint, showing just enough information to the user so the underlying type error is visible. All type aliases whose names are different, or whose underlying type is not exactly the same are recursively expanded while printing.

In practice, this will look like this:

```gleam
pub type UserId =
  Int

pub type Metadata =
  List(String)

pub type Expected =
  Result(UserId, Metadata)

pub fn main(given: Result(String, Metadata)) {
  let expected: Expected = given
}
```

```
Expected type:

    Expected

Found type:

    Result(String, Metadata)

Hint: The expected type expands to:

    Result(Int, Metadata)
```

I hope you like them! I spent a lot of time trying to get the behaviour just right.

~ 💜

PR #6137, opened 2026-08-05
Stopped unknown type errors cascading through usages when an invalid type alias is created

### LINKED ISSUE #6072: Using an invalid type alias reports cascading errors
The following code:

```gleam
pub type UnknownTypeAlias = UnknownType

pub type UsageOfUnknownTypeAlias = UnknownTypeAlias
```

Produces the error:

```
error: Unknown type
  ┌─ /src/main.gleam:1:29
  │
1 │ pub type UnknownTypeAlias = UnknownType
  │                             ^^^^^^^^^^^

The type `UnknownType` is not defined or imported in this module.

error: Unknown type
  ┌─ /src/main.gleam:3:36
  │
3 │ pub type UsageOfUnknownTypeAlias = UnknownTypeAlias
  │                                    ^^^^^^^^^^^^^^^^

The type `UnknownTypeAlias` is not defined or imported in this module.
```

I would expect the first error to be reported as the type `UnknownType` really doesn't exist. However with the second error, I wouldn't expect it to be reported to the user until they get the previous error fixed to prevent cascading errors.

Interestingly enough:

```gleam
pub type UnknownTypeAlias = UnknownType

pub type UnknownTypeAlias {}
```

reports

```
error: Duplicate type definition
  ┌─ /src/main.gleam:1:1
  │
1 │ pub type UnknownTypeAlias = UnknownType
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Redefined here
  ·
5 │ pub type UnknownTypeAlias {}
  │ ------------------------- First defined here

The type `UnknownTypeAlias` has been defined multiple times.
Names in a Gleam module must be unique so one will need to be renamed.
```

which acknowledges that the alias `UnknownTypeAlias` *has* already been defined which contradicts

```
error: Unknown type
  ┌─ /src/main.gleam:3:36
  │
3 │ pub type UsageOfUnknownTypeAlias = UnknownTypeAlias
  │                                    ^^^^^^^^^^^^^^^^

The type `UnknownTypeAlias` is not defined or imported in this module.
```

###### Tested on the Gleam Playground with Gleam version 1.17.0 and on Arch Linux with Gleam version 1.18.0

jamesdolan16:
Fixes https://github.com/gleam-lang/gleam/issues/6072;

Modified the code that registers type aliases so that when a type constructor cannot be found, instead of returning an error straight away, we use a new unbound type var for that alias, allowing later usages to remain valid thereby getting rid of the cascading unknown type error messages.

Also refactored the `both_errors_are_shown` test, as it appears that was there to ensure a previous fix related to this outputted a sensible error, which at the time was deemed to be the cascading behaviour. As this fix overrides that behaviour with the root cause/non-cascading message, I felt the test should be updated to reflect this.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1499,7 +1499,14 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
         let arity = parameters.len();
         let tryblock = || {
             hydrator.disallow_new_type_variables();
-            let type_ = hydrator.type_from_ast(resolved_type, environment, &mut self.problems)?;
+            let hydrator_result =
+                hydrator.type_from_ast(resolved_type, environment, &mut self.problems);
+
+            let type_ = match &hydrator_result {
+                Ok(type_) => type_.clone(),
+                Err(Error::UnknownType { .. }) => environment.new_unbound_var(),
+                Err(e) => return Err(e.clone()),
+            };
 
             environment
                 .names
@@ -1545,7 +1552,11 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
                 });
             }
 
-            Ok(())
+            if let Err(e) = hydrator_result {
+                Err(e)
+            } else {
+                Ok(())
+            }
         };
         let result = tryblock();
         self.record_if_error(result);
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1487,8 +1487,7 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
 
             let type_ = match &hydrator_result {
                 Ok(type_) => type_.clone(),
-                Err(Error::UnknownType { .. }) => environment.new_unbound_var(),
-                Err(e) => return Err(e.clone()),
+                Err(_) => environment.new_unbound_var(),
             };
 
             environment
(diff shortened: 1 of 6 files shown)


compiler-core/src/analyse.rs
```diff
@@ -1499,7 +1499,14 @@ impl<'a, A> ModuleAnalyzer<'a, A> {
...
             hydrator.disallow_new_type_variables();
-            let type_ = hydrator.type_from_ast(resolved_type, environment, &mut self.problems)?;
+            let hydrator_result =
+                hydrator.type_from_ast(resolved_type, environment, &mut self.problems);
+
+            let type_ = match &hydrator_result {
+                Ok(type_) => type_.clone(),
+                Err(Error::UnknownType { .. }) => environment.new_unbound_var(),
+                Err(e) => return Err(e.clone()),
+            };
```
lpil, line 1508: How come this is specialised to `UnknownType` instead of being for any error?

lpil (commented):
Hello! I've left a question inline. Thank you

compiler-core/src/analyse.rs
jamesdolan16: My thinking was that we would only want to create the unbound var type if the hydrator couldn't find the type, however I think I might have gotten a bit confused there as that makes no sense, I'll change that when I get chance 😆

jamesdolan16: @lpil I have got rid of the specialisation now

lpil (approved):
Thank you!!! Lovely work!! And thanks for waiting

ankddev:
@lpil I guess that you forgot to approve CI, so it isn't running

PR #6140, opened 2026-08-06
fix: show non-printable binaries as bit arrays in echo output

### LINKED ISSUE #6040: echo is too eager to show bit arrays as strings on Erlang
On Erlang strings and bit arrays have the same runtime implementation, so we rely on heuristics to decide which syntax should be used to print each one.

https://github.com/gleam-lang/gleam/blob/97b74f828020337dd5e78a11f837d8d59f6ad9ec/compiler-core/templates/echo.erl#L140-L158

Our current approach is a bit too keen to use the string syntax, with bit arrays like `<<1, 2, 3>>` being printed as `"\u{0001}\u{0002}\u{0003}"`.

How should we adjust the heuristics to show this as `<<1, 2, 3>>`? As that is much more likely to be the desired syntax. Perhaps we only use the `\u{}` syntax if there are some printable graphemes in the string?

Any change will likely need to ported to the stdlib for string.inspect too.

mmustafasenoglu:
Closes #6040

In Erlang, strings and bit arrays share the same runtime representation. The current `echo` implementation eagerly tries to display binaries as UTF-8 strings, which means bit arrays like `<<1, 2, 3>>` get printed as `"\u{0001}\u{0002}\u{0003}"`.

This changes the heuristic: a binary is only displayed as a string if it contains at least one printable character (ASCII 32-126, newline, carriage return, tab, or form feed). Non-printable binaries are shown as bit arrays.

**Before:**
```gleam
echo(<<1, 2, 3>>)
// "\u{0001}\u{0002}\u{0003}"
```

**After:**
```gleam
echo(<<1, 2, 3>>)
// <<1, 2, 3>>
```

Changes:
- `compiler-core/templates/echo.erl`: Added `inspect@has_printable` check in `inspect@binary` to only display as string when the binary contains printable characters

### DIFF
--- a/compiler-core/templates/echo.erl
+++ b/compiler-core/templates/echo.erl
@@ -84,11 +84,21 @@ inspect@bit_array_pieces(Bits, Acc) ->
 inspect@binary(Binary) ->
     case inspect@maybe_utf8_string(Binary, <<>>) of
         {ok, InspectedUtf8String} ->
-            InspectedUtf8String;
+            case inspect@has_printable(Binary) of
+                true -> InspectedUtf8String;
+                false ->
+                    Segments = [erlang:integer_to_list(X) || <<X>> <= Binary],
+                    ["<<", lists:join(", ", Segments), ">>"]
+            end;
         {error, not_a_utf8_string} ->
             Segments = [erlang:integer_to_list(X) || <<X>> <= Binary],
             ["<<", lists:join(", ", Segments), ">>"]
     end.
+
+inspect@has_printable(<<>>) -> false;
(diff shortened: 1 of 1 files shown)


ankddev:
Hello @mmustafasenoglu! Thank you, but that issue is under discussion, so we do not receive PRs for it at the moment. And could you disable that copilot review, I find it annoying?

mmustafasenoglu:
Closing — issue is still under discussion. Will wait for consensus before contributing. Thanks!

lpil:
Hi @mmustafasenoglu ! Your proposed heuristic seems reasonable to me.

I'd like the implementation to only do a single pass over the bit string though, rather than adding another as this PR does 🙏

PR #6141, opened 2026-08-08
fix: harden github action generated by gleam new

ghostdevv:

Apologies for not opening an issue first! I noticed that the generated GitHub workflow/action could do with a few quality of life and hardening steps. I'm not sure what area this would come under, and there didn't seem to be an easy way to grep through all of the previous changelogs so I just put it under bug fix for now - let me know if there's something better!


GearsDatapacks:
Unless fixing a specific bug (which from what I can tell, this isn't), the changelog entry should go under "build tool", as it's changing the output of the build tool.

I don't know enough about github actions to comment on the actual contents of the PR

khalidbelk:
Seems like the `concurrency` part is a good addition IMO.

SHA-pinning seems like a good idea for security for larger codebases like in this own repo's CI, but idk if it's worth "obfuscating" the yaml (given the low risks here) in a freshly scaffolded project's default CI file. Let's see what Louis / Jak think about it.

CHANGELOG.md
```diff
@@ -64,6 +64,9 @@
   types with a same-named import alias.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- Hardened the GitHub workflow generated by `gleam new`
+  ([Willow (GHOST)](https://github.com/ghostdevv))
+
```
khalidbelk, line 69: This shouldn't be under the **Bug fixes** section as @GearsDatapacks suggested. You may want to do this instead: 

```suggestion

### Build tool

- Improved the GitHub workflow generated by `gleam new`
  ([Willow (GHOST)](https://github.com/ghostdevv))
 
```

lpil:
Thank you but unplanned changes are not accepted, as indicated by the checkboxes you have not been able to tick. Please open a discussion with any proposals for changes or additions, once any have been agreed they can then be implemented 🙏

PR #6151, opened 2026-08-10
Added variant narrowing for constants within their module

### LINKED ISSUE #5310: Variant narrowing for constants
Type refinement currently works for constants inside `case` expressions, but only when the pattern explicitly matches the record constructor (not in all patterns). Matching on a constant and explicitly using the record constructor in the pattern allows accessing fields specific to that variant within that specific branch.

Since the value of a module constant is known at compile time, the compiler can guarantee which variant it holds. Therefore, accessing fields that exist on that specific variant should be type-safe globally, without requiring a `case` unwrapping or a specific pattern match.

### Code examples

**Current working behavior (Refinement in specific `case`):**
[Gleam playground](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIADgK4BGABPgJ5kQ0DqcVVANs8ADpI1W7LhH6CWCDtwAUcYqhoBJYgEp+AX3788SGPhoB3SSJoBeVsZkAWAExqk/SrQBmAgLYBDOdJU0+AmgB6QJoAFQALOBhDBCgAa2j8BAATD3oxGhwPGGYjKR4M8WFuGgBaAD4aGytCi3zpAH1fCpj8gDo5fWCaAEEcHAgYeCQAcxpnOAhOZJpkOnDmHT0PYhoPZ0IoGk98HEjRjM0kdRB1IA=)

```gleam
pub type Wibble {
  Wibble
  Wobble(int: Int)
}

const wobble = Wobble(42)

pub fn main() {
  // This works today
  case wobble {
    Wibble -> 24
    Wobble(_) -> wobble.int // Accessing field on the constant after matching
  }
}
```

_(Note: this generates invalid Erlang as reported in #5261, but works for the JavaScript target.)_

**Proposed behavior (Global refinement):**
[Gleam playground](https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIADgK4BGABPgJ5kQ0DqcVVANs8ADpI1W7LhH6CWCDtwAUcYqhoBJYgEp+AX3788SGPhoB3SSJoBeVsZkAWAExqk/SrQBmAgLYBDOdJU0+AmgB6QJoABSgEMgQYCAATBQAVAAs4GBoYJIQKTliaPDcyOG4aKggcDwoYmgByIykIappUmgBrJAQDAXwEEuZqiXrGj318wuL8ODcIADoxQ0sZuXwNEHUgA==)

```gleam
pub type Wibble {
  Wibble
  Wobble(int: Int)
}

const wobble = Wobble(42)

pub fn main() {
  // Proposed: This should compile because 'wobble' is known to be 'Wobble' at compile time.
  wobble.int
}
```
Currently, this produces the following error:

```text
error: Unknown record field
   ┌─ /src/main.gleam:10:10
   │
10 │   wobble.int
   │          ^^^ This field does not exist

    Wibble

It does not have fields that are common across all variants.

```

jamesdolan16:
fixes #5310 

Added functionality to `generalise_module_constant` that captures the inferred variant and then reapplies it to the environments copy of the generalised type. This allows for the inferred variant to carry forward within the module scope that the const was defined.

### DIFF
--- a/compiler-core/src/analyse.rs
+++ b/compiler-core/src/analyse.rs
@@ -1880,7 +1880,17 @@ fn generalise_module_constant(
         deprecation,
         implementations,
     } = constant;
+
+    let inferred_variant = type_.custom_type_inferred_variant();
+
     let type_ = type_::generalise(type_);
+
+    let mut environment_type = type_.clone();
+
+    if let Some(inferred_variant) = inferred_variant {
+        Arc::make_mut(&mut environment_type).set_custom_type_variant(inferred_variant);
+    }
+
     let variant = ValueConstructorVariant::ModuleConstant {
         documentation: doc.as_ref().map(|(_, doc)| doc.clone()),
         location,
@@ -1892,7 +1902,7 @@ fn generalise_module_constant(
     environment.insert_variable(
         name.clone(),
         variant.clone(),
-        type_.clone(),
+        environment_type,
         publicity,
         deprecation.clone(),
     );
--- a/compiler-core/src/type_/tests/custom_types.rs
+++ b/compiler-core/src/type_/tests/custom_types.rs
@@ -223,3 +223,50 @@ fn handle_fish(fish: Fish) {
 "#
     );
 }
+
+#[test]
+fn narrow_const_variant() {
+    assert_module_infer!(
+        "
+pub type Ty {
+  StrVar(str: String)
+  IntVar(int: Int)
+}
+
+pub const foo = IntVar(42)
+
+pub fn main() {
+  foo.int
+}
+",
+        vec![
+          ("IntVar", "fn(Int) -> Ty"),
+          ("StrVar", "fn(String) -> Ty"),
+          ("foo", "Ty"),
+          ("main", "fn() -> Int")
+        ],
+    );
+}
+
+#[test]
+fn const_variant_narrowing_is_not_public() {
+  assert_module_error!(
+    (
+      "other/module",
+      "
+pub type Ty { 
+  IntVar(int: Int)
+  StrVar(str: String)
+}
+pub const foobar = IntVar(42)
+"
+    ),
+    "
+import other/module
+
+pub fn main() {
+  module.foobar.int
+}
+"
+  );
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,10 @@
   expression now points to the entire function call.
   ([mmustafasenoglu](https://github.com/mmustafasenoglu))
 
+- Constants that are assigned a type variant are now narrowed within the module they
+  were defined.
+  ([James Dolan](https://github.com/jamesdolan16))
+
 ### Build tool
 
 - The build tool now stores its build cache in a more compact binary format,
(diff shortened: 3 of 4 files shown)


compiler-core/src/type_/tests/custom_types.rs
```diff
@@ -223,3 +223,50 @@ fn handle_fish(fish: Fish) {
...
+#[test]
+fn narrow_const_variant() {
+    assert_module_infer!(
+        "
+pub type Ty {
+  StrVar(str: String)
+  IntVar(int: Int)
+}
+
+pub const foo = IntVar(42)
```
lpil, line 236: No "foobar" in this codebase please, it's a rather grim WW2 reference.

If you could also remove the "ty", "str" and "var" abbreviations please, either using "type", "string", and "variant", or using nonsense words such as "wibble".

compiler-core/src/type_/tests/custom_types.rs
```diff
@@ -223,3 +223,50 @@ fn handle_fish(fish: Fish) {
...
+    ),
+    "
+import other/module
+
+pub fn main() {
+  module.foobar.int
+}
+"
+  );
+}
```
lpil, line 272: End of file newline missing here

compiler-core/src/type_/tests/custom_types.rs
```diff
@@ -223,3 +223,50 @@ fn handle_fish(fish: Fish) {
 "#
     );
 }
+
```
lpil, line 226: Can you add tests showing the variant inference being used and not being accessible in all other contexts too please 🙏 So record access expressions, record updates, guard record updates, constant record updates, and any I've forgotten.

compiler-core/src/analyse.rs
```diff
@@ -1880,7 +1880,17 @@ fn generalise_module_constant(
...
     } = constant;
+
+    let inferred_variant = type_.custom_type_inferred_variant();
+
     let type_ = type_::generalise(type_);
+
+    let mut environment_type = type_.clone();
+
+    if let Some(inferred_variant) = inferred_variant {
+        Arc::make_mut(&mut environment_type).set_custom_type_variant(inferred_variant);
```
lpil, line 1891: Only clone the `type_` if a variant is inferred, the clone is not used otherwise.

I think we can use `borrow_mut` instead of `make_mut` here too, which never clones.

compiler-core/src/analyse.rs
```diff
@@ -1880,7 +1880,17 @@ fn generalise_module_constant(
         deprecation,
         implementations,
     } = constant;
+
```
lpil, line 1883: Add a detailed comment to this section to explain what it is doing and why please

lpil (commented):
Lovely, thank you!! I've left a few comments inline

compiler-core/src/analyse.rs
jamesdolan16: @lpil thanks for the feedback, just wondering which bit you mean to `borrow_mut` instead of `make_mut`?

PR #6156, opened 2026-08-10
LS: make `Discard unused variable` work for string prefixes

### LINKED ISSUE #6120: "Discard unused variable" code action does not work for string prefix patterns
If you have some code that uses a string prefix pattern that binds the prefix to an unused variable, something like this:
```gleam
case greeting {
  "Hello" as hello <> name -> name
  _ -> "unknown"
}
```
Here, the `hello` variable is unused to the code action is offered. However, if it is applied, this gets produced:
```gleam
case greeting {
  "Hello" as <> name -> name
  _ -> "unknown"
}
```
Which is a syntax error. Instead it should be:
```gleam
case greeting {
  "Hello" <> name -> name
  _ -> "unknown"
}
```

ankddev:
* Closes #6120 


compiler-core/src/ast.rs
```diff
@@ -3650,6 +3667,13 @@ impl<A> HasLocation for Pattern<A> {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct StringPrefixLeftSideAssignment {
+    pub name: EcoString,
+    pub name_location: SrcSpan,
+    pub location: SrcSpan,
+}
```
lpil, line 3694: What are these fields? Could you document them please 🙏 

I suspect the name location span may be partially redundant and should be replaced with a position.

lpil (commented):
Thank you! I've left a tiny note inline.

ankddev:
@lpil Done!

ankddev:
@lpil merged https://github.com/gleam-lang/gleam/pull/6158, which is stacked on top of this one, so closing!

PR #6159, opened 2026-08-10
Fix main

giacomocavalieri:
There were a couple of compilation errors and failing tests on main somehow. This should fix them! I've also changed a little error in the changelog

### DIFF
--- a/hexpm/src/lib.rs
+++ b/hexpm/src/lib.rs
@@ -972,6 +972,7 @@ pub struct Release<Meta> {
     #[serde(alias = "checksum", deserialize_with = "deserialize_checksum")]
     pub outer_checksum: Vec<u8>,
     /// Security advisories affecting this release of the package
+    #[serde(default)]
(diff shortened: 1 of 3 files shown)


hexpm/src/lib.rs
```diff
@@ -972,6 +972,7 @@ pub struct Release<Meta> {
     #[serde(alias = "checksum", deserialize_with = "deserialize_checksum")]
     pub outer_checksum: Vec<u8>,
     /// Security advisories affecting this release of the package
+    #[serde(default)]
     pub security_advisories: Vec<SecurityAdvisory>,
```
giacomocavalieri, line 976: The original PR introducing this error had no tests so I'm not really sure if this is the correct fix or not! @ankddev any ideas?

ankddev: I think that it's fine, thank you!

ankddev (approved):

lpil (approved):
Oops, thank you!

PR #6163, opened 2026-08-11
Get ready to switch to the abstract format binary generator

giacomocavalieri:
This PR does a couple of things in preparation for the switch to the abstract format binary generator:

- The `SrcSpan` and `LineNumbers` definitions have been moved to their own separate module and they're no longer in code (this way the generator can depend on them)
- All the generator methods now take an additional location that will be used for annotations. The source printer just ignores them
- The code for echo is no longer in a template file but is generated using the Erlang generator like any other piece of code


lpil (approved):
Thank you Jak!!! I must admit, my eyes glazed over a few times reading the new echo module 😁

giacomocavalieri:
It's definitely not the nicest 😆

PR #6169, opened 2026-08-12
Binary abstract format

giacomocavalieri:
This PR closes #3705 
With this PR the compiler now produces binary `.abstr` files instead of a textual output!

I'm opening a Pull Request but there's still a couple of things that come to mind we need to take care of, let's start with the easy ones:

- [x] Update the CHANGELOG, from what I could measure locally this also makes the compiler faster overall. I'll make sure to measure and include these results as well.
- [x] Leave an option to produce textual files in the compiler. The playground and tour will need this to still display the compiled Erlang. I don't know what this would look like though. I'd love to hear any ideas on what this flag would look like and do.
- [ ] Test the source span annotations are correct. How do we do that? Do we write a third builder that produces a source map to be snapshotted? This will probably be a big amount of work and probably worth a separate PR to keep things manageable.
- [ ] Add more language tests. I would like to have way more language runtime tests to make sure we're not breaking stuff with the binary generator.

---


inoas-nbw:
>  With this PR the compiler now produces binary .abstr files instead of a textual output!

Does this mean no more readable Erlang files?
Would it be worth it to keep them around to be able to read/understand the underlying erlang and/or maybe use it to generate API docs for Erlang users?

giacomocavalieri:
> Does this mean no more readable Erlang files?

No, as the PR says we'd still have an option to produce textual Erlang files for display. Those are still useful for debugging and the Gleam playground

compiler-cli/src/publish.rs
```diff
@@ -808,7 +808,7 @@ pub struct ReleaseMetadata<'a> {
     version: &'a Version,
     description: &'a str,
     source_files: &'a [Utf8PathBuf],
-    generated_files: &'a [(Utf8PathBuf, String)],
+    generated_files: &'a [(Utf8PathBuf, Vec<u8>)],
```
lpil: Bytecode cannot be published to Hex as it's not portable, only source files can be published.

compiler-cli/templates/gleam@@compile.erl
```diff
@@ -90,15 +90,53 @@ worker_loop(Parent, Out) ->
...
+                                  erlang:send(Parent, failed)
+                              end;
+                          error ->
+                              log({failed, Module}),
+                              erlang:send(Parent, failed)
+                      end
+                    catch error:Error ->
+                      log({failed, Error, Module}),
+                      erlang:send(Parent, failed)
+                    end;
```
lpil: Jak you have written code with 10 levels of indentation and mixed 2 and 4 space indentation 😅

compiler-core/src/build/package_compiler.rs
```diff
@@ -392,8 +395,21 @@ where
 
         io.mkdir(&build_dir)?;
 
+        // We don't want to copy the precompiled erlang files of deps over, so
```
lpil:
```suggestion
        // We don't want to copy the precompiled Erlang files of deps over, so
```

compiler-core/src/codegen.rs
```diff
@@ -56,12 +57,20 @@ impl<'a> Erlang<'a> {
...
+        let name = format!("{erl_name}.abstr");
         let path = self.build_directory.join(&name);
         let line_numbers = LineNumbers::new(&module.code);
-        let output = erlang::module(&module.ast, line_numbers, root);
+
+        // We create the builder and set the line numbers so the ast will be
+        // properly annotated.
+        let erlang_module_name = ErlangModuleName::new(&module.name);
+        let mut builder = ErlangBinaryBuilder::new(Some(erlang_module_name));
+        builder.set_line_numbers(&line_numbers);
```
lpil: How come this isn't an argument to `new`?

lpil (commented):
I was nosy 👀

compiler-cli/src/publish.rs
giacomocavalieri: Do you think we should stop including generated files at all?

compiler-cli/templates/gleam@@compile.erl
giacomocavalieri: That's what I get for not using a formatter 😭

compiler-core/src/codegen.rs
giacomocavalieri: I tried doing that but that would force it into the Builder trait forcing a new lifetime in the builder definition. I was lazy and didn't want to update all the hundreds of references to that type with a new lifetime 😂

I can change it though, just let me know!

compiler-cli/src/publish.rs
lpil: Wouldn't that be a breaking change?

I don't know if anyone is using those files though.

compiler-core/src/codegen.rs
lpil: I don't think we have any reason to make `new` part of the trait, do we? The construction of the trait object is done prior to the code that is generic over the two Erlang generation backends.

giacomocavalieri: You're right, fixed!

changelog/v1.8.md
```diff
@@ -95,22 +95,21 @@ Dedicated to the memory of Len Pilfold.
   Usage: gleam deps tree [OPTIONS]
 
   Options:
-    -p, --package <PACKAGE>  Package to be used as the root of the tree
-    -i, --invert <PACKAGE>   Invert the tree direction and focus on the given package
-    -h, --help               Print help
+  -p, --package <PACKAGE> Package to be used as the root of the tree
```
lpil: Whitespace has been mistakenly removed in this file!

compiler-core/src/build/project_compiler.rs
```diff
@@ -57,6 +57,7 @@ pub struct Options {
     pub warnings_as_errors: bool,
     pub root_target_support: TargetSupport,
     pub no_print_progress: bool,
+    pub building_hex_tarball: bool,
```
lpil: Would be nice to use an enum rather than a bool here. Perhaps the existing `ErlangOutput`?

test-package-compiler/src/snapshots/test_package_compiler__generated_tests__variable_or_module.snap
```diff
@@ -2,50 +2,24 @@
 source: test-package-compiler/src/generated_tests.rs
 expression: "./cases/variable_or_module"
 ---
+//// /out/lib/the_package/_gleam_artefacts/main.abstr
+<1083 byte binary>
```
lpil: lol these snapshots are a bit questionable now 😁

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        self.etf.small_tuple(4);
+        self.etf.atom("attribute");
+        self.etf.usize(0);
+        self.etf.atom("module");
+        self.etf.atom(&module_name.0);
+    }
+
+    fn into_output(mut self) -> Vec<u8> {
+        // Before returning the binary representation we've gotta make sure
+        // we're ending the
```
lpil: oooh the suspense! Ending the what?

Knowing Jak I presume it is "coffee break not too early."

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+
+    type TypeSpec = Vec<EcoString>;
+
+    fn module_attribute(&mut self, module_name: ErlangModuleName) {
+        self.new_top_level_form();
+        // The EAF representation of `-module(Module).` is
+        // `{attribute, ANNOTATION, module, Module}`.
+        self.etf.small_tuple(4);
+        self.etf.atom("attribute");
+        self.etf.usize(0);
```
lpil:
```suggestion
        self.annotation(None);
```
Perhaps this?

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+    type UnionType = erlang_term_format::ListEnder;
+    type FunctionSpec = erlang_term_format::ListEnder;
+
+    type Block = erlang_term_format::ListEnder;
+
+    type FunctionTypeArguments = (erlang_term_format::ListEnder, erlang_term_format::ListEnder);
+
+    type TypeSpec = Vec<EcoString>;
+
+    fn module_attribute(&mut self, module_name: ErlangModuleName) {
```
lpil:
```suggestion
    fn module_declaration(&mut self, module_name: ErlangModuleName) {
```

`-module(...).` is a module attribute, but that's too vague a name as any `-NAME(...)` is a module attribute. The Erlang docs call it the module definition: https://www.erlang.org/doc/system/modules.html#pre-defined-module-attributes

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        // Before returning the binary representation we've gotta make sure
+        // we're ending the
+        if self.current_position().is_some() {
+            invalid_code_for_position!(self, "close module")
+        }
+        self.etf.end_list(self.module_ender, self.top_level_forms);
+        self.etf.into_vec()
+    }
+
+    fn export_attribute<Name: AsRef<str>>(
```
lpil, line 4005: Is there a reason all these `*_attribute` functions don't call a `module_attribute` function that does the shared parts?

```rust
self.module_attribute("export", None);
```

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        let options = self.etf.start_list();
+        let mut count = 0;
+        for argument in arguments {
+            count += 1;
+            self.etf.atom(argument);
+        }
+        self.etf.end_list(options, count);
+    }
+
+    fn file_attribute(&mut self, file: &str, line: u32) {
```
lpil: Now that we have annotations let's generate just a single file attribute for the whole module instead of one per function.

Maybe we should omit it entirely when compiling to source?

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        self.etf.small_tuple(4);
+        self.etf.atom("attribute");
+        self.annotation(None);
+        self.etf.atom("record");
+
+        self.etf.small_tuple(2);
+        self.etf.atom(record_name);
+        let record_fields = self.etf.start_list();
+
+        // The reps are left to be added separately and the list will be closed
```
lpil: What are reps?

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+            expected: ExpectedRecordFieldItem::Name,
+        });
+
+        // A :: T
+        //   becomes
+        // {typed_record_field,{record_field,ANNO,Rep(A)},Rep(T)}
+        self.etf.small_tuple(3);
+        self.etf.atom("typed_record_field");
+
+        // We stab out both tuples so that the only two things left to generate
```
lpil: Not sure what "stab out" means

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        self.etf.atom("attribute");
+        self.annotation(None);
+        self.etf.atom("spec");
+        self.etf.small_tuple(2);
+
+        {
+            self.etf.small_tuple(2);
+            self.etf.atom(name);
+            self.etf.usize(arity);
+        }
```
lpil: How come this is in a block?

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+
+        // After the name, we are expecting the representation of the type `T`,
+        // and only after that we can generate the list with the type
+        // parameters.
+        // So we will carry those type parameters around waiting for the type
+        // spec to be closed.
+        type_parameters
+            .into_iter()
+            // TODO) This doesn't feel great, probably could make the function take
+            // a vec of EcoStrings since that's the only way it is called.
```
lpil: Sounds like the right thing to do!

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1938 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        self.etf.atom("integer");
+        self.annotation(location);
+        self.etf.bigint(number);
+    }
+
+    /// This pushes the representation of a literal atom according to the Erlang
+    /// Abstract Format.
+    /// How is this different from:
+    /// - `self.builder.atom()`: the `erlang_term_format::Builder` can be used
+    ///   to push _bare_ atoms, this function pushes the _representation_ of an
```
lpil: This says "atom representation is different from atom because it's not an atom, it's an atom representation". It's circular, and I don't know what an atom representation is, or a bare atom.

From reading the uses I think a "representation" is when an atom or a float is not a sub-expression in the source code, but instead is a detail of the EAF representation of some other expression node. If that is right then it would be great to have that documented and include an example.

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1939 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+        self.etf.atom("remote_type");
+        self.annotation(None);
+
+        let outer_list = self.etf.start_list();
+        // Crucially, when it comes to a remote type, we require the
+        // _representation_ of the module and name. Those are atoms so we want
+        // the Erlang Abstract Format representation of an atom, NOT a bare
+        // Erlang atom. That's why here we call `self.atom` and not `self.etf.atom`:
+        // - `self.atom` produces the _representation_ of an atom
+        // - `self.etf.atom` produces a bare atom
```
lpil: I think this would be clearer with this change in terminology:

bare atom -> ETF atom
atom representation -> abstract forms atom

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1939 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+            // Crucially, when it comes to a call, we require the
+            // _representation_ of the module and name of the called thing.
+            // Those are atoms so we want the Erlang Abstract Format
+            // representation of an atom, NOT a bare Erlang atom.
+            // That's why here we call `self.atom` and not `self.etf.atom`:
+            // - `self.atom` produces the _representation_ of an atom
+            // - `self.etf.atom` produces a bare atom
+            self.atom_representation(None, &module.0);
+            self.atom_representation(None, function);
+        }
```
lpil: How come this is in a block?

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1939 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+            self.etf.small_tuple(4);
+            self.etf.atom("remote");
+            self.annotation(Some(location));
+            // Crucially, when it comes to a call, we require the
+            // _representation_ of the module and name of the called thing.
+            // Those are atoms so we want the Erlang Abstract Format
+            // representation of an atom, NOT a bare Erlang atom.
+            // That's why here we call `self.atom` and not `self.etf.atom`:
+            // - `self.atom` produces the _representation_ of an atom
+            // - `self.etf.atom` produces a bare atom
```
lpil: Same as with the previous instance of this comment

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1939 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+                self.etf.atom("function");
+                // Crucially, when it comes to a function reference, we require
+                // the _representation_ of the module, the name, and the arity
+                // That's why here we call `self.atom_representation` and
+                // `self.int_representation`, not `self.etf.atom` and
+                // `self.etf.usize`:
+                // - `self.atom_representation` produces the _representation_ of an atom
+                // - `self.etf.atom` produces a bare atom
+                // - `self.int_representation` produces the _representation_ of an int
+                // - `self.etf.usize` produces a bare int
```
lpil: Same as with the previous instances of this comment.

I'm not sure we really need this comment to be honest- I think it would be clear from reading the code that an EAF atom is expected, otherwise we would not be calling that method.

erlang-generation/src/lib.rs
```diff
@@ -3765,3 +3754,1939 @@ fn is_erlang_reserved_word(name: &str) -> bool {
...
+    }
+
+    fn float(&mut self, location: SrcSpan, value: f64) {
+        self.new_expression();
+        self.float_representation(Some(location), value);
+    }
+
+    fn atom(&mut self, location: SrcSpan, name: &str) {
+        self.new_expression();
+        self.atom_representation(Some(location), name);
```
lpil: I feel the use of the term "representation" isn't very clear, as all the methods are for representations.

My hunch is that this method name suffix was added to distinguish it from the un-suffixed versions (e.g. `atom` and `atom_representation`). I think it may be better to add an `_expression` suffix to the expression method instead, to match the `_pattern` suffix used for patterns.

lpil (commented):
Wow amazing!! Left a smattering of questions, and made some small edits.

I think a concurrent PR has introduced some failing tests?

test-package-compiler/src/snapshots/test_package_compiler__generated_tests__variable_or_module.snap
giacomocavalieri: Do you think we should remove them?

erlang-generation/src/lib.rs
giacomocavalieri: Yeah much better!

erlang-generation/src/lib.rs
giacomocavalieri: I didn't think about it! Fixing it

erlang-generation/src/lib.rs
giacomocavalieri: Yeah good idea!

test-package-compiler/src/snapshots/test_package_compiler__generated_tests__variable_or_module.snap
giacomocavalieri: Ah it was actually helpful to see how the size of the output changed as I changed the binary generator, I think we should keep them

compiler-core/src/build/project_compiler.rs
giacomocavalieri: Oh yeah much nicer!

lpil (approved):
Ahhh exciting!!!! Here we go!!!

PR #6170, opened 2026-08-12
fix: LSP panic on client disconnect

### LINKED ISSUE #6144: `gleam lsp` panics when the client disconnects
Closing stdin without `shutdown`/`exit` prints "Fatal compiler bug!", asks for
a crash report, and exits 101.

```
echo '' | gleam lsp
```

```
error: Fatal compiler bug!

Panic: language-server/src/server.rs:636
	LSP initialize: ProtocolError("disconnected channel", true)
```

After a completed `initialize` the panic shows at `messages.rs:201`,
`Receiving LSP message: RecvError`.

Sending `shutdown` then `exit` first exits 0 with no panic.

Gleam 1.18.1 on macOS.

ssenthilnathan3:

Closes #6144

Replaces `.expect()` on LSP transport operations (initialize, recv, send) with a dedicated `Error::LspTransport` variant, `Next::Stop`, and `let _ =` to exit cleanly on client disconnect.

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -197,6 +197,12 @@ pub enum Error {
     #[error("{error}")]
     GitInitialization { error: String },
 
+    #[error("Failed to send LSP message: {error}")]
+    LspMessageSendFailed { error: String },
+
+    #[error("Failed to receive LSP message: {error}")]
+    LspMessageReceiveFailed { error: String },
+
     #[error("io operation failed")]
     StandardIo {
         action: StandardIoAction,
@@ -1950,6 +1956,42 @@ with `gleam hex authenticate`."
                 );
                 vec![Diagnostic {
                     title: "Failed to initialize git repository".into(),
+                    text,
+                    hint: None,
+                    level: Level::Error,
+                    location: None,
+                }]
+            }
+
+            Error::LspMessageSendFailed { error } => {
+                let text = wrap_format!(
+                    "An error occurred when the Gleam language server \
+attempted to send a message to the client:
+
+    {error}
+
+Is your editor still running and connected to the language server?"
+                );
+                vec![Diagnostic {
+                    title: "Failed to send LSP message".into(),
+                    text,
+                    hint: None,
+                    level: Level::Error,
+                    location: None,
+                }]
+            }
+
+            Error::LspMessageReceiveFailed { error } => {
+                let text = wrap_format!(
+                    "An error occurred when the Gleam language server \
+attempted to receive a message from the client:
+
+    {error}
+
+Is your editor still running and connected to the language server?"
+                );
+                vec![Diagnostic {
+                    title: "Failed to send LSP message".into(),
                     text,
                     hint: None,
                     level: Level::Error,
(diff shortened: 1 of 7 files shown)


lpil (commented):
Thank you! These changes mean we are silently discarding errors, but we never do that. Any time interacting with the socket fails an error must be returned.

Could you also write a snapshot test that shows how the new error variant would be shown to the user. Thanks again!

ssenthilnathan3:
> Thank you! These changes mean we are silently discarding errors, but we never do that. Any time interacting with the socket fails an error must be returned.
> 
> Could you also write a snapshot test that shows how the new error variant would be shown to the user. Thanks again!

thanks for the review! i had a question about the four `sender.send()` sites. they're currently using `let _ =` because they're in fire-and-forget notification methods that don't return `Result`.

returning `Error::LspTransport` from these would require threading `Result` through ~8 function signatures in the call chain (`publish_diagnostics`, `publish_feedback`, `handle_notification`, `handle_message`, `run`, etc.).

should these notification paths also propagate the error, or is it acceptable to keep them fire-and-forget?

lpil:
No, it is never acceptable to discard an error, they must always be propagated to the user.

ssenthilnathan3:
> No, it is never acceptable to discard an error, they must always be propagated to the user.

Understood! Thanks for the prompt reply.

lpil:
Thank you!!

language-server/src/messages.rs
```diff
@@ -197,19 +197,24 @@ impl MessageBuffer {
...
+        // or if there is just a pause in typing.
         let message = match message {
             Some(message) => message,
             None => {
-                // A compile please message it added in the instance of this
+                // If the buffer was empty, the client disconnected without sending
+                // a shutdown request. Stop the server gracefully.
+                if was_empty {
+                    return Next::Stop;
+                }
```
lpil: Instead of doing this please immediately return an error in the empty-message-vector branch. This makes the code a bit simpler, and it also stops us silently discarding the error.

Put back the comment too, some useful information has been lost!

compiler-core/src/error.rs
```diff
@@ -197,6 +197,9 @@ pub enum Error {
     #[error("{error}")]
     GitInitialization { error: String },
 
+    #[error("{error}")]
+    LspTransport { error: String },
```
lpil: Let's make this more descriptive, as it's a bit sparse currently.


```suggestion
    #[error("Failed to send LSP message: {error}")]
    LspMessageSendFailed { error: String },
    
    #[error("Failed to receive LSP message: {error}")]
    LspMessageReceiveFailed { error: String },
```

lpil (commented):
Lovely, thank you! I've left a few notes inline.

Tag me when you're ready for another review.

lpil:
P.S. rebase on main for any unrelated build fixes

ssenthilnathan3:
@lpil I addressed the changes. should be ready for another review now. thanks!

lpil (approved):
Lovely!! Thank you!

I've pushed a commit to change the error message a little bit. Would you mind adding a changelog entry for this bug fix please. To say that the LS will no longer crash if the client disconnects unexpectedly 🙏

lpil (approved):
Fab work!! Thank you again!!

PR #6172, opened 2026-08-12
add -0.0 to the v2 list

giacomocavalieri:
We were discussing this earlier and decided that this breaking change would be something for v2, I've added it to the list so we won't forget!


lpil (approved):
Thanks!!

PR #6174, opened 2026-08-14
compile-package generates app file listing all modules, even colocated erlang and elixir files  

Papipo:
This PR addresses the second point of https://github.com/gleam-lang/gleam/issues/5945
(The first one is already working (compile-package does compile .ex and .erl files))

My first attempt at this was to use the source file names but those don't get copied in subsequent calls to compile-package if they haven't been modified since the last time. So I thought that anything that wasn't a gleam module in ebin must be erlang or elixir, no? :D


compiler-cli/src/compile_package.rs
```diff
@@ -36,8 +40,13 @@ pub fn command(options: CompilePackage) -> Result<()> {
         io.initialise_beam_compiler()?;
     }
 
+    let app_file = (!options.skip_beam_compilation).then(|| ErlangAppCodegenConfiguration {
+        include_dev_deps: false,
+        package_name_overrides: HashMap::new(),
+    });
```
lpil: I don't easily understand what this code does, could you write it more simply with `if` please 🙏

compiler-cli/src/compile_package.rs
```diff
@@ -36,8 +40,13 @@ pub fn command(options: CompilePackage) -> Result<()> {
...
+
     let target = match options.target {
-        Target::Erlang => TargetCodegenConfiguration::Erlang { app_file: None },
+        Target::Erlang => TargetCodegenConfiguration::Erlang { app_file },
```
lpil: Is `None` ever used now? I think it was only ever used by the package compiler, so now we can remove it entirely.

compiler-cli/src/compile_package.rs
```diff
@@ -36,8 +40,13 @@ pub fn command(options: CompilePackage) -> Result<()> {
         io.initialise_beam_compiler()?;
     }
 
+    let app_file = (!options.skip_beam_compilation).then(|| ErlangAppCodegenConfiguration {
+        include_dev_deps: false,
+        package_name_overrides: HashMap::new(),
```
lpil: The overrides are missing here.

compiler-core/src/build/package_compiler/tests.rs
```diff
@@ -88,3 +88,112 @@ pub fn same_package_defining_duplicate_module() {
 
     insta::assert_snapshot!(insta::internals::AutoName, output);
 }
+
+fn compile_once(fs: &InMemoryFileSystem, compile_beam_bytecode: bool) {
```
lpil: Can we not use the existing compilation code in this module?

compiler-core/src/build/package_compiler.rs
```diff
@@ -430,6 +434,36 @@ where
         Ok(())
     }
 
+    /// Every native (Erlang/Elixir) module that currently has a compiled
+    /// `.beam` file in the `ebin` directory but is not one of the Gleam
+    /// modules just compiled or read from cache.
```
lpil: I don't understand this sentence, is it missing a word somewhere?

compiler-core/src/io/memory.rs
```diff
@@ -110,6 +112,32 @@ impl InMemoryFileSystem {
...
+    pub fn write_elixir_module(&self, path: &Utf8Path, module_name: &str, body: &str) {
+        let code = format!("defmodule {module_name} do\n{body}\nend\n");
+        self.write(&Utf8Path::new("/src").join(path), &code)
+            .expect("write elixir module");
+        let _ = self
+            .external_module_names
+            .deref()
+            .borrow_mut()
+            .insert(path.to_path_buf(), format!("Elixir.{module_name}"));
+    }
```
lpil: These have nothing to do with an in memory file system generally, they are specific to some particular test. Please remove these.

CHANGELOG.md
```diff
@@ -116,6 +116,10 @@
   ([Amr Kadry](https://github.com/Amrkadry) and
   ([Louis Pilfold](https://github.com/lpil))
 
+- compile-package now generates a .app file in the ebin folder, listing all
+  .beam modules (this includes erlang and elixir ones).
+  ([Rodrigo Álvarez](https://github.com/Papipo))
```
lpil: Backticks around commands and put `gleam ` before it. No need to say the modules bit (as that is what a .app file is), but if that section was to be in there then Erlang and Elixir should be capitalised.

compiler-core/src/io/memory.rs
```diff
@@ -28,6 +28,7 @@ use camino::{Utf8Path, Utf8PathBuf};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct InMemoryFileSystem {
     files: Rc<RefCell<HashMap<Utf8PathBuf, InMemoryFile>>>,
+    external_module_names: Rc<RefCell<HashMap<Utf8PathBuf, String>>>,
```
lpil: This field doesn't seem to be used anywhere, it can be removed.

compiler-core/src/build/package_compiler.rs
```diff
@@ -430,6 +434,36 @@ where
...
+    ) -> Result<Vec<EcoString>, Error> {
+        let gleam_module_names: HashSet<EcoString> = modules
+            .iter()
+            .map(Module::erlang_name)
+            .chain(cached_module_names.iter().map(module_erlang_name))
+            .collect();
+        let entrypoint_name: EcoString = format!("{}@@main", self.config.name).into();
+
+        Ok(
+            files_with_extension(&self.io, &self.out.join("ebin"), "beam")
```
lpil, line 459: Now that we are looking at the filesystem we can remove all the other logic and do `files_with_extension` for all .beam modules, regardless of their origin. Otherwise we are tracking this additional information in order to remove them from that list and then to put them back in again.

compiler-core/src/build/package_compiler.rs
```diff
@@ -410,15 +414,15 @@ where
         // version and not the newly compiled version.
         Erlang::new(&build_dir, &include_dir).render(io.clone(), modules, self.root)?;
 
-        let native_modules: Vec<EcoString> = if self.compile_beam_bytecode {
+        if self.compile_beam_bytecode {
             written.extend(modules.iter().map(Module::compiled_erlang_path));
-            self.compile_erlang_to_beam(&written)?
+            let _ = self.compile_erlang_to_beam(&written)?;
```
lpil, line 420: Remove the `let _ = ` please.

lpil (commented):
Thank you!! Looking good, but I think we can simplify this all a bunch. I've left some notes inline.

Papipo:
Thanks for the review @lpil, my lack of rust-fu shows here I am afraid.

compiler-core/src/build/package_compiler.rs
Papipo: I don't understand it either, is a leftover from a previous iteration, removing.

lpil:
Is this ready for a review? If so please press the "ready for review" button and tag me 🙏

compiler-core/src/build/package_compiler.rs
Papipo: Should be fixed now

Papipo:
> Is this ready for a review? If so please press the "ready for review" button and tag me 🙏

Almost, I need to fix the fact that package name overrides are empty. The rest has been tackled.

Papipo:
I have rebased after fixing all the mentioned issues @lpil

compiler-core/src/build/package_compiler.rs
lpil: This hasn't been done 🔴

compiler-core/src/build/package_compiler.rs
```diff
@@ -409,29 +412,57 @@ where
...
+            // `app_file_config` can be `Some` even when `compile_beam_bytecode`
+            // is false, in which case nothing is ever written to `ebin` (and
+            // anything already there from an earlier build is stale). So we
+            // only trust the `ebin` scan when bytecode compilation actually
+            // ran this build; otherwise we fall back to the modules compiled
+            // in memory this run.
+            let compiled_modules = if self.compile_beam_bytecode {
+                self.modules_in_ebin()?
+            } else {
+                modules.iter().map(Module::erlang_name).collect()
```
lpil, line 435: This is incorrect in this case, due to the behaviour that the comment above describes.

Is there a situation in which it should be possible to have this combination? Or should we redesign the data structure here to make it impossible?

compiler-core/src/io/memory.rs
```diff
@@ -455,12 +455,37 @@ impl CommandExecutor for InMemoryFileSystem {
...
         &self,
-        _out: &Utf8Path,
+        out: &Utf8Path,
         _lib: &Utf8Path,
-        _modules: &HashSet<Utf8PathBuf>,
+        modules: &HashSet<Utf8PathBuf>,
         _stdio: Stdio,
     ) -> Result<Vec<String>, Error> {
-        Ok(Vec::new()) // Always succeed.
+        // Always succeed, pretending to have compiled every given module and
```
lpil, line 463: This doesn't seem to be used any more. As that is the case please make the `compile_beam` trait method return `Result<(), Error>`

lpil (commented):
Thank you!! I've left some notes inline.

compiler-core/src/build/package_compiler.rs
Papipo: Ah, I didn't remove it because the compiler complains about an unused Result.

Papipo: I see that this will be fixed once I remove the returned vector thingy

compiler-core/src/build/package_compiler.rs
Papipo: @lpil I knew there was something fishy here.

I think there is need for refactor, but to double check: if you compile to beam, you always want the app file, right? So it's always either compile to beam or not. And the app file will be generated in case there was beam compilation. There is no beam compilation + no app file, is there?

Papipo: Lemme know if you like how I tackled this in the last commit

compiler-cli/src/beam_compiler.rs
```diff
@@ -71,7 +71,7 @@ impl BeamCompilerInstance {
             match buf.trim() {
                 "gleam-compile-result-ok" => {
                     // Return Ok with the accumulated modules
-                    return Ok(accumulated_modules);
+                    return Ok(());
```
lpil, line 73: The comment above this is wrong now

compiler-cli/src/compile_package.rs
```diff
@@ -36,10 +39,24 @@ pub fn command(options: CompilePackage) -> Result<()> {
     if options.target.is_erlang() && !options.skip_beam_compilation {
         io.initialise_beam_compiler()?;
     }
-
+    let app_file = match options.skip_beam_compilation {
+        true => None,
+        false => {
+            let package_name_overrides = if paths.manifest().exists() {
+                let package_manifest = crate::dependencies::read_manifest_from_disc(&paths)?;
```
lpil: The manifest doesn't exist when using `compile-package` as it is created by the Gleam build tool, which this is not using.

We are going to have to make this a new flag for the CLI, so the build tool can pass in this data.

compiler-core/src/codegen.rs
```diff
@@ -140,14 +138,13 @@ impl<'a> ErlangApp<'a> {
...
-            .iter()
-            .map(|m| m.erlang_name())
-            .chain(cached_module_names.iter().map(module_erlang_name))
-            .chain(native_modules)
-            .unique()
+        // `compiled_modules` is resolved by the caller: either an `ebin` scan
+        // (which already includes modules left over from a previous build
+        // that weren't recompiled this time, so a warm rebuild doesn't shrink
+        // the `modules` list, see #5834) or the in-memory module list when no
+        // bytecode compilation happened this run.
```
lpil: I think this comment is out of date now. Probably could be removed entirely? I can't tell what the read is supposed to take from it.

compiler-core/src/io/memory.rs
```diff
@@ -455,12 +455,37 @@ impl CommandExecutor for InMemoryFileSystem {
...
+        let mut compiled = Vec::with_capacity(modules.len());
+        for path in modules {
+            let stem = path.file_stem().unwrap_or_default();
+            let name = if path.extension() == Some("ex") {
+                let camel_case = crate::strings::to_upper_camel_case(stem);
+                let mut chars = camel_case.chars();
+                let capitalised = match chars.next() {
+                    Some(first) => format!("{}{}", first.to_ascii_uppercase(), chars.as_str()),
+                    None => String::new(),
+                };
```
lpil, line 479: Isn't this a no-op? It looks to be doing what `to_upper_camel_case` already returned?

lpil (commented):
Thank you!! Left a handful of notes inline

compiler-cli/src/compile_package.rs
Papipo: Oh no 🤪

compiler-core/src/codegen.rs
Papipo: yeah, removing

compiler-core/src/io/memory.rs
Papipo: that function doesn't upcase the first char, which would be PascalCase

compiler-cli/src/compile_package.rs
Papipo: @lpil I have added the flag, let me know if it's ok this way

compiler-core/src/io/memory.rs
lpil: It does do that.

compiler-cli/src/beam_compiler.rs
```diff
@@ -70,8 +70,7 @@ impl BeamCompilerInstance {
         while let (Ok(_), Ok(None)) = (self.stdout.read_line(&mut buf), self.process.try_wait()) {
             match buf.trim() {
                 "gleam-compile-result-ok" => {
-                    // Return Ok with the accumulated modules
-                    return Ok(accumulated_modules);
+                    return Ok(());
```
lpil, line 72: The `accumulated_modules` value is no longer used, please remove it.

The Erlang compiler also needs to be updated to no longer print the names of the modules.

compiler-cli/src/lib.rs
```diff
@@ -782,6 +782,15 @@ pub struct CompilePackage {
     /// Skip Erlang to BEAM bytecode compilation
     #[arg(long = "no-beam")]
     skip_beam_compilation: bool,
+
+    /// Overrides for a dependency's OTP application name, in the form
+    /// `package=otp_app`.
+    ///
+    /// Required for any dependency whose OTP application name differs from
+    /// its Gleam package name. May be given multiple times.
```
lpil:
```suggestion
    /// its package name. May be given multiple times.
```

compiler-cli/src/lib.rs
```diff
@@ -782,6 +782,15 @@ pub struct CompilePackage {
...
+    ///
+    #[arg(verbatim_doc_comment, long = "otp-app-override")]
+    otp_app_overrides: Vec<String>,
```
lpil: Parse this data with the rest of the command line arguments please.

```suggestion
    otp_app_overrides: HashMap<String, String)>,
```

This clap method will help, you can use it in the clap attribute on this field. https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser

compiler-core/src/build/package_compiler.rs
```diff
@@ -26,7 +26,10 @@ use crate::{
     },
     codegen::{Erlang, ErlangApp, JavaScript, TypeScriptDeclarations},
     config::PackageConfig,
-    io::{BeamCompilerIO, CommandExecutor, FileSystemReader, FileSystemWriter, Stdio},
+    io::{
+        BeamCompilerIO, CommandExecutor, FileSystemReader, FileSystemWriter, Stdio,
+        files_with_extension,
```
lpil: Do not use unqualified imports for functions please.

compiler-core/src/build/package_compiler.rs
```diff
@@ -233,20 +234,15 @@ where
...
-    ) -> Result<Vec<EcoString>, Error> {
+    fn compile_erlang_to_beam(&mut self, modules: &HashSet<Utf8PathBuf>) -> Result<(), Error> {
         if modules.is_empty() {
             tracing::debug!("no_erlang_to_compile");
-            return Ok(Vec::new());
+            Ok(())
+        } else {
+            tracing::debug!("compiling_erlang");
+            self.io
+                .compile_beam(self.out, self.lib, modules, self.subprocess_stdio)
```
lpil: Undo this change where `else {` was added please

compiler-core/src/io/memory.rs
lpil: Remove this please.

compiler-core/src/error.rs
```diff
@@ -380,6 +380,9 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
     #[error("The --javascript-prelude flag must be given when compiling to JavaScript")]
     JavaScriptPreludeRequired,
 
+    #[error("Invalid --otp-app-override value `{input}`")]
+    InvalidOtpAppOverride { input: EcoString },
```
lpil: Remove this please, clap is used for CLI parsing.

test/compile_package1/Makefile
```diff
@@ -2,5 +2,6 @@
 build:
 	rm -rf out1 out2
 	cargo run -- compile-package --package app1 --target erlang --out out2 --lib .
-	cargo run -- compile-package --package app2 --target erlang --out out2 --lib .
+	cargo run -- compile-package --package app2 --target erlang --out out2 --lib . --otp-app-override app1=one_app
```
lpil: Undo this change please. The package compiler is tested in `test-package-compiler/`. Please add new tests there.

test/compile_package1/app2/gleam.toml
```diff
@@ -3,3 +3,6 @@
 
 name = "app2"
 version = "1.0.0"
+
+[dependencies]
+app1 = { path = "../app1" }
```
lpil: Undo this change please.

lpil (commented):

lpil:
I am editing the compile-package-tests package to make it easier to test with configuration. I will drop a message on your other PR when that is ready

compiler-core/src/io/memory.rs
Papipo: If I remove those lines I get this on the tests:

```
-old snapshot
+new results
────────────┬───────────────────────────────
    1     1 │ {application, a_package, [
    2     2 │     {vsn, "0.1.0"},
    3     3 │     {applications, []},
    4     4 │     {description, ""},
    5       │-    {modules, ['Elixir.SomeElixir',
          5 │+    {modules, ['Elixir.someElixir',
    6     6 │                a_module,
    7     7 │                some_erl]},
    8     8 │     {registered, []}
    9     9 │ ]}.
```

Papipo:
@lpil everything has been addressed except the camelCase thing because the function does not Upcase the first letter.

I finished this one before rebasing the other, so I am not sure if I should tackle testing differently.

lpil:
There's a lot of conflicts! Please rebase on main and tag me for a review 🙏

compiler-core/src/io/memory.rs
```diff
@@ -455,12 +455,37 @@ impl CommandExecutor for InMemoryFileSystem {
...
+        // Always succeed, pretending to have compiled every given module and
+        // reporting back the name it would be compiled to, the same way the
+        // real BEAM compiler reports back the name of each module it
+        // compiled. We also write out an (empty) `.beam` file per module,
+        // mirroring the real compiler, since callers may look for these in
+        // `ebin` afterwards.
+        let ebin = out.join("ebin");
+        let mut compiled = Vec::with_capacity(modules.len());
+        for path in modules {
+            let stem = path.file_stem().unwrap_or_default();
```
lpil: This only uses the file stem, but in Elixir the whole path is typically used. `app/login/user.ex` would be `Elixir.App.Login.User`.

compiler-core/src/io/memory.rs
lpil: Then fix the bug in the function!

Papipo: @lpil there is no bug, that's camelCase. If we want PascalCase, I can add a new function with this code.

Papipo: I was confused by the name of the function, the "upper" prefix is weird given we could just have `to_pascal_case`.

Papipo:
@lpil all done

compiler-cli/src/lib.rs
```diff
@@ -786,6 +786,25 @@ pub struct CompilePackage {
...
+    /// `package=otp_app`.
+    ///
+    /// Required for any dependency whose OTP application name differs from
+    /// its package name. May be supplied multiple times.
+    #[arg(
+        verbatim_doc_comment,
+        long = "otp-app-override",
+        value_parser = parse_otp_app_override
+    )]
+    otp_app_overrides: Vec<(String, String)>,
```
lpil, line 800: Parse this to the final format we want so we don't have to transform it further before using it.

```suggestion
    otp_app_overrides: HashMap<EcoString, EcoString>,
```

compiler-cli/templates/gleam@@compile.erl
```diff
@@ -18,10 +18,6 @@ compile_package_loop() ->
             {ok, {Lib, Out, Modules}} = erl_parse:parse_term(Tokens),
             case compile_package(Lib, Out, Modules) of
                 {ok, ModuleNames} ->
```
lpil, line 20: This variable is unused and this code still collects the module names.

compiler-core/src/build/package_compiler/tests.rs
```diff
@@ -1,51 +1,66 @@
...
     existing_modules: Vec<(&str, &str)>,
-) -> Outcome<Compiled, Error> {
+) -> (InMemoryFileSystem, Outcome<Compiled, Error>) {
     let fs = InMemoryFileSystem::new();
-    fs.write(
-        Utf8Path::new(&format!("/src/{module}.gleam")),
-        "pub fn main() -> Nil { Nil }",
-    )
-    .expect("write module");
+    //
```
lpil, line 29: Please undo these changes. Tests should go in `test-commands` and/or `test-package-compiler`, whichever is a better fit. `test-commands` use the real file system and BEAM compilers, `test-package-compiler` does not.

We don't want to change the structure of these test to fit one new test case.

test-package-compiler/src/lib.rs
```diff
@@ -144,3 +144,74 @@ pub fn run_package_compiler(package: &'static str) -> String {
         Err(error) => error,
     }
 }
+
+pub fn prepare_with_otp_app_overrides(
```
lpil, line 148: Remove this please. The existing test functions should be used, we don't want a second duplicate system.

```rust
Compilation::for_package(...)
  .otp_app_override("erl_hpack", "hpack")
```

test-package-compiler/src/tests.rs
```diff
@@ -339,3 +339,20 @@ fn src_only() {
...
+fn otp_app_override() {
+    use ecow::EcoString;
+    use std::collections::HashMap;
+
+    let overrides = HashMap::from([
+        (
+            EcoString::from("gleam_otp"),
+            EcoString::from("my_gleam_otp_app"),
+        ),
+        (EcoString::from("midas"), EcoString::from("my_midas_app")),
```
lpil, line 353: tiny thing: it's not possible to change these in Gleam, so let's not use real Gleam names. Fake names or names of real Erlang packages that do this would be my preference.

lpil (commented):
Thank you!! 💜  Left some more notes!

Papipo:
@lpil everything addressed and new conflicts solved

compiler-cli/src/lib.rs
lpil: This hasn't been done yet @Papipo

test-package-compiler/src/tests.rs
```diff
@@ -339,3 +339,17 @@ fn src_only() {
...
 }
+
+#[test]
+fn otp_app_override() {
+    let mut compiler = TestHarness::new();
+    compiler
+        .compile(
+            Compilation::for_package("erlang_app_generation")
+                .otp_app_override("hpack_erl", "hpack")
+                .otp_app_override("uuid_erl", "uuid"),
```
lpil, line 350: These packages have the wrong names in the .app file in the snapshot

lpil (commented):
Nearly there! Just a couple tiny things inline

compiler-cli/src/lib.rs
Papipo: @lpil switched to comma-separated `otp-app-names`

test-package-compiler/src/tests.rs
Papipo: I have added a specific fixture for this test and left the original alone

lpil (approved):
Perfect! Phew that was really fiddly to get right. Pesky clap.

Thank you again!!

Papipo:
> Perfect! Phew that was really fiddly to get right. Pesky clap.
> 
> Thank you again!!

Wasn't precisely the `Option` what made the arg optional?

lpil:
It's one of the ways, but you can provide a value instead. I think there's another API too, but I find it so hard to understand the Clap docs.

lpil:
oh! I broke it! One minute! I will fix

PR #6176, opened 2026-08-14
Add sized-chunks to deny.toml

Papipo:
Discussed over [discord](https://discord.com/channels/768594524158427167/1537823434762096740/1537837427568283779)

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -29,6 +29,14 @@ FIX-DEADLINE: 2026-10-08
 """
 
 [[advisories.ignore]]
+id = "RUSTSEC-2026-0255"
+reason = """
+sized-chunks depends on im, which is unmaintained
+
+FIX-DEADLINE: 2026-12-11
+"""
+
+[[advisories.ignore]]
 id = "RUSTSEC-2026-0248"
 reason = """
 The im crate is unmaintained.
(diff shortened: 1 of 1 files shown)


ankddev (approved):

deny.toml
```diff
@@ -28,6 +28,14 @@ https://github.com/gleam-lang/gleam/issues/5829
 FIX-DEADLINE: 2026-10-08
 """
 
+[[advisories.ignore]]
+id = "RUSTSEC-2026-0255"
+reason = """
+sized-chunks depends on im, which is unmaintained
```
lpil: Sized-chunks does not depend on im, and this CVE is not for im being unmaintained.

The URL of the issue for removing sized-chunks should be here too!

deny.toml
```diff
@@ -73,16 +81,16 @@ FIX-DEADLINE: 2027-04-11
...
-    "BSD-3-Clause",
-    "BSL-1.0",
-    "CC0-1.0",
-    "CDLA-Permissive-2.0",
-    "ISC",
-    "MIT",
-    "MPL-2.0",
-    "Unicode-3.0",
-    "Zlib",
+  "Apache-2.0",
```
lpil: No unrelated changes please

deny.toml
```diff
@@ -28,6 +28,16 @@ https://github.com/gleam-lang/gleam/issues/5829
 FIX-DEADLINE: 2026-10-08
 """
 
+[[advisories.ignore]]
+id = "RUSTSEC-2026-0255"
+reason = """
+sized-chunks is unmaintained
```
ankddev: IIRC, it's not about unmaintained status, but about vulnerability/unsoundness, isn't it? And I think that we should also say that it is used by im.

Papipo: Some memory related stuff and the possibility of panicking I think

Papipo: But I think it's mentioned elsewhere that it's unmaintained as well, I'll recheck

lpil: If you look up the CVE you can see what it is for.

RUSTSEC-2026-0251 is for it being unmaintained, this is for a soundness bug.

lpil (approved):
Thank you!!

PR #6178, opened 2026-08-15
LS: surround value import completions after dot with braces

### LINKED ISSUE #6167: LS: surround completions of values with `{}` directly after module name
For example,

```gleam
import gleam/option.|
```

When I trigger completion (cursor is denoted with `|`), I see values and types in `gleam/option` module. If I select `Option`, which is type, and accept completion, I get following code:

```gleam
```

But if I select value (function, const, ..), like `Some`, I get following (invalid) code:

```gleam
import gleam/option.Some
```

I think that it should not produce invalid code, so completion should look like:

```gleam
import gleam/option.{Some}
```

I would like to work on this, if this is a wanted thing.

ankddev:
* Closes #6167

### DIFF
--- a/language-server/src/tests/snapshots/gleam_language_server__tests__completion__completions_for_an_unqualified_import_without_braces.snap
+++ b/language-server/src/tests/snapshots/gleam_language_server__tests__completion__completions_for_an_unqualified_import_without_braces.snap
@@ -1,0 +1,35 @@
+---
+source: language-server/src/tests/completion.rs
+expression: import dep.
+---
+import dep.|
+
+
+----- Completion content -----
+Wibble
+  kind:   Class
+  detail: Type
+  sort:   4_Wibble
+  edits:
+    [0:11-0:11]: "{type Wibble}"
+Wibble
+  kind:   EnumMember
+  detail: Wibble
+  sort:   4_Wibble
+  desc:   dep
+  edits:
+    [0:11-0:11]: "Wibble"
+wibble
+  kind:   Constant
+  detail: String
+  sort:   4_wibble
+  desc:   dep
+  edits:
+    [0:11-0:11]: "wibble"
+wobble
+  kind:   Function
+  detail: fn() -> a
+  sort:   4_wobble
+  desc:   dep
+  edits:
+    [0:11-0:11]: "wobble"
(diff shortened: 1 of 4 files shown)


language-server/src/tests/snapshots/gleam_language_server__tests__completion__completions_for_an_unqualified_import_without_braces.snap
```diff
@@ -0,0 +1,9 @@
+---
+source: language-server/src/tests/completion.rs
+expression: "\nimport dep."
+---
+
+import dep.|
+
+
+----- Completion content -----
```
ankddev, line 8: For some reason, it doesn't show completions here. It shows them in the real editor, and in tests if I add braces :) Any ideas?

ankddev (commented):
And what do you think about ignoring this Clippy lint in this function? Or what should we do here?

```
error: this function has too many arguments (8/7)
    --> language-server/src/completer.rs:1352:5
     |
1352 | /     fn value_completion(
1353 | |         &self,
1354 | |         module_qualifier: Option<&str>,
1355 | |         module_name: &str,
...    |
1360 | |         priority: CompletionKind,
1361 | |     ) -> CompletionItem {
     | |_______________________^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/rust-1.97.0/index.html#too_many_arguments
     = note: `-D clippy::too-many-arguments` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::too_many_arguments)]`
```

language-server/src/tests/snapshots/gleam_language_server__tests__completion__completions_for_an_unqualified_import_without_braces.snap
lpil: Not immediately. Do we have other completion tests with similar syntax errors? What do the type tests look like?

ankddev: Ah, thank you! I found similar test, my bad that I didn't check this earlier 😅

lpil:
Hmm. Let's add the attribute to ignore that lint on that function for now

ankddev:
@lpil Done!

lpil (approved):
Lovely!! Thank you once again!!

Issue #6182, opened 2026-08-15
Matching on <<"":utf8>> generates incorrect JavaScript


daniellionel01:
On `gleam v1.18.1` this code:

```gleam
pub fn main() {
   echo case <<"ab":utf8>> {
     <<"":utf8, "ab":utf8>> -> "a"
     _ -> "b"
   }
 }
```

Outputs this, when run on the JavaScript target:

```
/app $ gleam run -tjs
   Compiled in 0.01s
    Running app.main
file:///app/app.mjs:20
     &&
     ^^

SyntaxError: Unexpected token '&&'
    at compileSourceTextModule (node:internal/modules/esm/utils:346:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:107:18)
    at #translate (node:internal/modules/esm/loader:540:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:587:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:162:19)

Node.js v22.19.0
```

In the generated JavaScript, it leaves a dangling `&&` between the other conditions of the `if` statement:
```js
export function main() {
  let _block;
  let $ = toBitArray([stringBits("ab")]);
  if (
    $.bitSize >= 0 &&
     &&
    $.bitSize === 16 &&
    $.byteAt(0) === 97 && $.byteAt(1) === 98
  ) {
    _block = "a";
  } else {
    _block = "b";
  }
  return echo(_block, undefined, "src/app.gleam", 2);
}
```

lpil:
Oops, thank you!

daniellionel01:
I think this issue can be closed since louis PR got merged

lpil:
Thanks!

PR #6186, opened 2026-08-16
Fix empty string segment codegen bug

lpil:

https://github.com/gleam-lang/gleam/issues/6182


compiler-core/src/exhaustiveness.rs
```diff
@@ -3862,39 +3864,64 @@ fn segment_matched_value(
...
+        // It is possible to write a bit array segment that will
+        // match an empty string. This doesn't do anything, but it
+        // is valid Gleam.
+        //
+        // The generated JavaScript code for a bit array segment looks
+        // like this:
+        //
+        //     if (SEGMENT_SIZE_CHECK && SEGMENT_MATCHED_CHECK) {
+        //
+        // The matchec value check being what this function returns.
```
jtdowney: Not sure what this line is trying to say but certainly matchec is a typo

compiler-core/src/javascript/tests/bit_arrays.rs
```diff
@@ -2878,3 +2878,16 @@ pub const value = <<0:size(0), 1:size(8)>>
...
     );
 }
+
+// https://github.com/gleam-lang/gleam/issues/6182
+#[test]
+fn match_on_empty_bit_array_string() {
+    assert_js!(
+        "
+pub fn main(x) {
+  let assert <<\"\":utf8>> = x
```
jtdowney: You could use raw strings here `r#"..." and avoid the quote escaping.

jtdowney (commented):
A pattern with an alias still miscompiles because it can introduce a binding that it never emits.

```
----- SOURCE CODE

pub fn main(x) {
  case x {
    <<"" as a:utf8>> -> a
    _ -> "b"
  }
}


----- COMPILED JAVASCRIPT
export function main(x) {
  if (x.bitSize === 0) {
    return a;
  } else {
    return "b";
  }
}
```

lpil:
Thanks John!

giacomocavalieri (approved):

PR #5991, opened 2026-07-18
omit unreachable case clauses in Erlang codegen

jackprogramsjp:
- [x] The changes in this PR have been discussed beforehand in an issue
- [x] The issue for this PR has been linked
- [x] Tests have been added for new behaviour
- [x] The changelog has been updated for any user-facing changes

Resolves https://github.com/gleam-lang/gleam/issues/5963

PR #6195, opened 2026-08-19
revert PR 5991

### LINKED ISSUE #6179: Unreachable branches cause erlang codegen to panic
When you write an unreachable branch in Erlang that includes `|` and uses a variable, the compiler panics.

For example this code:
```gleam
pub fn main() {
  echo case [5, 3] {
    [_, ..] -> [0]
    [9, ..rest] | rest -> rest
    _ -> []
  }
}
```

Compiles and runs fine on `v1.18.1`, but panics on nightly:

```sh
error: Fatal compiler bug!

Panic: compiler-core/src/erlang.rs:512
	variable not in scope
```

This has not been introduced to any release yet, since its still sitting in the main branch without a RC.

I was able to track it down to this PR: https://github.com/gleam-lang/gleam/pull/5991
This commit specifically: https://github.com/gleam-lang/gleam/commit/f366e99a7a83711c09c960530b843fe4272168a2

giacomocavalieri:
This PR fixes #6179 

In short the problem was:
- In a clause body whenever we reference a variable we use the location of the first time the variable is defined (if we're dealing with alternative patterns)
- If a branch is unreachable we don't generate the code for it, so the variable with that location is never added to the scope, even if later variables from the other alternatives are they are not the ones being referenced
- Then when we generate the code for the body and try and reference the variable it looks like it's not in scope and the compiler crashes

We could try and fix this by also keeping track for each variable in an alternative where the first occurrence happens, so that when we generate code for those we use that default location.
But that would be a lot of work to not generate unreachable branches on the Erlang target, since that is not very high value (it's already a Gleam warning so we can assume folks will fix it, the warning is silenced already on the Erlang side, and the Erlang compiler will most likely prune it anyway) the easiest solution was to just revert the changes introduces by #5991.


lpil (approved):
Thank you!

PR #6196, opened 2026-08-19
LS: do not import prelude when renaming to original name

### LINKED ISSUE #6033: LS: do not import prelude items when renaming to original name
When triggering rename over one of prelude types/values and leaving name as-is, the LS adds new import with import alias, so:

```gleam
pub fn go() -> Nil {
//             ^^^ Rename to Nil here
  Nil
//^^^ Rename to Nil here
}
```

Becomes this:

```gleam
import gleam.{Nil as Nil, type Nil as Nil}

pub fn go() -> Nil {
  Nil
}
```

Here two things happen:

1. It adds new import, even though this items are prelude
2. It adds alias with same name. I think that this is because of absence of import before rename

Reproducible as of nightly

0xda157:
fixes https://github.com/gleam-lang/gleam/issues/6033

### DIFF
--- a/language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_value_to_original_name_remove_item_import.snap
+++ b/language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_value_to_original_name_remove_item_import.snap
@@ -1,0 +1,21 @@
+---
+source: language-server/src/tests/rename.rs
+expression: "import gleam.{Nil as Null, type Nil as Wibble}\n\npub fn go() -> Wibble {\n  Null\n}"
+---
+----- BEFORE RENAME
+-- app.gleam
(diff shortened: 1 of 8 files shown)


language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_value_to_original_name_remove_item_import.snap
```diff
@@ -0,0 +1,21 @@
...
+}
+
+
+----- AFTER RENAME
+-- app.gleam
+import gleam.{Nil, type Nil as Wibble}
+
+pub fn go() -> Wibble {
+  Nil
+}
```
0xda157, line 21: it would be better to remove the actual import of `Nil` here, but that would be kinda difficult.

lpil (approved):
Lovely!! Thank you once again! You always do fab work

PR #6199, opened 2026-08-20
Fix crash with pipeline in constants

### LINKED ISSUE #6192: Compiler crash on nightly when using a pipe `|>` in a `const`
This is not happening on `v1.18.1`, but on the latest commit in the main branch.

This gleam program, although not valid, panics the compiler:

```gleam
const a = 1 |> 2

pub fn main() {
  Nil
}
```

Gives this error message on both targets:

```bash
error: Fatal compiler bug!

Panic: compiler-core/src/parse.rs:5220
        Token could not be converted to binop.
```

On `v1.18.1` you get this output instead:

```bash
$ gleam run
  Compiling gleam_stdlib
  Compiling gleeunit
error: Syntax error
  ┌─ /app/src/app.gleam:1:13
  │
1 │ const a = 1 |> 2
  │             ^^ I was not expecting this

Found `|>`, expected one of:
- An import, const, type, or function.
```

giacomocavalieri:
This PR fixes #6192

- [ ] The changelog has been updated for any user-facing changes (there's no user facing changes since the crash only happens on nightly)

### DIFF
--- a/compiler-core/src/parse/snapshots/gleam_core__parse__tests__pipeline_in_constant_does_not_panic.snap
+++ b/compiler-core/src/parse/snapshots/gleam_core__parse__tests__pipeline_in_constant_does_not_panic.snap
@@ -1,0 +1,18 @@
+---
+source: compiler-core/src/parse/tests.rs
+expression: pub const wibble = 1 |> wobble
+---
+----- SOURCE CODE
+pub const wibble = 1 |> wobble
+
+----- ERROR
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:22
+  │
+1 │ pub const wibble = 1 |> wobble
+  │                      ^^ I was not expecting this
+
+Found `|>`, expected one of: 
+- A binary operator
+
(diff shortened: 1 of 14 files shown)


compiler-core/src/parse/snapshots/gleam_core__parse__tests__pipeline_in_constant_does_not_panic.snap
```diff
@@ -0,0 +1,18 @@
...
+error: Syntax error
+  ┌─ /src/parse/error.gleam:1:22
+  │
+1 │ pub const wibble = 1 |> wobble
+  │                      ^^ I was not expecting this
+
+Found `|>`, expected one of: 
+- A binary operator
+
+Hint: Pipelines are not allowed in constants
```
0xda157, line 19: This makes it seem like binary operators are supported in constants, which they aren't currently as far as I understand (aside from `<>`).

lpil (approved):
Thanks!!!

PR #6202, opened 2026-08-20
make echo smarter

### LINKED ISSUE #6040: echo is too eager to show bit arrays as strings on Erlang

giacomocavalieri:
This PR fixes #6040

- [ ] Tests have been added for new behaviour (existing tests already covered this!)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -158,6 +158,10 @@
 - Fixed a bug where file permissions were not set in Hex tarballs.
   ([Giacomo Cavalieri](https://github.com/giacomocavalieri))
 
+- Fixed a bug where echo would print BitArrays like `<<1, 2, 3>>` as strings on
+  the Erlang target.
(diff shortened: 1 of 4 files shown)


lpil (approved):
Lovely!! Thank you!

PR #6204, opened 2026-08-21
Better error message for usage of private items

### LINKED ISSUE #5890: Provide better error when trying to use private item
For example, when compiling this code:

```gleam
// testing.gleam
import wibble.{wobble}

// wibble.gleam
const wobble = 0

```

The build tool gives this error message:

```
error: Unknown module value
  ┌─ C:\Users\user\projects\test_gleam\src\test_gleam.gleam:1:16
  │
1 │ import wibble.{wobble}
  │                ^^^^^^

The module `wibble` does not have a `wobble` value.
```

I find it confusing, because it has no information, whether item doesn't exists or it's just private. For instance, Rust compiler gives this error message:

```
error[E0603]: constant `wobble` is private
 --> src\main.rs:2:13
  |
2 | use wibble::wobble;
  |             ^^^^^^ private constant
  |
note: the constant `wobble` is defined here
 --> src\wibble.rs:1:1
  |
1 | const wobble: i32 = 0;
  | ^^^^^^^^^^^^^^^^^^^^^^
```

I'm not sure if it should say about visibility when importing/using an item from different package, but I think that it shouldn't.

ankddev:
- Closes #5890 
***

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -123,6 +123,32 @@
 - The compiler now provides hint about `|' in pattern matching
   ([n0kk23](https://github.com/n0kk23))
 
+- The compiler now provides better error message when trying to import or use
+  private value or type from modules within current package. For example:
+
+  ```gleam
+  import wibble
+  //     ^^^^^^ `wibble` is a module in same package
+
+  fn go() -> wibble.Wobble {
+    todo
+  }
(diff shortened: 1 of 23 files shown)


lpil (approved):
Fab!! Thank you!!!

PR #6205, opened 2026-08-21
Improve error message when adding a redundant dependency

giacomocavalieri:
I noticed that the error message when adding a package that is already a dependency of the project as a dev dependency was not great. With this PR the build tool now produces a nicer error message explaining what's wrong!

### DIFF
--- a/compiler-core/src/error.rs
+++ b/compiler-core/src/error.rs
@@ -441,6 +441,18 @@ file_names.iter().map(|x| x.as_str()).join(", "))]
 
     #[error("could not create temp file: {error}")]
     CouldNotCreateTempFile { error: String },
+
+    /// This happens when we try adding some packages as regular dependencies
+    /// (gleam add wibble wobble ...) but those are already listed in the dev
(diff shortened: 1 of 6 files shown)


lpil (commented):
Snapshot tests for the new errors always! I'll add them 👍

lpil (approved):
Thanks!!!

compiler-core/src/error.rs
```diff
@@ -2564,6 +2576,88 @@ add `gleam add {name}` in this project."
...
+                            "If you want to use this package as a regular \
+dependency, you can move it to the `dependencies` section of this project's \
+`gleam.toml`."
+                                .into(),
+                        ),
+                    }],
+                    packages => vec![Diagnostic {
+                        title: "Packages are already dev dependencies".into(),
+                        text: wrap_format!(
+                            "{} are already develpment dependencies of this project.",
```
ankddev: Typo
```suggestion
                            "{} are already development dependencies of this project.",
```

PR #6208, opened 2026-08-22
Suggest aliases for already imported modules

### LINKED ISSUE #5908: Import module suggestion suggests reimporting when module is aliased
Gleam version: `gleam 1.17.0`

```gleam
import gleam/io as gleam_io

pub fn main() -> Nil {
  io.println("Hello Joe!")
}
```

produces the error

```
error: Unknown module
  ┌─ src.gleam:4:3
  │
4 │   io.println("Hello Joe!")
  │   ^^

No module has been found with the name `io`.

Hint: Did you mean to import `gleam/io`?
```

Instead of suggesting to re-import the module we should suggest using the alias, in addition to other possible import suggestions. The `Import _` code action should also be updated or a new code action should be added for this.

ssenthilnathan3:

Closes #5908.

I updated the language server code action to replace the unknown module name with the existing alias.

Added regression tests to cover the compiler diagnostic and language server code action.

### DIFF
--- a/compiler-core/src/type_/environment.rs
+++ b/compiler-core/src/type_/environment.rs
@@ -962,13 +962,9 @@ impl Environment<'_> {
                 }
 
                 match &imported {
-                    // Don't suggest importing modules if they are already imported
-                    _ if self
-                        .imported_modules
-                        .contains_key(importable.split('/').next_back().unwrap_or(importable)) =>
-                    {
-                        None
-                    }
+                    // This lookup uses the full module path, rather than its local
+                    // name, so it also detects modules imported with an alias.
+                    _ if self.names.is_module_imported(importable) => None,
                     Imported::Type(name) if module_info.get_public_type(name).is_some() => {
                         Some(ModuleSuggestion::Importable(importable.clone()))
                     }
--- a/compiler-core/src/type_/printer.rs
+++ b/compiler-core/src/type_/printer.rs
@@ -222,6 +222,10 @@ impl Names {
             .map(|(_, location)| location)
     }
 
+    pub fn is_module_imported(&self, module_name: &str) -> bool {
+        self.imported_modules.contains_key(module_name)
+    }
+
     /// Check whether a particular type alias is reexporting an internal type,
     /// and if so register it so we can print it correctly.
     pub fn maybe_register_reexport_alias(
--- a/compiler-core/src/type_/tests/errors.rs
+++ b/compiler-core/src/type_/tests/errors.rs
@@ -2287,6 +2287,19 @@ pub fn main() {
 }
 
 #[test]
+fn unknown_module_does_not_suggest_importing_module_imported_with_alias() {
+    assert_module_error!(
+        ("gleam/io", "pub fn println(message: String) {}"),
+        "
+import gleam/io as gleam_io
+pub fn main() {
+  io.println(\"Hello, world!\")
+}
+",
+    );
+}
+
+#[test]
 fn unknown_module_suggest_typo_for_unimported_module() {
     assert_module_error!(
         ("wibble/wobble", "pub fn wubble() {}"),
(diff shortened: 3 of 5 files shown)


compiler-core/src/type_/environment.rs
```diff
@@ -978,24 +981,30 @@ impl Environment<'_> {
...
-
-        let threshold = std::cmp::max(module.chars().count() / 3, 1);
+            .filter_map(|suggestion| {
+                edit_distance(module, suggestion.last_name_component(), threshold)
+                    .map(|distance| (suggestion, distance))
+            });
+
+        let suggestions = suggestions.chain(self.imported_modules.iter().filter_map(
+            |(alias, (_, imported))| {
+                let imported_name = imported
```
lpil: This is not the "imported name" (the alias). This is the default alias for the module, but that might not be the alias that the programmer used.

compiler-core/src/type_/environment.rs
```diff
@@ -965,7 +967,8 @@ impl Environment<'_> {
                     // Don't suggest importing modules if they are already imported
                     _ if self
                         .imported_modules
-                        .contains_key(importable.split('/').next_back().unwrap_or(importable)) =>
+                        .values()
+                        .any(|(_, imported)| imported.name == *importable) =>
```
lpil: How come this has changed? I don't understand the purpose of this change, but it does mean that for each importable module all the already imported modules are iterated over, while previous it was only a hash lookup.

lpil (commented):
Hello! Thank you for this!

Sorry, this may be a me problem but I'm finding the changes a bit hard to understand. Could you talk me through them and add comments to explain the logic please? Thanks!

ssenthilnathan3:
> Hello! Thank you for this!
> 
> Sorry, this may be a me problem but I'm finding the changes a bit hard to understand. Could you talk me through them and add comments to explain the logic please? Thanks!

Sorry, I missed things here! I changed this because the existing lookup only recognizes the module when it’s imported under the default alias. With an explicit alias, we could end up treating the same module as importable again.

My first fix used the `ModuleInterface` in `imported_modules` since it already has the full module path. That fixed the behavior, but I missed that it meant scanning all imported modules for every candidate. Also, `imported_name` wasn’t a great name there since it was really the default alias.

I'm changing it to use the existing map keyed by the full module path instead, so we get the same behavior without adding that scan.

And I added a comment for the slightly non-obvious part: matching needs to consider both names, but the suggestion should use the local name.

compiler-core/src/type_/environment.rs
ssenthilnathan3: The change is needed because the old lookup misses explicitly aliased imports. 

You’re right about the scan though, `Environment::names` already maps full module paths, so i’ll use that instead and keep the lookup constant-time.

compiler-core/src/type_/environment.rs
ssenthilnathan3: You’re right, imported_name is actually the module’s default alias, while the map key is the local name used by the programmer. I’ll rename it to make that distinction clear. 

I'm also documenting why matching checks both but suggestions use the local name.

compiler-core/src/type_/environment.rs
```diff
@@ -953,7 +953,11 @@ impl Environment<'_> {
 
     /// Suggest modules to import or use, for an unknown module
     pub fn suggest_modules(&self, module: &str, imported: Imported) -> Vec<ModuleSuggestion> {
-        let mut suggestions = self
+        let threshold = std::cmp::max(module.chars().count() / 3, 1);
+
+        // Find modules that are not currently imported and whose default alias
+        // is similar to the unknown module name.
```
lpil: This isn't quite right. This ticket is to stop the compiler from suggesting that the programmer import a module that is already imported, not to change the suggestion system to ignore the programmer's aliases in favour of the default aliases.

lpil (commented):
Thank you, I understand much better now. I've left a comment inline about this approach 🙏

ssenthilnathan3:
@lpil I addressed the changes. Should be ready for another review now! Thanks!

PR #6209, opened 2026-08-22
Fix string prefix fall through

### LINKED ISSUE #6168: Gleam Javascript miscompiles some string pattern matches.
gleam 1.18.1

The pattern match below should take the `400` branch, but in gleam javascript it takes the `999` branch.

Also note this bug is not caused exclusively by guards (see second repro), it's about string matches "shadowing" prefix matches when they shouldn't.

```gleam
pub fn w1(s: String, x: Int) -> Int {
  case s, x {
    "ab", y if y > 4 -> 1
    "a" <> _, _ -> 400
    _, _ -> 999
  }
}

pub fn main() {
  echo w1("ab", 1)
}
```
Compiles to:
```js
export function w1(s, x) {
  if (s === "ab") {
    let y = x;
    if (y > 4) {
      return 1;
    } else {
      return 999;
    }
  } else if (s.charCodeAt(0) === 97) {
    return 400;
  } else {
    return 999;
  }
}
```

Second one: should be 502 but prints 999
```gleam
import gleam/string

pub type Shape {
  Circle(r: Int)
  Square(s: Int)
  Rect(w: Int, h: Int)
}

pub type Wrap {
  Wrapped(v: Shape)
  Empty
}

pub fn a(w: Wrap, s: String) -> Int {
  case w, s {
    _, "ab" -> 100
    Wrapped(Square(0)), "ab" <> _ -> 200
    Wrapped(_), "a" <> r -> 500 + string.length(r)
    _, _ -> 999
  }
}

pub fn main() {
  echo a(Wrapped(Rect(2, 1)), "abc")
}
```

jtdowney:
Closes #6168 

There were a few distinct but related bugs in #6168. Because those fixes caught more redundant patterns, the existing warning was expanded.


compiler-core/src/exhaustiveness.rs
```diff
@@ -1025,6 +1049,27 @@ pub enum BitArrayTest {
 }
 
 impl BitArrayTest {
+    /// Whether `move_unconditional_patterns` has something to do with this test
+    /// when it is the first one of a pattern, either popping it into the
+    /// branch's body or rewriting it into a test on the value it aliases.
+    ///
+    fn is_poppable(&self) -> bool {
```
lpil: Can we give this a more descriptive name? It's not clear what property of the data makes it "poppable", I think there's a bunch of context to this name that is elsewhere.

lpil (commented):
Thank you!! This PR looks really good to me, but with my lesser understanding of this area of the codebase I don't feel confident merging without the expert review of @GearsDatapacks or @giacomocavalieri. 🙏

compiler-core/src/exhaustiveness.rs
jtdowney: Done in 4d185a52585f5816653be367d5ddafbaa70efedd

lpil:
@GearsDatapacks @giacomocavalieri What did you think of this? 🙏

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__strings__overlapping_string_prefixes_with_guard.snap
```diff
@@ -16,8 +16,14 @@ pub fn classify(input: String, flag: Bool) -> String {
...
 ----- COMPILED JAVASCRIPT
 export function classify(input, flag) {
-  if (input.startsWith("aa") && flag) {
-    return "wibble";
+  if (input.startsWith("aa")) {
+    if (flag) {
+      return "wibble";
+    } else if (flag) {
+      return "wobble";
+    } else {
```
giacomocavalieri, line 24: It looks like this should never happen

compiler-core/src/exhaustiveness.rs
```diff
@@ -257,11 +249,20 @@ impl Branch {
...
+                        let prefix = prefix.clone();
+                        let rest = *rest;
+                        // Patterns live in an arena shared by every branch, and
+                        // splitting can leave several branches pointing at this
+                        // one. Taking the name out would bind it only in the
+                        // branch that happens to be compiled first, so give this
+                        // branch's check a copy of the pattern without the name.
+                        check.pattern = compiler.string_prefix_pattern(&prefix, rest);
+                        self.body.assign_literal_string(variable, prefix);
                         return true;
```
giacomocavalieri, line 264: I'm not sure I understand this, could you try and explain it to me again?

compiler-core/src/javascript/tests/snapshots/gleam_core__javascript__tests__strings__overlapping_string_prefixes_with_guard.snap
jtdowney: I believe this is an existing missing optimization, not a regression in this branch. It just shows up in this file because it was hidden by the bug that this branch fixes.

https://playground.gleam.run/#N4IgbgpgTgzglgewHYgFwEYA0IDGyAuES+aIADgK4BGABAGZI0DmCAFAB6o0CSxm9AGwCGTLgCEECAQEoaAWgB8NAMr4ocJExrAAOoxo4hMCDXba9NSzXQ04dQSPlKdIIS4tWA+rft1hWxRoXKnd9SxtAlwATUKsab0jcWJoAXz00lBSgA==

compiler-core/src/exhaustiveness.rs
jtdowney: The `std::mem::take(prefix_name)` that this replaces was the crux of the issue. It mutated the pattern in place, and patterns are shared through an arena. Splitting clones a branch into every path where it is still relevant, so several branches can point to the same pattern. Taking the name out bound it in whichever branch was compiled first and left the others referring to a name that was never bound.

jtdowney:
I rebased on main, which had some conflicts with #6186

compiler-core/src/exhaustiveness.rs
giacomocavalieri: Aha I see it does make sense, is the code written like that to make the borrow checker happy? I would have probably written it like this otherwise. I would also keep the comment shorter, it confuses me a little that the comment talks about a past wrong implementation detail
```suggestion
                        let Some(variable) = prefix_name.clone() else {
                            return true;
                        };
                        // We assign the literal value to the branch's body,
                        // then we can go on and replace the pattern with a
                        // simpler one, where there's no assignment.
                        self.body.assign_literal_string(variable, prefix.clone());
                        check.pattern = compiler.string_prefix_pattern(&prefix, *rest);
                        
return true;
```

compiler-core/src/exhaustiveness.rs
```diff
@@ -518,10 +544,11 @@ pub enum Pattern {
         prefix_name: Option<EcoString>,
         rest: Id<Pattern>,
     },
+    /// Binds the rest of a string whose prefix an earlier check has already
+    /// matched, so there is nothing left to test.
```
giacomocavalieri, line 546: Could you add an example to this highlighting what pattern this corresponds to?

jtdowney: Added in 239c4d39c22089cab7a34ee5311d44bba72b2257

compiler-core/src/exhaustiveness.rs
jtdowney: Yes, the locals are there for the borrow checker. I tried your suggestion, and it fails because `prefix` and `rest` still borrow from `compiler` when `string_prefix_pattern` needs another mutable borrow.

Good point on the comment, that is updated in 84f3951989c1c0de4cb3896449fbef96c8e24397.

giacomocavalieri (approved):

jtdowney:
@lpil I believe this is ready to merge. I rebased to resolve the changelog conflicts, but everything else is the same as what Jak approved.

lpil:
Thank you both!!!!! 💜

lpil (approved):
pew pew

PR #6210, opened 2026-08-23
Fix optional dep resolution

### LINKED ISSUE #5829: Migrate from the unmaintained proc-macro-error2 crate
https://rustsec.org/advisories/RUSTSEC-2024-0370

At time of writing, it is used by a single transitive dependency.

```
   $ cargo tree -i proc-macro-error2
   proc-macro-error2 v2.0.1
   └── i18n-embed-fl v0.9.2
       └── age v0.11.3
           └── gleam-core
```

jtdowney:
Closes #5829

While investigating #5829, I found the real reason the resolution tests destabilized. Upgrading `age` to 0.12 forces `rustc-hash` from 2.0.0 to 2.1.x, and `pubgrub`'s `Map` type is an `FxHashMap`. FxHasher iteration order is deterministic for a given `rustc-hash` version, but it is arbitrary, and 2.1 changed the hash function. The order pubgrub visits packages in changed, and some resolution tests started failing.

That only exposed the bug. The old code decided whether to apply an optional requirement by checking `!packages.contains_key(name)`, where `packages` is the lazily populated fetch cache, used as a proxy for "is this package in the solve":

https://github.com/gleam-lang/gleam/blob/68326fc52ba7ea44a6e5affe1a4047dd8bc84396/compiler-core/src/dependency.rs#L389-L410

Since `packages` caches what has been requested, if pubgrub tries a candidate version whose deps make it fetch `wibble` and then backtracks away from that branch entirely, `wibble` remains in the cache. Any optional requirement on `wibble` examined afterward sees `contains_key == true` and becomes a hard requirement, forcing `wibble` back into the tree that had just discarded it

This change remodels pubgrub's version type as either present or absent, with `Absent` as a sentinel that sorts below every real version. An optional requirement contributes `{Absent} U range` as its constraint, so it is satisfied either by the package staying out of the solve or by a version in the range. A mandatory requirement excludes `Absent`, so when pubgrub intersects a package's constraints, one mandatory requirement rules out absence, and the optional ranges then apply as ordinary version constraints.

### DIFF
--- a/compiler-core/src/dependency.rs
+++ b/compiler-core/src/dependency.rs
@@ -1,7 +1,14 @@
 // SPDX-License-Identifier: Apache-2.0
 // SPDX-FileCopyrightText: 2023 The Gleam contributors
 
-use std::{cell::RefCell, cmp::Reverse, collections::HashMap, rc::Rc};
+use std::{
+    cell::RefCell,
+    cmp::{Ordering, Reverse},
+    collections::HashMap,
+    fmt,
+    ops::Bound,
+    rc::Rc,
+};
 
 use crate::{Error, Result, manifest};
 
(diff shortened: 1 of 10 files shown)


compiler-core/src/dependency.rs
```diff
@@ -10,12 +10,60 @@ use hexpm::{
...
+}
+
+impl fmt::Display for Presence {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Presence::Absent => write!(f, "absent"),
+            Presence::Present(version) => version.fmt(f),
+        }
+    }
+}
```
lpil, line 70: We try to avoid implementing these traits as it hides where the logic is implemented, and it's very rare that there's a single text format for some data. For example, `version.fmt` returns a valid version in the Hex syntax, but this wrapper does not.

Where is this used?

edit: ah this is forced by the pubgrub crate's VersionSet I believe? I suppose it's not avoidable in that case.

compiler-core/src/dependency.rs
```diff
@@ -10,12 +10,60 @@ use hexpm::{
...
-use pubgrub::{Dependencies, Map};
+use pubgrub::{Dependencies, DerivationTree, Map};
 use thiserror::Error;
 
 pub type PackageVersions = HashMap<String, Version>;
 
-type PubgrubRange = pubgrub::Range<Version>;
+type PubgrubRange = pubgrub::Range<Presence>;
+
+/// The version a package resolves to, or `Absent` when the package is not part
```
lpil:
```suggestion
/// The version a package resolves to `Absent` when the package is not part
```

Is that right?

compiler-core/src/dependency.rs
```diff
@@ -10,12 +10,60 @@ use hexpm::{
...
+/// of the dependency tree at all.
+///
+/// Hex's optional requirements do not pull a package in, they only constrain
+/// one that something else has already pulled in. pubgrub has no way to say
+/// "this package, if it is here at all", so absence is modelled as a version
+/// that only optional requirements admit. A package resolves to `Absent`
+/// exactly when every requirement on it was optional.
+#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
```
lpil: Maybe this is me being a bit dim, but it took me quite a while to spot that this type implements `Ord` and how important that is to the resolution. Could we make this clearer with some documentation? Or perhaps an explicit implementation? 🙏

lpil (commented):
Wow this is fantastic work! I've looked into this multiple times now and I didn't get anywhere. Thank you so much!!

I've left some Qs inline, nothing major.

compiler-core/src/dependency.rs
jtdowney: It's required by pubgrub, as far as I'm aware, it's not actually used.

compiler-core/src/dependency.rs
jtdowney: Tried to make it clearer in 5f667436bdecc0009ebf62c409c0412627fc38fa

compiler-core/src/dependency.rs
jtdowney: Definitely not dim, it is the crux of the change. I had a comment about it, but it felt out of place, so I removed it. In 5f667436bdecc0009ebf62c409c0412627fc38fa I moved to an explicit implementation, which is better anyway since it is so important to the way it works.

jtdowney:
> Wow this is fantastic work! I've looked into this multiple times now and I didn't get anywhere. Thank you so much!!

The idea came from reading pubgrub-rs/pubgrub#120

lpil (approved):
Ah very good. Thank you!!

PR #6216, opened 2026-08-25
Turn remote constants into function calls

giacomocavalieri:
This is a change I've been discussing with Louis for a while.
This boils down to making sure that constants are never inlined across Erlang modules, making it so that the value of a constant is no longer part of the public API of a module. This way, changing a value of a constant, means we no longer have to recompile each module that depended on it to inline the new value


compiler-core/src/erlang/tests/snapshots/gleam_core__erlang__tests__consts__pub_const_equal_to_private_function.snap
```diff
@@ -14,7 +14,11 @@ expression: "\n          fn identity(a) {\n            a\n          }\n\n
 ----- COMPILED ERLANG
 -module(my@mod).
 -compile([no_auto_import, nowarn_ignored, nowarn_unused_vars, nowarn_unused_function, nowarn_nomatch, inline]).
--export([identity/1]).
+-export([id/0]).
+
+-spec id() -> fun((K) -> K).
+id() ->
+    fun identity/1.
```
lpil, line 21: It's cool how we not longer export private stuff with this change

lpil (approved):
Amazing, thank you!!!

PR #6217, opened 2026-08-25
ls: add code action to inline constant usage

### LINKED ISSUE #5891: LS: code action to inline constant
I would like to see an `Inline constant` code action. It would extract value from a constant, like with `Extract variable`. There is a [similar r-a code action](https://rust-analyzer.github.io/book/assists.html#inline_const_as_literal), though there it also needs to evaluate the constant.

### Example

```gleam
const answer = 0

fn wibble() {
  let wubble = wobble(answer)
  //                  ^^^^^^ Triggering here
}
```

Triggering the code action would produce this code:

```gleam
const answer = 0

fn wibble() {
  let wubble = wobble(0)
}
```

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/5891
***


ankddev (commented):
I didn't figure out how to implement inlining of unqualified constants in clause guards, so any help is appreciated!

language-server/src/code_action.rs
```diff
@@ -13329,3 +13332,230 @@ impl<'ast> ast::visit::Visit<'ast> for ConvertIntToDifferentBase<'ast> {
         self.int = Some((*location, base, int_value.clone()))
     }
 }
+
+pub struct InlineConstantUsage<'a> {
```
lpil: Can we call it "Inline constant value" please 🙏

lpil (commented):
Thank you! The code looks nice, but could you restrict it to only being able to inline constants from the same module, and then using the same approach as inline variable does. 🙏

ankddev:
Hello @lpil, thank you for review! I addressed all your comments. I also tried to make it work for unqualified usages in clause guards in this wip commit: https://github.com/gleam-lang/gleam/pull/6217/commits/81822818d559df1c2db63187d3e7ec68c06646a0, would like to hear if this is right direction (I also left few comments inline about it).

compiler-core/src/parse.rs
```diff
@@ -2007,7 +2007,11 @@ where
                 match self.parse_const_value_unit()? {
                     Some(const_val) => {
                         // Constant
-                        Ok(Some(ClauseGuard::Constant(const_val)))
+                        Ok(Some(ClauseGuard::Constant {
+                            location: const_val.location(),
+                            literal: const_val,
+                            module: "".into(),
```
ankddev, line 2013: I'm not sure where to get module here from.

PR #6218, opened 2026-08-25
Fix JavaScript names escaping a case clause

### LINKED ISSUE #6212: JavaScript code generator calls the wrong variable when a case pattern has the same name as a function
The gleam fuzzer strikes again! This program:

```gleam
fn wibble(_: Int) -> Float {
  0.0
}

pub fn main() {
  case 0 + 0 {
    wibble -> {
      echo wibble
    }
  }
  |> wibble
}

```

Crashes with this error message on the JavaScript target, but runs fine on Erlang:

```bash
$ gleam run -tjs
  Compiling app
   Compiled in 0.02s
    Running app.main
src/app.gleam:8
0
file:///app/build/dev/javascript/app/app.mjs:23
  return wibble$1(_pipe);
         ^

TypeError: wibble$1 is not a function
    at main (file:///app/build/dev/javascript/app/app.mjs:23:10)
    at file:///app/build/dev/javascript/app/gleam@@private_main_v1.18.1.mjs:2:1
    at ModuleJob.run (node:internal/modules/esm/module_job:345:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:651:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)

Node.js v22.19.0
```

This happens on gleam `v1.18.1` and nightly.

### LINKED ISSUE #6213: JavaScript code generator uses the wrong variable when a let inside a case arm shadows an outer variable
This Gleam program:

```gleam
pub fn main() {
  let m = 1
  let _ = case 0 + 0 {
    _ -> {
      let m = 99
      m
    }
  }
  echo m
}
```

Generates incorrect JavaScript. It should output 1, which it does on the Erlang target, but `m` is not shadowed correctly on the JavaScript target:
```bash
$ gleam run -terl
src/app.gleam:9
1

$ gleam run -tjs
src/app.gleam:9
99
```

This happens on gleam v1.18.1 and nightly.

jtdowney:
Closes #6212
Closes #6213

This is related to #5743 in that it is from when decision trees can be embedded into the parent scope without any if/else branches which create new scope for JS. I took the code largely introduced in #5746 and applied it for this case too.

Since this had popped up a few times I also went and looked for other cases and found one relating to guards. Bindings for guards are declared in outer scope.

For example, this is a problem because `b` gets declared twice:

```
----- SOURCE CODE

pub fn go(x) {
  case x {
    #(_, b) if b == 2 -> b
    #(_, b) -> b
  }
  let b = 5
  b
}

----- COMPILED JAVASCRIPT
export function go(x) {
  let b = x[1];
  if (b === 2) {
    b;
  } else {
    let b = x[1];
    b;
  }
  let b = 5;
  return b;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -212,6 +212,12 @@
   runtime crash when the shadowed name was a module function or constant.
   ([John Downey](https://github.com/jtdowney))
 
+- Fixed a bug where on the JavaScript target the variables a clause guard
+  needed would be declared in the enclosing scope of a case that always
+  matched, so a later `let` binding one of those names produced a duplicate
+  declaration and the whole module failed to parse.
+  ([John Downey](https://github.com/jtdowney))
+
 ## v1.18.1 - 2026-08-01
 
 - Fixed a bug where the Erlang code generator would generate wrong code when
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -206,6 +206,12 @@
   prefix alias.
   ([Andrey Kozhev](https://github.com/ankddev))
 
+- Fixed a bug where on the JavaScript target a name shadowed inside a case
+  clause that always matched would be used in place of the outer binding by the
+  code that followed the case expression. This could produce a wrong value, or a
+  runtime crash when the shadowed name was a module function or constant.
+  ([John Downey](https://github.com/jtdowney))
+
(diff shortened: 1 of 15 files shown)


lpil (approved):
Very nice! Super clean. Thank you again!!!

PR #6219, opened 2026-08-25
ls: remove import of prelude item when renaming to original name

### LINKED ISSUE #6201: LS: remove prelude import when prelude item is renamed to original name
Currently renaming `Null` to `Nil` here

```gleam
import gleam.{Nil as Null, type Nil as Wibble}

pub fn go() -> Wibble {
  Null
}
```

results in

```gleam
import gleam.{Nil, type Nil as Wibble}

pub fn go() -> Wibble {
  Nil
}
```

instead we could remove the unneeded import:

```gleam
import gleam.{type Nil as Wibble}

pub fn go() -> Wibble {
  Nil
}
```

see https://github.com/gleam-lang/gleam/blob/main/language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_value_to_original_name_remove_item_import.snap
somewhat related to https://github.com/gleam-lang/gleam/issues/6033

ankddev:
* Closes https://github.com/gleam-lang/gleam/issues/6201
***

### DIFF
--- a/language-server/src/rename.rs
+++ b/language-server/src/rename.rs
@@ -4,6 +4,7 @@
 use std::collections::HashMap;
 
 use ecow::EcoString;
+use itertools::Itertools;
 use lsp_server::ResponseError;
 use lsp_types::{Range, RenameParams, TextEdit, Uri as Url, WorkspaceEdit};
 
@@ -375,10 +376,8 @@ fn alias_references_in_module(
                 if name == &params.new_name {
                     if module_name == "gleam" {
                         // If new name is same as old name and item is from prelude,
-                        // we can delete entire import.
-                        rename_same_named_prelude_import_reference(
-                            module, layer, &mut edits, location,
-                        );
+                        // we can delete the entire import.
+                        rename_same_named_prelude_import_reference(&mut edits, module, location);
                     } else {
                         // If new name is same as old name, we can delete
                         // the `as ..` part.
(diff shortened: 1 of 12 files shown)


language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_type_to_original_name_remove_item_import.snap
```diff
@@ -0,0 +1,21 @@
...
+
+pub fn go() -> Wibble {
+               ↑▔▔▔▔▔  
+  Null
+}
+
+
+----- AFTER RENAME
+-- app.gleam
+import gleam.{Nil as Null, }
```
ankddev: I'm not sure if it's worth fixing this `, `, formatter should remove it. For a fix, we will need to track if it's the last import, and remove comma before it too.

language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_value_to_original_name_remove_full_import.snap
```diff
@@ -14,7 +14,7 @@ pub fn go() -> Nil {
 
 ----- AFTER RENAME
 -- app.gleam
-import gleam.{Nil}
+
```
ankddev, line 17: Is it good that there's newline here after removing import? Again, it will be removed by the formatter.

ankddev: Nevermind, I'll fix this!

language-server/src/tests/snapshots/gleam_language_server__tests__rename__rename_prelude_type_to_original_name_remove_item_import.snap
ankddev: Well, I'll fix this one too!

language-server/src/rename.rs
```diff
@@ -419,6 +430,100 @@ fn alias_references_in_module(
...
+        .collect_vec();
+
+    let renamed_index = imported_locations.binary_search(&location).ok();
+    let next_value_location =
+        renamed_index.and_then(|value_index| imported_locations.get(value_index + 1));
+    let previous_value_location = renamed_index.and_then(|value_index| {
+        value_index
+            .checked_sub(1)
+            .and_then(|previous_index| imported_locations.get(previous_index))
+    });
```
lpil, line 467: Instead of sorting, allocating a vector, binary searching and then indexing twice, let's iterate over the items once and use 2 mutable variables to hold the matching one and the one before it.

language-server/src/rename.rs
```diff
@@ -419,6 +430,100 @@ fn alias_references_in_module(
...
+
+    match (previous_value_location, next_value_location) {
+        // If there's an import following the renamed import, we need to delete everything before its start.
```
lpil, line 470: Wrap long lines please 🙏

lpil (commented):
Nice!! Thank you!! I've left a couple comments

PR #6222, opened 2026-08-26
Refactor `analysis` function

giacomocavalieri:
While trying to implement better cache fingerprinting I ran into a problem trying to make it work with the `analyse` function due to the way it was structured.

I think refactoring it into a structure makes the code a lot easier to follow, and will allow me to implement the public api fingerprinting much more easily!


compiler-core/src/build/package_compiler.rs
```diff
@@ -581,232 +579,380 @@ pub enum StdlibPackage {
...
+    name: EcoString,
+    origin: Origin,
+    path: Utf8PathBuf,
+    code: EcoString,
+    mtime: SystemTime,
+    dependencies: Vec<(EcoString, SrcSpan)>,
+}
+
+impl ModuleInfo {
+    fn into_typed_module(self, ast: TypedModule, extra: ModuleExtra) -> Module {
```
lpil: This is called "into_typed_module", but it takes a `TypedModule` as an argument instead of returning one.

compiler-core/src/build/package_compiler.rs
```diff
@@ -581,232 +579,380 @@ pub enum StdlibPackage {
...
+    skipped_modules: HashMap<EcoString, SkippedModule>,
+    /// Keeps track of the modules whose compilation failed because of some
+    /// error.
+    failed_modules: HashMap<EcoString, FailedModule>,
+    /// Keeps track of the modules that could be analysed successfully.
+    analysed_modules: Vec<Module>,
+}
+
+/// Struct to carry around cheap to clone data about a module, separately from
+/// its ast and `ModuleExtra` which would be expensive to clone.
```
lpil: This doc comment seems a bit misleading, it seems to suggest the purpose of this data structure is for it to be cloned, and for that to be cheap. `dependencies` and `path` are not cheap to clone, they are vectors of arbitrary length. The others are copyable or reference counted, which is cheap, but cloning them together is wasteful.

Looking at how this is used there's no point at which the whole thing is cloned, instead it is passed by reference and then individual fields are cloned, so I'm not sure the purpose of this struct.

I think this also makes worse the existing problem of module-related data structures having unclear names. They all contain information about a module, and there's multiple called `Module`, and now we have one called `ModuleInfo`, which is also vague name that could be applied to any of the others.

Perhaps this could be `ModuleSourceData` or something like that, and it be part of `UncompiledModule`, as that's when this data first all comes into scope.

compiler-core/src/build/package_compiler.rs
```diff
@@ -581,232 +579,380 @@ pub enum StdlibPackage {
...
+        );
     }
 }
 
 #[derive(Debug)]
 pub(crate) enum Input {
-    New(Box<UncompiledModule>),
-    Cached(CachedModule),
+    New { module: Box<UncompiledModule> },
+    Cached { module: CachedModule },
```
lpil, line 895: Is this change a stepping-stone towards additional fields being added in a later commit?

giacomocavalieri: Oops yeah, hope it's not a problem

lpil: yeah if there's more coming that's cool, I was just confused with only this

compiler-core/src/build/package_compiler.rs
```diff
@@ -581,232 +575,353 @@ pub enum StdlibPackage {
...
+                        input_path: path,
+                        mtime,
+                        origin,
+                        dependencies,
+                        extra,
+                        ast,
+                    };
+                    module.attach_doc_and_module_comments();
+
+                    let _ = self.incomplete_modules.remove(&module.name.clone());
```
lpil: `name` is being cloned here but then only used by reference, can remove the clone.

lpil (approved):
Nice! Thank you!!

PR #6225, opened 2026-08-27
Sort JS variant constant imports

jtdowney:
This was discussed on Discord a few weeks ago, but I forgot to make a PR until now. In a few of the investigations around decision trees, I have been trying to validate that things end up with the same SHA256 hash at the end of compilation, and I noticed that on some projects, files were inconsistent not because the decision tree created different code, but because the import paths were randomized by the hash map iteration. Not sure this really qualifies as a "bug" so much as a best practice to generate consistent code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,12 @@
   ([Amr Kadry](https://github.com/Amrkadry) and
   ([Louis Pilfold](https://github.com/lpil))
 
+- The ordering of import statements in generated JavaScript code is now stable.
+  Previously the same code compiled twice could produce different output, as
+  the singleton constants imported for a type's fieldless variants were emitted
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -247,6 +247,11 @@
   actually required.
   ([John Downey](https://github.com/jtdowney))
 
+- Fixed a bug where on the JavaScript target compiling the same code twice
+  could produce different output, as the singleton constants imported for a
+  type's fieldless variants were emitted in a different order on each run.
+  ([John Downey](https://github.com/jtdowney))
```
lpil: Not sure this counts as a bug, I'll move it to the compiler section

lpil (approved):
Oh nice catch!! Thank you

PR #6228, opened 2026-08-27
use `#[expect(...)]` instead of `#[allow(...)]`

### LINKED ISSUE #6227: internal: use `#[expect(...)]` instead of `#[allow(...)]`
### Discussed in https://github.com/gleam-lang/gleam/discussions/6226

<sup>Originally posted by **ankddev** August 27, 2026</sup>
`#[allow(...)]` is good, but it has disadvantage of not warning when it becomes redundant. I think that `#[expect(...)]` would be better here, because Clippy gives warning (or error in CI when setting it to deny warnings) for unused ones, so we don't leave redundant ignores across the codebase. See https://doc.rust-lang.org/reference/attributes/diagnostics.html#r-attributes.diagnostics.lint.level for more information about it. In local branch I found several redundant ones. I would like to open a PR if it's a wanted thing.</div>

ankddev:
* Closes https://github.com/gleam-lang/gleam/discussions/6226
* Closes https://github.com/gleam-lang/gleam/issues/6227
***

### DIFF
--- a/compiler-cli/src/add.rs
+++ b/compiler-cli/src/add.rs
@@ -68,7 +68,7 @@ pub fn command(paths: &ProjectPaths, packages_to_add: Vec<String>, dev: bool) ->
         );
 
         // False positive. This package doesn't use the indexing API correctly.
-        #[allow(clippy::indexing_slicing)]
+        #[expect(clippy::indexing_slicing)]
         {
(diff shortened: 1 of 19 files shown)


lpil (approved):
Thank you

PR #6229, opened 2026-08-27
Make compiler less eager to recompile stale modules

giacomocavalieri:
If a module depends on stale modules, but none of those had changes in its public api, then the module will not be recompiled. Recompiling the module is wasted effort: no new errors or warnings can appear. And the generated code won't change.

### DIFF
--- a/compiler-core/src/build/package_compiler.rs
+++ b/compiler-core/src/build/package_compiler.rs
@@ -6,8 +6,8 @@ mod tests;
 
 use crate::analyse::TargetSupport;
 use crate::ast::TypedModule;
-use crate::build::package_loader::CacheFiles;
-use crate::build::{ErlangOutput, module_erlang_name};
+use crate::build::package_loader::{CacheFiles, load_cached_module};
+use crate::build::{ApiFingerprint, ErlangOutput, module_erlang_name};
 
 use crate::error::{DefinedModuleOrigin, FailedModule, SkipReason, SkippedModule};
 
@@ -19,7 +19,7 @@ use crate::{
     Error, Result, Warning,
     ast::UntypedModule,
     build::{
-        Mode, Module, Origin, Outcome, SourceFingerprint, Target,
+        Mode, Module, Origin, Outcome, SourceFingerprint,
         elixir_libraries::ElixirLibraries,
         native_file_copier::NativeFileCopier,
         package_loader::{CodegenRequired, PackageLoader, StaleTracker},
@@ -199,9 +199,14 @@ where
 
         // Type check the modules that are new or have changed
         tracing::info!(count=%loaded.to_compile.len(), "analysing_modules");
-        let outcome =
-            PackageModulesAnalyser::new(&self, warnings, existing_modules, incomplete_modules)
-                .analyse(loaded.to_compile);
+        let outcome = PackageModulesAnalyser::new(
+            &self,
+            artefact_directory,
+            warnings,
+            existing_modules,
+            incomplete_modules,
+        )
+        .analyse(loaded.to_compile);
 
         let mut modules = match outcome {
             Outcome::Ok(modules) => modules,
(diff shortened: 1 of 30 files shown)


compiler-core/src/build/tests.rs
```diff
@@ -58,3 +64,429 @@ fn usable_build_tool_mix_then_rebar3() {
...
+        fingerprint(
+            r#"
+            pub const wibble = one
+            const one = two
+            const two = 2
+        "#
+        ),
+        fingerprint(
+            r#"
+            pub const wibble = 2
```
lpil, line 226: I didn't understand this test for a second. Huh, v cool!

compiler-core/src/build/tests.rs
```diff
@@ -58,3 +64,429 @@ fn usable_build_tool_mix_then_rebar3() {
...
+            r#"
+            pub fn wibble() -> Int { both_targets() }
+
+            @external(erlang, "wibble", "wobble")
+            @external(javascript, "wibble", "wobble")
+            fn both_targets() -> Int
+            "#
+        )
+    )
+}
```
lpil, line 492: Wow I kinda hate how much `assert_eq` looks like `assert_ne`. Thank you Rust

compiler-core/src/build/package_compiler.rs
```diff
@@ -326,6 +331,9 @@ where
                 codegen_performed: self.perform_codegen,
                 dependencies: module.dependencies.clone(),
                 fingerprint: SourceFingerprint::new(&module.code),
+                // TODO) this is computed in another place too, should we maybe
+                // add it as a field of the module?
```
lpil, line 335: TODO here!

giacomocavalieri: Oops! Do you think we should compute this once and save it rather than doing the work twice?

compiler-core/src/build/package_compiler.rs
```diff
@@ -562,11 +570,9 @@ pub enum StdlibPackage {
 
 /// A structure we use to hold data about used to analyse the packages of a
 /// module.
-struct PackageModulesAnalyser<'a> {
-    package_config: &'a PackageConfig,
-    target_support: TargetSupport,
-    target: Target,
-    ids: &'a UniqueIdGenerator,
+struct PackageModulesAnalyser<'a, 'package_compiler, IO> {
+    package_compiler: &'a PackageCompiler<'package_compiler, IO>,
```
lpil, line 574: This seems odd to me. Surely the analyser shouldn't have access to the whole compiler? Its domain is purely analysis.

compiler-core/src/build/package_compiler.rs
```diff
@@ -588,11 +594,18 @@ struct PackageModulesAnalyser<'a> {
...
     analysed_modules: Vec<Module>,
+
+    /// Keeps track of the module whose public api has changed since the last
+    /// build.
+    modules_with_new_public_api: HashSet<EcoString>,
 }
 
-impl<'a> PackageModulesAnalyser<'a> {
-    pub fn new<'package_compiler, IO>(
+impl<'a, 'package_compiler, IO: FileSystemWriter + FileSystemReader + CommandExecutor + Clone>
```
lpil, line 603: This is a HUGE increase in capabilities, but it doesn't seem like any of them are used?

lpil (commented):
It looks like the entire package compiler is being passed into the analyser, which breaks the levels of abstraction and responsibility we have in the codebase. Unless I'm reading it wrong it seems to be an easy fix as it is not actually used, and the small bits of package-related data can be passed in instead.

compiler-core/src/build/package_compiler.rs
lpil: Is there a reason not to?

giacomocavalieri: Not that I can think of! I've updated the code to compute it just once

PR #6233, opened 2026-08-27
Fix redundant comparison warning for floats that are zeroes

### LINKED ISSUE #6211: Wrong compiler warning for redundant zero comparison
Consider the following code:
```gleam
import gleam/io

pub fn main() -> Nil {
  case -0.0 == 0.0 {
    True -> io.println("True")
    _ -> io.println("False")
  }
}
```

When running this example via `gleam run`, then the following warning is produced:
```bash
warning: Redundant comparison
  ┌─ /home/jonas/Development/gleam/zero_comp/src/zero_comp.gleam:4:8
  │
4 │   case -0.0 == 0.0 {
  │        ^^^^^^^^^^^ This is always `True`

```

However, the printed output is "False".
This is probably due to the fact that starting from OTP 27, the behavior of zero comparison changed, see [here](https://erlangforums.com/t/in-erlang-otp-27-0-0-will-no-longer-be-exactly-equal-to-0-0/2586).
The changed behavior should be reflected in the compiler warning.

I used gleam version 1.18.1 and Erlang/OTP 28.

jackprogramsjp:

Closes https://github.com/gleam-lang/gleam/issues/6211


compiler-core/src/type_/tests/warnings.rs
```diff
@@ -4451,7 +4451,7 @@ fn float_literals_redundant_comparison_infinity() {
 
 #[test]
```
ankddev, line 4452: Could you link to the issue please, so future readers can read more about it?

lpil (approved):
Thank you!

PR #6236, opened 2026-08-28
replace toList with prepend calls

giacomocavalieri:
We already did that for lists with a tail, now we do it for all lists. This proves to be quite more efficient


lpil (approved):
Thank you!!

PR #6241, opened 2026-08-29
fix capture syntax elimination logic in formatting #6240

### LINKED ISSUE #6240: Formatter changes execution semantics by unsafely removing capture syntax on function calls
# Formatter changes execution semantics by unsafely removing capture syntax on function calls

## Reproducible Example
Gleam version: `1.18.1`
OS: Linux 7.1.8-arch1-3 x86_64

```gleam
import gleam/io

pub fn main() {
  let get_parser = fn() {
    io.println("evaluated!")
    fn(x) { x }
  }

  // Intent: We want to delay the evaluation (lazy)
  // same semantics code
  // let lazy = fn(x) { get_parser()(x) } 
  let lazy = get_parser()(_)
}
```

## Steps to reproduce & Actual Behavior:

1. Run gleam run before formatting. Nothing is printed to the console (as expected).

2. Run gleam format. The code is reformatted to:

```diff
-  let lazy = get_parser()(_)
+  let lazy = get_parser()
```

3. Run gleam run again. It now prints `evaluated!`.

## Expected Behavior

The formatter should not remove the `(_)` when the expression being captured is a function call, as it alters the semantics of the code.

## Suggestion

I looked into the formatter source code and found the likely culprit in compiler-core/src/format.rs (around line 2059 in v1.18.1), inside the fn_capture function:

```rs
            // The capture has a single unlabelled hole:
            //
            //     wibble |> wobble(_)
            //     list.map([], wobble(_))
            //
            // We want these to become:
            //
            //     wibble |> wobble
            //     list.map([], wobble)
            //
            (
                FnCapturePosition::RightHandSideOfPipe | FnCapturePosition::EverywhereElse,
                [argument],
            ) if argument.is_capture_hole() && argument.label.is_none() => self.expr(arena, fun),
```

While this optimization is perfectly safe when fun is a simple identifier (like wobble), it is unsafe when fun is a Call expression (like get_parser()).

Suggestion:
I think we should update the condition in the match guard to only apply this simplification when the left-hand side (fun) is a simple variable/identifier, and explicitly prevent it when it is a function call.

Tom-game-project:

Fixes #6240

Hello! I've encountered a bug where the formatter removes the capture syntax, and I have opened Issue #6240 for this. 

I haven't received a reply on the issue yet, but I went ahead and created this PR because I already figured out the fix and wanted to provide a concrete code to discuss.

I haven't updated the CHANGELOG.md yet because I wasn't sure if this fix requires an entry, or how you prefer to format it. Please let me know if I should add it!

### DIFF
--- a/format/src/lib.rs
+++ b/format/src/lib.rs
@@ -2078,6 +2078,13 @@ impl<'a, 'doc> Formatter<'a> {
             panic!("Function capture body found not to be a call in the formatter")
         };
 
+        let flag = match &**fun {
+            UntypedExpr::Fn {arguments, .. } =>
+                arguments.iter().any(|a| a.is_capture_hole()),
+            UntypedExpr::Var { .. } => true,
+            _ => false
+        };
+
         match (position, arguments.as_slice()) {
             // The capture has a single unlabelled hole:
             //
@@ -2092,7 +2099,9 @@ impl<'a, 'doc> Formatter<'a> {
             (
                 FnCapturePosition::RightHandSideOfPipe | FnCapturePosition::EverywhereElse,
                 [argument],
-            ) if argument.is_capture_hole() && argument.label.is_none() => self.expr(arena, fun),
+            ) if argument.is_capture_hole() && argument.label.is_none() &&
+                flag
+                => self.expr(arena, fun),
 
             // The capture is on the right hand side of a pipe and its first
             // argument it an unlabelled capture hole:
--- a/format/src/tests.rs
+++ b/format/src/tests.rs
@@ -3673,6 +3673,17 @@ fn function_captures_test() {
 }
 "
     );
+
+    assert_format_rewrite!(
+        "pub fn main() {
+  run()(_)
+}
+",
+        "pub fn main() {
+  run()(_)
+}
+"
+    );
 }
 
 #[test]


format/src/lib.rs
```diff
@@ -2078,6 +2078,13 @@ impl<'a, 'doc> Formatter<'a> {
             panic!("Function capture body found not to be a call in the formatter")
         };
 
+        let flag = match &**fun {
```
ankddev, line 2081: Give it more descriptive name, please

format/src/lib.rs
```diff
@@ -2078,6 +2078,13 @@ impl<'a, 'doc> Formatter<'a> {
...
+            UntypedExpr::Fn {arguments, .. } =>
+                arguments.iter().any(|a| a.is_capture_hole()),
+            UntypedExpr::Var { .. } => true,
+            _ => false
```
ankddev, line 2085: Enumerate all possible variants please, this helps a lot when we add new variant to that enum.

format/src/tests.rs
```diff
@@ -3671,6 +3671,17 @@ fn function_captures_test() {
         "pub fn main() {
   run(1, 2, _, 4, 5)
 }
+"
+    );
+
+    assert_format_rewrite!(
```
ankddev, line 3677: You could use `assert_format!`, so you don't need to duplicate code string.

ankddev (commented):
Hello @Tom-game-project! Thank you! I left some comments inline. And yes please for changelog. I guess that it could be located in the bug fixes section, you could see other entries for inspiration and format.

lpil (commented):
This change is incorrect and is implementing a bug! Please do not open PRs for changes that have not been agreed to 🙏

PR #6246, opened 2026-08-31
Fingerprint every path dep in one pass

### LINKED ISSUE #6244: Dependency resolution performed a second time by commands run after `gleam deps download`
[repro.zip](https://github.com/user-attachments/files/31644432/repro.zip)

Here's an example of a project which does not compile correctly. To reproduce easily the bug, go in the `to_build` folder, and run `rm -rf build && gleam deps download && gleam test`. If it builds correctly the first time, wait for a few seconds, and redo the same command. You should end up with a Rate limit exceeded for the `gleam test` command.

---

From my understanding, `gleam test` should not trigger a rate limit here, because after `gleam deps downoald`, `gleam test` should not download at all deps from Hex. The first line from `gleam test` is _always_ `Resolving versions`. So `gleam test` always reach for Hex, which can lead to a Rate Limit.

jtdowney:
Closes #6244

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -268,6 +268,11 @@
   comparison warning, which is incorrect on Erlang/OTP 27+.
   ([Jack Programs](https://github.com/jackprogramsjp))
 
+- Fixed a bug where a project with multiple path dependencies would perform
+  dependency resolution once per path dependency, so commands run after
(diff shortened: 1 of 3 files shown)


lpil (approved):
Very clean! Excellent work!! Thank you

PR #6282, opened 2026-09-07
Do not emit unneeded loop variables in tail-recursive functions

### LINKED ISSUE #6273: Gleam could be smarter about the way recursive functions are transformed on the JavaScript target
For a recursive function, Gleam performs TCO itself, transforming the function into loops that look a lot like this:

```js
function factorial(loop$a, loop$b, loop$i, $loop$n) {
  while (true) {
    let a = loop$a;
    let b = loop$b;
    let i = loop$i;
    let n = loop$n;

    if (i >= n) {
      return b;
    } else {
      loop$a = b;
      loop$b = a;
      loop$i = i + 1;
      loop$n = n;
    }
  }
}
```

2 things could be improved here:

1. unnecessary assignments (`n`) could be dropped
2. temporary variables could only be introduced if they are required, i.e. if the recursive call itself refers to previous values out-of-order

Applying those to this factorial function results in the following code:

```js
function factorial(a, b, i, n) {
  while (true) {
    if (i >= n) {
      return b;
    } else {
      let loop$a = b;
      b = a;
      i = i + 1;
      a = loop$a;
    }
  }
}
```

I applied both optimisations as an experiment to the lustre runtime:

| Variant | Unminified | Minified | Minified + gzip |
|---|---:|---:|---:|
| Gleam 1.18 | 111,488 B | 31,635 B | 8,192 B |
| w/ optimisations | 95,049 B (-14.75%) | 29,759 B (-5.93%) | 7,825 B (-4.48%) |

I also tested performance in both v8 and jsc, but could not measure a difference. Inspecting the v8 node graph revealed that both are lowered to the same code by TurboFan.

For this experiment I did implement this very badly in the actual compiler, so if this is something that's wanted I could clean that up :)

~ 💜

yoshi-monster:
Unfortunately, this turned out to be a bit worse than anticipated as it's a bit offset in real Lustre programs by the small-list optimisation, and I didn't realise I chose a non-representative example in the original issue.

Compiler | Unminified | Minified | Minified+Gzip
-- | -- | -- | --
v1.18.0 | 163,424 B | 54,423 B | 17,574 B
smaller-tco | 147,093 B (−9.99%) | 52,368 B (−3.78%) | 17,204 B (−2.11%)

Here I compare the routing example from Lustre, were the previous comparison was the diff benchmark. The absolute improvements were roughly accurate.

Closes #6273

### DIFF
--- a/compiler-core/src/javascript/expression.rs
+++ b/compiler-core/src/javascript/expression.rs
@@ -6,13 +6,15 @@ use vec1::Vec1;
 
 use super::{decision::ASSIGNMENT_VAR, *};
 use crate::{
-    ast::*,
+    ast::{visit, *},
     exhaustiveness::StringEncoding,
     type_::{
         ModuleValueConstructor, Type, TypedCallArg, ValueConstructor, ValueConstructorVariant,
+        error::{VariableDeclaration, VariableOrigin},
     },
 };
 use pretty_arena::*;
+use std::collections::HashSet;
 use std::sync::Arc;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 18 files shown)


compiler-core/src/javascript/expression.rs
```diff
@@ -342,6 +499,8 @@ impl<'module, 'a, 'doc> Generator<'module, 'a, 'doc> {
         body: &'a [TypedStatement],
         arguments: &'a [TypedArg],
     ) -> Document<'a, 'doc> {
+        self.loop_variable_arguments =
+            FunctionParameterReferences::collect(&self.function_name, body);
```
giacomocavalieri, line 503: Mmh so this means we have to go over the whole function body twice now. Did you measure how this would affect compile times?

yoshi-monster: I did a bunch of benchmarks and I couldn't really measure a meaningful difference.

While this branch does appear to be 5% slower than 1.18 it is basically the same as its parent, so it seems like the slowdown comes from another change unrelated to this PR.

| Benchmark | v1.18 | parent | PR |
|---|---|---|---|
| Lustre, full rebuild | 237ms | 248ms | 249ms |
| Lustre, edit diff.gleam | 64ms | 68ms | 67ms |
| iv, full rebuild | 60ms | 67ms | 67ms |
| 10k `add1` functions | 120ms | 125ms | 125ms |
| 10k `length` functions | 244ms | 353ms | 348ms |


I'm as confused as you are! I only have an explanation for actually tail-recursive functions because less code is getting generated, but it seems something else in main is slowing it down.

I let the robot run a bisect to find the first commit and it claims <https://github.com/gleam-lang/gleam/commit/73d4d16e0093084ec67d4114243f5b970f428d21> is 40% slower at generating `case` expressions, which would make sense because every tail-recursive function has one of those. For the lustre rebuild, this seems to contribute ~33% of its slowdown. The robot also pointed at the change to the `bitcode` dependency for the remaining 2/3rds.

yoshi-monster: Edit: I looked a bit further and the problem is that `restore_user_variables` is quadradic (number of functions * avg names per function) in this case, since `extend` copies every value from the old map to the new map.

I think instead of extend it should be safe to restore the map by replacing it? None of the tests broke at least when I tried that :D

giacomocavalieri: Woah that should definitely be fixed somehow thank you so much for looking into this

lpil: Wow, amazing work Rebecca!

PR #6283, opened 2026-09-07
Remove unreferenced snapshots

yoshi-monster:
Hello!

I noticed while working on #6282 that a bunch of snapshots are no longer referenced, so I went ahead and ran
```shell
cargo insta test --workspace --unreferenced=delete
```
As far as I can tell, those are truly all not referenced, and all tests still pass.


lpil:
Huh, did you run this on Windows? Last time we ran that it removed the Windows-only snapshots.

yoshi-monster:
Yeah, originally it removed those 3 windows-only snapshots that I restored in the force-push:

- config::docs_prefix_windows_dot_dot_string_path
- config::docs_windows_dot_dot_string_path
- error::tests::io_link_file_error_windows

I did not run it on Windows locally, but I assumed if the CI passes it's all good :)

lpil (approved):
Ah so you did a little manual fix, that makes sense. Thanks!

PR #6287, opened 2026-09-09
Add support for hex.pm's LicenseRef licences

### LINKED ISSUE #6284: Support hex.pm's LicenseRef-* licence identifiers
Hex supports SPDX licence identifiers, but [it also supports non-SPDX licences](https://hex.pm/docs/rebar3-publish#adding-metadata-to-srcmyappappsrc). By prefixing a licence identifier with `LicenseRef-`, a user can upload a package with a non-SPDX licence. However, Gleam's command line tool will exit with an error if a user's gleam.toml specifies any non-SPDX licence:

```
$ gleam run -- --help
error: File IO failure

    /src/gleam-licence-example/gleam.toml

    TOML parse error at line 4, column 13
  |
4 | licences = ["LicenseRef-my-licence"]
  |             ^^^^^^^^^^^^^^^^^^
LicenseRef-my-licence is not a known SPDX License identifier
```

I'd like to request that the gleam command line tool support non-SPDX licences, either by prefixing them with `LicenseRef-`. This would have no implications on any other community-run package indexes (like packages.gleam.run) which could, if desired, still require packages to use a SPDX licence. It would simply fix the current situation which is that gleam applications and libraries are only able to use SPDX licences, and nothing else.

This could be implemented simply and without breaking changes by altering the serde deserializer function to check for a licence string starting with `LicenseRef-` https://github.com/gleam-lang/gleam/blob/main/compiler-core/src/config.rs#L65

Alternatively, the constraints on licence identifiers could be removed entirely, leaving it up to the individual developer to decide if they want to use SPDX, or `LicenseRef-` to be compliant with hex.pm, or let them use whatever identifier they like.

absolutely-vivid:
Closes #6284

I didn't add any tests for this because I didn't see any existing licence related tests. I can add one if desired though.

### DIFF
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -66,16 +66,18 @@ impl<'de> serde::de::Visitor<'de> for SpdxLicenseVisitor {
     where
         E: serde::de::Error,
     {
-        if value.starts_with("LicenseRef-") {
-            return Ok(SpdxLicense {
-                licence: value.to_string(),
-            });
-        }
-
         match spdx::license_id(value) {
-            None => Err(serde::de::Error::custom(format!(
-                "{value} is not a known SPDX License identifier"
-            ))),
+            None => {
+                if is_valid_license_ref(value) {
+                    Ok(SpdxLicense {
+                        licence: value.to_string(),
+                    })
+                } else {
+                    Err(serde::de::Error::custom(format!(
+                        "{value} is not a known SPDX License identifier"
+                    )))
+                }
+            }
             Some(_) => Ok(SpdxLicense {
                 licence: value.to_string(),
             }),
@@ -1243,6 +1245,16 @@ fn is_valid_package_name(name: &str) -> bool {
     PACKAGE_NAME_PATTERN
         .get_or_init(|| Regex::new("^[a-z][a-z0-9_]*$").expect("Package name regex"))
         .is_match(name)
+}
+
+static LICENSE_REF_PATTERN: OnceLock<Regex> = OnceLock::new();
+
+// Check that a license string is a valid SPDX LicenseRef, according to:
+// https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/
+fn is_valid_license_ref(licence: &str) -> bool {
+    LICENSE_REF_PATTERN
+        .get_or_init(|| Regex::new(r"^LicenseRef-[A-Za-z0-9\.-]+$").expect("LicenseRef regex"))
+        .is_match(licence)
 }
 
 #[test]
--- a/compiler-core/src/config.rs
+++ b/compiler-core/src/config.rs
@@ -66,6 +66,12 @@ impl<'de> serde::de::Visitor<'de> for SpdxLicenseVisitor {
     where
         E: serde::de::Error,
     {
+        if value.starts_with("LicenseRef-") {
+            return Ok(SpdxLicense {
+                licence: value.to_string(),
+            });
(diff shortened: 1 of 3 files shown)


compiler-core/src/config.rs
```diff
@@ -66,6 +66,12 @@ impl<'de> serde::de::Visitor<'de> for SpdxLicenseVisitor {
     where
         E: serde::de::Error,
     {
+        if value.starts_with("LicenseRef-") {
+            return Ok(SpdxLicense {
```
lpil, line 70: It looks like starting with `LicenseRef-` isn't sufficient, the remainder of the string is constrained to `/^[A-Za-z\.-]+$/`. Verify that too please 🙏 

https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/

```
idstring = 1*(ALPHA / DIGIT / "-" / "." )

license-ref = ["DocumentRef-"(idstring)":"]"LicenseRef-"(idstring)
```

lpil (commented):
Looks great, thank you. I've left a note inline, and you've forgotten to push the tests!

absolutely-vivid:
Thanks for the review! I added a regex to check for a LicenseRef according to the ABNF you linked. I also added a test for valid LicenseRef licences, and a test for invalid LicenseRef licences.

compiler-core/src/snapshots/gleam_core__config__invalid_license_ref.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: compiler-core/src/config.rs
+assertion_line: 1560
+expression: output
+---
+TOML parse error at line 4, column 13
+  |
+4 | licences = ["LicenseRef-my_invalid_licence"]
+  |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+LicenseRef-my_invalid_licence is not a known SPDX License identifier
```
lpil, line 10: British English spelling please, and it's not just licence identifiers that are accepted here, so let's be a bit more descriptive.

```suggestion
This is not a valid SPDX licence identifier or custom licence reference.

Custom licence references must start with `LicenseRef-` and contain only
alphabetical characters, dashes, and dots.
```

We could even use the "did you mean" code to suggest a correct licence id from the list of SPDX licences!

lpil (commented):
Nice, thank you. Fab tests too.

I've suggested a change to the error text 🙏

absolutely-vivid:
Thanks for the compliment!

I Renamed the tests to use british spelling, changed the error message to use your suggestion and add a "did you mean?" suggestion if applicable, and added a new test for that.

I did have to export the compiler_core::error::did_you_mean function as public. I assume that's not acceptable (at least without a prior discussion and a doc comment) so I've marked the PR as a draft.

If exporting the function is fine then I can write a doc comment for it, and if not then I can write a public compiler_core::error::suggest_licence function so that did_you_mean itself isn't exported. Or we could extract the majority of the did_you_mean code into a public helper function in another module. I'm not sure what would be the preferred option among those three.

lpil:
Good question! I can see arguments either way. Do you have a preference?

absolutely-vivid:
After thinking about it, I think the best option is to make the did_you_mean function public, and document it. If we're using it to suggest a license from a different module then it's useful enough to be a public function. Extracting the string distance code would be premature, and a suggest_licence wrapper wouldn't be useful enough on its own to justify a whole new function.

lpil (approved):
Thank you!!!

PR #6291, opened 2026-09-10
fix compile-package command

### LINKED ISSUE #6281: `gleam compile-package` no longer produces erlang files (nightly)
Hi!

`gleam compile-package` is used by other tools to integrate with Gleam. For this, it's quite common for them to search for the produced Erlang source code files, compiling them with their own respective toolchains.

Nightly removed those, producing `.abstr` files instead. This breaks all the tools relying on the output of `compile-package`. I think `compile-package` should continue to produce .erl files, and put `.abstr` files behind an optional flag, if at all.

~ 💜

giacomocavalieri:
Closes #6281 
This makes sure we're compiling to textual format when using the `compile-package` command

- [X] The issue for this PR has been linked above

### DIFF
--- a/compiler-cli/src/compile_package.rs
+++ b/compiler-cli/src/compile_package.rs
@@ -40,7 +40,7 @@ pub fn command(options: CompilePackage) -> Result<()> {
     let target = match options.target {
         Target::Erlang => TargetCodegenConfiguration::Erlang {
             app_file: None,
-            output: ErlangOutput::Binary,
+            output: ErlangOutput::Textual,
         },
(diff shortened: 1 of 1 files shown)


lpil (commented):
Can we have a test for this? To make sure we don't break it in future 🙏

giacomocavalieri:
Would it be a test running the compile-package command and checking the `.erl` file exists?

lpil:
Yeah I think that's the way to go

lpil (commented):
We never add new shell based tests! Please add to `test-commands` instead 🙏

lpil (approved):
Thank youuuuuuuu!!

PR #6303, opened 2026-09-13
Validate package names in `gleam remove`

### LINKED ISSUE #6278: Have `gleam remove` validate package names.
Thank you for wonderful language, I am enjoying it a lot.

I have installed a package to my project, played with it a little bit and as I found out that I do not need it I wanted to remove it. Somehow naturally I have searched history for the installation command and just changed the subcommand `install` to `remove`. To my surprise it told me that the package is not a dependency. Soon I have realized that I need to specify the package without the version constraint.

I believe the CLI should accept the `package@version` variant and just remove the `package` (if I understand it correctly that we cannot have 2 versions of the same package installed at the same time).

Alternatively the CLI should explicitly reject the `@version` variant and inform the user that they need to specify the package without the version.

Failing silently seems misleading and can cause confusion.

```
➜  gleavent git:(main) ✗ gleam remove gleam_otp@1
error: Package not found

These packages are not dependencies of your package so they could not
be removed.

  - gleam_otp@1


➜  gleavent git:(main) ✗ gleam remove gleam_otp  
    Removed gleam_otp
```

My env:
```
MacOS 26.6.2
gleam 1.18.1
```

tomvoet:
Closes #6278

This adds validation to package names for `gleam remove`.
Package name validation is done first, so it fails fast in case of a malformed name.
If a version is included, like described in #6278, the error looks like the following:

```
$ gleam remove gleam_otp@1 lustre@4 Wibble
error: Invalid package names

These package names are invalid so they could not be removed.

  - gleam_otp@1
  - lustre@4
  - Wibble

Did you mean one of these:

  - gleam_otp
  - lustre

Package names must start with a lowercase letter and may only contain
lowercase letters, numbers and underscores.
```

I thought about increasing the scope of the suggestions to include things like:
- `gleam-otp` to `gleam_otp`
- `Wibble` to `wibble`

But I left those for now, as I didn't want to get too carried away. 
Furthermore, another possible avenue for improvement could be checking against the project's actual dependencies.
These are things I could definitely add, but of course only if this is something that's actually wanted.

I felt a bit ambivalent regarding the pluralization, there are a few error messages that have this and others that don't. 
If you have any specific flavour you like more I'd be happy to make the corresponding changes. (see `gleam_core__error__tests__removed_package_names_invalid_without_suggestions.snap` for singular)

Lastly: Currently, the tests cover the three error message branches. Since there is no command-level test on `gleam remove` I decided not to add one, happy to add coverage to that as well though.

PS.: Thanks for the wonderful language, I've been using it for around 2 years and I've been especially enjoying gleam_otp >= 1 💜


compiler-core/src/error.rs
```diff
@@ -1327,6 +1331,61 @@ If you want to overwrite {text_files}, delete {text_pronoun} and run the command
...
+            Error::RemovedPackageNamesInvalid { packages } => {
+                let (plural, introduction) = if packages.len() == 1 {
+                    (
+                        "",
+                        "This package name is invalid so it could not be removed.",
+                    )
+                } else {
+                    (
+                        "s",
+                        "These package names are invalid so they could not be removed.",
```
lpil:
```suggestion
                        "This is not a valid package:",
                    )
                } else {
                    (
                        "s",
                        "These are not valid package names:",
```

compiler-core/src/error.rs
```diff
@@ -1327,6 +1331,61 @@ If you want to overwrite {text_files}, delete {text_pronoun} and run the command
...
+                    )
+                };
+
+                let suggestions: Vec<_> = packages
+                    .iter()
+                    .filter_map(|package| {
+                        let (name, _version) = package.split_once('@')?;
+                        is_valid_package_name(name).then_some(name)
+                    })
+                    .collect();
```
lpil: Can you use the existing code for selecting "did you mean" suggestions please 🙏 

We can select from the names of the existing dependencies.

lpil (commented):
Nice, thank you!! I've left some small notes inline.

compiler-core/src/error.rs
tomvoet: Done. One "drawback" though: when a project has only one dependency, `did_you_mean` always suggests it, no matter the name difference. So `gleam remove Wibble` in a new project gives you "Did you mean `gleam_stdlib`?". 
I'd say it's not too important since removing anything other than the only installed package doesn't make sense either. I can change this behaviour if you want tho :)

lpil: That's the desired behaviour, aye! It suggests the only option when there is only one, because that's the only possibility.

compiler-core/src/error/tests.rs
```diff
@@ -72,6 +72,34 @@ fn test_output_files_already_exist() {
...
+            ],
+        ),
+        ("without_suggestions", vec!["Wibble".to_string()]),
+    ];
+    for (group, packages) in packages {
+        let err = Error::RemovedPackageNamesInvalid {
+            packages,
+            dependencies: dependencies.clone(),
+        };
+        assert_snapshot!(
```
lpil, line 96: Only one test per test function please, and a maximum of one snapshot. Please split this into multiple functions, which will also make the test easier to understand. It is a bit daunting currently!

lpil (commented):
Thank you! I've pushed a commit to change the error format to be inline with the existing errors, and left a comment inline on the tests. Once that's sorted this can be merged 💜

PR #6307, opened 2026-09-14
Fix/compiler capture bug

### LINKED ISSUE #6242: Compiler Bug: A captured function is evaluated lazily instead of strictly
# Compiler Bug: A captured function is evaluated lazily instead of strictly

Related issues #6240

## Reproducible Example
Gleam version: `1.18.1`
OS: Linux 7.1.8-arch1-3 x86_64

```gleam
import gleam/io

pub fn main() {
  let get_parser = fn() {
    io.println("evaluated!")
    fn(x) { x }
  }

  // these two lines currently behave the same
  //let lazy = fn(x) { get_parser()(x) }
  let lazy = get_parser()(_)
  // This evaluate `get_parser` four times
  lazy(1)
  lazy(1)
  lazy(1)
  lazy(1)
}
```

## Actual

```sh
evaluated!
evaluated!
evaluated!
evaluated!
```

## Expected

```sh
evaluated!
```

```gleam
let lazy1 = get_parser()
let lazy2 = get_parser()(_)
```

`lazy1` behaves the same as `lazy2` instead

Tom-game-project:
Closes #6242

## Overview

In both the Erlang and JavaScript code generators, when handling TypedExpr::Fn during expression codegen, It accidentally outputs code that runs lazily. To resolve this, for problematic cases, the compiler now emits code that assigns <expr> to a temporary variable to evaluate it first before returning the function.

## Additional Test

I added a test using `echo` to verify that the function is actually called. Without this fix, it wouldn't print anything.


test-output/cases/function_capture_side_effect/src/main.gleam
```diff
@@ -0,0 +1,29 @@
...
+  // UntypedExpr::FieldAccess
+  let _v = func().f(_)
+
+  let func = fn() {
+    echo "hello"
+    #(Nil, fn(x) { x })
+  }
+  // UntypedExpr::TupleIndex
+  let _v = func().1(_)
+}
```
lpil, line 35: Good idea to have tests like this!

Could you please extract each test case into its own function (each called from `main`), change `"hello"` to something more descriptive that is immediately clear that it should be printed + which of the 3 it is, add another echo to each function that should not be printed, remove the comments that reference private compiler internals, add comments that explain what this test is doing and why, and add a link to the issue please 🙏

lpil (commented):
Thank you!!

I'm not sure that solving this twice in each generator is the best approach. Perhaps we could instead rewrite the tree during analysis to regular anonymous functions, and then remove all special knowledge inside the generators of the capture expression. That seems a lot nicer.

The only thing we would need to be careful about is preserving the positional information in the AST so hover etc continues to work correctly in the LSP.

Could we get some regular JS and Erlang snapshot tests too please.

Tom-game-project:
Thank you for the review!

I agree, and I had the same concern while implementing this.

In my initial local experiment, I tried rewriting the Untyped AST right after parsing. However, as pointed out in [this comment](https://github.com/gleam-lang/gleam/issues/6242#issuecomment-5500695307), that approach broke the formatter and wasn't the right place to do it.

Since the underlying transformation itself works, moving that rewrite to an appropriate layer during analysis makes a sense. Before diving into the implementation, could you point me to where that layer would ideally be? Since this is my first contribution to the Gleam compiler, I've been reading through `./docs/compiler/README.md` to navigate the codebase—where does this transformation best fit in relation to the pipeline described there?

test-output/cases/function_capture_side_effect/src/main.gleam
Tom-game-project: I changed the echo output from the unrelated "hello" to "eagerly n" to explicitly reflect the expected eager evaluation.

To test cases where echo should not be executed, I added a message on line 92 so that any difference between lazy and eager evaluation becomes visible.
This allows us to verify that the expression is evaluated once when the function capture is defined, and not re-evaluated during subsequent calls.

Tom-game-project:
I'll try implementing this in ./compiler-core/src/type_/expression.rs (likely within infer_fn), as other transformations like pipe desugaring seem to be handled around there.

PR #6320, opened 2026-09-17
Add byte indexing and an ASCII fast path to the lexer

jtdowney:
I reworked the lexer to use byte indexing with an ASCII fast path. Names, numbers, comments, and string contents now use source slices where possible, avoiding intermediate strings.

CRLF and lone carriage returns still normalize to newlines. This also fixes a bug with the final token's end position when the source ends with a multi-byte character. I can't imagine this bug has ever been impactful.

I added benchmarks for the standard library, emojindex (something with a lot of Unicode), and squirrel (something with large multiline strings). They use `dhat` to measure allocations and `criterion` to measure lexer speed. I didn't add it to CI, but you can run them manually with `cargo bench -p gleam-benches`.

It won't make a notable difference to overall compile times, but it definitely cleans up a lot of unnecessary allocations.

| Package | Lexer time | Allocations | Bytes allocated | Peak bytes live |
| --- | ---: | ---: | ---: | ---: |
| gleam_stdlib | 0.881 ms (-32%) | 2,513 (-87%) | 138 KB (-73%) | 345 (-4%) |
| emojindex | 5.534 ms (-35%) | 43,547 (-78%) | 1.434 MB (-69%) | 352 (-96%) |
| squirrel | 0.809 ms (-32%) | 2,352 (-90%) | 81 KB (-81%) | 1,391 (-60%) |


lpil (approved):
Oh this is so nice!!! What a great idea! Thank you

PR #6324, opened 2026-09-19
make compiler faster

giacomocavalieri:
All the investigation work was done by @yoshi-monster! We were copying all the keys over when it would have been enough to just reuse the old map https://github.com/gleam-lang/gleam/pull/6282#discussion_r4053033111


lpil (commented):
Keep the method interface instead of making internals public please 🙏

giacomocavalieri:
Done!

lpil (approved):
Thank you!!

PR #6325, opened 2026-09-19
Log Hex OAuth refresh rejection details

jtdowney:
I want to investigate further when people report issues with Hex requiring logins more often than expected, but the error details aren't being recorded. This change logs the error details under trace, so we'll have more to work with.


lpil (commented):
Nice! Any reason not to print this in the regular error? In case the user doesn't have logging enabled?

That aside, what issue are you working on? I thought that was no longer happening since Hex fixed their session expiry bug.

jtdowney:
> Nice! Any reason not to print this in the regular error? In case the user doesn't have logging enabled?

I'm certainly open to that. I hadn't really given much thought to what the UI would look like, because my thought was just to log it out. Let me set this as a draft and think it through.
 
> That aside, what issue are you working on? I thought that was no longer happening since Hex fixed their session expiry bug.

I wasn't aware Hex had fixed a bug. Do you have a link for that? I actually hit a random logout yesterday, which inspired me to open a PR for it.

PR #6329, opened 2026-09-21
Some smaller code quality improvements

alii:
Hello! This is a small (& very slightly drive-by) PR that makes some small code quality improvements. I used a homegrown linter that finds places in a Rust codebase where one could rely on Rust's type system more for correctness and better modelling of the domain.

Each small change is broken down into a single commit which I hope will make this easier to review, although overall this is a small PR. The 

1. 3afd64a1838728e5da811649f16ee2ee77839732 Simple, removes an exact duplicate copy of a function
2. 4cddc28a7ab7c92ca63cef8d8ac40c60cac7d2a1 Swap operand names that behaved correctly today, but were named incorrectly
3. 03852d0852775451e93811e20d49866e64147f81 We can avoid an error that will never happen with `Utf8PathBuf::from` instead of `Utf8PathBuf::from_str`
4. 6db75160c1e70a86774b34f2b4b78e9e2e67a97f Better represent the `parameters_location` property of `AddMissingTypeParameter` instead of having an extra flag that could be true even without any parameter source span


language-server/src/code_action.rs
```diff
@@ -12213,16 +12213,20 @@ pub struct AddMissingTypeParameter<'a> {
     params: &'a CodeActionParams,
     edits: TextEdits<'a>,
     /// The source location where the parameters should be defined.
-    /// This might be a zero-length span if there are no parameters yet,
-    /// or it might cover the already existing type parameter definitions.
-    parameters_location: Option<SrcSpan>,
-    /// If the type definition already had existing parameters before.
-    has_existing_parameters: bool,
+    parameters_location: Option<TypeParametersLocation>,
```
lpil, line 12216: I don't understand this change, please undo it.

lpil (commented):
Thank you, I've left one note inline!

PR #6330, opened 2026-09-21
Dedupe exact repeated matches

alii:
This is a followup to #6329 which removes exact duplicate match expressions, so that we no longer risk them falling out of sync if, for example, a contributor updates one and forgets to update any others.

This was not included in #6329 because it is in my opinion lower value, and is a bigger/noisier diff, making it harder to review.


compiler-core/src/type_/expression.rs
```diff
@@ -4956,6 +4956,32 @@ enum StaticComparison {
     CantTell,
 }
 
+/// Compares each pair of expressions: the result is only `CertainlyEqual` if
+/// every pair is.
+fn static_compare_pairwise<'a>(
```
lpil, line 4961:
```suggestion
fn static_compare_all<'a>(
```

compiler-core/src/type_.rs
```diff
@@ -1511,6 +1521,14 @@ impl Deprecation {
     pub fn is_deprecated(&self) -> bool {
         matches!(self, Self::Deprecated { .. })
     }
+
+    /// The deprecation message, or an empty string if it's not deprecated.
+    pub fn message(&self) -> &str {
```
lpil, line 1526: Never use an empty string to mean absence. Always use the `Option` type.

erlang-generation/src/lib.rs
```diff
@@ -3474,6 +3471,14 @@ impl ErlangSourceBuilder {
         }
     }
 
+    /// Starts a new pattern or expression, depending on the kind of list.
+    fn new_of_kind(&mut self, list_kind: &ListKind) {
```
lpil, line 3475: I don't think this make function makes much sense. The name is very general but it's specific to lists. I can't think of a better design, so let's remove this one.

pretty-arena/src/lib.rs
```diff
@@ -558,6 +558,17 @@ pub enum NestMode {
...
+impl NestMode {
+    /// The indentation to use when nesting a document by `amount`, when the
+    /// current indentation is `indent`.
+    fn apply(self, indent: isize, amount: isize) -> isize {
+        match self {
+            NestMode::Increase => indent + amount,
+            NestMode::Set => amount,
+        }
+    }
+}
```
lpil, line 570: This makes the code less clear, please undo this change.

lpil (commented):
Hello! This PR is largely fine (I've left some notes inline), but I'd like to touch on this statement:

> so that we no longer risk them falling out of sync if, for example, a contributor updates one and forgets to update any others.

Multiple things having the same code structure does not mean they are the same and need to be changed the same way in future, but coupling them together means they are forced to be. This is only beneficial if the rules of the domain mean they have to always be the same, otherwise it's coupling (which makes code less flexible) and abstraction (which most often makes code harder to understand).

PR #6335, opened 2026-09-22
Use forward slashes for paths in Hex package metadata

jtdowney:
While investigating Hex packages for a different project I discovered a few that shipped metadata.config files that didn't parse with `file:consult/1`. One was a Gleam package published from Windows, and the file list used forward slashes instead of backslashes. When Erlang reads the term, it sees a string with escape codes and chokes on unknown ones like `<<"src\caldav\internal\xml_request.gleam">>`. 


lpil (commented):
Great find! Thank you!!

I think you forgot to push the tests?

jtdowney:
I updated the existing test to show the bug. The existing test built the path with a forward slash embedded instead of using the join method which uses the platform native slash. With that change, the test continued to pass on Unix but failed on Windows.

lpil (approved):
Oh sorry! I misread that part of the code. Thank you!

